Issue #866, opened 2019-02-05
LinkAll causes NullReferenceException in AssemblyUtils.GetAssemblyFileVersionAttribute method


vikram-ma:
**Which Version of MSAL are you using ?**
MSAL 2.7
(Same issue happens with MSAL 2.6.2 as well)

**Platform**
Xamarin.iOS

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**
This is a new app, planning to release soon with linking enabled to reduce binary size.

**Repro**
1. Add MSAL reference (version 2.7 or 2.6.2, haven't tested other versions) to Xamarin application.
2. Create PublicClientApplication class

```csharp
var pca = new PublicClientApplication(clientId)
    {RedirectUrl=$"msal{clientId}://auth"};
```

**Expected behavior**
Should work with as before with full linking enabled.

**Actual behavior**
PublicClientApplication constructor fails and throws exception.

**Note**
After Link All is enabled, even if Microsoft.Identity.Client is linking is skipped (using LinkDescription file) it still crashes.
Link Framework SDKs only and Don't Link option works.

**Possible Solution**
Not sure

**Additional context/ Logs / Screenshots**
Stacktrace
```
at Microsoft.Identity.Client.Utils.AssemblyUtils.GetAssemblyFileVersionAttribute () [0x00019] in <33f2914531b0404fa49b0e603071a533>:0
 at Microsoft.Identity.Client.ClientApplicationBase..ctor (System.String clientId, System.String authority, System.String redirectUri, System.Boolean validateAuthority, Microsoft.Identity.Client.Core.IServiceBundle serviceBundle) [0x0009c] in <33f2914531b0404fa49b0e603071a533>:0
 at Microsoft.Identity.Client.PublicClientApplication..ctor (Microsoft.Identity.Client.Core.IServiceBundle serviceBundle, System.String clientId, System.String authority) [0x0000e] in <33f2914531b0404fa49b0e603071a533>:0
 at Microsoft.Identity.Client.PublicClientApplication..ctor (System.String clientId, System.String authority) [0x00000] in <33f2914531b0404fa49b0e603071a533>:0
[...]

PR #1210, opened 2019-06-07
Stop logging file version - fix for #866

bgavrilMS:
Assembly version is always the same with file version and informational version in MSAL. No point in logging all of them - it's slow and it causes issues like #866


jmprieur:
is it related to supportability?

bgavrilMS:
Passing UI Automation build https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1210

PR #1212, opened 2019-06-11
[DO NOT MERGE / WIP] Integrate WAM with UWP apps

MarkZuber:


henrik-me:
nit: There are a few things that bothers me about the inheritance and call patterns e.g. 
IPublicClientApplicationExecutor executor = new WamPublicClientExecutor(
                wamPublicClientApplication.ServiceBundle,
                wamPublicClientApplication);
and what follows from these changes to the ctr update to use IAppConfig.
Assuming those things are work in progress and thus not adding more details now.

src/Microsoft.Identity.Client/WamPublicClientApplication.cs
```diff
@@ -0,0 +1,167 @@
...
+        // TODO(WAM): implement these
+        public Task<IAccount> GetAccountAsync(string identifier) => throw new NotImplementedException();
+        public Task<IEnumerable<IAccount>> GetAccountsAsync() => throw new NotImplementedException();
+        public Task RemoveAsync(IAccount account) => throw new NotImplementedException();
+
+        public AcquireTokenByIntegratedWindowsAuthParameterBuilder AcquireTokenByIntegratedWindowsAuth(IEnumerable<string> scopes) => throw new NotImplementedException();
+        public AcquireTokenByUsernamePasswordParameterBuilder AcquireTokenByUsernamePassword(IEnumerable<string> scopes, string username, SecureString password) => throw new NotImplementedException();
+
+        #region Unsupported With WAM
+        public AcquireTokenWithDeviceCodeParameterBuilder AcquireTokenWithDeviceCode(IEnumerable<string> scopes, Func<DeviceCodeResult, Task> deviceCodeResultCallback) => throw new NotImplementedException();
```
henrik-me: >throw new NotImplementedException(); [](start = 171, length = 36)

Should we use the obsolete attribute instead to show a compiler error instead of a run-time error? I suppose we may have to throw a platformnotimplementedexception to follow the pattern used elsewhere allowing developers to work with interfaces instead of the concrete classes.

src/Directory.Build.props
```diff
@@ -144,6 +144,7 @@
     <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.3.0" />
     <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
     <PackageReference Include="System.Runtime.Serialization.Json" Version="4.3.0" />
+    <PackageReference Include="System.Runtime.WindowsRuntime" Version="4.3.0" />
```
henrik-me: ><PackageReference Include="System.Runtime.WindowsRuntime" Version="4.3.0" /> [](start = 4, length = 76)

is this needed?

src/Directory.Build.props
```diff
@@ -84,6 +83,7 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop)'">
     <DefineConstants>$(DefineConstants);DESKTOP</DefineConstants>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
henrik-me: ><TargetFrameworkVersion>v4.5</TargetFrameworkVersion> [](start = 4, length = 53)

is this needed?

henrik-me:
Looks interesting :)

src/Microsoft.Identity.Client/WamPublicClientApplication.cs
jmprieur: this is not a public class ? so the obsolete attribute does not matter much here? (or do I mis-understand something?)

PR #1216, opened 2019-06-13
Adding java to cache compat testing

MarkZuber:


MarkZuber:
The code infra is in place for running Java tests now but the java tests are currently failing locally.  Need to work with Petro/Santiago to get that resolved.  Once it's working locally, I'll update the build definition to include the proper java/maven calls for bringing in the library and then we can do a final review of this for checkin.

Issue #1196, opened 2019-06-03
Confidential client specifying Instance in ConfidentialClientApplicationOptions fails to retrieve OIDC metadata and acquire tokens


jelledruyts:
**Which Version of MSAL are you using ?**
MSAL.NET 3.0.8

**Platform**
netcore22

**What authentication flow has the issue?**
* Web App
    * [X] Authorization code
    * [X] OBO
* Web API
    * [X] OBO

**Is this a new or existing app?**
This is a new app

**Repro**

```csharp
var options = new ConfidentialClientApplicationOptions
{
    ClientId = "<client-id>",
    ClientSecret = "<client-secret>",
    TenantId = "<tenant-id>",
    Instance = "https://login.microsoftonline.com/"
};
var app = ConfidentialClientApplicationBuilder.CreateWithApplicationOptions(options).Build();
var result = app.AcquireTokenForClient(new [] { "https://graph.microsoft.com/.default" }).ExecuteAsync().Result;
```

**Expected behavior**
Get a token result.

**Actual behavior**
Error: "AADSTS90002: Tenant 'v2.0' not found. This may happen if there are no active subscriptions for the tenant. Check with your subscription administrator."
The URL being called to retrieve OIDC configuration is in fact missing the tenant:
GET https://login.microsoftonline.com//v2.0/.well-known/openid-configuration
Note the double slash before v2.0 where the tenant id should be.

**Possible Solution**
If you do not set the "Instance" property of the ConfidentialClientApplicationOptions, it actually works fine.

```csharp
var options = new ConfidentialClientApplicationOptions
{
    ClientId = "<client-id>",
    ClientSecret = "<client-secret>",
    TenantId = "<tenant-id>",
    // WORKAROUND - remove: Instance = "https://login.microsoftonline.com/"
};
var app = ConfidentialClientApplicationBuilder.CreateWithApplicationOptions(options).Build();
var result = app.AcquireTokenForClient(new [] { "https://graph.microsoft.com/.default" }).ExecuteAsync().Result;
```

Typically you will however bind the options to Configuration (in the "AzureAd" scope), which usually will have the Instance set. A workaround in that case to explicitly set the Instance property of the options object to null after having bound it to configuration.

```csharp
var options = new ConfidentialClientApplicationOptions();
Configuration.Bind("AzureAd", options);
options.Instance = null; // WORKAROUND
```

But of course this workaround should not be necessary at all.

[...]

PR #1224, opened 2019-06-18
Fix 1196 handling of Instance value in builder with trailing slash

MarkZuber:
Fixes https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1196

### DIFF
--- a/src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
+++ b/src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
@@ -345,6 +345,7 @@ namespace Microsoft.Identity.Client
 
             if (!string.IsNullOrWhiteSpace(Config.Instance))
             {
+                Config.Instance = Config.Instance.TrimEnd(' ', '/');
                 return Config.Instance;
             }
 
--- a/tests/Microsoft.Identity.Test.Unit.net45/AppConfigTests/ConfidentialClientApplicationBuilderTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit.net45/AppConfigTests/ConfidentialClientApplicationBuilderTests.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Drawing.Text;
 using System.Security.Cryptography.X509Certificates;
+using System.Security.Permissions;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Test.Common;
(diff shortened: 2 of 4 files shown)


tests/Microsoft.Identity.Test.Unit.net45/AppConfigTests/ConfidentialClientApplicationBuilderTests.cs
```diff
@@ -40,6 +42,55 @@ public void TestConstructor()
             Assert.AreEqual(null, cca.AppConfig.TenantId);
         }
 
+        private ConfidentialClientApplicationOptions CreateConfidentialClientApplicationOptions()
+        {
+            return new ConfidentialClientApplicationOptions
+            {
+                ClientId = MsalTestConstants.ClientId,
+                ClientSecret = "the_client_secret",
```
jennyf19: >the_client_secret" [](start = 32, length = 18)

nit: can you add these as constants?

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -345,6 +345,7 @@ private string GetDefaultAuthorityInstance()
 
             if (!string.IsNullOrWhiteSpace(Config.Instance))
             {
+                Config.Instance = Config.Instance.TrimEnd(' ', '/');
```
henrik-me: >Config.Instance = Config.Instance.TrimEnd(' ', '/'); [](start = 16, length = 52)

any reason we are not asking the developer to specify this correctly? exception?

MarkZuber: Our code later on assumes that this has no / for joining, and when we parse we split on "/".  So this is for us to not screw up.  It's legitimate syntax for the developer to supply the URI as https://login.microsoftonline.com/ or https://login.microsoftoneline.com (no trailing slash).  So this code just makes sure we do the right thing later on when combining these URI paths and parsing them.  

For example, Path.Combine actually solves this for file system paths and does the right thing around making sure there's only one / or \ delimiter.  But new Uri() does NOT do this and so we can end up with https://login.microsoftonline.com//tenantid and then when we parse out the tenant id (by doing a split on '/') we get the empty string for tenant. 

So, bottom line, I don't think we should throw an exception here, we're just keeping values tidy for later processing internally.

henrik-me: ok. thanks for explaining. 

---
In reply to: [295450204](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1224#discussion_r295450204) [](ancestors = 295450204)

Issue #1154, opened 2019-05-16
iOS acquiretokensilent test failing due to test app UI


trwalke:
**Which Version of MSAL are you using ?**
3.0.8

**Platform**
xamarin iOS

**What authentication flow has the issue?**
silent auth

**Is this a new or existing app?**
xamarin ios test app

**Repro**
run appcenter test with acquire token silent test enabled for iOS

**Expected behavior**
app should acquire a token for one resource to get a multi resource refresh token.
The app should then acquire a token for a different resource and receive a second access token

**Actual behavior**
Error is thrown saying "User was not selected"

**Possible Solution**

**Additional context/ Logs / Screenshots**
[LOG.txt](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/files/3189177/LOG.txt)
![Screenshot_1](https://user-images.githubusercontent.com/30090357/57890412-74f96780-77ec-11e9-9216-00e2aada9508.png)

jmprieur:
@trwalke is this issue fixed now? if yes, you might want to close it?

trwalke:
@jmprieur no, not fixed yet.

jmprieur:
@trwalke: isn't it highest priority to fix the failing test, than doing the new feature work?

jmprieur:
this is now fixed and available in MSAL.NET 4.1

Issue #1069, opened 2019-04-12
Fundamentals: Enable B2C UI automations on iOS


jennyf19:
The B2C UI automation tests on iOS are all ignored. They should be running w/the Android tests.

jmprieur:
@henrik-me. do we know why they are ignored?

jennyf19:
@jmprieur it's because the UI Ids are different on iOS then Android.

henrik-me:
Updating priority. Should be part of the next release after 4.0.

jennyf19:
Local accounts are not working due to bug on Xamarin.UITest w/iOS, as elements are placed on the same text field.

jmprieur:
this is now available with MSAL.NET 4.1

Issue #1026, opened 2019-03-28
Secure official Facebook account for testing


jennyf19:
Currently failing at `B2CFacebookMicrosoftAuthorityAcquireTokenTest` because we need to prove we are a user and not a bot.

jennyf19:
I am working w/Lab team on this.

jennyf19:
We have a lab account, needs testing.

jennyf19:
latest UI tests are targeting FB now w/the new user.

PR #1226, opened 2019-06-21
UIAutomation updates for iOS tests

jennyf19:
- Fix AT Silent on iOS. We were not selecting the user, so not sure how it was working on android...also, tapping"Done" here and there did the trick...will resolve this [issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1154)

- Rename B2C test methods so they are shorter and easier to read (IMO)
- Fix "Async" naming w/iOS tests
- Update some of the test ignore comments
- Include ROPC B2C test (i forgot to add "click the button" in the earlier Android tests) for Android and iOS

**Known issues**
- **B2C local account does not work on iOS**. There is a bug w/the Xamarin.UITest framework in that, sometimes, the username and password are put on the same field  in iOS, but not in Android. Am working w/support on finding a solution, as adding in waits and separating out the tapping from the text entering did not resolve the issue. Could also be a zoom issue. [Work item for b2c iOS tests](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1069)
image

**Next steps**
- B2C Facebook tests are ignored still. Just got an account from the lab this week, but need to try it out first. Will come in later PR to enable in iOS and Android. [Work item found here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1026)


tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -70,12 +70,10 @@ public void RunAllTests()
                 AcquireTokenADFSV2019InteractiveNonFederatedTest,
 
                 B2CLocalAccountAcquireTokenTest,
-                B2CROPCLocalAccountAcquireTokenTest
-                // Google Auth does not support embedded webview from b2clogin.com authority.
-                // App Center cannot run system browser tests yet, so this test can only be run in system browser locally.
-                //B2CGoogleB2CLoginAuthorityAcquireTokenTest,
-
+                B2CROPCLocalAccountAcquireTokenTest,
```
bgavrilMS: Is there any code path covered by this test that is not covered by the similar test you wrote for desktop? AFAIK ROPC is identical between all platforms where it is enabled, the only risks here are that we disable the ROPC entry point on mobile.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -251,7 +278,9 @@ public void AcquireTokenADFSV2019InteractiveFederatedTest()
         [Test]
         public void AcquireTokenADFSV3InteractiveFederatedTest()
         {
-            _mobileTestHelper.AcquireTokenInteractiveTestHelper(_xamarinController, LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV3).GetAwaiter().GetResult());
+            _mobileTestHelper.AcquireTokenInteractiveTestHelper(
```
bgavrilMS: Similar to my previous comment - do we expect tests hitting ADFSV3, ADFSv2 etc. to catch any issues? I'd say no because: 

- we have a mobile test hitting AAD, which tests the mobile web UI
- we have desktop tests hitting all the ADFS users

The only thing we would ever catch is EVO sending custom UI for mobiles and this UI being broken. @henrik-me - is this worth it?

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
```diff
@@ -74,8 +77,14 @@ public void AcquireTokenSilentTestHelper(ITestController controller, LabResponse
             AcquireTokenInteractiveHelper(controller, labResponse, CoreUiTestConstants.UiBehaviorLogin);
             VerifyResult(controller);
 
-            //acquire token for 2nd resource with refresh token
-            SetInputData(controller, labResponse.AppId, CoreUiTestConstants.DefaultScope, CoreUiTestConstants.UiBehaviorLogin);
+            controller.Tap(CoreUiTestConstants.SelectUser, XamarinSelector.ByAutomationId);
```
bgavrilMS: At some point, we should stop using a test app with a complex UI and adopt a simpler strategy, similar to how mobile unit tests exist: 

- the test app has a drop-down that shows a list of test names and a button to run the test
- the automation iterates through this drop-down and runs each test
- all configuration is in memory, rather than having a zillion of input boxes, switches, tabs and buttons. 

We can even take a step further and use the Xamarin.Test infrastructure for running unit tests, where the test logic is injected from the test to the app and the test runner handles exceptions, failures etc. 

@MarkZuber @henrik-me @trwalke - thoughts?

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
```diff
@@ -74,8 +77,14 @@ public void AcquireTokenSilentTestHelper(ITestController controller, LabResponse
...
+            controller.Tap(labResponse.User.Upn);
+
+            if (_platform == Xamarin.UITest.Platform.iOS)
+            {
+                controller.Tap("Done");
```
bgavrilMS: This seems to magical...

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
```diff
@@ -282,13 +290,29 @@ public void SetAuthority(ITestController controller, string authority)
...
 
         public void PerformB2CLocalAccountSignInFlow(ITestController controller, LabUser user, UserInformationFieldIds userInformationFieldIds)
         {
             controller.EnterText(CoreUiTestConstants.WebUpnB2CLocalInputId, 20, user.Upn, XamarinSelector.ByHtmlIdAttribute);
 
-            controller.EnterText(userInformationFieldIds.GetPasswordInputId(true), user.GetOrFetchPassword(), XamarinSelector.ByHtmlIdAttribute);
+            if (_platform == Xamarin.UITest.Platform.iOS)
+            {
+                controller.Tap(CoreUiTestConstants.B2CWebPasswordId, XamarinSelector.ByHtmlIdAttribute);
+                Thread.Sleep(1000);
```
bgavrilMS: magic

tests/Microsoft.Identity.Test.iOS.UIAutomation/iOSTests.cs
```diff
@@ -53,36 +52,38 @@ public void InitializeBeforeTest()
...
             {
-                AcquireTokenTestAsync,
-                //AcquireTokenSilentTest,
-                AcquireTokenADFSV3InteractiveFederatedTestAsync,
-                AcquireTokenADFSV3InteractiveNonFederatedTestAsync,
-                AcquireTokenADFSV4InteractiveFederatedTestAsync,
-                AcquireTokenADFSV4InteractiveNonFederatedTestAsync,
-                AcquireTokenADFSV2019InteractiveFederatedTestAsync,
-                AcquireTokenADFSV2019InteractiveNonFederatedTestAsync,
+                AcquireTokenTest,
```
bgavrilMS: This is not resetting internal caches such as instance discovery cache and OIDC cache.

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
trwalke: We had the idea to do something like this a while back but we never got around to doing it. The problem we have now is that we are using apps designed for customers to test out MSAL rather than apps designed specifically for testing. One great example of how to do this is with our old desktop automation app. This app had a handful of buttons to initiate different scenarios and if you click on one of these buttons, you were able to pass in a json string with all of the configuration properties for that scenario. This brought the complexity of that app way down as we only needed to focus on automating the web interactions rather than configuring the app with a series of buttons.

an example of what the configuration looked like:
{"authority":"https://login.microsoftonline.com/common","client_id":"3c1e0e0d-b742-45ba-a35e-01c664e14b16","redirect_uri":"urn:ietf:wg:oauth:2.0:oob","resource":"00000003-0000-0ff1-ce00-000000000000","prompt_behavior":"Always"}

MarkZuber: My $.02: The less UI we have to automate, the better.  And a button or drop list item per test makes it clear what the test breakdowns being executed.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
jennyf19: we have this test on desktop and in integration tests. I do see value in having it on mobile as we have run into the issue where we broke scenario on iOS or android and the UI tests picked it up. Since it's not complicated, i don't see the hurt in having it.

---
In reply to: [296148945](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1226#discussion_r296148945) [](ancestors = 296148945)

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
jennyf19: I like the idea of having a drop down. Can look into that at a later date, but not part of this PR.

---
In reply to: [296304990](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1226#discussion_r296304990) [](ancestors = 296304990)

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
jennyf19: this was advised by support, and this test is currently ignored, but I will remove the waits for now.

---
In reply to: [296153527](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1226#discussion_r296153527) [](ancestors = 296153527)

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
MarkZuber: @jennyf19  -- agree with "not this pr".  but forward thinking, simplifying our ui automation makes sense.

jennyf19: definitely.

---
In reply to: [296309507](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1226#discussion_r296309507) [](ancestors = 296309507)

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
jennyf19: ? on iOS the tabs are on the bottom, so if we have picker on the screen, we have to hit the done button in order to see the tabs again. 

---
In reply to: [296153282](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1226#discussion_r296153282) [](ancestors = 296153282)

bgavrilMS: Sorry, I know what this does and you've explained it to me. It's magical because it behaves differently on iOS than on Android and because it has absolutely nothing to do with our test. Most of the logic in the tests are about pressing buttons, filling forms and switching switches. It's impossible to understand what a test does by reading its source code, and it's also very difficult to maintain these tests.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
bgavrilMS: The hurt in having it is having to maintain it and the extra time it adds to the test run. But since you added it and it's low risk of causing issues, let's keep it.

tests/Microsoft.Identity.Test.iOS.UIAutomation/iOSTests.cs
bgavrilMS: This hasn't been resolved.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -46,6 +47,7 @@ public void InitializeBeforeTest()
             _app = AppFactory.StartApp(_platform, "com.Microsoft.XFormsDroid.MSAL");
             _xamarinController.Application = _app;
             _mobileTestHelper = new MobileTestHelper(_platform);
+            TestCommon.ResetInternalStaticCaches();
```
bgavrilMS: This is good, but for the "fast run" test, it will not call ResetInternalStatiCaches() before running each individual test.

jennyf19:
[integration tests passed here PR build](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=360146)
[UI tests passed here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=360152)

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
jennyf19: done.

tests/Microsoft.Identity.Test.iOS.UIAutomation/iOSTests.cs
jennyf19: done.

tests/Microsoft.Identity.Test.Common/InternalsVisibleTo.cs
```diff
@@ -6,4 +6,6 @@
...
 [assembly:InternalsVisibleTo("DesktopTestApp,         PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
 [assembly:InternalsVisibleTo("XForms,         PublicKey=00240000048000009400000006020000002400005253413100040000010001002D96616729B54F6D013D71559A017F50AA4861487226C523959D1579B93F3FDF71C08B980FD3130062B03D3DE115C4B84E7AC46AEF5E192A40E7457D5F3A08F66CEAB71143807F2C3CB0DA5E23B38F0559769978406F6E5D30CEADD7985FC73A5A609A8B74A1DF0A29399074A003A226C943D480FEC96DBEC7106A87896539AD")]
+[assembly: InternalsVisibleTo("Test.Microsoft.Identity.iOS.Msal.UIAutomation, PublicKey=00240000048000009400000006020000002400005253413100040000010001002d96616729b54f6d013d71559a017f50aa4861487226c523959d1579b93f3fdf71c08b980fd3130062b03d3de115c4b84e7ac46aef5e192a40e7457d5f3a08f66ceab71143807f2c3cb0da5e23b38f0559769978406f6e5d30ceadd7985fc73a5a609a8b74a1df0a29399074a003a226c943d480fec96dbec7106a87896539AD")]
```
henrik-me: Why are these needed?

henrik-me:
>     public class MobileTestHelper

Should there have been a MobileTestHelper for iOS and one for Android? avoiding switches? Perhaps with a factory that would create the right one?

---
Refers to: tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs:17 in 66bd1d4. [](commit_id = 66bd1d4731cc9fa7778f2a16d05ee294cfa308c3, deletion_comment = False)

tests/Microsoft.Identity.Test.Common/InternalsVisibleTo.cs
jennyf19: this is for accessing Test.Common to access the method which cleans out the instance dis cache

Issue #1211, opened 2019-06-10
Enhancement: ADAL to MSAL transition: setting Correlation Id.


msminski:
**Is your feature request related to a problem? Please describe.**
With ADAL, our current implementation of WebAuth library assigns our own correlation Id value to the AuthenticationContext instance before calling to AAD. This way, we could assure that one correlation Id is used throughout the flow. In our attempt to transition to MSAL, we've found out that ConfidentialClientApplication(or other related classes) class doesn't seem to have property CorrelationId anymore, while MsalServiceException class does have the property. 

**Describe the solution you'd like**
We'd like to know if this change(missing CorrelationId property in ConfidentialClientApplication class) is intended so that partners could only use the correlation ids came back from AAD side, or is this a feature that could be requested to be added back.

bgavrilMS:
AFAIK a correlation ID is generated for each request and:

- MSAL will log it
- MSAL will send it to AAD, and if AAD encounters an error it will also display the correlation ID

This should be sufficient to track problems across both MSAL (client side) and the Identity Provider (AAD). 

@jmprieur to confirm

jmprieur:
@msminski : can you describe your scenarios a bit more: do you need to trace things beyond authentication? 
you mention that you'd like the CorrelationId to be on ConfidentialClientApplication, do you really mean for the duration of the application? or would you need more something like .WithExternalCorrelationId(guid) on each AcquireTokenXX request?

msminski:
@jmprieur: we definitely meant to ask for the latter, something like .WithExternalCorrelationId(guid) on each AcquireTokenXX request. The purpose of this is to provide our partners an option to trace their scenarios that involve authentication with single correlation Id. Previously we achieved it by setting CorrealtionId in AuthenticationContext instance, but now that I think about how ConfidentialClientApplication is representing application duration, it would make more sense to make the second approach you mentioned above(WithExternalCorrelationId(guid)).

Thank you

jmprieur:
+ @MarkZuber  and @henrik-me  as maybe the right answer to the scenario is not the correlation id, but the telemetry API?

henrik-me:
@jmprieur : I believe it would be ok to allow developers to specify a correlation id. I don't think the answer here is in the telemetry.

jmprieur:
This makes sense @henrik-me.
The only question is: at the app building time? token aquisition time ? or both?

jmprieur:
Proposed spec

[...]

PR #1235, opened 2019-06-26
Implementing WithCorrelationId method

trwalke:
Implementation of [#1211](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1211)


src/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -328,6 +328,17 @@ public T WithB2CAuthority(string authorityUri)
             return (T)this;
         }
 
+        /// <summary>
+        /// Sets the correlation id to be used in the authentication request.
+        /// </summary>
+        /// <param name="correlationId">correlation id of the authentication request</param>
+        /// <returns></returns>
+        public T WithCorrelationId(Guid correlationId)
```
trwalke: @jmprieur @henrik-me Using GUID here for the correlation ID. Do we want to keep it as is or change it to a string input instead? The authenticationResult in this PR also contains a correlation id of type GUID

henrik-me: GUID is great

---
In reply to: [297839588](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1235#discussion_r297839588) [](ancestors = 297839588)

MarkZuber: MATS requires a guid for it's correlation ids, so my $.02 is that we make this a guid explicitly, as you have here.  That way we can format it appropriately for what the telemetry system and services are expecting (there's a AsMatsCorrelationId or something like that helper method in the .net mats telemetry code).

src/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
```diff
@@ -14,11 +14,11 @@ internal class AcquireTokenCommonParameters
 
         public ApiEvent.ApiIds ApiId { get; set; } = ApiEvent.ApiIds.None;
         public Guid TelemetryCorrelationId { get; set; }
+        public Guid UserProvidedTelemetryCorrelationId { get; set; }
```
henrik-me: >Telemetry [](start = 32, length = 9)

This is not necessarily a telemetry id... It's a CorrelationId which has been user provided. Pls. remove the word Telemetry

src/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -32,7 +32,7 @@ internal override async Task<AuthenticationResult> ExecuteAsync(CancellationToke
                 var msalAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
                 if (msalAccessTokenItem != null)
                 {
-                    return new AuthenticationResult(msalAccessTokenItem, null);
+                    return new AuthenticationResult(msalAccessTokenItem, null, AuthenticationRequestParameters.RequestContext.TelemetryCorrelationId);
```
henrik-me: is this intentional?

src/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -41,7 +41,7 @@ internal override async Task<AuthenticationResult> ExecuteAsync(CancellationToke
                 MsalAccessTokenCacheItem msalAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
                 if (msalAccessTokenItem != null)
                 {
-                    return new AuthenticationResult(msalAccessTokenItem, null);
+                    return new AuthenticationResult(msalAccessTokenItem, null, AuthenticationRequestParameters.RequestContext.TelemetryCorrelationId);
```
henrik-me: intentional?

src/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
trwalke: yes. the correlation ID needs to be made available on the AuthenticationResult so I pass it into the constructor

src/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
henrik-me: >UserProvidedTelemetryCorrelationId [](start = 20, length = 34)

This is the new one, seems like you have now started taking the param above and using that in a number of places. is that intentional?

src/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -98,7 +98,7 @@ internal override async Task<AuthenticationResult> ExecuteAsync(CancellationToke
                 if (msalAccessTokenItem != null)
                 {
                     var msalIdTokenItem = await CacheManager.GetIdTokenCacheItemAsync(msalAccessTokenItem.GetIdTokenItemKey()).ConfigureAwait(false);
-                    return new AuthenticationResult(msalAccessTokenItem, msalIdTokenItem);
+                    return new AuthenticationResult(msalAccessTokenItem, msalIdTokenItem, AuthenticationRequestParameters.RequestContext.TelemetryCorrelationId);
```
henrik-me: >TelemetryCorrelationId [](start = 137, length = 22)

?

src/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -340,7 +351,7 @@ internal void ValidateAndCalculateApiId()
             Validate();
             CommonParameters.ApiId = CalculateApiEventId();
             CommonParameters.ApiTelemId = ApiTelemetryId;
-            CommonParameters.TelemetryCorrelationId = Guid.NewGuid();
+            CommonParameters.TelemetryCorrelationId = (CommonParameters.UserProvidedTelemetryCorrelationId != null) ? CommonParameters.UserProvidedTelemetryCorrelationId : new Guid();
```
MarkZuber: you should use Guid.NewGuid() instead of "new Guid()".  I don't believe the Guid constructor actually calls CoCreateGuid()

src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -43,7 +44,8 @@ public partial class AuthenticationResult
             string tenantId,
             IAccount account,
             string idToken,
-            IEnumerable<string> scopes)
+            IEnumerable<string> scopes,
+            string correlationID)
```
MarkZuber: nit:  correlationId

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -1262,6 +1262,38 @@ public void EnsurePublicApiSurfaceExistsOnInterface()
             CheckBuilderCommonMethods(byRefreshTokenBuilder);
         }
 
+        [TestMethod]
+        public void CheckUserProvidedCorrlationIDTest()
```
henrik-me: >Corrlation [](start = 37, length = 10)

nit: missing e

src/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -141,6 +145,11 @@ internal AuthenticationResult(MsalAccessTokenCacheItem msalAccessTokenCacheItem,
         /// </summary>
         public IEnumerable<string> Scopes { get; }
 
+        /// <summary>
+        /// Gets the correlation id used for the request.
+        /// </summary>
+        public string CorrelationID { get; }
```
MarkZuber: nit:  CorrelationId  (.net naming conventions have ID as Id)

Do we want this to be TelemetryCorrelationId or is this more broad.  We're calling it TelemetryCorrelationId elsewhere.   
Should this also be a Guid type since that's how we're sending it in?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -64,7 +64,8 @@ public void MockConfidentialClientApplication_AcquireToken()
                 {
                     "scope1",
                     "scope2"
-                });
+                },
+                new Guid().ToString());
```
MarkZuber: Guid.NewGuid().   We should also always specify the format of the guid as we expect it.  Due to differences in how ToString() works on different platforms (some require IFormatProvider), there's a AsMatsCorrelationId() extension method in Microsoft.Identity.Client.TelemetryCore.Internal.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -1262,6 +1262,38 @@ public void EnsurePublicApiSurfaceExistsOnInterface()
...
         }
 
+        [TestMethod]
+        public void CheckUserProvidedCorrlationIDTest()
+        {
+            using (var harness = CreateTestHarness())
+            {
+                harness.HttpManager.AddInstanceDiscoveryMockHandler();
+                var correlationId = new Guid();
+                PublicClientApplication app = PublicClientApplicationBuilder.Create(MsalTestConstants.ClientId)
```
MarkZuber: nit:  please wrap these to the next line tabbed in 4 spaces to reduce width for code reviews:

PublicClientApplication app = PUblicClientApplicationBuilder
    .Create(foo)
    .WithAuthority(bar)
    .etc();

src/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
trwalke: yes. This is used to set the correlation id that will be used by the authentication request.

src/Microsoft.Identity.Client/AuthenticationResult.cs
trwalke: Yes. it will be a Guid.
I think CorrelationId is sufficient on its own as it is a more general term.
@henrik-me @jmprieur thoughts?

henrik-me: CorrelationId, is the right term imo. can be telemetry or not.

src/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
henrik-me: hmm, that is not what the user passed in. that seems odd to me

src/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
henrik-me: as mentioned elsewhere... this is the telemetrycorrelationid and not the new id. Also why are you creating this as a new Guid?

---
In reply to: [297841976](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1235#discussion_r297841976) [](ancestors = 297841976)

src/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
trwalke: I dont think so but perhaps I am mistaken? I checked the code and the TelemetryCorrelationId is the same as the CorrelationId. It seems like we have two different names for the same value in the code.

src/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
trwalke: We have two names for the same value so this is a little confusing. the telemetrycorrelationid and the correlationid are the same. Also, I create a new guid because we are exposing the correlation id as a GUID on the AuthenticationResult but the correlation id in the requestContext is exposed as a string so i have to make the conversion.

---
In reply to: [298004949](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1235#discussion_r298004949) [](ancestors = 298004949,297841976)

src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
```diff
@@ -98,7 +99,7 @@ private async Task AcquireAuthorizationAsync(CancellationToken cancellationToken
             AuthenticationRequestParameters.RedirectUri = _webUi.UpdateRedirectUri(AuthenticationRequestParameters.RedirectUri);
             var authorizationUri = CreateAuthorizationUri(true);
 
-            var uiEvent = new UiEvent(AuthenticationRequestParameters.RequestContext.TelemetryCorrelationId);
+            var uiEvent = new UiEvent(AuthenticationRequestParameters.RequestContext.CorrelationId.AsMatsCorrelationId());
```
henrik-me: >AsMatsCorrelationId() [](start = 99, length = 21)

why is this needed?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/ConfidentialClientApplicationTests.cs
henrik-me: hmm, still not sure I get this. 

---
In reply to: [297843566](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1235#discussion_r297843566) [](ancestors = 297843566)

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -1262,6 +1263,40 @@ public void EnsurePublicApiSurfaceExistsOnInterface()
             CheckBuilderCommonMethods(byRefreshTokenBuilder);
         }
 
+        [TestMethod]
+        public void CheckUserProvidedCorrelationIDTest()
+        {
+            using (var harness = CreateTestHarness())
+            {
+                harness.HttpManager.AddInstanceDiscoveryMockHandler();
+                var correlationId = new Guid();
```
MarkZuber: Guid.NewGuid()

MarkZuber: new Guid() always returns Guid.Null (00000-00000-00000-...)

trwalke: Missed this one. thanks

src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
trwalke: I changed the requestContext's correlation id to a Guid as it is the main mechanism in which the correlationId moves through the project. Some classes that use the correlation Id, such as the UiEvent class require a string for the correlation Id because its base type is Dictionary<string, string> and a Guid would incompatible since it cannot be added to the collection.

Issue #1148, opened 2019-05-14
Improve CA Error Handling


henrik-me:
### Is your feature request related to a problem? Please describe.
# Problem statement

One of common status codes returned from authentication libraries in silent mode today is InvalidGrant. This status code means that the application should call the authentication library again, but in interactive mode. Additional user interaction is required before authentication token can be issued. Note: some versions of libraries also return InteractionRequired status, with exactly the same semantics. The rest of the document will refer to InvalidGrant only, and same decisions apply to InteractionRequired.

Over time, many conditions accumulated in the broad category covered by InvalidGrant status code. As a result, InvalidGrant has become very general, and it is hard for applications to build user experience that correctly handles all the conditions that result in InvalidGrant. Some of those conditions are easy for users to resolve (e.g. accept Terms of Use with a single click), and some cannot be resolved with the current configuration (e.g. the machine in question needs to connect to a specific corporate network). 

Depending on how complicated and involved required user interaction is, apps may want to show different user experience for different levels of difficulty of user interaction. For example, if the app is trying to show multiple resources to the user at the same time (e.g. items in a collection returned by a search result), the app may choose to not display specific results for which authentication flow to resolve InvalidGrant condition is too intrusive, but may choose to enable resolution of InvalidGrant condition if resolution is quick and simple.

Applications today do not have a way to distinguish between different classes of conditions that cause InvalidGrant, and therefore only can have a very generic way to handle this state, which leads to end user confusion and user experience dead ends in some of application flows.

For more details and description of scenarios, please see [Improving CA Error Experience in Office.pptx](https://microsoft-my.sharepoint-df.com/:p:/p/nagupt/ETbeSzCWPuFAntdQshc084gBc4MpyG3rtJavntPs3fPwlg?rtime=40dSAjTF1kg)

# Requirements

1. Apps must be able to distinguish between several classes of InvalidGrant condition. See "API Changes" section for the detailed list.
1. Decouple complexity of various error conditions on the server from client apps. Server must retain capability to quickly iterate and add new authentication flows and conditions, without requiring client apps to be changed in any way.
1. Extensibility: provide a way for future classes of InvalidGrant condition to be communicated to apps without the need to service authentication libraries.

# Solution

## API Changes

MSALs will expose additional classification of InvalidGrant condition. This classification will be returned as a string, with the following meaning and recommended handling:

| Classification    | Meaning           | Recommended handling |
|-------------------|-------------------|----------------------|
[...]

PR #1236, opened 2019-06-26
Add "Classification" to UI Required Exception #1148

bgavrilMS:


src/Microsoft.Identity.Client/OAuth2/OAuthConstants.cs
```diff
@@ -83,6 +83,59 @@ internal static class OAuth2Error
         public const string AuthorizationPending = "authorization_pending";
     }
 
+    internal static class OAuth2SubError
+    {
+        /// <summary>
+        /// Condition can be resolved by user interaction during the interactive authentication flow.
+        /// </summary>
+        public const string BasicAction = "basic_action";
```
bgavrilMS: Unfortunately I don't know how to "repro" each of these, so I couldn't test E2E. Still working on this, but unit tests should give some confidence.

src/Microsoft.Identity.Client/MsalUiRequiredException.cs
```diff
@@ -14,6 +15,37 @@ namespace Microsoft.Identity.Client
...
+        /// </summary>
+        public const string UserPasswordExpired = OAuth2SubError.UserPasswordExpired;
+
+        /// <summary>
+        /// User consent is missing, or has been revoked.
+        /// See https://aka.ms/msal-net-uirequiredexception-classification for details
+        /// </summary>
+        public const string ConsentRequired = OAuth2SubError.ConsentRequired;
+
+
```
trwalke: nit: extra line

src/Microsoft.Identity.Client/MsalUiRequiredException.cs
```diff
@@ -40,6 +72,37 @@ public class MsalUiRequiredException : MsalServiceException
...
         public MsalUiRequiredException(string errorCode, string errorMessage, Exception innerException) : base(errorCode, errorMessage, innerException)
         {
+            switch (base.SubError)
+            {
+                case OAuth2SubError.BasicAction:
+                case OAuth2SubError.AdditionalAction:
+                case OAuth2SubError.MessageOnly:
+                case OAuth2SubError.ConsentRequired:
+                case OAuth2SubError.UserPasswordExpired:
+
```
trwalke: nit: are the extra lines here intentional?

tests/Microsoft.Identity.Test.Unit.net45/ExceptionTests/MsalExceptionTests.cs
```diff
@@ -3,15 +3,268 @@
...
+        private const string ExCode = "exCode";
+        private const string ExMessage = "exMessage";
+
+        private const string JsonError = @"{ ""error"":""invalid_tenant"", ""suberror"":""some_suberror"",
+            ""claims"":""some_claims"",
+            ""error_description"":""AADSTS90002: Tenant 'x' not found. "", ""error_codes"":[90002],""timestamp"":""2019-01-28 14:16:04Z"",
+            ""trace_id"":""43f14373-8d7d-466e-a5f1-6e3889291e00"",
+            ""correlation_id"":""6347d33d-941a-4c35-9912-a9cf54fb1b3e""}";
+
+
```
trwalke: nit: extra line

src/Microsoft.Identity.Client/MsalServiceException.cs
```diff
@@ -123,7 +132,7 @@ public MsalServiceException(string errorCode, string errorMessage, int statusCod
             Claims = claims;
         }
 
-        private HttpResponse _httpResponse;
+        #endregion
```
jennyf19: do we need the region?

src/Microsoft.Identity.Client/MsalUiRequiredException.cs
```diff
@@ -14,6 +15,37 @@ namespace Microsoft.Identity.Client
     /// </summary>
     public class MsalUiRequiredException : MsalServiceException
     {
+        /// <summary>
+        /// Condition can be resolved by user interaction during the interactive authentication flow.
+        /// See https://aka.ms/msal-net-uirequiredexception-classification for details
```
jennyf19: >uirequiredexception [](start = 40, length = 19)

this is super card to read, can we camel case it?

tests/Microsoft.Identity.Test.Unit.net45/ExceptionTests/MsalExceptionTests.cs
```diff
@@ -3,15 +3,268 @@
...
+using Microsoft.Identity.Client.Utils;
+using Microsoft.Identity.Test.Common.Core.Helpers;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Identity.Test.Unit.ExceptionTests
 {
     [TestClass]
     public class MsalExceptionTests
     {
+        private const string ExCode = "exCode";
```
jennyf19: >exCode [](start = 39, length = 6)

can you put some of these in the const file for tests?

src/Microsoft.Identity.Client/MsalServiceException.cs
bgavrilMS: Regions are an editor feature, no effect on build or runtime. So we definitely don't "need" it, but I like grouping things in regions. 

There are projects that make extensive use of regions - "members", "constructors", "public surface", "private methods" are typical regions. It makes scrolling easier, and you can also glance at how complex a class is by looking at how many items are in the public region (hint: only 1 method and some ctor should ideally be public).

tests/Microsoft.Identity.Test.Unit.net45/ExceptionTests/MsalExceptionTests.cs
bgavrilMS: These constants are private, i.e. they are only used in this class. I can't see other tests that would benefit from them yet.

jmprieur:
@bgavrilMS 
shoudn't we do something in the case of BadToken, and possibly TokenExpired? like removing the token from the cache?

bgavrilMS:
@jmprieur - that's a pretty good idea

src/Microsoft.Identity.Client/MsalServiceException.cs
henrik-me: I vote for no need for regions

---
In reply to: [298062673](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1236#discussion_r298062673) [](ancestors = 298062673)

src/Microsoft.Identity.Client/MsalUiRequiredException.cs
```diff
@@ -14,6 +15,37 @@ namespace Microsoft.Identity.Client
...
+        /// </summary>
+        public const string BasicAction = OAuth2SubError.BasicAction;
```
henrik-me: Is this the right place for these? Not all the InvalidGrant suberrors can be mitigated by UI. I think these should be in a different context. Perhaps InvalidGrantClassifications ?

src/Microsoft.Identity.Client/OAuth2/OAuthConstants.cs
```diff
@@ -83,6 +83,59 @@ internal static class OAuth2Error
         public const string AuthorizationPending = "authorization_pending";
     }
 
+    internal static class OAuth2SubError
```
henrik-me: >OAuth2SubError [](start = 26, length = 14)

Can we call these InvalidGrantClassifications ?

src/Microsoft.Identity.Client/MsalUiRequiredException.cs
```diff
@@ -40,6 +72,38 @@ public class MsalUiRequiredException : MsalServiceException
...
         public MsalUiRequiredException(string errorCode, string errorMessage, Exception innerException) : base(errorCode, errorMessage, innerException)
         {
+           
+        }
+
+        /// <summary>
+        /// Classification of the conditional access error, enabling you to do more actions or inform the user depending on your scenario. See https://aka.ms/msal-net-uirequiredexception-classification for details.
+        /// </summary>
+        /// <remarks>This class <see cref="MsalUiRequiredException"/> lists most classification strings as constants. </remarks>
+        public string Classification
```
henrik-me: IMO, we should stick to suberror publicliy if we plan to combine this with with other sub errors. If we want to stick to this being related to invalid grant classifications then we should imo name it InvalidGrantClassifications, as we are then clear that we link the protocol to the property. 

As mentioned above, I prefer for now to stick with sub-error as that seems to be more flexible. 

Also, really not sure if this is an MsalUiRequiredException in all cases... as there are suberrors which essentially should be a different type of error as it can't be resulved with us showing Ui)

src/Microsoft.Identity.Client/OAuth2/OAuthConstants.cs
bgavrilMS: This is an internal class and the name is taken from the server side (this is how they call them)

Issue #891, opened 2019-02-18
.WithCertificate with /common audience scenario is broken


svrooij:
**Which Version of MSAL are you using ?**
`<PackageReference Include="Microsoft.Identity.Client" Version="2.7.0" />` 

**Platform**
netcore 2.2

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [x] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Code based on this sample https://github.com/microsoftgraph/aspnetcore-connect-sample

**Is this a new or existing app?**
It's a new app, I took created a new empty dotnet core web application and took the code from the above sample as an inspiration.

**Repro**

```csharp
// Azure options method
// ClientCredentials with Key when filled and with certificate when not filled (both created in te portal correctly!!)
public ClientCredential GetClientCredential() {
    if(!string.IsNullOrEmpty(ClientSecret)) return new ClientCredential(ClientSecret);
    if(!string.IsNullOrEmpty(CertificateThumbprint)) {
        X509Certificate2 cert = GetCertificate();
        if(cert != null)
            return new ClientCredential(new ClientAssertionCertificate(cert));
    }
    throw new ArgumentException("No client secret or certificate configured");
}
// Actual code.
var cca = new ConfidentialClientApplication(
    _azureOptions.ClientId,
    _azureOptions.GetAuthority(),
    redirecUrl,
    _azureOptions.GetClientCredential(),
    new Helpers.SessionTokenCache(identifier, memoryCache).GetCacheInstance(),
    null);

var result = await cca.AcquireTokenByAuthorizationCodeAsync(code, scopes);


```

**Expected behavior**
Get an accesstoken

**Actual behavior**
Exception thrown (when certificate set)
```
"AADSTS700023: Client assertion audience claim does not match Realm issuer.
Trace ID: 36e0fa3f-437e-45b5-825e-472930cc7100
Correlation ID: fa45528a-9bde-4eaa-9fd1-ab7c44d15348
Timestamp: 2019-02-18 22:35:46Z"
```

Access Token when the secret is set. (To check the client is created correctly and it doesn't have anything to do with redirect uri, invalid client.....)

**Possible Solution**
[...]

PR #1237, opened 2019-06-27
Fix jwtAudience value having {tenantid} instead of common during clie…

jennyf19:
Fix jwtAudience value having {tenantid} instead of common during clie…nt credential w/certificate flow

- add some consts and move log messages to const
- add u tests

[Issue 891](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/891)

### DIFF
--- a/src/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs
+++ b/src/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs
@@ -78,13 +78,13 @@ namespace Microsoft.Identity.Client.Instance
             endpoints = new AuthorityEndpoints(
                 edr.AuthorizationEndpoint.Replace(Constants.Tenant, tenant),
                 edr.TokenEndpoint.Replace(Constants.Tenant, tenant),
-                ReplaceNonTenantSpecificValueWithCorrectTenant(edr, tenant));
+                ReplaceNonTenantSpecificValueWithTenant(edr, tenant));
 
             Add(authorityInfo, userPrincipalName, endpoints);
             return endpoints;
         }
 
-        internal string ReplaceNonTenantSpecificValueWithCorrectTenant(TenantDiscoveryResponse endpoints, string tenant)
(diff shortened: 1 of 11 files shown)


src/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs
```diff
@@ -58,32 +58,52 @@ public AuthorityEndpointResolutionManager(IServiceBundle serviceBundle, bool sho
...
-                edr.Issuer.Replace("{tenant}", tenant));
+                edr.AuthorizationEndpoint.Replace(Constants.Tenant, tenant),
+                edr.TokenEndpoint.Replace(Constants.Tenant, tenant),
+                CreateSelfSignedJwtAudienceEndpointFromIssuer(edr, tenant));
 
             Add(authorityInfo, userPrincipalName, endpoints);
             return endpoints;
         }
 
+        internal string CreateSelfSignedJwtAudienceEndpointFromIssuer(TenantDiscoveryResponse endpoints, string tenant)
```
henrik-me: This method needs explanation in my mind. It needs to say why this is needed

jennyf19: i agree now that i look at it. how about `ReplaceNonTenantSpecificValueWithCorrectTenant`? 

---
In reply to: [298006468](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1237#discussion_r298006468) [](ancestors = 298006468)

henrik-me: Like that much better. Perhaps just ReplaceNonTenantSpecificValueWithTenant

---
In reply to: [298222183](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1237#discussion_r298222183) [](ancestors = 298222183,298006468)

henrik-me:
Approving with suggestions to rename as mentioned in the comment

jennyf19:
[integration tests passed here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=361616)

PR #1240, opened 2019-06-27
throw MsalUiRequiredException only on some suberrors

bgavrilMS:


src/Microsoft.Identity.Client/InvalidGrantClassification.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        internal const string ClientMismatch = "client_mismatch";
+
+        /// <summary>
+        /// Internal to MSALs. Indicates that device should be re-registered.
+        /// </summary>
+        internal const string DeviceAuthenticationFailed = "device_authentication_failed";
+
+        private static ISet<string> s_nonUiInteractionException = new HashSet<string>(
+                new[] { ClientMismatch, ProtectionPolicyRequired  },
```
bgavrilMS: I could only identify these 2 sub-statues where we should not throw MsalUiRequired exception, because they will not fix anything. We don't treat ProtectionPolicyRequired anywhere, but we have special handling for ClientMismatch.

src/Microsoft.Identity.Client/InvalidGrantClassification.cs
```diff
@@ -0,0 +1,83 @@
...
+                StringComparer.OrdinalIgnoreCase);
+
+        internal static bool IsUiInteractionRequired(string subError)
+        {
+            if (string.IsNullOrEmpty(subError))
+            {
+                return true;
+            }
+
+            return !s_nonUiInteractionException.Contains(subError);
```
henrik-me: Right now we are assuming that errors can be solved with ui interaction unless otherwise specified. Is this the best assumption? We would also start by saying those exceptions we know can be solved with ui interaction should be raised as such? It would make more sense to me to use this approach as it's easier to ship a fix where we add new UIInteractionREquired exceptions than when we remove those.

src/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+            }
+
+
+            ex.ResponseBody = httpResponse?.Body;
+            ex.StatusCode = httpResponse != null ? (int)httpResponse.StatusCode : 0;
+            ex.Headers = httpResponse?.Headers;
+
+            ex.Claims = oAuth2Response?.Claims;
+            ex.CorrelationId = oAuth2Response?.CorrelationId;
+            ex.SubError = oAuth2Response?.SubError;
```
henrik-me: >ex.SubError = oAuth2Response?.SubError; [](start = 12, length = 39)

Should we filter this to the sub errors that we know?

src/Microsoft.Identity.Client/MsalServiceException.cs
```diff
@@ -246,5 +216,36 @@ internal override void PopulateObjectFromJson(JObject jobj)
...
+                    case Client.InvalidGrantClassification.BadToken:
+                    case Client.InvalidGrantClassification.TokenExpired:
+                    case Client.InvalidGrantClassification.ProtectionPolicyRequired:
+                    case Client.InvalidGrantClassification.ClientMismatch:
+                    case Client.InvalidGrantClassification.DeviceAuthenticationFailed:
+                        return string.Empty;
+
+                    // Forward compatibility - new sub-errors bubble through
+                    default:
+                        return SubError;
```
henrik-me: perhaps we should flip this to only expose what we know? Perhaps there will be more like the above where we don't expose the invalid grant classification... not sure. Alternatively we can expose everything....

henrik-me:
Lots of great stuff imo. I thinks it requires a short conversation to finalize this.

src/Microsoft.Identity.Client/InvalidGrantClassification.cs
```diff
@@ -0,0 +1,104 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Sub-errors send by AAD to indicate than user interaction is required. See https://aka.ms/msal-net-UiRequiredExceptionfor details.
+    /// </summary>
+    public static class InvalidGrantClassification
```
henrik-me: >public [](start = 4, length = 6)

Mark as internal for now?

src/Microsoft.Identity.Client/InvalidGrantClassification.cs
bgavrilMS: So we're discussing between an "allow list" vs "block list", with the current implementation showing a block list. The advantage of a block list is around backwards compat - if EVO decides to introduce another suberror, for example "bad_token" becomes "bad_token" and "really_bad_token", then an allow list will break our customers because "really_bad_token" will not throw an UIRequiredException.

src/Microsoft.Identity.Client/MsalServiceException.cs
henrik-me: to be considered. currently leaning towards not exposing it and not causing a ui require exception by default

---
In reply to: [298366905](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1240#discussion_r298366905) [](ancestors = 298366905)

Issue #1188, opened 2019-05-31
Add .WithClientAssertion to support client assertions


henrik-me:
ADAL supports ClientAssertions. We will need to support the same in MSAL.
See more:

- [ADAL FAQ](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/ADAL-.NET-Q-&-A#what-overload-of-acquiretoken-should-i-call)
- [ADAL ClientAssertion Description](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/wiki/Client-credential-flows#clientassertion-in-adalnet)

Note: This looks very similar to UserAssertion

**Describe the solution you'd like**
We should follow the new model of .With in the API
With Claims:
```csharp

Dictionary<string, string> claims = new[] 
{
 { "client_ip", ipAddres}
};

IClientAssertion clientAssertion = CreateClientAssertion(certificate, claims);
var app = ConfidentialClientApplicationBuilder.Create(clientId)
                      .WithClientAssertion(clientAssertion))
                      .Build();
```

With Signed assertion:
```csharp
string SignedAssertion = 
          string.Concat(token, ".", UrlEncodeSegment(credential.Sign(certificate, token)));

IClientAssertion clientAssertion = CreateClientAssertion(SignedAssertion);
var app = ConfidentialClientApplicationBuilder.Create(clientId)
                      .WithClientAssertion(clientAssertion))
                      .Build();
```

**Describe alternatives you've considered**
N/A

**Additional context**
We have a current ask from customers for adding this to be able to move to MSAL.

henrik-me:
@jmprieur

jmprieur:
@henrik-me : something like ^^

bgavrilMS:
Shouldn't this go through an API review? 
I don't understand what the constructor with certificate and claims is supposed to do and how it is different than the .WithCertificate option.

trwalke:
This is different from the .WithCertificate option because we allow the user to pass in their own claims to be signed by the certificate rather than signing a predefined set of claims for the authentication call.

I dont think this should go through the api review as we are simply bringing MSAL to parity with ADAL. However, it may be warranted since we are allowing the user to pass in the claims(Dictionary<string, string> claims) which is not done in ADAL.
@henrik-me thoughts on this?

jmprieur:
[...]

PR #1242, opened 2019-06-28
Updating client WithClaintClaims api.

trwalke:
addressing follow up comments for the WithClientClaims api
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1188

### DIFF
--- a/src/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -70,12 +70,14 @@ namespace Microsoft.Identity.Client
 
         /// <summary>
         /// Sets the certificate associated with the application along with the specific claims to sign.
-        /// The required set of claims will need to be provided as well for a succsesful authentication request. See https://aka.ms/msal-net-client-assertion
+        /// By default, this will merge the <paramref name="claimsToSign"/> with the default required set of claims needed for authentication.
+        /// If <paramref name="mergeWithDefaultClaims"/> is set to false, you will need to provide the required default claims. See https://aka.ms/msal-net-client-assertion
         /// </summary>
         /// <param name="certificate">The X509 certificate used as credentials to prove the identity of the application to Azure AD.</param>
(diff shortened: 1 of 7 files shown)


src/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -70,12 +70,14 @@ public ConfidentialClientApplicationBuilder WithCertificate(X509Certificate2 cer
...
+        /// If <paramref name="mergeWithDefaultClaims"/> is set to false, you will need to provide the required default claims. See https://aka.ms/msal-net-client-assertion
         /// </summary>
         /// <param name="certificate">The X509 certificate used as credentials to prove the identity of the application to Azure AD.</param>
         /// <param name="claimsToSign">The claims to be signed by the provided certificate.</param>
+        /// <param name="mergeWithDefaultClaims">Determines whether or not to merge <paramref name="claimsToSign"/> with the default claims required for authentication.</param>
         /// <remarks>You should use certificates with a private key size of at least 2048 bytes. Future versions of this library might reject certificates with smaller keys. </remarks>
-        public ConfidentialClientApplicationBuilder WithClientClaims(X509Certificate2 certificate, IDictionary<string, string> claimsToSign)
+        public ConfidentialClientApplicationBuilder WithClientClaims(X509Certificate2 certificate, IDictionary<string, string> claimsToSign, bool mergeWithDefaultClaims = true)
```
trwalke: Overall the same functionality. the only difference here is we are specifying if MSAL should append default claims with a Boolean instead of a separate method.
@henrik-me @jmprieur

henrik-me: >mergeWithDefaultClaims [](start = 146, length = 22)

Assuming true is the right default... seems like this was different in ADAL but I believe this is ok. just to be aware that we may have to highlight that in the wiki page. Travis are you updating the wiki page?

henrik-me:
approving with a follow up to update the wiki

Issue #1199, opened 2019-06-03
Fundamentals: Ensure we have standard/cross library cache tests in the repo


henrik-me:
**Expected behavior**
All tests described for basic cache compat should be part of the list of unit tests.
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FUnifiedSchema%2Ftestcases&version=GBdev

All tests should be green, and the test should be clearly marked as "UnifiedSchema_Validation", ensuring we can properly report on the state of these tests.

MarkZuber:
fixed and merged.

PR #1243, opened 2019-06-28
Add unified schema validation tests

MarkZuber:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1199

I've tried to keep these tests very similar to the iOS tests located here: 
https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/blob/master/IdentityCore/tests/integration/MSIDCacheSchemaValidationTests.m

So that we can keep in sync with them easier.

### DIFF
--- a/src/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
+++ b/src/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Identity.Client.Cache.Items
         }
 
         internal MsalAccessTokenCacheItem(
-            string prefferedCacheEnv,
+            string preferredCacheEnv,
             string clientId,
             string scopes,
             string tenantId,
@@ -50,7 +50,7 @@ namespace Microsoft.Identity.Client.Cache.Items
             string userId = null)
             : this()
         {
-            Environment = prefferedCacheEnv;
+            Environment = preferredCacheEnv;
             ClientId = clientId;
             NormalizedScopes = scopes;
             TenantId = tenantId;
--- a/tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedSchemaValidationTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedSchemaValidationTests.cs
@@ -99,8 +99,6 @@ namespace Microsoft.Identity.Test.Unit.CacheTests
             string cachedAt = CoreHelpers.DateTimeToUnixTimestamp(currentDate);
 
             // 1. Verify payload
-
-            // TODO: ios tests do not have client_info key in them.  Should they?
             var expectedJsonObject = new JObject
             {
                 ["secret"] = "<removed_at>",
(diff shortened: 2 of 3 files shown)


src/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -50,7 +50,7 @@ internal MsalAccessTokenCacheItem()
             string userId = null)
             : this()
         {
-            Environment = prefferedCacheEnv;
+            Environment = preferredCacheEnv;
```
bgavrilMS: :)

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedSchemaValidationTests.cs
```diff
@@ -0,0 +1,878 @@
...
+                ["secret"] = "<removed_rt>",
+                ["credential_type"] = "RefreshToken",
+                ["environment"] = "login.microsoftonline.com",
+                ["home_account_id"] = "9f4880d8-80ba-4c40-97bc-f7a23c703084.f645ad92-e38d-4d1a-b510-d1b09a74a8ca",
+                ["client_id"] = "b6c69a37-df96-4db0-9088-2ab96e1d8215",
+                ["client_info"] = "eyJ1aWQiOiI5ZjQ4ODBkOC04MGJhLTRjNDAtOTdiYy1mN2EyM2M3MDMwODQiLCJ1dGlkIjoiZjY0NWFkOTItZTM4ZC00ZDFhLWI1MTAtZDFiMDlhNzRhOGNhIn0",
+                ["family_id"] = "1",
+            };
+
+            AssertAreJsonStringsEquivalent(expectedJsonObject.ToString(), credential.ToJsonString());
```
bgavrilMS: I probably didn't fully implement the FRT serialization on iOS because FOCI is disabled on iOS. In hindsight, I should have had this throw a NotImplException. Should be simple to fix - see how we create the key for non-iOS.

MarkZuber: Do we actually generate a refresh token AND a FRT item in .NET?  The tests olga has have a RefreshToken and a FamilyRefreshToken as separate items.  This test is returning refreshtoken-1-- which is what i'd expect for a family token and not to get this at all.  i'm not sure why this is a test since we shouldn't have two items.

MarkZuber: Updated comments in code and I'm ignoring this test for now since we don't support the scenario in .NET of writing both.

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedSchemaValidationTests.cs
```diff
@@ -0,0 +1,842 @@
...
+
+    [TestClass]
+    [TestCategory("UnifiedSchema_Validation")]
+    public class UnifiedSchemaValidationTests : TestBase
+    {
+        private const string ClientId = "b6c69a37-df96-4db0-9088-2ab96e1d8215";
+        private const string B2CClientId = "0a7f52dd-260e-432f-94de-b47828c3f372";
+        private const string B2CScopes = "https://iosmsalb2c.onmicrosoft.com/webapitest/user.read";
+        private const string AuthorityUri = "https://login.microsoftonline.com/common";
+        private const string MsalEnvironment = "login.microsoftonline.com";
```
jennyf19: what about b2c and custom domains?

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedSchemaValidationTests.cs
```diff
@@ -0,0 +1,842 @@
...
+
+            // 1. Verify payload
+            var expectedJsonObject = new JObject
+            {
+                ["secret"] = "<removed_at>",
+                ["target"] = "https://iosmsalb2c.onmicrosoft.com/webapitest/user.read",
+                ["extended_expires_on"] = extExpiresOn,
+                ["ext_expires_on"] = extExpiresOn,
+                ["credential_type"] = "AccessToken",
+                ["environment"] = "login.microsoftonline.com",
```
jennyf19: this is confusing, because it's a b2c account...if i run the desktop test w/b2c, i get: `msidlabb2c.b2clogin.com` as the environment...and login.microsoftonline.com is being deprecated, so we should use `something.b2clogin.com` and what about custom domains?

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedSchemaValidationTests.cs
MarkZuber: this is for now just a replica of what the iOS team has.  i can talk to olga about adding tests here for custom domains and expanding b2c tests

jennyf19: hmmm...interesting. just doesn't reflect what customers are using.

---
In reply to: [299257205](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1243#discussion_r299257205) [](ancestors = 299257205)

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedSchemaValidationTests.cs
```diff
@@ -746,15 +714,13 @@ public void TestSchemaComplianceForIDToken_WhenMSSTSResponse_WithAADAccountAndFo
             Assert.AreEqual(2003, key.iOSType);
         }
 
+        // TODO: We have not, on .NET, implemented saving two different fresh tokens in FOCI case (one FRT and one MRT).  So this test won't succeed.
+        // Marking it ignored now until we get through other PRs active for this to get resolved.
```
bgavrilMS: @SomkaPe is driving an redesign of FOCI that states that an FRT will be saved as an FRT alone. IMO the schema tests will need to be updated.

PR #1250, opened 2019-07-03
[Instance Metadata] If network discovery fails, use well-known or aut…

bgavrilMS:
…hority

This will improve the reliablity of MSAL - if we can't perform an actual network instance discovery, then fall back to well-known aliases. If those fail, fall back to use only the provided authority. 

Does not impact authority validation, which will still propagate the failure.

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
@@ -3,8 +3,10 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Instance.Discovery;
+using Microsoft.Identity.Test.Common.Core.Helpers;
 using Microsoft.Identity.Test.Common.Core.Mocks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NSubstitute;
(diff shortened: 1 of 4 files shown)


tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -119,6 +122,80 @@ public async Task KnownMetadataProviderIsCheckedSecondAsync()
...
+            _knownMetadataProvider.GetMetadata("some_env.com", Enumerable.Empty<string>()).Returns(_expectedResult);
+
+            // network fails with something other than invalid_instance exception
+            _networkMetadataProvider
+                .When(x => x.FetchAllDiscoveryMetadataAsync(Arg.Any<Uri>(), _testRequestContext))
+                .Do(x => throw new MsalServiceException("endpoint_busy", "some exception message"));
+
+
+            // Act
+            var actualResult = await _discoveryManager.GetMetadataEntryAsync(
```
jennyf19: nit: extra line above

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -119,6 +122,80 @@ public async Task KnownMetadataProviderIsCheckedSecondAsync()
...
+            // Arrange
+            _staticMetadataProvider = new StaticMetadataProvider();
+
+            // no known metadata 
+            _knownMetadataProvider.GetMetadata(null, null).ReturnsForAnyArgs((InstanceDiscoveryMetadataEntry)null);
+
+            // network fails with something other than invalid_instance exception
+            _networkMetadataProvider
+                .When(x => x.FetchAllDiscoveryMetadataAsync(Arg.Any<Uri>(), _testRequestContext))
+                .Do(x => throw new MsalServiceException("endpoint_busy", "some exception message"));
```
jennyf19: >endpoint_busy [](start = 57, length = 13)

is this an actual exception we can get?

bgavrilMS: No, it's an exception that is different from "invalid_instance".

PR #1252, opened 2019-07-04
[Instance Metadata] Allow users to specify their own instance metadata

bgavrilMS:


src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -37,6 +38,40 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
             return (T)this;
         }
 
+#pragma warning disable CS1570 // XML comment has badly formed XML
+        /// <summary>
+        /// Allows developers to configure details about the authorities used, bypassing the need for MSAL
+        /// to perform instance discovery, e.g. https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fv2.0%2Fauthorize
```
jmprieur: actually you need to replace & by `&amp;`

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -37,6 +38,40 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
...
+        /// MSAL uses the aliases to identify tokens that were issued to equivalent authorities, as tokens issued for one authority in the alias set is 
+        /// valid for all other aliases. For more details see https://aka.ms/msal-net-custom-instance-metadata
+        /// </summary>
+        /// <remarks>
+        /// For security reasons, do not use this option if the authority is not a constant.
```
jmprieur: what does it mean for the authority to be constant?

I would write something like:
"Don't use this method unless you have specific needs that MSAL.NET's default behavior (fetching the instance metadata from the Azure AD discovery endpoint) does not fullfil"

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -11,26 +11,53 @@
 
 namespace Microsoft.Identity.Client.Instance.Discovery
 {
+    /// <summary>
+    /// Priority order of metadata providers: 
+    /// 
+    /// 1. Static cache (this is populated from the network)
+    /// 2. Well-known cache if all enviroments present in the token cache are known (this is hardcoded into msal)
+    /// 3. Cache stored in token cache (Not currently implemented)
+    /// 4. User provided AAD instance metadata
```
jmprieur: Add `<see cref="AbstractApplicationBuilder.WithInstanceDiscoveryMetadata"/>` ?

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: I'll add an aka.ms link instead. If I use &amp then clicking on the link from the comments in VS doesn't work (our customers can easily do this when they click "go to definition"). Not sure how many people people use a comments browser nowadays. 

aka.ms will solve both scenarios.

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: The reason this whole validate authority was introdcued was because in the old days you could send an unauthenticated request to a priotected API and it would reply with an Unauthorized with a header "Authority=www.someauthority.com". Developers take the authority and pass it on to ADAL / MSAL. 

A hacked API could give send you to a hacked authority.

bgavrilMS: Will update the comment though.

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
bgavrilMS: Ok, but pls note that this isn't public API.

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -37,6 +38,37 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
             return (T)this;
         }
 
+        /// <summary>
+        /// Allows developers to configure details about the authorities used, bypassing the need for MSAL
```
henrik-me: > bypassing the need for MSAL [](start = 78, length = 28)

I'm assuming this is not what it is. Rather this is about allow customers to set their own valid authorities and thus they have to be careful when crafting this document?

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -37,6 +38,37 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
...
+        /// to perform instance discovery, e.g. https://aka.ms/aad-instance-discovery
+        /// MSAL uses the aliases to identify tokens that were issued to equivalent authorities, as tokens issued for one authority in the alias set is 
+        /// valid for all other aliases. For more details see https://aka.ms/msal-net-custom-instance-metadata
+        /// </summary>
+        /// <remarks>
+        /// Don't use this method unless you have specific needs that MSAL.NET's default behavior (fetching the instance metadata from the Azure AD discovery endpoint) does not fulfill.
+        /// Has no effect on ADFS or B2C authorities, only for AAD authorities</remarks>
+        /// <param name="instanceDiscoveryJson"></param>
```
henrik-me: Per question above: Should clarify that when using this the developer takes the responsibility of this document holding critical information on valid authorities and this can be a security concern!

Note: This functionality needs to be properly documented in the thread model and highlighted in the next review

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -95,16 +122,24 @@ public async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntryAsync(string a
...
-
-                    if (entry != null)
+                    // If the user provides metadata, don't go to the network
+                    if (_userMetadataProvider != null)
                     {
-                        return entry;
+                        entry = _userMetadataProvider.GetMetadata(environment);
                     }
+                    else
+                    {
```
henrik-me: suggestion: reduce nesting by and return early

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -95,16 +122,24 @@ public async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntryAsync(string a
...
-                    entry = await FetchNetworkMetadataOrFallbackAsync(requestContext, authorityUri, environment).ConfigureAwait(false);
-
-                    if (entry != null)
+                    // If the user provides metadata, don't go to the network
+                    if (_userMetadataProvider != null)
                     {
-                        return entry;
+                        entry = _userMetadataProvider.GetMetadata(environment);
                     }
+                    else
```
henrik-me: perhaps the priority selection should be extracted to a different class? that has that responsibility?

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -95,16 +122,24 @@ public async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntryAsync(string a
                         return entry;
                     }
 
-                    entry = await FetchNetworkMetadataOrFallbackAsync(requestContext, authorityUri, environment).ConfigureAwait(false);
-
-                    if (entry != null)
+                    // If the user provides metadata, don't go to the network
+                    if (_userMetadataProvider != null)
```
henrik-me: it's not clear from the code why we would ever get to the user provided metadata, as it's not visible in which cases the static metadata would not return a value (e.g. I think of the static metadata provider as some static fields we have set in the code, however I also remember that we have something that sets these values the first time we fetch the data from somewhere). Wondering if there is somehow that can be made more clear from the code?

tests/Microsoft.Identity.Test.Unit.net45/AppConfigTests/ConfidentialClientApplicationBuilderTests.cs
```diff
@@ -221,5 +223,40 @@ public void TestConstructor_WithCertificate_X509Certificate2()
 
             Assert.IsNotNull(cca.AppConfig.ClientCredentialCertificate);
         }
+
+        [TestMethod]
```
henrik-me: >[TestMethod] [](start = 8, length = 12)

Test suggestion: Can we add a test that validates that no http calls are made?

tests/Microsoft.Identity.Test.Unit.net45/AppConfigTests/ConfidentialClientApplicationBuilderTests.cs
```diff
@@ -221,5 +223,40 @@ public void TestConstructor_WithCertificate_X509Certificate2()
...
+                                                  .WithAuthority("https://some.authority/bogus/", true)
+                                                  .Build());
+            Assert.AreEqual(ex.ErrorCode, MsalError.ValidateAuthorityOrCustomMetadata);
+        }
+
+        [TestMethod]
+        public void TestConstructor_BadInstanceMetadata()
+        {
+            var ex = AssertException.Throws<MsalClientException>(() => ConfidentialClientApplicationBuilder.Create(MsalTestConstants.ClientId)
+                                                  .WithInstanceDicoveryMetadata("{bad_json_metadata")
```
henrik-me: >{ [](start = 81, length = 1)

nit: ?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
```diff
@@ -147,49 +148,57 @@ public void AcquireTokenSilentScopeAndUserOverloadDefaultAuthorityTest()
...
-                    .ExecuteAsync(CancellationToken.None);
-
-                AuthenticationResult result = task.Result;
-                Assert.IsNotNull(result);
-                Assert.AreEqual(MsalTestConstants.DisplayableId, result.Account.Username);
-                Assert.AreEqual(MsalTestConstants.Scope.AsSingleString(), result.Scopes.AsSingleString());
+               
+
+                var app = PublicClientApplicationBuilder.Create(MsalTestConstants.ClientId)
+                                                                           .WithAuthority(new Uri(MsalTestConstants.AuthorityGuestTenant), false)
```
henrik-me: Why false?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
```diff
@@ -147,49 +148,57 @@ public void AcquireTokenSilentScopeAndUserOverloadDefaultAuthorityTest()
...
+                //Task<AuthenticationResult> task = app
+                //    .AcquireTokenSilent(
+                //        MsalTestConstants.Scope.ToArray(),
+                //        new Account(MsalTestConstants.UserIdentifier, MsalTestConstants.DisplayableId, null))
+                //    .ExecuteAsync(CancellationToken.None);
+
+                //AuthenticationResult result = task.Result;
+                //Assert.IsNotNull(result);
+                //Assert.AreEqual(MsalTestConstants.DisplayableId, result.Account.Username);
+                //Assert.AreEqual(MsalTestConstants.Scope.AsSingleString(), result.Scopes.AsSingleString());
```
henrik-me: Seems like the test is not complete?

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: Yeah, I thought about combining this with authority validation. IMO these are separate issues. Authority validation means: 
- [AAD] an instance discovery call, which MSAL performs anyway
- [B2C] not defined
- [ADFS] something ADFS specific

IMO we should deprecate the "validateAuthority" flag from all `WithAuthority()` builder methods, except for `WithAdfsAuthority()`. It makes no sense to have an authority validation flag on AAD since we perform validation anyway. 

Allowing the user to provide "valid authorities" does indeed bypass the `validate authority` mechanism.

bgavrilMS: Updated comments.

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
bgavrilMS: Ok, I renamed the Static provider as NetworkCache provider and rewrote a bit the logic, hopefully it will be more clear.

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
bgavrilMS: The responsability of this class is to call providers in the correct order.

src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
bgavrilMS: I also moved more logic into the network provider, which now populates the cache on its own.

tests/Microsoft.Identity.Test.Unit.net45/AppConfigTests/ConfidentialClientApplicationBuilderTests.cs
bgavrilMS: We need a test that checks if the user configures a bad json. This is it.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
bgavrilMS: It doesn't make sense to set validate metadata = true and to have this set. So they'll have to set validate flag to false, otherwise I throw, i.e. these 2 settings are mutually exclusive.  I think this will require developers to understand if they really need to do this, read the docs etc.

bgavrilMS: CC @jmprieur to whom I have discussed about this, but I believe it is an important detail.

bgavrilMS: @henrik-me - the exception message is: 

_"You have configured custom instance metadata, but the validateAuthority flag is set to true. These are mutually exclusive. Set the validateAuthority flag to false. See https://aka.ms/msal-net-custom-instance-metadata for more details."_

bgavrilMS:
Addressed all comments, please have another look

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -37,6 +38,41 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
...
+        /// For more details see https://aka.ms/msal-net-custom-instance-metadata
+        /// </summary>
+        /// <remarks>
+        /// Developers take responsibility for authority validation if they use this method. Should not be used when the authority is not know in advance. 
+        /// Has no effect on ADFS or B2C authorities, only for AAD authorities</remarks>
+        /// <param name="instanceDiscoveryJson"></param>
+        /// <returns></returns>
+        public T WithInstanceDicoveryMetadata(string instanceDiscoveryJson)
+        {
+            if (string.IsNullOrEmpty(instanceDiscoveryJson))
```
MarkZuber: please use { } even on one-liners

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -272,20 +308,25 @@ internal virtual void Validate()
             // Validate that we have a client id
             if (string.IsNullOrWhiteSpace(Config.ClientId))
             {
-                throw new InvalidOperationException(MsalErrorMessage.NoClientIdWasSpecified);
+                throw new MsalClientException("no_client_id", MsalErrorMessage.NoClientIdWasSpecified);
```
MarkZuber: is the "no_client_id" the MsalErrorCode?  If so, shouldn't this be a constant in MsalErrorCode.cs?

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -308,7 +349,7 @@ private void TryAddDefaultAuthority()
             Config.AuthorityInfo = new AuthorityInfo(
                     AuthorityType.Aad,
                     new Uri($"{defaultAuthorityInstance}/{defaultAuthorityAudience}").ToString(),
-                    true);
+                    false /* default authority is well know, there is no need for validation */);
```
MarkZuber: typo:  "well known" (missing n)

src/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -41,6 +42,7 @@ public string ClientVersion
 
         public IHttpManager HttpManager { get; internal set; }
 
+
```
MarkZuber: nit: please remove extra blank line

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: Actually our new code style, which is copied from the .net core guys, specifically enables this. It is the reason we moved to it - @brentschmaltz uses this in their projects and we wanted to have the same code standard across C# libs.

src/Microsoft.Identity.Client/Instance/Discovery/IUserMetadataProvider.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.Identity.Client.Core;
```
MarkZuber: need top of file license comment

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: That's what I get from trying to be nice :D :D . I'll fix this.

src/Microsoft.Identity.Client/Instance/Discovery/NetworkCacheMetadataProvider.cs
```diff
@@ -2,17 +2,21 @@
...
+    internal class NetworkCacheMetadataProvider : INetworkCacheMetadataProvider
     {
         private static readonly ConcurrentDictionary<string, InstanceDiscoveryMetadataEntry> s_cache =
              new ConcurrentDictionary<string, InstanceDiscoveryMetadataEntry>();
 
-        public InstanceDiscoveryMetadataEntry GetMetadata(string environment)
+        public InstanceDiscoveryMetadataEntry GetMetadata(string environment, ICoreLogger logger)
         {
             s_cache.TryGetValue(environment, out InstanceDiscoveryMetadataEntry entry);
+            logger.Verbose($"[Instance Discovery] Tried to use network cache provider " +
```
MarkZuber: the + here is going to allocate 3 strings and combine them together under the covers.  this gets expensive and it's only for code readability here.  i believe you can use \ or something to define the single string constant around wrapping lines instead of doing a runtime concat of strings.  please fix this in this and other cases of logging in this PR.

src/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -273,5 +273,19 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
         public const string AuthorizeEndpointWasNotFoundInTheOpenIdConfiguration = "Authorize endpoint was not found in the openid configuration";
         public const string TokenEndpointWasNotFoundInTheOpenIdConfiguration = "Token endpoint was not found in the openid configuration";
         public const string IssuerWasNotFoundInTheOpenIdConfiguration = "Issuer was not found in the openid configuration";
+        public const string InvalidUserInstanceMetadata = "The json containing instance metadata could not be parsed. " +
```
MarkZuber: don't use "+" in the constants either, this is a runtime allocation cost.

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -151,12 +170,12 @@ public async Task AuthorityValidationFailure_IsRethrown_Async()
         public async Task NetworkProviderFailures_AreIgnored_Async()
         {
             // Arrange
-            _staticMetadataProvider = new StaticMetadataProvider();
-            _knownMetadataProvider.GetMetadata("some_env.com", Enumerable.Empty<string>()).Returns(_expectedResult);
+            _networkCacheMetadataProvider = new NetworkCacheMetadataProvider();
+            _knownMetadataProvider.GetMetadata("some_env.com", Enumerable.Empty<string>(), Arg.Any<ICoreLogger>()).Returns(_expectedResult);
```
MarkZuber: should "some_env.com" be a constant?

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -234,8 +253,46 @@ public async Task NetworkProviderIsCalledLastAsync()
 
             // Assert
             Assert.AreSame(_expectedResult, actualResult, "The known metadata provider should be queried second");
-            _knownMetadataProvider.Received(1).GetMetadata("some_env.com", otherEnvs);
-            _ = _networkMetadataProvider.Received(1).FetchAllDiscoveryMetadataAsync(authorityUri, _testRequestContext);
+            _knownMetadataProvider.Received(1).GetMetadata("some_env.com", otherEnvs, Arg.Any<ICoreLogger>());
+            _ = _networkMetadataProvider.Received(1).GetMetadataAsync(authorityUri, _testRequestContext);
```
MarkZuber: the _ is going to get the task and this won't be awaited within the bounds of the test.  Is this what you intended?  why are you not awaiting for the result of the call within this test?

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -234,8 +253,46 @@ public async Task NetworkProviderIsCalledLastAsync()
...
+               .ConfigureAwait(false);
+
+            // Assert
+            Assert.AreSame(_expectedResult, actualResult, "The user metadata provider should be queried second");
+            Assert.AreSame(_expectedResult, actualResult2, "The user metadata provider should be queried second");
+            _userMetadataProvider.Received(2).GetMetadataOrThrow("some_env.com", Arg.Any<ICoreLogger>());
+            _knownMetadataProvider.DidNotReceiveWithAnyArgs().GetMetadata(null, null, null);
+            _networkMetadataProvider.DidNotReceiveWithAnyArgs().GetMetadataAsync(
+                Arg.Any<Uri>(),
+                Arg.Any<RequestContext>()).GetAwaiter().GetResult();
```
MarkZuber: you are in an async method.  you should call await here instead of using GetAwaiter().GetResult()

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -327,6 +328,44 @@ public void AcquireTokenSameUserTest()
...
+        {
+            string instanceMetadataJson = File.ReadAllText(
+                ResourceHelper.GetTestResourceRelativePath("CustomInstanceMetadata.json"));
+
+            using (var harness = CreateTestHarness())
+            {
+                // No instance discovery is made - it is important to not have this mock handler added
+                // harness.HttpManager.AddInstanceDiscoveryMockHandler();
+
+                PublicClientApplication app = PublicClientApplicationBuilder.Create(MsalTestConstants.ClientId)
```
MarkZuber: nit: these read better in code reviews if they're wrapped like on line 356 (4-space tab indented next line instead of aligned under .Create)

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
MarkZuber: nope.  totally disagree.  this causes bugs and reduces readability.  we need to revert that change.

jmprieur: I agree with @MarkZuber on that (and disagree with Brent). New comers coming to the code or PMs won't pay attention. Having the {} on one lines makes it safer to change the block

bgavrilMS: Ok, I don't have strong opinions on this particular item, but we had a conversation over email on this and made this decisions. Please reopen that conversation or start a new one,

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
bgavrilMS: Normally yes, but with NSubstitute it's different. See details here: https://stackoverflow.com/questions/30996024/check-calls-received-for-async-method/31021430#31021430

bgavrilMS:
@MarkZuber - address all comments, please have another look.

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: Personally I disagree that is reduces readability - because you can fit more important lines of code on the screen, i.e. 

```csharp 
if (arg == null) 
{
   throw new ArgumentNullException("arg");
}
```
is less readable than

```csharp
if (arg == null) 
   throw new ArgumentNullException("arg");
```

I agree it is potentially more dangerous for developers who are not familiar with C# / C / Java.

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
bgavrilMS: In product code, I'm all for constants. In test code, I tend to use strings inline because of readability. Tests are very difficult to read anyway and they also use much more constants than product code. Being able to read a test from top to bottom seems more important to me than changing the value of a test constant, which happens very rarely.

bgavrilMS:
Issue #1259 1259

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -37,6 +38,41 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
...
+        /// </summary>
+        /// <remarks>
+        /// Developers take responsibility for authority validation if they use this method. Should not be used when the authority is not know in advance. 
+        /// Has no effect on ADFS or B2C authorities, only for AAD authorities</remarks>
+        /// <param name="instanceDiscoveryJson"></param>
+        /// <returns></returns>
+        public T WithInstanceDicoveryMetadata(string instanceDiscoveryJson)
+        {
+            if (string.IsNullOrEmpty(instanceDiscoveryJson))
+                throw new ArgumentNullException(instanceDiscoveryJson);
```
MarkZuber: for now, for consistency with the entire rest of our codebase which follows this rule, can you put the {} on here and we'll start the mail thread up?

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
MarkZuber: Can you please add that note as a comment here?  Readability wise this stands out as not directly explained by reading the code and this pointer will be helpful to future people reading this test.

src/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -594,6 +596,36 @@ public static class MsalError
...
+        /// <para>What happens?</para>You've specified a client ID that is not a <see cref="Guid"/>
+        /// <para>Mitigation</para>Use the application ID (a Guid) from the application portal as client ID in this SDK
+        /// </summary>
+        public const string ClientIdMustBeAGuid = "client_id_must_be_guid";
+
+        /// <summary>
+        /// <para>What happens?</para>You have configured both Mats and a telemetry config. 
+        /// <para>Mitigation</para>Only one telememtry mechanism can be configured.
+        /// </summary>
+        public const string MatsOrTelemetryCallback = "mats_or_telemetry_callback";
```
henrik-me: >mats_or_ [](start = 55, length = 8)

Not in favor of exposing MATS as a thing externally, can't we stick to calling it TelemetryCallback?

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -272,20 +308,25 @@ internal virtual void Validate()
...
             {
-                throw new InvalidOperationException(MsalErrorMessage.NoClientIdWasSpecified);
+                throw new MsalClientException(MsalError.NoClientId, MsalErrorMessage.NoClientIdWasSpecified);
             }
 
             //Adfs does not require client id to be in the form of a Guid
             if (Config.AuthorityInfo?.AuthorityType != AuthorityType.Adfs && !Guid.TryParse(Config.ClientId, out Guid clientIdGuid))
             {
-                throw new InvalidOperationException(MsalErrorMessage.ClientIdMustBeAGuid);
+                throw new MsalClientException(MsalError.ClientIdMustBeAGuid, MsalErrorMessage.ClientIdMustBeAGuid);
```
henrik-me: >MsalClientException [](start = 26, length = 19)

I suppose this is kind of a breaking change... though it's technically more correct. @JM, thoughts on exposed exception types?

henrik-me:
I would really have liked the provider ordering explained somewhere. perhaps having class explanations for all the providers (even though internal) will help clarify

henrik-me:
There are a decent amount of tests. Should this also be added to the e2e test suite?

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
bgavrilMS: It looks like in newer versions of NSubstitute we can also use await for this assertion, so I'll change it to that.

src/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: Theoretically yes, exceptions are part of the contract, but this isn't enforced like in Java, it's more on us to make a judgement. 

I don't expect developers to write exception handling code for things like using a badly form client ID. MsalClient exception will give a better experience.

bgavrilMS:
@henrik-me - I added an extra test to cover what happens if the user provides metadata that does not cover their authority - see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/61a3667188d3660995da7c5ef703e584d2e09863/tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs#L375

In E2E tests we don't mock the http calls, but the 2 tests in the file above are very close to E2E tests. 

The [InstanceDiscoveryManager ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/61a3667188d3660995da7c5ef703e584d2e09863/src/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs#L15) documents the order of the providers. I added a link to the spec as well.

PR #1254, opened 2019-07-05
UiRequiredException Classification

bgavrilMS:


src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Details about the cause of an <see cref="MsalUiRequiredException"/>, giving a hint about what the user can expect when 
+    /// they go through interactive authentication. See https://aka.ms/msal-net-UiRequiredException for details.
+    /// </summary>
+    public enum UiRequiredExceptionClassification
```
bgavrilMS: An enum is a friendlier way to discover all possible options

src/Microsoft.Identity.Client/MsalUiRequiredException.cs
```diff
@@ -47,12 +47,26 @@ public class MsalUiRequiredException : MsalServiceException
...
+                if (string.Equals(base.SubError, MsalError.MessageOnly, StringComparison.OrdinalIgnoreCase))
+                    return UiRequiredExceptionClassification.MessageOnly;
+
+                if (string.Equals(base.SubError, MsalError.ConsentRequired, StringComparison.OrdinalIgnoreCase))
+                    return UiRequiredExceptionClassification.ConsentRequired;
+
+                if (string.Equals(base.SubError, MsalError.UserPasswordExpired, StringComparison.OrdinalIgnoreCase))
+                    return UiRequiredExceptionClassification.UserPasswordExpired;
+
+                return UiRequiredExceptionClassification.None;
```
bgavrilMS: We now use an allow-list to map between sub-error and the enum.

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// Condition can be resolved by user interaction during the interactive authentication flow.
+        /// See https://aka.ms/msal-net-UiRequiredExceptionfor details
+        /// </summary>
+        BasicAction,
+
+        /// <summary>
+        /// Condition can be resolved by additional remedial interaction with the system, outside of the interactive authentication flow.
+        /// See https://aka.ms/msal-net-UiRequiredExceptionfor details
+        /// </summary>
+        AdditionalAction,
```
henrik-me: not a ui exception classification

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
henrik-me: >UiRequiredExceptionClassification [](start = 16, length = 33)

I like the enum, however the name here is challenging. Perhaps this has to stay as ExceptionClassification or we need two enum's one for ui and one for none ui. or this one could be called InvalidGrantClassification :)

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
bgavrilMS: I disagree, UI is required to display this "Additional action". However, developers can give a visual hint to the user that they won't be able to login to the app anytime soon :). 

From the spec: 

_Call AcquireTokenInteractively() to show a message that explains the remedial action. Calling application may choose to hide flows that require additional_action if the user is unlikely to complete the remedial action._

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
bgavrilMS: There are 2 sources of UIRequiredException: 

- client side MSAL logic, such as calling AcquireTokenSilent(string upn) with an invalid upn. CLassificaiton will be `UiRequiredExceptionClassification.None`
- server side logic, where AAD returns an "invalid_grant" exception with an additional sub_error, which is translated to the `UiRequiredExceptionClassification.BasicAction`, UiRequiredExceptionClassification.PasswordExpired` etc. 

@jmprieur - can you help break the naming dispute ?

src/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -590,6 +590,61 @@ public static class MsalError
         /// </summary>
         public const string ClientCredentialAuthenticationTypesAreMutuallyExclusive = "Client_Credential_Authentication_Types_Are_Mutually_Exclusive";
 
+        #region InvalidGrant suberrors
+        /// <summary>
+        /// Condition can be resolved by user interaction during the interactive authentication flow.
```
jmprieur: Condition? not issue?

src/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -590,6 +590,61 @@ public static class MsalError
...
+        /// See https://aka.ms/msal-net-UiRequiredExceptionfor details
```
jmprieur: there is a typo (you need a space before for)

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
jmprieur: ExceptionClassification ? Or ServiceExceptionClassification?

src/Microsoft.Identity.Client/MsalError.cs
bgavrilMS: This is an internal comment now, I will fix it nontheless.

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
bgavrilMS: Updated the comments to better reflect this.

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
bgavrilMS: @shoatman - can you also have a look at this please?

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Details about the cause of an <see cref="MsalUiRequiredException"/>, giving a hint about what the user can expect when 
+    /// they go through interactive authentication. See https://aka.ms/msal-net-UiRequiredException for details.
+    /// </summary>
+    public enum UiRequiredExceptionClassification
+    {
+        /// <summary>
+        /// No further details are provided. It is possible that the user will be able to resolve the issue by launching interactive authentication.
+        /// See https://aka.ms/msal-net-UiRequiredExceptionfor details
+        /// </summary>
+        None,
```
bgavrilMS: MSAL throw UiRequiredException is also thrown when the user calls AcquireTokenSilent with: 

- account = null
- a login_hint that cannot be matched to 1 (and only 1) account

This is breaks the pattern that all client side exceptions are MsalClientException, in order to provide a more uniform experience (most users don't care why the UiRequiredException is thrown). It's also a downside for controlling the flow using exceptions. 

I think we can improve the user experience if we added more options to this enum for the "client" cases, something like "InvalidAccount"

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// See https://aka.ms/msal-net-UiRequiredException for details.
+        /// </summary>
+        UserPasswordExpired,
+
+        /// <summary>
+        /// <see cref="AcquireTokenInteractiveParameterBuilder.WithPrompt(Prompt)"/> was used with Prompt.Never value, 
+        /// however this could not be honored by the server. Please use a different prompt behavior, such as <see cref="Prompt.SelectAccount"/>
+        /// See https://aka.ms/msal-net-UiRequiredException for details.
+        /// </summary>
+        PromptNeverFailed,
```
bgavrilMS: @henrik-me - I added these extra 2 root causes for UiRequiredException. The first comes from AAD rejecting a prompt=never interactive auth. The second comes from multiple failure conditions in AcquireTokenSilent. Let me know if this makes sense, especially the public comments.

src/Microsoft.Identity.Client/UiRequiredExceptionClassification.cs
henrik-me: IMO: not the intention of the UIRequiredException. Suppose it will work regardless and developers will get used to this.

---
In reply to: [300998582](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1254#discussion_r300998582) [](ancestors = 300998582)

henrik-me:
Are all scenarios fully covered in tests? Hard to see from this PR alone

Issue #1201, opened 2019-06-04
"Key not valid for use in specified state" - happens in .Net Framework 4.7.2 only, not on .Net Core 2.2


William-Au-Yeung:
**Which Version of MSAL are you using ?**
Microsoft.Identity.Client Version=3.0.8

**Platform**

my library in .Net Core
and a web job console in .Net Framework 4.7.2

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [x] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**

C. This is a new app or experiment

**Repro**

```csharp
            SqlConnectionStringBuilder connectionStringBuilder = new SqlConnectionStringBuilder();
            connectionStringBuilder.DataSource = this.Settings.DataSource;
            connectionStringBuilder.InitialCatalog = this.Settings.InitialCatalog;
            connectionStringBuilder.Encrypt = this.Settings.Encrypt;
            connectionStringBuilder.TrustServerCertificate = this.Settings.TrustServerCertificate;

            SqlConnection connection = new SqlConnection(connectionStringBuilder.ConnectionString);
            connection.AccessToken = this.Settings.TokenAuthenticator.GetAccessTokenAsync(DefaultScope).Result;

        public async Task<string> GetAccessTokenAsync(IEnumerable<string> scopes)
        {
            X509Certificate2 certificate = this.GetCertificate();
            ConfidentialClientApplicationOptions option = new ConfidentialClientApplicationOptions
            {
                AzureCloudInstance = this.CloudInstance,
                ClientId = this.ApplicationId,
                TenantId = this.TenantId,
            };
            IConfidentialClientApplication app = ConfidentialClientApplicationBuilder
                .CreateWithApplicationOptions(option)
                .WithCertificate(certificate)
                .Build();

            AuthenticationResult result = await app.AcquireTokenForClient(scopes)
                .ExecuteAsync();

            return result.AccessToken;
        }

```

**Expected behavior**
No exception. I should be able to authenticate a connection to my Azure SQL Database. I have run the same code on .Net Core console project, and it has been working.

**Actual behavior**
I got this exception in my .Net Framework 4.7.2 console project:
[...]

PR #1275, opened 2019-07-17
Updating certificate signing on .NET 4.7+

trwalke:
Modifying how we create the RSACryptoServiceProvider to avoid running into issues with a cert that has a non exportable key. 
fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1201

This change will allow us to support certificates with a non exportable key.

@henrik-me @jmprieur This fix will not work with certs with a CNG key. I am not able to confirm if the azure portal supports CNG keys or not so this should not be merged yet.


tests/Microsoft.Identity.Test.Integration.net45/Microsoft.Identity.Test.Integration.net45.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net462</TargetFramework>
+    <TargetFramework>net472</TargetFramework>
```
trwalke: We have a certificate signing method that behaves differently on >4.6.2 and 4.7+. This change will allow our certificate tests to run on 4.7 and hit the new api for testing. Is this ok or should I create an additional project that targets 4.7 and keep this one at 4.6.2.

trwalke: @henrik-me @bgavrilMS @MarkZuber

bgavrilMS: If we have code paths that behave differently on .net46 and .net47, we should have them both tested, i.e. create a simple test project that targets net46 with the tests that go through that code path and keep net47 for all tests.

PR #1279, opened 2019-07-23
remove enablebroker flag from api

MarkZuber:
Removing the EnableBroker flag from the builder so we can release 4.2.  Broker will be back for 4.3 with a few fixes.


MarkZuber:
tests failing are only maven not existing on build server.  merging and will investigate java cache failures separately.

Issue #1270, opened 2019-07-15
[Bug] Using PublicClientApplicationBuilder and WithAuthority(string cloudInstanceUri, string tenant, bool validateAuthority) throws exception when acquiring token


Crazyconv:
**Which Version of MSAL are you using ?**
MSAL 4.1.0

**Platform**
net45

**What authentication flow has the issue?**
* Desktop / Mobile
    * Username Password

**Is this a new or existing app?**
This is a new app

**Repro**

```csharp
string tenantId = "<tenantId";
string username = "<username>";
string password = "<password>";
string clientId = "<clientId>";

string[] permissions = new[] {"<permission>"};
IPublicClientApplication app = PublicClientApplicationBuilder.Create(clientId).WithAuthority("https://login.microsoftonline.com", tenantId).Build();
AuthenticationResult result = await app.AcquireTokenByUsernamePassword(permissions, username, this.ToSecureString(password)).ExecuteAsync();
```

**Expected behavior**
An AuthenticationResult result is returned.

**Actual behavior**
Exception is thrown.
Microsoft.Identity.Client.MsalServiceException: "AADSTS90002: Tenant 'v2.0' not found. This may happen if there are no active subscriptions for the tenant. Check with your subscription administrator.\r\nTrace ID: 024c3425-f6b7-4f00-a95b-8b0a14325400\r\nCorrelation ID: d2d97559-d4ab-4517-9a3f-bec02f562a50\r\nTimestamp: 2019-07-15 07:33:33Z"

**Possible Solution**
In [this line](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e2bcfcbb41d529fdb34af7550ce122e70e9b4566/src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs#L94), `string.Format(CultureInfo.InvariantCulture, "{0}/{1}/", cloudInstanceUri, tenant)` produces `https://login.microsoftonline.com//<tenantId>`, and so `CanonicalAuthority` of the `AuthorityInfo` object becomes `https://login.microsoftonline.com//`.

Updating the line to `string.Format(CultureInfo.InvariantCulture, "{0}{1}/", cloudInstanceUri, tenant)` should fix the issue. However I am not sure about other possible urls that can be passed into the method so not sure whether it will break other user cases.

trwalke:
[...]

PR #1281, opened 2019-07-23
Resolving issue with authority api adding an additional "/" to the authority.

trwalke:
Resolving issue with authority api adding an additional "/" to the end of the authority.

Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1270

### DIFF
--- a/src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
+++ b/src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Identity.Client
         internal static AuthorityInfo FromAadAuthority(Uri cloudInstanceUri, Guid tenantId, bool validateAuthority)
         {
 #pragma warning disable CA1305 // Specify IFormatProvider
-            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}/{1}/", cloudInstanceUri, tenantId.ToString("D")), validateAuthority);
+            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}{1}/", cloudInstanceUri, tenantId.ToString("D")), validateAuthority);
 #pragma warning restore CA1305 // Specify IFormatProvider
         }
 
@@ -91,7 +91,7 @@ namespace Microsoft.Identity.Client
             {
                 return FromAadAuthority(cloudInstanceUri, tenantId, validateAuthority);
             }
-            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}/{1}/", cloudInstanceUri, tenant), validateAuthority);
+            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}{1}/", cloudInstanceUri, tenant), validateAuthority);
         }
 
         internal static AuthorityInfo FromAadAuthority(
--- a/src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
+++ b/src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Identity.Client
         internal static AuthorityInfo FromAadAuthority(Uri cloudInstanceUri, Guid tenantId, bool validateAuthority)
         {
 #pragma warning disable CA1305 // Specify IFormatProvider
-            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}{1}/", cloudInstanceUri, tenantId.ToString("D")), validateAuthority);
+            return FromAuthorityUri(new Uri(cloudInstanceUri, tenantId.ToString("D")).ToString(), validateAuthority);
 #pragma warning restore CA1305 // Specify IFormatProvider
         }
 
@@ -91,7 +91,7 @@ namespace Microsoft.Identity.Client
             {
                 return FromAadAuthority(cloudInstanceUri, tenantId, validateAuthority);
             }
-            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}{1}/", cloudInstanceUri, tenant), validateAuthority);
+            return FromAuthorityUri(new Uri(cloudInstanceUri, tenant).ToString(), validateAuthority);
         }
 
         internal static AuthorityInfo FromAadAuthority(
--- a/src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
+++ b/src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Identity.Client
         internal static AuthorityInfo FromAadAuthority(Uri cloudInstanceUri, Guid tenantId, bool validateAuthority)
         {
 #pragma warning disable CA1305 // Specify IFormatProvider
-            return FromAuthorityUri(new Uri(cloudInstanceUri, tenantId.ToString("D")).ToString(), validateAuthority);
+            return FromAuthorityUri(new Uri(cloudInstanceUri, tenantId.ToString("D")).AbsoluteUri, validateAuthority);
 #pragma warning restore CA1305 // Specify IFormatProvider
         }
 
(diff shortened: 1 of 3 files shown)


tests/Microsoft.Identity.Test.Common/MsalTestConstants.cs
```diff
@@ -29,7 +29,7 @@ internal static class MsalTestConstants
         public const string Uid = "my-uid";
         public const string Utid = "my-utid";
         public const string Common = "common";
-        public const string TenantId = "e56cat29e-b008-4cea-b6f0-48facatsd64a";
```
trwalke: This was an invalid GUID

src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -81,7 +81,7 @@ internal static AuthorityInfo FromAuthorityUri(string authorityUri, bool validat
         internal static AuthorityInfo FromAadAuthority(Uri cloudInstanceUri, Guid tenantId, bool validateAuthority)
         {
 #pragma warning disable CA1305 // Specify IFormatProvider
-            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}/{1}/", cloudInstanceUri, tenantId.ToString("D")), validateAuthority);
+            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}{1}/", cloudInstanceUri, tenantId.ToString("D")), validateAuthority);
```
trwalke: No need for additional slash here. According to the Uri specs,
 `"Note that the absolute path cannot be empty; if none is present in the original URI, it MUST be given as "/" (the server root)."` [Spec](https://tools.ietf.org/html/rfc2616#section-5.1.2)
Thus, there will always be a "/" appended to the uri so we dont need the additional one here. Otherwise you end up with "https://login.microsoftonline.com//tenantID"

henrik-me: I don't understand the explanation, however if it means that the uri class will always return the uri including the trailing / that is great. Is that what the above means? 

---
In reply to: [306497215](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1281#discussion_r306497215) [](ancestors = 306497215)

trwalke: yes. if you create a uri using the uri class it will always append a trailing "/"

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
```diff
@@ -318,6 +318,31 @@ public void CanonicalAuthorityInitTest()
...
+            var fullAuthority = String.Concat(host, "/" , MsalTestConstants.TenantId, "/");
+
+            var publicClient = PublicClientApplicationBuilder.Create(MsalTestConstants.ClientId)
+                                                             .WithAuthority(host, MsalTestConstants.TenantId)
+                                                             .BuildConcrete();
+
+            Assert.AreEqual(publicClient.Authority, fullAuthority);
+
+            publicClient = PublicClientApplicationBuilder.Create(MsalTestConstants.ClientId)
+                                                         .WithAuthority(host, new Guid(MsalTestConstants.TenantId))
```
henrik-me: >.WithAuthority(host, new Guid(MsalTestConstants.TenantId)) [](start = 57, length = 58)

what is this testing that the above is not testing? Why new Guid?

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
```diff
@@ -318,6 +318,31 @@ public void CanonicalAuthorityInitTest()
             Assert.AreEqual(UriCustomPortTailSlash, authority.AuthorityInfo.CanonicalAuthority);
         }
 
+        [TestMethod]
+        public void TenantSpecificAuthorityInitTest()
```
henrik-me: Will anything fail? Can you pls. ensure we can do the right thing when calling ARM resources? - I believe they have to have the double /

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
trwalke: The above test uses a different constructor on the AuthorityInfo class that did not run into this bug which is why it did not fail. Also the withAuthority requires a Guid so I am simply passing a new GUID here that is created from our test guid string

henrik-me: thanks

---
In reply to: [306517901](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1281#discussion_r306517901) [](ancestors = 306517901)

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
trwalke: Yes, this test will fail if the value that is passed into withAuthority does not match the authority on the application object after it is constructed. 

For the arm resources, if they use the 
`WithAuthority(Uri authorityUri, bool validateAuthority = true)` 
method, it will bypass this code altogether use whatever the user passed in as authorityUri. also, do you know if the double / will be at the end? (https://login.microsoftonline.com/tenant//) or is t supposed to be (https://login.microsoftonline.com//tenant/)?

henrik-me: all your tests shows happy path. please add the test cases that doesn't end up in a happy scenario. Explicitly I would like an ARM test case that validates that arm works.

---
In reply to: [306521638](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1281#discussion_r306521638) [](ancestors = 306521638)

henrik-me: coming to think of it I believe I thought of resources. Still would like to see test cases that ensures it's possible to set a path which includes extra / before and after tenant id. This was possible before, so it's possible this is being used somewhere

---
In reply to: [306529133](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1281#discussion_r306529133) [](ancestors = 306529133,306521638)

src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
bgavrilMS: While your solution works, it's still a bit cryptic because developers do not know how Uri behaves. Consider using the Uri ctor instead: 

```csharp
new Uri(Uri baseUri, string relativeUri)
```

see https://stackoverflow.com/questions/372865/path-combine-for-urls

trwalke: This is much more elegant. Thanks!

src/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -81,7 +81,7 @@ internal static AuthorityInfo FromAuthorityUri(string authorityUri, bool validat
         internal static AuthorityInfo FromAadAuthority(Uri cloudInstanceUri, Guid tenantId, bool validateAuthority)
         {
 #pragma warning disable CA1305 // Specify IFormatProvider
-            return FromAuthorityUri(string.Format(CultureInfo.InvariantCulture, "{0}{1}/", cloudInstanceUri, tenantId.ToString("D")), validateAuthority);
+            return FromAuthorityUri(new Uri(cloudInstanceUri, tenantId.ToString("D")).ToString(), validateAuthority);
```
bgavrilMS: Prefer using uri.AbsoluteUri instead of ToString(), see http://faithlife.codes/blog/2010/08/uritostring_must_die/

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
bgavrilMS: Agreed with Henrik, we need a regression test for this.

PR #1284, opened 2019-07-24
reenable system browser as default on ios and android

jennyf19:


MarkZuber:
Would rename this bug for future archaeology.  You're re-enabling "embedded web view = false" on ios and android, not re-enabling systembrowser = false.

MarkZuber:
PR tests passed

PR #1290, opened 2019-07-24
add 4.2.1 changelog

MarkZuber:


MarkZuber:
> I thought we were just going to put the issue # and not the entire url...not a big issue right now...more of question of how we want to do the change log :)

We certainly can do that, i was just following existing convention.  Agree that [####] is more readable, but the nice thing about the entire URL is also that this markdown works anywhere and not just on github (which does the auto issue number translation)

jennyf19:
>We certainly can do that, i was just following existing convention. Agree that [####] is more readable, but the nice thing about the entire URL is also that this markdown works anywhere and not just on github (which does the auto issue number translation)

good point. I guess @jmprieur  can decide then as he'll probably be most interested in formatting for this. thx.

Issue #1280, opened 2019-07-23
Update iOS Broker Logic and re-enable WithBroker


MarkZuber:
Current broker logic always does interactive flow first and only if msauth:// is in the auth code response (meaning broker isn't installed and MAM/CA policy is required) will we launch the broker.  

Broker logic should do this in InteractiveRequest:

If IsBrokerEnabled (by developer)
	if CanOpenUrl(msauth://) {
	   send request to broker
	} else {
    try interactive flow
  }
}	else {
  show interactive ui login
	if during interactive login we get a msauth:// challenge in the auth code, install/launch the broker (regardless if it's enabled by the developer) or fail if it's not installed.
}

jennyf19:
Included in [4.3.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.3.0) release

PR #1295, opened 2019-07-26
Reenable iOS broker

jennyf19:
- fix broker implementation
- add more test coverage

@bgavrilMS and I pair-programmed part of this. We can now mock the broker response e2e. 

Updated broker logic:
If developer does ->
`.WithBroker()`, then we will go through the broker flow
OR
if we get msauth:// back w/the auth code, we will go through broker because it will be required for that user

else -> do regular AT flow

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1280


src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
```diff
@@ -270,5 +264,20 @@ private void VerifyAuthorizationResult()
...
+        {
+            if (_authorizationResult.Code != null &&
+               !string.IsNullOrEmpty(_authorizationResult.Code) &&
+               _authorizationResult.Code.StartsWith(BrokerParameter.AuthCodePrefixForEmbeddedWebviewBrokerInstallRequired, StringComparison.OrdinalIgnoreCase) ||
+               _authorizationResult.Code.StartsWith(ServiceBundle.Config.RedirectUri, StringComparison.OrdinalIgnoreCase))
+            {
+                AuthenticationRequestParameters.RequestContext.Logger.Info(LogMessages.BrokerInvocationRequired);
+                return true;
+            }
+
```
MarkZuber: nit: remove extra blank lines

src/Microsoft.Identity.Client/Platforms/iOS/iOSPlatformProxy.cs
```diff
@@ -156,5 +157,15 @@ public override int GetMatsOsPlatformCode()
         }
 
         protected override IFeatureFlags CreateFeatureFlags() => new iOSFeatureFlags();
+
+        public override IBroker CreateBroker()
+        {
+            if (base.CreateBroker() == base.OverloadBrokerForTest)
```
MarkZuber: Why do you need to call base.CreateBroker() instead of just checking if OverloadBrokerForTest is non-null?  If someone has set the test hook, shouldn't you just use it?

jennyf19: definitely . thanks.

---
In reply to: [308433527](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1295#discussion_r308433527) [](ancestors = 308433527)

PR #1302, opened 2019-07-30
Set exception on TCS so it is thrown in correct place

crmann1:


src/Microsoft.Identity.Client/Platforms/net45/WebUI.cs
```diff
@@ -35,8 +35,18 @@ internal abstract class WebUI : IWebUI
...
+                    authorizationResult = Authenticate(authorizationUri, redirectUri);
+                }
+                catch (Exception e)
+                {
+                    // Need to catch the exception here and put on the TCS which is the task we are waiting on so that
+                    // the exception comming out of Authenticate is correctly thrown.
+                    ((TaskCompletionSource<object>)tcs).TrySetException(e);
+                }
+
+                 ((TaskCompletionSource<object>)tcs).TrySetResult(null);
```
crmann1: I am actually going to move this onto line 41

bgavrilMS: Was about to comment on that.

crmann1:
> This looks good, we were not handling the exception properly. Have you met errors here?

Yes we have, network down and some others so some people are hitting it.

Issue #1276, opened 2019-07-19
[Bug] MSAL.NET throws a MsalClientException and not a MsalServiceException for the reset password exception from AAD B2C


jennyf19:
**Which Version of MSAL are you using ?**
First noticed in 4.1, but not sure which version introduced the change in behavior

**Platform**
mobile, desktop, web app

**Repro**

example, include this code:
```CSharp
catch (MsalServiceException ex)
          {
                try
                {
                          if (ex.Message.Contains("AADB2C90118"))
                          {
                                   authResult = await (app as 
                                         publicClientApplication).AcquireTokenInteractive(App.ApiScopes)
                                        .WithParentActivityOrWindow(new WindowInteropHelper(this).Handle)
                                        .WithAccount(GetAccountByPolicy(accounts, App.PolicySignUpSignIn))
                                        .WithPrompt(Prompt.SelectAccount)
                                        .WithB2CAuthority(App.AuthorityResetPassword)
                                        .ExecuteAsync();
}
```
Can be tried w/B2C desktop sample.

**Expected behavior**
MSAL.NET should throw a `MsalServiceException` with the following error message:
`AADB2C90118: The user has forgotten their password.`

**Actual behavior**
MSAL.NET throws a `MsalClientException` instead

bgavrilMS:
So what is happening is: 

- Upon clicking on that URI, the server returns an url containg "ErrorCode=access_denied" and "ErrorDescription = AADB2C90118: The user has forgotten their password"
- This translates to an AuthorizationResult with Status = ProtocolError
- Based on this AuthorizationResult, we throw an MsalClient exception (logic [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/caa9dfd1f15452c70eb90d3020b86525d091beee/src/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs#L270))

Note sure when such a regression would have occured, since the logic that throws exceptions has not been touched since we split ADAL from MSAL. 

I agree that we should look at these exceptions again - this is indeed a server exception.

bgavrilMS:
Workaround: catch MsalException instead.

jennyf19:
@bgavrilMS Thanks for investigating. I'll update the sample

jennyf19:
[...]

PR #1305, opened 2019-07-31
change client exception back to service

jennyf19:
[1276](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1276)

@jmprieur and @henrik-me we need to discuss this change and versioning.

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
@@ -801,7 +801,7 @@ namespace Microsoft.Identity.Test.Unit.PublicApiTests
                         .ExecuteAsync(CancellationToken.None)
                         .ConfigureAwait(false);
                 }
-                catch (MsalServiceException exc)
+                catch (MsalClientException exc)
                 {
                     Assert.IsNotNull(exc);
                     Assert.AreEqual("authentication_canceled", exc.ErrorCode);
--- a/tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
@@ -801,7 +801,7 @@ namespace Microsoft.Identity.Test.Unit.PublicApiTests
                         .ExecuteAsync(CancellationToken.None)
                         .ConfigureAwait(false);
                 }
-                catch (MsalClientException exc)
+                catch (MsalServiceException exc)
                 {
                     Assert.IsNotNull(exc);
                     Assert.AreEqual("authentication_canceled", exc.ErrorCode);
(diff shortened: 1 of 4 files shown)


tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -801,7 +801,7 @@ public async Task AcquireTokenWithAuthenticationCanceledTestAsync()
                         .ExecuteAsync(CancellationToken.None)
                         .ConfigureAwait(false);
                 }
-                catch (MsalClientException exc)
+                catch (MsalServiceException exc)
                 {
                     Assert.IsNotNull(exc);
                     Assert.AreEqual("authentication_canceled", exc.ErrorCode);
```
henrik-me: >Assert.AreEqual("authentication_canceled", exc.ErrorCode); [](start = 20, length = 58)

I thought this was a client exception?

henrik-me: I'm puzzled that there are not other tests that are failing as part of this change... 

---
In reply to: [309523923](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1305#discussion_r309523923) [](ancestors = 309523923)

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -817,6 +817,47 @@ public async Task AcquireTokenWithAuthenticationCanceledTestAsync()
             Assert.Fail("Should not reach here. Exception was not thrown.");
         }
 
+        [TestMethod]
+        [Description("Test for AcquireToken with user resetting password")]
+        public async Task B2CAcquireTokenWithResetPasswordTestAsync()
```
henrik-me: >B2CAcquireTokenWithResetPasswordTestAsync [](start = 26, length = 41)

Nice to see a good regression test!

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
jennyf19: @henrik-me I'm puzzled too..only two tests failed as a result of the change. I'll try out the scenario above and report back tomorrow

jennyf19: @henrik-me - take a look at this now. we now get a client exception, but not sure if this is the best way to handle it...also, before this change, were we doing service exceptions? i agree, does not seem right.

henrik-me: Did you confirm that this is the actual user canceling/closing the dialog exception? If yes, then I believe the fix is right as it's ensures the abort is a client exception. Do we have any other scenarios that will cause this exception to be thrown? I'm wondering if any other scenario's exists that will go through this code path... 

---
In reply to: [309890518](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1305#discussion_r309890518) [](ancestors = 309890518)

jennyf19: @henrik-me i tested that actually canceling/closing the dialog throws this exception, so anytime the user cancels, not sure how ests defines that, we will throw a client exception here.

henrik-me: ESTS doesn't define that. we do. Which is why I'm curios if we will ever end in a situation where we will see a service exception from that piece of code. The service exceptions are thrown elsewhere. Anyway, I believe this fix is safe and I can't currently imagine there is a real breaking change anywhere.

jennyf19:
[test pass here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=378513)
[ui tests pass here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=378515)

Issue #592, opened 2018-07-16
Missing MsalClientException.NetworkNotAvailableError on iOS & Android


KarinBerg:
For platform Xamarin.iOS and Xamarin.Android there is no check for an available network connection in class WebUI. So in this case an HttpRequestException is thrown and not the expected MsalClientException with ErrorCode "NetworkNotAvailableError".

Please add an network connection check for this platforms to be consistent over all platforms.

bgavrilMS:
+ @jmprieur for triage

henrik-me:
Will require a bit of refactoring. I assume this is possible especially with the latest updates in the dev branch. @bgavrilMS can you pls. take a look at this?

henrik-me:
@bgavrilMS : still relevant? Putting this in 4.3.
@jmprieur : fix or close? (#Supportability)

bgavrilMS:
I had a look at this and I propose we fix it by removing the custom exception from .net (this is a breaking behavioural change though), but then again so is the proposed change.

The reasons: 

- Network is down exceptions are fairly common and app developers should treat them. Even a generic "Something went wrong" message would do. 

- MSAL makes many calls to the network, such as discovery calls. It would be a perf hit to try to read the state of the network before each HTTP call. 

- Catching HTTP exceptions and repackaging them would also be a breaking change to our 

- Detecting that the network is down is tricky - see how it's been done in Xamarin Essentials - https://github.com/xamarin/Essentials/blob/master/Xamarin.Essentials/Connectivity/Connectivity.android.cs - also note that it does not guarantee that a device has Internet, it just shows that a connection is available. 

All in all, I don't think it's MSAL's responsibility to provide internet detection capabilities. App developers should not have to have special catch blocks for MSAL. 

This is easier: 

```csharp
try
{
   // call MSAL
   // call some other logic that makes HTTP calls
}
catch HttpException
  // Handle all unexpected exceptions, including HTTP issues
```

Than custom handling: 


```csharp
try
{
   // call MSAL
   // call some other logic that makes HTTP calls
}
catch (MsalClientException ex) when ex.ErrorCode == NetworkIsDown
{
   // handle MSAL
}
catch HttpException
{
 // handle all other Http issues
}
```

jmprieur:
This makes sense to me.
Let's provide guidance, then. Let's show this in our sample and the Mobile scenario landing page, as well as the error handling wiki/docs.ms page?

[...]

PR #1306, opened 2019-07-31
Stop custom handling network down error

bgavrilMS:
…Essentials for that.

Fixes bug https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/592 by providing a **consistent** approach to HTTP errors. See more details in the bug.

Note that this is theoretically a breaking change because we are changing exception logic, but in practice it should not make a difference, because if the network is down, MSAL will fail with an Http exception when making a discovery call.

### DIFF
--- a/src/Microsoft.Identity.Client/MsalError.cs
+++ b/src/Microsoft.Identity.Client/MsalError.cs
@@ -470,6 +470,7 @@ namespace Microsoft.Identity.Client
         /// <para>Mitigation [App development]</para> In the application you could either inform the user that there are network issues
         /// or retry later
         /// </summary>
+        [Obsolete("MSAL no longer throws this error - it will allow the HttpClient exceptions to propagate. App developers may write their own logic for detecting access to the network issues, for example by using Xamarin.Essentials. ")]
         public const string NetworkNotAvailableError = "network_not_available";
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -470,6 +470,7 @@ public static class MsalError
         /// <para>Mitigation [App development]</para> In the application you could either inform the user that there are network issues
         /// or retry later
         /// </summary>
+        [Obsolete("MSAL no longer throws this error - it will allow the HttpClient exceptions to propagate. App developers may write their own logic for detecting access to the network issues, for example by using Xamarin.Essentials. ")]
```
MarkZuber: Is this change related to a bug/issue around this area?  Curious what prompted the change.

jennyf19: @MarkZuber I think this ->[592](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/592)

henrik-me:
@jmprieur : Do you consider this a breaking change that would require a major version bump? I'm generally ok with this change as long as we provide guidance on this in the wiki and to be considered to include handling in samples.

PR #1307, opened 2019-07-31
Fix regression from PR #1275

marcre:
Fix a regression from [PR#1275](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1275) where if certificate.PrivateKey returns an instance of RSACryptoServiceProvider using a legacy CSP that doesn't support SHA256 the signature fails.  The fix is to use the existing helper method that checks for unsupported provider types and create a new instance of RSACryptoServiceProvider with a supported provider type.

### DIFF
--- a/src/Microsoft.Identity.Client/Platforms/net45/NetDesktopCryptographyManager.cs
+++ b/src/Microsoft.Identity.Client/Platforms/net45/NetDesktopCryptographyManager.cs
@@ -122,7 +122,9 @@ namespace Microsoft.Identity.Client.Platforms.net45
         /// <returns><see cref="RSACryptoServiceProvider"/> initialized with private key from <paramref name="certificate"/></returns>
         private static RSACryptoServiceProvider GetCryptoProviderForSha256(X509Certificate2 certificate)
         {
-            return (RSACryptoServiceProvider) certificate.PrivateKey;
+            // We need to call the other GetCryptoProviderForSha256 method in case certificate.PrivateKey is using
+            // one of the providers that doesn't support SHA256
+            return GetCryptoProviderForSha256((RSACryptoServiceProvider)certificate.PrivateKey);
         }
 
         // Copied from ACS code


src/Microsoft.Identity.Client/Platforms/net45/NetDesktopCryptographyManager.cs
```diff
@@ -122,7 +122,9 @@ public byte[] SignWithCertificate(string message, X509Certificate2 certificate)
         /// <returns><see cref="RSACryptoServiceProvider"/> initialized with private key from <paramref name="certificate"/></returns>
         private static RSACryptoServiceProvider GetCryptoProviderForSha256(X509Certificate2 certificate)
         {
-            return (RSACryptoServiceProvider) certificate.PrivateKey;
+            // We need to call the other GetCryptoProviderForSha256 method in case certificate.PrivateKey is using
+            // one of the providers that doesn't support SHA256
+            return GetCryptoProviderForSha256((RSACryptoServiceProvider)certificate.PrivateKey);
```
trwalke: @bgavrilMS perhaps we can expand our E2E test coverage to include certificates with various private key types?

jennyf19:
@MarkZuber is this going in 4.3?

MarkZuber:
> @MarkZuber is this going in 4.3?

Seems like it meets the bar for 4.3.  @henrik-me ?

trwalke:
We are not ready to take this yet. It needs to be verified before we make the changes.

MarkZuber:
> We are not ready to take this yet. It needs to be verified before we make the changes.

OK.  then please don't merge this until we have that validation and this can move to 4.3.1.  thx.

trwalke:
yeah, Brent and I just had a conversation about it and he offered refactored code that will optimize our signing operation. Im going to send a demo package to @marcre

PR #1314, opened 2019-08-01
Fix regression from PR #1275

trwalke:
Recreating PR as I do not have permission to modify @marcre's fork

"Fix a regression from PR#1275 where if certificate.PrivateKey returns an instance of RSACryptoServiceProvider using a legacy CSP that doesn't support SHA256 the signature fails. The fix is to use the existing helper method that checks for unsupported provider types and create a new instance of RSACryptoServiceProvider with a supported provider type." - @marcre

Previous PR https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1307


trwalke:
@brentschmaltz has suggested additional changes to be made on top of @marcre's fix so those are reflected in this PR

henrik-me:
Did you validate the original scenario?

trwalke:
> Did you validate the original scenario?

Yes, i asked the customer to validate the fix with a private preview package.

henrik-me:
Can we pls. ensure we have our own regression test?

trwalke:
> Can we pls. ensure we have our own regression test?
Ok, We have a test for certs already but our test cert did not have the same properties. Basically we need to run that test multiple times with different certificate types. Ill see if I can create some certs with various private key types

bgavrilMS:
Code look good, needs more E2E tests, potentially one for net48 as well.

PR #1318, opened 2019-08-06
yaml build

bgavrilMS:


bgavrilMS:
just an experiment...

jmprieur:
was the experiement succesful @bgavrilMS ?

bgavrilMS:
The experiment was "move to yaml builds" to adopt the modern technique of "build definition as code". Seems easy enough to do, we would need to spend a day or so to move all our builds from the graphical definition to yaml.

PR #1322, opened 2019-08-06
[DO NOT MERGE / WIP] PoP

bgavrilMS:


src/Microsoft.Identity.Client/Internal/PoP/IPoPCryptoProvider.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Internal.PoP
+{
+    public interface IPoPCryptoProvider
```
bgavrilMS: I did not want to take a dependency on Wilson just yet, so I am separating Wilson behind this interface.

src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -297,6 +301,49 @@ internal async Task ResolveAuthorityEndpointsAsync()
...
+            //var parameters = key.Rsa == null ? key.Parameters : key.Rsa.ExportParameters(false);
+            //return "{\"kty\":\"RSA\",\"n\":\"" + Base64UrlEncoder.Encode(parameters.Modulus) + "\",\"e\":\"" + Base64UrlEncoder.Encode(parameters.Exponent) + "\",\"alg\":\"" + algorithm + "\",\"kid\":\"" + key.KeyId + "\"}";
+
+
+            // return "{\"kty\":\"RSA\",\"alg\":\"" + algorithm + "\",\"kid\":\"" + keyId + "\"}";
+            return "{\"kid\":\"" + keyId + "\"}";
+
+        }
+
+        private static string CreateCnfRequest(IPoPCryptoProvider popCryptoProvider)
```
bgavrilMS: This is the where I create a req_cnf according to https://tools.ietf.org/html/draft-ietf-oauth-pop-key-distribution-07#page-9

bgavrilMS: Sadly eSTS replies with a bearer token :(

src/Microsoft.Identity.Client/Internal/PoP/IPoPCryptoProvider.cs
yordan-msft: why not just take a string as 'Key' param?

src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
```diff
@@ -201,7 +201,7 @@ public JWTHeaderWithCertificate(ClientCredentialWrapper credential, bool sendCer
 #endif
             }
 
-            [DataMember(Name = JsonWebTokenConstants.ReservedHeaderParameters.X509CertificateThumbprint)]
```
bgavrilMS: Minor refactoring, just renaming a constant.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -283,8 +285,12 @@ async Task<MsalAccessTokenCacheItem> ITokenCacheInternal.FindAccessTokenAsync(Au
                     MsalErrorMessage.MultipleTokensMatched);
             }
 
+            msalAccessTokenCacheItem = FilterATByKeyId(msalAccessTokenCacheItem, requestParams);
```
bgavrilMS: PoP tokens are bound to a kid, so if a different public / private key pair is used - do not return the token.

yordan-msft: what if there are multiple tokens bound to different KIDs? Possible?
Code will throw exception before filter is applied.

bgavrilMS: It is possible to have multiple ATs bound to the same key, for example if they are for different client IDs or for different enviroments or different scopes.

But at this point in the `FindAccessToken` logic it's guaranteed that there are 0 or 1 tokens found in the cache. If more are found, an exception is thrown (see line 283).

bgavrilMS: In this case, the access tokens will override each other. I chose this strategy because with PoP the key should be stable, i.e. one key per machine. The key can be lost, => new one will be generated, so we need to check AT is bound to same key. 

A different strategy would be to add `kid` to the access token cache key. This is considerably more complex and will break cache schema, i.e. old MSALs running side by side with new MSALs will fail. 

Third idea is to store PoP tokens separately from Bearer tokens. PoP tokens cache key would include `kid`. Adds quite a bit of complexity, but also flexibility. 

What do you think? Is it worth maybe discussing around this during spec review tomorrow?

jennyf19:
> using System;

header info

---
Refers to: src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs:1 in fa667de. [](commit_id = fa667de097acb5911823f0fca526ecfbc85599da, deletion_comment = False)

src/client/Microsoft.Identity.Client/PoP/HttpPopClaimTypes.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace Microsoft.Identity.Client.PoP
+{
+    internal static class HttpPopClaimTypes
```
jennyf19: >p [](start = 32, length = 1)

nit: elsewhere you have PoP

jennyf19:
> namespace Microsoft.Identity.Client.PoP

header info

---
Refers to: src/client/Microsoft.Identity.Client/PoP/HttpPopClaimTypes.cs:1 in fa667de. [](commit_id = fa667de097acb5911823f0fca526ecfbc85599da, deletion_comment = False)

src/client/Microsoft.Identity.Client/Internal/JsonWebTokenConstants.cs
```diff
@@ -31,13 +24,28 @@ internal class ReservedClaims
             public const string JwtIdentifier = "jti";
         }
 
-        internal class ReservedHeaderParameters
+        internal static class ReservedHeaderParameters
```
jennyf19: >static [](start = 17, length = 6)

does this have to static?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -313,6 +319,31 @@ async Task<MsalAccessTokenCacheItem> ITokenCacheInternal.FindAccessTokenAsync(Au
...
+
+            string requestKid = authenticationRequest.AuthenticationScheme.KeyId;
+            if (string.IsNullOrEmpty(item.Kid) && string.IsNullOrEmpty(requestKid))
+            {
+                authenticationRequest.RequestContext.Logger.Verbose("Bearer token found");
+                return item;
+            }
+
+            if (string.Equals(item.Kid, requestKid))
+            {
```
jennyf19: how is the key id generated for the pop token? or we can sync offline on this.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -313,6 +319,31 @@ async Task<MsalAccessTokenCacheItem> ITokenCacheInternal.FindAccessTokenAsync(Au
...
+        private MsalAccessTokenCacheItem FilterATByKeyId(MsalAccessTokenCacheItem item, AuthenticationRequestParameters authenticationRequest)
+        {
+            if (item == null)
+            {
+                return null;
+            }
+
+            string requestKid = authenticationRequest.AuthenticationScheme.KeyId;
+            if (string.IsNullOrEmpty(item.Kid) && string.IsNullOrEmpty(requestKid))
+            {
```
jennyf19: is that the only way to identify a bearer token?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -166,7 +168,7 @@ internal AuthenticationResult(MsalAccessTokenCacheItem msalAccessTokenCacheItem,
         /// </example>
         public string CreateAuthorizationHeader()
         {
-            return Oauth2AuthorizationHeader + AccessToken;
+            return _authenticationScheme?.AuthorizationHeaderPrefix ?? "Bearer" + " " + AccessToken;
```
jennyf19: >Bearer [](start = 72, length = 6)

this will only and always be for a bearer token? Can you use the BearerAuthSchemePrevix value here instead "Bearer"

jmprieur: why would an AuthorizationHeaderPrefix be null?

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private readonly Uri _uriToBind;
+        private readonly IPoPCryptoProvider _popCryptoProvider;
+
+        /// <summary>
+        /// Creates POP tokens, i.e. tokens that are bound to an HTTP request and are digitally signed.
+        /// </summary>
+        /// <param name="uriToBind">Uri describing the HTTP request</param>
+        /// <param name="popCryptoProvider">Crypto used to sign the POP token</param>
+        public PoPAuthenticationScheme(Uri uriToBind, IPoPCryptoProvider popCryptoProvider)
```
jmprieur: No HTTP verb?

jmprieur: We also considered passing the HttpRequest (which would contain both the URI and the Http Verb. Would that make sense?

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -21,6 +23,23 @@ internal AbstractPublicClientAcquireTokenParameterBuilder(IPublicClientApplicati
             PublicClientApplicationExecutor = publicClientApplicationExecutor;
         }
 
+        public AbstractPublicClientAcquireTokenParameterBuilder<T> WithPoPAuthenticationScheme(Uri uri) // TODO: naming and comments
```
jmprieur: **Proposal for XML comment**
summary: Modifies the token acquisition request so that the acquired token is a Proof of Possession token (PoP), rather than a bearer token. A key/pair is generated by MSAL.NET. See https://aka.ms/msal-net-pop
param: uri: URI for which to request a Proof of Possession token is aquired
See also: AuthorizationResult.CreateAuthorizationHeader

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -21,6 +23,23 @@ internal AbstractPublicClientAcquireTokenParameterBuilder(IPublicClientApplicati
...
+        {
+            CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithScheme);
+            CommonParameters.AuthenticationScheme = new PoPAuthenticationScheme(uri, null); // TODO add a default crypto provider
+
+            return this;
+        }
+
+        // Allows testing the PoP flow with any crypto. Consider making this public.
+        internal AbstractPublicClientAcquireTokenParameterBuilder<T> WithPoPAuthenticationScheme(Uri protectedUri, IPoPCryptoProvider popCryptoProvider)
```
jmprieur: **Proposal for XML comment**
summary: Modifies the token acquisition request so that the acquired token is a Proof of Possession token (PoP), rather than a bearer token. The keys used in signing the Pop token are provided by popCryptoProvider. See https://aka.ms/msal-net-pop
uri: URI for which to request a Proof of Possession token is aquired
popCryptoProvider: key provider used to sign the Pop token.
See also: AuthorizationResult.CreateAuthorizationHeader

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -166,7 +168,7 @@ internal AuthenticationResult(MsalAccessTokenCacheItem msalAccessTokenCacheItem,
         /// </example>
         public string CreateAuthorizationHeader()
```
jmprieur: Shouldn't we also expose a string property with the Token Type (AuthorizationHeaderPrefix) ?
I'm thinking supportability here, if customers don't realize they got a Pop token (because they called a method, and have not realized the pop modifier was used) and try to feed it into an Authorization header as a bearer token. Under the debugger they would immediately see that this is a Pop token?

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -21,6 +23,23 @@ internal AbstractPublicClientAcquireTokenParameterBuilder(IPublicClientApplicati
             PublicClientApplicationExecutor = publicClientApplicationExecutor;
         }
 
+        public AbstractPublicClientAcquireTokenParameterBuilder<T> WithPoPAuthenticationScheme(Uri uri) // TODO: naming and comments
+        {
+            CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithScheme);
```
MarkZuber: do we want to know which scheme in the telemetry?  Perhaps "WithPoPScheme"?  (and we can look to make WithBearerScheme the default)

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
MarkZuber: Agree with @jmprieur here.  the _authenticationScheme can't be null (you validate that in the constructor).  So this should always be:  return _authenticationScheme.AuthorizatonHeaderPrefix;

src/client/Microsoft.Identity.Client/BearerAuthenticationScheme.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Microsoft.Identity.Client.Cache.Items;
+
+namespace Microsoft.Identity.Client
+{
+    internal class BearerAuthenticationScheme : IAuthenticationScheme
+    {
+        public string AuthorizationHeaderPrefix => "Bearer";
```
MarkZuber: public const string

MarkZuber: In the earlier code (CreateAuthorizationHeader) you're doing "Bearer" + " " + AccessToken.  So do we need a method on the IAuthenticationScheme for GetAuthorizationHeader() that does the proper formatting here?  This will just get you the prefix which isn't enough to actually create the header in AuthenticationResult.CreateAuthorizationHeader().

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -81,8 +85,13 @@ internal string TenantId
...
         internal string ExpiresOnUnixTimestamp { get; set; }
         internal string ExtendedExpiresOnUnixTimestamp { get; set; }
-        public string UserAssertionHash { get; set; }
+        internal string UserAssertionHash { get; set; }
  
+        /// <summary>
+        /// Used when the token is bound to a public / private key pair which is identified by a key id (kid). 
+        /// Currently used by PoP tokens
+        /// </summary>
+        internal string Kid { get; private set; }
```
MarkZuber: i know the spec calls it a "kid" (because they're trying to save json data transfer space).  Can we call it KeyId in our code?

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -128,6 +137,7 @@ internal static MsalAccessTokenCacheItem FromJObject(JObject j)
                 ExpiresOnUnixTimestamp = expiresOn.ToString(CultureInfo.InvariantCulture),
                 ExtendedExpiresOnUnixTimestamp = extendedExpiresOn.ToString(CultureInfo.InvariantCulture),
                 UserAssertionHash = JsonUtils.ExtractExistingOrEmptyString(j, StorageJsonKeys.UserAssertionHash),
+                Kid = JsonUtils.ExtractExistingOrEmptyString(j, StorageJsonKeys.Kid)
```
MarkZuber: Will need to update the cache spec to include this new storage key as a reserved keyword.

src/client/Microsoft.Identity.Client/IAuthenticationScheme.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Prefix for the HTTP header that has the token. E.g. "Bearer" or "POP"
+        /// </summary>
+        string AuthorizationHeaderPrefix { get; }
+
+        /// <summary>
+        /// Extra paramters that are added to the request to the /token endpoint
+        /// </summary>
+        /// <returns>Name and values of params</returns>
+        System.Collections.Generic.IDictionary<string, string> GetTokenRequestParams();
```
MarkZuber: add a using block to the top for System.Collections.Generic and just have IDictionary<string, string> here.

src/client/Microsoft.Identity.Client/Internal/JsonWebTokenConstants.cs
MarkZuber: It should be since it's just a class holder for consts.

src/client/Microsoft.Identity.Client/PoP/HttpPopClaimTypes.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// Access token with response cnf
+        /// https://tools.ietf.org/html/draft-ietf-oauth-signed-http-request-03#section-3
+        /// </summary>
+        public const string At = "at";
+
+        /// <summary>
+        /// Http method (GET or POST)
+        /// https://tools.ietf.org/html/draft-ietf-oauth-signed-http-request-03#section-3
+        /// </summary>
+        public const string M = "m";
```
MarkZuber: suggestion:  since these are constants in our code, should we spell them out a bit more for readability?  For example, call this "Method" instead of "M"?

src/client/Microsoft.Identity.Client/PoP/HttpPopClaimTypes.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// Uri path
+        /// https://tools.ietf.org/html/draft-ietf-oauth-signed-http-request-03#section-3
+        /// </summary>
+        public const string P = "p";
+
+        #endregion
+    }
+
+    internal static class PopRequest
```
MarkZuber: move class to its own file since it's not nested.

src/client/Microsoft.Identity.Client/PoP/IPoPCryptoProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+        ///  "n": "0vx7ag...",
+        ///  "e":"AQAB",
+        ///  "alg":"RS256",
+        ///  "kid":"2011-04-29"}
+        /// </summary>
+        /// <returns></returns>
+        string GetPublicKeyJwk();
+
+        /// <summary>
+        /// Sings the byte array using the private key
```
MarkZuber: typo:  Signs

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
```
MarkZuber: need open source header (ensure they're in every file)

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="uriToBind">Uri describing the HTTP request</param>
+        /// <param name="popCryptoProvider">Crypto used to sign the POP token</param>
+        public PoPAuthenticationScheme(Uri uriToBind, IPoPCryptoProvider popCryptoProvider)
+        {
+            _uriToBind = uriToBind;
+            _popCryptoProvider = popCryptoProvider ??
+                throw new ArgumentNullException(nameof(popCryptoProvider));
+        }
+
+        public string AuthorizationHeaderPrefix { get { return "PoP"; } }
```
MarkZuber: suggestion:  public string AuthorizationHeaderPrefix => "PoP";

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
```diff
@@ -0,0 +1,74 @@
...
+            return CreateJWS(payload.ToString(Json.Formatting.None), header.ToString(Json.Formatting.None));
+        }
+
+        /// <summary>
+        /// Creates a JWS (json web signature) as per: https://tools.ietf.org/html/rfc7515
+        /// Format: header.payload.signed_payload
+        /// </summary>
+        private string CreateJWS(string payload, string header)
+        {
+            var message = Base64UrlHelpers.Encode(Encoding.UTF8.GetBytes(header)) + "." + Base64UrlHelpers.Encode(payload);
```
MarkZuber: instead of all of the string concat (with +) can you do a string.Format() here to avoid multiple buffer copies?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -313,6 +319,31 @@ async Task<MsalAccessTokenCacheItem> ITokenCacheInternal.FindAccessTokenAsync(Au
             return null;
         }
 
+        private MsalAccessTokenCacheItem FilterATByKeyId(MsalAccessTokenCacheItem item, AuthenticationRequestParameters authenticationRequest)
```
MarkZuber: nit:  avoid abbreviations.  FilterAccessTokenByKeyId

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -313,6 +319,31 @@ async Task<MsalAccessTokenCacheItem> ITokenCacheInternal.FindAccessTokenAsync(Au
...
+                return item;
+            }
+
+            if (string.Equals(item.Kid, requestKid))
+            {
+                authenticationRequest.RequestContext.Logger.Verbose("PoP token found");
+                return item;
+            }
+
+            authenticationRequest.RequestContext.Logger.Info(
```
MarkZuber: is the $"" style string going to throw CultureInfo violations in the release build due to not using String.Format?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: Sure, let's sync. I haven't coded this part yet. It can be fairly complex, so for now I'll focus on a simple key generation strategy for out .net classic / Windows platform:

- generate a public / private key pair
- store it in a file on disk, encrypted with DPAPI

In the future we need to figure out a strategy for: 

- Linux, Mac (no DPAPI there)
- Confidential Clients (they can't rely on files, since they can be distributed systems)
- Mobile platforms

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: This is a good point. 

I think I will add an explicit "TokenType" in the cache. After all, this is returned by AAD in the token response. 

For example, AAD will have the ability to issue SSH "tokens", which are also bound to keys.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: There is actually a ctor which we expose so that our customers can test. I can't expose _authenticationScheme there. I will expose the actual prefix though.

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
bgavrilMS: Ack. I remember that we wanted to have the verb as being optional, so this is the minimum required to get this to work.

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
bgavrilMS: Yes, I've updated it already, it is part of the POP spec.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: Agreed

src/client/Microsoft.Identity.Client/BearerAuthenticationScheme.cs
bgavrilMS: We don't need this abtraction for Bearer, POP or SSH. For SSH calling `CreateAuthorizationHeader()` would just throw.

henrik-me:
I would like to make sure that we for now mark all public changes for this effort as experimental/preview as they are subject to change

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -45,7 +45,8 @@ public partial class AuthenticationResult
             IAccount account,
             string idToken,
             IEnumerable<string> scopes,
-            Guid correlationId)
+            Guid correlationId,
+            string tokenType = "Bearer")
```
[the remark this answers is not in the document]
bgavrilMS: Ack

henrik-me:
@bgavrilMS : Seems like the tokentype parameter was not marked as experimental in the ctr or the ctr with the param wasn't made internal. Is that on purpose? Should I expect a PR to update?

bgavrilMS:
@henrik-me - I can update that ctor to mark token type as experimental. Please note that the ctor is there only to help people test, we don't expect developers to new up an AuthenitcationResult in other scenarios. So the ctor is just an enumeration of all the public properties, experimental or not.

bgavrilMS:
@henrik-me  - marked param as experimental

PR #1323, opened 2019-08-07
Jennyf/update dev app to handle IsSystemWebviewAvailable

jennyf19:


jennyf19:
build passed [here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=381100)

PR #1325, opened 2019-08-07
broker update

jennyf19:


jennyf19:
[pr build passed here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=381151)

Issue #1328, opened 2019-08-08
[Bug] Claims not sent to /authroize endpoint 


bgavrilMS:
**Which Version of MSAL are you using ?**
4.3

**Platform**
all

**Problem**
According to the OIDC spec https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter claims is a parameter relevant to the /authorize endpoint. 

MSAL only sends this param to the /token endpoint. 

```csharp
pca.AcquireTokenInteractive(...).WithClaims(claims);
```

Actual: claims param sent in request to the **/token** endpoint but not to the **/authorize** endpoint
Expected: claims param needs to be sent to both

bgavrilMS:
Spec on claims: https://identitydivision.visualstudio.com/DefaultCollection/IDDP/_git/first-party-docs/pullrequest/537?_a=files&path=%2Fdocfx-src%2Fdevex%2Fsdks%2Fclientcapabilities.md
 
Claims should actually be sent to both /authorize and /token endpoints

jmprieur:
@bgavrilMS @henrik-me @shoatman : does it have an impact on AdditionalClaimsToConsentTo?

jmprieur:
makes MSAL.NET OIDC compliant

henrik-me:
Included in the 4.4.0 release

PR #1330, opened 2019-08-08
Send claims as query param to /authorize endpoint as per OIDC spec

bgavrilMS:
Fix for #1328


bgavrilMS:
@shoatman - Could you confirm the behaviour in this PR since you have a lot of context around claims? 

MSAL is sending the Claims param only to the /token endpoint. This PR ensures Claims are also sent to the /authorize endpoint.

PR #1337, opened 2019-08-12
Fix for #891 - JWT audience = tenanted oauth endpoint

bgavrilMS:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/891
Includes a regression test

### DIFF
--- a/src/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs
+++ b/src/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs
@@ -8,6 +8,8 @@ using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.TelemetryCore.Internal;
+using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Instance
 {
@@ -40,10 +42,6 @@ namespace Microsoft.Identity.Client.Instance
 
             requestContext.Logger.Info(LogMessages.ResolvingAuthorityEndpointsFalse);
 
-            var authorityUri = new Uri(authorityInfo.CanonicalAuthority);
-            string path = authorityUri.AbsolutePath.Substring(1);
-            string tenant = path.Substring(0, path.IndexOf("/", StringComparison.Ordinal));
-
             var endpointManager = OpenIdConfigurationEndpointManagerFactory.Create(authorityInfo, _serviceBundle);
 
             string openIdConfigurationEndpoint = await endpointManager.ValidateAuthorityAndGetOpenIdDiscoveryEndpointAsync(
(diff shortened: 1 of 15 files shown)


src/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs
```diff
@@ -74,34 +72,41 @@ public AuthorityEndpointResolutionManager(IServiceBundle serviceBundle, bool sho
...
 
             Add(authorityInfo, userPrincipalName, endpoints);
             return endpoints;
         }
 
-        internal string ReplaceNonTenantSpecificValueWithTenant(TenantDiscoveryResponse endpoints, string tenant)
+        // Used in WithCertificate to create an audience claim in JWT sent by MSAL to EVO
+        // ADAL uses the token endpoint as audience (tenanted or not)
+        // MSAL had been using the issuer, which does not work when tenantnless. But continue to use issuer for ADFS and B2C.
+        private static string GetSelfSignedJwtAudience(string issuer, string tokenEndpoint, string tenantId, AuthorityType authorityType)
```
bgavrilMS: This is the actual fix. If AAD, use the token endpoint. Else use the issuer. 

Needs more testing with ADFS and B2C to understnad if we can use the token endpoint consistenty.

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/StringExtensionTests.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    [TestClass]
+    public class StringExtensionTests
+    {
+        [TestMethod]
+        public void StringReplace()
+        {
+            string original1 = "https://login.microsoftonline.com/{tenant}/oauth2";
+            string original2 = "https://login.microsoftonline.com/{tenant_id}/oauth2";
+            string expected = "https://login.microsoftonline.com/common/oauth2";
```
MarkZuber: you should have tests with multiple {xxx} values in the template string to ensure it handles multiple insertion as well.

src/Microsoft.Identity.Client/Utils/StringExtensions.cs
```diff
@@ -16,5 +17,72 @@ public static byte[] ToByteArray(this string stringInput)
...
+                return str;
+            }
+
+            if (oldValue == null)
+            {
+                throw new ArgumentNullException(nameof(oldValue));
+            }
+            if (oldValue.Length == 0)
+            {
+                throw new ArgumentException("String cannot be of zero length.");
```
MarkZuber: "oldValue cannot be of zero length"

src/Microsoft.Identity.Client/Utils/StringExtensions.cs
```diff
@@ -16,5 +17,72 @@ public static byte[] ToByteArray(this string stringInput)
         {
             return new UTF8Encoding().GetBytes(stringInput);
         }
+
```
MarkZuber: question:  are you concerned about memory fragmentation here for just doing a "while string.contains(oldValue) { string.replace(oldValue, newValue) } type of algorithm?  these strings aren't super long and don't have a lot of templatization so just using the existing string replace instead of building it up via stringbuilders might be less error prone

src/Microsoft.Identity.Client/Utils/StringExtensions.cs
bgavrilMS: Will update this algorithm.

src/Microsoft.Identity.Client/Utils/StringExtensions.cs
bgavrilMS: I'll get something simpler.

bgavrilMS:
UI Automation failing because of a Facebook B2C login - unrelated.

Issue #1345, opened 2019-08-17
[Bug] MSAL does not allow you to bypass the internal token cache like ADAL does


bgavrilMS:
**Which Version of MSAL are you using ?**
4.3.1

**Platform**
all

** Issue **
In ADAL, we could disable the token cache like this: 

```csharp
 var authContext = new AuthenticationContext(authority, tokenCache: null); 
```

In MSAL this is not possible, although the internal code has checks such as `if (HasCache)`
It is not possible beacause there is no way to set the cache to null. 

```csharp
var pca = ... // create public client app
pca.UserTokenCache = null; // does not compile, there is only a getter
````

** Proposal **
Either support this scenario (no token cache) or remove the internal logic around it, as it complicates code.

bgavrilMS:
Looking at this error code, it seems that this was a conscious decision - MSAL does not support cache disabling. 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e730464b889f509380948e3beaf4a0685832a61a/src/Microsoft.Identity.Client/MsalError.cs#L71

bgavrilMS:
Closing as by design. The cleanup done is not customer facing and should not be included in the release notes.

PR #1348, opened 2019-08-19
Cleanup null cache non-used code paths

bgavrilMS:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1345

This is leftover from ADAL. In MSAL, you cannot set the TokenCache to null. There is a bit of product code and quite a few tests that internally set it to null, but it not a valid scenario.

### DIFF
--- a/src/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -31,15 +31,13 @@ namespace Microsoft.Identity.Client.Cache
             AuthenticationRequestParameters requestParams, 
             ITelemetryManager telemetryManager)
         {
-            TokenCacheInternal = tokenCacheInternal;
+            TokenCacheInternal = tokenCacheInternal ?? throw new ArgumentNullException(nameof(requestParams));
             _requestParams = requestParams ?? throw new ArgumentNullException(nameof(requestParams));
             _telemetryManager = telemetryManager ?? throw new ArgumentNullException(nameof(telemetryManager));
         }
 
         #region ICacheSessionManager implementation
         public ITokenCacheInternal TokenCacheInternal { get; }
-
-        public bool HasCache => TokenCacheInternal != null; 
 
         public async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync()
         {
--- a/src/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Identity.Client.Cache
             AuthenticationRequestParameters requestParams, 
             ITelemetryManager telemetryManager)
         {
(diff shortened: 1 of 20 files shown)


src/Microsoft.Identity.Client/Cache/ICacheSessionManager.cs
```diff
@@ -14,7 +14,6 @@ namespace Microsoft.Identity.Client.Cache
     internal interface ICacheSessionManager
     {
         ITokenCacheInternal TokenCacheInternal { get; }
-        bool HasCache { get; }
```
bgavrilMS: Always true in MSAL

src/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -212,29 +210,11 @@ protected async Task<AuthenticationResult> CacheTokenResponseAndCreateAuthentica
...
-                AuthenticationRequestParameters.RequestContext.Logger.Info("Saving Token Response to cache..");
+            AuthenticationRequestParameters.RequestContext.Logger.Info("Saving Token Response to cache..");
+
+            var tuple = await CacheManager.SaveTokenResponseAsync(msalTokenResponse).ConfigureAwait(false);
+            return new AuthenticationResult(tuple.Item1, tuple.Item2, AuthenticationRequestParameters.RequestContext.CorrelationId);
 
-                var tuple = await CacheManager.SaveTokenResponseAsync(msalTokenResponse).ConfigureAwait(false);
-                return new AuthenticationResult(tuple.Item1, tuple.Item2, AuthenticationRequestParameters.RequestContext.CorrelationId);
-            }
-            else
```
bgavrilMS: This code was only ever used by tests, but there is no reason for this.

src/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -68,6 +68,7 @@ public static class MsalError
         /// This error code comes back from <see cref="ClientApplicationBase.AcquireTokenSilent(System.Collections.Generic.IEnumerable{string}, IAccount)"/> calls when
         /// the user cache had not been set in the application constructor. This should never happen in MSAL.NET 3.x as the cache is created by the application
```
bgavrilMS: This comment shows that this is an ADAL concept that was "pushed" into MSAL. We just never cleaned it up after the divorce.

src/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
```diff
@@ -31,16 +31,14 @@ internal class CacheSessionManager : ICacheSessionManager
             AuthenticationRequestParameters requestParams, 
             ITelemetryManager telemetryManager)
         {
-            TokenCacheInternal = tokenCacheInternal;
+            TokenCacheInternal = tokenCacheInternal ?? throw new ArgumentNullException(nameof(requestParams));
```
MarkZuber: shouldn't this be nameof(tokenCacheInternal)?

bgavrilMS:
@MarkZuber - fixed ArgNull exception message with correct param name

Issue #1349, opened 2019-08-19
[Bug] OBO flow always tries to get a token from the cache


bgavrilMS:
**Which Version of MSAL are you using ?**
Note that to get help, you need to run the latest version. Preview version are also ok. 
For ADAL, please log issues to https://github.com/AzureAD/azure-activedirectory-library-for-dotnet
4.3.1

**Platform**
ConfidentialClient

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [x] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Repro**

```csharp
// All our samples suggest to use the pattern ATS then OBO i.e. 
try 
{
 // AcquireTokenSilent ...
}
catch 
{
 // AcquireTokenOnBehalfOf
}

```

**Expected behavior**
AcquireTokenOnBehalfOf should never look in the token cache

**Actual behavior**
AcquireTokenOnBehalfOf looks in the cache for an AT, see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/46c7ab85572de6e5fcdd5fd794d0d47069b62720/src/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs#L41

The comment there is 
```
            // look for access token in the cache first.
            // no access token is found, then it means token does not exist
            // or new assertion has been passed. We should not use Refresh Token
            // for the user because the new incoming token may have updated claims
            // like mfa etc.
```

bgavrilMS:
@henrik-me @jmprieur @jennyf19 - as far as I remember calling OBO without AcquireTokenSilent works fine for our customers.

OBO will make a specialized search in the cache, as it searches for an access token associated with the current assertion. I am not sure we should be calling AcquireTokenSilent here, we even have a comment in our code explaining the reason why AcquireTokenSilent is discouraged: 

`Look for access token in the cache first.
 If no access token is found, then it means token does not exist
 or new assertion has been passed. 
 We should not use Refresh Token
 for the user because the new incoming token may have updated claims
 like mfa etc.`

@yordan-msft - could you advise on this matter please?

bgavrilMS:
CC @SomkaPe as you've looked at OBO in detail

yordan-msft:
[...]

PR #1360, opened 2019-08-28
circumvent cache look up for OBO

jennyf19:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1349

### DIFF
--- a/src/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
+++ b/src/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
@@ -7,7 +7,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 using Microsoft.Identity.Client.Core;
-using Microsoft.Identity.Client.Cache.Items;
 using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 
@@ -29,20 +28,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         internal override async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
         {
             await ResolveAuthorityEndpointsAsync().ConfigureAwait(false);
-
-            // look for access token in the cache first.
-            // no access token is found, then it means token does not exist
-            // or new assertion has been passed. We should not use Refresh Token
-            // for the user because the new incoming token may have updated claims
-            // like mfa etc.
-
-            MsalAccessTokenCacheItem msalAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
-            if (msalAccessTokenItem != null)
-            {
-                return new AuthenticationResult(msalAccessTokenItem, null, AuthenticationRequestParameters.RequestContext.CorrelationId);
-            }
-
-
+            
             var msalTokenResponse = await SendTokenRequestAsync(GetBodyParameters(), cancellationToken).ConfigureAwait(false);
             return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -29,20 +28,7 @@ internal class OnBehalfOfRequest : RequestBase
...
-
-            MsalAccessTokenCacheItem msalAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
-            if (msalAccessTokenItem != null)
-            {
-                return new AuthenticationResult(msalAccessTokenItem, null, AuthenticationRequestParameters.RequestContext.CorrelationId);
-            }
-
-
+            
             var msalTokenResponse = await SendTokenRequestAsync(GetBodyParameters(), cancellationToken).ConfigureAwait(false);
```
jmprieur: Do I understand correctly that:
1. this is now applying the recommended pattern?
2. But this is a breaking change for customers who were not applying this pattern?

bgavrilMS: Yes!

src/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -29,20 +28,7 @@ internal class OnBehalfOfRequest : RequestBase
         internal override async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
         {
             await ResolveAuthorityEndpointsAsync().ConfigureAwait(false);
-
-            // look for access token in the cache first.
-            // no access token is found, then it means token does not exist
-            // or new assertion has been passed. We should not use Refresh Token
-            // for the user because the new incoming token may have updated claims
-            // like mfa etc.
```
bgavrilMS: @jennyf19  - do you understand this comment? The caching logic here was added for a purpose, but seems to be forgotten.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/ClientCredentialWithCertTest.cs
```diff
@@ -173,7 +173,9 @@ public async Task JsonWebTokenWithX509PublicCertSendCertificateOnBehalfOfTestAsy
                 Assert.IsNotNull(result.AccessToken);
 
                 appCacheAccess.AssertAccessCounts(0, 0);
-                userCacheAccess.AssertAccessCounts(1, 1);
+                userCacheAccess.AssertAccessCounts(0, 1);
+
+                harness.HttpManager.AddMockHandler(CreateTokenResponseHttpHandler(false));
```
bgavrilMS: There are OBO tests that are commented. I think this is a good oportunity to bring them back to life.

src/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
jmprieur: Would it mean that our guidance is wrong?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/ClientCredentialWithCertTest.cs
bgavrilMS: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/bd8b7f693a01a058322198d4deb782e3eba831fe/tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/OboFlowTests.cs#L17

bgavrilMS: I think they were commented out when we moved to common core or smth. So they are pretty out of date possibly.

PR #1363, opened 2019-08-29
Move Microsoft.Identity.Client.* to src/client to make room for more …

MarkZuber:
…projects

We want to be able to pull in Microsoft.identity.web (with different naming) into src and potentially consolidate other extensions into this repo.  But the build.props and build.targets are specific to building the multi-targeting needed for Microsoft.Identity.Client and Microsoft.Identity.Client.Ref.  So this PR just moves those down from src\ into src\client so that we can put other projects we plan to ship underneath src\.

No code changes besides the project references have been made.


MarkZuber:
The one integration test failed due to timeout with chromedriver and not the change.

henrik-me:
did you validate a release build and the nuget package created?

MarkZuber:
> did you validate a release build and the nuget package created?

It will work, but the build definition has the path to the csproj hardcoded, so we'll need to update it to be src/client/ instead of src/ 

Let's talk in standup about how we want to manage the build projects because changing this as is will make it so this build definition won't build earlier versions (if we needed to do a patch).  Again really wishing the build definition was checked into the branch instead of being globally defined.

PR #1367, opened 2019-09-03
Enabling Device Code flow with personal accounts

jmprieur:
Removes the restriction common => organizations that was set up when Device Code flow did not work yet for Microsoft personal accounts.

When this PR is in, we also need to:
- Update https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Device-Code-Flow#constraints
- Update the following page in docs.ms: scenario-desktop-acquire-token.md (several locations in the page)
- remove references to AADSTS90133
- update docs.ms: scenario-desktop-app-registration.md
- update v2-supported-account-types.md
- update https://github.com/Azure-Samples/active-directory-dotnetcore-devicecodeflow-v2/blob/master/device-code-flow-console/appsettings.json#L16

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
@@ -49,15 +49,8 @@ namespace Microsoft.Identity.Client.Internal.Requests
             client.AddBodyParameter(OAuth2Parameter.Scope, deviceCodeScopes.AsSingleString());
             client.AddQueryParameter(OAuth2Parameter.Claims, AuthenticationRequestParameters.Claims);
 
-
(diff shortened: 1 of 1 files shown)


bgavrilMS:
LGTM @jmprieur, but needs a unit test. A bit surprised if one doesn't break. I can add one.

jmprieur:
@bgavrilMS : apparently 2 tests failed. 
Actually, I did this to check the state of the service (by changing , and I still got the service error ...
AADSTS90133: Device Code flow is not supported under /common or /consumers endpoint.

It seems it's not yet supported :(

bgavrilMS:
Yes, 2 integration tests that use Device Code Flow are failing. Good tests :)

bgavrilMS:
I have tested this and it seems to work. However I get the consent screen shown to me every time. I will commit the MSAL changes and will follow up on the consent screen.

PR #1370, opened 2019-09-04
Create YAML devops build pipelines

MarkZuber:
This PR creates YAML build pipelines that are a mirror of our exact current devops pipelines.  These are now setup as separate jobs with (YAML) in the title.  We'll run these in parallel for a while and then disable the old ones to have these be our defaults.

I've run a PR and a release build and they both work.  I verified that the release build creates the proper signed nuget package for microsoft.identity.client.  I haven't added the extensions assemblies or new tests to the builds yet, as this PR is just to get a mirror/snapshot of our existing builds and then we can modify them as needed in later PRs.

I will need to work with folks to get the proper TestSecret for the UI Automation but I believe I have the rest of the variables needed to get things setup properly.


build/pipeline-pullrequest.yaml
```diff
@@ -0,0 +1,95 @@
...
+- powershell: |
+   $kvSecretBytes = [System.Convert]::FromBase64String('$(AzureADIdentityDivisionTestAgent)')
+   $certCollection = New-Object System.Security.Cryptography.X509Certificates.X509Certificate2Collection
+   $certCollection.Import($kvSecretBytes, $null, [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::Exportable)
+
+   $protectedCertificateBytes = $certCollection.Export([System.Security.Cryptography.X509Certificates.X509ContentType]::Pkcs12)
+   $pfxPath = '$(Build.SourcesDirectory)' + "\TestCert.pfx"
+   [System.IO.File]::WriteAllBytes($pfxPath, $protectedCertificateBytes)
+
+   Import-PfxCertificate -FilePath $pfxPath -CertStoreLocation Cert:\LocalMachine\My
```
henrik-me: >Cert:\LocalMachine\My [](start = 63, length = 21)

Should the certs be in the user store instead? Also seems like the certs are not removed again when the build is complete, should they be?

trwalke: Ive seen instances where the cert cannot be found if in the user store for some unknown reason so I usually put it in local machine. So far, tests haven't had an issue finding them there. Also, this will be running on a hosted agent which will be wiped after we finish using it so I think its ok to leave the certs there.

build/pipeline-uiautomation.yaml
```diff
@@ -0,0 +1,249 @@
...
+    - '*'
+    exclude:
+    - master
+
+variables:
+  BuildPlatform: 'any cpu'
+  BuildConfiguration: 'debug'
+  FastRun: true
+
+#BUILD PHASE
```
trwalke: Have you verified that the build phases and the app center phases run in parallel?

MarkZuber: yes

build/pipeline-uiautomation.yaml
```diff
@@ -0,0 +1,249 @@
+
+# UI Automation builds should be triggered on every branch except master (so we don't trigger again after a merge)
```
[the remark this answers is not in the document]
MarkZuber: resolved as we discussed offline.

Issue #1365, opened 2019-09-02
[Bug] AADSTS50020 error if signing out then into another tenant without restarting the app


osolo:
**Which Version of MSAL are you using ?**
MSAL 4.3.1

**Platform**
all

**What authentication flow has the issue?**
* Desktop / Mobile
    * [X] Interactive

**Description**

In UWP apps (possibly also in non-UWP), if you execute the following sequence, an AADSTS50020 is issued:

1. `AcquireTokenInteractive()` [user logs in successfully to CompanyA.com]
2. `AcquireTokenSilent()` [this is required for the bug to appear]
3. Log out (see code below)
4. `AcquireTokenInteractive()` [user tries to login into CompanyB.com]

The user should be able to log into CompanyB, however we get this error:

> AADSTS50020: User account '[someemail]' from identity provider 'https://sts.windows.net/[some_guid]/' does not exist in tenant '[some name]' and cannot access the application '[some_guid]'(AppName) in that tenant. The account needs to be added as an external user in the tenant first. Sign out and sign in again with a different Azure Active Directory user account. 

Logout looks like this:
```csharp
var accounts = await _graphApp.GetAccountsAsync();
while (accounts.Any())
{
	await _graphApp.RemoveAsync(accounts.FirstOrDefault());
	accounts = await _graphApp.GetAccountsAsync();
}

```

**Additional Information**
1. If you restart the app between steps 3 and 4, things work as expected (user can log into CompanyB).
2. Adding `.WithAuthority(AadAuthorityAudience.AzureAdMultipleOrgs)` to the app builder does not resolve the problem.

I have provided a [sample app](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/files/3566873/MSALTest.zip) to demonstrate this problem.  You must supply your own AppId.

bgavrilMS:
I do not understand the repro steps, please clarify. You have a user e.g. john@companyA.com that is member of 2 tenants - companyA and companyB (home tenant companyA and guest tenant companyB). 

How does the user login to "companyA" and to "companyB"?

osolo:
Separate users in separate tenants.

For example, using the provided app, log in with your work (@microsoft.com) account.  Then log out and try to log in using any private/home Microsoft Account or an account on any other tenant (such as a demo tenant).

bgavrilMS:
Can repro.

bgavrilMS:
Can repro on other platforms, not just UWP.

[...]

PR #1372, opened 2019-09-05
Fix bug 1365 - make Authority objects immutable

bgavrilMS:
Consider reviewing commit by commit.
Fixes bug #1365 , caused by mutating the authority in the Config because it points to the same AuthorityInfo from the Request, which is being mutated. Solution - make all these objects immutable.

This paves the way for a bigger refactor - removing `AuthorityInfo` altogether. `Authority` object consists just of static builders and validations, plus a single member `AuthorityInfo`. Will take ~1 day to refactor, propose to do it as part of the story to remove the "validate authority" concept from MSAL. Thoughts?


src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -3,10 +3,14 @@
 
 using System;
 using System.Globalization;
+using Microsoft.Identity.Client.Instance;
 using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client
 {
+    /// <remarks>
+    /// This class must be kept immutable
```
bgavrilMS: There is a test for it now..

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -57,12 +61,32 @@ internal class AuthorityInfo
             }
         }
 
+        private AuthorityInfo(string host, string canonicalAuthority, AuthorityType authorityType, string userRealmUriPrefix, bool validateAuthority)
```
bgavrilMS: Copy ctor since we don't want several authorities to point to the same AuthorityInfo (even though AuthorityInfo is now immutable).

tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
```diff
@@ -46,6 +47,31 @@ public static void AreEqual(DateTimeOffset expected, DateTimeOffset actual, Time
...
         }
 
+        public static void IsImmutable<T>()
+        {
+            Assert.IsTrue(IsImmutable(typeof(T)));
+        }
+
+        private static bool IsImmutable(Type type)
+        {
+            if (type == typeof(string) || type.GetTypeInfo().IsPrimitive || type.GetTypeInfo().IsEnum)
```
bgavrilMS: Well, sort of immutable - if you have primitive types, you can always change them, i.e. ImmutableType.MyInt++;

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -18,29 +21,37 @@ internal abstract class Authority
...
+            }
+
+            // Don't reuse the same authority info, instead copy it
+            // to prevent objects updating each other's details
+            AuthorityInfo = new AuthorityInfo(authorityInfo);
         }
 
         public AuthorityInfo AuthorityInfo { get; }
 
-        protected IServiceBundle ServiceBundle { get; }
```
bgavrilMS: Having the servicebundle here is confusing and violates SRP. In fact, a single Authority ctor needs this, to be able to validate that the authority from the request is similar to the authority from the config. Much simpler to modify the affected ctor than to pass ServiceBundle everywhere.

tests/Microsoft.Identity.Test.Unit.net45/PlatformProxyPerformanceTests.cs
```diff
@@ -42,7 +42,7 @@ private void ValidateMethodPerformance(long maxMilliseconds, string name, Action
             }
         }
 
-        private const long AllowedMilliseconds = 10;
+        private const long AllowedMilliseconds = 20;
```
henrik-me: does this mean that we take a performance hit? what is causing this?

tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
```diff
@@ -46,6 +47,31 @@ public static void AreEqual(DateTimeOffset expected, DateTimeOffset actual, Time
             Assert.IsTrue(DictionariesAreEqual(dict1, dict2, valueComparer));
         }
 
+        public static void IsImmutable<T>()
+        {
+            Assert.IsTrue(IsImmutable(typeof(T)));
+        }
+
+        private static bool IsImmutable(Type type)
```
henrik-me: >IsImmutable [](start = 28, length = 11)

do we need to make these checks?

bgavrilMS: I like to keep my objects immutable explicitly, it's very easy to add a setter on a property by mistake. The root cause of this bug was mutating the authority

tests/Microsoft.Identity.Test.Unit.net45/PlatformProxyPerformanceTests.cs
bgavrilMS: I think 10ms is just too strict. After all this involves reflection.

Anyway, this is smth that randomly fails every now and then, so I though to bump the threshold

henrik-me: ok

tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
henrik-me: ok

henrik-me:
Thoughts on test failures?

trwalke:
> Thoughts on test failures?

@henrik-me The failure here is from an MFA error on the B2C local edit policy test. I have resolved it with the lab yesterday. Should be green from this point forward

PR #1376, opened 2019-09-06
Convert YAML builds into templates to share build steps

MarkZuber:
Main thing I still need to fix is the warning on release builds around SemVer 2.0.  If folks know how to disable this warning, that would be helpful.

C:\Program   Files\dotnet\sdk\2.2.105\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(202,5):   Error NU5105: The package version '4.3.1-internal20190906.4yamltemplates'   uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on   legacy clients. Change the package version to a SemVer 1.0.0 string. If the   version contains a release label it must start with a letter. This message   can be ignored if the package is not intended for older clients.
--

### DIFF
--- a/build/pipeline-releasebuild.yaml
+++ b/build/pipeline-releasebuild.yaml
@@ -21,299 +21,49 @@ variables:
   BuildConfiguration: 'release'
 
 steps:
-- task: securedevelopmentteam.vss-secure-development-tools.build-task-policheck.PoliCheck@1
-  displayName: 'Run PoliCheck'
-  inputs:
-    targetType: F
-    optionsFTPATH: 'build/policheck_filetypes.xml'
-  continueOnError: true
+# Run pre-build code analysis (policheck, credscan, etc)
+- template: template-prebuild-code-analysis.yaml
 
-- task: securedevelopmentteam.vss-secure-development-tools.build-task-credscan.CredScan@2
-  displayName: 'Run CredScan'
-  inputs:
-    suppressionsFile: 'build/credscan-exclusion.json'
-    debugMode: false
-
-- task: securedevelopmentteam.vss-secure-development-tools.build-task-postanalysis.PostAnalysis@1
-  displayName: 'Post Analysis'
-  inputs:
-    CredScan: true
-    PoliCheck: true
-
-- task: NuGetToolInstaller@0
-  displayName: 'Use NuGet 4.6.2'
-  inputs:
-    versionSpec: 4.6.2
-
-#Your build pipeline references an undefined variable named ‘TestSecret’. Create or edit the build pipeline for this YAML file, define the variable on the Variables tab. See https://go.microsoft.com/fwlink/?linkid=865972
+# Run Nuget Tool Installer
+- template: template-install-nuget.yaml
 
 - powershell: |
    $secret = '$(TestSecret)'
    $secret | Out-File $(Build.SourcesDirectory)/tests/Microsoft.Identity.Test.LabInfrastructure/data.txt
   displayName: 'Install Keyvault Secrets'
 
-- task: stevedower.python.PythonScript.PythonScript@1
-  displayName: 'Install MSAL.Python PIP'
-  inputs:
-    arguments: '-m pip install msal'
+# Configure Python
+- template: template-python-config.yaml
 
-- task: VSBuild@1
-  displayName: 'NuGet restore LibsAndSamples.sln'
-  inputs:
-    solution: LibsAndSamples.sln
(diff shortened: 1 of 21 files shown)


trwalke:
have you tried adding "<NoWarn>NU5105</NoWarn>" to the project to disable the warning?

build/pipeline-releasebuild.yaml
```diff
@@ -21,299 +21,38 @@ variables:
...
+- template: template-pack-and-sign-all-nugets.yaml
 
-- task: VSBuild@1
-  displayName: 'NuGet restore LibsAndSamples.sln'
-  inputs:
-    solution: LibsAndSamples.sln
-    vsVersion: '16.0'
-    msbuildArgs: '/t:restore'
-    configuration: '$(BuildConfiguration)'
+# Publish nuget packages and symbols
```
bgavrilMS: " ... to VSTS package manager" (I hope :D)

build/pipeline-releasebuild.yaml
```diff
@@ -21,299 +21,38 @@ variables:
   BuildConfiguration: 'release'
 
 steps:
```
bgavrilMS: doesn't the release build need the bootstrap template?

build/template-publish-analysis-and-cleanup.yaml
```diff
@@ -0,0 +1,30 @@
...
+    projectNameIDENTITYDIVISION: IDDP
+    areaPath: 'IDDP\DevEx-Client-SDK\DotNet'
+    iterationPath: 'IDDP\Unscheduled'
+    uploadAPIScan: false
+    uploadFortifySCA: false
+    uploadFxCop: false
+    uploadModernCop: false
+    uploadPREfast: false
+    uploadTSLint: false
+  continueOnError: true
```
bgavrilMS: we should break the build if TSA reports errors

build/template-sign-binary.yaml
```diff
@@ -0,0 +1,70 @@
...
+         "toolName": "signtool.exe",
+         "toolVersion": "6.2.9304.0"
+       }
+     ]
+    SessionTimeout: 20
+    VerboseLogin: true
+  timeoutInMinutes: 10
+
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@3
+  displayName: 'Run BinSkim ${{ parameters.Pattern }}'
```
bgavrilMS: Good idea to group binary validation as part of the signing template.

src/Directory.build.props
```diff
@@ -8,6 +8,9 @@
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)/../build/SolutionWideAnalyzerConfig.ruleset</CodeAnalysisRuleSet>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+
+    <NoWarn>NU5105</NoWarn>
```
bgavrilMS: Please add a comment on why this is needed

build/template-nuget-pack.yaml
```diff
@@ -0,0 +1,13 @@
...
+  BuildConfiguration: 'release'
+  ProjectPath: ''
+
+steps:
+- task: VSBuild@1
+  displayName: 'Pack ${{ parameters.ProjectPath }}'
+  inputs:
+    solution: ${{ parameters.ProjectPath }}
+    vsVersion: '15.0'
+    msbuildArgs: '/t:pack /p:nobuild=true /p:IncludeSymbols=true'
```
bgavrilMS: Note that I had found and logged [a bug with the msbuild.sdk.extras](https://github.com/onovotny/MSBuildSdkExtras/issues/136) where /t:pack would actually rebuild the project even if /p:nobuild=true is passed. The workaround was to pack twice (see details in the bug). 

So just a heads up to check that the assemblies in the nuget package are actually signed. If they are not - it's probably because of this.

build/pipeline-releasebuild.yaml
MarkZuber: Yes, it's there on line 28.  The release build does the code analysis first to look at the source.

build/template-nuget-pack.yaml
MarkZuber: I've validated that the binaries are signed.  What's running here is exactly what we have running in our existing builds, just broken up into yaml scripts.

PR #1378, opened 2019-09-09
[iOS] Support UINavigationController with empty stack

davidjohnoliver:
This change permits calling AcquireAuthorizationAsync() with an app RootViewController that is a UINavigationController with an empty navigation stack.

The top-level view controller will now be returned, rather than trying to recursively call FindCurrentViewController() on a null VisibleViewController (which causes a NullReferenceException).

## How to reproduce the bug
Build the solution at https://github.com/davidjohnoliver/active-directory-Uno and run the iOS head.

### DIFF
--- a/src/client/Microsoft.Identity.Client/UI/CoreUIParent.cs
+++ b/src/client/Microsoft.Identity.Client/UI/CoreUIParent.cs
@@ -65,10 +65,9 @@ namespace Microsoft.Identity.Client.UI
                 UITabBarController tabBarController = (UITabBarController)callerViewController;
                 return FindCurrentViewController(tabBarController.SelectedViewController);
             }
-            else if (callerViewController is UINavigationController)
+            else if (callerViewController is UINavigationController uiNavigationController && uiNavigationController.VisibleViewController != null)
             {
-                UINavigationController navigationController = (UINavigationController)callerViewController;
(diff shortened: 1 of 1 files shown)


bgavrilMS:
LGTM

bgavrilMS:
Contribution is much appreciated @davidjohnoliver ; Please let us know if building / running tests in MSAL was difficult.

henrik-me:
@MarkZuber @trwalke  : seems like the build is failing. 
@bgavrilMS : can you pls. help validate the change manually?

henrik-me:
change looks fine to me

davidjohnoliver:
Thanks @bgavrilMS , I intended to reply to your question re: tests but forgot. 

I did see that there were a few UI tests for iOS, but they seem to only cover the main authentication flow. I didn't know if creating a new UI test was warranted in this case. Let me know if it's something you think should be added.

bgavrilMS:
@davidjohnoliver - the UI tests are pretty complicated and we don't expect contributors to write any of those. We don't (currently) run unit tests on mobile platforms, but most of our code is common and is unit tested. Your fix is in the part that can't be unit tested. 

I plan to manually validate and merge your change soon. Should go in the next release, which is tentatively end of week.

PR #1380, opened 2019-09-10
Component governance updates and build integration

bgavrilMS:
1. integrate CG in all builds
2. Update some dependencies that have security issues

### DIFF
--- a/src/client/Directory.Build.props
+++ b/src/client/Directory.Build.props
@@ -117,6 +117,15 @@
     <EmbeddedResource Include="$(PathToMsalSources)\Properties\Microsoft.Identity.Client.rd.xml" />
   </ItemGroup>
 
+  <ItemGroup Label="Dependency overrides due to security vulnerabilities">
+    <!-- 4.3.0 has CVE-2019-0657 -->
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
+    <!-- 4.1.3 has CVE-2018-0786 -->
+    <PackageReference Include="System.ServiceModel.Http" Version="4.5.3" Condition="'$(TargetFramework)' == '$(TargetFrameworkUap)'" />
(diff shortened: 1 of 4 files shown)


tests/CacheCompat/CommonCache.Test.MsalJava/pom.xml
```diff
@@ -24,7 +24,7 @@
         <dependency>
             <groupId>com.microsoft.azure</groupId>
             <artifactId>msal4j</artifactId>
-            <version>0.4.0-preview</version>
+            <version>0.5.0-preview</version>
```
bgavrilMS: CG also checks the status of our friends using Maven :)

src/client/Directory.Build.props
```diff
@@ -117,6 +117,15 @@
     <EmbeddedResource Include="$(PathToMsalSources)\Properties\Microsoft.Identity.Client.rd.xml" />
   </ItemGroup>
 
+  <ItemGroup Label="Dependency overrides due to security vulnerabilities">
+    <!-- 4.3.0 has CVE-2019-0657 -->
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
```
bgavrilMS: There are 2 ways of addressing this CVE: 

- update the .net core runtime (not applicable to us, since we are a library that needs to run on any 2.1.x runtime)
- upgrade the private package explicitly

src/client/Directory.Build.props
```diff
@@ -117,6 +117,15 @@
...
+    <!-- 4.1.3 has CVE-2018-0786 -->
+    <PackageReference Include="System.ServiceModel.Http" Version="4.5.3" Condition="'$(TargetFramework)' == '$(TargetFrameworkUap)'" />
```
bgavrilMS: I bumped the version of these packages only in UAP (and have tested things still work). Theoretically we can upgrade all the frameworks, but only UWP brings this 4.1.3 version of the libraries. 

Frameworks have metapackages associated with them, so we do not normally see these packages referenced directly.

Issue #1357, opened 2019-08-26
Part 1: Modify iOS broker to handle iOS 13 changes


jennyf19:
iOS 13 has made changes around the authenticator's ability to verify applications. Currently, MSAL.NET relies on the sourceApplication to verify requests, this will need to be modified.

bgavrilMS:
What is the user impact?

bgavrilMS:
Just trying to understand the impact here. So you mean that once people upgrade to iOS 13, pre-existing apps that use the broker will stop working? Sounds like a high priority item then.

henrik-me:
@jennyf19 : I created a new issue to track part 2 as this issue is linked to the 4.4.0 release.
Part2: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1396

henrik-me:
Cloned the issue to fix the rest in the next release

biozal:
@jennyf19  was this fully tested?  After upgrading to this we no longer work with broker errors:  
{Microsoft.IdentityModel.Clients.ActiveDirectory.AdalServiceException: Unencrypted broker response hash did not match the expected hash
  at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.OAuth2.TokenResponse.GetResult (System.DateTimeOffset expiresOn, System.DateTimeOffset extendedExpiresOn) [0x0015f] in <0d6c65707ae94d29ab4fb4ec5df8169f>:0 
  at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform.iOSBroker.ResultFromBrokerResponse (System.Collections.Generic.IDictionary`2[TKey,TValue] responseDictionary) [0x00132] in <0d6c65707ae94d29ab4fb4ec5df8169f>:0 
  at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform.iOSBroker.ProcessBrokerResponse () [0x0008a] in <0d6c65707ae94d29ab4fb4ec5df8169f>:0 
  at Microsoft.IdentityModel.Clients.ActiveDirectory.Internal.Platform.iOSBroker.AcquireTokenUsingBrokerAsync (System.Collections.Generic.IDictionary`2[TKey,TValue] brokerPayload) [0x0026b] in <0d6c65707ae94d29ab4fb4ec5df8169f>:0 
[...]

PR #1383, opened 2019-09-10
Part 1: iOS 13 broker support

jennyf19:
**Part 1: iOS 13 broker support**
- Generate a nonce to send in the broker request, only if msauthv3 (iOS broker 6.3.19+) is installed on the device
- Check that the nonce in the broker response matches the nonce sent in the request
- Update info.plist in dev app for msauthv3

More info on these required changes [here](https://github.com/AzureAD/microsoft-authentication-library-for-objc/wiki/iOS-13-and-macOS-10.15-support-in-MSAL)

**Tested w/:**
- msauthv3 (iOS broker 6.3.19+)
- msauthv2 (iOS broker 6.3.12)

**TODO**: 
- Test w/iOS 13 and XCode 11 where source application returns null
- Create aka.ms link
- Update wiki and/or link to iOS team's docs

**Part 2:** Will include handling an application token, storing it in keychain and sending in the broker responses, but will do ADAL.NET work first and then come back to Part 2. Part 2 is not as urgent.

[issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1357)

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerParameter.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerParameter.cs
@@ -35,8 +35,8 @@ namespace Microsoft.Identity.Client.Internal.Broker
         public const string Force = "force";
 
         public const string BrokerInstallUrl = "broker_install_url";
-        public const string BrokerV2 = "msauthv2://";
-        public const string BrokerV3 = "msauthv3://";
+        public const string UriSchemeBrokerV2 = "msauthv2://";
+        public const string UriSchemeBrokerV3 = "msauthv3://";
         public const string AuthCodePrefixForEmbeddedWebviewBrokerInstallRequired = "msauth://";
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerParameter.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerParameter.cs
@@ -36,6 +36,7 @@ namespace Microsoft.Identity.Client.Internal.Broker
 
         public const string BrokerInstallUrl = "broker_install_url";
         public const string BrokerV2 = "msauthv2://";
+        public const string BrokerV3 = "msauthv3://";
         public const string AuthCodePrefixForEmbeddedWebviewBrokerInstallRequired = "msauth://";
     }
 }
--- a/src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
+++ b/src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
@@ -101,12 +101,13 @@ namespace Microsoft.Identity.Client.OAuth2
 
         internal static MsalTokenResponse CreateFromBrokerResponse(Dictionary<string, string> responseDictionary)
         {
-            if (responseDictionary.ContainsKey(BrokerResponseConst.ErrorMetadata))
+            if (responseDictionary.ContainsKey(BrokerResponseConst.BrokerErrorCode) ||
+                responseDictionary.ContainsKey(BrokerResponseConst.BrokerErrorDescription))
             {
                 return new MsalTokenResponse
                 {
                     Error = responseDictionary[BrokerResponseConst.BrokerErrorCode],
-                    ErrorDescription = CoreHelpers.UrlDecode(responseDictionary[BrokerResponseConst.ErrorMetadata])
+                    ErrorDescription = CoreHelpers.UrlDecode(responseDictionary[BrokerResponseConst.BrokerErrorDescription])
                 };
             }
 
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
@@ -24,14 +24,33 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
-        /// Returns if the response is from the broker app
+        /// Returns if the response is from the broker app. See https://aka.ms/msal-net-ios-13-broker
+        /// for more details.
         /// </summary>
         /// <param name="sourceApplication">application bundle id of the broker</param>
         /// <returns>True if the response is from broker, False otherwise.</returns>
         public static bool IsBrokerResponse(string sourceApplication)
         {
-            Debug.WriteLine("IsBrokerResponse Called with sourceApplication {0}", sourceApplication);
-            return sourceApplication != null && sourceApplication.Equals("com.microsoft.azureauthenticator", StringComparison.OrdinalIgnoreCase);
+            Debug.WriteLine("IsBrokerResponse called with sourceApplication {0}", sourceApplication);
+
+            if (sourceApplication != null && sourceApplication.Equals("com.microsoft.azureauthenticator", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+            else if (string.IsNullOrEmpty(sourceApplication))
+            {
+                // For iOS 13+, SourceApplication will not be returned
+                // Customers will need to install iOS broker >= 6.3.19
+                // MSAL.NET will generate a nonce (guid), which broker will
+                // return in the response. MSAL.NET will validate a match in iOSBroker.cs
+                // So if SourceApplication is null, just return, MSAL.NET will throw a 
+                // specific error message if the nonce does not match.
+                return true; 
+            }
+            else
+            {
+                return false;
+            }
         }
 
         /// <summary>
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Identity.Client
             {
                 return true;
             }
-            else if (string.IsNullOrEmpty(sourceApplication))
+            if (string.IsNullOrEmpty(sourceApplication))
             {
                 // For iOS 13+, SourceApplication will not be returned
                 // Customers will need to install iOS broker >= 6.3.19
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
@@ -45,12 +45,10 @@ namespace Microsoft.Identity.Client
                 // return in the response. MSAL.NET will validate a match in iOSBroker.cs
                 // So if SourceApplication is null, just return, MSAL.NET will throw a 
                 // specific error message if the nonce does not match.
-                return true; 
+                return true;
             }
-            else
-            {
-                return false;
-            }
+
+            return false;
         }
 
         /// <summary>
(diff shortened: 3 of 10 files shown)


src/client/Microsoft.Identity.Client/Internal/Broker/BrokerParameter.cs
```diff
@@ -36,6 +36,7 @@ internal class BrokerParameter
 
         public const string BrokerInstallUrl = "broker_install_url";
         public const string BrokerV2 = "msauthv2://";
+        public const string BrokerV3 = "msauthv3://";
```
bgavrilMS: These constants are uri schemes, so they should be named appropriately.

src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
```diff
@@ -24,14 +24,33 @@ public static bool SetAuthenticationContinuationEventArgs(NSUrl url)
...
         {
-            Debug.WriteLine("IsBrokerResponse Called with sourceApplication {0}", sourceApplication);
-            return sourceApplication != null && sourceApplication.Equals("com.microsoft.azureauthenticator", StringComparison.OrdinalIgnoreCase);
+            Debug.WriteLine("IsBrokerResponse called with sourceApplication {0}", sourceApplication);
+
+            if (sourceApplication != null && sourceApplication.Equals("com.microsoft.azureauthenticator", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+            else if (string.IsNullOrEmpty(sourceApplication))
```
bgavrilMS: if you don't need `else` becasue you return. Code is clearer like: 

`if (...) { return true} if (...) { return false } if (..) return x`

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -27,6 +27,9 @@ internal class iOSBroker : NSObject, IBroker
 
         private readonly ICoreLogger _logger;
         private readonly ICryptographyManager _cryptoManager;
+        private string BrokerRequestNonce { get; set; }
```
bgavrilMS: You don't need properties for these.

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -46,9 +49,19 @@ public bool CanInvokeBroker(CoreUIParent uiParent)
 
             uiParent.CallerViewController.InvokeOnMainThread(() =>
             {
-                result = UIApplication.SharedApplication.CanOpenUrl(new NSUrl(BrokerParameter.BrokerV2));
+                result = UIApplication.SharedApplication.CanOpenUrl(new NSUrl(BrokerParameter.BrokerV3));
```
bgavrilMS: `result` is an overloaded term, please use somehting with more local meaning, like `canOpenUrl`

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -46,9 +49,19 @@ public bool CanInvokeBroker(CoreUIParent uiParent)
...
+                if (result)
+                {
+                    BrokerRequestNonce = Guid.NewGuid().ToString();
+                    _logger.Info(iOSBrokerConstants.iOSBrokerv3Installed + result);
+                }
             });
 
+            if (!result)
+            {
+                result = UIApplication.SharedApplication.CanOpenUrl(new NSUrl(BrokerParameter.BrokerV2));
```
bgavrilMS: Why can this be on a non-UI thread but the call at line 52 must be on the UI thread?

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -46,9 +49,19 @@ public bool CanInvokeBroker(CoreUIParent uiParent)
 
             uiParent.CallerViewController.InvokeOnMainThread(() =>
             {
-                result = UIApplication.SharedApplication.CanOpenUrl(new NSUrl(BrokerParameter.BrokerV2));
+                result = UIApplication.SharedApplication.CanOpenUrl(new NSUrl(BrokerParameter.BrokerV3));
+                if (result)
+                {
+                    BrokerRequestNonce = Guid.NewGuid().ToString();
```
bgavrilMS: Needs refactoring. The method `CanInvokeBroker` suggest it is a `query` type of method (i.e. read some stuff, give me a result). However, you actually generate a nonce, so it creates state. 

I suggest you decouple the nonce creation from the CanInvoke method. If you are worried about performacen, then feel free to cache the values of `canInvokeV2Broker` and `canInvokeV3Broker` - but I would not be worried at this point.

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -72,6 +85,11 @@ private void AddIosSpecificParametersToPayload(Dictionary<string, string> broker
             brokerPayload[iOSBrokerConstants.BrokerKey] = encodedBrokerKey;
             brokerPayload[iOSBrokerConstants.MsgProtocolVer] = BrokerParameter.MsgProtocolVersion3;
 
+            if (!string.IsNullOrEmpty(BrokerRequestNonce))
```
bgavrilMS: No, it would be better to check if the V3 broker can be used and to generate the nonce here.

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -27,6 +27,9 @@ internal class iOSBroker : NSObject, IBroker
 
         private readonly ICoreLogger _logger;
         private readonly ICryptographyManager _cryptoManager;
+        private string BrokerRequestNonce { get; set; }
+        private string BrokerResponseNonce { get; set; }
+        public static bool DoesNonceMatch { get; set; } = false;
```
bgavrilMS: There doesn't seem to be any need to have this property as public? Or even have it at all.

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -180,9 +192,33 @@ private MsalTokenResponse ResultFromBrokerResponse(Dictionary<string, string> re
                 };
             }
 
+            if (!string.IsNullOrEmpty(BrokerRequestNonce))
+            {
+                ValidateBrokerResponseNonceWithRequestNonce(responseDictionary);
+                if (!DoesNonceMatch)
```
bgavrilMS: Validate method can return a boolean, there is no need for `DoesNonceMatch`

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -180,9 +192,33 @@ private MsalTokenResponse ResultFromBrokerResponse(Dictionary<string, string> re
                 };
             }
 
+            if (!string.IsNullOrEmpty(BrokerRequestNonce))
```
bgavrilMS: verification should be at the top of this method (if this fails, there's no point in doing all the other calculations). It should also encapsulate all the validatin logic, i.e. including line 195.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -101,12 +101,13 @@ public long RefreshIn
 
         internal static MsalTokenResponse CreateFromBrokerResponse(Dictionary<string, string> responseDictionary)
         {
-            if (responseDictionary.ContainsKey(BrokerResponseConst.ErrorMetadata))
+            if (responseDictionary.ContainsKey(BrokerResponseConst.BrokerErrorCode) ||
```
bgavrilMS: This is a not negligible amount of logic that should be unit tested (it's not ios code)

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerParameter.cs
jennyf19: good point. thanks,

---
In reply to: [323181834](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1383#discussion_r323181834) [](ancestors = 323181834)

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
jennyf19: i was trying to access this elsewhere, but didn't remove the properties later. thx.

---
In reply to: [323183904](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1383#discussion_r323183904) [](ancestors = 323183904)

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
jennyf19: thanks for the suggestions...maybe it's better? also...cannot test it because my mac gave out. have to reinstall and update a bunch of stuff :(

---
In reply to: [323184596](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1383#discussion_r323184596) [](ancestors = 323184596)

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
jennyf19: once i can test again, i can pull some actual decoded broker responses and write some tests. 

---
In reply to: [323189847](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1383#discussion_r323189847) [](ancestors = 323189847)

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -42,19 +44,36 @@ public bool CanInvokeBroker(CoreUIParent uiParent)
                 throw new MsalClientException(MsalError.UIViewControllerRequiredForiOSBroker, MsalErrorMessage.UIViewControllerIsRequiredToInvokeiOSBroker);
             }
 
-            var result = false;
+            bool canStartBroker = false;
```
bgavrilMS: canStartV3Broker

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -154,18 +179,25 @@ private MsalTokenResponse ResultFromBrokerResponse(Dictionary<string, string> re
...
             string expectedHash = responseDictionary[iOSBrokerConstants.ExpectedHash];
             string encryptedResponse = responseDictionary[iOSBrokerConstants.EncryptedResponsed];
             string decryptedResponse = BrokerKeyHelper.DecryptBrokerResponse(encryptedResponse, _logger);
             string responseActualHash = _cryptoManager.CreateSha256Hash(decryptedResponse);
             byte[] rawHash = Convert.FromBase64String(responseActualHash);
             string hash = BitConverter.ToString(rawHash);
 
+            if (!string.IsNullOrEmpty(_brokerRequestNonce))
+            {
+                if (!ValidateBrokerResponseNonceWithRequestNonce(responseDictionary))
```
bgavrilMS: Please move `if (!string.IsNullOrEmpty(_brokerRequestNonce))` inside `ValidateBrokerResponseNonceWithRequestNonce`

Issue #1384, opened 2019-09-11
[Bug] Client Credential should not be allowed over tenantless authorities


bgavrilMS:
**Which Version of MSAL are you using ?**
4.3.x

**What authentication flow has the issue?**
Client creds

**Repro**

1. Create CCA with tenantless AAD authority (e.g. common) 
2. AcquireTokenForClient  -> fetches a token, for tenant T1
3. AcquireTokenForClient  with authority override set tenanted authority T1

**Actual:** Second call to AcquireTokenForClient should fetch the AT in the cache
**Expected:** MSAL talks to Evo to get a new token

Note: discussed this with Adrian Frei and Evo's stance is that client_credentials should never have he authority set to `common`. From MSAL perspective, we cannot get the actual tenant ID because we normally get it from the ID Token, which is not present in response.

jmprieur:
@hpsin : can you please confirm that we do want to disallow `organizations` in Client credentials in next major version of MSAL.NET

bgavrilMS:
(and `common` as well)

jmprieur:
common for sure (it does not make sense for client creds)

hpsin:
Yes - `common` and `organizations` are both aliases over multiple tenants, and do not specify which tenant a token should be requested for.  For security and reliability, apps should explicitly indicate what tenant they want to get an app-only token for instead of having AAD do best-effort guesses.

hpsin:
`consumers` is ok, as app tokens are valid there and `consumers` is a single tenant. Please do not block use of `consumers`.

jmprieur:
Thanks @hpsin

bgavrilMS:
I'd like to take this as a non-breaking change. It looks like ESTS-R blocks it anyway.

maliksahil:
We should go forward with this, but can we start with a warning in a major version, before an error or disallow in the next?

bgavrilMS:
We already log an error message for this case, and this has been the case for a while.

PR #1387, opened 2019-09-13
Fix for #1384 - do not allow /common authorities on client creds

bgavrilMS:
Fix for #1384 
Can be reviewed but do not commit this before it goes through API governance.

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
+++ b/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
@@ -278,6 +278,10 @@ namespace Microsoft.Identity.Client
             " Potential issue: the redirect URI is valid, but it has been configured for the wrong app type." +
             " Check the configuration of the app being used in the app registration portal." +
             " See https://aka.ms/msal-net-invalid-client for details. ";
+        
+        internal const string ClientCredentialExplicitTenantId =
+            "You're using the client credential flow (for a daemon app) and you didn't configure an authority or you are trying to use an AAD authority with a general audience (`common`, `organizations` or `consumers`). " +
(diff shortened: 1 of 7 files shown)


src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -278,6 +278,10 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
             " Potential issue: the redirect URI is valid, but it has been configured for the wrong app type." +
             " Check the configuration of the app being used in the app registration portal." +
             " See https://aka.ms/msal-net-invalid-client for details. ";
+        
+        internal const string ClientCredentialExplicitTenantId =
+            "You didn't configure an authority or you are trying to use an AAD authority with a general audience (`common`, `organizations` or `consumers`). " +
```
jmprieur: You're using the client credential flow (for a daemon app), and you didn't ..

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -278,6 +278,10 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
...
+            "Specify the tenant id with your authority. You can get tenant id from the app registration portal. Use a method like .WithAuthority(audience, tenantID) to specify the authority.";
```
jmprieur: or maybe instead of (or in addition to) "You can get tenant id from the app registration portal. Use a method like .WithAuthority(audience, tenantID) to specify the authority.", add ". See https://aka.ms-msal-net-client-credentials" ?

henrik-me:
For the spirit of this change it should be aligned across libraries which support client credential flows.

PR #1389, opened 2019-09-18
[WIP] server side http telemetry

jennyf19:
@bgavrilMS

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.Identity.Client.TelemetryCore;
 using Microsoft.Identity.Client.TelemetryCore.Internal;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 
@@ -25,7 +26,18 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
 
         public void AddApiTelemetryFeature(ApiTelemetryFeature feature)
         {
-            _apiTelemetry[MatsConverter.AsString(feature)] = "true";
+            _apiTelemetry[MatsConverter.AsString(feature)] = TelemetryConstants.True;
+        }
+
+        public void AddApiTelemetryFeature(ApiTelemetryFeature feature, bool isTrue)
+        {
+            string telemetryEnabled = TelemetryConstants.False;
+            if (isTrue)
+            {
+                telemetryEnabled = TelemetryConstants.True;
+            }
+
+            _apiTelemetry[MatsConverter.AsString(feature)] = telemetryEnabled;
         }
 
         public IEnumerable<KeyValuePair<string, string>> GetApiTelemetryFeatures()
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
@@ -9,7 +9,6 @@ using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.OAuth2;
-using Microsoft.Identity.Client.TelemetryCore;
 using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Internal.Requests
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.TelemetryCore;
 using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Internal.Requests
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
@@ -236,9 +236,6 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 authorizationRequestParameters[OAuth2Parameter.Prompt] = _interactiveParameters.Prompt.PromptValue;
             }
 
-            authorizationRequestParameters[TelemetryConstants.XClientLastRequest] = ServiceBundle.TelemetryManager.FetchAndResetPreviousHttpTelemetryContent();
-            authorizationRequestParameters[TelemetryConstants.XClientCurrentTelemetry] = ServiceBundle.TelemetryManager.FetchAndResetCurrentHttpTelemetryContent();
-
             return authorizationRequestParameters;
         }
 
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
@@ -16,6 +16,7 @@ using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.UI;
 using Microsoft.Identity.Client.Utils;
 using Microsoft.Identity.Client.TelemetryCore.Internal;
+using Microsoft.Identity.Client.TelemetryCore;
 
 namespace Microsoft.Identity.Client.Internal.Requests
 {
@@ -234,6 +235,9 @@ namespace Microsoft.Identity.Client.Internal.Requests
             {
                 authorizationRequestParameters[OAuth2Parameter.Prompt] = _interactiveParameters.Prompt.PromptValue;
             }
+
+            authorizationRequestParameters[TelemetryConstants.XClientLastRequest] = ServiceBundle.TelemetryManager.FetchAndResetPreviousHttpTelemetryContent();
+            authorizationRequestParameters[TelemetryConstants.XClientCurrentTelemetry] = ServiceBundle.TelemetryManager.FetchAndResetCurrentHttpTelemetryContent();
 
             return authorizationRequestParameters;
         }
--- a/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
+++ b/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Runtime.Serialization;
+
+namespace Microsoft.Identity.Client.TelemetryCore.Internal
+{
+    [DataContract]
+    internal class HttpTelemetryContent
+    {
+        [DataMember]
+        public string LastErrorCode { get; set; }
+
+        [DataMember]
+        public int UnreportedErrorCount { get; set; }
+
+        [DataMember]
+        public string ApiId { get; set; }
+
+        [DataMember]
+        public string CorrelationId { get; set; }
+
+        public void ResetLastErrorCode()
+        {
+            LastErrorCode = string.Empty;
+        }
+    }
+}
--- a/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
+++ b/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
@@ -1,28 +1,57 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System.Runtime.Serialization;
+using Microsoft.Identity.Client.TelemetryCore.Internal.Constants;
+using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 
 namespace Microsoft.Identity.Client.TelemetryCore.Internal
 {
-    [DataContract]
     internal class HttpTelemetryContent
     {
-        [DataMember]
-        public string LastErrorCode { get; set; }
-
-        [DataMember]
-        public int UnreportedErrorCount { get; set; }
-
-        [DataMember]
-        public string ApiId { get; set; }
-
-        [DataMember]
-        public string CorrelationId { get; set; }
-
-        public void ResetLastErrorCode()
+        public HttpTelemetryContent(EventBase evt)
         {
-            LastErrorCode = string.Empty;
+            if (evt != null)
+            {
+                evt.TryGetValue(MsalTelemetryBlobEventNames.ApiIdConstStrKey, out string apiId);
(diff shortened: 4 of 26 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.TelemetryCore;
```
bgavrilMS: not needed?

src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
```diff
@@ -235,6 +236,9 @@ private Uri CreateAuthorizationUri(bool addPkceAndState = false)
                 authorizationRequestParameters[OAuth2Parameter.Prompt] = _interactiveParameters.Prompt.PromptValue;
             }
 
+            authorizationRequestParameters[TelemetryConstants.XClientLastRequest] = ServiceBundle.TelemetryManager.FetchAndResetPreviousHttpTelemetryContent();
```
bgavrilMS: I though we are sending this as headers? Here you are sending this as form params. Also, is it confirmed that the server accepts telemetry on the /authorize endpoint?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -16,4 +16,11 @@ internal static class TelemetryError
     //    public const string MsalCacheEventCount = "msal.cache_event_count";
     //    public const string MsalUiEventCount = "msal.ui_event_count";
     //}
+
+    internal static class TelemetryConstants
+    {
+        public const string HttpTelemetrySchemaVersion1 = "1|";
```
bgavrilMS: Please set the pipe as a different constant.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -64,6 +72,8 @@ public void StartEvent(EventBase eventToStart)
 
         public void StopEvent(EventBase eventToStop)
         {
+            ProcessEventsForPreviousHttpTelemetryContent(eventToStop); //try movnig up to start
```
bgavrilMS: Please consolidate - set it on start or on stop.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -137,6 +147,93 @@ public void Flush(string correlationId)
             Callback?.Invoke(eventsToFlush.Cast<Dictionary<string, string>>().ToList());
         }
 
+        public string FetchAndResetPreviousHttpTelemetryContent()
+        {
+            if (!string.IsNullOrEmpty(_previousTelemetryPayload.ApiId))
+            {
+                string csv = CreateCSVForPreviousHttpTelem();
+                _currentTelemetryPayload.ResetLastErrorCode();
```
bgavrilMS: Should be `_previousTelemetryPayload `

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -137,6 +147,93 @@ public void Flush(string correlationId)
...
+
+            string csvString = string.Join(",", myValues);
+            csvString = TelemetryConstants.HttpTelemetrySchemaVersion1 + csvString;
+            return csvString;
+        }
+
+        private string CreateCSVForCurrentHttpTelem()
+        {
+            // csv expected format:
+            // 1|api_id|platform_config
```
bgavrilMS: The actual fomat seems to be 1|api_id|correlation_id

src/client/Microsoft.Identity.Client/WsTrust/WsTrustWebRequestManager.cs
```diff
@@ -26,8 +28,10 @@ public WsTrustWebRequestManager(IHttpManager httpManager)
         /// <inheritdoc/>
         public async Task<MexDocument> GetMexDocumentAsync(string federationMetadataUrl, RequestContext requestContext)
         {
+            IDictionary<string, string> headers = CreateHttpTelemetryHeaders(requestContext);
+
             var uri = new UriBuilder(federationMetadataUrl);
-            HttpResponse httpResponse = await _httpManager.SendGetAsync(uri.Uri, null, requestContext.Logger).ConfigureAwait(false);
+            HttpResponse httpResponse = await _httpManager.SendGetAsync(uri.Uri, headers, requestContext.Logger).ConfigureAwait(false);
```
bgavrilMS: I don't think the Ws Trust endpoint supports this, nor do I think they will ever support this, so might as well not send the data.

bgavrilMS: That said, we should definitely send the MsalIDParameters here.

src/client/Microsoft.Identity.Client/WsTrust/WsTrustWebRequestManager.cs
```diff
@@ -109,16 +113,37 @@ public async Task<MexDocument> GetMexDocumentAsync(string federationMetadataUrl,
...
 
             return httpResponse.StatusCode == System.Net.HttpStatusCode.OK
                 ? JsonHelper.DeserializeFromJson<UserRealmDiscoveryResponse>(httpResponse.Body)
                 : null;
         }
+
+        private IDictionary<string, string> CreateHttpTelemetryHeaders(RequestContext requestContext)
+        {
+            IDictionary<string, string> httpTelemetryHeader = new Dictionary<string, string>();
+            httpTelemetryHeader[TelemetryConstants.XClientLastRequest] = requestContext.ServiceBundle.TelemetryManager.FetchAndResetPreviousHttpTelemetryContent();
```
bgavrilMS: An authentication flow involves talking to the several endpoints - discovery, oidc, ws-trust, authroize, token. However, I do not think the server side is ready to accept data on all of these - my understanding is that it only accepts the data on /token (and maybe /authorize?). 

And since you're resetting the last error code everytime you send the telemetry data, we'll most likely lose it. So I propose either: 

1. Only send telemetry on /token endpoint OR
2. Only reset the last error code when sending to /token endpoint

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpMessageHandler.cs
```diff
@@ -96,6 +98,13 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
                 }
             }
 
+            if (ExpectedHeaders != null)
+            {
+                Assert.IsTrue(ActualRequestMessage.Headers.Contains(TelemetryConstants.XClientLastRequest));
```
bgavrilMS: This assertion is not strong enough. You can expose `ActualHeaders` and then validate that you're sending the correct telemetry in a variatey of use cases.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpMessageHandler.cs
```diff
@@ -96,6 +98,13 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
...
+                Assert.IsTrue(ActualRequestMessage.Headers.Contains(TelemetryConstants.XClientCurrentTelemetry));
+                ExpectedHeaders = null;
```
bgavrilMS: Why is this needed?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -137,6 +147,93 @@ public void Flush(string correlationId)
...
+        private void ProcessEventsForPreviousHttpTelemetryContent(EventBase events)
+        {
+            foreach (var ev in events)
+            {
+                if (string.Equals(ev.Key, MsalTelemetryBlobEventNames.ApiIdConstStrKey))
+                {
+                    _previousTelemetryPayload.ApiId = ev.Value;
+                }
+
+                if (string.Equals(ev.Key, MsalTelemetryBlobEventNames.ApiErrorCodeConstStrKey))
```
bgavrilMS: I don't understand where this data is written since you've just defined this key, but you are not using it anywhere.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpMessageHandler.cs
jennyf19: i added more tests

---
In reply to: [325641770](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1389#discussion_r325641770) [](ancestors = 325641770)

src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
jennyf19: as we discussed...only sending to /token endpoint now

---
In reply to: [325635909](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1389#discussion_r325635909) [](ancestors = 325635909)

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
jennyf19: the format was right...i removed correlation_id

---
In reply to: [325638151](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1389#discussion_r325638151) [](ancestors = 325638151)

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
jennyf19: not sure i follow.

---
In reply to: [325790729](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1389#discussion_r325790729) [](ancestors = 325790729)

src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
jennyf19: image

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -137,6 +146,93 @@ public void Flush(string correlationId)
             Callback?.Invoke(eventsToFlush.Cast<Dictionary<string, string>>().ToList());
         }
 
+        public string FetchAndResetPreviousHttpTelemetryContent()
+        {
+            if (!string.IsNullOrEmpty(_previousTelemetryPayload.ApiId))
+            {
+                string csv = CreateCSVForPreviousHttpTelem();
+               // _previousTelemetryPayload.ResetLastErrorCode();
```
MarkZuber: dead code?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -137,6 +146,93 @@ public void Flush(string correlationId)
...
+                return string.Empty;
+            }
+        }
+
+        public string FetchAndResetCurrentHttpTelemetryContent()
+        {
+            if (!string.IsNullOrEmpty(_currentTelemetryPayload.ApiId))
+            {
+                string csv = CreateCSVForCurrentHttpTelem();
+               //_currentTelemetryPayload.ResetLastErrorCode();
```
MarkZuber: dead code?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -64,6 +70,9 @@ public void StartEvent(EventBase eventToStart)
 
         public void StopEvent(EventBase eventToStop)
         {
+            ProcessEventsForCurrentHttpTelemetryContent(eventToStop);
```
MarkZuber: Does this need locking?  We could have more than one request going on at the same time which means we could be calling StopEvent in the telemetry manager more than once.  and this call pattern isn't idempotent or atomic.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
```diff
@@ -25,7 +26,18 @@ internal class AcquireTokenCommonParameters
 
         public void AddApiTelemetryFeature(ApiTelemetryFeature feature)
         {
-            _apiTelemetry[MatsConverter.AsString(feature)] = "true";
+            _apiTelemetry[MatsConverter.AsString(feature)] = TelemetryConstants.True;
+        }
+
+        public void AddApiTelemetryFeature(ApiTelemetryFeature feature, bool isTrue)
```
bgavrilMS: Is this feature (tracking the value of some flags) needed? It seems like a scope creep. Please remove if.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -0,0 +1,76 @@
...
+                evt.TryGetValue(MsalTelemetryBlobEventNames.ForceRefreshId, out string forceRefresh);
+
+                ApiId = apiId ?? string.Empty;
+                CorrelationId = correlationId ?? string.Empty;
+                LastErrorCode = errorCode ?? string.Empty;
+                ForceRefresh = forceRefresh ?? string.Empty;
+            }
+        }
+
+        public string LastErrorCode { get; set; } = string.Empty;
```
bgavrilMS: Do you need setters and default values?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -0,0 +1,76 @@
...
+        public string GetCsvAsPrevious()
+        {
+            if (string.IsNullOrWhiteSpace(ApiId))
+            {
+                return string.Empty;
+            }
+
+            // csv expected format:
+            // 1|api_id,correlation_id,last_error_code|platform_config
+            string[] myValues = new string[] {
```
bgavrilMS: try to refer to local variables without the pronoun `my`. Instead, give a name that has meaning in the local context, for example `dataToCollect`

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            string csvString = string.Join(",", myValues);
+            csvString = $"{TelemetryConstants.HttpTelemetrySchemaVersion1}{TelemetryConstants.HttpTelemetryPipe}{csvString}{TelemetryConstants.HttpTelemetryPipe}";
+            return csvString;
+        }
+
+        public string GetCsvAsCurrent()
+        {
+            // csv expected format:
+            // 1|api_id,force_refresh|platform_config
```
bgavrilMS: Hmm, there is a mismatch in semantics between MATS and HTTP telemetry. In fact, there are 3 values for all flags similar to `forceRefresh`

- user doesn't care and did not set it
- user set it to true
- user set it to false

MATS captures this 3-state because it will not log any event when user doesn't set a value explicitly. HTTP merges the meaning of "user did not set it" and "user set it to false".

PR #1405, opened 2019-09-27
Component governance on net core / net standard

bgavrilMS:
upgrade affected package only on affected platforms


bgavrilMS:
@GeoK  - I remember you did some similar work on your libraries. Is this a good enough approach for upgrading packages that have security issues in .net core metapackage?

PR #1411, opened 2019-10-03
update msbuild sdk

jennyf19:
thanks @bgavrilMS  for finding this quick and easy fix to my 72 build errors. #guru

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,5 @@
 {
-  "msbuild-sdks": {
-    "MSBuild.Sdk.Extras": "1.6.61"
-  }
+    "msbuild-sdks": {
+        "MSBuild.Sdk.Extras": "2.0.41"
+    }
 }


global.json
```diff
@@ -1,5 +1,5 @@
 {
-  "msbuild-sdks": {
-    "MSBuild.Sdk.Extras": "1.6.61"
-  }
+    "msbuild-sdks": {
+        "MSBuild.Sdk.Extras": "2.0.41"
```
bgavrilMS: Why not upgrade to the latest version 2.0.54 ? 

Not sure if the builds are ok, but it does state that .NET Core 3 SDK is required, so we might need to upgrade the builds to install that if they are failing.

jennyf19: we synced off line...updating caused other warnings..esp w/xamarin ios/android,

Issue #1098, opened 2019-04-19
UWP authentication sometimes crashes


tipa:
**Which Version of MSAL are you using ?**
MSAL 3.0.5-preview

**Platform**
UWP

In my UWP app, I only had _internetClient_ capability enabled and it works ok for most users. However, some of my users get an error like 

`{Microsoft.Identity.Client.MsalException: WAB authentication failed ---> System.IO.FileNotFoundException: The specified protocol is unknown. (Exception from HRESULT: 0x800C000D) at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task) ...(TRUNCATION)... --- End of stack trace from previous location where exception was thrown --- at UserDetailsClient.MainPage.d__2.MoveNext() ErrorCode: authentication_ui_failed}
`

Therefore I investigated about what could be the problem and found out it could be related to the app capabilites.
Could you please clarify regarding the capabilites needed for UWP.

[Here](https://github.com/Azure-Samples/active-directory-b2c-xamarin-native/issues/28#issuecomment-462495404), @jennyf19 states these 4 capabilities are needed:
```
  <Capabilities>
    <Capability Name="internetClient" />
    <Capability Name="privateNetworkClientServer" />
    <uap:Capability Name="enterpriseAuthentication" />
    <uap:Capability Name="userAccountInformation" />
  </Capabilities>
```

However, in [this ](https://github.com/Azure-Samples/active-directory-dotnet-native-uwp-v2/blob/msal3x/active-directory-dotnet-native-uwp-v2/Package.appxmanifest )example, these 4 capabilities are listed (_sharedUserCertificates_ instead of _userAccountInformation_):
```
  <Capabilities>
    <Capability Name="internetClient" />
    <Capability Name="privateNetworkClientServer" />
    <uap:Capability Name="enterpriseAuthentication" />
    <uap:Capability Name="sharedUserCertificates" />
  </Capabilities>
```
Then again, in [this ](https://github.com/Azure-Samples/active-directory-xamarin-native-v2/blob/msal3x/UserDetailsClient/UserDetailsClient.UWP/Package.appxmanifest) example, only this capability is listed:
```
  <Capabilities>
    <Capability Name="internetClient" />
  </Capabilities>
```

[...]

PR #1415, opened 2019-10-04
Retry calls to WAB and better error message - #1098

bgavrilMS:


tests/Microsoft.Identity.Test.Unit.net45/UtilTests/RetryOperationHelperTest.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.Collections.Generic;
```
jennyf19: license info

PR #1424, opened 2019-10-08
Fix #1422 - handle protected scopes better

### LINKED ISSUE #1422: [Bug] MSAL should not stop developers from adding reserved scopes
MSAL 4.4

**Platform**
all

Client Credentials and all others.

- See discussion here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1403
-  Gist of it - In client credentials on ADFS, <resource>/.default does not work. Instead, users need to add "openid"

**Actual**: MSAL throws a client exception stating that users should not add their own reserved scope "openid"
**Expected**: MSAL should not throw, instead it should just merge scopes

bgavrilMS:
Fix for #1422

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -5,12 +5,11 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
-using Microsoft.Identity.Client.Core;
-using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
-using Microsoft.Identity.Client.OAuth2;
-using Microsoft.Identity.Client.Utils;
-using System;
 using Microsoft.Identity.Client.Cache.Items;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
+using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Internal.Requests
 {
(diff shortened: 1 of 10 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -83,6 +83,14 @@ protected override void EnrichTelemetryApiEvent(ApiEvent apiEvent)
             apiEvent.IsConfidentialClient = true;
         }
 
+        protected override SortedSet<string> GetDecoratedScope(SortedSet<string> inputScope)
+        {
+            // Client credentails should not add the reserved scopes
```
jennyf19: credentials

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -83,6 +83,14 @@ protected override void EnrichTelemetryApiEvent(ApiEvent apiEvent)
...
+            // "openid", "profile" and "offline_access" 
+            // becaue AT is on behalf of an app (no profile, no IDToken, no RT)
```
jennyf19: because

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -87,24 +87,21 @@ private void LogRequestStarted(AuthenticationRequestParameters authenticationReq
...
         }
 
-        protected SortedSet<string> GetDecoratedScope(SortedSet<string> inputScope)
+        protected virtual SortedSet<string> GetDecoratedScope(SortedSet<string> inputScope)
         {
-            SortedSet<string> set = new SortedSet<string>(inputScope.ToArray());
-            set.UnionWith(ScopeHelper.CreateSortedSetFromEnumerable(OAuth2Value.ReservedScopes));
+            // OAuth spec states that scopes are case sensitive, but 
+            // merge the reserved scopes in a case insensitive way, to 
+            // avoid sending things like "openid OpenId" (note that EVO is tollerant of this)
```
jennyf19: nit: should you put "Microsoft identity platform" instead of EVO? more specific and other IdPs maybe case sensitive and following the spec

tests/Microsoft.Identity.Test.Integration.net45/Infrastructure/SeleniumExtensions.cs
```diff
@@ -28,7 +28,7 @@ public static IWebDriver CreateDefaultWebDriver()
 
             // ~2x faster, no visual rendering
             // remove when debugging to see the UI automation
-            options.AddArguments("headless");
+            //options.AddArguments("headless");
```
jennyf19: remove?

PR #1440, opened 2019-10-14
Jennyf/ios

jennyf19:


src/client/Microsoft.Identity.Client/Platforms/iOS/SystemWebview/SystemWebUI.cs
```diff
@@ -72,10 +72,14 @@ public void Authenticate(Uri authorizationUri, Uri redirectUri, RequestContext r
                         {
                             asWebAuthenticationSession.PresentationContextProvider =
```
henrik-me: >asWebAuthenticationSession.PresentationContextProvider = [](start = 28, length = 56)

can the presentationcontext be set in the contructor? then perhaps it will not be required to sync to the ui/main thread?

bgavrilMS: I tried this but the call to `.Start()` call seems to be required to be on the UI thread. I assume that when the auth session is built using a presentationContext, Start() will try to access a property of the context.

Reading the MSAL.ObjC code I believe they expect all the logic around the system UI to occur on the UI thread: https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/blob/23c936a528cbd94b5f60c9b2d2623596bb5dec67/IdentityCore/src/oauth2/MSIDWebviewFactory.m#L97

bgavrilMS: I will commit this so that I can get a new package for testing. We can revisit our options at standup today,.

PR #1442, opened 2019-10-15
cleanup for presentationContext

jennyf19:
i could not find any official documentation on .Start() and the main ui thread...also, Xamarin has not come out w/docs yet for the presentationContext and iOS 13 methods.


henrik-me:
nice :)

Issue #1294, opened 2019-07-26
[Feature Request] HasTokens property on the Token cache serialization argument


TiagoBrenck:
**Is your feature request related to a problem? Please describe.**
As discussed on this thread, https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/130, an empty `MsalCacheItemBase` JSON is generated when clearing the token cache. This is not that harmful but we could avoid saving this empty JSON structure.

**Describe the solution you'd like**
As suggested by the developer, we could have a property like `IsEmpty` on `TokenCacheNotificationArgs` that could be checked before saving the token in memory (or DB).

**Additional context**
https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/130

jmprieur:
Adding supportability as we can avoid questions like "why do we save empty caches" when an account is removed.

jmprieur:
Requires more spec.

bgavrilMS:
We discussed offline with @jmprieur about this. The term `is_empty` is probably too overloaded, since the cache can contain: 

- tokens, 
- accounts  
- app metadata.
- other artefacts in the future (extensibility)

Today, it makes sense to lower COGS of apps integrating with MSAL if there are no accounts. No accounts means no tokens. This holds even if an FRT is present, you still need an account to use it...

So I would propose that for the user cache, IsEmpty means "no accounts". 
The app cache is different - it only holds access tokens. There are not Accounts, IDTokens, RTs. So IsEmpty means "no **unexpired** access tokens". 

We still need to find a name for the flag that conveys these 2 meanings, but I think the semantics should be the ones I just mentioned.

bgavrilMS:
user cache is empty if no accounts exist 
app cache does not matter

bgavrilMS:
Impact: reduce COGS of apps

bgavrilMS:
I'm working on this but I'm slightly concerned about the naming. So we can only tell the developer if the cache contains tokens in the notifications that comes AFTER the update (OnAfterUpdate). 

But what about the other 2 callbacks - `OnBeforeWrite` and `OnBeforeAccess`? Here we are firing the callback before the token is read from disk. Does it make sense to tell the user if there are tokens in the cache in memory?

jmprieur:
yes, @bgavrilMS I think it's a good proposal.

PR #1448, opened 2019-10-16
Expose IsApplicationTokenCache

bgavrilMS:
This will simplify extension libraries such as Microsoft.Identity.Web

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCache.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.cs
@@ -45,6 +45,9 @@ namespace Microsoft.Identity.Client
         ITokenCacheAccessor ITokenCacheInternal.Accessor => _accessor;
         ILegacyCachePersistence ITokenCacheInternal.LegacyPersistence => LegacyCachePersistence;
 
+        bool IsAppTokenCache { get; }
+        bool ITokenCacheInternal.IsApplicationTokenCache => IsAppTokenCache;
+
         private readonly SemaphoreSlim _semaphoreSlim = new SemaphoreSlim(1, 1);
 
         SemaphoreSlim ITokenCacheInternal.Semaphore => _semaphoreSlim;
@@ -54,11 +57,12 @@ namespace Microsoft.Identity.Client
         /// The recommended way to get a cache is by using <see cref="IClientApplicationBase.UserTokenCache"/>
         /// and <c>IConfidentialClientApplication.AppTokenCache</c> once the app is created.
         /// </summary>
-        public TokenCache() : this((IServiceBundle)null)
+        [Obsolete("The recommended way to get a cache is by using IClientApplicationBase.UserTokenCache or IClientApplicationBase.AppTokenCache")]
+        public TokenCache() : this((IServiceBundle)null, false)
         {
         }
 
-        internal TokenCache(IServiceBundle serviceBundle)
+        internal TokenCache(IServiceBundle serviceBundle, bool isApplicationTokenCache)
         {
             var proxy = serviceBundle?.PlatformProxy ?? PlatformProxyFactory.CreatePlatformProxy(null);
             _accessor = proxy.CreateTokenCacheAccessor();
@@ -81,6 +85,8 @@ namespace Microsoft.Identity.Client
             SetIosKeychainSecurityGroup(serviceBundle.Config.IosKeychainSecurityGroup);
 #endif // iOS
 
+            IsAppTokenCache = isApplicationTokenCache;
+
             // Must happen last, this code can access things like _accessor and such above.
             ServiceBundle = serviceBundle;
         }
@@ -88,12 +94,10 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// This method is so we can inject test ILegacyCachePersistence...
         /// </summary>
-        /// <param name="serviceBundle"></param>
-        /// <param name="legacyCachePersistenceForTest"></param>
-        internal TokenCache(IServiceBundle serviceBundle, ILegacyCachePersistence legacyCachePersistenceForTest)
-            : this(serviceBundle)
+        internal TokenCache(IServiceBundle serviceBundle, ILegacyCachePersistence legacyCachePersistenceForTest, bool isApplicationTokenCache)
+            : this(serviceBundle, isApplicationTokenCache)
         {
-            LegacyCachePersistence = legacyCachePersistenceForTest;
+            LegacyCachePersistence = legacyCachePersistenceForTest;            
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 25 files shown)


src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -43,5 +45,12 @@ public sealed partial class TokenCacheNotificationArgs
         /// Not all cache operations modify the state of the cache.
         /// </summary>
         public bool HasStateChanged { get; internal set; }
+
+        /// <summary>
+        /// Indicates wheather the cache change occured in the <see cref="ClientApplicationBase.UserTokenCache" />
+        /// or in the <see cref="ConfidentialClientApplication.AppTokenCache"/>. The latter is used in Client Credential grant, 
+        /// i.e. <see cref="ConfidentialClientApplication.AcquireTokenForClient(System.Collections.Generic.IEnumerable{string})"/>
+        /// </summary>
+        public bool IsApplicationTokenCache { get; }
```
bgavrilMS: Alternatively, we could invert the flag, expose `IsUserTokenCache` if you think it's more intuitive.

jmprieur: this looks good to me.

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -43,5 +45,12 @@ public sealed partial class TokenCacheNotificationArgs
         /// Not all cache operations modify the state of the cache.
         /// </summary>
         public bool HasStateChanged { get; internal set; }
+
+        /// <summary>
+        /// Indicates whether the cache change occured in the UserTokenCache
+        /// or in the AppTokenCache. The latter is used in Client Credential grant, 
+        /// which is not available on all platforms.
```
henrik-me: A link to a aka.ms page describing why this is needed might be a good idea?

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -43,5 +45,12 @@ public sealed partial class TokenCacheNotificationArgs
...
+        /// </summary>
+        public bool IsApplicationTokenCache { get; }
```
henrik-me: >Token [](start = 33, length = 5)

nit: It may not need to say token?

src/client/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -45,6 +45,9 @@ public sealed partial class TokenCache : ITokenCacheInternal
         ITokenCacheAccessor ITokenCacheInternal.Accessor => _accessor;
         ILegacyCachePersistence ITokenCacheInternal.LegacyPersistence => LegacyCachePersistence;
 
+        bool IsAppTokenCache { get; }
```
henrik-me: >bool IsAppTokenCache { get; } [](start = 8, length = 29)

private? also do we need this? Can't we simply use the one on ITokenCacheInternal?

henrik-me:
Please provide more information on how this will help developers. Links to relevant issue conversations, wiki pages etc.

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
jennyf19: agree w/henrik.

henrik-me:
@jmprieur : It's not obvious from the PR or related issue that this is the intention (IMO). 

In we agreed in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1294 that we needed a bit more spec to make sure the feature was the right one? Perhaps the issue should just be updated with this information?

src/client/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -54,11 +57,12 @@ public sealed partial class TokenCache : ITokenCacheInternal
         /// The recommended way to get a cache is by using <see cref="IClientApplicationBase.UserTokenCache"/>
         /// and <c>IConfidentialClientApplication.AppTokenCache</c> once the app is created.
         /// </summary>
-        public TokenCache() : this((IServiceBundle)null)
+        [Obsolete("The recommended way to get a cache is by using IClientApplicationBase.UserTokenCache or IClientApplicationBase.AppTokenCache")]
```
henrik-me: >[Obsolete("The recommended w [](start = 8, length = 28)

Will this not be sort of a breaking change?

henrik-me: Can we change it now with-out updating the major version?

bgavrilMS: I don't think developers use this ctor. We can remove the Obsolete attribute if you think this will cause breaking changes.

henrik-me: I don't think it's a matter of whether I think this is a breaking change or not. I think it is. @jmprieur can we ship with this change with-out bumping the major version?

jmprieur: @henrik-me  @bgavrilMS 
I agree that this is a breaking change (as if dvelopers have enabled warning as errors in build, they will be broken).
Now I don't think of any scenario where people would want to use this constructor. This is actually bad if they use it, so they want to know, and the fix is easy enough

I don't feel that I'd want to bump up the major number for this, though as this would provoke some resistance, so I'm fine with taking this theorical breaking change (I hope not many people will be affected, and if they are they have to react anyway), without moving to 5.x

henrik-me: Can we please add an aka.ms link to the additional cache page in the wiki or in the ms.docs ?

---
In reply to: [339514356](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1448#discussion_r339514356) [](ancestors = 339514356)

jmprieur: @bgavrilMS @jennyf19 @henrik-me, I've created a new aka.ms for this:
https://aka.ms/msal-net-web-token-cache

src/client/Microsoft.Identity.Client/TokenCache.cs
bgavrilMS: The pattern used here is: 

- ITokenCacheInternal is implemented as an explicit interface. `internal` interfaces can only be explicit, since the TokenCache object is public. 
- When using explicit interfaces, one needs to make an explicit cast to the interface to be able to access the methods / properties on that interface. 
- Use the trick above to overcome the explicit cast. This makes the code more readable, otherwise it would be littered with interface casting.

src/client/Microsoft.Identity.Client/TokenCache.cs
bgavrilMS: And I've created https://aka.ms/msal-net-app-cache-serialization

bgavrilMS: I'll use yours Jean-Marc, mine points to the wiki, which I've updated.

PR #1450, opened 2019-10-16
Enable system browser on xamarin macos

bgavrilMS:
The default remains the embedded browser


henrik-me:
>     internal class TcpInterceptor : ITcpInterceptor

As you are in this area... I think adding more logging may be needed based on discussions we had earlier today?

---
Refers to: src/client/Microsoft.Identity.Client/Platforms/Shared/DefaultOSBrowser/TcpInterceptor.cs:23 in d224a6f. [](commit_id = d224a6fa32ef0f363c0e26514a5372982585cbb5, deletion_comment = False)

henrik-me:
Consider if we can add logging to understand why we would not be getting the callback (based on the redirect uri)

henrik-me:
Aside: Seems like the non yaml build stuff can be removed

bgavrilMS:
Add more logging.

Issue #1449, opened 2019-10-16
[Bug] Xamarin.Android typeload exception with MSAL 4.5.0


bgavrilMS:
**Which Version of MSAL are you using ?**
Note that to get help, you need to run the latest version. Preview version are also ok. 
For ADAL, please log issues to https://github.com/AzureAD/azure-activedirectory-library-for-dotnet
4.5.0

**Platform**
android only

**Repro**
1. Update to latest VS (RTM)
2. Create a new XForms solution
3. Add MSAL 4.5.0 (just add the nuget packge)
4. Run

VS 16.3.5
Xamarin: 16.3.0.277
Xamarin.Android.SDK 10.0.3.0

UNHANDLED EXCEPTION:
10-16 13:18:43.263 I/MonoDroid(23861): System.TypeInitializationException: The type initializer for 'Attribute' threw an exception. ---> System.TypeLoadException: Could not resolve type with token 01000022 from typeref (expected class 'Microsoft.Identity.Client.Ref.Resource' in assembly 'Microsoft.Identity.Client, Version=4.5.0.0, Culture=neutral, PublicKeyToken=0a613f4dd989e8ae')
10-16 13:18:43.263 I/MonoDroid(23861): at Android.Runtime.ResourceIdManager.UpdateIdValues () [0x00097] in <11a340ccc8de43f09c97400139266ef5>:0
10-16 13:18:43.263 I/MonoDroid(23861): at Xamarin.Forms.Platform.Android.Resource+Attribute..cctor () [0x012a2] in D:\a\1\s\Xamarin.Forms.Platform.Android\Resources\Resource.Designer.cs:1579
10-16 13:18:43.263 I/MonoDroid(23861): --- End of inner exception stack trace ---
10-16 13:18:43.263 I/MonoDroid(23861): at (wrapper managed-to-native) System.Object.__icall_wrapper_mono_generic_class_init(intptr)
10-16 13:18:43.263 I/MonoDroid(23861): at Xamarin.Forms.Platform.Android.AppCompat.NavigationPageRenderer.ActionBarHeight () [0x00000] in D:\a\1\s\Xamarin.Forms.Platform.Android\AppCompat\NavigationPageRenderer.cs:476
[...]

PR #1454, opened 2019-10-17
Fix for #1449 - do not include Android Resources in dll

bgavrilMS:


jennyf19:
@bgavrilMS I assume once this is merged, we will run a 4.5.1 release...correct?

bgavrilMS:
I think we should. I am now running a release build on the topic branch and plan to ask people to validate first. But it works on my machine (TM) so we can definitely ship ... when did this ever go wrong? :D

jennyf19:
let's merge it and run it on master, to save time and we can validate. looks like a legit fix.

PR #1458, opened 2019-10-18
Integration tests hardening and more logging

bgavrilMS:


tests/Microsoft.Identity.Test.Common/Core/Helpers/TraceLogger.cs
```diff
@@ -0,0 +1,98 @@
...
+            _prefix = prefix;
+        }
+
+        public Guid CorrelationId => Guid.Empty;
+
+        public string ClientName => "MSAL.Test";
+
+        public string ClientVersion => "0";
+
+        public bool PiiLoggingEnabled => true;
```
trwalke: Is this to test Pii logging with our test accounts?

bgavrilMS: Normally we use the LogCallback to do logging, but some of our internal objects, use the actual ICoreLogger object. This is just a logger that logs via Trace.WriteLine (which end up in the test log)

PR #1464, opened 2019-10-19
Update cache tests to enable the java to .net and python tests

henrik-me:
A few issues:
* Maven was not found as the process is a cmd, 
* The process runner was not able to run the cmd with-out a complete path to the cmd. 
* The latest Java library had introduced some updates to the public api
* Python executable location improved to enable local test runs (should not be needed but was not working locally on my box)

TODO:
* Lookup env var M2_HOME to find maven installation
* Add logging to the Java cache test

### DIFF
--- a/tests/CacheCompat/CommonCache.Test.MsalJava/pom.xml
+++ b/tests/CacheCompat/CommonCache.Test.MsalJava/pom.xml
@@ -24,7 +24,7 @@
         <dependency>
             <groupId>com.microsoft.azure</groupId>
             <artifactId>msal4j</artifactId>
-            <version>0.5.0-preview</version>
+            <version>0.6.0-preview</version>
         </dependency>
         <dependency>
             <groupId>com.google.code.gson</groupId>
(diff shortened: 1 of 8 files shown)


henrik-me:
Verified on a release build and the cache tests passes, however one integration test fails:
SeleniumGetAuthCode_RedeemForAt_CommonAuthority_Async

tests/CacheCompat/CommonCache.Test.MsalJava/pom.xml
```diff
@@ -24,7 +24,7 @@
         <dependency>
             <groupId>com.microsoft.azure</groupId>
             <artifactId>msal4j</artifactId>
-            <version>0.5.0-preview</version>
+            <version>0.6.0-preview</version>
```
bgavrilMS: We should look into version ranges / using LATEST, but as a separate work item, as it does not seem to be trivial.

bgavrilMS:
I’ve made some changes to harden that test, the stuff done here has no impact.

bgavrilMS:
Tests are passing, this can be checked in.

henrik-me:
> Tests are passing, this can be checked in.

I will be optimizing the code looking for Maven and if tests passes I will merge. thanks

Issue #1461, opened 2019-10-18
Question - Where can I find Release Notes?


udlose:
I'm not able to find Release Notes for updates to the Microsoft.Identity.Client nuget package.  Do they exist anywhere?  TIA

jennyf19:
@udlose Like these [release notes](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases) or the [changelog.txt](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/changelog.txt)? we also have blog posts for some of the releases, in the side bar under ["News"](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/msal-net-4.3)....or is there something else you are looking for? Any suggestions on a more searchable place to find them are appreciated as well.

jmprieur:
@udlose : they are also all referenced from the roadmap: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki#roadmap

udlose:
@jennyf19 , @jmprieur 
Thanks. I'd normally expect to find the Release Notes on the default page for the GitHub site. If not the Release Notes themselves, then at least a link to them.

jmprieur:
Thanks for the feedback; @udlose 
I've updated the README.md to make it clearer

bgavrilMS:
@jmprieur - for example out see https://www.nuget.org/packages/Moq/

jennyf19:
Thanks @bgavrilMS

PR #1465, opened 2019-10-21
Add PackageReleaseNotes to MSAL

bgavrilMS:
#1461

### DIFF
--- a/src/client/Directory.Build.props
+++ b/src/client/Directory.Build.props
@@ -35,6 +35,7 @@
     <Authors>Microsoft</Authors>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=844761</PackageProjectUrl>
+    <PackageReleaseNotes>A changelog is available at https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/changelog.txt </PackageReleaseNotes>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
     <Description>
       This package contains the binaries of the Microsoft Authentication Library for .NET (MSAL.NET).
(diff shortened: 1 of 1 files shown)


src/client/Directory.Build.props
```diff
@@ -35,6 +35,7 @@
     <Authors>Microsoft</Authors>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=844761</PackageProjectUrl>
+    <PackageReleaseNotes>A changelog is available at https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/changelog.txt </PackageReleaseNotes>
```
jmprieur: Thanks @bgavrilMS this is an awesome improvement.
I wonder if we should not add 
```
and the roadmap (including blog post for releases) from https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki#roadmap
```

bgavrilMS: Ok, but let's exclude the parenthesis - too many words for such a small place.

src/client/Directory.Build.props
```diff
@@ -35,6 +35,7 @@
     <Authors>Microsoft</Authors>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://go.microsoft.com/fwlink/?linkid=844761</PackageProjectUrl>
+    <PackageReleaseNotes>The changelog is available at https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/changelog.txt and the roadmap at https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki#roadmap </PackageReleaseNotes>
```
jennyf19: this is pointing to adal's release notes...am i missing something? @jmprieur @bgavrilMS

jmprieur: Good catch, Jenny. My bad

jennyf19: @jmprieur i fixed it.

jmprieur: @jennyf19 thank you.

PR #1466, opened 2019-10-21
Fix #1407 - /devicecode endpoint sends errors with 200 OK status

### LINKED ISSUE #1407: [Bug] Device Code Flow fails with misleading error message if app is misconfigured
MSAL 4.4.0

    * [x] Device code flow (browserless)

**Repro**

1. Register an app and do not change the "Allow public client flows" setting which enables Device Code Flow, IWA and ROPC
2. Use MSAL to perform Device Code Flow 

**Expected behavior**
Friendly exception explaining the issue

**Actual behavior**
MSALClientException with code "code_expired"

bgavrilMS:
Fix #1407

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
@@ -51,7 +51,9 @@ namespace Microsoft.Identity.Client.Internal.Requests
             var response = await client.ExecuteRequestAsync<DeviceCodeResponse>(
                                builder.Uri,
                                HttpMethod.Post,
-                               AuthenticationRequestParameters.RequestContext).ConfigureAwait(false);
+                               AuthenticationRequestParameters.RequestContext, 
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -51,7 +51,9 @@ internal override async Task<AuthenticationResult> ExecuteAsync(CancellationToke
             var response = await client.ExecuteRequestAsync<DeviceCodeResponse>(
                                builder.Uri,
                                HttpMethod.Post,
-                               AuthenticationRequestParameters.RequestContext).ConfigureAwait(false);
+                               AuthenticationRequestParameters.RequestContext, 
+                               // Normally AAD responds with an error HTTP code, but /devicecode endpoint sends errors on 200OK
```
jmprieur: how weird!

bgavrilMS: Yes. I thought of discussing this with ESTS folks, but I doubt they will change the behaviour of a production endpoint.

PR #1470, opened 2019-10-24
Add support for releasing the extension out of band with MSAL

bgavrilMS:


build/pipeline-releasebuild.yaml
```diff
@@ -39,6 +39,9 @@ steps:
     BuildPlatform: '$(BuildPlatform)'
     BuildConfiguration: '$(BuildConfiguration)'
     MsalClientSemVer: $(MsalClientSemVer)
+    # Leave empty to release the desktop extenion with a project refrence to msal
```
jmprieur: extenion => extension

build/pipeline-releasebuild.yaml
```diff
@@ -39,6 +39,9 @@ steps:
...
+    # Or set a version to take a dependency on a pre-released MSAL NuGet package
```
jmprieur: pre-released? do you mean previously released ?

src/extensions.msal/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -19,17 +20,22 @@
...
-    <MsalClientSemVer Condition="'$(MsalClientSemVer)' != ''">$(MsalClientSemVer)-preview</MsalClientSemVer>
-    <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
-    <Version>$(MsalClientSemVer)</Version>
+    <DesktopExtensionVersion Condition="'$(ExtensionUsesMsalFromNuGetVersion)' == '' and '$(MsalClientSemVer)' != ''">
+      $(MsalClientSemVer)-preview
+    </DesktopExtensionVersion>
+    <DesktopExtensionVersion  Condition="'$(ExtensionUsesMsalFromNuGetVersion)' != ''" >
+      $(ExtensionUsesMsalFromNuGetVersion)-preview
+    </DesktopExtensionVersion>
+    <DesktopExtensionVersion Condition="$(DesktopExtensionVersion) == ''">4.5.0-internal</DesktopExtensionVersion>
```
jmprieur: Is it something that is hardcoded?

src/extensions.msal/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -82,8 +88,12 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All" />
   </ItemGroup>
 
-  <ItemGroup>
+  <!-- Set the variable MsalPackageReferenceVersion to an MSAL version to relesae the extension out of band with MSAL-->
```
jennyf19: Relesae -> release

src/extensions.msal/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
bgavrilMS: Yes, but I will remove this as there are so many defaults now ...

src/extensions.msal/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
```diff
@@ -3,17 +3,16 @@
   <PropertyGroup>
     <TargetFrameworkNetDesktop>net45</TargetFrameworkNetDesktop>
     <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
-    <TargetFrameworkNetCore2>netcoreapp2.1</TargetFrameworkNetCore2>
-    <TargetFrameworkNetCore3>netcoreapp3.0</TargetFrameworkNetCore3>
```
bgavrilMS: Looks like the extension was supposed to be built for .net core 2 and .net core 3, but there is no .net core 3 logic right now, so reverting this. Also, the flags were wrong. We need a lot of scrutiny for releasing the extension from here ...

bgavrilMS:
After further inspection I've decided to drop this. The ADAL + MSAL desktop extensions share a lot of code between them and there are many unanswered questions related to shipping in sync with ADAL / MSAL and out of band. The ADAL extension has not been ported to ADAL, and clearly it will be updated much more frequently than ADAL itself. 

If we are going to add all this functionality into the libraries, then we need to find an answer to the shared functionality problem. Until then, the existing repo https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet and out of band releases are more tractable. 

CC @henrik-me

PR #1475, opened 2019-10-25
iOS Broker Part II

jennyf19:
- get application token in broker response
- save in keychain
- send in subsequent requests to v3 broker

tested w/iOS v3 broker

@bgavrilMS maybe you'll have time to take a look at this. also one for [ADAL](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/pull/1672), but the changes are the same, so whatever comments made here, i will update in ADAL as well. 
Am also going to confirm w/iOS team tomorrow that this works as expected and they see the right stuff on their end. Also, am not sure which specific use case or test should be run to ensure the expected behavior happens, so will find out tomorrow. wanted to give you time to review, if you have time. 😼 thx.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
@@ -15,6 +15,7 @@ using Microsoft.Identity.Client.Internal.Broker;
 using Microsoft.Identity.Client.UI;
 using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
 using System.Globalization;
+using Security;
 
 namespace Microsoft.Identity.Client.Platforms.iOS
 {
@@ -96,11 +97,18 @@ namespace Microsoft.Identity.Client.Platforms.iOS
             {
                 _brokerRequestNonce = Guid.NewGuid().ToString();
                 brokerPayload[iOSBrokerConstants.BrokerNonce] = _brokerRequestNonce;
+
+                string applicationToken = ReadBrokerApplicationTokenFromKeychain(brokerPayload);
+
+                if (!string.IsNullOrEmpty(applicationToken))
+                {
+                    brokerPayload[iOSBrokerConstants.ApplicationToken] = applicationToken;
+                }
             }
 
             if (brokerPayload.ContainsKey(iOSBrokerConstants.Claims))
             {
-                brokerPayload.Add(iOSBrokerConstants.SkipCache, BrokerParameter.SkipCache);
+                brokerPayload[iOSBrokerConstants.SkipCache] = BrokerParameter.SkipCache;
                 string claims = Base64UrlHelpers.Encode(brokerPayload[BrokerParameter.Claims]);
                 brokerPayload[BrokerParameter.Claims] = claims;
             }
@@ -198,6 +206,14 @@ namespace Microsoft.Identity.Client.Platforms.iOS
                         ErrorDescription = MsalErrorMessage.BrokerNonceMismatch
                     };
                 }
+
+                if (responseDictionary.ContainsKey(iOSBrokerConstants.ApplicationToken))
+                {
+                    TryWriteBrokerApplicationTokenToKeychain(
+                        responseDictionary[BrokerResponseConst.ClientId], 
+                        responseDictionary[iOSBrokerConstants.ApplicationToken]);
+                }
+
                 brokerTokenResponse = MsalTokenResponse.CreateFromBrokerResponse(responseDictionary);
             }
             else
@@ -224,8 +240,8 @@ namespace Microsoft.Identity.Client.Platforms.iOS
                 string brokerResponseNonce = brokerResponseDictionary[BrokerResponseConst.iOSBrokerNonce];
 
                 bool ok = string.Equals(
-                    brokerResponseNonce, 
-                    _brokerRequestNonce, 
+                    brokerResponseNonce,
+                    _brokerRequestNonce,
                     StringComparison.InvariantCultureIgnoreCase);
 
                 if (!ok)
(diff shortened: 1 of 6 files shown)


src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
```diff
@@ -222,6 +240,27 @@ private SecStatusCode Save(IiOSKey key, string payload)
             return secStatusCode;
         }
 
+        internal SecStatusCode Save(string clientIdAsKey, string applicationToken)
```
bgavrilMS: SaveBrokerApplicationToken

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
```diff
@@ -147,6 +146,25 @@ public IEnumerable<MsalAccountCacheItem> GetAllAccounts()
...
+            };
+
+            SecRecord record = SecKeyChain.QueryAsRecord(queryRecord, out SecStatusCode resultCode);
+
+            if (resultCode == SecStatusCode.Success)
+            {
+                return record.ValueData.ToString(NSStringEncoding.UTF8);
+            }
+
+            return null;
```
bgavrilMS: Consider using `SecRecord TryGetBrokerApplicationToken(string clientId, out string appToken)` to be able to log SecRecord

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
```diff
@@ -147,6 +146,25 @@ public IEnumerable<MsalAccountCacheItem> GetAllAccounts()
                 .ToList();
         }
 
+        internal string GetBrokerApplicationToken(string clientId)
+        {
+            var queryRecord = new SecRecord(SecKind.GenericPassword)
```
bgavrilMS: Changes are ok, but it's not clear (i.e. please add some comments) 

a. what is the app token
b. what is the consequence of not having an app token. Today you treat keychaing access errors silently, so it is optional? I'd recommend we make it mandatory - it's better to fail 

KeyChain access has been notoriously bad - I think we need more logging here.

jennyf19: iOS team is not throwing if they have a [keychain error w/the apptoken](https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/523/files), they are just logging the info here. the consequence is a possible extra consent screen depending on if other parameters are not met. i don't think throwing makes sense here.
I have added some comments
we are logging the result of trying to save the application token, so we'll know if it's a duplicate or something else happened.

---
In reply to: [339137282](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1475#discussion_r339137282) [](ancestors = 339137282)

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
jennyf19: i think it makes more sense to return the secstatuscode, then the record, which may or may not have been saved to the keychain. at least this way, we will know if it's not a success, what the error is.

---
In reply to: [339124688](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1475#discussion_r339124688) [](ancestors = 339124688)

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
jennyf19: i was just following the other methods here, but i also agree it makes sense to specify it. thx.

---
In reply to: [339124173](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1475#discussion_r339124173) [](ancestors = 339124173)

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
bgavrilMS: Fair enough.

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -96,6 +97,13 @@ private void AddIosSpecificParametersToPayload(Dictionary<string, string> broker
             {
                 _brokerRequestNonce = Guid.NewGuid().ToString();
                 brokerPayload[iOSBrokerConstants.BrokerNonce] = _brokerRequestNonce;
+
+                string applicationToken = ReadBrokerApplicationTokenFromKeychain(brokerPayload);
+
+                if (!string.IsNullOrEmpty(applicationToken))
+                {
+                    brokerPayload.Add(iOSBrokerConstants.ApplicationToken, applicationToken);
```
henrik-me: >brokerPayload.Add [](start = 20, length = 17)

Can there be more than one application token in the payload? If not any reason to not assign the value directly in the dictionary?

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBroker.cs
```diff
@@ -242,6 +258,32 @@ private bool ValidateBrokerResponseNonceWithRequestNonce(Dictionary<string, stri
             return true;
         }
 
+        private void WriteBrokerApplicationTokenToKeychain(string clientId, string applicationToken)
+        {
+            iOSTokenCacheAccessor iOSTokenCacheAccessor = new iOSTokenCacheAccessor();
+
+            SecStatusCode secStatusCode = iOSTokenCacheAccessor.SaveBrokerApplicationToken(clientId, applicationToken);
```
henrik-me: >SecStatusCode secStatusCode = iOSTokenCacheAccessor.SaveBrokerApplicationToken(clientId, applicationToken); [](start = 12, length = 107)

could this result in an exception or will a failure result always be in SecStatusCode? if not the log statement may not be executed

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
```diff
@@ -147,6 +146,27 @@ public IEnumerable<MsalAccountCacheItem> GetAllAccounts()
...
+        internal SecStatusCode TryGetBrokerApplicationToken(string clientId, out string appToken)
+        {
+            var queryRecord = new SecRecord(SecKind.GenericPassword)
+            {
+                AccessGroup = _keychainGroup,
+                Account = clientId,
+                Service = iOSBrokerConstants.iOSBroker,
+            };
+
+            SecRecord record = SecKeyChain.QueryAsRecord(queryRecord, out SecStatusCode resultCode);
```
henrik-me: nit: why not secStatusCode :)

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
```diff
@@ -222,6 +242,34 @@ private SecStatusCode Save(IiOSKey key, string payload)
             return secStatusCode;
         }
 
+        internal SecStatusCode SaveBrokerApplicationToken(string clientIdAsKey, string applicationToken)
```
henrik-me: >SaveBrokerApplicationToken [](start = 31, length = 26)

TrySaveBrokerApplicationToken ... and make sure it doesn't throw an exception since you arr returning a status code?

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSTokenCacheAccessor.cs
jennyf19: because

---
In reply to: [339276501](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1475#discussion_r339276501) [](ancestors = 339276501)

Issue #1456, opened 2019-10-17
[Bug] AcquireTokenSilent always calls a new refresh token when .WithAuthority is set on it.


peombwa:
**Which Version of MSAL are you using ?**
MSAL 4.5.0

**Platform**
net45, netcore

**What authentication flow has the issue?**
AcquireTokenSilent

Other? - please describe;

**Is this a new or existing app?**
This is a new app or experiment

**Repro**

```csharp
public async Task<AuthenticationResult> GetAuthTokenForGraphAsync(string clientId, string cacheFileName, string cacheFilePath)
{
	IPublicClientApplication publicClientApp = PublicClientApplicationBuilder
						   .Create(clientId)
						   .Build();
						   
	StorageCreationProperties storageCacheProperty = new StorageCreationPropertiesBuilder(cacheFileName, cacheFilePath, clientId)
		.Build();

	MsalCacheHelper msalCacheHelper = await MsalCacheHelper.CreateAsync(storageCacheProperty);
	msalCacheHelper.RegisterCache(publicClientApp.UserTokenCache);
	
	List<string> scopes = new List<string> { "User.Read" };

	var accounts = await publicClientApp.GetAccountsAsync();
	try
	{
		return await publicClientApp.AcquireTokenSilent(scopes, accounts.FirstOrDefault())
		// .WithAuthority(publicClientApp.Authority) causes MSAL to always call a new refresh token even if the token cache is in a valid state.
				.WithAuthority(publicClientApp.Authority)
				.ExecuteAsync();
	}
	catch (MsalUiRequiredException)
	{
		return await publicClientApp.AcquireTokenWithDeviceCode(scopes,
					deviceCodeResult =>
					{
						Console.WriteLine(deviceCodeResult.Message);
						return Task.FromResult(0);
					}).ExecuteAsync();
	}
}

var authenticationProvider = new DelegateAuthenticationProvider(async (request) =>
{
	string clientId = YOUR_CLIENT_ID;
	string cacheFileName = "UserCacheFileName.bin3";
	string cacheFilePath = YOUR_CACHE_FILE_PATH;
	AuthenticationResult result = await GetAuthTokenForGraphAsync(clientId, cacheFileName, cacheFilePath);
	request.Headers.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", result.AccessToken);
});

GraphServiceClient graphClient = new GraphServiceClient(authenticationProvider);

[...]

PR #1478, opened 2019-10-27
Fix for #1456 - ignore common authority override

bgavrilMS:


src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -36,22 +27,65 @@ protected Authority(AuthorityInfo authorityInfo)
...
         #region Builders
-        public static Authority CreateAuthorityWithOverride(AuthorityInfo requestAuthorityInfo, AuthorityInfo configAuthorityInfo)
+
+        /// <summary>
+        /// Figures out the authority based on the authority from the config and the authority from the request, 
+        /// and optionally the homeAccountTenantId, which has an impact on AcquireTokenSilent
+        /// 
+        /// The algorithm is: 
+        /// 
+        /// 1. If there is no request authority (i.e. no authority override), use the config authority.
```
jennyf19: >request [](start = 30, length = 7)

nit: the meaning of "request" authority is not very clear here, esp from a customer perspective. is it the authority used in the request?

tests/devapps/NetCoreTestApp/Program.cs
```diff
@@ -107,88 +107,89 @@ 1. IWA
...
-                        return;
-                    default:
-                        break;
+                                BrowserRedirectSuccess = new Uri("https://www.bing.com?q=why+is+42+the+meaning+of+life")
+                            };
+
+                            var cts = new CancellationTokenSource();
+                            authTask = pca.AcquireTokenInteractive(s_scopes)
+                                .WithSystemWebViewOptions(options)
+                                 .WithExtraQueryParameters("dc=PROD-WST-CORP")
```
jennyf19: do you want to leave this in?

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/UnifiedCacheTests.cs
```diff
@@ -77,6 +80,7 @@ public void UnifiedCache_MsalStoresToAndReadRtFromAdalCache()
                             TestConstants.s_scope.ToArray())
                     });
 
+
                 // Using RT from Adal cache for silent call
```
jennyf19: nit: remove extra line?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -1200,6 +1200,8 @@ public async Task MultiTenantWithAuthorityOverrideAsync()
             }
         }
 
+
```
jennyf19: nit: remove extra lines?

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -36,22 +27,65 @@ protected Authority(AuthorityInfo authorityInfo)
...
         {
-            switch (requestAuthorityInfo.AuthorityType)
+            if (configAuthorityInfo == null)
+            {
+                throw new ArgumentNullException(nameof(configAuthorityInfo));
+            }
+
+            switch (configAuthorityInfo.AuthorityType)
             {
+                // ADFS and B2C are tenantless, no need to consider tenant
```
jennyf19: >B2C are tenantless [](start = 28, length = 18)

What you mean by tenantless here? B2C has a tenant.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
bgavrilMS: This is not a public comment, the class is internal. Request Authority is the same as "authority override"  - it is the authority that is specified in the AcquireToken* builder, e.g. 

` AcquireTokenSilent(...).WithAuthority(...)`

THis is different from the config authority, which is the one specified on the client object: 

`var pca = PublicClientApplicationBuilder.Create("clientid").WithAuthority(...)`

bgavrilMS: Happy to rephrase in a way that makes more sense, please advise.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
bgavrilMS: From MSAL perspective, I don't think it does. If you use .WithTentantId() or overloads of WithAuthority() that take a tenant, I don't think it does anything to a B2C authority. 

Tenantless means "common", "organizations" or "consumers". Not sure how else to call it ... "commonish" ? :)

jennyf19: idk...but definitely don't say "commonish"

jmprieur: not-tenanted authorties?

PR #1494, opened 2019-11-07
Update changelog for 4.7.0

trwalke:


changelog.txt
```diff
@@ -1,3 +1,8 @@
+4.7.0
+============
+New Features:
+Added Subject Name + Issuer authentication to the acquire token by authorization code and acquire token by refresh token flows with the WithSendX5C() api on the confidential client application.
+
```
jennyf19: nit: can you set it up like the other notes, w/a similar format, we have issue in bold and customer impact after that and then a link to the issue. thx.

changelog.txt
```diff
@@ -1,3 +1,8 @@
+4.7.0
+============
+New Features:
+**Added Subject Name + Issuer authentication to the acquire token by authorization code and acquire token by refresh token flows with the WithSendX5C() api on the confidential client application.** All confidential client authentication flows will not have acces to this feature. [Issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1490)
```
jennyf19: "will now have access" ? or not have access? also add another "s" to access :) thx.  looks cool

trwalke: should be "will now have access"

jennyf19: okay..because it says "will not"

Issue #1471, opened 2019-10-24
[Feature Request] Run all the tests with MSA as well as AAD (when suitable)


jmprieur:
**Is your feature request related to a problem? Please describe.**
We don't run MSA tests enough. Find out if we can do something in the infrastructure to run tests with both kind of accounts (when suitable = no IWA, UP, Daemon)

jmprieur:
PR is https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1511

jennyf19:
in 4.8.1 release

Issue #1459, opened 2019-10-18
[Fundamentals] Xamarin ui tests are failing when run on android 9 and 10


trwalke:
**Which Version of MSAL are you using ?**
latest

**Platform**
xamarin android, xamarin iOS

**Expected behavior**
The Xamarin.UiTes framework should be able to perform our interactive tests in appcenter on android 9 and 10

**Actual behavior**
The current implementation Xamarin.Uitest framework fails to properly navigate around the webviews during the app center test on android 9 and 10

**Possible Solution**
Upgrade the xamarin.Uitest package to version 3

PR #1409, opened 2019-10-02
Test automation - device code grant for MSA user

bgavrilMS:

PR #1496, opened 2019-11-09
Jennyf/msa tests

jennyf19:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1459
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1471

Android app center tests...but these might fail.
image

Actually, this whole thing will probably fail.

### DIFF
--- a/build/pipeline-uiautomation.yaml
+++ b/build/pipeline-uiautomation.yaml
@@ -121,7 +121,7 @@ stages:
       inputs:
         restoreSolution: LibsMacOS.sln
 
-    - task: XamariniOS@2
+    - task: XamariniOS@1
       displayName: 'Build Xamarin.iOS solution LibsMacOS.sln'
       inputs:
         solutionFile: LibsMacOS.sln
--- a/build/pipeline-uiautomation.yaml
+++ b/build/pipeline-uiautomation.yaml
@@ -121,7 +121,7 @@ stages:
       inputs:
         restoreSolution: LibsMacOS.sln
 
-    - task: XamariniOS@1
+    - task: XamariniOS@2
       displayName: 'Build Xamarin.iOS solution LibsMacOS.sln'
       inputs:
         solutionFile: LibsMacOS.sln
--- a/tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
+++ b/tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
@@ -62,8 +62,12 @@ namespace Microsoft.Identity.Test.UIAutomation
                 AcquireTokenSilentTest,
 
                 AcquireTokenADFSV3InteractiveFederatedTest,
+                AcquireTokenADFSV3InteractiveManagedTest,
+                AcquireTokenMsaUserTest,
                 AcquireTokenADFSV4InteractiveFederatedTest,
+                AcquireTokenADFSV4InteractiveManagedTest,
                 AcquireTokenADFSV2019InteractiveFederatedTest,
+                AcquireTokenADFSV2019InteractiveManagedTest,
 
                 B2CLocalAccountAcquireTokenTest,
                 //B2CFacebookMicrosoftLoginAcquireTokenTest,
@@ -125,6 +129,21 @@ namespace Microsoft.Identity.Test.UIAutomation
                 _xamarinController,
                 LabUserHelper.GetDefaultUserAsync().GetAwaiter().GetResult(),
                 CoreUiTestConstants.AcquireTokenSilent);
+        }
+
+        /// <summary>
+        /// Runs through the standard acquire token silent flow
+        /// for an Msa user
+        /// </summary>
+        [Test]
+        public void AcquireTokenMsaUserTest()
+        {
+            TestCommon.ResetInternalStaticCaches();
+
+            _mobileTestHelper.AcquireTokenTestHelper(
+                _xamarinController,
+                LabUserHelper.GetMsaUserAsync().GetAwaiter().GetResult(),
+                CoreUiTestConstants.AcquireTokenInteractive);
         }
 
         /// <summary>
@@ -248,6 +267,20 @@ namespace Microsoft.Identity.Test.UIAutomation
         }
 
         /// <summary>
+        /// Runs through the standard acquire token ADFSV4 Managed flow
+        /// </summary
+        [Test]
+        public void AcquireTokenADFSV4InteractiveManagedTest()
+        {
+            TestCommon.ResetInternalStaticCaches();
+
+            _mobileTestHelper.AcquireTokenTestHelper(
+                _xamarinController,
+                LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV4).GetAwaiter().GetResult(),
+                CoreUiTestConstants.ADFSv4NonFederated);
+        }
+
+        /// <summary>
         /// Runs through the standard acquire token ADFSV2019 Federated flow
         /// </summary
         [Test]
@@ -262,6 +295,20 @@ namespace Microsoft.Identity.Test.UIAutomation
         }
 
         /// <summary>
+        /// Runs through the standard acquire token ADFSV2019 Managed flow
+        /// </summary
+        [Test]
+        public void AcquireTokenADFSV2019InteractiveManagedTest()
+        {
+            TestCommon.ResetInternalStaticCaches();
+
+            _mobileTestHelper.AcquireTokenTestHelper(
+                _xamarinController,
+                LabUserHelper.GetAdfsUserAsync(FederationProvider.ADFSv2019).GetAwaiter().GetResult(),
+                CoreUiTestConstants.ADFSv2019NonFederated);
+        }
+
+        /// <summary>
         /// Runs through the standard acquire token ADFSV3 Federated flow
         /// </summary>
         [Test]
@@ -273,6 +320,20 @@ namespace Microsoft.Identity.Test.UIAutomation
                 _xamarinController,
                 LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV3).GetAwaiter().GetResult(),
                 CoreUiTestConstants.ADFSv3Federated);
+        }
+
+        /// <summary>
+        /// Runs through the standard acquire token ADFSV3 Managed flow
+        /// </summary>
+        [Test]
+        public void AcquireTokenADFSV3InteractiveManagedTest()
+        {
+            TestCommon.ResetInternalStaticCaches();
+
+            _mobileTestHelper.AcquireTokenTestHelper(
+                _xamarinController,
+                LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV3).GetAwaiter().GetResult(),
+                CoreUiTestConstants.ADFSv3NonFederated);
         }
 
         private static void LogMessage(string message, StringBuilder stringBuilderMessage)
--- a/tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
+++ b/tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
@@ -62,12 +62,9 @@ namespace Microsoft.Identity.Test.UIAutomation
                 AcquireTokenSilentTest,
 
                 AcquireTokenADFSV3InteractiveFederatedTest,
-                AcquireTokenADFSV3InteractiveManagedTest,
                 AcquireTokenMsaUserTest,
                 AcquireTokenADFSV4InteractiveFederatedTest,
-                AcquireTokenADFSV4InteractiveManagedTest,
                 AcquireTokenADFSV2019InteractiveFederatedTest,
-                AcquireTokenADFSV2019InteractiveManagedTest,
 
                 B2CLocalAccountAcquireTokenTest,
                 //B2CFacebookMicrosoftLoginAcquireTokenTest,
@@ -267,20 +264,6 @@ namespace Microsoft.Identity.Test.UIAutomation
         }
 
         /// <summary>
-        /// Runs through the standard acquire token ADFSV4 Managed flow
-        /// </summary
-        [Test]
-        public void AcquireTokenADFSV4InteractiveManagedTest()
-        {
-            TestCommon.ResetInternalStaticCaches();
-
-            _mobileTestHelper.AcquireTokenTestHelper(
-                _xamarinController,
-                LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV4).GetAwaiter().GetResult(),
-                CoreUiTestConstants.ADFSv4NonFederated);
-        }
-
-        /// <summary>
         /// Runs through the standard acquire token ADFSV2019 Federated flow
         /// </summary
         [Test]
@@ -295,20 +278,6 @@ namespace Microsoft.Identity.Test.UIAutomation
         }
 
         /// <summary>
-        /// Runs through the standard acquire token ADFSV2019 Managed flow
-        /// </summary
-        [Test]
-        public void AcquireTokenADFSV2019InteractiveManagedTest()
-        {
-            TestCommon.ResetInternalStaticCaches();
-
-            _mobileTestHelper.AcquireTokenTestHelper(
-                _xamarinController,
-                LabUserHelper.GetAdfsUserAsync(FederationProvider.ADFSv2019).GetAwaiter().GetResult(),
-                CoreUiTestConstants.ADFSv2019NonFederated);
-        }
-
-        /// <summary>
         /// Runs through the standard acquire token ADFSV3 Federated flow
         /// </summary>
         [Test]
@@ -320,20 +289,6 @@ namespace Microsoft.Identity.Test.UIAutomation
                 _xamarinController,
                 LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV3).GetAwaiter().GetResult(),
                 CoreUiTestConstants.ADFSv3Federated);
-        }
-
-        /// <summary>
-        /// Runs through the standard acquire token ADFSV3 Managed flow
-        /// </summary>
-        [Test]
-        public void AcquireTokenADFSV3InteractiveManagedTest()
-        {
-            TestCommon.ResetInternalStaticCaches();
-
-            _mobileTestHelper.AcquireTokenTestHelper(
-                _xamarinController,
-                LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV3).GetAwaiter().GetResult(),
-                CoreUiTestConstants.ADFSv3NonFederated);
         }
 
         private static void LogMessage(string message, StringBuilder stringBuilderMessage)
--- a/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
@@ -390,7 +390,7 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
 
             SecureString securePassword = new NetworkCredential("", user.GetOrFetchPassword()).SecurePassword;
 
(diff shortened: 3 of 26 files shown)


tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -62,8 +62,12 @@ public void RunAllTests()
                 AcquireTokenSilentTest,
 
                 AcquireTokenADFSV3InteractiveFederatedTest,
+                AcquireTokenADFSV3InteractiveManagedTest,
```
bgavrilMS: There is no such thing as adfsv3 managed.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -62,8 +62,12 @@ public void RunAllTests()
...
+                AcquireTokenMsaUserTest,
                 AcquireTokenADFSV4InteractiveFederatedTest,
+                AcquireTokenADFSV4InteractiveManagedTest,
```
bgavrilMS: there is no such thing as adfsv4 managed

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -62,8 +62,12 @@ public void RunAllTests()
...
                 AcquireTokenADFSV2019InteractiveFederatedTest,
+                AcquireTokenADFSV2019InteractiveManagedTest,
```
bgavrilMS: adfs2019 supports 2 modes: federated (i.e. through AAD) or direct. There is no such thing as "managed"

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -247,6 +266,20 @@ public void AcquireTokenADFSV4InteractiveFederatedTest()
                 CoreUiTestConstants.ADFSv4Federated);
         }
 
+        /// <summary>
+        /// Runs through the standard acquire token ADFSV4 Managed flow
+        /// </summary
+        [Test]
+        public void AcquireTokenADFSV4InteractiveManagedTest()
```
bgavrilMS: Please remove these ADFS tests.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -247,6 +266,20 @@ public void AcquireTokenADFSV4InteractiveFederatedTest()
...
+        {
+            TestCommon.ResetInternalStaticCaches();
+
+            _mobileTestHelper.AcquireTokenTestHelper(
+                _xamarinController,
+                LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV4).GetAwaiter().GetResult(),
```
bgavrilMS: This is not a managed user (there is no such thing), this is a federated user.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
jennyf19: I didn't think so but we had tests for them. This makes my day,! I will remove them!

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
bgavrilMS: adfsv3 and adfsv4 are supported via federation only, MSAL cannot "directly" talk to them.

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/ConfidentialClientAuthorizationTests.cs
```diff
@@ -75,6 +75,17 @@ public async Task SeleniumGetAuthCode_RedeemForAt_CommonAuthority_Async()
             await RunTestForUserAsync(labResponse, $"https://login.microsoftonline.com/{TenantId}").ConfigureAwait(false);
         }
 
+        [TestMethod]
+        // Regression test for: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/891
+        public async Task SeleniumGetAuthCode_RedeemForAt_CommonAuthority_MsaUser_Async()
```
bgavrilMS: Is it really worth running this rather expensive test with an MSA user? (genuine question) I think we want to make sure happy paths for all our flows work with MSA users, but need to exercise some causion around old regression tests.

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/FociTests.cs
```diff
@@ -49,9 +49,22 @@ public void TestInitialize()
...
         [TestMethod]
-        public async Task FociSingInSignOutAsync()
+        public async Task FociSignInSignOutAsync()
         {
             LabResponse labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
+            await FociSignInSignOutAsync(labResponse).ConfigureAwait(false);          
+        }
+
+        [TestMethod]
+        public async Task FociSignInSignOutMsaUserAsync()
```
bgavrilMS: Hmm, does this test pass? Foci with client_ids that have MSA passthrough enabled....

tests/Microsoft.Identity.Test.iOS.UIAutomation/iOSTests.cs
```diff
@@ -60,15 +60,19 @@ public void RunAllTests()
                 AcquireTokenInteractiveAndSilentTest,
 
                 AcquireTokenADFSV3InteractiveFederatedTest,
+                AcquireTokenADFSV3InteractiveManagedTest,
```
bgavrilMS: Same observation aroudn ADFS tests

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -54,6 +54,15 @@ public async Task Interactive_AADAsync()
             await RunTestForUserAsync(labResponse).ConfigureAwait(false);
         }
 
+        [TestMethod]
+        public async Task Interactive_MsaUser_Async()
```
bgavrilMS: We should also have an ROPC test with MSA, where we should assert that it fails gracefully.

bgavrilMS:
I'm a bit surprised that you are not getting into trouble with the consent screen around MSA. I had trouble "clicking" on the consent button, because it is not visible on screen. Have a look at my PR for details https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1409/files

bgavrilMS:
A few other tests that need to be covered: 

[OBO ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f7bba97c12301287c2a1994910ffa8711ddb5409/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs#L22)

[ROPC](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f7bba97c12301287c2a1994910ffa8711ddb5409/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/UsernamePasswordIntegrationTests.cs#L21)

[refresh_token](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f7bba97c12301287c2a1994910ffa8711ddb5409/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/SilentAuthTests.cs#L17) Note that MSA does not work with ROPC, so you'll need to get a token with another method, e.g. device code flow or interactive, then perform a silent call with force refresh

jennyf19:
amazing, right? lol.

---
In reply to: [552077521](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1496#issuecomment-552077521) [](ancestors = 552077521)

jennyf19:
in progress.

---
In reply to: [552077921](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1496#issuecomment-552077921) [](ancestors = 552077921)

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/ConfidentialClientAuthorizationTests.cs
jennyf19: we can discuss.

---
In reply to: [344433906](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1496#discussion_r344433906) [](ancestors = 344433906)

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
henrik-me: what were those testing? perhaps the name of the test was wrong?

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
henrik-me: same comment as above

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -390,7 +390,7 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
 
             SecureString securePassword = new NetworkCredential("", user.GetOrFetchPassword()).SecurePassword;
 
-            var msalPublicClient = PublicClientApplicationBuilder.Create(publicClientID).WithAuthority(TestConstants.AuthorityOrganizationsTenant).WithRedirectUri("urn:ietf:wg:oauth:2.0:oob").Build();
+            var msalPublicClient = PublicClientApplicationBuilder.Create(publicClientID).WithAuthority(TestConstants.AuthorityOrganizationsTenant).WithRedirectUri("https://login.microsoftonline.com/common/oauth2/nativeclient").Build();
```
henrik-me: I'm surprised there is not a const available to be used here given the number of places you have been updating consts

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
```diff
@@ -65,6 +70,60 @@ public async Task OBO_WithCache_MultipleUsers_Async()
             await RunOnBehalfOfTestAsync(aadUser2, true).ConfigureAwait(false);
         }
 
+        [TestMethod]
+        [Timeout(2 * 60 * 1000)] // 2 min timeout
```
henrik-me: >[Timeout(2 * 60 * 1000)] // 2 min timeout [](start = 8, length = 41)

any reason there has to be this timeout?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
```diff
@@ -65,6 +70,60 @@ public async Task OBO_WithCache_MultipleUsers_Async()
...
+            try
+            {
+                authResult = await cca
+                  .AcquireTokenSilent(s_scopes, msaUser.Upn)
+                  .ExecuteAsync()
+                  .ConfigureAwait(false);
+            }
+            catch (MsalUiRequiredException)
+            {
+                Assert.IsFalse(false, "ATS should have found a token, but it didn't");
```
henrik-me: suggestion: No abbreviations ?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
```diff
@@ -65,6 +70,60 @@ public async Task OBO_WithCache_MultipleUsers_Async()
             await RunOnBehalfOfTestAsync(aadUser2, true).ConfigureAwait(false);
         }
 
+        [TestMethod]
+        [Timeout(2 * 60 * 1000)] // 2 min timeout
+        public async Task OBO_WithCache_MSA_Async()
```
henrik-me: >OBO_WithCache_MSA_Async [](start = 26, length = 23)

I was hoping that this would basically be calling an existing test with the only difference being the actual user? Are refactoring's planned?

henrik-me:
are refactoring's planned to ensure that tests are written only once and only the user is replaced per test?

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/ConfidentialClientAuthorizationTests.cs
henrik-me: I would like to see the numbers telling us more precisely how expensive it really is before optimizing.

---
In reply to: [344960019](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1496#discussion_r344960019) [](ancestors = 344960019,344433906)

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/DeviceCodeFlowIntegrationTest.cs
```diff
@@ -88,5 +88,31 @@ public async Task DeviceCodeFlowAdfsTestAsync()
             Assert.IsNotNull(result);
             Assert.IsTrue(!string.IsNullOrEmpty(result.AccessToken));
         }
+
+        [TestMethod]
+        [Timeout(2 * 60 * 1000)] // 2 min timeout
```
henrik-me: >2 * 60 * 1000 [](start = 17, length = 13)

?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
jennyf19: the issue is we can't do ROPC w/msa accounts, so have to do another flow, like device code or something to get an RT. so it's not just a matter of running the same test w/an MSA account.

---
In reply to: [345534322](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1496#discussion_r345534322) [](ancestors = 345534322)

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
jennyf19: it's doing device code flow, so that's why the time out is there. i can think about a different way of doing this.

---
In reply to: [345533947](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1496#discussion_r345533947) [](ancestors = 345533947)

jennyf19:
I'm trying to do that when possible, but most of our tests do ROPC< which we can't do w/MSA.

---
In reply to: [553199348](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1496#issuecomment-553199348) [](ancestors = 553199348)

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
bgavrilMS: No, I think the lab API used to be confusing. You could query the lab to give you a user that was marked as "managed" and "Adfs" and the lab would return some user. We thought this is a valid permutation, but having discussed with the lab team, this is not. I believe the new API throws an exception if trying to get a "managed adfs" user.

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/ConfidentialClientAuthorizationTests.cs
bgavrilMS: An interactive login takes with Selenium takes about 5 seconds. Most of our tests are more complex, they perform the flow serveral times etc. Test time does not vary too much: 

image

Entire Integration test suite: ~3 min on .net 45, ~2 min on .net core
We don't seem to be running the .net47 and .net48 tests :( I'll fix this.

bgavrilMS: The particular test you are pointing at is a pretty complex one, it takes 30 seconds (our most expensive test)

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
bgavrilMS: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9b6ed7aa6fba9f7063e1e72a2a1a609f620f1332/src/client/Microsoft.Identity.Client/Core/Constants.cs#L16

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
bgavrilMS: There are several levels of timeout: 

1. Each Selenium operation has an implicit timeout (I believe 10 sec). So if clicking on a button takes more than 10s -> fail. 
2. Some Selenium operations have custom timeouts (e.g. wait for a button to be enabled for 20 seconds)
3. TcpListener has I believe a 5 min timeout, in case smth happens to the browser / redirect etc.

However, the test can hang outside Selenium and the tcp listener - bugs in the tcp listener / Selenioum syncronization, bugs in MSAL etc. This happens rarely, but when it happens I'd rather not have to wait 1h for the entire build to timeout.

MarkZuber:
> #1459
> #1471
> 
> Android app center tests...but these might fail.
> image
> 
> Actually, this whole thing will probably fail.

yup, it probably will.  :P

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
```diff
@@ -70,60 +70,6 @@ public async Task OBO_WithCache_MultipleUsers_Async()
             await RunOnBehalfOfTestAsync(aadUser2, true).ConfigureAwait(false);
         }
 
-        [TestMethod]
```
bgavrilMS: Can you ignore them instead? We should fix timeouts.

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/FociTests.cs
```diff
@@ -55,14 +55,6 @@ public async Task FociSignInSignOutAsync()
             await FociSignInSignOutAsync(labResponse).ConfigureAwait(false);          
         }
 
-        [TestMethod]
-        public async Task FociSignInSignOutMsaUserAsync()
```
bgavrilMS: This is a pretty important test to have deleted. ...

build/pipeline-uiautomation.yaml
```diff
@@ -121,7 +121,7 @@ stages:
       inputs:
         restoreSolution: LibsMacOS.sln
 
-    - task: XamariniOS@2
+    - task: XamariniOS@1
```
bgavrilMS: What does this do?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/OboTests.cs
jmprieur: [Ignore] and add a github issue to fix them later

build/pipeline-uiautomation.yaml
jennyf19: was trying to fix the yaml build issues. found it in a xamarin forum but it didn't work.

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/FociTests.cs
jennyf19: yeah, was trying to get some tests to pass, as all the ones w/device code flow are timing out and just slowing down this process.

jennyf19:
@henrik-me @bgavrilMS made a second PR w/out the xamarin changes.

PR #1500, opened 2019-11-13
Integration test fixes

bgavrilMS:


tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
```diff
@@ -87,55 +87,6 @@ public async Task FailedAuthorityValidationTestAsync()
             Assert.AreEqual("invalid_instance", exception.ErrorCode);
         }
 
-        [TestMethod]
-        public void FooTest()
```
jmprieur: why having removed these? or moved to the other folder?
They are not headless tests?

PR #1503, opened 2019-11-14
Fix cache tests

bgavrilMS:
The crux of the change is around using client_id and tenant_id as input params for the tests instead of constants.


tests/CacheCompat/CommonCache.Test.MsalV3/Program.cs
```diff
@@ -23,36 +23,37 @@ private class MsalV3CacheExecutor : AbstractCacheExecutor
...
-                    testInputData.StorageType,
-                    app.UserTokenCache,
-                    CommonCacheTestUtils.AdalV3CacheFilePath,
-                    CommonCacheTestUtils.MsalV2CacheFilePath,
-                    CommonCacheTestUtils.MsalV3CacheFilePath);
-
-                IEnumerable<IAccount> accounts = await app.GetAccountsAsync().ConfigureAwait(false);
+               
 
                 var results = new List<CacheExecutorAccountResult>();
```
trwalke: too many spaces above

tests/CacheCompat/CommonCache.Test.Common/LabUserData.cs
```diff
@@ -1,21 +1,25 @@
...
         {
-        }
-
-        public LabUserData(string upn, string password)
-        {
-            Upn = upn;
-            Password = password;
+            Upn = upn ?? throw new ArgumentNullException(nameof(upn));
+            Password = password ?? throw new ArgumentNullException(nameof(clientId));
+            ClientId = clientId ?? throw new ArgumentNullException(nameof(clientId));
```
trwalke: should this be nameof(password?)

tests/CacheCompat/CommonCache.Test.Common/LabUserData.cs
```diff
@@ -1,21 +1,25 @@
...
+            TenantId = tenantId ?? throw new ArgumentNullException(nameof(tenantId));
         }
 
         public string Upn { get; set; }
         public string Password { get; set; }
+
+        public string ClientId { get; }
```
trwalke: Extra space above?

tests/CacheCompat/CommonCache.Test.Common/LabUserData.cs
```diff
@@ -1,21 +1,25 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+
 namespace CommonCache.Test.Common
 {
     public class LabUserData
```
trwalke: is it not possible to use the lab response classes that already have this data? This class seems redundant unless I am missing something?

tests/CacheCompat/CommonCache.Test.MsalJava/src/main/java/TestInput.java
```diff
@@ -29,5 +23,11 @@
 
         @SerializedName("Password")
         String password;
+
+         @SerializedName("Authority")
```
trwalke: nit: extra space

tests/CacheCompat/CommonCache.Test.AdalV5/Program.cs
```diff
@@ -22,40 +22,37 @@ private class AdalV5CacheExecutor : AbstractCacheExecutor
...
-                var authenticationContext = new AuthenticationContext(app.Authority, tokenCache);
-
-                var items = authenticationContext.TokenCache.ReadItems().ToList();
-                Console.WriteLine("here come the cache items!: {0}", tokenCache.Count);
-                foreach (var item in items)
-                {
-                    Console.WriteLine("here's a cache item!");
-                    Console.WriteLine(item.DisplayableId);
-                }
+                //var items = authenticationContext.TokenCache.ReadItems().ToList();
```
trwalke: dead code?

tests/CacheCompat/CommonCache.Test.Common/LabUserData.cs
trwalke: line 13

tests/CacheCompat/CommonCache.Test.Common/LabUserData.cs
bgavrilMS: I originally thought the same, but this class is more of an adapter for the Lab objects. This gets serialized and re-serlialized to JSON and passed to python and to java. Also, I'd need to make changes to each cache test project to take a dependnecy on LabInfra dll, which is not trivial. 

So all in all - yes, I agree with you that this is a redundant object, but in practical terms it simplifies the design of the cache tests.

Issue #1506, opened 2019-11-14
[Bug] Interactive auth with Edge system browser sometimes hangs


bgavrilMS:
**Which Version of MSAL are you using ?**
MSAL 4.7

**Platform**
netcore or net classic

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Repro**

Configure interactive auth to use the system browser and also configure it to use Edge.

```csharp
  var options = new SystemWebViewOptions()
{
     OpenBrowserAsync = SystemWebViewOptions.OpenWithEdgeBrowserAsync
};

var cts = new CancellationTokenSource();
authTask = pca.AcquireTokenInteractive(s_scopes)
                        .WithSystemWebViewOptions(options)
                        .ExecuteAsync(cts.Token);
```

**Actual behavior**
Sometimes the browser hangs at an url similar to: 
`https://login.microsoftonline.com/common/reprocess?ctx=rQIIAW2Q`

Eventually an MsalClientException occurs, stating: 

`Could not extract the query from the authorization response - check Pii enabled logs for details`

This happens about 50% of the time.

**Possible Solution**
It looks like the TCP listener is invoked, but no data is available. Calling `ReadAsync` seems to block. After some time or if the browser is refreshed, the tcp listener reads 0 bytes.

bgavrilMS:
P1 as it is blocking customers (Storage Explorer in particular)

bgavrilMS:
CC @crmann1

jmprieur:
@bgavrilMS : we did not ship 4.8 yet? we might just want to mark it as fixed (and keep it open)

bgavrilMS:
I merged the fix and it looks like GitHub closes the associated issue...

bgavrilMS:
I think this is because GitHub inspects the PR comments and if it finds smth like "fixes bug xxx" then it automatically closes bug xxx. Smart, but not customizable ... :)

https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

jmprieur:
oh, ok. thanks for the explantion, @bgavrilMS !

jmprieur:
Released today in MSAL.NET 4.7.1

PR #1509, opened 2019-11-17
System Browser - use HttpListner instead of TcpListner

bgavrilMS:
Should fix https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1506

It looks like there are several TCP connections being created by the browser and they behave differently in various browsers (well, Edge being the odd-ball there). I found a solution to this by cycling through several tcp connections, but the overall code became rather complex. 

So I re-wrote the interceptor using HttpListner rather than TcpListner. We weren't able to use this before because it has been made available in .net core 2+, whereas at the time this feature came out we were still supporting .net core 1.x


src/client/Microsoft.Identity.Client/Platforms/Shared/DefaultOSBrowser/HttpListenerInterceptor.cs
```diff
@@ -0,0 +1,107 @@
...
+using System;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Platforms.Shared.Desktop.OsBrowser;
+
+namespace Microsoft.Identity.Client.Platforms.Shared.DefaultOSBrowser
+{
+    internal class HttpListnerInterceptor : IUriInterceptor
```
bgavrilMS: The overall logic is much simpler since there is no more parsing of messages

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -129,7 +129,7 @@ public override ITokenCacheAccessor CreateTokenCacheAccessor()
             return new InMemoryTokenCacheAccessor();
         }
 
-        protected override IWebUIFactory CreateWebUiFactory() => new NetStandardWebUIFactory();
```
bgavrilMS: We never actually used this factory in .net standard

src/client/Microsoft.Identity.Client/Platforms/Shared/DefaultOSBrowser/HttpListenerInterceptor.cs
jmprieur: I think there is typo?
HttpListnerInterceptor => HttpList**e**nerInterceptor  ?

jennyf19: @jmprieur ---wow...good catch.

PR #1513, opened 2019-11-21
remove use of wrong default redirect uri in tests urn:ietf:wg:oauth:2…

jennyf19:
….0:oob


src/client/Microsoft.Identity.Client/Core/Constants.cs
```diff
@@ -13,7 +13,7 @@ internal static class Constants
         public const int CodeVerifierByteSize = 32;
 
         public const string UapWEBRedirectUri = "https://sso"; // only ADAL supports WEB
-        public const string DefaultRedirectUri = "urn:ietf:wg:oauth:2.0:oob";
```
henrik-me: >DefaultRedirectUri [](start = 28, length = 18)

this is technically a breaking change - perhaps we should not make this change unless telemetry data shows this redirect uri is no longer used? This means when upgrading the library they need to ensure the redirect uri is added.

PR #1523, opened 2019-11-28
Update pipeline-uiautomation.yaml

trwalke:
Disabling iOS appcenter tests for now. A new signing cert for the Xamarin build task needs to be created.


jennyf19:
@trwalke I think we should just leave them failing until this is fixed and we also need an issue opened (unless you already did it).

PR #1525, opened 2019-11-29
POP implementation 

bgavrilMS:
What this contains: 

- Full implementation of POP authentication scheme
- Key management for net45
- Public APIs, exposed in net45 only (final API surface still TBD)
- Lots of tests

What it doesn't contain: 
- Latest token cache storage strategy for pop tokens

How to review this: 
- set aside at least 1-2h
- read the [Evo spec](https://msazure.visualstudio.com/One/_git/ESTS-Docs?path=%2FProtocols%2FAuthentication%2FOAuth2.0%2FProofofPossession%2FNon-Implicit%20Access%20Token%20Proof%20of%20Possession.md&version=GBmaster)
- Review the E2E in commit https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/c3f747d02ffd5eaa97026236884df90534e0b6c9
- Run the integration test, and inspect the Evo traffic and decode tokens
- Review commit by commit.


src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -377,9 +377,9 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
                 return item;
             }
 
-            if (string.Equals(item.KeyId, requestKid))
+            if (string.Equals(item.KeyId, requestKid, StringComparison.OrdinalIgnoreCase))
             {
-                authenticationRequest.RequestContext.Logger.Verbose("PoP token found");
+                authenticationRequest.RequestContext.Logger.Verbose("Keyed token found");
```
jennyf19: >Keyed [](start = 69, length = 5)

nit: how about "Token found, bound to a key"

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
```diff
@@ -0,0 +1,132 @@
...
+                    new JProperty(PoPClaimTypes.JWK, popAssertion))));
+
+            return CreateJWS(payload.ToString(Json.Formatting.None), header.ToString(Json.Formatting.None));
+        }
+
+        private static string CreateNonce()
+        {
+            // Guid with no hyphens
+#if NETSTANDARD || WINDOWS_APP
+            return Guid.NewGuid().ToString("N");
```
jennyf19: >Guid.NewGuid().ToString("N") [](start = 19, length = 28)

how is this different from doing Guid.NewGuid()?

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
```diff
@@ -0,0 +1,132 @@
...
+                new JProperty(PoPClaimTypes.Host, _requestUri.Host),
+                new JProperty(PoPClaimTypes.Path, _requestUri.AbsolutePath),
+                new JProperty(PoPClaimTypes.Nonce, CreateNonce()),
+                new JProperty(PoPClaimTypes.Cnf, new JObject(
+                    new JProperty(PoPClaimTypes.JWK, popAssertion))));
+
+            return CreateJWS(payload.ToString(Json.Formatting.None), header.ToString(Json.Formatting.None));
+        }
+
+        private static string CreateNonce()
```
jennyf19: do we need to verify that the nonce is valid? the spec mentions 1-5 min lifetime.

tests/devapps/NetFxConsoleTestApp/Program.cs
```diff
@@ -309,24 +345,115 @@ 1. IWA
...
+                    .WithPoPAuthenticationScheme(new Uri(PoPUri), s_popMethod);
+            }
+
+            return builder as T;
+        }
+
+        private static async Task FetchTokenAndCallPoPApiAsync(Task<AuthenticationResult> authTask)
+        {
+            var result = await authTask.ConfigureAwait(false);
+            Console.BackgroundColor = ConsoleColor.DarkGreen;
```
jennyf19: nit: move the display stuff to a separate method to resuse

tests/devapps/NetFxConsoleTestApp/Program.cs
```diff
@@ -309,24 +345,115 @@ 1. IWA
...
+        {
+            var httpClient = new HttpClient();
+            HttpResponseMessage response;
+            var request = new HttpRequestMessage(HttpMethod.Post, PoPValidatorEndpoint);
+
+            request.Headers.Add("Authorization", authHeader);
+            request.Headers.Add("Secret", s_secretForPoPValidationRequest);
+            request.Headers.Add("Authority", "https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/");
+            request.Headers.Add("ClientId", s_clientIdForPublicApp);
+
```
jennyf19: why is all this needed? is this something developers would need to write as well in order to use PoP?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/PoPTests.cs
```diff
@@ -0,0 +1,185 @@
...
+        [TestMethod]
+        public async Task POPAsync()
+        {
+            var labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
+            await RunHappyPathTestAsync(labResponse).ConfigureAwait(false);
+        }
+
+        private async Task RunHappyPathTestAsync(LabResponse labResponse)
+        {
+            var user = labResponse.User;
```
jennyf19: nit: is there a reason to make this a separate variable? instead of just labResponse.User.Upn , etc...?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/PoPTests.cs
```diff
@@ -0,0 +1,185 @@
...
+            HttpResponseMessage response;
+            var request = new HttpRequestMessage(HttpMethod.Post, PoPValidatorEndpoint);
+
+            request.Headers.Add("Authorization", popAuthHeader);
+            request.Headers.Add("Secret", _popValidationEndpointSecret);
+            request.Headers.Add("Authority", "https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/");
+            request.Headers.Add("ClientId", clientId);
+
+            // the URI the POP token is bound to
+            request.Headers.Add("ShrUri", uri.ToString());
```
jennyf19: >Shr [](start = 33, length = 3)

what's shr?

tests/Microsoft.Identity.Test.Unit.net45/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        private static string AssertSimpleClaim(JwtSecurityToken jwt, string expectedKey, string optionalExpectedValue = null)
+        {
+            string value = jwt.Claims.Single(c => c.Type.Equals(expectedKey, StringComparison.InvariantCultureIgnoreCase)).Value;
+            if (optionalExpectedValue != null)
+            {
+                Assert.AreEqual(optionalExpectedValue, value);
+            }
+            return value;
+        }
+    }
```
jennyf19: do you have any tests for catching the ests errors for failed requests? spec says ESTS will respond with Bad Request if the req_cnf parameter is provided when token_type=pop and the input value is one of the following:
Not URL encoded
Not containing a kid field
Containing more elements than kid

is it possible to have tests for that? do we do anything special in msal for these cases? also, this may be more middleware related...still wrapping my head around this. :)

jennyf19:
>                     string.Format(

do we have tests for this? finding a token w/ the wrong key?

---
Refers to: src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs:387 in c3f747d. [](commit_id = c3f747d02ffd5eaa97026236884df90534e0b6c9, deletion_comment = False)

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
bgavrilMS: I believe this is the responsability of the verifier (i.e. Wilson). MSAL should simply generate a random number so that the same request cannot be played twice.

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
bgavrilMS: Guid.NewGuid() returns an object of type `Guid` so we need to tranform it into a string. .NET can do that for us because each and every object implements a method called `ToString`, although note that in most cases you get a string represenation like `this is an array` instead of `1 2 3`. In case of GUIDs, .net folks have implemented ToString for us and it returns a string in the format `xxxx-xxxx-xxxxxxxxx-xxxx`. But I do not want hyphens on there (this is what I discussed with George), and the ToString("N") takes care of that.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/PoPTests.cs
bgavrilMS: No, I guess not, it's just that typing `labResposne.User.property` everytime is kind of long...

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/PoPTests.cs
bgavrilMS: George has setup a small web api endpoint where we pass an HttpRequest with a POP token plus some headers and it checks that the POP token is according to those headers. So in this case, it will check that

`popToken["bound_uri"] == headers["ShrUri"]`

bgavrilMS: (otherwise I also don't know what `shr` stands for :))

tests/Microsoft.Identity.Test.Unit.net45/pop/PopAuthenticationSchemeTests.cs
bgavrilMS: This is a very good question. In this PR, we do not allow developers to specify their own key, because on Windows we can do that for them. But it is a natural extensibility point which we will undertake sooner or later. 

Because we do key management, I don't expect ESTS to ever return these errors, because MSAL obeys the rules above. However, implementers other than MSAL (e.g. if you use call the Web APIs directly), then you might hit those if you do not package the key correctly.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -21,6 +26,47 @@ internal AbstractPublicClientAcquireTokenParameterBuilder(IPublicClientApplicati
...
+        /// <param name="httpRequestMessage">An HTTP request to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request.</param>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</item>
+        /// <item> Add the PoP token in an Authorization header, just like a bearer token. See <seealso cref="AuthenticationResult.CreateAuthorizationHeader"/> for details.</item>
+        /// <item> The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters). </item>
+        /// <item> MSAL creates, reads and stores a key securely on behalf of the Windows user. </item>
+        /// </list>
+        /// </remarks>
+        public T WithPoPAuthenticationScheme(HttpRequestMessage httpRequestMessage)
```
bgavrilMS: @jmprieur @henrik-me - Please review the public API. There are 2 possibilities for the param `HttpRequestMessage` (which embeds uri, method etc.) and `uri` + `method` explicitly. 

I like `HttpRequestMessge` more because in the future it is possible that we will bind not only the Uri and the Method, but also the forms params and headers, which users could "turn on", similar to how Wilson will validate the PoP tokens.

Also, we may consider adding the `Authentication: PoP ey...` header directly to the HttpRequestMessage.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/PoPTests.cs
jennyf19: lol

---
In reply to: [352708502](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1525#discussion_r352708502) [](ancestors = 352708502)

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/PoPTests.cs
jennyf19: it's SignedHttpRequest...I asked George :)

---
In reply to: [352711137](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1525#discussion_r352711137) [](ancestors = 352711137)

jennyf19: I see...thanks

---
In reply to: [352710544](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1525#discussion_r352710544) [](ancestors = 352710544)

tests/Microsoft.Identity.Test.Unit.net45/pop/PopAuthenticationSchemeTests.cs
jennyf19: makes sense. thanks.

---
In reply to: [352715068](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1525#discussion_r352715068) [](ancestors = 352715068)

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
jennyf19: got it. thanks.

---
In reply to: [352705636](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1525#discussion_r352705636) [](ancestors = 352705636)

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
jennyf19: cool. thanks for the explanation.

---
In reply to: [352707770](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1525#discussion_r352707770) [](ancestors = 352707770)

tests/devapps/NetFxConsoleTestApp/Program.cs
bgavrilMS: No, this is because the SHR (George's test validator) is flexible and allows us to test any kind of PoP token. 

Users would not need to set any of this up./

jennyf19: i see... very cool

---
In reply to: [352758457](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1525#discussion_r352758457) [](ancestors = 352758457)

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
jmprieur: I agree @bgavrilMS that we should use the abstractions available in .NET, and this has potential to evolve. We'll just need to document what members are used today. 

Adding the authentication to the HttpRequestMessage (in a re-entrant way), would also be good. We'd also probably be able to provide an end to end with the HttpRequestMessage (calling the resource, getting a 403 and an www.Authenticate header, parsing it, and finding that it needs Pop?

src/client/Microsoft.Identity.Client/PoP/PoPAuthenticationScheme.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                return hash.ComputeHash(Encoding.UTF8.GetBytes(cannonicalJwk));
+            }
+        }
+
+        /// <summary>
+        /// Creates a JWS (json web signature) as per: https://tools.ietf.org/html/rfc7515
+        /// Format: header.payload.signed_payload
+        /// </summary>
+        private string CreateJWS(string payload, string header)
```
jennyf19: this is what sal verifies?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AuthenticationSchemeTests.cs
```diff
@@ -32,6 +33,7 @@ public async Task Interactive_WithCustomAuthScheme_ThenSilent_Async()
             // Arrange
             var authScheme = Substitute.For<IAuthenticationScheme>();
             authScheme.AuthorizationHeaderPrefix.Returns("BearToken");
```
jennyf19: um....lol. "BearToken"? I know you didn't add this, but it's pretty funny

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AuthenticationSchemeTests.cs
```diff
@@ -89,6 +91,41 @@ public async Task Interactive_WithCustomAuthScheme_ThenSilent_Async()
             }
         }
 
+        [TestMethod]
+        public async Task WrongTokenType_Async()
+        {
+            // Arrange
+            var authScheme = Substitute.For<IAuthenticationScheme>();
+            authScheme.AuthorizationHeaderPrefix.Returns("BearToken");
```
jennyf19: "BearerToken"?

PR #1533, opened 2019-12-04
[DO NOT MERGE] Enabling android appcenter builds

trwalke:


tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -107,10 +108,19 @@ public void AcquireTokenTest()
...
+            try
+            {
+                _mobileTestHelper.AcquireTokenTestHelper(
+                    _xamarinController,
+                    LabUserHelper.GetDefaultUserAsync().GetAwaiter().GetResult(),
+                    CoreUiTestConstants.AcquireTokenInteractive);
+            }
+            catch (FileLoadException ex)
+            {
+#pragma warning disable CA2201 // Do not raise reserved exception types
```
bgavrilMS: Please try to avoid these directives that disable best practices. You can read about them on msdn and they usually give you a pretty good idea on how to proceed. In this case, the solution is to throw an `InvalidOperationException` or some other more focused exception.

tests/Microsoft.Identity.Test.Android.UIAutomation/androidTests.cs
```diff
@@ -107,10 +108,19 @@ public void AcquireTokenTest()
...
+                throw new Exception(ex.FusionLog);
```
bgavrilMS: By doing this you lose the original exception details, which may be important. All .net exceptions have a concept of wrapping other exceptionsd, which you can use here. I would rewrite this as: 

```csharp
throw new FileLoadException($"A FileLoad exception happened. This usually occurs when a dependency DLL is not found (e.g. Newtonsoft.Json.dll). See inner exception for details. Fusion Log: {ex.FusionLog}. ", ex);

```

trwalke: There changes are only for testing appcenter stuff, will be removed later. please ignore

bgavrilMS: There's not reason to not leave this code in permanently.

PR #1534, opened 2019-12-05
[WIP] fix possible null ref in non-broker related calls when source applica…

jennyf19:
…tion is null

first reported in [ADAL](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/issues/1673)
now w/ios 13, sourceApplication can be null, and if null, we are throwing a null ref later in the code for non broker responses, as the SemaphoreSlim was never created during the creation of the ios broker object.

still need to test this scenario w/msal

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using Foundation;
 using Microsoft.Identity.Client.Platforms.iOS;
 
@@ -57,8 +58,21 @@ namespace Microsoft.Identity.Client
         /// <param name="url"></param>
         public static void SetBrokerContinuationEventArgs(NSUrl url)
         {
+            // url should always be present
+            // if coming from broker, will be the broker response
+            // check to prevent null ref later
+            if (url == null)
+            {
+                return;
+            }
+
             Debug.WriteLine("SetBrokercontinuationEventArgs Called with Url {0}", url);
-            iOSBroker.SetBrokerResponse(url);
+
+            string urlString = string.Format(CultureInfo.CurrentCulture, url.ToString());
+            if (urlString.Contains(iOSBrokerConstants.IdentifyiOSBrokerFromResponseUrl))
+            {
+                iOSBroker.SetBrokerResponse(url);
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
```diff
@@ -57,8 +58,21 @@ public static bool IsBrokerResponse(string sourceApplication)
         /// <param name="url"></param>
         public static void SetBrokerContinuationEventArgs(NSUrl url)
         {
+            // url should always be present
```
bgavrilMS: I still don't understand when this url can ever be null. If it is an error, shouldn't we expose it to the user?

src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
```diff
@@ -57,8 +58,21 @@ public static bool IsBrokerResponse(string sourceApplication)
...
+            // check to prevent null ref later
+            if (url == null)
+            {
+                return;
+            }
+
             Debug.WriteLine("SetBrokercontinuationEventArgs Called with Url {0}", url);
-            iOSBroker.SetBrokerResponse(url);
+
+            string urlString = string.Format(CultureInfo.CurrentCulture, url.ToString());
```
bgavrilMS: Don't use CurrentCulture for URLs. Use an invariant culture.

src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
```diff
@@ -57,8 +58,21 @@ public static bool IsBrokerResponse(string sourceApplication)
...
+            if (urlString.Contains(iOSBrokerConstants.IdentifyiOSBrokerFromResponseUrl))
+            {
+                iOSBroker.SetBrokerResponse(url);
```
bgavrilMS: I think this may lead to a deadlock. I don't understand under which circumstance this condition would not be hit but  but it looks to me that you start a broker flow, which acquires a semaphore, then when the continuation is hit, you may or may not release the semaphore. If you don't release the semaphore the app is blocked.

src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
```diff
@@ -57,8 +58,21 @@ public static bool IsBrokerResponse(string sourceApplication)
         /// <param name="url"></param>
         public static void SetBrokerContinuationEventArgs(NSUrl url)
         {
+            // url should always be present
+            // if coming from broker, will be the broker response
+            // check to prevent null ref later
+            if (url == null)
```
bgavrilMS: Can we have a chat about this please?

src/client/Microsoft.Identity.Client/Platforms/iOS/AuthenticationContinuationHelper.cs
jennyf19: discussed off-line

---
In reply to: [354580232](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1534#discussion_r354580232) [](ancestors = 354580232)

PR #1557, opened 2020-01-07
fix sso on uap and update uwp test app

jennyf19:
Using the test app, select the null redirect uri, which will use https://sso and then WEB auth to handle "sso". I signed in w/corp account and cloud account, and both show as previously signed in accounts using the UWP stand-alone test app. Similar to how we handle this in ADAL.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
@@ -191,7 +191,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 {
                     // Hack: STS does not yet send back the suberror on these platforms because they are not in an allowed list,
                     // so the best thing we can do is to consider all errors as client_mismatch.
-#if NETSTANDARD || UAP || MAC
+#if NETSTANDARD || WINDOWS_APP || MAC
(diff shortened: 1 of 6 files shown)


bgavrilMS:
Good stuff!

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -191,7 +191,7 @@ private async Task<MsalTokenResponse> TryGetTokenUsingFociAsync(CancellationToke
                 {
                     // Hack: STS does not yet send back the suberror on these platforms because they are not in an allowed list,
                     // so the best thing we can do is to consider all errors as client_mismatch.
-#if NETSTANDARD || UAP || MAC
+#if NETSTANDARD || WINDOWS_APP || MAC
                     ex?.GetType();  // avoid the "variable 'ex' is declared but never used" in this code path.
```
henrik-me: >ex?.GetType(); [](start = 20, length = 14)

should probably have been fixed by not specifying the variable in the catch paragraph instead of wasted cpu time?

henrik-me: > ex?.GetType();  [](start = 19, length = 16)

could have put the #if's around the catch instead and have this path not use the parameter, or specify an ignore flag instead. waste of cpu

PR #1563, opened 2020-01-08
Update DLL signing metadata - name and info

bgavrilMS:


bgavrilMS:
@jmprieur - the old values are the "default" ones. You'd need an advanced tool to see these values, like NuGet Package Explorer 
image

PR #1564, opened 2020-01-08
updating telemetry schema from v1 to v2

jennyf19:
Updating the telemetry schema from v1 to v2 - here is a link to the [internal pr doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/995?isPreview=true&path=%2FTelemetry%2FMSALServerSideTelemetry.md&_a=files&fullScreen=true)

Main change is in the previous requests, we now need to keep all the previous requests (api_id and correlation_id plus any corresponding error codes) up to 4kb for telemetry.

@bgavrilMS and I will discuss this tomorrow.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -128,6 +128,10 @@ namespace Microsoft.Identity.Client.Internal.Requests
                         apiEvent.TenantId = authenticationResult.TenantId;
                         apiEvent.AccountId = authenticationResult.UniqueId;
                         apiEvent.WasSuccessful = true;
+                        if (AuthenticationRequestParameters.ApiId == ApiEvent.ApiIds.AcquireTokenSilent)
+                        {
+                            AuthenticationRequestParameters.RequestContext.ServiceBundle.TelemetryManager.SuccessfulSilentCallCount ++;
+                        }
                         return authenticationResult;
                     }
                     catch (MsalException ex)
--- a/src/client/Microsoft.Identity.Client/TelemetryCore/ITelemetryManager.cs
+++ b/src/client/Microsoft.Identity.Client/TelemetryCore/ITelemetryManager.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Identity.Client.TelemetryCore
     internal interface ITelemetryManager
     {
         TelemetryCallback Callback { get; }
-
+        int SuccessfulSilentCallCount { get; set; }
         TelemetryHelper CreateTelemetryHelper(EventBase eventBase);
 
         void StartEvent(EventBase eventToStart);
--- a/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
+++ b/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using Microsoft.Identity.Client.Core;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Constants;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 
@@ -18,35 +20,66 @@ namespace Microsoft.Identity.Client.TelemetryCore.Internal
                 evt.TryGetValue(MsalTelemetryBlobEventNames.ApiErrorCodeConstStrKey, out string errorCode);
                 evt.TryGetValue(MsalTelemetryBlobEventNames.ForceRefreshId, out string forceRefresh);
 
-                ApiId = apiId ?? string.Empty;
-                CorrelationId = correlationId ?? string.Empty;
-                LastErrorCode = errorCode ?? string.Empty;
+                ApiId.Add(apiId ?? string.Empty);
+                CorrelationId.Add(correlationId ?? string.Empty);
+                LastErrorCode.Add(errorCode ?? string.Empty);
                 ForceRefresh = forceRefresh ?? string.Empty;
             }
         }
 
-        public string LastErrorCode { get; set; } = string.Empty;
-        public string ApiId { get; set; } = string.Empty;
-        public string CorrelationId { get; set; } = string.Empty;
-        public string ForceRefresh { get; set; } = string.Empty;
-
-        public string GetCsvAsPrevious()
+        public /*For testing*/ HttpTelemetryContent(bool shouldClearCache)
         {
-            if (string.IsNullOrWhiteSpace(ApiId))
+            if (shouldClearCache)
+            {
+                ApiId?.Clear();
+                CorrelationId?.Clear();
+                LastErrorCode?.Clear();
+            }
+        }
+
+        public string ForceRefresh { get; set; } = string.Empty;
+        public static List<string> ApiId { get; set; } = new List<string>();
+        public static List<string> CorrelationId { get; set; } = new List<string>();
+        public static List<string> LastErrorCode { get; set; } = new List<string>();
+
+        public string GetCsvAsPrevious(int successfulSilentCallCount)
+        {
+            if (ApiId == null)
             {
                 return string.Empty;
             }
 
             // csv expected format:
-            // 1|api_id,correlation_id,last_error_code|platform_config
-            string[] myValues = new string[] {
-                ApiId,
-                CorrelationId,
-                LastErrorCode};
+            // 2|silent_successful_count|failed_requests|errors|platform_fields
+            // failed_request can be further expanded to include:
+            // api_id_1,correlation_id_1,api_id_2,correlation_id_2|error_1,error_2
+            string apiIdCorIdData = CreateApiIdAndCorrelationIdContent();
+            string data = string.Format(CultureInfo.InvariantCulture,
+                $"{TelemetryConstants.HttpTelemetrySchemaVersion2}{TelemetryConstants.HttpTelemetryPipe}{successfulSilentCallCount}{TelemetryConstants.HttpTelemetryPipe}" +
+                $"{CreateFailedRequestsContent(apiIdCorIdData)}" +
+                $"{TelemetryConstants.HttpTelemetryPipe}");
 
-            string csvString = string.Join(",", myValues);
-            csvString = $"{TelemetryConstants.HttpTelemetrySchemaVersion1}{TelemetryConstants.HttpTelemetryPipe}{csvString}{TelemetryConstants.HttpTelemetryPipe}";
-            return csvString;
+            if (data.Length <= 3800)
+            {
+                return data;
+            }
+            else
+            {
+                return data.Substring(0, length: 3800);
+            }                
+        }
+
+        private string CreateApiIdAndCorrelationIdContent()
+        {
+            var apiIdAndCorId = ApiId.Concat(CorrelationId)
+                                        .ToList();
+            return string.Join(",", apiIdAndCorId);
+        }
+
+        private string CreateFailedRequestsContent(string apiIdAndCorIdCsv)
+        {
+            string errorCodeCsv = string.Join(",", LastErrorCode);
+            return $"{apiIdAndCorIdCsv}{TelemetryConstants.HttpTelemetryPipe}{errorCodeCsv}";
         }
 
         public string GetCsvAsCurrent()
@@ -54,13 +87,11 @@ namespace Microsoft.Identity.Client.TelemetryCore.Internal
             // csv expected format:
             // 1|api_id,force_refresh|platform_config
             string[] myValues = new string[] {
-                ApiId,
-                ConvertFromStringToBitwise(ForceRefresh),
-                };
+                ApiId.FirstOrDefault(),
+                ConvertFromStringToBitwise(ForceRefresh)};
 
             string csvString = string.Join(",", myValues);
-            csvString = $"{TelemetryConstants.HttpTelemetrySchemaVersion1}{TelemetryConstants.HttpTelemetryPipe}{csvString}{TelemetryConstants.HttpTelemetryPipe}";
-            return csvString;
+            return $"{TelemetryConstants.HttpTelemetrySchemaVersion2}{TelemetryConstants.HttpTelemetryPipe}{csvString}{TelemetryConstants.HttpTelemetryPipe}";
         }
 
         private string ConvertFromStringToBitwise(string value)
--- a/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
+++ b/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
@@ -38,9 +38,9 @@ namespace Microsoft.Identity.Client.TelemetryCore.Internal
         }
 
         public string ForceRefresh { get; set; } = string.Empty;
-        public static List<string> ApiId { get; set; } = new List<string>();
-        public static List<string> CorrelationId { get; set; } = new List<string>();
-        public static List<string> LastErrorCode { get; set; } = new List<string>();
+        public List<string> ApiId { get; set; } = new List<string>();
+        public List<string> CorrelationId { get; set; } = new List<string>();
+        public List<string> LastErrorCode { get; set; } = new List<string>();
 
         public string GetCsvAsPrevious(int successfulSilentCallCount)
         {
@@ -65,7 +65,10 @@ namespace Microsoft.Identity.Client.TelemetryCore.Internal
             }
             else
             {
-                return data.Substring(0, length: 3800);
+                ApiId?.Clear();
+                CorrelationId?.Clear();
+                LastErrorCode?.Clear();
+                return string.Empty;
             }                
         }
 
@@ -85,7 +88,7 @@ namespace Microsoft.Identity.Client.TelemetryCore.Internal
         public string GetCsvAsCurrent()
         {
             // csv expected format:
-            // 1|api_id,force_refresh|platform_config
+            // 2|api_id,force_refresh|platform_config
             string[] myValues = new string[] {
                 ApiId.FirstOrDefault(),
                 ConvertFromStringToBitwise(ForceRefresh)};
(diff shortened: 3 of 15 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -135,6 +135,10 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                         apiEvent.TenantId = authenticationResult.TenantId;
                         apiEvent.AccountId = authenticationResult.UniqueId;
                         apiEvent.WasSuccessful = true;
+                        if (AuthenticationRequestParameters.ApiId == ApiEvent.ApiIds.AcquireTokenSilent)
```
bgavrilMS: Please move this to SilentRequest subclass. Typically when there is a class hierarchy, having code that is like

```
if (subtype is X) then do stuff
```

is a code smell. It suggests that either the class hierarchy is wrong or that there needs to be more virtual methods (hooks) in the hierarchy. In this case it's pretty simple, I think you can move this to SilentRequest

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,78 @@ public HttpTelemetryContent(EventBase evt)
...
+            string apiIdCorIdData = CreateApiIdAndCorrelationIdContent();
+            string data = string.Format(CultureInfo.InvariantCulture,
+                $"{TelemetryConstants.HttpTelemetrySchemaVersion2}{TelemetryConstants.HttpTelemetryPipe}{successfulSilentCallCount}{TelemetryConstants.HttpTelemetryPipe}" +
+                $"{CreateFailedRequestsContent(apiIdCorIdData)}" +
+                $"{TelemetryConstants.HttpTelemetryPipe}");
 
-            string csvString = string.Join(",", myValues);
-            csvString = $"{TelemetryConstants.HttpTelemetrySchemaVersion1}{TelemetryConstants.HttpTelemetryPipe}{csvString}{TelemetryConstants.HttpTelemetryPipe}";
-            return csvString;
+            if (data.Length <= 3800)
```
trwalke: >3800 [](start = 31, length = 4)

add a variable to describe what this is.

Also, why 3800?

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/Telemetry/TelemetryHelperTests.cs
```diff
@@ -14,7 +14,7 @@ public class TelemetryHelperTests
     {
         private const string CorrelationId = "thetelemetrycorrelationid";
         private const string ClientId = "theclientid";
-        private _TestEvent _trackingEvent;
+        internal _TestEvent _trackingEvent;
```
trwalke: Is this being used somewhere outside of the class? If not, does it need to be internal?

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/Telemetry/TelemetryHelperTests.cs
```diff
@@ -39,7 +39,7 @@ public void HandleTelemetryEvents(List<Dictionary<string, string>> events)
             }
         }
 
-        private class _TestEvent : EventBase
```
trwalke: Is this being used somewhere outside of the class? If not, does it need to be internal?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
trwalke: i see, is this the 4kb data cap?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,78 @@ public HttpTelemetryContent(EventBase evt)
...
+            if (shouldClearCache)
+            {
+                ApiId?.Clear();
+                CorrelationId?.Clear();
+                LastErrorCode?.Clear();
+            }
+        }
+
         public string ForceRefresh { get; set; } = string.Empty;
+        public static List<string> ApiId { get; set; } = new List<string>();
```
bgavrilMS: I think this can be done without statics.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,78 @@ public HttpTelemetryContent(EventBase evt)
...
+                $"{TelemetryConstants.HttpTelemetryPipe}");
 
-            string csvString = string.Join(",", myValues);
-            csvString = $"{TelemetryConstants.HttpTelemetrySchemaVersion1}{TelemetryConstants.HttpTelemetryPipe}{csvString}{TelemetryConstants.HttpTelemetryPipe}";
-            return csvString;
+            if (data.Length <= 3800)
+            {
+                return data;
+            }
+            else
```
bgavrilMS: Else is not needed. I would rewrite it as 
`if (data.Len>3800) { return ...} return data;

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,78 @@ public HttpTelemetryContent(EventBase evt)
...
+            {
+                return data.Substring(0, length: 3800);
```
bgavrilMS: Is this really valid telemetry data? How do you know if you are cutting the string in the right place? I think it would be safer to not send data if len > 3800 or come with a proper algorithm to send a batch.

src/client/Microsoft.Identity.Client/TelemetryCore/ITelemetryManager.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Identity.Client.TelemetryCore
     internal interface ITelemetryManager
     {
         TelemetryCallback Callback { get; }
-
+        int SuccessfulSilentCallCount { get; set; }
```
bgavrilMS: This object is not responsible for knowing telemetry details. It manages the telemetry operations. This property needs to be somewhere else.

bgavrilMS: Probably HttpTelemetryContent

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/HttpTelemetryTests.cs
```diff
@@ -64,63 +69,171 @@ public async Task AcquireTokenInteractive_UnknownError_FollowedByASuccessfulToke
...
+            HttpTelemetryContent httpTelemetryContent = CreateHttpTelemetryContent();
+            List<string> errorCodes = new List<string>(new string[] { MsalError.AuthenticationFailed, MsalError.UserMismatch, MsalError.InvalidGrantError });
+            PopulateHttpTelemetryContent(errorCodes);
+            HttpTelemetryContent.ApiId.Add("1005");
+            HttpTelemetryContent.CorrelationId.Add("b1e662cf-8efb-4e13-b89a-71e845bbb62f");
+
+            Assert.AreEqual(csvString, httpTelemetryContent.GetCsvAsPrevious(1));
+        }
+
+        [TestMethod]
```
bgavrilMS: These tests are ok, but we need to have higher level tests that check this. We need to assert on the telemetry headers on E2E tests, either unit tests or integration tests. 

I'd expect to have a test that does smth like: 

1. Acquire Token Interactive (or U/P or whatever)
-> check telemetry 
2. Acquire Token Silent
-> check telemetry
3. Acquire Token Silent again, with an expired AT
-> check telemetry

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,81 @@ public HttpTelemetryContent(EventBase evt)
...
+            string apiIdCorIdData = CreateApiIdAndCorrelationIdContent();
+            string data = string.Format(CultureInfo.InvariantCulture,
+                $"{TelemetryConstants.HttpTelemetrySchemaVersion2}{TelemetryConstants.HttpTelemetryPipe}{successfulSilentCallCount}{TelemetryConstants.HttpTelemetryPipe}" +
+                $"{CreateFailedRequestsContent(apiIdCorIdData)}" +
+                $"{TelemetryConstants.HttpTelemetryPipe}");
 
-            string csvString = string.Join(",", myValues);
-            csvString = $"{TelemetryConstants.HttpTelemetrySchemaVersion1}{TelemetryConstants.HttpTelemetryPipe}{csvString}{TelemetryConstants.HttpTelemetryPipe}";
-            return csvString;
+            if (data.Length <= 3800)
```
bgavrilMS: No magic numbers please. Please have a constant documenting this number. Also pls read https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/

bgavrilMS: Pretty sure the chars in that header are 1 byte / char.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
henrik-me: return to reduce nesting I suppose is what that is called. which is a great idea. Not sure this is the right thing to do though. This confuses me.

---
In reply to: [364811803](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1564#discussion_r364811803) [](ancestors = 364811803)

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,81 @@ public HttpTelemetryContent(EventBase evt)
...
+            }
+        }
+
         public string ForceRefresh { get; set; } = string.Empty;
+        public List<string> ApiId { get; set; } = new List<string>();
+        public List<string> CorrelationId { get; set; } = new List<string>();
+        public List<string> LastErrorCode { get; set; } = new List<string>();
 
-        public string GetCsvAsPrevious()
+        public string GetCsvAsPrevious(int successfulSilentCallCount)
```
henrik-me: > public string GetCsvAsPrevious(int successfulSilentCallCount) [](start = 7, length = 62)

Should collect 4k data and send that. The remaining data should stay. not sure why you clear things

bgavrilMS: It's fairly complex and time consuming to keep creating a telemetry string and to measure it to see it is not > 4k. 4k is a LOT of data, so we're talking about a very rare corner case. IMO a proxy such as a number of api / http events is a better match for this.

henrik-me: Fair point. However the time this is needed is when things goes wrong and it has been a long time since the last successful connection. I think nothing should prevent a proxy approach to this.

---
In reply to: [366462971](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1564#discussion_r366462971) [](ancestors = 366462971)

src/client/Microsoft.Identity.Client/TelemetryCore/ITelemetryManager.cs
bgavrilMS: Having looked more at this, all the telemetry data you need is the telemetry manager - in the event collections.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,81 @@ public HttpTelemetryContent(EventBase evt)
...
+            if (shouldClearCache)
+            {
+                ApiId?.Clear();
+                CorrelationId?.Clear();
+                LastErrorCode?.Clear();
+            }
+        }
+
         public string ForceRefresh { get; set; } = string.Empty;
+        public List<string> ApiId { get; set; } = new List<string>();
```
bgavrilMS: None of these are needed. All the data is in the TelemeteryManager, in the event collections.

tests/Microsoft.Identity.Test.Integration.net45/Infrastructure/TestHttpClientFactory.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.Net.Http;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Test.Integration.net45.Infrastructure
+{
+    public class TestHttpClientFactory : IMsalHttpClientFactory
```
bgavrilMS: `HttpSnifferClientFactory` ? Or smth like that :) SmellyFactory..

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -190,6 +205,8 @@ private async Task<AuthenticationResult> RunTestForUserAsync(LabResponse labResp
             userCacheAccess.AssertAccessCounts(1, 1); // the assert calls GetAccounts
             Assert.IsFalse(userCacheAccess.LastNotificationArgs.IsApplicationCache);
 
+            AssertTelemetryHeaders(factory, LastRequestExpectedTelemetryBasic);
```
bgavrilMS: So this is a bit concerning. It seems that even though the test is driving MSAL thorugh a lot, the telemetry values stay the same?

tests/Microsoft.Identity.Test.Integration.net45/Infrastructure/HttpTelemetryRecorder.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Identity.Client.TelemetryCore;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Integration.net45.Infrastructure
+{
+    public class HttpTelemetryRecorder
```
bgavrilMS: This is looking much better. I would add some methods that make assertions, smth like 

`AssertCurrentTelemetry(expectedApiId, expectedErrorCode ...)`

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -211,9 +229,12 @@ private async Task<AuthenticationResult> RunTestForUserAsync(LabResponse labResp
             userCacheAccess.AssertAccessCounts(3, 3);
             Assert.IsFalse(userCacheAccess.LastNotificationArgs.IsApplicationCache);
 
+            AssertTelemetryHeaders(factory, LastRequestExpectedTelemetryApiCorId);
```
bgavrilMS: It would be easier if we had a higher level assertion call similar to `userCacheAccess.AssertAccessCounts(3, 3)`. See prev comment

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
```diff
@@ -18,49 +20,81 @@ public HttpTelemetryContent(EventBase evt)
                 evt.TryGetValue(MsalTelemetryBlobEventNames.ApiErrorCodeConstStrKey, out string errorCode);
                 evt.TryGetValue(MsalTelemetryBlobEventNames.ForceRefreshId, out string forceRefresh);
 
-                ApiId = apiId ?? string.Empty;
-                CorrelationId = correlationId ?? string.Empty;
-                LastErrorCode = errorCode ?? string.Empty;
+                ApiId.Add(apiId ?? string.Empty);
```
pmaytak: Should these properties remain as strings rather than lists? As far as I saw, HttpTelemetryContent is instantiated for each EventBase instance and nothing is ever added to these lists besides in here.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
pmaytak: C# char is 16bit by default?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
pmaytak: The spec shows that the telemetry data can have multiple errors in it, right? In that sense, the length is not super consistent. What happens if we send an oversized request, does it return error? Then we lose the data either way. Can we really truncate it though (the csv will be malformed?)? We could consider using StringBuilder and/or Length property, maybe it's more performant.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryManager.cs
```diff
@@ -137,7 +138,7 @@ public string FetchAndResetPreviousHttpTelemetryContent()
             {
                 var httpTelemetryContent = new HttpTelemetryContent(_mostRecentStoppedApiEvent);
                 _mostRecentStoppedApiEvent = null;
-                return httpTelemetryContent.GetCsvAsPrevious();
+                return httpTelemetryContent.GetCsvAsPrevious(SuccessfulSilentCallCount);
```
pmaytak: The spec mentions that the requests should be deleted if there was an error in receiving by server side; do we handle that scenario?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
bgavrilMS: Completely agree, cutting the data will result in malformed CSV. Might as well not send it :). Not sure about Lenght property on StringBuilder. In C# strings are UTF-8, so size of char is 1 or 2 bytes (see https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/). Might as well consider it 2 bytes or transform to byte array and measure that. 

During telemetry meeting there was discussions about aggregating this data. Not sure if any proposal were sent out on this?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/HttpTelemetryContent.cs
bgavrilMS: 2 bytes.

Issue #1543, opened 2019-12-12
[Feature Request] Migrate PKey Auth implementation to MSAL.NET


trwalke:
PKey auth is needed to handle device authentication on mobile devices where browsers do not support client TLS and broker is not being used. The implementation should result in the the same behavior as ADAL.NET.

We can and should do the same for desktop. This is required for certain CA scenarios. Basic idea is we return not supported as that is what CA policies are asking for.
Align with support in other desktop libraries e.g.: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/720166

NOTE: System browsers will not be supported. Users will have to use the broker instead.

E2E validation:
Step-by-step policy configuration to get the optional-device-authentication:

1.	Create a CA policy for a particular user/app. In my example it’s All users / Sharepoint Online.
2.	Select Conditions -> Device State -> Configure: “Yes”
  a.	‘Include’ tab is selected with ‘All device state’
3.	Go to the exclude tab
  a.	Select any (or both) checkboxes
4.	Choose Grant Control = MFA.

With this configuration we are saying that all users going to Sharepoint should perform MFA, except if their device is compliant or Hybrid Domain joined.
So, what we do in ESTS is to attempt a best-effort device authentication to get the device properties, and would skip this policy if the device is compliant or hybrid domain joined.

When we attempt a device authentication in the silent endpoint (/token endpoint) we would expect a response with either the device certificate or with empty details. If it’s empty, we would just fallback to the MFA challenge.
The problem with Xamarin apps is that they are not responding with an empty device info, they just appear to be stuck in the client, and we never have a chance to perform MFA.

henrik-me:
We should ensure to expand this to all platforms as certain CA policies needs us to return not supported  as a response to the protocol.

jmprieur:
If the client does not answer to the challenge to the protocol, there is an exception. We need to return an empty value. We need to do it for Mobile, Desktop, UWP, and .NET Core.
This is only when using the embedded web view

jmprieur:
On hold waiting for the test environment to be ready in the Lab

cheenamalhotra:
Will this feature need any changes in consuming libraries or it will be implicitly handled by MSAL?

trwalke:
No changes needed, MSAL will automatically start detecting PKey Auth challenges and respond to them assuming the proper device cert is available.

PR #1567, opened 2020-01-09
Migration of PKeyAuth to MSAL

trwalke:
Implementation of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1543


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
```diff
@@ -199,5 +199,10 @@ public virtual IPoPCryptoProvider GetDefaultPoPCryptoProvider()
         {
             throw new NotImplementedException();
         }
+
+        public virtual IDeviceAuthManager CreateDeviceAuthManager()
+        {
+            throw new NotImplementedException();
```
bgavrilMS: What will happen on platforms where you do not define a platform proxy, like .net core, .net standard or xamarin.mac?

src/client/Microsoft.Identity.Client/Platforms/Shared/Apple/AppleDeviceAuthManager.cs
```diff
@@ -4,14 +4,15 @@
...
 namespace Microsoft.Identity.Client.Platforms.Shared.Apple
 {
-    internal class DeviceAuthHelper
+    internal class AppleDeviceAuthManager : IDeviceAuthManager
     {
-        public static bool CanHandleDeviceAuthChallenge { get { return false; } }
+        public bool CanHandleDeviceAuthChallenge { get { return false; } }
 
-        public static Task<string> CreateDeviceAuthChallengeResponseAsync(IDictionary<string, string> challengeData)
+        public Task<string> CreateDeviceAuthChallengeResponseAsync(IDictionary<string, string> challengeData)
```
bgavrilMS: Why doesn't this method just throw or do nothing since CanHandle.. returns false..

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
bgavrilMS: I suggest that you create a NotSupportedDeviceAuthManager for these platforms for now.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -311,5 +338,32 @@ private Uri CreateFullEndpointUri(Uri endPoint)
...
+
+            return data;
+        }
+
+        private bool IsDeviceAuthChallenge(HttpResponse response, bool respondToDeviceAuthChallenge)
+        {
+            return _deviceAuthManager.CanHandleDeviceAuthChallenge
+                   && response != null
+                   && respondToDeviceAuthChallenge
+                   && response?.Headers != null
```
bgavrilMS: response is not null so you don't need to use the `?.` operator.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -89,6 +95,24 @@ internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, R
 
                 if (response.StatusCode != HttpStatusCode.OK || expectErrorsOn200OK)
                 {
+                    //Check if we can perform device auth
+                    if (IsDeviceAuthChallenge(response, respondToDeviceAuthChallenge))
```
bgavrilMS: This block of code should be unit tested.

src/client/Microsoft.Identity.Client/Platforms/net45/Native/X509Native.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Security.Cryptography;
+using Microsoft.Win32.SafeHandles;
+
+namespace Microsoft.Identity.Client.Platforms.net45.Native
+{
+    internal static class X509Native
```
bgavrilMS: I wonder if all this native will work from .net core - on Windows only of course. It might - consider opening an issue so that we keep track of this. IMO .net classic is less important than .net core.

src/client/Microsoft.Identity.Client/Platforms/net45/NetDesktopDeviceAuthManager.cs
```diff
@@ -0,0 +1,157 @@
...
+            Task<string> resultTask =
+                Task.Factory.StartNew(
+                    () =>
+                    {
+                        return string.Format(CultureInfo.InvariantCulture, authHeaderTemplate, authToken,
+                            challengeData["Context"],
+                            challengeData["Version"]);
+                    });
+
+            return await resultTask.ConfigureAwait(false);
```
bgavrilMS: YOu don't need to do this trick with the factory, it's simpler to have just a 

` Task.FromResult(stringResult)`

src/client/Microsoft.Identity.Client/Platforms/net45/NetDesktopDeviceAuthManager.cs
```diff
@@ -0,0 +1,157 @@
...
+namespace Microsoft.Identity.Client.Platforms.net45
+{
+    internal class NetDesktopDeviceAuthManager : IDeviceAuthManager
+    {
+        public bool CanHandleDeviceAuthChallenge
+        {
+            get { return true; }
+        }
+
+        public async Task<string> CreateDeviceAuthChallengeResponseAsync(IDictionary<string, string> challengeData)
```
bgavrilMS: I do not understand how much code you have written and how much was just copied from ADAL - please, next time you port code, please let reviewers know so that we don't review copied code. 

The issue here are unit tests: 

- if you copied this code, it's likely that you don't understand it enough to write unit tests. Are there unit tests in ADAL around this? If so, they need to be copied. If not, we need to asses the importance of this code, i.e. how likely it is to change. If it is likely to have bugs or to evolve, we need tests. I would ask @henrik-me for an opinion of shipping this without unit tests.

src/client/Microsoft.Identity.Client/Instance/Discovery/NetworkMetadataProvider.cs
```diff
@@ -73,7 +77,7 @@ private void CacheInstanceDiscoveryMetadata(InstanceDiscoveryResponse instanceDi
           Uri authority,
           RequestContext requestContext)
         {
-            var client = new OAuth2Client(requestContext.Logger, _httpManager, _telemetryManager);
+            var client = new OAuth2Client(requestContext.Logger, _httpManager, _telemetryManager, _deviceAuthManager);
```
bgavrilMS: I don't think this is necessary. You can just pass null here, i.e. the device auth manager should be null since you don't need to perform the extra complexity of device auth for instance discovery.

src/client/Microsoft.Identity.Client/Core/ServiceBundle.cs
```diff
@@ -44,7 +44,7 @@ internal class ServiceBundle : IServiceBundle
                 TelemetryManager = new TelemetryManager(config, PlatformProxy, config.TelemetryCallback);
             }
 
-            InstanceDiscoveryManager = new InstanceDiscoveryManager(HttpManager, TelemetryManager, shouldClearCaches, config.CustomInstanceDiscoveryMetadata);
+            InstanceDiscoveryManager = new InstanceDiscoveryManager(HttpManager, TelemetryManager, shouldClearCaches, config.CustomInstanceDiscoveryMetadata, PlatformProxy.CreateDeviceAuthManager());
```
bgavrilMS: I don't think this should be added to instance discovery

src/client/Microsoft.Identity.Client/Platforms/net45/NetDesktopDeviceAuthManager.cs
henrik-me: unit tests are needed. e2e tests are needed

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -21,7 +21,23 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <DefineConstants>$(DefineConstants);NETSTANDARD_RUNTIME</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net45|AnyCPU'">
```
henrik-me: >Debug|net45|AnyCPU' [](start = 80, length = 19)

really?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -21,7 +21,23 @@
...
+  <ItemGroup>
+    <Compile Include="PlatformsCommon\Interfaces\IDeviceAuthManager.cs" />
+    <Compile Include="Platforms\net45\NetDesktopDeviceAuthManager.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Remove="PlatformsCommon\Interfaces\IDeviceAuthManager.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Remove="Platforms\net45\NetDesktopDeviceAuthManager.cs" />
+  </ItemGroup>
```
henrik-me: what about netcore?

henrik-me:
Did you explore calling into .net instead?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -21,7 +21,23 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <DefineConstants>$(DefineConstants);NETSTANDARD_RUNTIME</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net45|AnyCPU'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
   <ItemGroup>
     <ReferenceAssemblyProjectReference Include="..\Microsoft.Identity.Client.Ref\Microsoft.Identity.Client.Ref.csproj" />
   </ItemGroup>
+  <ItemGroup>
```
bgavrilMS: None of these changes are required. Please delete all this stuff and you should notice that it just works (TM).

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
bgavrilMS: @trwalke - please add this MSBuild property in the existing property group for .NET45 which can be found here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/a013e8c10b7c10bbfac84e318000d591f3a35373/src/client/Directory.Build.props#L87

src/client/Microsoft.Identity.Client/OAuth2/DeviceAuthJWTResponse.cs
```diff
@@ -0,0 +1,74 @@
...
+        public DeviceAuthHeader(string base64EncodedCertificate)
+        {
+            this.Alg = "RS256";
+            this.Type = "JWT";
+            this.X5c = new List<string>();
+            this.X5c.Add(base64EncodedCertificate);
+        }
+
+        [DataMember(Name = "x5c", IsRequired = true)]
+        public List<string> X5c { get; set; }
```
bgavrilMS: Consider exposing an `IList` instead.

src/client/Microsoft.Identity.Client/OAuth2/DeviceAuthJWTResponse.cs
```diff
@@ -0,0 +1,74 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Runtime.Serialization;
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.OAuth2
+{
+    [DataContract]
```
bgavrilMS: Please rebase your change and start using the newtonsoft.json attributes (DataContract -> JsonObject, DataMember -> JsonProperty). These also work with newtonsoft, but it be more consistent to use the native ones.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
trwalke: Yes

src/client/Microsoft.Identity.Client/Platforms/Shared/Apple/AppleDeviceAuthManager.cs
trwalke: There are some places in code that call the CreateDeviceAuthChallengeResponseAsync directly without checking CanHandle for iOS and android. The response returned here tells AAD to bypass the PKeyAuth check.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
trwalke: yes

src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -30,7 +30,7 @@ internal override async Task<AuthenticationResult> ExecuteAsync(CancellationToke
         {
             await ResolveAuthorityEndpointsAsync().ConfigureAwait(false);
 
-            var client = new OAuth2Client(ServiceBundle.DefaultLogger, ServiceBundle.HttpManager, ServiceBundle.TelemetryManager);
+            var client = new OAuth2Client(ServiceBundle.DefaultLogger, ServiceBundle.HttpManager, ServiceBundle.TelemetryManager, ServiceBundle.PlatformProxy.CreateDeviceAuthManager());
```
bgavrilMS: Is PKAuth required on this endpoint? If not, prefer to pass null.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -16,22 +16,30 @@
...
 using Microsoft.Identity.Client.Instance.Discovery;
 using Microsoft.Identity.Client.TelemetryCore.Internal;
+using System.Linq;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
 
 namespace Microsoft.Identity.Client.OAuth2
 {
     internal class OAuth2Client
     {
+        private const string DeviceAuthHeaderName = "x-ms-PKeyAuth";
```
bgavrilMS: There are some classes with constants around, please move these constants there, preferably in a new class, smth like PKAuthConstants.DeviceAuthHeaderName

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -58,12 +66,14 @@ public async Task<MsalTokenResponse> GetTokenAsync(Uri endPoint, RequestContext
...
 
-
-
-        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false)
+        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false, bool respondToDeviceAuthChallenge = true)
         {
             bool addCorrelationId = requestContext != null && !string.IsNullOrEmpty(requestContext.Logger.CorrelationId.ToString());
-            AddCommonHeaders(requestContext, addCorrelationId);
+            //No need to re-add headers for the replayed PKeyAuth request
+            if (respondToDeviceAuthChallenge)
```
bgavrilMS: `sendDeviceAuthParams` ?

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -58,12 +66,14 @@ public async Task<MsalTokenResponse> GetTokenAsync(Uri endPoint, RequestContext
             return await ExecuteRequestAsync<MsalTokenResponse>(endPoint, HttpMethod.Post, requestContext).ConfigureAwait(false);
         }
 
-
-
-        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false)
+        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false, bool respondToDeviceAuthChallenge = true)
         {
             bool addCorrelationId = requestContext != null && !string.IsNullOrEmpty(requestContext.Logger.CorrelationId.ToString());
-            AddCommonHeaders(requestContext, addCorrelationId);
```
bgavrilMS: I think this is wrong, but please correct me if my judgement fails - we should ALWAYS add the correlation ID. You probably want to prevent only the PKAuth header from being added.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -140,6 +171,9 @@ private bool AddCommonHeaders(RequestContext requestContext, bool addCorrelation
 
             _headers.Add(TelemetryConstants.XClientLastTelemetry, _telemetryManager.FetchAndResetPreviousHttpTelemetryContent());
             _headers.Add(TelemetryConstants.XClientCurrentTelemetry, _telemetryManager.FetchCurrentHttpTelemetryContent(requestContext.CorrelationId.AsMatsCorrelationId()));
+
+            //add pkeyauth header
+            _headers.Add(DeviceAuthHeaderName, DeviceAuthHeaderValue);
```
bgavrilMS: So we add this on all platforms, even if a deviceManager does not exist? I would argue that if a device manager does not  exist, then we do not add the header? 

From an implementation detail, I would let the deviceManager itself add this header, that way the NullDeviceManager would not nothing, while the other managers can add it.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -89,6 +99,27 @@ internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, R
 
                 if (response.StatusCode != HttpStatusCode.OK || expectErrorsOn200OK)
                 {
+                    //Check if we can perform device auth
+                    if (_deviceAuthManager != null && IsDeviceAuthChallenge(response, respondToDeviceAuthChallenge))
```
bgavrilMS: Please have a look at the `NULL object pattern` (e.g. [here](https://www.c-sharpcorner.com/article/null-object-design-pattern/)). 

Instead of littering the code with `if (_deviceManager != null) then ...` , we can implement a NullDeviceManager : IDeviceManager. This object's behavior is "do nothing". It will provide a much better abstraction.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -291,7 +291,7 @@ internal async Task ResolveAuthorityEndpointsAsync()
             IDictionary<string, string> additionalBodyParameters,
             CancellationToken cancellationToken)
         {
-            OAuth2Client client = new OAuth2Client(ServiceBundle.DefaultLogger, ServiceBundle.HttpManager, ServiceBundle.TelemetryManager);
+            OAuth2Client client = new OAuth2Client(ServiceBundle.DefaultLogger, ServiceBundle.HttpManager, ServiceBundle.TelemetryManager, ServiceBundle.PlatformProxy.CreateDeviceAuthManager());
```
bgavrilMS: Why don't you expose `ServiceBundle.DeviceManager` like all the other objects in ServiceBundle? The ServiceBundle's lifetime is the same as the `pca` / `cca` objects, which seems better suited for DeviceManager. I do not see a need to have a new object like this for every request.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -311,5 +345,32 @@ private Uri CreateFullEndpointUri(Uri endPoint)
...
+
+        private bool IsDeviceAuthChallenge(HttpResponse response, bool respondToDeviceAuthChallenge)
+        {
+            return _deviceAuthManager.CanHandleDeviceAuthChallenge
+                   && response != null
+                   && respondToDeviceAuthChallenge
+                   && response.Headers != null
+                   && response.StatusCode == HttpStatusCode.Unauthorized
+                   && response.Headers.Contains(WwwAuthenticateHeader)
+                   && response.Headers.GetValues(WwwAuthenticateHeader).FirstOrDefault()
```
bgavrilMS: .First since you already know it contains the stuff.

bgavrilMS: Please make sure you compare string in a case insensitive way - see https://stackoverflow.com/questions/444798/case-insensitive-containsstring

bgavrilMS: I would use .Any() instead of .First() in the unlikely event that the server returns 2 values there.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
bgavrilMS: Can we have the deviceManager take over the entire responsability of looking at an HTTPResponse and returning a Header? THe OAuth client shouldn't do this. 

Smth like 

`string header = await _deviceManager.TryCreateDeviceAuthChallengeAsync(response)`

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -311,5 +345,32 @@ private Uri CreateFullEndpointUri(Uri endPoint)
                 }
             }
         }
+
+        private IDictionary<string, string> ParseChallengeData(HttpResponse response)
+        {
+            IDictionary<string, string> data = new Dictionary<string, string>();
```
bgavrilMS: Dictionaries are case senstive by default, so when you call thinkgs like `ContainsKey` it behaves case-sensitive, which is not what you want. Case insensitive dictionaries - https://stackoverflow.com/questions/13988643/case-insensitive-dictionary-with-string-key-type-in-c-sharp

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -311,5 +345,32 @@ private Uri CreateFullEndpointUri(Uri endPoint)
...
+            string wwwAuthenticate = response.Headers.GetValues(WwwAuthenticateHeader).FirstOrDefault();
```
bgavrilMS: Use .Single if you are sure there is only one value. Otherwise, extract the right value using .Where

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -311,5 +345,32 @@ private Uri CreateFullEndpointUri(Uri endPoint)
                 }
             }
         }
+
+        private IDictionary<string, string> ParseChallengeData(HttpResponse response)
```
bgavrilMS: None of the logic in this class is unit tested. Every block of code needs to be unit tested, especially code like this.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs
```diff
@@ -217,5 +218,10 @@ private static bool IsChromeEnabled()
         }
 
         public override bool UseEmbeddedWebViewDefault => false;
+
+        public override IDeviceAuthManager CreateDeviceAuthManager()
+        {
+            return new AndroidDeviceAuthManager();
```
bgavrilMS: There is no need to create dummy device managers for each platform. Just create a `NullDeviceAuthManager` for all platforms that doesn't support this.

src/client/Microsoft.Identity.Client/Platforms/Android/EmbeddedWebview/AndroidDeviceAuthManager.cs
```diff
@@ -4,17 +4,16 @@
...
     {
-        public static bool CanHandleDeviceAuthChallenge {
-            get { return false; }
-        }
+        public bool CanHandleDeviceAuthChallenge { get { return false; } }
 
-        public static Task<string> CreateDeviceAuthChallengeResponseAsync(IDictionary<string, string> challengeData)
+        public Task<string> CreateDeviceAuthChallengeResponseAsync(IDictionary<string, string> challengeData)
         {
             return Task.FromResult(string.Format(CultureInfo.InvariantCulture, @"PKeyAuth Context=""{0}"",Version=""{1}""", challengeData[BrokerConstants.ChallangeResponseContext], challengeData[BrokerConstants.ChallangeResponseVersion]));
```
bgavrilMS: This is doing too much for a "null" object.

src/client/Microsoft.Identity.Client/Platforms/Android/EmbeddedWebview/AndroidDeviceAuthManager.cs
```diff
@@ -4,17 +4,16 @@
...
-    internal class DeviceAuthHelper
+    internal class AndroidDeviceAuthManager : IDeviceAuthManager
     {
-        public static bool CanHandleDeviceAuthChallenge {
-            get { return false; }
-        }
+        public bool CanHandleDeviceAuthChallenge { get { return false; } }
 
-        public static Task<string> CreateDeviceAuthChallengeResponseAsync(IDictionary<string, string> challengeData)
+        public Task<string> CreateDeviceAuthChallengeResponseAsync(IDictionary<string, string> challengeData)
```
bgavrilMS: Can be made to do more, e.g. `Task DecorateRequestWithDeviceCodeChallengeAsync(HttpRequestMessage)`.

src/client/Microsoft.Identity.Client/Platforms/net45/Native/BCryptNative.cs
```diff
@@ -0,0 +1,354 @@
...
+   //
+    // Public facing enumerations
+    //
+
+    /// <summary>
+    ///     Padding modes 
+    /// </summary>
+    [SuppressMessage("Microsoft.Design", "CA1027:MarkEnumsWithFlags", Justification = "Public use of the enum is not as flags")]
+    [SuppressMessage("Microsoft.Design", "CA1008:EnumsShouldHaveZeroValue", Justification = "The native BCRYPT_PAD_NONE value is 1, not 0, and this is for interop.")]
+    public enum AsymmetricPaddingMode
```
bgavrilMS: I believe Brent recommended you to call into .NET FWK classes instead of pivoking.

src/client/Microsoft.Identity.Client/Platforms/netstandard13/NetStandard13DeviceAuthManager.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+
+namespace Microsoft.Identity.Client.Platforms.netstandard13
+{
+    internal class NetStandard13DeviceAuthManager : IDeviceAuthManager
```
bgavrilMS: Why are there 2 of these? netstandard13 and netstnadard?

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IDeviceAuthManager.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.PlatformsCommon.Interfaces
+{
+    internal interface IDeviceAuthManager
```
bgavrilMS: As discussed above, these objects need to do more so that other objects are less responsible.

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCoreDeviceAuthManager.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+
+namespace Microsoft.Identity.Client.Platforms.netcore
+{
+    internal class NetCoreDeviceAuthManager : IDeviceAuthManager
+    {
+        public bool CanHandleDeviceAuthChallenge { get { return false; } }
```
bgavrilMS: I was under the impression that PKAuth should work on .NET Core. If this is not the case, please make this assumption clear and perhaps open a new feauture / bug to add the missing functionality. In general .Net Core is much more important than .Net Classic and UWP, so there needs to be a strong argument for implementing smth on .Net Fx but not on .NET core.

jmprieur: Why wouldn't it work in .NET Core?

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -58,12 +66,14 @@ public async Task<MsalTokenResponse> GetTokenAsync(Uri endPoint, RequestContext
...
-
-        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false)
+        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false, bool respondToDeviceAuthChallenge = true)
         {
             bool addCorrelationId = requestContext != null && !string.IsNullOrEmpty(requestContext.Logger.CorrelationId.ToString());
-            AddCommonHeaders(requestContext, addCorrelationId);
+            //No need to re-add headers for the replayed PKeyAuth request
+            if (respondToDeviceAuthChallenge)
+            {
+                AddCommonHeaders(requestContext, addCorrelationId);
```
bgavrilMS: @henrik-me @jmprieur  - please clarify if we want to do PKAuth every time? This results in a pretty big pref impact.

This seems to effectively add an extra HTTP call to the `/token` endpoint. Not to mention all the crypto operations which also take time. I don't know if @trwalke measured the perf impact of PKAuth - perhaps he can add some numbers.

src/client/Microsoft.Identity.Client/Platforms/Mac/AuthenticationAgentNSWindowController.cs
```diff
@@ -211,7 +214,7 @@ void DecidePolicyForNavigation(WebView webView, NSDictionary actionInformation,
                 }
 
                 Dictionary<string, string> keyPair = CoreHelpers.ParseKeyValueList(query, '&', true, false, null);
-                string responseHeader = DeviceAuthHelper.CreateDeviceAuthChallengeResponseAsync(keyPair).Result;
+                string responseHeader = _deviceAuthManager.CreateDeviceAuthChallengeResponseAsync(keyPair).Result;
```
bgavrilMS: How come you're adding DeviceAuth to a single embedded UI in the whole stack - the one that is probably not used by anybody- MAC? 

1. Can you clarify that the device auth logic needs to occur on the /authroize endpoint? 
2. Was this tested ?

tests/Microsoft.Identity.Test.Unit.net45/OAuthClientTests.cs
```diff
@@ -164,6 +167,54 @@ public void OAuthClient_FailsWithServiceExceptionWhenResponseIsHttpNotFound()
                 });
         }
 
+        [TestMethod]
+        public void PKeyAuthSuccsesResponseTest()
```
bgavrilMS: There 3 types of tests that you need to write / modify: 

1. Integration tests - these are great to ensure that EVO keeps it's part of the contract. I don't think you need to write new tests, you just need to make sure the HTTP traffic is happening as you expect - with PKAuth challenges and headers. You can use this class to sniff the HTTP traffic in integration tests https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/1c03719723c125d94399c6496dd6db9d988a93a4

2. End to End unit tests like this one - these are great if you are able to mock IDeviceManager (see how we've mocked IBroker for an example). This allows you to test the OAuth2Client by manipualting the traffic. Potentially not needed if you move more logic to IDeviceManager

3. Conventional Unit tests which look at objects in isolation. In this case, the various implementations of IDeviceManager, especially the .net classic one needs to be tested. Consider starting with a well know public / private key pair and asserting the header values.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
trwalke: Yes, we will always send the correlation id. What is happening here is the pkeyauth code is replaying the same request after being issued a challenge with additional pkeyauth headers. We don't want to re-add common headers to the request because the common headers have already been added to it the first time

src/client/Microsoft.Identity.Client/Platforms/Mac/AuthenticationAgentNSWindowController.cs
trwalke: I am not sure what is going on here exactly. Mark took some components from ADAl's pkey auth classes and added them here for this class. i am assuming mark tested this when he added it.

bgavrilMS: Pretty sure you have to get rid of this since the devicemanager on MAC does nothing.

src/client/Microsoft.Identity.Client/Platforms/netstandard13/NetStandard13DeviceAuthManager.cs
trwalke: I remaned this but the second file stuck around for some reason. will be removed

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -16,7 +16,7 @@ internal class DeviceAuthHelper
         public static IDictionary<string, string> ParseChallengeData(HttpResponse response)
         {
             IDictionary<string, string> data = new Dictionary<string, string>();
-            string wwwAuthenticate = response.Headers.GetValues(PKeyAuthConstants.WwwAuthenticateHeader).FirstOrDefault();
+            string wwwAuthenticate = response.Headers.GetValues(PKeyAuthConstants.WwwAuthenticateHeader).Single();
```
bgavrilMS, line 19: It's probably safer to return empty dictionary if header does not exist.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
trwalke: Surprisingly, according to the spec the values for PKeyAuth are actually case sensitive so this is the correct comparison.

trwalke:
Core functionality updated, tests are coming shortly. Still working out a few issues with them

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
bgavrilMS: In that case, it would be good to add a comment about this, as it is certain nobody is going to remember this spec detail :)

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
trwalke: The header will exist in this method. there is a method before that ensures the headers exist before parsing them.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
trwalke: Wanted to clarify, this line is adding a predefined pkey auth header to the request that lets the STS know that the client is capable of handling PKeyAuth. The only time MSAL will actually perform the challenge is when authentication fails with a very specific error that contains the PKeyAuth challenge data. This will be the only time MSAL will sign the challenge and then replay the request. ADAL is using the same logic for this here, however, it is still up for debate if we want to go in a different direction.

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
trwalke: i agree, ill update

src/client/Microsoft.Identity.Client/Core/ServiceBundle.cs
```diff
@@ -47,6 +47,7 @@ internal class ServiceBundle : IServiceBundle
             InstanceDiscoveryManager = new InstanceDiscoveryManager(HttpManager, TelemetryManager, shouldClearCaches, config.CustomInstanceDiscoveryMetadata);
             WsTrustWebRequestManager = new WsTrustWebRequestManager(HttpManager);
             AuthorityEndpointResolutionManager = new AuthorityEndpointResolutionManager(this, shouldClearCaches);
+            DeviceAuthManager = PlatformProxy != null ? PlatformProxy.CreateDeviceAuthManager() : null;
```
bgavrilMS, line 50: PlatformProxy  is guaranteed not null, see line 33.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
bgavrilMS: Let's add some headers, they are only used for telemetry purposes, so it's better to have them.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -26,12 +29,14 @@ internal class OAuth2Client
...
         private readonly ITelemetryManager _telemetryManager;
+        private readonly IDeviceAuthManager _deviceAuthManager;
 
-        public OAuth2Client(ICoreLogger logger, IHttpManager httpManager, ITelemetryManager telemetryManager)
+        public OAuth2Client(ICoreLogger logger, IHttpManager httpManager, ITelemetryManager telemetryManager, IDeviceAuthManager deviceAuthManager)
         {
             _headers = new Dictionary<string, string>(MsalIdHelper.GetMsalIdParameters(logger));
             _httpManager = httpManager ?? throw new ArgumentNullException(nameof(httpManager));
             _telemetryManager = telemetryManager ?? throw new ArgumentNullException(nameof(telemetryManager));
+            _deviceAuthManager = deviceAuthManager;
```
bgavrilMS, line 39: You already are using the "null object pattern", which is great. So I think you should use it here as well, i.e. if deviceAuthManager is null, use the NullDeviceAuthmanager. This will avoid you from having to continously check `if (deviceAuthManager != null) ...`

bgavrilMS: Or make sure to always pass in a DeviceManager, that will also work. There is one attached to the service bundle...

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -89,6 +96,18 @@ internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, R
 
                 if (response.StatusCode != HttpStatusCode.OK || expectErrorsOn200OK)
                 {
+                    //Check if we can perform device auth
+                    if (sendDeviceAuthParams)
+                    {
+                        string responseHeader = await _deviceAuthManager.CreateDeviceAuthChallengeResponseAsync(response, endpointUri).ConfigureAwait(false);
```
bgavrilMS, line 102: I read throught the implementation and it seems to be ok for .net classic, but not ok for .net core, android etc. as the logic in `DeviceAuthHelper.GetBypassChallengeResponse` (in NullDeviceAuthManager` does not make any verification if the headers are there. 

In the .net 45 implementation you use `IsDeviceAuthChallenge` which is pretty defensive, but in the NullDeviceAuthManger you no longer use this.

bgavrilMS: Also I would like to suggest a slightly different pattern to make the code more readable. The method `CreateXXX` implies that a response header will always be created. But in reality, this is not the case. 

So instead you can have a method that is more like `bool TryCreateXXX(..., out responseHeader)`. This is typical pattern in C#, see for example [here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.8).

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -26,12 +29,14 @@ internal class OAuth2Client
         private readonly Dictionary<string, string> _queryParameters = new Dictionary<string, string>();
         private readonly IHttpManager _httpManager;
         private readonly ITelemetryManager _telemetryManager;
+        private readonly IDeviceAuthManager _deviceAuthManager;
 
-        public OAuth2Client(ICoreLogger logger, IHttpManager httpManager, ITelemetryManager telemetryManager)
+        public OAuth2Client(ICoreLogger logger, IHttpManager httpManager, ITelemetryManager telemetryManager, IDeviceAuthManager deviceAuthManager)
```
bgavrilMS, line 34: This code is quite old and needs to be rebased. There is now a `TokenClient` higher-level class that is responsible for the communication with /token endpoint. Not sure if you saw that. 

It's not clear to me if this PKAuth challenge only comes from the /token endpoint or from all the endpoints? If it's only from the /token endpoint, then probably it's better to move the logic to `TokenClient` class.

src/client/Microsoft.Identity.Client/Platforms/Mac/AuthenticationAgentNSWindowController.cs
bgavrilMS: `CreateDeviceAuthChallengeResponseAsync` return null here, but you do not check for this.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/NullDeviceAuthManager.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+
+namespace Microsoft.Identity.Client.PlatformsCommon.Shared
+{
+    internal class NullDeviceAuthManager : IDeviceAuthManager
```
bgavrilMS, line 12: `DeviceAuthHelper.GetBypassChallengeResponse(response)` is still a response, but CanHandleDeviceAuthChallenge  is set to false?

tests/Microsoft.Identity.Test.Unit.net45/OAuthClientTests.cs
```diff
@@ -164,6 +167,56 @@ public void OAuthClient_FailsWithServiceExceptionWhenResponseIsHttpNotFound()
...
+                //Initiates PKeyAuth challenge which will trigger an additional request sent to AAD to satisfy the PKeyAuth challenge
+                harness.HttpManager.AddMockHandler(
+                    new MockHttpMessageHandler
+                    {
+                        ExpectedMethod = HttpMethod.Post,
+                        ResponseMessage = MockHelpers.CreatePKeyAuthChallengeResponse()
+                    });
+                harness.HttpManager.AddSuccessTokenResponseMockHandlerForPost(TestConstants.AuthorityCommonTenant);
+
+                AuthenticationResult result = app
```
bgavrilMS, line 199: This test only checks that an additional request to the token endpoint is made. It does not check what headers are sent, any of the NetDesktopDeviceManager functionality etc.

src/client/Microsoft.Identity.Client/Core/ServiceBundle.cs
```diff
@@ -87,6 +87,8 @@ internal class ServiceBundle : IServiceBundle
 
         public IDeviceAuthManager DeviceAuthManager { get; }
 
+        public IHttpTelemetryManager HttpTelemetryManager { get; }
```
bgavrilMS, line 90: I think this is a merging problem. This is not part of your commit :(

src/client/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs
```diff
@@ -153,7 +153,7 @@ private void Add(AuthorityInfo authorityInfo, string userPrincipalName, Authorit
              string openIdConfigurationEndpoint,
              RequestContext requestContext)
         {
-            var client = new OAuth2Client(requestContext.Logger, _serviceBundle.HttpManager, _serviceBundle.TelemetryManager, _serviceBundle.DeviceAuthManager);
+            var client = new OAuth2Client(requestContext.Logger, _serviceBundle.HttpManager, _serviceBundle.MatsTelemetryManager, _serviceBundle.DeviceAuthManager);
```
bgavrilMS, line 156: I think you may need to re-merge.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/NullDeviceAuthManager.cs
trwalke: removed the CanHandleDeviceAuthChallenge  as it is no longer needed

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
bgavrilMS: Still active comment - please leave those headers there.

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.PlatformsCommon
+{
+    internal class DeviceAuthHelper
+    {
+        public static IDictionary<string, string> ParseChallengeData(HttpResponse response)
+        {
+            IDictionary<string, string> data = new Dictionary<string, string>();
+            string wwwAuthenticate = response.Headers.GetValues(PKeyAuthConstants.WwwAuthenticateHeader).Single();
```
bgavrilMS, line 20: Careful with `Single`, it will crash if there isn't exactly 1 header with that name. Consider using `SingleOrDefault` and then if wwwAuthenticate is null, return an empty dict.

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+            wwwAuthenticate = wwwAuthenticate.Substring(PKeyAuthConstants.PKeyAuthName.Length + 1);
+            List<string> headerPairs = CoreHelpers.SplitWithQuotes(wwwAuthenticate, ',');
+            foreach (string pair in headerPairs)
+            {
+                List<string> keyValue = CoreHelpers.SplitWithQuotes(pair, '=');
+                data.Add(keyValue[0].Trim(), keyValue[1].Trim().Replace("\"", ""));
```
bgavrilMS, line 26: If keyValue has 0 or 1 components, this code will throw. Code needs to be more defensive.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -73,10 +78,14 @@ public async Task<MsalTokenResponse> GetTokenAsync(Uri endPoint, RequestContext
             return await ExecuteRequestAsync<MsalTokenResponse>(endPoint, HttpMethod.Post, requestContext).ConfigureAwait(false);
         }
 
-        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false)
+        internal async Task<T> ExecuteRequestAsync<T>(Uri endPoint, HttpMethod method, RequestContext requestContext, bool expectErrorsOn200OK = false, bool sendDeviceAuthParams = true)
```
bgavrilMS, line 81: As discussed offline, this is not the right place to invoke the pkauth manager because of 2 reasons: 

- this class is responsible for making all calls to STS - /token, instance, oidc, ws-trust, /device_code
- Single Responsibility Principle - this low level class should not gain another responsibility, it's already doing a lot...

So my proposal is to move this logic to TokenClient.cs which handles communication with /token endpoint only.

trwalke: Lets discuss this one offline, i ran into some uncertainty while trying to make this change

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -17,13 +17,21 @@ internal class DeviceAuthHelper
         public static IDictionary<string, string> ParseChallengeData(HttpResponse response)
         {
             IDictionary<string, string> data = new Dictionary<string, string>();
-            string wwwAuthenticate = response.Headers.GetValues(PKeyAuthConstants.WwwAuthenticateHeader).Single();
+            string wwwAuthenticate = response.Headers.GetValues(PKeyAuthConstants.WwwAuthenticateHeader).SingleOrDefault();
             wwwAuthenticate = wwwAuthenticate.Substring(PKeyAuthConstants.PKeyAuthName.Length + 1);
```
bgavrilMS, line 21: you can still get a null ref at line 21

src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.PlatformsCommon.Shared;
```
henrik-me, line 12: Are these still needed (done several places)?

src/client/Microsoft.Identity.Client/OAuth2/DeviceAuthJWTResponse.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    [JsonObject]
+    internal class DeviceAuthPayload
+    {
+        public DeviceAuthPayload(string audience, string nonce)
+        {
+            this.Nonce = nonce;
+            this.Audience = audience;
+            var timeSpan = (DateTime.UtcNow - new DateTime(1970, 1, 1, 0, 0, 0));
```
henrik-me, line 40: >new DateTime(1970, 1, 1, 0, 0, 0) [](start = 46, length = 33)

would it make sense to have this as a named lazy var (or perhaps even static)?

src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
trwalke: Not needed. removing

src/client/Microsoft.Identity.Client/OAuth2/DeviceAuthJWTResponse.cs
trwalke: yes it does, ill change

tests/Microsoft.Identity.Test.Unit.net45/OAuthClientTests.cs
```diff
@@ -124,6 +127,65 @@ public void OAuthClient_FailsWithServiceExceptionWhenResponseIsHttpNotFound()
                 });
         }
 
+#if DESKTOP
+        [TestMethod]
+        public void PKeyAuthSuccsesResponseTest()
```
henrik-me, line 132: >PKeyAuthSuccsesResponseTest() [](start = 20, length = 29)

no asserts?

Issue #1552, opened 2019-12-19
[Feature Request] Use "mobile" configuration for iOS Xamarin embedded webview


henrik-me:
Apple’s iPadOS (iOS 13 for iPad) switched to use macOS user-agent for all browsers by default.
This has caused some disruption in conditional access scenarios that depend on the User-Agent string. 

For UIWebView no changes are required

**To use Mobile user-agent with WKWebView when running on iPadOS see below sample code:**
```obj-c
WKWebViewConfiguration *config = [WKWebViewConfiguration new];
config.defaultWebpagePreferences.preferredContentMode = WKContentModeMobile;
WKWebView *webView = [[WKWebView alloc] initWithFrame:rootView.frame configuration:config];
```
Note: To explore the same change for ADAL

bgavrilMS:
@henrik-me - do you have more details around this? What is the scenario that you are talking about?

jennyf19:
released in 4.8.1

PR #1569, opened 2020-01-09
fix for wkwebview on ipad

jennyf19:
Fix for: #1552
Test: none


henrik-me:
Automation: should an e2e test for ipad be added to our test suite? I believe today we are only testing on iPhone's

bgavrilMS:
The commit message is not descriptive enough :)

henrik-me:
hmm, I think it's one of the more precise ones I've seen for a long time in this repo

---
In reply to: [574181314](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1569#issuecomment-574181314) [](ancestors = 574181314)

PR #1589, opened 2020-01-21
fix for null ref during interactive flow w/out android broker

jennyf19:
@bgavrilMS this resolves the null ref during interactive without broker. i know you want to do more refactoring, but wasn't sure if we wanted a quick fix or something more long term. here is a quick fix.
- tested w/system browser and embedded webview. I don't have the v2 android broker on my device, so didn't test any broker related flows.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -93,7 +93,7 @@ namespace Microsoft.Identity.Client.Platforms.Android
                 _logger.Verbose("User is specified for silent token request. Starting silent broker request");
                 string silentResult = _brokerHelper.GetBrokerAuthTokenSilently(brokerPayload, _activity);
                 _androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
-                _readyForResponse.Release();
+                _readyForResponse?.Release();
                 return;
             }
             else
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -93,7 +93,7 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
                 _logger.Verbose("User is specified for silent token request. Starting silent broker request");
                 string silentResult = _brokerHelper.GetBrokerAuthTokenSilently(brokerPayload, _activity);
                 _androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
-                _readyForResponse.Release();
```
bgavrilMS: This may fix the immediate problem, but the architectural issue remains - why do we end up calling broker code when the broker isn't active i.e. `.WithBroker(false)` ?

jennyf19: yes...agreed and i think separating out things is a good idea. like i mentioned earlier, this is a temporary fix and we should consider something more long term. but that can happen now. wasn't sure on the urgency of this.

jennyf19: agreed...we are doing it automatically in the `SetAuthenticationContinuatoinEventArgs`. will look into a fix there as well.

Issue #1586, opened 2020-01-21
[Bug] Xamarin: Make library linker-safe


tipa:
**Which Version of MSAL are you using ?**
MSAL 4.8.0

**Platform**
Xamarin.Android & Xamarin.iOS

**Repro**
Enable full linking in Xamarin project
iOS:
![image](https://user-images.githubusercontent.com/2863198/72796742-271ce500-3c40-11ea-8b0e-8f85c9c7f32b.png)
Android:
![image](https://user-images.githubusercontent.com/2863198/72796763-30a64d00-3c40-11ea-94ea-6a46bece5cc9.png)

**Expected behavior**
Library works without any (or minimal) adjustments needed

**Actual behavior**
Usage of the library fails (see error below) because needed classes (or their constructors/methods) were removed/linked away. Affected classes are, amongst others, `Microsoft.Identity.Client.Core.IdToken`, `Microsoft.Identity.Client.Instance.TenantDiscoveryResponse`, `Microsoft.Identity.Client.Instance.Discovery.InstanceDiscoveryResponse`.
Previously (with v4.7.1) I only had to manually add class `Microsoft.Identity.Client.BrowserTabActivity` to the [LinkDescription.xml](https://docs.microsoft.com/en-us/xamarin/cross-platform/deploy-test/linker), but with v4.8.0, many more classes need to be excluded from linking manually.

**Possible Solution**
It would be great if the SDK could be made linker-safe so that I wouldn't have to care and exclude classes manually. Afaik, this can be done by Attributes as described [here](https://docs.microsoft.com/en-us/xamarin/ios/deploy-test/linker) or [here](https://montemagno.com/optimizing-xamarin-apps-libraries-with-the-linker/)

**Additional context/ Logs / Screenshots**
> MSAL.Xamarin.iOS.4.8.0.0.MsalClientException: 
> ErrorCode: json_parse_failed
> Microsoft.Identity.Client.MsalClientException: Failed to parse the returned id token. ---> Microsoft.Identity.Json.JsonSerializationException: Unable to find a constructor to use for type Microsoft.Identity.Client.Core.IdToken. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'ver', line 1, position 7.
[...]

PR #1590, opened 2020-01-22
Linker

bgavrilMS:
tracking bug: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1586
Things that don't work now: json exception.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
+++ b/src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
@@ -4,10 +4,10 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
-#if ANDROID_BUILDTIME
+#if ANDROID
 using Android;
 #endif
-#if iOS_BUILDTIME
+#if iOS
 using Foundation;
 #endif
 
@@ -20,6 +20,6 @@ using Foundation;
 // COM, set the ComVisible attribute to true on that type.
 [assembly: ComVisible(false)]
 [assembly: CLSCompliant(true)]
-#if ANDROID_BUILDTIME || iOS_BUILDTIME
+#if ANDROID || iOS
 [assembly: LinkerSafe]
 #endif
--- a/src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
+++ b/src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
@@ -4,6 +4,12 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
+#if ANDROID
+using Android;
+#endif
+#if iOS
+using Foundation;
+#endif
 
 // Version and Metadata are set at build time from msbuild properties defined in the csproj
 
@@ -14,3 +20,6 @@ using System.Runtime.InteropServices;
 // COM, set the ComVisible attribute to true on that type.
 [assembly: ComVisible(false)]
 [assembly: CLSCompliant(true)]
+#if iOS || ANDROID
+[assembly: LinkerSafe]
+#endif
--- a/src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
+++ b/src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
@@ -4,10 +4,10 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
-#if ANDROID
+#if ANDROID_BUILDTIME
 using Android;
 #endif
-#if iOS
+#if iOS_BUILDTIME
 using Foundation;
 #endif
 
@@ -20,6 +20,6 @@ using Foundation;
 // COM, set the ComVisible attribute to true on that type.
 [assembly: ComVisible(false)]
 [assembly: CLSCompliant(true)]
-#if iOS || ANDROID
+#if ANDROID_BUILDTIME || iOS_BUILDTIME
 [assembly: LinkerSafe]
 #endif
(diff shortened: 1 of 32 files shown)


src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
```diff
@@ -4,6 +4,12 @@
 using System;
 using System.Reflection;
 using System.Runtime.InteropServices;
+#if ANDROID_BUILDTIME
```
bgavrilMS: Please use `if ANDROID` as we not hiding anything here.

bgavrilMS: This is probably the reason why the fix isn't working.

bgavrilMS:
Here's what I think it happening with the json exceptions:

1. The LINKER looks at objects like ClientInfo and thinks - their constructors are are not used anywhere, so I can remove them! This is because MSAL doesn't use those constructors directly, but Newtonsoft invokes them thorough reflection. But Newtonsoft can't deserialize if constructors are removed :)
2. I bet that the Linker knows about the old [DataContract] attributes but doesn't know about the new [JsonObject] attributes.

So the solution is either: 
1. to add the #if ANDROID || IOS [Preserve] attribute to ALL classes that we deserlialize, i.e. all classes that are decorated with  [JsonObject]. OR
2. Alternatively, we can add [assembly: LinkerSafe] to tell the linker to leave the entire assembly alone. 


If we want to play nice with Xamarin, we should go with [Preserve] attributes. This will allow Xamarin to remove unused code form MSAL and produce a small apk. If we want to be more future proof, [assembly: LinkerSafe] is the answer - this will tell Xamarin to leave MSAL alone and if we add a new object that we deserialize, then we wouldn't have to add the [Preseve] attribute as well. 

In light of the fact that MSAL does not use external dependencies, the size gains should be minimal, but we can check, i.e. check the size of an APK with each solution.

src/client/Microsoft.Identity.Client/Properties/AssemblyInfo.cs
trwalke: We tried doing this with #if ANDROID but that did not seem to work. The only thing that did work was adding preserve to each of the classes. We can remove this from the AssemblyInfo

bgavrilMS:
LGTM

tests/Microsoft.Identity.Test.Core.UIAutomation/MobileTestHelper.cs
```diff
@@ -280,10 +280,12 @@ public void VerifyResult(ITestController controller)
                 }
                 else if (result.Contains(CoreUiTestConstants.TestResultFailureMessage))
                 {
+                    controller.Tap("LogPage");
```
bgavrilMS: You don't need all those `else` clauses because you are either returning or throwing, which breaks the flow anyway.

bgavrilMS:
UI Automation failure shows that we have problems selecting some field... Runs in progress with this latest change and a change before all the serialization work. 

- Error while performing WaitForElement(Class("WKWebView").Css("#i0116"), "Timeout waiting for web element with css id: i0116", 00:00:30, 00:00:01, 00:00:01)Exception: System.TimeoutException: Timeout waiting for web element with css id: i0116  at Xamarin.UITest.Shared.WaitForHelper.WaitForAny[T] (System.Func`1[TResult] query, System.String timeoutMessage, System.Nullable`1[T] timeout, System.Nullable`1[T] retryFrequency, System.Nullable`1[T] postTimeout) [0x0001d] in <2a16c16730a54859bda72c6bc1c728f7>:0   at Xamarin.UITest.iOS.iOSApp+<>c__DisplayClass67_0.<WaitForElement>b__0 () [0x00054] in <2a16c16730a54859bda72c6bc1c728f7>:0   at Xamarin.UITest.Utils.ErrorReporting.With[T] (System.Func`1[TResult] func, System.Object[] args, System.String memberName) [0x0000e] in <2a16c16730a54859bda72c6bc1c728f7>:0 Failed to find UPN input. Attempting to click on UPN from select account screen23-01-2020 02:07:16.211 +01:00 - 57569 - Waiting for element matching Class("WKWebView").Css("#idlab@msidlab4.onmicrosoft.com").23-01-2020 02:07:46.500 +01:00 - 87856 - Error while performing WaitForElement(Class("WKWebView").Css("#idlab@msid

hig-dev:
I would not use something like:
```
#if iOS
using Foundation;
#endif
#if ANDROID
using Android.Runtime;
#endif
```

Instead declare a PreserveAttribute class and use that in the code:

```
public sealed class PreserveAttribute : System.Attribute {
    public bool AllMembers;
    public bool Conditional;
}
```

Then you can use it like:

```
    [PreserveAttribute(AllMembers = true)]
    public class ExampleClassToPreserve
    {
     ....
    }
```

This is a much cleaner solution.
Source: https://docs.microsoft.com/en-us/xamarin/ios/deploy-test/linker?tabs=macos#controlling-the-linker

tests/devapps/XForms/XForms/TestPage.xaml.cs
```diff
@@ -311,7 +311,13 @@ private void CreateExceptionMessage(Exception exception)
                 acquireResponseLabel.Text = "Exception - " + exception.Message;
             }
 
-            Console.WriteLine(exception);
+            Console.WriteLine(string.Format(CultureInfo.InvariantCulture, "Exception -\nMessage: {0}\nStack Trace: {1}",
```
bgavrilMS: If you just replace the entire block with 

```csharp
Console.WriteLine(exception);
```

it will be enough. The exception.ToString() method takes care of outputting the message, the stack trace and the inner exception details

PR #1591, opened 2020-01-22
[Do not merge yet] Broker refactoring proposal

trwalke:
I made a few changes to how MSAL processes responses on android to avoid running broker code unnecessarily.


src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -32,14 +32,16 @@ public static void SetAuthenticationContinuationEventArgs(int requestCode, Resul
...
 
-                    AndroidBroker.SetBrokerResult(data, (int)resultCode);
-
-
             AuthorizationResult authorizationResult;
             if (data.Action != null && data.Action.Equals("ReturnFromEmbeddedWebview", StringComparison.OrdinalIgnoreCase))
             {
                 authorizationResult = ProcessFromEmbeddedWebview(requestCode, resultCode, data);
             }
+            else if (!String.IsNullOrEmpty(data.GetStringExtra(BrokerConstants.BrokerResultV2)))
```
bgavrilMS: yes, this is looks much more like the root cause of the problem.

PR #1605, opened 2020-01-31
Broker Updates

trwalke:
This PR Enables broker, fixes an exception issue and reapplies an older PR that was lost due to the branch being deleted.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1591

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -26,11 +26,6 @@ namespace Microsoft.Identity.Client.Platforms.Android
     [AndroidNative.Runtime.Preserve(AllMembers = true)]
     internal class AndroidBroker : IBroker
     {
-        //Since the response for the broker interactive call is returned in the AuthenticationContinuationHelper, this class needs to send the response to 
-        //the broker and wait for the AuthenticationContinuationHelper to return the response to the broker thread. This sephamore is used to make the MSAL code wait
-        //for the response. The AuthenticationContinuationHelper will trigger the continueation of the broker thread once it sets the rsultEX value. 
-        //Because of this, resultEX is made to be static. However, there will only ever be one broker authentication happening at once.
-        private static SemaphoreSlim _readyForResponse = null;
         private static MsalTokenResponse _androidBrokerTokenResponse = null;
         //Since the correlation ID is not returned from the broker response, it must be stored at the beginning of the authentication call and reinjected into the response at the end.
         private static string _correlationId;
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -58,12 +52,10 @@ public async Task<MsalTokenResponse> AcquireTokenUsingBrokerAsync(Dictionary<str
...
             _androidBrokerTokenResponse = null;
             _correlationId = AndroidBrokerHelper.GetValueFromBrokerPayload(brokerPayload, BrokerParameter.CorrelationId);
-            //Need to disable warning for non awaited async call.
+
             try
             {
-#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
-                new TaskFactory().StartNew(() => AcquireTokenInternalAsync(brokerPayload));
-#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+                await Task.Run(() => AcquireTokenInternalAsync(brokerPayload)).ConfigureAwait(false);
```
trwalke: Exceptions were getting lost when I used TaskFactory().StartNew() so i had to revert
@bgavrilMS

bgavrilMS: Why not just `await AcquireTokenInteranalAsync` ?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -43,7 +38,6 @@ public AndroidBroker(CoreUIParent uiParent, ICoreLogger logger)
             _activity = uiParent?.Activity;
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _brokerHelper = new AndroidBrokerHelper(Application.Context, logger);
-            _readyForResponse = new SemaphoreSlim(0);
         }
 
         public bool CanInvokeBroker()
```
bgavrilMS: Can you test what happens if you invoke the broker twice? I.e. call AcquireTokenInteractive and then call it imediatly again, without completing the broker flow. This should be a test for both iOS and Android

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -74,11 +66,10 @@ public async Task<MsalTokenResponse> AcquireTokenUsingBrokerAsync(Dictionary<str
                     throw new MsalException(MsalError.AndroidBrokerOperationFailed, ex.Message, ex);
             }
 
-            await _readyForResponse.WaitAsync().ConfigureAwait(false);
```
bgavrilMS: Could you test what happens now if the broker is not installed but users configures WithBroker(true) ?

trwalke:
Manual testing resulted in errors so I will need to revert the async change and add the semephore back in. I spoke with henrik about this and he agrees that it is the best way to go for now. I am moving these changes here https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1610 because jenny has some changes that she made to broker code to fix a bug.

PR #1608, opened 2020-01-31
[In Progress - Not ready to merge] iOS/Android broker refactor

jennyf19:
fix null ref when .WithBroker(true) but broker not installed
fix wpj requirement from auth code

Mainly review these two classes:  `InteractiveRequest` and `BrokerInteractiveRequest` classes and `ExecuteAsync()` and `SendTokenRequestToBrokerAsync()` in the broker interactive request

**Test cases for iOS Broker:**
- Developer sets `.WithBroker(true)`
User logs in to app.
Broker is not installed on device.
User is presented with regular auth flow and signs in.

- WPJ requirement - broker not installed
User logs into app
User is presented with regular auth flow
Auth code is returned like this: `msauth://wpj?username=idlabmamca%40msidlab4.onmicrosoft.com&app_link=itms%3a%2f%2fitunes.apple.com%2fapp%2fazure-authenticator%2fid983156458%3fmt%3d8`
MSAL will open the broker in the AppStore
>This scenario does not work. ests does not have xamarin ios/android on it's list of supporting MAM, so you have to replace the auth code w/the above to test the experience, as we are currently presented with the wrong log-in screen

- WPJ requirement - broker installed
User logs into app
User is presented with regular auth flow
Auth code is returned like this: `msauth://wpj?username=idlabmamca%40msidlab4.onmicrosoft.com&app_link=itms%3a%2f%2fitunes.apple.com%2fapp%2fazure-authenticator%2fid983156458%3fmt%3d8`
User logs in via broker

- Interactive LogIn (no broker)
User logs in to app.
User is presented with regular auth flow and signs in.

**Test Cases for Android Broker (@trwalke to add them here):**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerInteractiveRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerInteractiveRequest.cs
@@ -7,7 +7,6 @@ using Microsoft.Identity.Client.Internal.Requests;
 using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.UI;
 using Microsoft.Identity.Client.Utils;
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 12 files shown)


src/client/Microsoft.Identity.Client/Internal/Broker/BrokerInteractiveRequest.cs
```diff
@@ -10,47 +10,85 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Identity.Client.Internal.Broker
 {
-    internal class BrokerInteractiveRequest
+    internal class BrokerInteractiveRequest : InteractiveRequest
```
bgavrilMS: TLDR; it should be cleaner to derive from `RequestBase`

What does RequestBase provide? 
- access to ServiceBundle, config params and request params
- scope validation
- logging and telememetry

and hooks for: 

- caching token response -> possibly used by broker request
- resolve authority 
- send request to /token endpoint with headers, body params -> not used by broker

bgavrilMS: When you derive from RequestBase, check what methods you should be overrideing and what methods you just need to call.

src/client/Microsoft.Identity.Client/Internal/Requests/InteractiveRequest.cs
```diff
@@ -26,11 +27,11 @@ internal class InteractiveRequest : RequestBase
 
         private readonly SortedSet<string> _extraScopesToConsent;
         private readonly IWebUI _webUi;
-        private AuthorizationResult _authorizationResult;
         private string _codeVerifier;
         private string _state;
         private readonly AcquireTokenInteractiveParameters _interactiveParameters;
         private MsalTokenResponse _msalTokenResponse;
+        internal AuthorizationResult AuthResult { get; private set; }
```
bgavrilMS: Let's keep this private. If needed, InteractiveRequest can pass it in the constructor of BrokerInteractiveRequest.

bgavrilMS:
The 2 interactive requests - WebUI / Broker should not know about each other. The current InteractiveRequest already does way too much. In order for an object to do less, another object must intervene and take responsibility - in this case an "orchestrator" type of object need to switch between WebUI / Broker requests. 

I will propose a PR and we can sync to discuss the details.

jennyf19:
@trwalke you might want to do your android fixes into @bgavrilMS 's branch or into master later...

Issue #1614, opened 2020-02-06
[Feature Request] Add app link support to MSAL broker on Xamarin Android


trwalke:
When the android broker is not installed on the device and MSAL tries to do an interactive sign in with an account with conditional access policies applied, MSAL should be able to parse the App link response from AAD and navigate to the broker application in the Android appstore.

trwalke:
released in 4.9.0
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.9.0

PR #1616, opened 2020-02-07
Adding app link support for MDM/MAM enrollment

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1614


src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -80,6 +80,19 @@ public async Task<MsalTokenResponse> AcquireTokenUsingBrokerAsync(Dictionary<str
 
         private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerPayload)
         {
+            if (brokerPayload.ContainsKey(BrokerParameter.BrokerInstallUrl))
+            {
+                _logger.Info("Android Broker - broker payload contains install url");
+
+                var appLink = AndroidBrokerHelper.GetValueFromBrokerPayload(brokerPayload, BrokerParameter.BrokerInstallUrl);
```
bgavrilMS: nit: not a big fan of this helper, a null is easier to track than an empty string when debugging.

PR #1623, opened 2020-02-11
use msauthv2 again

jennyf19:


jennyf19:
yes...i just pinged @bgavrilMS ...`msauthv3` is not a valid broker scheme, it's just used to detect which version of broker is installed on the app, because based on that, we have to send additional parameters. We can send `msauth` or `msauthv2`, and we are using `msauthv2`. Olga said they will improve the error messaging for this on their end, as there were no issues appearing in the logs. 

we were doing the right thing. cc: @jmprieur

bgavrilMS:
Thanks for following up @jennyf19 . 

@henrik-me @jmprieur - after we do a bit more testing, how do you feel about pushing a 4.9-preview version out there, for example EOD today? There are a few people who can help test this and we already have bug reports that we cannot repro.

PR #1627, opened 2020-02-12
[do not merge] dub keychain fix in broker key fix

jennyf19:
check for min version of ios 13 for embedded webview issue

>Will get iOS 10 device from iOS team to test at 11:30.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
@@ -15,7 +15,6 @@ namespace Microsoft.Identity.Client.Platforms.iOS
     {
         internal static byte[] GetRawBrokerKey(ICoreLogger logger)
         {
-            byte[] brokerKey = null;
             SecRecord record = new SecRecord(SecKind.GenericPassword)
             {
                 Generic = NSData.FromString(iOSBrokerConstants.LocalSettingsContainerName),
@@ -27,13 +26,14 @@ namespace Microsoft.Identity.Client.Platforms.iOS
             };
 
             NSData key = SecKeyChain.QueryAsData(record);
+            byte[] brokerKey;
             if (key == null)
             {
                 AesManaged algo = new AesManaged();
                 algo.GenerateKey();
                 byte[] rawBytes = algo.Key;
                 NSData byteData = NSData.FromArray(rawBytes);
-                record = new SecRecord(SecKind.GenericPassword)
+                SecRecord secRecord = new SecRecord(SecKind.GenericPassword)
                 {
                     Generic = NSData.FromString(iOSBrokerConstants.LocalSettingsContainerName),
                     Service = iOSBrokerConstants.BrokerKeyService,
@@ -44,7 +44,18 @@ namespace Microsoft.Identity.Client.Platforms.iOS
                     ValueData = byteData
                 };
 
-                var result = SecKeyChain.Add(record);
+                var result = SecKeyChain.Add(secRecord);
+
+                if (result == SecStatusCode.DuplicateItem)
+                {
+                    logger.Info(iOSBrokerConstants.FailedToSaveBrokerKey + result);
+                    result = SecKeyChain.Remove(record);
+                    if (result == SecStatusCode.Success)
+                    {
+                        result = SecKeyChain.Add(secRecord);
+                    }
+                }
+
                 if (result != SecStatusCode.Success)
                 {
                     logger.Info(iOSBrokerConstants.FailedToSaveBrokerKey + result);
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
@@ -54,10 +54,7 @@ namespace Microsoft.Identity.Client.Platforms.iOS.EmbeddedWebview
         {
             WKWebViewConfiguration wkconfg = new WKWebViewConfiguration() { };
 #if !IS_APPCENTER_BUILD
-            if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
-            {
-                wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
-            }
+            wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
 #endif
             _wkWebView = new WKWebView(View.Bounds, wkconfg)
             {
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
@@ -54,7 +54,10 @@ namespace Microsoft.Identity.Client.Platforms.iOS.EmbeddedWebview
         {
             WKWebViewConfiguration wkconfg = new WKWebViewConfiguration() { };
 #if !IS_APPCENTER_BUILD
-            wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
+            if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
+            {
+                wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
+            }
 #endif
             _wkWebView = new WKWebView(View.Bounds, wkconfg)
             {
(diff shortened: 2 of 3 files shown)


src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
```diff
@@ -44,7 +44,18 @@ internal static byte[] GetRawBrokerKey(ICoreLogger logger)
                     ValueData = byteData
                 };
 
-                var result = SecKeyChain.Add(record);
+                var result = SecKeyChain.Add(secRecord);
+
+                if (result == SecStatusCode.DuplicateItem)
```
jennyf19: if the above key is null, we try to add the record created earlier, but the two keys (Service and Account) are the same, so a DuplicateItem error is throw. If we try to do `SecStatusCode.Update(record, secRecord)`, there is a new error because secRecord has a few field -> ValueData, so it's best to just delete the original record and add the new one. I was able to repro this by setting the key to null.

src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
```diff
@@ -27,13 +26,14 @@ internal static byte[] GetRawBrokerKey(ICoreLogger logger)
             };
 
             NSData key = SecKeyChain.QueryAsData(record);
+            byte[] brokerKey;
             if (key == null)
             {
                 AesManaged algo = new AesManaged();
```
bgavrilMS: this is disposable so please use 

```csharp
using (AesManaged algo...) 
{
// do stuff
}
```

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
```diff
@@ -54,7 +54,10 @@ protected WKWebView PrepareWKWebView()
         {
             WKWebViewConfiguration wkconfg = new WKWebViewConfiguration() { };
 #if !IS_APPCENTER_BUILD
-            wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
+            if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
+            {
+                wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
```
bgavrilMS: I suggest that this change goes out in a different commit. No new review necessary, ship it!

src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
```diff
@@ -44,7 +44,18 @@ internal static byte[] GetRawBrokerKey(ICoreLogger logger)
                     ValueData = byteData
                 };
 
-                var result = SecKeyChain.Add(record);
+                var result = SecKeyChain.Add(secRecord);
+
+                if (result == SecStatusCode.DuplicateItem)
+                {
+                    logger.Info(iOSBrokerConstants.FailedToSaveBrokerKey + result);
+                    result = SecKeyChain.Remove(record);
```
henrik-me: >result = SecKeyChain.Remove(record); [](start = 20, length = 36)

we are only here if they key is null (after querying the data store). Will this really do anything for this case? Should this have been remove secRecord instead?

src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
henrik-me: there are more of those around. is this important? Generally I would say the GC can handle this one ? if we are afraid of running out of resources it may be a good idea to force dispose it, otherwise I would leave it to the GC

---
In reply to: [378471346](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1627#discussion_r378471346) [](ancestors = 378471346)

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
```diff
@@ -54,7 +54,10 @@ protected WKWebView PrepareWKWebView()
         {
             WKWebViewConfiguration wkconfg = new WKWebViewConfiguration() { };
 #if !IS_APPCENTER_BUILD
```
henrik-me: why keeping this?

jennyf19:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1628

src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
bgavrilMS: It is best practice to dispose of IDisposables, there are even static analyses rules around it (I'll look for one). There are 2 problems with not disposing of IDposable imo: 

1. GC can be slow. A second gen GC pass is invoked only about once per second. In a website running many requests per second, this can become a bottleneck.
2. Not all objects that implement IDisposable implement the entire pattern that protects against developers who forget to call `Dispose()`. If an IDisposable does not implement the full pattern, but it contains unmanaged resources, then that resource will not be cleaned up.

PS: there's a [static](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1063?view=vs-2019) analysis rule for the second point :). I'll look at enabling it.

henrik-me: Yeah, my point is. how needed is it regardless? it may implement IDisposable because they have non managed resources and that's the only way to free those resources. perhaps they are able to reuse those resources if we don't dispose the object. We have had this code around for years and to my knowledge no issues so far. I would not prioritize this very high

---
In reply to: [378479075](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1627#discussion_r378479075) [](ancestors = 378479075)

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
```diff
@@ -54,7 +54,10 @@ protected WKWebView PrepareWKWebView()
...
-            wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
+            if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
```
henrik-me: what is the default value? for different iOS versions? perhaps a note on that here or a link to get that info from apple?

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
jennyf19: [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1629](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1629) new PR here

---
In reply to: [378472216](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1627#discussion_r378472216) [](ancestors = 378472216)

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
jennyf19: this is only applicable in ios 13+

---
In reply to: [378509016](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1627#discussion_r378509016) [](ancestors = 378509016)

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
jennyf19: idk. i removed it in the other pr (link somewhere here)

---
In reply to: [378475327](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1627#discussion_r378475327) [](ancestors = 378475327)

src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
jennyf19: For some reason, the customer is getting a dubplicate keychain error here, so even though the key is null, there is still something in the keychain. i wasn't able to repro this naturally, so am waiting for customer response.

---
In reply to: [378472489](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1627#discussion_r378472489) [](ancestors = 378472489)

src/client/Microsoft.Identity.Client/Platforms/iOS/BrokerKeyHelper.cs
jennyf19: i'll make a work item for this. I agree that it's too much right now

---
In reply to: [378507050](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1627#discussion_r378507050) [](ancestors = 378507050,378479075)

PR #1632, opened 2020-02-13
Android Broker Fixes

trwalke:
Fixing threading issue in interactive broker flow on android.
Fixing null ref during silent broker flow on android.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -95,7 +95,18 @@ namespace Microsoft.Identity.Client.Platforms.Android
                 {
                     _logger.Verbose("User is specified for silent token request. Starting silent broker request");
                     string silentResult = await _brokerHelper.GetBrokerAuthTokenSilentlyAsync(brokerPayload, _activity).ConfigureAwait(false);
-                    s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
+                    if (!string.IsNullOrEmpty(silentResult))
+                    {
+                        s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
+                    }
+                    else
+                    {
+                        s_androidBrokerTokenResponse = new MsalTokenResponse
+                        {
+                            Error = MsalError.BrokerResponseReturnedError,
+                            ErrorDescription = "Failed to acquire token silently from the broker.",
+                        };
+                    }
                     return;
                 }
                 else
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -96,7 +96,6 @@ namespace Microsoft.Identity.Client.Platforms.Android
                     _logger.Verbose("User is specified for silent token request. Starting silent broker request");
                     string silentResult = await _brokerHelper.GetBrokerAuthTokenSilentlyAsync(brokerPayload, _activity).ConfigureAwait(false);
                     s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
-                    s_readyForResponse?.Release();
                     return;
                 }
                 else
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -59,7 +59,8 @@ namespace Microsoft.Identity.Client.Platforms.Android
             }
             catch (Exception ex)
             {
-                _logger.Error("Broker Operation Failed to complete.");
+                _logger.Error("Broker Operation Failed to complete. In order to perform brokered authentication," +
+                    " you need to ensure that you have installed either Intune Company Portal (Version 5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater).");
                 if (ex is MsalException)
                     throw;
                 else
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Identity.Client.Platforms.Android
             }
             catch (Exception ex)
             {
-                _logger.Error("Broker Operation Failed to complete. In order to perform brokered authentication," +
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -96,7 +96,6 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
                     _logger.Verbose("User is specified for silent token request. Starting silent broker request");
                     string silentResult = await _brokerHelper.GetBrokerAuthTokenSilentlyAsync(brokerPayload, _activity).ConfigureAwait(false);
                     s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
-                    s_readyForResponse?.Release();
```
bgavrilMS: I don't understand who releases the semaphore if it isn't done here?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -246,7 +246,7 @@ public async Task InitiateBrokerHandshakeAsync(Activity callerActivity)
         private Handler GetPreferredLooper(Activity callerActivity)
         {
             var myLooper = Looper.MyLooper();
-            if (myLooper != null && callerActivity.MainLooper != myLooper)
+            if (myLooper != null && callerActivity != null && callerActivity.MainLooper != myLooper)
```
bgavrilMS: This fix looks good to me.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -96,7 +96,6 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
                     _logger.Verbose("User is specified for silent token request. Starting silent broker request");
                     string silentResult = await _brokerHelper.GetBrokerAuthTokenSilentlyAsync(brokerPayload, _activity).ConfigureAwait(false);
                     s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
```
bgavrilMS: I don't think this covers the case where the broker returns an error result.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
bgavrilMS: @jennyf19 - this piece of code is somewhat related to threading. The "looper" is smth equivalent to the UI / non-UI thread in WPF / WinForm apps. This piece of code tries to use the "looper" of the calling activity, or the main looper. The bug is that callerActivity seems that it can be null.

jennyf19: @bgavrilMS yeah, i remember the looper.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -96,7 +96,6 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
...
-                    s_readyForResponse?.Release();
                     return;
```
bgavrilMS: @jennyf19 - it took me a while to figure out what's going on. Suffice to say that we use a semaphore like a lock. If you call "wait", the lock becomes "busy". If you call "release", it becomes available. Waiting on a busy lock "blocks" until the lock becomes available again. The semaphore starts as "busy".

When doing a Silent auth, there is no need really for a lock, because the broker just returns the result. So the bug was that the lock got released, but there's nothing waiting on that lock. A subsequent interactive request would fail because at line 142 (below) would simply continue, since the lock is not busy.

`await s_readyForResponse.WaitAsync()`

bgavrilMS: All this manual locking is very tedious and bug prone. Tasks are a much better abstraction that can be used to do multi-threading and asyncronous.

jennyf19: i see. thanks for the explanation @bgavrilMS

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
bgavrilMS: I had no idea such a thing exists. Funky name :)

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
trwalke: The error case is being handled by the CreateMsalTokenResponseFromResult method. we check to see if we get an account back from the account manager and if we dont we log an error. however, it wouldnt hurt to also add that error to the MSAL token response.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -95,7 +95,18 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
                 {
                     _logger.Verbose("User is specified for silent token request. Starting silent broker request");
                     string silentResult = await _brokerHelper.GetBrokerAuthTokenSilentlyAsync(brokerPayload, _activity).ConfigureAwait(false);
-                    s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(silentResult);
+                    if (!string.IsNullOrEmpty(silentResult))
```
bgavrilMS: This is not what I meant but it is a good catch, as returning null would probably create problems downstream.

What I meant is that `CreateMsalTokenResponseFromResult` only handles the happy path when the broker returns a successful result. If the result has an error, you need to extract it, like at line 170. Preferably without having to duplicate code.

Also, this method is now too big, please see if you can split it into smaller pieces.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
trwalke: you are exactly right @bgavrilMS

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
bgavrilMS: I see now looking upstream that you return NULL if the broker does not return a successfuil response. Would be better if we can extract the actual error like you do with in the interactive broker flow. If possible.

PR #1649, opened 2020-02-19
Resolving null ref in IAccount silent flow for broker.

trwalke:


src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -58,6 +58,12 @@ private async Task<MsalTokenResponse> SendAndVerifyResponseAsync()
 
         internal void CreateRequestParametersForBroker()
         {
+            if (string.IsNullOrEmpty(_silentParameters.LoginHint) && string.IsNullOrEmpty(_silentParameters.Account.Username))
```
bgavrilMS: How can this happen? `AcquireTokenSilent` either takes an IAccount or a LoginHint. Can Account.Username ever be null?

bgavrilMS: Oh, yes, it can be null... https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7a7b5a1c6d3769740a10afb83184a2c91befc324/src/client/Microsoft.Identity.Client/IAccount.cs#L20

bgavrilMS: What does Android team do here? This experience of throwing an exception is not great - if MSAL were to manage its own cache (instaed of going to the broker), it would be able to get you a token even if IAccount.Username is null. Tokens are indexed by home_account_id, not by username. Is there a way to pass home_account_id to broker?

trwalke: Android bypasses its own cache for silent requests. Also, the broker requires a username to look up the account in the account manager. I dont think there is a way to use the home account id alone

trwalke: We can follow this release up with 4.9.1 to handle this more gracefully

jennyf19: I might need to see the end to end of this to understand it, but shouldn't you throw an MsalUiRequiredException here, with the fancy error message, because what/how will the developer address this error? how will they know the login hint or username if there is no user and they are doing a silent call at app start up (like we recommend)...does that make sense?

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -285,6 +285,7 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
             "You can modify the configuration in the application registration portal. See https://aka.ms/msal-net-invalid-client for details. ";
         public const string SSHCertUsedAsHttpHeader = "MSAL was configured to request SSH certificates from AAD, and these cannot be used as an HTTP authentication header. Developers are responsible for transporting the SSH certificates to the target machines.";
         public const string BrokerApplicationRequired = "Installation of broker failed. The broker application must be installed to continue authentication";
+        public const string NoUPNForSilentAuth = "Unable to query the android account manager because no username is provided in the silent authentication request. Please try using AcquireTokenSilent(IEnumerable<string> scopes, string loginHint). See https://aka.ms/Brokered-Authentication-for-Android for details";
```
jennyf19: nit: suggestion: `NoUpnForSilentBrokerAuth`

And, this can be because no username or login hint was passed correct? like both were null?

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
bgavrilMS: So, if account.Username is null, you cannot perform `AcquireTokenSilent` via broker ever. It might even be better to try to look in MSAL's cache, it's quite possible that you have a token. 

But anyway, we're not designing the feature, I'm just trying to understand what the Android folks are doing - since they designed the feature. Do you know if they also throw / fail if loginHint is null and Username is null?

trwalke: @bgavrilMS , you are correct, the broker can use the "home_account_id" and the "local_account_id" to look for the account, I need to change the conditional in the android broker to look for "home_account_id" and "local_account_id" instead of just the username.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -159,7 +166,14 @@ public async Task<string> GetBrokerAuthTokenSilentlyAsync(IDictionary<string, st
 
         public void GetBrokerAccountInfo(IDictionary<string, string> brokerPayload, Activity callerActivity)
         {
-            var loginHint = GetValueFromBrokerPayload(brokerPayload, BrokerParameter.LoginHint);
+            string loginHint = string.Empty;
+            if (string.IsNullOrEmpty(GetValueFromBrokerPayload(brokerPayload, BrokerParameter.LoginHint)))
+            {
+                _logger.Error(MsalErrorMessage.NoUPNOrAccountIDForSilentAuth);
+                throw new MsalClientException(MsalError.NoUsernameOrAccountIDProvidedForSilentAndroidBrokerAuthentication, MsalErrorMessage.NoUPNOrAccountIDForSilentAuth);
```
trwalke: @bgavrilMS it should not be possible to have both the username and home account id to be null correct? what do you think i should throw here

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -429,6 +429,11 @@ public static class MsalError
         /// </summary>
         public const string AndroidBrokerOperationFailed = "android_broker_operation_failed";
 
+        /// <summary>
+        /// Unable to query the android account manager because no username or home account id is provided in the silent authentication request.
```
pmaytak: Capital 'Android', capital 'ID'?

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -285,6 +285,7 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
             "You can modify the configuration in the application registration portal. See https://aka.ms/msal-net-invalid-client for details. ";
         public const string SSHCertUsedAsHttpHeader = "MSAL was configured to request SSH certificates from AAD, and these cannot be used as an HTTP authentication header. Developers are responsible for transporting the SSH certificates to the target machines.";
         public const string BrokerApplicationRequired = "Installation of broker failed. The broker application must be installed to continue authentication";
+        public const string NoUPNOrAccountIDForSilentAuth = "Unable to query the android account manager because no username or home account id is provided in the silent authentication request.";
```
pmaytak: Same as previous comment.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
bgavrilMS: Yes, that is a correct. ATS requies the user to pass in a login hint or an IAccount.

jennyf19: >MsalClientException [](start = 26, length = 19)

I guess i'm still confused here. so if there is no login hint, we throw an exception, but how does the developer resolve the issue? if they are doing 
try { ATSilent() } catch(MsalUiRequiredException ){ ATInteractive() }

They won't catch this error, unless later we are bubbling it up as an MsalUiRequiredException. Or is it required to provide a login hint w/the silent flow? if so, we should check earlier and fail if it's not provided.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -159,7 +166,14 @@ public async Task<string> GetBrokerAuthTokenSilentlyAsync(IDictionary<string, st
...
+            }
+
+            loginHint = GetValueFromBrokerPayload(brokerPayload, BrokerParameter.LoginHint);
```
jennyf19: you're getting this value twice...maybe move this earlier and if loginHint is null, then go through the if statement above, so you're not doing this twice if there is a login hint provided.

jennyf19:
>                 throw new MsalException(MsalError.NoAndroidBrokerAccountFound, "Please add the selected account to the broker");

sorry...this error message is confusing. does the customer add the account to the broker? how? shouldn't the broker open and they add their account? 

---
Refers to: src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs:191 in 94291cb. [](commit_id = 94291cbb45652fbd7cf58912f3282abbce85a69a, deletion_comment = False)

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
jennyf19: also, this would be a breaking change, right?

---
In reply to: [382131373](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#discussion_r382131373) [](ancestors = 382131373)

trwalke:
> > ```
> >             throw new MsalException(MsalError.NoAndroidBrokerAccountFound, "Please add the selected account to the broker");
> > ```
> 
> sorry...this error message is confusing. does the customer add the account to the broker? how? shouldn't the broker open and they add their account?
> 
> Refers to: src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs:191 in [94291cb](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/94291cbb45652fbd7cf58912f3282abbce85a69a). [](commit_id = [94291cb](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/94291cbb45652fbd7cf58912f3282abbce85a69a), deletion_comment = False)

That depends on the prompt behavior. if the user has login as the prompt then broker will start the sign in flow. if they chose no prompt then it will do nothing.

jennyf19:
so to use broker silent flow, they have to pass a loginhint or Prompt.ForceLogin? because silent flow doesn't take any prompts, so how would they pass that in?

---
In reply to: [589201616](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#issuecomment-589201616) [](ancestors = 589201616)

trwalke:
> so to use broker silent flow, they have to pass a loginhint or Prompt.ForceLogin? because silent flow doesn't take any prompts, so how would they pass that in?
> 
> In reply to: [589201616](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#issuecomment-589201616) [](ancestors = 589201616)

@jennyf19 you are correct, I misspoke earlier. if they are using the silent flow the user needs to pass in either a login hint or an Iaccount object with the homeAccountID. If they do not satisfy these requirements we should throw a UI required exception. Also, if they do provide this information and the broker simply does not have an account then we should throw a uirequired exception as well. Ill update to make sure this happens.

jennyf19:
sounds good. was helpful to sync off-line too. :) 

---
In reply to: [589216558](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#issuecomment-589216558) [](ancestors = 589216558)

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
jennyf19: should you just be checking for a null account? if the account is null, you can't have the home account id, right? so just check for a null account. in other words, can an account have an environment value, but no username and no home account id?

---
In reply to: [381623173](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#discussion_r381623173) [](ancestors = 381623173)

trwalke: Discussed offline, adding comment to better explain the scenario here

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
trwalke: Discussed offline. removing exception

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
trwalke: Yes. I modified this a little and now it is used in a msaluirequired exception instead of a client exception

jennyf19: for this error message, the only time they'd get this is if using their own IAccount object, which might not have all the stuff in it...can you add an aka.ms link or have something more actionable? like what should the developer do to mitigate this or how did this happen in the first place? thx.

---
In reply to: [382296441](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#discussion_r382296441) [](ancestors = 382296441)

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -68,6 +74,8 @@ internal void CreateRequestParametersForBroker()
             string extraQP = string.Join("&", _authenticationRequestParameters.ExtraQueryParameters.Select(x => x.Key + "=" + x.Value));
             BrokerPayload.Add(BrokerParameter.ExtraQp, extraQP);
             BrokerPayload.Add(BrokerParameter.ExtraOidcScopes, BrokerParameter.OidcScopesValue);
+            BrokerPayload.Add(BrokerParameter.HomeAccountId, _silentParameters.Account?.HomeAccountId?.Identifier);
```
jennyf19: >? [](start = 101, length = 1)

you don't need the null check here anymore as you're already checking above for HomeAccountId, it should be there

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -88,12 +96,12 @@ internal void ValidateResponseFromBroker(MsalTokenResponse msalTokenResponse)
             {
                 _authenticationRequestParameters.RequestContext.Logger.Info(
                     LogMessages.ErrorReturnedInBrokerResponse(msalTokenResponse.Error));
-                throw new MsalServiceException(msalTokenResponse.Error, MsalErrorMessage.BrokerResponseError + msalTokenResponse.ErrorDescription);
+                throw new MsalUiRequiredException(msalTokenResponse.Error, MsalErrorMessage.BrokerResponseError + msalTokenResponse.ErrorDescription);
```
jennyf19: why are these MsalUiRequiredExceptions? you should still throw serviceException right? can the developer do anything to mitigate these types of errors?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -153,13 +160,21 @@ public async Task<string> GetBrokerAuthTokenSilentlyAsync(IDictionary<string, st
...
 
         public void GetBrokerAccountInfo(IDictionary<string, string> brokerPayload, Activity callerActivity)
         {
-            var loginHint = GetValueFromBrokerPayload(brokerPayload, BrokerParameter.LoginHint);
+            string loginHint = GetValueFromBrokerPayload(brokerPayload, BrokerParameter.LoginHint);
+            if (string.IsNullOrEmpty(loginHint))
+            {
+                //While it is very unlikely that we end up with both the homeAccountID and the username equal to null
+                //We should still notify the user via log message if this were to occur.
+                _logger.Error(MsalErrorMessage.NoUPNOrAccountIDForSilentBrokerAuth);
```
jennyf19: >MsalErrorMessage.NoUPNOrAccountIDForSilentBrokerAuth [](start = 30, length = 52)

about this error message, is loginHint actually the homeAccountID and/or UPN? because i thought we were getting those from the IAccount object and login hint was given by the developer, so should this error message actually be .NoLoginHintProvided?

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
jmprieur: @trwalke : if the response from the broker is an error that can be solved by the user re-signing-in (or re consenting), then it's fine to throw an `MsalUiRequiredException`
**otherwise**, if this is an error because MSAL does something wrong we should **not** throw an `MsalUiRequiredException` because this might cause an infinite loop.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -153,13 +160,21 @@ public async Task<string> GetBrokerAuthTokenSilentlyAsync(IDictionary<string, st
...
+            }
+
+            loginHint = GetValueFromBrokerPayload(brokerPayload, BrokerParameter.LoginHint);
```
jennyf19: > loginHint  [](start = 11, length = 11)

this can be removed now because you have the value in line 169

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -153,13 +160,21 @@ public async Task<string> GetBrokerAuthTokenSilentlyAsync(IDictionary<string, st
                 }
             }
 
-            _logger.Info("Android Broker AddAccount didn't return any results. ");
+            _logger.Info("Android Broker didn't return any results.");
```
jmprieur: do we want to be more precise (to the silent authentication request) ?

jennyf19:
i think i need to walk through this w/my android device, but don't want to block the PR either. if you want to run a build from this branch for testing, that works too. 

---
In reply to: [589220763](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#issuecomment-589220763) [](ancestors = 589220763,589216558)

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
henrik-me: +1: to the aka.ms link!

---
In reply to: [382302642](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#discussion_r382302642) [](ancestors = 382302642,382296441)

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
henrik-me: tracking issue?

---
In reply to: [381614475](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#discussion_r381614475) [](ancestors = 381614475)

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
trwalke: @jmprieur Depends on what the error message is. the broker may return a error if either its cache is not populated with a token or if the service returns an error. However, I just realized that I should perform a check on the account manager before sending the request to the broker to check if we have an account in the broker cache. once I do that, only service exceptions will be caught here. ill flip these back

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -58,6 +58,12 @@ private async Task<MsalTokenResponse> SendAndVerifyResponseAsync()
 
         internal void CreateRequestParametersForBroker()
         {
+            if (string.IsNullOrEmpty(_silentParameters.LoginHint) && string.IsNullOrEmpty(_silentParameters.Account?.HomeAccountId?.Identifier))
+            {
+                throw new MsalUiRequiredException(MsalError.NoUsernameOrAccountIDProvidedForSilentAndroidBrokerAuthentication, MsalErrorMessage.NoUPNOrAccountIDForSilentBrokerAuth);
+            }
```
henrik-me: which tests has been added to ensure we will explore this scenario?

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -68,6 +74,8 @@ internal void CreateRequestParametersForBroker()
             string extraQP = string.Join("&", _authenticationRequestParameters.ExtraQueryParameters.Select(x => x.Key + "=" + x.Value));
             BrokerPayload.Add(BrokerParameter.ExtraQp, extraQP);
             BrokerPayload.Add(BrokerParameter.ExtraOidcScopes, BrokerParameter.OidcScopesValue);
+            BrokerPayload.Add(BrokerParameter.HomeAccountId, _silentParameters.Account?.HomeAccountId?.Identifier);
+            BrokerPayload.Add(BrokerParameter.LocalAccountId, _silentParameters.Account?.HomeAccountId?.ObjectId);
```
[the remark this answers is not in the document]
trwalke: Yes, it is the oid. this requirement is set by the broker as a required parameter for silent requests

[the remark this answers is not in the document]
henrik-me: What is the local account id needed for?

---
In reply to: [382305518](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#discussion_r382305518) [](ancestors = 382305518)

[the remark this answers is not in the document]
trwalke: Let me speak with android about how they handle guest accounts

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
henrik-me: Are these going to be service exceptions again? If not are these causing a potential regression for the iOS broker which has already shipped?

---
In reply to: [382314137](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#discussion_r382314137) [](ancestors = 382314137)

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
trwalke: There are log messages beforehand that indicate that the silent flow has started but it would not hurt to be more precise here.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
henrik-me: Is this ensuring that there is no null ref exception thrown in iOS as well as Android? and thus people will now get this instead of a null ref? Could this be thrown for other reasons? If so has iOS been fully regression tested?

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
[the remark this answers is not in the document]
trwalke: @henrik-me I believe they are using the local account id as a key to look up the account in their cache

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
trwalke: The login hint is considered the username for the broker so we can use either the loginhint or the UPN from the account object. So for the broker silent requests. we need to have either the UPN(which may come from the login hint) or the homeaccount ID

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
trwalke: This is only for android. iOS does not have a silent broker flow and it simply checks the cache.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
trwalke: These will be flipped back. also, ios does not support silent broker auth so we are good here

jennyf19:
Ok...i ran through this. I would just add this check earlier for a null homeAccountId, so that way, if it's not there, we throw earlier and don't go through all the broker code.

```cSharp
protected override void Validate()
        {
            base.Validate();
            if (Parameters.Account == null && string.IsNullOrWhiteSpace(Parameters.LoginHint))
            {
                throw new MsalUiRequiredException(
                    MsalError.UserNullError, 
                    MsalErrorMessage.MsalUiRequiredMessage, 
                    null, 
            }
            else if( Parameters.Account.HomeAccountId == null && string.IsNullOrWhiteSpace(Parameters.LoginHint)) // Android broker specific case
            {
                throw new MsalUiRequiredException(
                    MsalError.UserNullError,
                    MsalErrorMessage.MsalUiRequiredMessage,
                    null,
            }
```
        }

---
In reply to: [589399757](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#issuecomment-589399757) [](ancestors = 589399757,589220763,589216558)

jennyf19:
also, i tried w/no account and a loginHint and we get back a "[]" account, but then addAccount fails and it's surfaced as an MsalUiRequiredException, so i think that makes sense.

---
In reply to: [589502434](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1649#issuecomment-589502434) [](ancestors = 589502434,589399757,589220763,589216558)

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -58,6 +58,12 @@ private async Task<MsalTokenResponse> SendAndVerifyResponseAsync()
 
         internal void CreateRequestParametersForBroker()
         {
+            if (string.IsNullOrEmpty(_silentParameters.LoginHint) && string.IsNullOrEmpty(_silentParameters.Account?.HomeAccountId?.Identifier))
```
bgavrilMS: This should never happen - we should always have either a LoginHint or an IAccount. 
If the user decides to pass in their own object that implements IAccount (bummer), and their implemenation has HomeAccountId null, then we have to throw. Best to do that in SilentRequest.cs - there's a Validate method there.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
trwalke: This exception can also be thrown if the account is not present in the cache

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
trwalke: removing check here

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
trwalke: didnt see that method in silentrequest but it is in the acquireTokenSilentParameterBuilder. moved it there

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -173,29 +172,44 @@ public void GetBrokerAccountInfo(IDictionary<string, string> brokerPayload, Acti
...
-                        brokerPayload.Add(BrokerParameter.LocalAccountId, accountData[BrokerResponseConst.LocalAccountId].ToString());
-                        var acc = account;
+                        _logger.Info("Found broker account in Android account manager.");
                         return;
                     }
                 }
             }
+
+            _logger.Info("The requested account does not exist in the Android account manager.");
+            throw new MsalUiRequiredException(MsalError.NoAndroidBrokerAccountFound, MsalErrorMessage.NoAndroidBrokerAccountFound);
```
trwalke: This method will now look up the account information provided by the request and verify that the account exists in the account manager. if these checks fail then we know for certain the broker does not have the account in its cache so we can throw a ui required exception without sending any requests to the broker.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -68,7 +69,9 @@ internal void CreateRequestParametersForBroker()
             string extraQP = string.Join("&", _authenticationRequestParameters.ExtraQueryParameters.Select(x => x.Key + "=" + x.Value));
             BrokerPayload.Add(BrokerParameter.ExtraQp, extraQP);
             BrokerPayload.Add(BrokerParameter.ExtraOidcScopes, BrokerParameter.OidcScopesValue);
-            BrokerPayload.Add(BrokerParameter.LoginHint, _silentParameters.LoginHint);
+            BrokerPayload.Add(BrokerParameter.HomeAccountId, _silentParameters.Account?.HomeAccountId?.Identifier);
+            BrokerPayload.Add(BrokerParameter.LocalAccountId, _silentParameters.Account?.HomeAccountId?.ObjectId);
+            BrokerPayload.Add(BrokerParameter.Username, !string.IsNullOrEmpty(_silentParameters.Account?.Username) ? _silentParameters.Account?.Username : _silentParameters.LoginHint);
```
trwalke: Login hint and username are interchangeable here. Also updating the broker parameter to username.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -179,6 +179,14 @@ protected override void Validate()
                     null, 
                     UiRequiredExceptionClassification.AcquireTokenSilentFailed);
             }
+            else if (Parameters.Account.HomeAccountId == null && string.IsNullOrWhiteSpace(Parameters.LoginHint)) // Android broker specific case
```
bgavrilMS: 1. Parameters.Account can be null, possible null ref exception there. 
2. Use `IsNullOrEmpty` to check nullity of strings
3. `else` is not required, previous block is guaranteed to exit 
4. This is not an Android specific case, it's valid for non-broker scenarios as well. 

`if (string.IsNullOrWhiteSpace(Paramters.Account?.HomeAccountId) && string.IsNullOrWhiteSpace(Paramters.LoginHint))`

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -173,29 +172,44 @@ public void GetBrokerAccountInfo(IDictionary<string, string> brokerPayload, Acti
...
             if (!string.IsNullOrEmpty(accounts))
             {
                 dynamic authResult = JArray.Parse(accounts);
 
                 foreach (JObject account in authResult)
                 {
                     var accountData = account[BrokerResponseConst.Account];
-                    if ((accountData[BrokerResponseConst.UserName]).ToString() == loginHint)
+
+                    if ((accountData[BrokerResponseConst.UserName]).ToString() == username)
```
bgavrilMS: Do not compare strings like this. Use string.Equals please.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -173,29 +172,44 @@ public void GetBrokerAccountInfo(IDictionary<string, string> brokerPayload, Acti
...
+                    {
+                        brokerPayload[BrokerParameter.HomeAccountId] = accountData[BrokerResponseConst.HomeAccountId].ToString();
+                        brokerPayload[BrokerParameter.LocalAccountId] = accountData[BrokerResponseConst.LocalAccountId].ToString();
+                        _logger.Info("Found broker account in Android account manager.");
```
bgavrilMS: Please make this log message distinct from the one at line 205

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
bgavrilMS: Please rename the method and add a comment so that it is clear that it is only releavent for Silent requests. Non-silent request logic should not throw this exception.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
pmaytak: Why? '==' operator calls Equals to do ordinal comparison anyway. Unless there are consequences of Xamarin translating it differently.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
trwalke: 1. fixed
2. HomeAccountId is not a string. Parameters.LoginHint is a string but i am checking that with IsNullOrEmpty
3. removed
4. removed comment

Issue #1639, opened 2020-02-18
[Bug] WithClaims should bypass token cache


bgavrilMS:
**Which Version of MSAL are you using ?**
Note that to get help, you need to run the latest version. Preview version are also ok. 
For ADAL, please log issues to https://github.com/AzureAD/azure-activedirectory-library-for-dotnet
4.8.1

**Platform**
all

**What authentication flow has the issue?**
AquireTokenSilent

**Repro**

```csharp
               await app
                    .Silent(TestConstants.s_scope)
                    .WithClaims(TestConstants.Claims)```
```

**Expected behavior**
An AT token from the cache MUST not be returned. The refresh_token grant must be performed so that an AT with those claims is returned.

Cache is bypassed by ADAL and by MSAL for Android 

**Actual behavior**
Cache is not bypassed

Part of the work for Client Capabilities https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1545

jmprieur:
wow @bgavrilMS : is the claims a key of the cache? how does it work today: we should see infinite loops?

bgavrilMS:
claims is not a key of the cache. Today if you pass Claims to AcquierTokenSilent, it would return you an access token that is not tied to those claims. So Claims only works with `AcquireTokenInteractive` really.

jmprieur:
Oh, yes that makes sense. today we get a claims challenge and call AcquireTokenInteractive.
Good point for client capabilities;

trwalke:
released in 4.9.0
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.9.0

PR #1652, opened 2020-02-20
Claims bypass the token cache

bgavrilMS:
Fix for #1639


tests/Microsoft.Identity.Test.Unit.net45/CoreTests/OAuth2Tests/ClaimsTest.cs
```diff
@@ -2,19 +2,256 @@
...
+                var account = (await app.GetAccountsAsync().ConfigureAwait(false)).Single();
+                var result = await app.AcquireTokenSilent(TestConstants.s_scope, account)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+            }
+        }
+
+
+        [TestMethod]
+        public async Task ClaimsAreSentTo_AuthorizationEndpoint_And_TokenEndpoint_Async()
```
bgavrilMS: The rest of these tests were just moved

tests/devapps/NetFxConsoleTestApp/Program.cs
```diff
@@ -20,6 +20,22 @@ namespace NetFx
 {
     public class Program
     {
+        private const string Claims = @" {
+   ""userinfo"":
```
jmprieur: @bgavrilMS : is this the way customers can today request claims? (without a specific API) ?

bgavrilMS: Yes, that's what the OIDC spec states, and I'm pretty sure AAD does the right thing. https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter 
```json
{
   "userinfo":
    {
     "given_name": {"essential": true},
     "nickname": null,
     "email": {"essential": true},
     "email_verified": {"essential": true},
     "picture": null,
     "http://example.info/claims/groups": null
    },
   "id_token":
    {
     "auth_time": {"essential": true},
     "acr": {"values": ["urn:mace:incommon:iap:silver"] }
    }
  }
```

And Client Capabilities are a special claim for the access token:

```json
{
   "userinfo":
    {
     "given_name": {"essential": true},
     "nickname": null,
     "email": {"essential": true},
     "email_verified": {"essential": true},
     "picture": null,
     "http://example.info/claims/groups": null
    },
   "id_token":
    {
     "auth_time": {"essential": true},
     "acr": {"values": ["urn:mace:incommon:iap:silver"] }
    },
    "access_token":
    { 
      "xms_cc": { "values": ["llt"] }
    }
  }
```

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -31,7 +31,7 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             MsalAccessTokenCacheItem cachedAccessTokenItem = null;
             var logger = AuthenticationRequestParameters.RequestContext.Logger;
 
-            if (!_clientParameters.ForceRefresh)
+            if (!_clientParameters.ForceRefresh && !AuthenticationRequestParameters.HasClaims)
```
pmaytak: Should HasClaims be changed to HaveClaims, so it reads as AuthenticationRequestParameters.HaveClaims? Or if it refers to the request, for consistency, then it's fine. (Just a thought, I'm eh about it.)

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/OAuth2Tests/ClaimsTest.cs
```diff
@@ -2,19 +2,256 @@
...
+                var result = await app.AcquireTokenSilent(TestConstants.s_scope, account)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+
+                Trace.WriteLine("3. Silent Request + WithClaims does not return an AT from the cache");
+                harness.HttpManager.AddInstanceDiscoveryMockHandler(TestConstants.AuthorityUtidTenant);
+                harness.HttpManager.AddMockHandlerForTenantEndpointDiscovery(TestConstants.AuthorityUtidTenant);
+                harness.HttpManager.AddSuccessTokenResponseMockHandlerForPost(TestConstants.AuthorityUtidTenant);
+
+                result = await app.AcquireTokenSilent(TestConstants.s_scope, account)
```
pmaytak: For this test method and following one, how do we assert that step 2 and 3 have hit or bypassed the cache respectively?

bgavrilMS: Hmm, this is a very good question Peter. The way it works is ... if an access token is found in the cache, it is returned imediately. Otherwise, the cache logic is to find a refresh token and make an Http request to AAD called the "refersh_token" grant. 

So if the cache is hit -> no extra network calls occur
If the cache is not hit -> 3 network calls occur (2 for discovery, 1 for /token)

That `using` at line 31 ensures that all mock handlers added (lines 48-50) are actually consumed.

bgavrilMS: But I agree that this is too subtle, let me try to find a way to make these assertions in a more obvious way.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
bgavrilMS: In my mind, it's the request that has claims. Naming is so hard :)

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/OAuth2Tests/ClaimsTest.cs
bgavrilMS: Ideally we'd want to mock ICacheSessionManager, but exposing that is pretty tricky. I think I will let these tests be as they are for now and add more comments and assertions to explain what they do.

pmaytak: Ok, I understand. I'll take a look at that harness class. Yea, it's not super straightforward. Agree that for now, it's fine, since it's consistent. For the future, we should maybe think about improving this flow.

PR #1656, opened 2020-02-24
Removing unneeded semaphore release

trwalke:
Removing unneeded semaphore as it has the potential to cause threading issues in the future.


src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -145,9 +145,6 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
             catch (Exception ex)
             {
                 _logger.ErrorPiiWithPrefix(ex, "Broker invocation failed.");
-
-
-                s_readyForResponse.Release();
```
bgavrilMS: static locks are hard :)

trwalke: indeed

PR #1666, opened 2020-02-26
Fixing broker cancellation error code.

trwalke:
Fixing broker cancellation error code.
Broker error codes are now properly added to exceptions

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -160,17 +160,32 @@ namespace Microsoft.Identity.Client.Platforms.Android
                     return;
                 }
 
-                if (resultCode != (int)BrokerResponseCode.ResponseReceived)
+                switch (resultCode)
                 {
-                    s_androidBrokerTokenResponse = new MsalTokenResponse
-                    {
-                        Error = MsalError.BrokerResponseReturnedError,
-                        ErrorDescription = data.GetStringExtra(BrokerConstants.BrokerResultV2),
-                    };
-                }
-                else
-                {
-                    s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(data.GetStringExtra(BrokerConstants.BrokerResultV2));
+                    case (int)BrokerResponseCode.ResponseReceived:
+                        s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(data.GetStringExtra(BrokerConstants.BrokerResultV2));
+                        break;
+                    case (int)BrokerResponseCode.UserCancelled:
+                        s_androidBrokerTokenResponse = new MsalTokenResponse
+                        {
+                            Error = MsalError.AuthenticationCanceledError,
+                            ErrorDescription = MsalErrorMessage.AuthenticationCanceled,
+                        };
+                        break;
+                    case (int)BrokerResponseCode.BrowserCodeError:
+                        s_androidBrokerTokenResponse = new MsalTokenResponse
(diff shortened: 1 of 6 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -37,7 +37,7 @@ public static void SetAuthenticationContinuationEventArgs(int requestCode, Resul
             {
                 authorizationResult = ProcessFromEmbeddedWebview(requestCode, resultCode, data);
             }
-            else if (!String.IsNullOrEmpty(data.GetStringExtra(BrokerConstants.BrokerResultV2)))
+            else if (!String.IsNullOrEmpty(data.GetStringExtra(BrokerConstants.BrokerResultV2)) || requestCode == BrokerConstants.BrokerRequestId)
```
trwalke: The issue with the cancellation error was when the user cancels the authentication, there is no "BrokerResultV2". I added an extra check to check the request code that is added to the activity when it is launched. only broker adds a request code at the moment. both the system browser and embedded webview on android use a request value of zero. @bgavrilMS do you think its safe to only check for the request code here?

bgavrilMS: Could you add a comment on what the requestCode == 1001 does?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -160,17 +160,32 @@ internal static void SetBrokerResult(Intent data, int resultCode)
...
+                        s_androidBrokerTokenResponse = new MsalTokenResponse
+                        {
+                            Error = data.GetStringExtra(BrokerConstants.BrokerResultErrorCode),
+                            ErrorDescription = data.GetStringExtra(BrokerConstants.BrokerResultV2),
+                        };
+                        break;
+                    default:
+                        s_androidBrokerTokenResponse = new MsalTokenResponse
+                        {
+                            Error = BrokerConstants.BrokerUnknownErrorCode,
```
trwalke: @bgavrilMS this is the same error returned by android

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -160,17 +160,32 @@ internal static void SetBrokerResult(Intent data, int resultCode)
...
-                else
-                {
-                    s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(data.GetStringExtra(BrokerConstants.BrokerResultV2));
+                    case (int)BrokerResponseCode.ResponseReceived:
+                        s_androidBrokerTokenResponse = CreateMsalTokenResponseFromResult(data.GetStringExtra(BrokerConstants.BrokerResultV2));
+                        break;
+                    case (int)BrokerResponseCode.UserCancelled:
+                        s_androidBrokerTokenResponse = new MsalTokenResponse
+                        {
+                            Error = MsalError.AuthenticationCanceledError,
```
bgavrilMS: I don't see any changes to the iOS broker - assume this is separate PR. I remember that iOS broker does not handle this case explicitly, it just throws whatever error code it gets from the broker. 

So there are 2 ways of fixing this: 

1. make changes to the iOS broker so that it throws this exact error code 
2. make changes to the Android broker so that it throws the same error code with ios

Personally I like option 1 more because it's closer to webview error code. However it's not exactly backwards compatible. Option 2 is more backwards compatible.

PR #1668, opened 2020-02-27
MSAL android broker GetAccountsAsync() will now query the account man…

trwalke:
MSAL android broker GetAccountsAsync() will now query the account manager instead of the internal cache.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -17,6 +17,8 @@ using Microsoft.Identity.Client.Utils;
 using Microsoft.Identity.Client.PlatformsCommon.Factories;
 using Microsoft.Identity.Client.ApiConfig.Executors;
 using Microsoft.Identity.Client.Cache;
+using Microsoft.Identity.Client.Internal.Broker;
+using Microsoft.Identity.Client.UI;
 
 namespace Microsoft.Identity.Client
 {
@@ -85,6 +87,15 @@ namespace Microsoft.Identity.Client
         {
             RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
             IEnumerable<IAccount> accounts = Enumerable.Empty<IAccount>();
+
+#if ANDROID
+            if (AppConfig.IsBrokerEnabled)
+            {
+                var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
+                accounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
+                return accounts;
+            }
+#endif
             if (UserTokenCache == null)
             {
                 requestContext.Logger.Info("Token cache is null or empty. Returning empty list of accounts.");
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -88,14 +88,14 @@ namespace Microsoft.Identity.Client
             RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
             IEnumerable<IAccount> accounts = Enumerable.Empty<IAccount>();
 
-#if ANDROID
-            if (AppConfig.IsBrokerEnabled)
+
+            if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
                 accounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
                 return accounts;
             }
-#endif
+
             if (UserTokenCache == null)
             {
                 requestContext.Logger.Info("Token cache is null or empty. Returning empty list of accounts.");
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
@@ -26,7 +26,9 @@ namespace Microsoft.Identity.Client.Internal.Broker
         public const string LocalAccountId = "local_account_id";
         public const string UserName = "username";
         public const string iOSBrokerNonce = "broker_nonce"; // included in request and response with iOS Broker v3
-        
+        public const string Environment = "mEnvironment";
+
+
         public const string iOSBrokerUserCancellationErrorCode = "-50005";
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Identity.Client.Internal.Broker
         public const string LocalAccountId = "local_account_id";
         public const string UserName = "username";
         public const string iOSBrokerNonce = "broker_nonce"; // included in request and response with iOS Broker v3
-        public const string Environment = "mEnvironment";
+        public const string Environment = "environment";
 
 
         public const string iOSBrokerUserCancellationErrorCode = "-50005";
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
@@ -28,7 +28,6 @@ namespace Microsoft.Identity.Client.Internal.Broker
         public const string iOSBrokerNonce = "broker_nonce"; // included in request and response with iOS Broker v3
         public const string Environment = "environment";
 
-
         public const string iOSBrokerUserCancellationErrorCode = "-50005";
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Identity.Client.Internal.Broker
 
         Task<MsalTokenResponse> AcquireTokenUsingBrokerAsync(Dictionary<string, string> brokerPayload);
 
+        //This method is only available to brokers that have the BrokerSupportsSilentFlow flag enabled
         Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID);
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
@@ -13,5 +13,7 @@ namespace Microsoft.Identity.Client.Internal.Broker
         bool CanInvokeBroker();
 
         Task<MsalTokenResponse> AcquireTokenUsingBrokerAsync(Dictionary<string, string> brokerPayload);
+
+        Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID);
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
@@ -23,5 +23,10 @@ namespace Microsoft.Identity.Client.Internal.Broker
         {
             throw new PlatformNotSupportedException(MsalErrorMessage.BrokerNotSupportedOnThisPlatform);
         }
+
+        public Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID)
+        {
+            throw new PlatformNotSupportedException(MsalErrorMessage.BrokerNotSupportedOnThisPlatform);
+        }
     }
 }
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -101,8 +101,7 @@ namespace Microsoft.Identity.Client.Platforms.Android
                         s_androidBrokerTokenResponse = new MsalTokenResponse
                         {
                             Error = MsalError.BrokerResponseReturnedError,
-                            ErrorDescription = "Failed to acquire token silently from the broker. In order to perform brokered authentication on android" +
-                    " you need to ensure that you have installed either Intune Company Portal (Version 5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater).",
+                            ErrorDescription = "Failed to acquire token silently from the broker." + MsalErrorMessage.AndroidBrokerCannotBeInvoked,
                         };
                     }
                     return;
@@ -232,8 +231,7 @@ namespace Microsoft.Identity.Client.Platforms.Android
 
         private void HandleBrokerOperationError(Exception ex)
         {
-            _logger.Error("In order to perform brokered authentication on android" +
-                          " you need to ensure that you have installed either Intune Company Portal (Version 5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater).");
+            _logger.Error(MsalErrorMessage.AndroidBrokerCannotBeInvoked);
             if (ex is MsalException)
                 throw ex;
             else
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -213,5 +213,27 @@ namespace Microsoft.Identity.Client.Platforms.Android
 
             return MsalTokenResponse.CreateFromBrokerResponse(response);
         }
+
+        public async Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID)
+        {
+            Dictionary<string, string> brokerPayload = new Dictionary<string, string>();
+            brokerPayload.Add(BrokerParameter.ClientId, clientID);
+
+            try
+            {
+                await _brokerHelper.InitiateBrokerHandshakeAsync(_activity).ConfigureAwait(false);
+
+                return _brokerHelper.GetBrokerAccountsInAccountManager(brokerPayload);
+            }
+            catch (Exception ex)
+            {
+                _logger.Error("Broker Operation Failed to complete. In order to perform brokered authentication on android" +
+                    " you need to ensure that you have installed either Intune Company Portal (Version 5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater).");
+                if (ex is MsalException)
+                    throw;
+                else
+                    throw new MsalClientException(MsalError.AndroidBrokerOperationFailed, ex.Message, ex);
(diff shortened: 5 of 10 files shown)


src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
```diff
@@ -26,7 +26,9 @@ internal class BrokerResponseConst
         public const string LocalAccountId = "local_account_id";
         public const string UserName = "username";
         public const string iOSBrokerNonce = "broker_nonce"; // included in request and response with iOS Broker v3
-        
+        public const string Environment = "mEnvironment";
```
jmprieur: why the m?

trwalke: typo. fixed

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -85,6 +87,15 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
         {
             RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
             IEnumerable<IAccount> accounts = Enumerable.Empty<IAccount>();
+
+#if ANDROID
+            if (AppConfig.IsBrokerEnabled)
+            {
+                var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
+                accounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
```
jennyf19: >GetAccountsAsync(AppConfig.ClientId) [](start = 40, length = 36)

question: GetAccountsAsync returns the accounts based on the clienttId here? shouldn't it be based on the user?

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -85,6 +87,15 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
         {
             RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
             IEnumerable<IAccount> accounts = Enumerable.Empty<IAccount>();
+
+#if ANDROID
+            if (AppConfig.IsBrokerEnabled)
+            {
+                var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
```
jennyf19: >null [](start = 70, length = 4)

it's okay to have a null broker here?

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
```diff
@@ -23,5 +23,10 @@ public Task<MsalTokenResponse> AcquireTokenUsingBrokerAsync(Dictionary<string, s
         {
             throw new PlatformNotSupportedException(MsalErrorMessage.BrokerNotSupportedOnThisPlatform);
         }
+
+        public Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID)
```
jennyf19: can this be unit tested?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -213,5 +213,27 @@ private static MsalTokenResponse CreateMsalTokenResponseFromResult(string broker
...
         }
+
+        public async Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID)
+        {
+            Dictionary<string, string> brokerPayload = new Dictionary<string, string>();
+            brokerPayload.Add(BrokerParameter.ClientId, clientID);
+
+            try
+            {
+                await _brokerHelper.InitiateBrokerHandshakeAsync(_activity).ConfigureAwait(false);
```
jennyf19: >_activity [](start = 65, length = 9)

how is this activity getting created correctly if there was a null broker instance created?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -213,5 +213,27 @@ private static MsalTokenResponse CreateMsalTokenResponseFromResult(string broker
...
+
+                return _brokerHelper.GetBrokerAccountsInAccountManager(brokerPayload);
+            }
+            catch (Exception ex)
+            {
+                _logger.Error("Broker Operation Failed to complete. In order to perform brokered authentication on android" +
```
jennyf19: >_logger [](start = 16, length = 7)

shouldn't this be something more like "silent auth w/android broker failed because there were no accounts in the account manager"?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -213,5 +213,27 @@ private static MsalTokenResponse CreateMsalTokenResponseFromResult(string broker
...
+                    " you need to ensure that you have installed either Intune Company Portal (Version 5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater).");
+                if (ex is MsalException)
+                    throw;
+                else
+                    throw new MsalClientException(MsalError.AndroidBrokerOperationFailed, ex.Message, ex);
```
jennyf19: can you add logging through here? in which scenarios would an MsalException be thrown?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -197,14 +197,14 @@ public void CheckForBrokerAccountInfoInAccountManager(IDictionary<string, string
 
                     if (string.Equals((accountData[BrokerResponseConst.UserName]).ToString(), username, StringComparison.OrdinalIgnoreCase))
                     {
-                        brokerPayload[BrokerParameter.HomeAccountId] = accountData[BrokerResponseConst.HomeAccountId].ToString();
-                        brokerPayload[BrokerParameter.LocalAccountId] = accountData[BrokerResponseConst.LocalAccountId].ToString();
+                        brokerPayload[BrokerParameter.HomeAccountId] = accountData[BrokerResponseConst.HomeAccountId]?.ToString();
```
jennyf19: instead of pulling these out twice, can you store the values in a local variable?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -216,6 +216,52 @@ public void CheckForBrokerAccountInfoInAccountManager(IDictionary<string, string
             throw new MsalUiRequiredException(MsalError.NoAndroidBrokerAccountFound, MsalErrorMessage.NoAndroidBrokerAccountFound);
         }
 
+        /// <summary>
+        /// This method will acquire all of the accounts in the account manager that have an access token for the given client ID.
+        /// </summary>
```
jennyf19: why is this by client id?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -216,6 +216,52 @@ public void CheckForBrokerAccountInfoInAccountManager(IDictionary<string, string
...
+        public IEnumerable<IAccount> GetBrokerAccountsInAccountManager(IDictionary<string, string> brokerPayload)
```
jennyf19: isn't this code similar to some other code in this class? can some of it be moved to a similar method?

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
trwalke: it is not a null broker, the activity that is passed in is null which is fine for non interactive broker operations

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
trwalke: no, there is no user specified as a parameter because this flow was designed to get all users from the broker. The broker requires a client ID when making a request to get all accounts because it separates accounts by client ID.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
trwalke: not a null broker, just a null activity.

trwalke: since this is not an interactive operation no activity is needed

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
trwalke: no, this code is only designed to get the accounts from the broker. no authentication here.

trwalke: However, i can update the message to be more specific to get accounts

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
trwalke: The exception will be thrown if the broker handshake operation fails (this is already logged in a different method) or if Xamarin were to throw some exception from its code.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
trwalke: because the broker separates application caches by client ID. so you need to pass in the client id to get all of the accounts for the current application.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
trwalke: They are not pulled out twice, only once. The first if statement has a return at the end so after the values are pulled out the first time, the second if statement will never happen.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
bgavrilMS: I think the iOS broker doesn't support this. You should add an extra condition to the `if` on line 92: 

`if AppConfig.BrokerEnabled && PlatformProxy.BrokerSupportsSilentCall `

`GetAccounts` does not change for iOS broker.

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
```diff
@@ -13,5 +13,7 @@ internal interface IBroker
         bool CanInvokeBroker();
 
         Task<MsalTokenResponse> AcquireTokenUsingBrokerAsync(Dictionary<string, string> brokerPayload);
+
+        Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID);
```
bgavrilMS: Please add a comment stating that is only supported in conjuction with BrokerSupportsSilentFlow flag

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
bgavrilMS: We should never throw MsalException - only `MsalClientException` and `MsalServiceException`. `MsalException` exists only as a base class.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -213,5 +213,27 @@ private static MsalTokenResponse CreateMsalTokenResponseFromResult(string broker
...
+            {
+                await _brokerHelper.InitiateBrokerHandshakeAsync(_activity).ConfigureAwait(false);
+
+                return _brokerHelper.GetBrokerAccountsInAccountManager(brokerPayload);
+            }
+            catch (Exception ex)
+            {
+                _logger.Error("Broker get accounts operation failed to complete. In order to perform brokered authentication on android" +
+                    " you need to ensure that you have installed either Intune Company Portal (Version 5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater).");
+                if (ex is MsalException)
```
bgavrilMS: This piece of error handling is repeated at line 62 in `AcquireTokenUsingBrokerAsync` - please consolidate, e.g. via a private method.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -195,16 +183,19 @@ public void CheckForBrokerAccountInfoInAccountManager(IDictionary<string, string
...
-                        brokerPayload[BrokerParameter.LocalAccountId] = accountData[BrokerResponseConst.LocalAccountId].ToString();
+                        brokerPayload[BrokerParameter.HomeAccountId] = accountDataHomeAccountID;
+                        brokerPayload[BrokerParameter.LocalAccountId] = accountDataLocalAccountID;
                         _logger.Info("Found broker account in Android account manager using the provided login hint.");
                         return;
                     }
                     
-                    if (string.Equals((accountData[BrokerResponseConst.HomeAccountId]).ToString(), homeAccountId) &&
-                         string.Equals((accountData[BrokerResponseConst.LocalAccountId]).ToString(), localAccountId))
+                    if (string.Equals(accountDataHomeAccountID, homeAccountId) &&
```
bgavrilMS: Please add Culture to string comparisons.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -216,6 +207,58 @@ public void CheckForBrokerAccountInfoInAccountManager(IDictionary<string, string
...
+            if (string.IsNullOrEmpty(accounts))
+            {
+                _logger.Info("Android account manager didn't return any accounts.");
+            }
+
+            List<IAccount> brokerAccounts = new List<IAccount>();
+
+            if (!string.IsNullOrEmpty(accounts))
+            {
+                dynamic authResult = JArray.Parse(accounts);
```
bgavrilMS: Why do you need this `dynamic` ? We have better ways of performing JSON deserialization. 

When you do `Json.Deserialize<T>()`, reflection is only used once. Subsequent calls to Deserialize<T> will no longer reflect on the type T. 

I believe this optimization is bypassed if you use `dynamic`

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
```diff
@@ -216,6 +207,58 @@ public void CheckForBrokerAccountInfoInAccountManager(IDictionary<string, string
...
+
+                foreach (JObject account in authResult)
+                {
+                    var accountData = account[BrokerResponseConst.Account];
```
bgavrilMS: I think there's potential for some null refs here and line 250-253. Consider using `HasKey` to avoid this.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -85,6 +87,15 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
         {
             RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
             IEnumerable<IAccount> accounts = Enumerable.Empty<IAccount>();
+
+#if ANDROID
```
bgavrilMS: Please remove if #Android (see comment below). And unit test this piece of code - you need to assert that `broker.GetAccountsAsync(clientID)` is called. There already exist infrastructure to inject a mocked broker, so this should be fairly easy.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
trwalke: ah, i for got about this parameter. yes it will be cleaner

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
trwalke: yes

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
trwalke: The JSON deserialization methods we have now do not work on Android. Whenever i tried to use them, they resulted in a null value. dynamic is the only thing that I could get to work

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
trwalke: yes but the PlatformProxy.BrokerSupportsSilentCall requires a mocked platform proxy. how do I achieve that?

bgavrilMS: The way I would do it is to add an extra param to the existing `SetBrokerForTest` [hook](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/86c5572232f2657eba6eb1919a749874877a996b/src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IPlatformProxy.cs#L111)

`void SetBrokerForTest(IBroker broker, bool? supportsSilentFlow);`

It's important to use a nullable boolean because there are 3 states (null for not set, true or false for the possible values). Product uses null. Test uses values.

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
bgavrilMS: Yes, we added a unit test showing that GetAccounts calls this method.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
jennyf19: did you see this comment?

---
In reply to: [385631543](https://github.com/azuread/microsoft-authentication-library-for-dotnet/pull/1668#discussion_r385631543) [](ancestors = 385631543)

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBrokerHelper.cs
pmaytak: Interesting behavior. I am curious to find out the reason for that.

bgavrilMS: There are many other places where we deserialize using Deserialize<T>() in Android, for example when communicating with EVO. Newtonsoft.Json definitely works on Android....

pmaytak: I would create an issue to fix this and at least document.

bgavrilMS: Well, this piece of code is not perf sensitive (public client flows generally are not), so not sure we can log a perf bug. Otherwise, this has the cool benefit that it works :) and [refactoring, not on the backlog](https://ronjeffries.com/xprog/articles/refactoring-not-on-the-backlog/).

PR #1672, opened 2020-03-03
Enabling Arlington automation tests

trwalke:


tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -30,17 +30,28 @@ public class ConfidentialClientIntegrationTests
     {
         private static readonly string[] s_scopes = { "User.Read" };
         private static readonly string[] s_oboServiceScope = { "api://23c64cd8-21e4-41dd-9756-ab9e2c23f58c/access_as_user" };
+        private static readonly string[] s_arlingtonOBOServiceScope = { "https://arlmsidlab1.us/IDLABS_APP_Confidential_Client/user_impersonation" };
         private static readonly string[] s_keyvaultScope = { "https://vault.azure.net/.default" };
         private static readonly string[] s_adfsScopes = { "openid", "profile" };
 
         //TODO: acquire scenario specific client ids from the lab resonse
+        private const string _publicClientIDOBO = "be9b0186-7dfd-448a-a944-f771029105bf";
```
bgavrilMS: Please use the consistent naming, as suggested by Visual Studio. Constants use **P**ascal**C**ased

bgavrilMS: What is a PublicClient OBO?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -162,14 +181,16 @@ public async Task ConfidentialClientWithRSACertificateTestAsync()
         }
 
         [TestMethod]
-        public async Task ConfidentialClientWithClientSecretTestAsync()
+        [DataRow(ConfidentialClientID, TestAuthority, "default", DisplayName ="Public Cloud")]
```
bgavrilMS: Can you rename the existing constants as "PublicClould" to be more consistent with this model, e.g. `PublicCloudAuthority` etc.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -57,6 +68,14 @@ public void TestInitialize()
             {
                 _keyVault = new KeyVaultSecretsProvider();
                 _ccaSecret = _keyVault.GetSecret(TestConstants.MsalCCAKeyVaultUri).Value;
+                _arlingtonCCASecret = _keyVault.GetSecret(TestConstants.MsalArlingtonCCAKeyVaultUri).Value;
+
+                if (SecretData == null)
+                {
+                    SecretData = new Dictionary<string, string>();
+                    SecretData.Add("default", _ccaSecret);
```
bgavrilMS: Why don't we reuse the existing `AzureCloudInstance` enum ? It will be clearer.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -99,7 +118,7 @@ public async Task ConfidentialClientWithCertificateTestAsync()
             var confidentialClientAuthority = TestAuthority;
 
             confidentialApp = ConfidentialClientApplicationBuilder
-                .Create(ConfidentialClientID)
+                .Create(_arlingtonConfidentialClientIDOBO)
```
bgavrilMS: Shouldn't this test run on the public cloud as well?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -383,27 +412,37 @@ public async Task AcquireTokenWithClientSecretFromAdfsAsync()
             Assert.IsNull(authResult.IdToken);
         }
 
-        private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
+        private async Task RunOnBehalfOfTestAsync(LabResponse labResponse, bool usGov = false)
```
bgavrilMS: Use the AzureCloudInstance enum as suggested above.

tests/Microsoft.Identity.Test.Integration.net45/Infrastructure/SeleniumExtensions.cs
```diff
@@ -28,7 +28,7 @@ public static IWebDriver CreateDefaultWebDriver()
 
             // ~2x faster, no visual rendering
             // remove when debugging to see the UI automation
-            options.AddArguments("headless");
+            //options.AddArguments("headless");
```
bgavrilMS: pls undo

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -57,6 +68,14 @@ public void TestInitialize()
             {
                 _keyVault = new KeyVaultSecretsProvider();
                 _ccaSecret = _keyVault.GetSecret(TestConstants.MsalCCAKeyVaultUri).Value;
+                _arlingtonCCASecret = _keyVault.GetSecret(TestConstants.MsalArlingtonCCAKeyVaultUri).Value;
+
+                if (SecretData == null)
```
bgavrilMS: The idea of having a dictionary where the key is the Cloud and the data is some could specific detail is pretty good. But we should abstract this further because right now these tests do too much - they know how to pick between Public Cloud and Arlington data and also they know how the test logic. This breaks the Single Responsability Principle and is a code smell that there is a missing object.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -162,14 +181,16 @@ public async Task ConfidentialClientWithRSACertificateTestAsync()
         }
 
         [TestMethod]
-        public async Task ConfidentialClientWithClientSecretTestAsync()
+        [DataRow(ConfidentialClientID, TestAuthority, "default", DisplayName ="Public Cloud")]
+        [DataRow(_arlingtonConfidentialClientIDOBO, _arlingtonAuthority, "arlington", DisplayName ="Arlington Cloud")]
```
bgavrilMS: DataRow is cool, but it should be used consistently - also in the ROPC and other tests.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
bgavrilMS: Next ask might be to add tests for Germany cloud. `bool` is not a good choice here, an enum is.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -181,8 +197,11 @@ private async Task RunHappyPathTestAsync(LabResponse labResponse)
                 .ConfigureAwait(false);
 
             // TODO: you can now assert each request and response
-            var (req, res) = factory.RequestsAndResponses.Single(x => x.Item1.RequestUri.AbsoluteUri == "https://login.microsoftonline.com/organizations/oauth2/v2.0/token");
-            var telemetryValue = req.Headers.Where(h => h.Key == "x-client-last-telemetry");
+            if (!USGov)
+            {
+                var (req, res) = factory.RequestsAndResponses.Single(x => x.Item1.RequestUri.AbsoluteUri == "https://login.microsoftonline.com/organizations/oauth2/v2.0/token");
+                var telemetryValue = req.Headers.Where(h => h.Key == "x-client-last-telemetry");
```
bgavrilMS: Why are we not sending telemetry to the US cloud?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
henrik-me: I have certainly asked for this previously so not an unreal scenario

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
trwalke: this is a temporary piece of code. the lab needs to make updates to properly return the Arlington values in the api. I have reached out to them already and this should be resolved soon. Once resolved, we can simply use the data in the lab response rather than needing to use if statements/data rows to change the test logic depending on the environment. 

Issue tracking here
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1706

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -163,25 +177,32 @@ public async Task AcquireTokenWithFederatedUsernameIncorrectPasswordAsync()
...
                 .Create(labResponse.App.AppId)
-                .WithHttpClientFactory(factory)
-                .WithAuthority(_authority)
-                .Build();
+                .WithHttpClientFactory(factory);
 
+            switch (labResponse.User.AzureEnvironment)
+            {
+                case AzureEnvironment.azureusgovernment:
+                    builder.WithAuthority(labResponse.Lab.Authority + "organizations");
```
bgavrilMS: Interesting, so for Arlinghton the authority is already `organizations` ?

bgavrilMS: You actually have to use `organizations` for Public AAD as well. So this is common code, i.e. you don't need a `switch`

bgavrilMS: You can just use 

`builder.WithAuthority(labResponse.Lab.Authority, "organizations");`

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -164,12 +175,27 @@ public async Task ConfidentialClientWithRSACertificateTestAsync()
...
         public async Task ConfidentialClientWithClientSecretTestAsync()
         {
-            var confidentialClientAuthority = TestAuthority;
+            await RunTestWithClientSecretAsync(PublicClientConfidentialClientID,
+                                                           PublicCloudTestAuthority,
+                                                           _publicCloudCcaSecret).ConfigureAwait(false);
+        }
+
+        [TestMethod]
+        public async Task ArlingtonConfidentialClientWithClientSecretTestAsync()
```
bgavrilMS: Would be good to test the `.WithCertificate` option as well. It's probably more imporant than `WithSecret`

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/DeviceCodeFlowIntegrationTest.cs
```diff
@@ -71,7 +86,18 @@ public async Task DeviceCodeFlowMsaTestAsync()
         private async Task AcquireTokenWithDeviceCodeFlowAsync(LabResponse labResponse, string userType)
         {
             Trace.WriteLine($"Calling AcquireTokenWithDeviceCodeAsync with {0}", userType);
-            var pca = PublicClientApplicationBuilder.Create(labResponse.App.AppId).Build();
+            var builder = PublicClientApplicationBuilder.Create(labResponse.App.AppId);
+
+            switch (labResponse.User.AzureEnvironment)
+            {
+                case AzureEnvironment.azureusgovernment:
+                    builder.WithAuthority(labResponse.Lab.Authority + labResponse.Lab.TenantId);
```
bgavrilMS: This is a valid test, but it is different than the main Happy Path device code flow. This test is using a `tenanted authority`. The AAD test is using `common`. 

Note that if you do not specify a .WithAuthority, the `common` public cloud authority is used. So easiest for your tests is to use

`.WithAuthority(labResponse.Lab.Authority)` ... no `switch` needed

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -170,10 +185,20 @@ private async Task<AuthenticationResult> RunTestForUserAsync(LabResponse labResp
...
             else
             {
-                pca = PublicClientApplicationBuilder
+                var builder = PublicClientApplicationBuilder
                     .Create(labResponse.App.AppId)
-                    .WithRedirectUri(SeleniumWebUI.FindFreeLocalhostRedirectUri())
-                    .Build();
+                    .WithRedirectUri(SeleniumWebUI.FindFreeLocalhostRedirectUri());
+
+                switch(labResponse.User.AzureEnvironment)
```
bgavrilMS: Same comment as before, just use `WithAuthority` for all cases.

tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/DeviceCodeFlowIntegrationTest.cs
trwalke: This is the intended path to test. [JAVA](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/520eebfc32829c005452d349762e20fe710769f9/src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java#L45)

PR #1691, opened 2020-03-09
Fix disabled unit test

bgavrilMS:


tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -766,25 +766,17 @@ public void GetAuthority_AccountWithIdPassed_TenantedAuthorityUsed()
             Assert.AreEqual(TestConstants.AuthorityTestTenant, authority.AuthorityInfo.CanonicalAuthority);
         }
 
+        [TestMethod]
         public async Task AcquireTokenSilentNullAccountErrorTestAsync()
```
pmaytak: `AcquireTokenSilentNullAccountErrorTestAsync`

Separate with underscores?

Issue #1696, opened 2020-03-11
[Bug]  Certain error messages are not returned from the Andorid Broker


trwalke:
**Which Version of MSAL are you using ?**
4.9.0

**Platform**
xamarin android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Repro**
Create a service error with a misconfigured app. The, using broker, make an interactive call.

**Expected behavior**
Msal should return an MSALServiceException with the proper error information

**Actual behavior**
Msal return a service exception with an error message stating that the broker returned an error without the specific broker error message

**Possible Solution**

**Additional context/ Logs / Screenshots**
Add any other context about the problem here, such as logs and screebshots. Logging is described at https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging

PR #1697, opened 2020-03-11
Broker error message fix

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1696

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
@@ -170,11 +170,18 @@ namespace Microsoft.Identity.Client.Platforms.Android
                         break;
                     case (int)BrokerResponseCode.BrowserCodeError:
                         dynamic errorResult = JObject.Parse(data.GetStringExtra(BrokerConstants.BrokerResultV2));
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -169,10 +169,12 @@ internal static void SetBrokerResult(Intent data, int resultCode)
                         };
                         break;
                     case (int)BrokerResponseCode.BrowserCodeError:
+                        dynamic errorResult = JObject.Parse(data.GetStringExtra(BrokerConstants.BrokerResultV2));
```
bgavrilMS: Code needs to be more defensive: 

- make sure `data` and `data.GetStringExtra` are not null
- make sure `errorResult` has those keys that you look to. If they don't you still need to produce an "unknown error", setting Error and ErrorDescription to null is not a good user experience.

trwalke: data has a null check earlier in the code.
There are validation methods on the interactive and silent request that will catch the null value and replace it with the unknown error so using a null here is ok.

PR #1703, opened 2020-03-13
Enabled POP on AcquireTokenSilent

bgavrilMS:


henrik-me:
Assuming tests are already in place

bgavrilMS:
Yes, I've got integration tests for this. But didn't realize it was still marked as internal... Test assembly have "internals visible" friendship with MSAL ...

Issue #1676, opened 2020-03-04
[Bug] When using the Android broker, RemoveAccount does nothing


bgavrilMS:
**Which Version of MSAL are you using ?**
4.9-preview

**Platform**
android

**Repro**

1. (optional) AcquireTokenInteractive with broker
2. GetAccounts - this actually gets accounts from the broker
3. call RemoveAccount on one of the accounts

**Expected behavior**
The account should be removed from the broker or at least the access token should be removed.

**Actual behavior**
Nothing happens

Note: When using the Android broker, MSAL no longer controls the accounts. Instead, it uses delegates this responsability to the broker. `GetAccounts` works, but `RemoveAccount` doesn't call into the broker in 4.9.

henrik-me:
Triage: We need to call remove account on broker and explore in further details on the behavior of the RemoveAccount api.

harmjonker:
Application starts with select-account screen  (as per preview release the acquiresilent call crashes we direct the acquiretoken call to the account select step)  

![Screenshot_20200305-214431_Company Portal](https://user-images.githubusercontent.com/55669455/76024828-0d073f80-5f2c-11ea-8c21-a5a72eeec7a6.jpg)
=> I choose 'Account toevoegen'  (add new account)

I then shows an ADFS screen (as least I think this is an AFDS screen) with only one of the accounts. 
![Screenshot_20200305-214522_Company Portal](https://user-images.githubusercontent.com/55669455/76025090-8010b600-5f2c-11ea-9bda-fb19616b0096.jpg)

 I then enter a new account name (truus.ea11@nslogin.nl)

![Screenshot_20200305-214441_Company Portal](https://user-images.githubusercontent.com/55669455/76024961-463faf80-5f2c-11ea-96db-8c66d69cbd2f.jpg)

After clicking 'Volgende' ( Next) control returns to my app with the harm.jonker@nslogin.nl account as selected account.

trwalke:
This fix will now allow users to remove account data from the android broker for a specific client id. once the account is removed, it will no longer show up in the results of GetAccountsAcync().

trwalke:
Released in 4.10

PR #1712, opened 2020-03-16
Enabling the Remove account api on MSAL for andorid broker

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1676

### DIFF
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Identity.Client
             if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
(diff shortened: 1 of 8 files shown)


src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -139,7 +139,7 @@ public async Task RemoveAsync(IAccount account)
             if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
-                broker.RemoveAccountAsync(AppConfig.ClientId, account);
+                await broker.RemoveAccountAsync(AppConfig.ClientId, account).ConfigureAwait(false);
```
bgavrilMS: Hmm, do we have a logger without the requestContext? We should add a bit of logging here...

trwalke: Ah, you might be right. let me double check

trwalke: The logger will be injected by the platform proxy so everything will be logged

Issue #1708, opened 2020-03-16
[Bug] Client Credentials flow with certificate not working with ADFS


bgavrilMS:
**Which Version of MSAL are you using ?**
4.9

**Platform**
confidential client

**What authentication flow has the issue?**
confidential client

**Repro**

```csharp
var cca = // create confidential client with certificate
await cca.AcquireTokenForClient(...)
```

**Expected behavior**
tokens

**Actual behavior**
bad cert error

**Possible Solution**
After investigation with PowerShell team, we believe the aud claim in the signed assertion we send to the authority is wrong. The aud claim should be the token endpoint, but we use the issuer.

bgavrilMS:
This is blocking Azure Stack

jmprieur:
@bgavrilMS : we use the issuer only in the case of ADFS? or even Azure AD? 
I agree the audience cannot be the issuer.

jmprieur:
@bgavrilMS @trwalke @jennyf19 @henrik-me @aiwangmicrosoft 
I propose we take a 4.9.1 (with Pop ATS)

bgavrilMS:
@jmprieur - given that we had a new feature in that changes the public API (instance metadata uri), I think it has to be called `4.10`

bgavrilMS:
@jmprieur - we use the token_endpoint value for `aud` for AAD. For ADFS for some reason it is set to use the issuer. 

But this isn't the only problem. I think the header that we send is wrong - we use `kid` but both Wilson and ADAL uses `x5t`.

henrik-me:
Seems like a regression from PoP? We used to use x5t/x5c (SNI) and did not have the kid previously?

PR #1714, opened 2020-03-17
Fix for #1708 - use token_endpoint as issuer and x5t in header

bgavrilMS:
This fixes the ADFS + confidential client scenario

Tested on an Azure Stack env which uses ADFS 2019. 

Needs an integration test, for which I am discussing with the lab team. Will follow up separately.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
@@ -205,7 +205,7 @@ namespace Microsoft.Identity.Client.Internal
 #endif
             }
 
-            [JsonProperty(PropertyName = JsonWebTokenConstants.ReservedHeaderParameters.X5T)]
+            [JsonProperty(PropertyName = JsonWebTokenConstants.ReservedHeaderParameters.X509CertificateThumbprint)]
             public string X509CertificateThumbprint { get; set; }
 
             [JsonProperty(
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
```diff
@@ -205,9 +210,25 @@ public JWTHeaderWithCertificate(ClientCredentialWrapper credential, bool sendCer
...
             }
 
-            [JsonProperty(PropertyName = JsonWebTokenConstants.ReservedHeaderParameters.KeyId)]
+            /// <summary>
+            /// x5t = base64 url encoded cert thumbprint 
+            /// </summary>
+            /// <remarks>
+            /// Mandatory for ADFS 2019
+            /// </remarks>
+            [JsonProperty(PropertyName = JsonWebTokenConstants.ReservedHeaderParameters.X509CertificateThumbprint)]
```
bgavrilMS: As @henrik-me  pointed out, this is a regression introduced by the PoP work (a longer way to say to "introduced by Bogdan" ). I added x5t back and also added kid. Kid is optional but it is added by Wilson - so we are now compliant with Wilson. We have integration tests for AAD which are all passing.

src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
```diff
@@ -191,6 +191,11 @@ public JWTHeaderWithCertificate(ClientCredentialWrapper credential, bool sendCer
                 : base(credential)
             {
                 X509CertificateThumbprint = Credential.Thumbprint;
```
jennyf19: could Credential be null above? line 193.

Issue #1505, opened 2019-11-14
[Feature Request] Remove RT if suberror is "bad_token" to avoid overloading ESTS


bgavrilMS:
If the RT has expired or has been tampered with, ESTS will return a "bad_token" suberror. MSAL does not currently take action on it, but it should delete the RT. The reason for this is explained by Dejan: 

_Based on our investigations into busy client loops, we know that some apps tend to sit in loops in the background and try to acquire token silently even though the auth library is returning an error. If the library knows that the RT is bad and the call will never succeed, we can cut these loops locally and not talk to the server. Loops like this have contributed to significant livesite outages in the past._

jmprieur:
@shoatman do you have any perspective on why we should not do that?

shoatman:
No.  bad_token is a bad_token and can be deleted.  With that said, I would expect to see something similar if you attempted to send an RT from WW to a sovereign cloud.  So I'd also add some tests to ensure that RTs are only ever being sent to the cloud that issued them.

trwalke:
Resolved in [4.11.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.11.0)

PR #1720, opened 2020-03-19
Remove account from cache for bad token error

neha-bhargava:
#1505

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
@@ -125,7 +125,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 if (MsalError.BadToken.Equals(e.SubError, StringComparison.OrdinalIgnoreCase))
                 {
                     await CacheManager.TokenCacheInternal.RemoveAccountAsync(AuthenticationRequestParameters.Account, AuthenticationRequestParameters.RequestContext).ConfigureAwait(false);
-                    logger.Warning("Removed the account from cache because of bad_token error.");
+                    logger.Warning("Failed to refresh access token because the refresh token is invalid, removing account from cache.");
                     throw;
                 }
 
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
@@ -120,7 +120,15 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 return await RefreshRtOrFailAsync(cancellationToken).ConfigureAwait(false);
             }
             catch (MsalServiceException e)
-            {
+            {           
+                //Remove the account from cache in case of bad_token sub error
+                if (MsalError.BadToken.Equals(e.SubError, StringComparison.OrdinalIgnoreCase))
+                {
+                    await CacheManager.TokenCacheInternal.RemoveAccountAsync(AuthenticationRequestParameters.Account, AuthenticationRequestParameters.RequestContext).ConfigureAwait(false);
+                    logger.Warning("Removed the account from cache because of bad_token error.");
+                    throw;
+                }
+
                 bool isAadUnavailable = e.IsAadUnavailable();
 
                 logger.Warning($"Refreshing the RT failed. Is AAD down? {isAadUnavailable}. Is there an AT in the cache that is usable? {cachedAccessTokenItem != null}");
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -120,7 +120,15 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                 return await RefreshRtOrFailAsync(cancellationToken).ConfigureAwait(false);
             }
             catch (MsalServiceException e)
-            {
+            {           
+                //Remove the account from cache in case of bad_token sub error
+                if (MsalError.BadToken.Equals(e.SubError, StringComparison.OrdinalIgnoreCase))
+                {
+                    await CacheManager.TokenCacheInternal.RemoveAccountAsync(AuthenticationRequestParameters.Account, AuthenticationRequestParameters.RequestContext).ConfigureAwait(false);
+                    logger.Warning("Removed the account from cache because of bad_token error.");
```
trwalke: Should be more like. "Failed to refresh access token because the refresh token is invalid. removing account from cache".

@bgavrilMS  thoughts?

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -120,7 +120,15 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+                    throw;
```
trwalke: @bgavrilMS should this throw or allow the code to continue and throw at line 143?

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
neha-bhargava: This will sound better. I will update it.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
bgavrilMS: I think it's fine, as an MsalUIRequiredException will be thrown back to the user, which makes sense since the RT can't be used

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
```diff
@@ -46,6 +47,59 @@ public async Task NullAccount_EmptyLoginHintAsync()
...
+
+                var accounts = await app.GetAccountsAsync().ConfigureAwait(false);
+                Assert.IsFalse(accounts.Any());
+
+                var refreshTokens = app.UserTokenCacheInternal.Accessor.GetAllRefreshTokens();
+                Assert.IsFalse(refreshTokens.Any());
+            }
+        }
+
+        private static void AddHttpMocks(MockHttpManager httpManager)
```
jennyf19: >AddHttpMocks [](start = 28, length = 12)

since this mock actually creates a invalid_grant error, could you modify the name to include that? it's a little misleading if just reading the code from the test that the invalid_grant part will be configured here.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
```diff
@@ -46,6 +47,59 @@ public async Task NullAccount_EmptyLoginHintAsync()
...
+
+                var tokenCacheHelper = new TokenCacheHelper();
+                tokenCacheHelper.PopulateCache(app.UserTokenCacheInternal.Accessor, addSecondAt: false);
+
+                Trace.WriteLine("2. Configure AAD to respond with the typical Invalid Grant error and Bad Token sub error");
+                AddHttpMocks(harness.HttpManager);
+                var account = new Account(TestConstants.s_userIdentifier, TestConstants.DisplayableId, null);
+
+                // Act
+                var excepion = await AssertException.TaskThrowsAsync<MsalUiRequiredException>(() => app
```
jennyf19: nit: spelling -> exception.

neha-bhargava: Changed :)

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
neha-bhargava: Good point. Updated the name.

PR #1727, opened 2020-03-24
ADFS client assertion test with Wilson

bgavrilMS:


bgavrilMS:
I'm going to merge this in since our build is failing. It's only test code.

PR #1730, opened 2020-03-25
use ADAL generic value

jennyf19:
we need to use the adal generic so devices using adal and msal will not be impacted. but just for writing, lookup should be just service and account.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBrokerConstants.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBrokerConstants.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Identity.Client.Platforms.iOS
         public const string BrokerKeyLabel = "Broker Key Label";
         public const string BrokerKeyComment = "Broker Key Comment";
         public const string BrokerKeyStorageDescription = "Storage for broker key";
-        public const string LocalSettingsContainerName = "MicrosoftIdentityClient";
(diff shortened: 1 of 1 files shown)


jmprieur:
is it a breaking change? for MSAL only apps?

jennyf19:
@jmprieur as discussed off-line, not a breaking change as the look up has not changed.

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSBrokerConstants.cs
```diff
@@ -36,7 +36,7 @@ internal class iOSBrokerConstants
         public const string BrokerKeyLabel = "Broker Key Label";
         public const string BrokerKeyComment = "Broker Key Comment";
         public const string BrokerKeyStorageDescription = "Storage for broker key";
-        public const string LocalSettingsContainerName = "MicrosoftIdentityClient";
+        public const string LocalSettingsContainerName = "ActiveDirectoryAuthenticationLibrary";
```
bgavrilMS: Same questions as @jmprieur asked:

- what is the impact of this change to existing apps? is it a breaking change? will it result in users having to login again?

jennyf19: i sent you an email with the details.

jennyf19:
closing...not doing this as discussed.

Issue #715, opened 2018-12-10
The scope parameter should be less strict in some of the AcquireTokenXXX methods (allow an empty set of scopes)


jmprieur:
**Is your feature request related to a problem? Please describe.**
In Web scenarios, the authorization code is requested by the ASP.NET / ASP.NET core middlewre, and redeeming the code is often done with MSAL.NET, but **only** to have a token in the cache, so that it can be used from controllers, using `AcquireTokenSilentAsync`. For instance in the *Web app calls web api* sample

```CSharp
 // This initializes the ASP.NET / ASP.NET Core part
options.ResponseType = "id_token code";
options.Scope.Add("offline_access");
options.Scope.Add("User.Read"); 

// Handling the auth redemption by MSAL.NET so that a token is available in the token cache
// where it will be usable from Controllers later (through the TokenAcquisition service)
var handler = options.Events.OnAuthorizationCodeReceived;
options.Events.OnAuthorizationCodeReceived = async context =>
{
 var _tokenAcquisition = context.HttpContext.RequestServices.GetRequiredService<ITokenAcquisition>();
 await _tokenAcquisition.AddAccountToCacheFromAuthorizationCode(context, options.Scope);
 await handler(context);
 };
```

Note that, as the name of the `AddAccountToCacheFromAuthorizationCode` method tells us, the intent is just to add an account to the cache from the authorization code. For more context see [active-directory-aspnetcore-webapp-openidconnect-v2 in Startup.cs#L67](https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/blob/ee3627031bad6027c69cc79a5ee43af8a9eb4802/Startup.cs#L67)

The simplified implementation of `AddAccountToCacheFromAuthorizationCode` is something like this:

```CSharp
/// <summary>
/// Scopes which are already requested by MSAL.NET. they can't be re-requested;
/// </summary>
private string[] scopesRequestedByMsalNet = new string[] { "openid", "profile", "offline_access" };

var result = await application.AcquireTokenByAuthorizationCodeAsync(context.ProtocolMessage.Code, scopes.Except(scopesRequestedByMsalNet));
```

[...]

PR #1735, opened 2020-03-31
Enable AcquireToken* with 0 scopes (just login) - #715

bgavrilMS:
See #715  for more details.

In practice this doesn't mean anything for AAD - an AccessToken for the `email` scope is still generated. 
For B2C however, the token response does not contain an access token, so the `AuthenticationResult` object will have AccessToken set to NULL. 
If there is no access token, `AcquireTokenSilent` will ALWAYS refresh the RT.


src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -1,225 +1,228 @@
...
-        {
-            CredentialType = StorageJsonValues.CredentialTypeAccessToken;
-        }
-
-        internal MsalAccessTokenCacheItem(
-            string preferredCacheEnv,
-            string clientId,
-            MsalTokenResponse response,
-            string tenantId,
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bgavrilMS: Hmm, VS detected a problem with spaces. There are no changes in this file!

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccountCacheItem.cs
```diff
@@ -18,22 +18,21 @@ internal MsalAccountCacheItem()
...
         internal MsalAccountCacheItem(
             string preferredCacheEnv,
-            MsalTokenResponse response,
+            string clientInfo,
+            IdToken idToken,
             string preferredUsername,
             string tenantId)
             : this()
         {
-            var idToken = IdToken.Parse(response.IdToken);
```
bgavrilMS: Small optimization - we seem to be parsing the IdToken like crazy. This is pretty expensive - it involves splitting a string and parsing a piece of JSON. Not a big deal, but when done n times ...

src/client/Microsoft.Identity.Client/Cache/Items/MsalIdTokenCacheItem.cs
```diff
@@ -63,7 +66,12 @@ internal MsalIdTokenCacheItem()
                                     IsAdfs ? string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/", Environment, "adfs") :
                                     string.Format(CultureInfo.InvariantCulture, "https://{0}/{1}/", Environment, TenantId ?? "common");
 
-        internal IdToken IdToken => IdToken.Parse(Secret);
```
bgavrilMS: This is a great example of dangerous property use :) Callers of this property imply that the token is already parsed. It's not. A method would have been better here. 

In the end I chose to cache the parsed value.

src/client/Microsoft.Identity.Client/ITokenCacheInternal.cs
```diff
@@ -30,17 +30,17 @@ internal interface ITokenCacheInternal : ITokenCache, ITokenCacheSerializer
         /// </summary>
         /// <returns></returns>
         Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem>> SaveTokenResponseAsync(
-            AuthenticationRequestParameters authenticationRequestParameters,
```
bgavrilMS: Just renames ...

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -28,6 +29,13 @@ internal class ClientCredentialRequest : RequestBase
 
         protected override async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
         {
+            if (AuthenticationRequestParameters.Scope == null || !AuthenticationRequestParameters.Scope.Any())
```
bgavrilMS: Client Creds and OBO do not support getting 0 scopes. It doesn't make sense.

tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/OboFlowTests.cs
```diff
@@ -1,932 +0,0 @@
...
-using System.Threading.Tasks;
-using Microsoft.Identity.Client;
-using Microsoft.Identity.Client.Internal.Cache;
-using Microsoft.Identity.Client.Internal.Http;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Test.MSAL.NET.Unit.Mocks;
-
-namespace Test.MSAL.NET.Unit.RequestsTests
-{
-    class OboFlowTests
```
bgavrilMS: These are just some ADAL tests for OBO that were never really properly moved. They are useless now.

src/client/Microsoft.Identity.Client/Cache/Items/MsalIdTokenCacheItem.cs
jmprieur: why not change to a method, then?

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -303,6 +302,8 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
         public const string CustomMetadataInstanceOrUri = "You have configured your own instance metadata using both an Uri and a string. Only one is supported. " +
             "See https://aka.ms/msal-net-custom-instance-metadata for more details.";
 
+        public const string ScopesRequired = "At least one scope needs to be requested for this authentication flow.";
```
jmprieur, line 305: LGTM.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
```diff
@@ -162,7 +163,8 @@ public static void AddMockHandlerContentNotFound(this MockHttpManager httpManage
                 });
         }
 
-        public static MockHttpMessageHandler AddMockHandlerSuccessfulClientCredentialTokenResponseMessage(this MockHttpManager httpManager)
+        public static MockHttpMessageHandler AddMockHandlerSuccessfulClientCredentialTokenResponseMessage(
+            this MockHttpManager httpManager)
```
trwalke, line 167: Are we intending to follow this pattern going forward?

Also, there seem to be a few methods below that are not following this

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -69,26 +69,42 @@ public partial class AuthenticationResult
...
+
                 Account = new Account(
-                    msalAccessTokenCacheItem.HomeAccountId,
+                    homeAccountId,
                     username,
-                    msalAccessTokenCacheItem.Environment);
+                    environment);
+            }
+
+            if (msalAccessTokenCacheItem!= null)
```
trwalke, line 94: So if the user passed in a scope of "openid", this will simply return the ID token only? Is it the same if the user passes in no scopes at all?

henrik-me:
Will review the entire change, however as I tried to understand the issue as well as the description provided for the PR the following came to mind:
Assuming we are returning simply the ID token should an AT not exists? The IDT has an expiry similarly to what an AT has. Hitting the STS continuously doesn't seem desirable, it should be possible to apply similar logic to the IDT as what we already have for the AT.

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccountCacheItem.cs
henrik-me: Something which should be done as optimizations everywhere then? string operations seems to be one of the operations which causes the most CPU and memory usage...

---
In reply to: [401011046](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1735#discussion_r401011046) [](ancestors = 401011046)

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -172,13 +172,6 @@ public static class MsalError
         /// </summary>
         public const string InvalidOwnerWindowType = "invalid_owner_window_type";
 
-        // TODO: InvalidServiceUrl does not seem to be used anywhere?
-
-        /// <summary>
-        /// Invalid service URL.
-        /// </summary>
-        public const string InvalidServiceUrl = "invalid_service_url";
```
henrik-me, line 180: >public const string InvalidServiceUrl = "invalid_service_url"; [](start = 8, length = 62)

Like the cleanup, however perhaps use obsolete instead, it's a breaking change to remove them if used or not?  (comment relevant in a few places)

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -49,7 +49,6 @@ internal static class MsalErrorMessage
         public const string MultipleTokensMatched =
             "The cache contains multiple tokens satisfying the requirements. Try to clear token cache";
 
-        public const string NoDataFromSTS = "No data received from security token service";
```
henrik-me, line 52: see comment in MsalError.cs

trwalke:
What happens if the user passes in { "openid", "profile", "offline_access" }? will MSAL just return an ID token?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -69,26 +69,42 @@ public partial class AuthenticationResult
...
+                AccessToken = authenticationScheme.FormatAccessToken(msalAccessTokenCacheItem);
+                ExpiresOn = msalAccessTokenCacheItem.ExpiresOn;
+                ExtendedExpiresOn = msalAccessTokenCacheItem.ExtendedExpiresOn;
+                Scopes = msalAccessTokenCacheItem.ScopeSet;
+                IsExtendedLifeTimeToken = msalAccessTokenCacheItem.IsExtendedLifeTimeToken;
+                TokenType = msalAccessTokenCacheItem.TokenType;
             }
 
-            AccessToken = authenticationScheme.FormatAccessToken(msalAccessTokenCacheItem);
             UniqueId = msalIdTokenCacheItem?.IdToken?.GetUniqueId();
```
henrik-me, line 104: >UniqueId [](start = 12, length = 8)

seems like this doesn't fallback to the subject in case the IDT is null? Is the doc wrong on the fallback value?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -69,26 +69,42 @@ public partial class AuthenticationResult
...
 
-            if (msalAccessTokenCacheItem.HomeAccountId != null)
+            if (homeAccountId != null)
             {
-                string username = msalAccessTokenCacheItem.IsAdfs ? msalIdTokenCacheItem?.IdToken.Upn : msalIdTokenCacheItem?.IdToken?.PreferredUsername;
+                string username = null;
+                if (msalIdTokenCacheItem != null)
+                {
+                    username = msalIdTokenCacheItem.IsAdfs ? 
+                        msalIdTokenCacheItem?.IdToken.Upn :
```
jennyf19, line 84: >IdToken [](start = 46, length = 7)

null check here as well?

src/client/Microsoft.Identity.Client/Cache/Items/MsalIdTokenCacheItem.cs
henrik-me: (perhaps because it would require additional changes in other places of the code. we generally use the lazy pattern in a number of places. do you prefer methods for this?)

---
In reply to: [401278439](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1735#discussion_r401278439) [](ancestors = 401278439)

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -61,7 +61,8 @@ public TokenClient(AuthenticationRequestParameters requestParams)
                     "ScopeSet was missing from the token response, so using developer provided scopes in the result. ");
             }
 
-            if (!string.Equals(
+            if (!string.IsNullOrEmpty(response.TokenType) &&
```
henrik-me, line 64: >response.TokenType) [](start = 38, length = 19)

when is this null or empty? are we using ""/null for idtoken? should we have an "idtoken" type?

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
henrik-me: @trwalke, do you mean newline after the ( ?

---
In reply to: [401286305](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1735#discussion_r401286305) [](ancestors = 401286305)

trwalke: @henrik-me yes

bgavrilMS: There are no hard limits on the max length of a method / line, but generally I'd say that code should be readable without having to scroll horizontally (on a typical 1080p resolution, with VS having Test Explorer and Solution Explorer open).

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: MSAL always adds the following 3 scopes to a request: `openid`, `email` and `offline_access`. If the user passes  just`openid` it makes no difference to us, we still send the 3 scopes above. 

So passing 0 scopes or any combination of `openid`, `email`, `offline_access` results in the same behaviour:

- AAD issues an access token, id token, refresh token
- B2C does NOT issue an access token, and MSAL crashes with null reference exceptions.

bgavrilMS:
@henrik-me @jmprieur - the contentious point here is the behaviour of AcquireTokenSilent behavior. There are 3 possibilities: 

1. [CURRENT] always perform the `refresh_token` flow
2. do not allow  `AcquireTokenSilent` in this case. Users will have to call `AcquireTokenInteractive` and most likely the browser has session state that can be used
3. use the "exp" claim of the IdToken to perform the strategy of "refresh the id token only if expired". 

Note that this only affects B2C, as AAD seems to generate an AT (still testing on ADFS). I will ping the Android folks to see if they made a decision on this as I remember they support "Login" flow.

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccountCacheItem.cs
bgavrilMS: Yes, and now I've done the same for ClientInfo, which is also a small piece of json that we were parsing like crazy.

Issue #1685, opened 2020-03-06
[Bug] Potential cache consistency issues in multi-threaded env


bgavrilMS:
**Which Version of MSAL are you using ?**
4.9

**Platform**
.net and .net core

**What authentication flow has the issue?**
All, but confidential flows affected the most

**Repro**

In high load enviroments, where the token cache is shared by many threads, cache inconsistency can occur because the access to the underlying token cache is not fully syncronized. In particular, the in-memory cache uses normal dictionaries, which are not thread safe if read and written to at the same time.

This could become a problem if the same token cache is shared between multiple threads (singleton CCA in web apps) or processes (dev tools cache).

Note: we've not yet had users complaining about this.

bgavrilMS:
Example problematic call: 

```csharp
            await RefreshCacheForReadOperationsAsync(CacheEvent.TokenTypes.Account).ConfigureAwait(false);
            return await TokenCacheInternal.GetAccountsAsync(authority, _requestParams.RequestContext).ConfigureAwait(false);

```

GGorelko:
The issues I'm currently facing with similarly are related to the bug above (package version 4.9). The web app is spinning up multiple threads to send requests acquiring access tokens for Confidential client and I can see from time to time exceptions like below (any hints how to work around this)?:

> 
> System.AggregateException: One or more errors occurred. ---> System.ArgumentNullException: Value cannot be null.
> Parameter name: source
>    at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
>    at Microsoft.Identity.Client.TelemetryCore.TelemetryManager.Flush(String correlationId)
>    at Microsoft.Identity.Client.Internal.Requests.RequestBase.<RunAsync>d__14.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at Microsoft.Identity.Client.ApiConfig.Executors.ConfidentialClientExecutor.<ExecuteAsync>d__3.MoveNext()
> --- End of stack trace from previous location where exception was thrown ---
>    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1740, opened 2020-04-03
Fix for #1685 - thread safety in concurrent cache access

bgavrilMS:


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryTokenCacheAccessor.cs
```diff
@@ -8,114 +8,148 @@
...
+        }
+
+        #region Add
         public void SaveAccessToken(MsalAccessTokenCacheItem item)
         {
-            _accessTokenCacheDictionary[item.GetKey().ToString()] = item;
+            string key = item.GetKey().ToString();
+
+            // if a conflict occurs, pick the latest value
+            _accessTokenCacheDictionary.AddOrUpdate(key, item, (k, oldValue) => item);
```
bgavrilMS, line 49: AddOrUpdate, TryRemove etc. are thread safe way of accessing a concurrent dictionary.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -735,9 +735,9 @@ async Task ITokenCacheInternal.ClearAsync()
             {
                 TokenCacheNotificationArgs args = new TokenCacheNotificationArgs(this, ClientId, null, true, (this as ITokenCacheInternal).IsApplicationCache);
 
-                await (this as ITokenCacheInternal).OnBeforeAccessAsync(args).ConfigureAwait(false);
                 try
                 {
+                    await (this as ITokenCacheInternal).OnBeforeAccessAsync(args).ConfigureAwait(false);
```
bgavrilMS, line 740: Moving these inside the try block because if the user adds a callback for OnBeforeAccess that throws exception, then catches them, MSAL will not fire `OnAfterAccessAsync`. This would break the extensions, as it would leave a permanent file lock on disk (and block VS, Powershell etc.)

CC @schaabs

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryTokenCacheAccessor.cs
```diff
@@ -8,114 +8,148 @@
...
+            _logger = logger ?? throw new System.ArgumentNullException(nameof(logger));
+        }
+
+        #region Add
         public void SaveAccessToken(MsalAccessTokenCacheItem item)
         {
-            _accessTokenCacheDictionary[item.GetKey().ToString()] = item;
+            string key = item.GetKey().ToString();
+
+            // if a conflict occurs, pick the latest value
```
bgavrilMS, line 48: As an alternative to this, we could pick the token that has the oldest expiration date ....

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryTokenCacheAccessor.cs
```diff
@@ -8,114 +8,148 @@
...
+            }
         }
 
         public void DeleteRefreshToken(MsalRefreshTokenCacheKey cacheKey)
         {
-            _refreshTokenCacheDictionary.Remove(cacheKey.ToString());
+            if (!_refreshTokenCacheDictionary.TryRemove(cacheKey.ToString(), out _))
+            {
+                _logger.InfoPii(
+                    $"Cannot delete an refresh token because it was already deleted. Key {cacheKey}",
```
pmaytak, line 125: typo: `an` > `a`.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -735,9 +735,9 @@ async Task ITokenCacheInternal.ClearAsync()
...
                     await (this as ITokenCacheInternal).OnBeforeWriteAsync(args).ConfigureAwait(false);
```
henrik-me, line 741: Should this be in try/finally similar to what you have been doing elsewhere?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
henrik-me: is this really doing what you expect? 

---
In reply to: [403098828](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1740#discussion_r403098828) [](ancestors = 403098828)

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: It is, in the finally we need to make sure to call `OnAfterAccess`.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: It's a fair point, I think I can write a unit test that solidify this behaviour.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
henrik-me: how about ensuring both of the onbefore calls?

---
In reply to: [405060863](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1740#discussion_r405060863) [](ancestors = 405060863)

bgavrilMS: I'm not sure. If in `BeforeAccess` the developer throws an exception, does it make sense to still invoke `BeforeWrite` ?

bgavrilMS: In general I don't see the purpose of `BeforeWrite` since we have the `HasCacheChanged` flag on `BeforeAccess`. Just makes things harder to maintain ...

henrik-me: I suppose I agree regarding the need for both events. Should we get rid of it? #deprecate, perhaps an opt-in (or possibly not worth considering)? 

My thinking was we should be guaranteeing both to be called. We don't know what folks are using those events for. Perhaps read locks? 

---
In reply to: [405065050](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1740#discussion_r405065050) [](ancestors = 405065050)

Issue #1653, opened 2020-02-24
[Feature Request] Upgrade library to use AndroidX


tranb3r:
The official stable NuGet packages for AndroidX have been released.
Could you please upgrade the msal lib to use AndroidX?

More information from xamarin here:
https://devblogs.microsoft.com/xamarin/xamarin-forms-xamarin-essentials-go-androidx/

@jennyf19 has kindly listed some scenarios that we should test: 

- [x] Android 9 + interactive/silent auth
   - [x] Embedded webview
   - [x] System browser
- [x] Android 10 + interactive/silent auth
   - [x] Embedded webview
   - [x] System browser
- [x] Broker Android 9 + interactive/silent auth
- [x] Broker Android 10 + interactive/silent auth
- [x] Create a Xamarin Android app from scratch w/new package
  - [x] Embedded webview
  - [x] System browser
- [ ] Update Xamarin samples 
    -  [ ] [B2C Xamarin](https://github.com/Azure-Samples/active-directory-b2c-xamarin-native)
    -  [ ] [Xamarin](https://github.com/Azure-Samples/active-directory-xamarin-native-v2)
- [ ] Send the build to a few customers who are interested in testing before official release

bgavrilMS:
This seems to be the official recommendation from the Xamarin team. All Xamarin libraries should support this. 

It is not clear to me if the upgrade won't break backwards compatibility and also how keeping this version (i.e. not upgrading to AndroidX) affects applications. @tranb3r  - what is your opinion on this?

tranb3r:
My opinion is that:
- most developers are eager to migrate to androidx asap
- backwards compatibility should not be a major issue, assuming you follow the recommendations and add a new target for android10 (with androidx) while keeping the existing android9 target (with support libs).

Anyway, I'd rather ask @redth and @jamesmontemagno for their opinion !

jamesmontemagno:
Our recommendation is to use AndroidX when compile mode is set to Android 10. All AndroidX packages use the migration so they can work side-by-side existing libraries that haven't updated. Mostly this will help when developers are using Xamarin.Forms and Xamarin.Essentials so eventually over time Support Libs will fade away.

tranb3r:
Any progress ?

bgavrilMS:
So this involves providing support for android10 framework. We are currently on android9. There are 2 ways of doing this: 

1. Add a new target to MSAL 

PROS: backwards compatibility
[...]

PR #1745, opened 2020-04-06
GH-1653 AndroidX Support

jamesmontemagno:
#1653

Updated to build for Android 9 and Android 10.

Android 10 supports AndroidX work including the services.

Work to be done would be test integration. This should be easily accomplished by creating 2 head projects. One compiles against 9 and one against 10.


bgavrilMS:
Will take a look as soon as possible. @jamesmontemagno - I assume the best way to test this is to have an app target Android 9 and then move to Android 10 to see the migration?

jamesmontemagno:
Yeah, There is a built in migration tool too: https://docs.microsoft.com/en-us/xamarin/android/platform/androidx

When you have the latest Xamarin.Forms (4.5) and re-target to compile against 10.0 using PackageReferences it will use AndroidX at the core.

neha-bhargava:
@jamesmontemagno  I performed the testing on emulator and it works as expected for embedded web view. But seems like the system browser (chrome) does not have custom tabs. Is it something known with AndroidX?

jamesmontemagno:
I mean it could be possible I missed something in there... 

I did add: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1745/files#diff-9aef39af4a36c0197ee96e49281fcf9cR40 but yeah that should work to open it.

jamesmontemagno:
Should be fixed up, basically ADAL does a check to see if anything can open a custom tab and if not then don't do anything. We have to use the older check.

neha-bhargava:
> Should be fixed up, basically ADAL does a check to see if anything can open a custom tab and if not then don't do anything. We have to use the older check.

There was a check in the app I was using but should not the custom tabs be available anyway? I am using an emulator and the browser is chrome.

jamesmontemagno:
I mean technically users could remove browsers that don't support chrome custom tabs.

In that case it defaults back to non-custom tabs. I am not sure I am in love with this check, but I didn't put it there :)

bgavrilMS:
The phone not having a browser that supports custom tabs is a known issue. There are some phone manufacturers that ship Android phones with such browsers. I believe we have a method called "IsSystemWebViewAvailable" that makes this check

For now, the important thing is that this works on phones with Chrome installed. A second test is to try what happens with Chrome uninstalled - I'd expect a nice exception and that flag to be false. @jennyf19 can confirm this.

jennyf19:
Things should work without custom tabs, here's a list of [browsers](https://github.com/azuread/microsoft-authentication-library-for-dotnet/wiki/Android-system-browser) and their results, this is from a year or two ago, however.
I haven't read through this issue though, if you want me to do so, lmk.

jamesmontemagno:
Correct all of those browser support custom tabs which my change does nothing beside upgrade to the android x version of the library. So whatever worked before should continue to work.

bgavrilMS:
Thanks @jamesmontemagno - we'll have a look at retesting

jennyf19:
Thanks @jamesmontemagno I'll pull your branch and take a look.

jennyf19:
@jamesmontemagno We had some issues w/our test app, which have sorted out now. 
This PR looks good to me. We need to test one more scenario (android + broker).

One question, for you, if I have two apps w/the same clientId, redirectUri, and android scheme on the same device, right now, when using system browser, I get a pop-up asking me which app I want to open. With Android X updates, I'm not seeing that pop-up. Is this expected?
image

jamesmontemagno:
@redth any thoughts^?

bgavrilMS:
@jennyf19 - in your scenario with 2 apps using the same redirect uri, what is the behavior on Android X? Does it freeze?

jennyf19:
@bgavrilMS it just tries to open one of the two apps, so the user doesn't get to select which one, and the user is shown a blank screen.

jamesmontemagno:
When you have multiple apps using the same redirect you should see the pop up from the OS. However, if you already selected one once it may have it selected as the default.

That is just android base OS level and the AndroidX package for chrome custom tabs wouldn't really have any impact there is my thinking.

jennyf19:
@jamesmontemagno that makes sense. I guess you'd go into settings to remove the pre-selected app? just so we can document this better.

jennyf19:
@bgavrilMS @jamesmontemagno I resolved the merge conflict. i think this is ready to go unless we want to test broker first. i don't think it matters too much either way, as we can resolve any issues as they come up if needed. my 2 cents.

bgavrilMS:
Propose to merge and test the broker scenarios as part of our release testing.

bgavrilMS:
Merging this as it will improve our testing.

PR #1746, opened 2020-04-06
Integration tests now capture MSAL logs and HTTP traffic

bgavrilMS:


tests/Microsoft.Identity.Test.Integration.net45/Infrastructure/ClientAppTestExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.Identity.Test.Integration.net45.Infrastructure
+{
+    public static class ClientAppTestExtensions
+    {
+        internal static PublicClientApplicationBuilder WithTestLogging(this PublicClientApplicationBuilder builder)
+        {
+            var httpClientFactory = new HttpSnifferClientFactory();
+
+            return builder
+                .WithLogging((lvl, msg, pii) => Trace.WriteLine($"[MSAL][{lvl}] {msg}"), LogLevel.Verbose, enablePiiLogging: true)
```
pmaytak, line 16: I assume we don't have PII in tests, so default true is fine?

bgavrilMS: Well, I log the Http request and the Http response, so there is some sensitive stuff in there, but related to some test users in the lab, so nothing impactful. 

That said, the test logs live on ADO where they are private. I did not see the test logs being shared from ADO to GitHub / public view.

bgavrilMS: Well spotted, btw.

Issue #901, opened 2019-02-21
`AcquireTokenByIntegratedWindowsAuthAsync(scopes)` should be available on .NET Core, when running on Windows


jmprieur:
**Is your feature request related to a problem? Please describe.**
Partner quote:
*"As a .NET Core or .NET standard developer, if I call an API called `AcquireTokenByIntegratedWindowsAuthAsync(scopes)`, and it doesn’t work because I might not be on windows, that’s a fairly frustrating experience.*

*The kind of work required to get the UPN is something like the following:*

```CSharp
static string GetUPN()
{
 var builder = new StringBuilder(1024);
 var userNameSize = builder.Capacity;
 if (GetUserNameEx(ExtendedNameFormat.UserPrincipal, builder, ref userNameSize) == 0)
 {
  Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
 }
 return builder.ToString();
}
 
[DllImport("secur32.dll", CharSet = CharSet.Auto, SetLastError = true)]
static extern int GetUserNameEx(ExtendedNameFormat nameFormat,
                                                    StringBuilder userName, ref int userNameSize);
 
public enum ExtendedNameFormat
{
    Unknown = 0,
    /// <summary>
    /// The user principal name (for example, someone@example.com).
    /// </summary>
    UserPrincipal = 8,
}
```
 *Lots of customers won’t know how to call the other override (because they won't know how to get the unp), and they’ll just be frustrated with .NET Core. It feels like there shouldn’t be an arbitrary restriction here."*

**Describe the solution you'd like**
On .NET Core, for the Windows platforms, `AcquireTokenByIntegratedWindowsAuthAsync(scopes)` should work (without passing the username). It should throw a meaningful PlatformNotImplemented exception on Mac and Linux

**Additional context**
The .NET Core team is committed to providing an excellent experience with .NET Core 3.0

henrik-me:
Unassigning Travis. Open for all. Not part of current planned release

neha-bhargava:
Fixed in [MSAL 4.12.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.12.0)

PR #1753, opened 2020-04-08
Nebharg/iwa net core

neha-bhargava:
#901


jmprieur:
@jmprieur : to change the docs on docs.ms
@neha-bhargava will you change the wiki?

neha-bhargava:
@jmprieur Sure, I will change the wiki.

PR #1769, opened 2020-04-17
Fixed filtering by upn and uniqueid

petrochuk:
When upn provided as a filter GetAllAdalEntriesForMsal is returning full list when upn is not found. Same problem with uniqueid. This creates issue when cache contains other accounts and library tries to use refresh token from other that account provided in login hint resulting in "Returned user identifier does not match the sent user identifier when saving the token to the cache" errors


jmprieur:
Thanks @petrochuk. Would you mind providing a unit test?

petrochuk:
> Thanks @petrochuk. Would you mind providing a unit test?

Added unit tests

Issue #1770, opened 2020-04-18
[Feature Request] Add ability to disable ADAL CacheFallbackOperations


petrochuk:
**Feature plan**
1. Measure improvement (start with baseline and explore improvements for a small and large cache sizes)
2. Flag to disable ADAL cache fallback (.WithDisableAdalCache - suggestions welcome)
  a. Include this flag in the ApplicationOptions
3. Add telemetry (in .net specific section) to let us know if this feature is used or not.

_Will not be doing:
Auto detect if adal fallback is used, if not disable after first cache call (as Bogdan suggested)_

**Customer challenge:**
Some legacy ADAL operations are very CPU and memory intensive. For example WriteAdalRefreshToken deserializes entire token cache from byte array to Json, adds item and then serializes Json back to byte array. There should be ability through application configuration to disable CacheFallbackOperations and LegacyCachePersistence for applications that do not need it

Our current work around:

`
            // Clearing old ADAL cache prior to each token request to save memory and CPU
            app.UserTokenCache.SetBeforeAccess((args) => args.TokenCache.DeserializeAdalV3(null));
`

bgavrilMS:
I would love to implement this and even set this behaviour as default.

jmprieur:
i agree, @bgavrilMS that we should implement it. 
What about:
`.WithSerializeAdalCache(bool? serialializeAdalV3Cache = null)`

Behavior depending on `serialializeAdalV3Cache`
| serialializeAdalV3Cache | Behavior |
|  ---- | ----- |  
| `null` |  depends on the platform (see below)  | 
| `true` |  Serializes the ADAL cache  |
| `true` |  Serializes the ADAL cache  |

I'm not sure about using `false` by default as this would be a breaking change, but we could be clever:
- serialize by default for Xamarin.iOS, Xamarin.Android, UWP where we provide cache serialization
- do not serialize for .NET Core and .NET Framework where developers need to customize token cache serialization, and throw with a meaningfull exception if `DeserializeAdalV3` and `SerializeAdalV3` are called ?

henrik-me:
@jmprieur : Seems like your table has a small bug as there are 2 entries for true, also I don't understand null (perhaps an enum if we really need more options). 
nit: How about WithAdalCacheSerialization(false) ? 
Also we should add an attribute ensuring we fix this to be the default behavior for the next major version. Also most of our samples should set this especially for web scenarios.

jmprieur:
[...]

PR #1772, opened 2020-04-20
Fix adal cache filtering by unique ID

bgavrilMS:


src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -476,11 +476,10 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
                 return CacheFallbackOperations.GetAdalEntryForMsal(
                     Logger,
                     LegacyCachePersistence,
-                    metadata.PreferredCache,
                     aliases,
                     requestParams.ClientId,
                     upn,
-                    requestParams.Account.HomeAccountId?.Identifier);
+                    requestParams.Account.HomeAccountId?.ObjectId);
```
bgavrilMS, line 482: This was a subtle in an area of the code that wasn;'t tested - how MSAL actually saves the ADAL cache. ADAL cache expects the object ID, MSAL wsas setting the whole "objId.tenantId". 

The code used to work because the original implementation had logic as "if I am given a unique ID, try to filter by it, but if none match, continue with all tokens choose one randomly."

Our cache tests work because they use only 1 token.

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -298,31 +299,30 @@ internal static class CacheFallbackOperations
...
                 upn,
                 uniqueId);
 
-            List<MsalRefreshTokenCacheItem> filteredByPrefEnv = adalRts.Where
-                (rt => rt.Environment.Equals(preferredEnvironment, StringComparison.OrdinalIgnoreCase)).ToList();
+            IEnumerable<IGrouping<string, MsalRefreshTokenCacheItem>> rtGroupsByEnv = adalRts.GroupBy(rt => rt.Environment.ToLowerInvariant());
 
-            if (filteredByPrefEnv.Any())
+            // if we have more than 1 RT per env, there is smth wrong with the ADAL cache
+            if (rtGroupsByEnv.Any(g => g.Count() > 1))
```
bgavrilMS, line 318: ADAL fallback never fails, as we catch all exceptions and log them. But if multiple tokens are found in the cache, it is a legitimate error (someone tampered with the cache) or bug in MSAL

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -772,6 +772,12 @@ public static class MsalError
         /// </summary>
         public const string CertWithoutPrivateKey = "cert_without_private_key";
 
+        /// <summary>
+        /// <para>What happens?</para>The ADAL cache is invalid as it contains multiple refresh token entries for one user.
+        /// <para>Mitigation</para>Delete the ADAL cache. If you do not maintain an ADAL cache, this may be a bug in MSAL.
```
jennyf19, line 777: >this may be a bug in MSA [](start = 96, length = 24)

do we know if it's a bug or not? seems like this is something we should be able to test and know for sure...?

bgavrilMS: If you manually edit a cache file, you could end up with this. Otherwise, with the current code, I don't see how you can. But there so many combinations of tokens in ADAL - with client info, without client info, with / without upn etc. etc. etc. It's impossible to know for sure. I am also doubting if I should add that comment about "possible MSAL bug", but the reality is there are bugs in the ADAL fallback.

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
petrochuk: I agree. This is why we need this: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1770

src/client/Microsoft.Identity.Client/MsalError.cs
petrochuk: I agree. This is why we need this: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1770

PR #1793, opened 2020-04-29
Automation updates

trwalke:
Automation updates for device auth tests


build/pipeline-ci.yaml
```diff
@@ -6,16 +6,33 @@ trigger:
...
+      - msbuild
+      - visualstudio
+    steps:
+
+    - template: template-ci-and-pr.yaml
+
+  - job: 'RunDeviceAuthTests'
+    dependsOn: 'BuildAndTest'
+    pool:
+        name: DevEx Win8 Automation
```
bgavrilMS, line 30: Please add a comment here on why an on-prem agent is needed.

build/pipeline-ci.yaml
```diff
@@ -6,16 +6,33 @@ trigger:
...
+        demands: agent.name -equals MSIDWIN802
```
bgavrilMS, line 31: Seems a bit fragile o use `agent.name`, especially given that there are 3 agents provisioned that can run these tests. Please configure to use the pool instead of a specific agent (e.g. https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/pools-queues?view=azure-devops&tabs=yaml%2Cbrowser)

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            var msalPublicClient = PublicClientApplicationBuilder
+                .Create(_publicClientId)
+                .Build();
+
+            //Act
+            var authResult = msalPublicClient.AcquireTokenByUsernamePassword(new[] { "user.read" }, _deviceAuthuser, new NetworkCredential("", LabUserHelper.FetchUserPassword("msidlab8")).SecurePassword)
+            .WithClaims(JObject.Parse(_claims).ToString())
+            .ExecuteAsync(CancellationToken.None).Result;
+
+            //Assert
```
bgavrilMS, line 35: How do we know this test does anything? I don't see any assertion that indicates that device auth was performed. 

One idea would be to use https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b84322c945e0b8e5f3e237ea9b9e695b50fd1a36/tests/Microsoft.Identity.Test.Integration.net45/Infrastructure/HttpSnifferClientFactory.cs#L11 to interpret the traffic

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
```diff
@@ -0,0 +1,43 @@
...
+    [TestClass]
+    public class DeviceAuthenticationTests
+    {
+        private const string _claims = "{\"access_token\":{\"deviceid\":{\"essential\":true}}}";
+        private const string _publicClientId = "4b0db8c2-9f26-4417-8bde-3f0e3656f8e0";
+        private const string _deviceAuthuser = "idlabca@msidlab8.onmicrosoft.com";
+
+        [TestMethod]
+
+        public void PKeyAuthNonInteractiveTest()
```
bgavrilMS, line 23: What happens if we run this test on a Win10 developer machine? We discussed offline that tests should not fail on developer machines, but it is acceptable to finish as inconclusive via `Assert.Inconclusive`. This is what we do on the cache ext project and works quite well.

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Identity.Test.LabInfrastructure;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Integration.net45.HeadlessTests
+{
+    [TestClass]
+    public class DeviceAuthenticationTests
+    {
+        private const string _claims = "{\"access_token\":{\"deviceid\":{\"essential\":true}}}";
+        private const string _publicClientId = "4b0db8c2-9f26-4417-8bde-3f0e3656f8e0";
```
bgavrilMS, line 18: was under the impression that we should not hardcode client ID and upn - they should come from the lab.

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public void PKeyAuthNonInteractiveTest()
+        {
+            //Arrange
+            var msalPublicClient = PublicClientApplicationBuilder
+                .Create(_publicClientId)
+                .Build();
+
+            //Act
+            var authResult = msalPublicClient.AcquireTokenByUsernamePassword(new[] { "user.read" }, _deviceAuthuser, new NetworkCredential("", LabUserHelper.FetchUserPassword("msidlab8")).SecurePassword)
```
bgavrilMS, line 31: Will there be an `AcquireTokenByIntegratedWindowsAuth` as well?

jennyf19, line 31: >"msidlab8" [](start = 175, length = 10)

this should come from the lab right? +1 on bogdan's earlier comment on clientId.

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            .WithClaims(JObject.Parse(_claims).ToString())
```
jennyf19, line 32: will device auth user always have these claims? i'm trying to think of a failure case. what if you try device auth but don't have the claims? or try with the claims but the wrong user?

build/pipeline-ci.yaml
trwalke: Creating a new pool for just one agent is a bit too much in my opinion. If we were to follow this going forward we would end up with many pools with only 1 agent in them. The only way agent.name will change is if someone removes the agent from the pool entirely. which only I and a few other people have the permission to do which is why using agent.name has never given us issues in the past.

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
trwalke: I could not get this account when I tried using the queries but i just remembered that we can query for accounts directly, Ill update

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
trwalke: I missed this. will add

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
trwalke: Separate item for the moment

trwalke: yes

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
trwalke: To initiate device auth you add this claim specifically. However, I get your point about the failure case. ill try to add

build/pipeline-ci.yaml
bgavrilMS: Not sure I understand what you mean by "creating a new pool". It looks to me that there already exists a pool with 3 agents that were configured for this scenario: 

image

bgavrilMS: My question is really is - why do you have to add ` demands: agent.name -equals MSIDWIN802` - and thus limit this test on one specific agent. Can't these tests run on any of those 3 agents?

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/SilentAuthTests.cs
```diff
@@ -47,6 +47,7 @@ public async Task SilentAuth_ForceRefresh_Async()
         }
 
         [TestMethod]
+        [Ignore]
```
bgavrilMS, line 50: I have an IcM open for this problem, let's not ignore this test.

bgavrilMS:
@trwalke - when you get a bit of time to look at this, let's close it. PRs that stay open for too long are difficult to merge back into master.

build/pipeline-ci.yaml
trwalke: Discussed offline

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/SilentAuthTests.cs
bgavrilMS, line 50: I think this one is already ignored, you'll have to merge.

Issue #1775, opened 2020-04-22
GetAccounts always returns 0 accounts if broker is not installed


harmjonker:
If we do a acquiretoken request with broker=true, what is the exact behavior in case there is no broker installed ? 
I thought it would do the login without broker, but in the 4.11 release we notice that it fails to get a token. Is the intended behavior that it indeed fails and one should retry with broker=false?

bgavrilMS:
Is this Android or iOS , or both?

As far as I remember, if the broker is not installed, MSAL will attempt to use a browser. Note that on Android the redirect uri might be different in this case, so you need to register 2 redirect uris. Or use  the embedded browser via `.WithUseEmbeddedWebUi(true)`.

CC @trwalke as he's looking in this general area now

harmjonker:
We are only developing for Android at the moment and do use .WithUseEmbeddedWebUi. I will look for log fragment to show the case. We don't see it falling back to the browser.

alexander-papen:
Before calling AqcuireTokenSilent(), we query the accounts using GetAccountsAsync()
The returned account is passed as argument to the AquireTokenSilent call.

However, GetAcountsAsync() does not provide any account, see logs:

MSAL Warning False (False) MSAL 4.11.0.0 MSAL.Xamarin.Android 28 [04/22/2020 11:10:12] (UnknownClient: 0.0.0.0) No Authenticator found on the device.
MSAL Error False (False) MSAL 4.11.0.0 MSAL.Xamarin.Android 28 [04/22/2020 11:10:12] (UnknownClient: 0.0.0.0) Android broker is not installed so no accounts will be returned.

Next an AqcuireTokenInteractive is called. 
Although the documentation says that the MSAL lib will redirect the user somehow to the installation location of the broker application,  in fact the WebView login is displayed and the user can login.
It is _'assumed'_ that the MSAL lib wil work now without broker functionality, and the account returned from MSAL lib will be stored/cached by the MSAL lib.

It is expected that the next call to query the accounts, will return the account that was previously loggedin. 
But in stead GetAccountsAsync() returns no accounts. see logs (same as above): 

MSAL Warning False (False) MSAL 4.11.0.0 MSAL.Xamarin.Android 28 [04/22/2020 11:12:49] (UnknownClient: 0.0.0.0) No Authenticator found on the device.
[...]

PR #1797, opened 2020-04-30
Allowing android broker flow to return accounts from local MSAL cache when broker is not installed and WithBroker() is true

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1775

Validated with manual testing.
Adding a new manual test for this to the spreadsheet

I also verified that this is the same behavior on MSAL Andorid

### DIFF
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -91,8 +91,11 @@ namespace Microsoft.Identity.Client
             if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
-                accounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
-                return accounts;
+                if (broker.CanInvokeBroker())
+                {
+                    accounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
+                    return accounts;
+                }
             }
 
             if (UserTokenCache == null)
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -91,8 +91,11 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
             if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
-                accounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
-                return accounts;
+                if (broker.CanInvokeBroker())
```
trwalke, line 94: Checking to see if it is possible to invoke broker at all. if not, this will now continue on to the normal getAccountsAsync functionality W/O broker

jennyf19:
did you already test this w/iOS and Android broker scenarios? if yes, LGTM.

tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/BrokerRequestTests.cs
```diff
@@ -184,6 +184,7 @@ public async Task BrokerGetAccountsTestAsync()
             var mockBroker = Substitute.For<IBroker>();
             var expectedAccount = Substitute.For<IAccount>();
             mockBroker.GetAccountsAsync(TestConstants.ClientId).Returns(new[] { expectedAccount });
+            mockBroker.CanInvokeBroker().Returns(true);
```
jennyf19, line 187: do you have a test where .CanInvokeBroker().Returns(false) and ensures the other route is taken?

trwalke: good point, I added that

PR #1801, opened 2020-05-04
Update codeql.yml

bgavrilMS:


bgavrilMS:
Thx Jenny, but the build isn't working :(. I don't think GitHub's agents have the software that exist on Azure DevOps agents ... I'm going to abandon this for now...

jennyf19:
ok..we have the same issue in identity.web. 
let me know if you figure it out.

PR #1802, opened 2020-05-04
Enabling silent auth fallback for android broker flow

trwalke:
Needed to add additional check to silent request logic so that the broker silent flow will fallback to MSAL when the broker is not installed.

Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1809

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// Indicates whether or not the current application object is configured to use brokered authentication.
         /// </summary>
-        bool IsBrokerConfiguredByUser { get; }
+        bool IsBrokerConfigured { get; }
 
         /// <summary>
         /// The name of the calling application for telemetry purposes.
--- a/src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
@@ -67,8 +67,9 @@ namespace Microsoft.Identity.Client
         IDictionary<string, string> ExtraQueryParameters { get; }
 
         /// <summary>
+        /// Indicates whether or not the current application object is configured to use brokered authentication.
         /// </summary>
-        bool IsBrokerEnabled { get; }
+        bool IsBrokerConfiguredByUser { get; }
 
         /// <summary>
         /// The name of the calling application for telemetry purposes.
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
             {
                 if (_canExecuteBrokerValue == null)
                 {
-                    _canExecuteBrokerValue = AuthenticationRequestParameters.IsBrokerConfiguredByUser && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
+                    _canExecuteBrokerValue = AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
                 }
 
                 return (bool)_canExecuteBrokerValue;
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
@@ -20,7 +20,6 @@ namespace Microsoft.Identity.Client.Internal.Requests
     {
         private readonly AcquireTokenSilentParameters _silentParameters;
         private const string TheOnlyFamilyId = "1";
-        private Lazy<IBroker> _broker;
 
         public SilentRequest(
             IServiceBundle serviceBundle,
@@ -29,7 +28,6 @@ namespace Microsoft.Identity.Client.Internal.Requests
             : base(serviceBundle, authenticationRequestParameters, silentParameters)
         {
             _silentParameters = silentParameters;
-            _broker = new Lazy<IBroker>(() => ServiceBundle.PlatformProxy.CreateBroker(null));
         }
 
         private async Task<IAccount> GetSingleAccountForLoginHintAsync(string loginHint)
@@ -77,9 +75,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         internal async override Task PreRunAsync()
         {
 
-            if (ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth() &&
-                AuthenticationRequestParameters.IsBrokerConfigured &&
-                _broker.Value.CanInvokeBroker())
+            if (AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 return;
             }
@@ -97,7 +93,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         {
             var logger = AuthenticationRequestParameters.RequestContext.Logger;
             MsalAccessTokenCacheItem cachedAccessTokenItem = null;
-            if (ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth() && AuthenticationRequestParameters.IsBrokerConfigured && _broker.Value.CanInvokeBroker())
+            if (AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var msalTokenResponse = await ExecuteBrokerAsync(cancellationToken).ConfigureAwait(false);
                 return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
     {
         private readonly AcquireTokenSilentParameters _silentParameters;
         private const string TheOnlyFamilyId = "1";
+        private bool? _canExecuteBrokerValue;
 
         public SilentRequest(
             IServiceBundle serviceBundle,
@@ -28,6 +29,19 @@ namespace Microsoft.Identity.Client.Internal.Requests
             : base(serviceBundle, authenticationRequestParameters, silentParameters)
         {
             _silentParameters = silentParameters;
+        }
+
+        private bool _canExecuteBroker
+        {
+            get
+            {
+                if (_canExecuteBrokerValue == null)
+                {
+                    _canExecuteBrokerValue = AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
+                }
+
+                return (bool)_canExecuteBrokerValue;
+            }
         }
 
         private async Task<IAccount> GetSingleAccountForLoginHintAsync(string loginHint)
@@ -75,7 +89,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         internal async override Task PreRunAsync()
         {
 
-            if (AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
+            if (_canExecuteBroker)
             {
                 return;
             }
@@ -93,7 +107,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         {
             var logger = AuthenticationRequestParameters.RequestContext.Logger;
             MsalAccessTokenCacheItem cachedAccessTokenItem = null;
-            if (AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
+            if (_canExecuteBroker)
             {
                 var msalTokenResponse = await ExecuteBrokerAsync(cancellationToken).ConfigureAwait(false);
                 return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
@@ -31,13 +31,13 @@ namespace Microsoft.Identity.Client.Internal.Requests
             _silentParameters = silentParameters;
         }
 
-        private bool _canExecuteBroker
+        private bool BrokerIsInstalledAndSupportsSilentAuth
         {
             get
             {
                 if (_canExecuteBrokerValue == null)
                 {
-                    _canExecuteBrokerValue = AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
+                    _canExecuteBrokerValue = AuthenticationRequestParameters.IsBrokerConfiguredByUser && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
                 }
 
                 return (bool)_canExecuteBrokerValue;
@@ -89,7 +89,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         internal async override Task PreRunAsync()
         {
 
-            if (_canExecuteBroker)
+            if (BrokerIsInstalledAndSupportsSilentAuth)
             {
                 return;
             }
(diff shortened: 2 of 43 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -28,6 +29,7 @@ internal class SilentRequest : RequestBase
             : base(serviceBundle, authenticationRequestParameters, silentParameters)
         {
             _silentParameters = silentParameters;
+            _broker = new Lazy<IBroker>(() => ServiceBundle.PlatformProxy.CreateBroker(null));
```
trwalke, line 32: null here indicates that the broker will run in the main context rather than create a NullBroker

jennyf19: i'm not sure about making a broker object for each silent call, esp when it's only used for android. let me think about this.

---
In reply to: [419790230](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419790230) [](ancestors = 419790230)

trwalke: That will not be the case, this will only be created on android and only when the broker is enabled with the WithBroker(true).

This is only initialized when all three checks are made in the conditionals that use it.

example

"ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth() &&
                AuthenticationRequestParameters.IsBrokerConfigured &&
                **_broker**.Value.CanInvokeBroker()"

henrik-me: in that case I believe you should be getting the broker via the service bundle? or an additional param on the silentrequest?

jennyf19: Or can you check in the CanBrokerSupportSilentAuth() method to see if it can be invoked as well? because doesn't matter if it supports it and it can't be invoked right?

---
In reply to: [419799949](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419799949) [](ancestors = 419799949)

jennyf19: I meant the actual _broker object. I don't think we should create that with each silent call. 

---
In reply to: [419798837](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419798837) [](ancestors = 419798837)

trwalke: Yes, that is what i am referring to. since it is a lazy object. it will only be created when it is used on silent broker requests, which can only happen on android when the broker is enabled

henrik-me: or ServiceBundle.PlatformProxy.CanInvokedBroker()

trwalke: > or ServiceBundle.PlatformProxy.CanInvokeBroker()

@henrik-me this is what I am looking at now

jennyf19: Or this:
```csharp
 public override bool CanBrokerSupportSilentAuth()
        {
            IBroker broker = CreateBroker(null);
            if (broker.CanInvokeBroker())
            {
                return true;
            }
            else
            {
                return false;
            }
        }
```

---
In reply to: [419805856](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419805856) [](ancestors = 419805856)

jennyf19: then in ClientApplicationBase...we might have the same issue if broker is not available...idk.
```csharp
 public async Task<IEnumerable<IAccount>> GetAccountsAsync()
        {
            RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
            IEnumerable<IAccount> accounts = Enumerable.Empty<IAccount>();

            {
                if (broker.CanInvokeBroker())
                {
                    accounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
                    return accounts;
                }
            }
...
```
so if you check in CanBrokerSupportSilentAuth that broker is also on the device, you won't have to worry about that here as well.

---
In reply to: [419806308](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419806308) [](ancestors = 419806308,419805856)

trwalke: No, the client application base is doing the same thing the silentReqeust is doing. that is where I made the fix initially.

```CSharp
                if (broker.CanInvokeBroker())
```

trwalke: As you can see in the ClientApplicationBase, "ServiceBundle.PlatformProxy.CreateBroker(null);" will only be called on Android and when it is enabled via WithBroker()

jennyf19: yes i understand. just saying that you can't support silent auth w/out having a broker on the device, so checking in the support silent auth method makes sense to me. idk. i'm just not a fan of the broker as part of the silent request constructor.

---
In reply to: [419808536](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419808536) [](ancestors = 419808536)

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -75,9 +77,12 @@ private async Task<IAccount> GetAccountFromParamsOrLoginHintAsync(AcquireTokenSi
         internal async override Task PreRunAsync()
         {
 
-            if (ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth() && 
-                AuthenticationRequestParameters.IsBrokerConfigured)
+            if (ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth() &&
+                AuthenticationRequestParameters.IsBrokerConfigured &&
```
jennyf19, line 81: nit: you can store this bool and then use it later, so you don't have to do all the checks again in `ExecuteAsync()`

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
trwalke: I agree that checking in the CanBrokerSupportSilentAuth() method is a good idea. reworking the code now.
@jennyf19

jennyf19: ok...i think bogdan will probably have a better idea though. i just would like to move any intialization out of the constructor is all. 


---
In reply to: [419811605](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419811605) [](ancestors = 419811605)

trwalke: he might, but I think it is a good starting point atleast

trwalke: There really isnt a good way of getting around the creation of the new broker object to check if the broker is installed unless we store a static somewhere which I would prefer not to do. putting this logic in the CanBrokerSupportSilentAuth() is a good way to mitigate this issue by ensuring this will only happen on android and when the broker is enabled by the customer. I think this is a good solution with the current design we have.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -30,6 +31,19 @@ internal class SilentRequest : RequestBase
             _silentParameters = silentParameters;
         }
 
+        private bool _canExecuteBroker
```
jennyf19, line 34: We know `PreRunAsync()` will always be called before `ExecuteAsync`, so I would set `private bool _canExecuteBrokerSilentRequest = false` by default, and then in `PreRunAsync()`, if it's true, the value will change, then just check the value of `_canExecuteBrokerRequest` in `ExecuteAsync()`...what do you think?
```csharp
  private void _IsBrokerAvailableForSilentRequest
        {
                if (!_canExecuteBrokerSilentRequest)
                {
                   _canExecuteBrokerSilentRequest = AuthenticationRequestParameters.IsBrokerConfigured && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
                }         
        }
```

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs
```diff
@@ -233,7 +233,16 @@ public override IBroker CreateBroker(CoreUIParent uIParent)
 
         public override bool CanBrokerSupportSilentAuth()
         {
-            return true;
+            IBroker broker = CreateBroker(null);
```
jennyf19, line 236: i like this better. but would like to get BG's thoughts before we merge.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
jennyf19: Are there any tests that can be done for this using the IBroker? Like, it should go through a normal flow and not try to use android, so we can at least test for that, correct? I'm not too familiar w/the android implementation.

---
In reply to: [419826655](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#discussion_r419826655) [](ancestors = 419826655)

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs
```diff
@@ -233,7 +233,16 @@ public override IBroker CreateBroker(CoreUIParent uIParent)
...
+
+            if (broker.CanInvokeBroker())
```
bgavrilMS, line 238: nit: `return broker.CanInvokeBroker()`

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
bgavrilMS: Agreed with testing comment, every line in SilentAuth needs to be unit tested.

bgavrilMS, line 34: Incorrect naming of a property

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -20,6 +20,7 @@ internal class SilentRequest : RequestBase
     {
         private readonly AcquireTokenSilentParameters _silentParameters;
         private const string TheOnlyFamilyId = "1";
+        private bool? _canExecuteBrokerValue;
```
bgavrilMS, line 23: There is a better way to do lazy initialization using `Lazy<T>`. You can see an example here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b84322c945e0b8e5f3e237ea9b9e695b50fd1a36/src/client/Microsoft.Identity.Client/Cache/Items/MsalIdTokenCacheItem.cs#L19

While I'm normally not bothered by doing things "manually", I think using Lazy<T> would make the code more readable.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs
bgavrilMS: Me too.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -75,9 +89,10 @@ private async Task<IAccount> GetAccountFromParamsOrLoginHintAsync(AcquireTokenSi
         internal async override Task PreRunAsync()
         {
 
-            if (ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth() && 
-                AuthenticationRequestParameters.IsBrokerConfigured)
+            if (_canExecuteBroker)
```
bgavrilMS, line 92: I think we need to look at naming. It's getting confusing as we have about 4 verbs around broker existence, invocation etc: canExecute, canSwitchTo, CanBrokerSupportSilentAuth and canInvoke. It's very difficult to reason on what the "code" does. The typical solution for this to define the concepts and associate a single name for each concept: 

1. broker is configured by the user 
2. broker can be invoked - if it exists & we can switch to it 
3. broker supports silent auth 
4. can execute broker which is a combination of 2 & 3

For 4, I would suggest `BrokerIsInstalledAndSupportsSilentAuth()` or smth similar. Have a look at the rest and see if you can consolidate naming.

tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/SilentRequestTests.cs
```diff
@@ -109,43 +109,58 @@ public async Task ExpiredTokenRefreshFlowTestAsync()
...
+        }
+
+        [TestMethod]
+        public void BrokerSilentRequestBrokerNotConfiguredAndNotInstalledTestAsync()
+        {
+            //Broker is not configured by user and is not installed
+            BrokerSilentRequestTestExecutor(false, false);
+        }
+
+        public void BrokerSilentRequestTestExecutor(bool brokerConfiguredByUser, bool brokerIsInstalledAndInvokable)
```
trwalke, line 131: Trying to consolidate test code here.

tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/SilentRequestTests.cs
```diff
@@ -109,43 +109,58 @@ public async Task ExpiredTokenRefreshFlowTestAsync()
 
         [TestMethod]
         public void BrokerSilentRequestBrokerRequiredTestAsync()
+        {
+            //Broker is not configured by user but is installed
```
trwalke, line 113: These tests are testing the various combinations broker configuration values to simulate different authentication scenarios. These scenarios should result in MSAL using its own internal cache as a fallback because the broker is either not installed or not configured by the user.

src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
```diff
@@ -67,8 +67,9 @@ public interface IAppConfig
         IDictionary<string, string> ExtraQueryParameters { get; }
 
         /// <summary>
+        /// Indicates whether or not the current application object is configured to use brokered authentication.
         /// </summary>
-        bool IsBrokerEnabled { get; }
+        bool IsBrokerConfiguredByUser { get; }
```
jennyf19, line 72: >IsBrokerConfiguredByUser [](start = 13, length = 24)

nit: idk about this renaming. mainly  because the "user" does not configure the broker, the app developer does. So it's slightly misleading. @bogavril@microsoft.com do you have any ideas?

bgavrilMS: `IsBrokerConfigured` is probably enough ...

tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -93,7 +93,7 @@ public static SortedSet<string> s_scope
 
         public const string DisplayableId = "displayable@id.com";
         public const string RedirectUri = "urn:ietf:wg:oauth:2.0:oob";
-        public const string MobileDefaultRedirectUri = "msal4a1aa1d5-c567-49d0-ad0b-cd957a47f842://auth"; // in msidentity-samples-testing tenant -> PublicClientSample
+        public const string MobileDefaultRedirectUri = "msal4a1aa1d5-c567-99d0-ad0b-cd957a470042://auth"; // in msidentity-samples-testing tenant -> PublicClientSample
```
jennyf19, line 96: >99d0 [](start = 74, length = 4)

check these values.

tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/SilentRequestTests.cs
```diff
@@ -109,43 +109,58 @@ public async Task ExpiredTokenRefreshFlowTestAsync()
...
             using (var harness = new MockHttpTestHarness(TestConstants.AuthorityHomeTenant))
             {
-                _tokenCacheHelper.PopulateCache(harness.Cache.Accessor);
-                harness.ServiceBundle.PlatformProxy.SetBrokerForTest(CreateMockBroker());
+                _tokenCacheHelper.PopulateCache(harness.Cache.Accessor,
+                    TestConstants.Uid,
+                    TestConstants.Utid,
+                    TestConstants.ClientId,
+                    TestConstants.ProductionPrefCacheEnvironment,
+                    "Broker@broker.com");
```
jennyf19, line 140: nit: can you use a const here, just in this method is fine.

jennyf19:
still not convinced on the names, but otherwise lgtm. thanks, travis.

---
In reply to: [405755732](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1802#issuecomment-405755732) [](ancestors = 405755732)

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidBroker.cs
```diff
@@ -38,7 +38,7 @@ public AndroidBroker(CoreUIParent uiParent, ICoreLogger logger)
             _brokerHelper = new AndroidBrokerHelper(Application.Context, logger);
         }
 
-        public bool CanInvokeBroker()
+        public bool BrokerIsInstalledAndInvokable()
```
jennyf19, line 41: could this be `IsBrokerInstalledAndInvokable()`?

trwalke: sure

jennyf19:
thanks @trwalke :)

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
bgavrilMS: Would have been nice to address this as suggested.

PR #1804, opened 2020-05-05
Bogavril/ui req throt

bgavrilMS:
UI required throttling provider. 

Spec: https://identitydivision.visualstudio.com/devex/_git/AuthLibrariesApiReview?version=GBdev&path=%2FService%20protection%2FIntial%20set%20of%20protection%20measures.md&_a=preview

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -614,6 +614,12 @@ namespace Microsoft.Identity.Client
         /// </summary>
         public const string ClientCredentialAuthenticationTypesAreMutuallyExclusive = "Client_Credential_Authentication_Types_Are_Mutually_Exclusive";
 
+        /// <summary>
+        /// <para>What happens?</para>You configured MSAL confidential client authentication without an authentication type (Certificate, Secret, Client Assertion)
+        /// <para>Mitigation</para>Either call ConfidentialClientApplicationBuilder.WithClientSecret, ConfidentialClientApplicationBuilder.WithCertificate, ConfidentialClientApplicationBuilder.WithClientAssertion
+        /// </summary>
+        public const string ClientCredentialAuthenticationTypeMustBeDefined = "Client_Credentials_Required_In_Confidential_Client_Application";
+
         #region InvalidGrant suberrors
         /// <summary>
         /// Issue can be resolved by user interaction during the interactive authentication flow.
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -19,6 +19,15 @@ namespace Microsoft.Identity.Client
         /// </summary>
         public const string InvalidGrantError = "invalid_grant";
 
+        /// <summary>
+        /// Standard OAuth2 protocol error code. It indicates that the application needs to expose the UI to the user
+        /// so that the user does an interactive action in order to get a new token.
+        /// <para>Mitigation:</para> If your application is a <see cref="T:IPublicClientApplication"/> call <c>AcquireTokenInteractive</c>
+        /// perform an interactive authentication. If your application is a <see cref="T:ConfidentialClientApplication"/> chances are that the Claims member
+        /// of the exception is not empty. See <see cref="P:MsalServiceException.Claims"/> for the right mitigation
+        /// </summary>
+        public const string InteractionRequired = "interaction_required";
+
 #if !DESKTOP && !NET_CORE
 #pragma warning disable CS1574 // XML comment has cref attribute that could not be resolved
 #endif
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Identity.Client
 
         /// <summary>
         /// Standard OAuth2 protocol error code. It indicates that the application needs to expose the UI to the user
-        /// so that the user does an interactive action in order to get a new token.
+        /// so the user can do an interactive log-in to get a token with updated claims.
         /// <para>Mitigation:</para> If your application is a <see cref="T:IPublicClientApplication"/> call <c>AcquireTokenInteractive</c>
         /// perform an interactive authentication. If your application is a <see cref="T:ConfidentialClientApplication"/> chances are that the Claims member
         /// of the exception is not empty. See <see cref="P:MsalServiceException.Claims"/> for the right mitigation
--- a/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
+++ b/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
@@ -24,19 +24,17 @@ namespace Microsoft.Identity.Client
             MsalServiceException ex = null;
             var oAuth2Response = JsonHelper.TryToDeserializeFromJson<OAuth2ResponseBase>(httpResponse?.Body);
 
-            if (string.Equals(oAuth2Response?.Error, MsalError.InvalidGrantError, StringComparison.OrdinalIgnoreCase))
+            if (IsInvalidGrant(oAuth2Response) || IsInteractionRequired(oAuth2Response))
             {
-                if (IsUiInteractionRequired(oAuth2Response?.SubError))
-                {
-                    ex = new MsalUiRequiredException(errorCode, errorMessage, innerException);
-                }
+                ex = new MsalUiRequiredException(errorCode, errorMessage, innerException);
             }
+
 
             if (string.Equals(oAuth2Response?.Error, MsalError.InvalidClient, StringComparison.OrdinalIgnoreCase))
             {
                 ex = new MsalServiceException(
                     MsalError.InvalidClient,
-                    MsalErrorMessage.InvalidClient + " Original exception: " + oAuth2Response?.ErrorDescription, 
+                    MsalErrorMessage.InvalidClient + " Original exception: " + oAuth2Response?.ErrorDescription,
                     innerException);
             }
 
@@ -57,7 +55,19 @@ namespace Microsoft.Identity.Client
             return ex;
         }
 
-        internal static bool IsUiInteractionRequired(string subError)
+        private static bool IsInteractionRequired(OAuth2ResponseBase oAuth2Response)
+        {
+            return string.Equals(oAuth2Response?.Error, MsalError.InvalidGrantError, StringComparison.OrdinalIgnoreCase);
+
+        }
+
+        private static bool IsInvalidGrant(OAuth2ResponseBase oAuth2Response)
+        {
+            return string.Equals(oAuth2Response?.Error, MsalError.InvalidGrantError, StringComparison.OrdinalIgnoreCase)
+                             && IsInvalidGrantSubError(oAuth2Response?.SubError);
+        }
+
+        private static bool IsInvalidGrantSubError(string subError)
         {
             if (string.IsNullOrEmpty(subError))
             {
--- a/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
+++ b/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
@@ -57,8 +57,7 @@ namespace Microsoft.Identity.Client
 
         private static bool IsInteractionRequired(OAuth2ResponseBase oAuth2Response)
         {
-            return string.Equals(oAuth2Response?.Error, MsalError.InvalidGrantError, StringComparison.OrdinalIgnoreCase);
-
+            return string.Equals(oAuth2Response?.Error, MsalError.InteractionRequired, StringComparison.OrdinalIgnoreCase);
         }
 
         private static bool IsInvalidGrant(OAuth2ResponseBase oAuth2Response)
--- a/src/client/Microsoft.Identity.Client/MsalThrottledUiRequiredException.cs
+++ b/src/client/Microsoft.Identity.Client/MsalThrottledUiRequiredException.cs
@@ -6,7 +6,7 @@ namespace Microsoft.Identity.Client
     /// <summary>
     /// Exception type thrown when MSAL detects that an application is trying to acquire a token even 
     /// though an <see cref="MsalUiRequiredException"/> was recently thrown. 
-    /// To mitigate this, when an <see cref="MsalUiRequiredException"/> is encountered,
+    /// To mitigate this, when a <see cref="MsalUiRequiredException"/> is encountered,
     /// the application should switch to acquiring a token interactively. To better understand
     /// why the <see cref="MsalUiRequiredException" /> was thrown, inspect the <see cref="MsalUiRequiredException.Classification"/>
     /// property.
--- a/src/client/Microsoft.Identity.Client/MsalThrottledUiRequiredException.cs
+++ b/src/client/Microsoft.Identity.Client/MsalThrottledUiRequiredException.cs
@@ -4,24 +4,28 @@
 namespace Microsoft.Identity.Client
 {
     /// <summary>
-    /// Exception type thrown when MSAL detects that an application is trying to acquire a token too often, as a result of: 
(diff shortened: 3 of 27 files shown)


src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -24,19 +24,17 @@ internal class MsalServiceExceptionFactory
             MsalServiceException ex = null;
             var oAuth2Response = JsonHelper.TryToDeserializeFromJson<OAuth2ResponseBase>(httpResponse?.Body);
 
-            if (string.Equals(oAuth2Response?.Error, MsalError.InvalidGrantError, StringComparison.OrdinalIgnoreCase))
+            if (IsInvalidGrant(oAuth2Response) || IsInteractionRequired(oAuth2Response))
```
bgavrilMS, line 27: Turns out that MSALs should also throw `UI Req Exception` if error_code is "interaction_required"

src/client/Microsoft.Identity.Client/MsalThrottledUiRequiredException.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// To mitigate this, when an <see cref="MsalUiRequiredException"/> is encountered,
+    /// the application should switch to acquiring a token interactively. To better understand
+    /// why the <see cref="MsalUiRequiredException" /> was thrown, inspect the <see cref="MsalUiRequiredException.Classification"/>
+    /// property.
+    /// 
+    /// The properties of this exception are identical to the original exception
+    /// 
+    /// For more details see https://aka.ms/msal-net-throttling
+    /// </summary>
+    public class MsalThrottledUiRequiredException : MsalUiRequiredException
```
bgavrilMS, line 18: @rayluo - not sure if the spec makes it clear, but @SomkaPe argued for having explicit exceptions. In java / .net we used inheritance to not break people. In python it's up to you, but it should be clear that throttling occurs otherwise it will be difficult to fix support cases etc.

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -19,6 +19,15 @@ public static class MsalError
         /// </summary>
         public const string InvalidGrantError = "invalid_grant";
 
+        /// <summary>
+        /// Standard OAuth2 protocol error code. It indicates that the application needs to expose the UI to the user
+        /// so that the user does an interactive action in order to get a new token.
```
jennyf19, line 24:
```suggestion
        /// so the user can do an interactive log-in to get a token with updated claims.
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -19,6 +19,15 @@ public static class MsalError
...
+        /// <para>Mitigation:</para> If your application is a <see cref="T:IPublicClientApplication"/> call <c>AcquireTokenInteractive</c>
+        /// perform an interactive authentication. If your application is a <see cref="T:ConfidentialClientApplication"/> chances are that the Claims member
```
jennyf19, line 26:
```suggestion
        /// to perform an interactive authentication. If your application is a <see cref="T:ConfidentialClientApplication"/> chances are that the Claims member
```

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
jennyf19: makes sense.

---
In reply to: [420305866](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1804#discussion_r420305866) [](ancestors = 420305866)

src/client/Microsoft.Identity.Client/MsalThrottledUiRequiredException.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Exception type thrown when MSAL detects that an application is trying to acquire a token even 
+    /// though an <see cref="MsalUiRequiredException"/> was recently thrown.
```
jennyf19, line 8:
```suggestion
    /// though a <see cref="MsalUiRequiredException"/> was recently thrown. 
```

src/client/Microsoft.Identity.Client/MsalThrottledUiRequiredException.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Exception type thrown when MSAL detects that an application is trying to acquire a token even 
+    /// though an <see cref="MsalUiRequiredException"/> was recently thrown. 
+    /// To mitigate this, when an <see cref="MsalUiRequiredException"/> is encountered,
```
jennyf19, line 9:
```suggestion
    /// To mitigate this, when a <see cref="MsalUiRequiredException"/> is encountered,
```

trwalke:
>     /// <item>Afetr receving an RetryAfter header</item>

spelling

---
Refers to: src/client/Microsoft.Identity.Client/OAuth2/Throttling/SingletonThrottlingManager.cs:18 in d57d82b. [](commit_id = d57d82b319971aca7d55d2d57aea2e1045746cff, deletion_comment = False)

src/client/Microsoft.Identity.Client/OAuth2/Throttling/RetryAfterProvider.cs
```diff
@@ -4,20 +4,19 @@
...
 using Microsoft.Identity.Client.Internal.Requests;
-using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 
 namespace Microsoft.Identity.Client.OAuth2.Throttling
 {
     internal class RetryAfterProvider : IThrottlingProvider
     {
 
-        internal ThrottlingCache Cache { get; } // internal for test only
+        internal ThrottlingCache ThrottlingCache { get; } // internal for test only
```
jennyf19, line 13: my dyslexia is kicking in an i keep reading throttling cattle. lol.

tests/Microsoft.Identity.Test.Unit.net45/ExceptionTests/MsalExceptionTests.cs
```diff
@@ -17,6 +17,7 @@
 
 namespace Microsoft.Identity.Test.Unit.ExceptionTests
 {
+
     [TestClass]
```
trwalke, line 21: nit: empty line above

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
```diff
@@ -0,0 +1,118 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
jennyf19, line 2:
```suggestion
// Licensed under the MIT License.

```

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
```diff
@@ -0,0 +1,118 @@
...
+                    $"[Throttling] Exception thrown because of throttling rule UiRequired - thumbprint: {thumbprint}",
+                    $"[Throttling] Exception thrown because of throttling rule UiRequired ");
+
+                // mark the exception for logging purposes                
+                throw new MsalThrottledUiRequiredException(uiException);
+            }
+        }
+
+        private static bool IsSupported(AuthenticationRequestParameters requestParams)
+        {
```
jennyf19, line 81: nit: maybe a comment here or earlier about this check.

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
```diff
@@ -0,0 +1,118 @@
...
+                logger.WarningPii(
+                    $"[Throttling] Exception thrown because of throttling rule UiRequired - thumbprint: {thumbprint}",
+                    $"[Throttling] Exception thrown because of throttling rule UiRequired ");
+
+                // mark the exception for logging purposes                
+                throw new MsalThrottledUiRequiredException(uiException);
+            }
+        }
+
+        private static bool IsSupported(AuthenticationRequestParameters requestParams)
```
trwalke, line 80: I think IsRequestSupported() might be a more clear name

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
```diff
@@ -0,0 +1,118 @@
...
+                logger.Info($"[Throttling] MsalUiRequiredException encountered - " +
+                    $"throttling for {s_uiRequiredExpiration.TotalSeconds} seconds");
+
+                var thumbprint = GetRequestStrictThumbprint(bodyParams,
+                    requestParams.AuthorityInfo.CanonicalAuthority,
+                    requestParams.RequestContext.ServiceBundle.PlatformProxy.CryptographyManager);
+                var entry = new ThrottlingCacheEntry(ex, s_uiRequiredExpiration);
+                ThrottlingCache.AddAndCleanup(thumbprint, entry, logger);
+            }
+        }
```
jennyf19, line 44: can you explain when we throttle based on the ui required exception and how it's different from the regular ui exception we throw after AT S?

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            if (bodyParams.TryGetValue(OAuth2Parameter.RefreshToken, out string rt) && 
+                !string.IsNullOrEmpty(rt))
+            {
+                sb.Append(crypto.CreateSha256Hash(rt) + ThrottleCommon.KeyDelimiter);
+            }
+
+            return sb.ToString();
+        }
+    }
```
jennyf19, line 117: Does B2C support throttling?

tests/Microsoft.Identity.Test.Unit.net45/Throttling/ThrottlingAcceptanceTests.cs
```diff
@@ -233,10 +250,113 @@ await app.AcquireTokenSilent(new[] { "Other.Scopes" }, account).ExecuteAsync()
...
+        /// Sleep for at least DefaultUIRequired + 1 seconds.
+        /// Repeat the same request again - expectation is the request should be issued on the server.
+        /// If a request had the response with Retry After header, then a different request with a different strict thumbprint should work
+        /// </summary>
+        [TestMethod]
+        public async Task UiRequiredThrottling_Async()
+        {
+            using (var httpManagerAndBundle = new MockHttpAndServiceBundle())
+            {
+                var app = await SetupAndAquireOnceAsync(
```
jennyf19, line 269: sp: `Acquire`

tests/Microsoft.Identity.Test.Unit.net45/Throttling/ThrottlingAcceptanceTests.cs
```diff
@@ -233,10 +250,113 @@ await app.AcquireTokenSilent(new[] { "Other.Scopes" }, account).ExecuteAsync()
...
+                throttlingManager.SimulateTimePassing(
+                    UiRequiredProvider.s_uiRequiredExpiration + 
+                    TimeSpan.FromSeconds(1));
+
+                httpManagerAndBundle.HttpManager.AddTokenResponse(TokenResponseType.Valid);
+                await app.AcquireTokenSilent(TestConstants.s_scope, account).ExecuteAsync()
+                       .ConfigureAwait(false);
+
+                AssertThrottlingCacheEntryCount(throttlingManager, uiRequiredEntryCount: 0);
+            }
```
jennyf19, line 306: can the uiRequiredEntryCount ever get above 1?

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                sb.Append((clientId ?? "") + ThrottleCommon.KeyDelimiter);
+            }
+            sb.Append(authority + ThrottleCommon.KeyDelimiter);
+            if (bodyParams.TryGetValue(OAuth2Parameter.Scope, out string scopes))
+            {
+                sb.Append((scopes ?? "") + ThrottleCommon.KeyDelimiter);
+            }
+
+            if (bodyParams.TryGetValue(OAuth2Parameter.RefreshToken, out string rt) &&
```
trwalke, line 109: so it looks like we are using the RT to create a thumbprint for each of the cached exceptions. How does this work with the android broker because it does not return an RT?

jennyf19:
Looks good. Might have a few questions for you later in the week. thx.

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
bgavrilMS: Sure. The idea behind client-side throttling is to prevent developers from repeating some requests that **we know** will fail. 

Imagine the following scenario: 

- you app retries `AcquireTokenSilent` a few times. If it fails, it goes to `AcquireTokenInteractive`
- the tenant admin enables MFA
- because of MFA, existing refresh tokens become unusable
- when trying the refresh_token grant, AAD errors out with "invalid_grant" "MFA was enabled, please login interactively"
- a badly written app ignores this and simply retries with the same refresh token. Over and over. These requests are destined to fail, but they still put stress on AAD. 

With client throttling, **MSAL** detects that the application is trying to make the same bad request again and stop the request before it reaches AAD. This reduces the load on AAD, without affecting customer experience (hopefully).

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
bgavrilMS: Client-side throttling is based on the following signals from the token providres: 

- HTTP 429 and RetryAfter header
- HTTP 5xx
- HTTP 400 and error_code of "invalid_grant" / "interaction_required" (which MSAL translates as MsalUiRequiredException). 

Now, I don't know if B2C sends RetryAfter headers. I think it does send "invalid_grant" errors though. 

Definitely something to follow up with B2C folks, thanks for pointing this out @jennyf19

tests/Microsoft.Identity.Test.Unit.net45/Throttling/ThrottlingAcceptanceTests.cs
bgavrilMS: Yes. I will add a test to show this. The entries in the dictionary depend on: 

- client ID
- authority
- refresh token
- scopes

So if you make 2 requests that result in a UIRequired exception, and the 2 requests differ because of any of the above params, there will be 2 entries. 

Added a test where we request fewer scopes - it shows 2 entries.

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
bgavrilMS: This is a great question! I was initially thinking to check if "IsBrokerRequest" but then I realized that broker requests do not use the "TokenClient". So we should be good. 

The broker requests communicate with brokers through different methods, not using HTTP calls. TokenClient class, where all the throttling methods are, are not used by broker.

src/client/Microsoft.Identity.Client/OAuth2/Throttling/UiRequiredProvider.cs
jennyf19: Got it. Thanks. very helpful.

PR #1807, opened 2020-05-07
Update some log statements and stop logging default client name

bgavrilMS:


src/client/Microsoft.Identity.Client/Internal/Broker/BrokerSilentRequest.cs
```diff
@@ -83,25 +81,22 @@ internal void CreateRequestParametersForBroker()
 
         internal void ValidateResponseFromBroker(MsalTokenResponse msalTokenResponse)
         {
-            _authenticationRequestParameters.RequestContext.Logger.Info(LogMessages.CheckMsalTokenResponseReturnedFromBroker);
+            _logger.Info(LogMessages.CheckMsalTokenResponseReturnedFromBroker);
             if (msalTokenResponse.AccessToken != null)
             {
-                _authenticationRequestParameters.RequestContext.Logger.Info(
-                    LogMessages.BrokerResponseContainsAccessToken +
-                    msalTokenResponse.AccessToken.Count());
```
bgavrilMS, line 91: AccessToken.Count() = the number of chars in an access token. 
Log message looks like this "Token response contains access token count 2035" 
App developer and @bgavrilMS  super confused.

src/client/Microsoft.Identity.Client/Internal/MsalLogger.cs
```diff
@@ -38,7 +38,7 @@ internal class MsalLogger : ICoreLogger
             ClientVersion = clientVersion ?? string.Empty;
 
             ClientInformation = string.Empty;
-            if (!string.IsNullOrEmpty(clientName))
+            if (!string.IsNullOrEmpty(clientName) && !ApplicationConfiguration.DefaultClientName.Equals(clientName))
```
bgavrilMS, line 41: This will stop logging that "UnknownClient - 0.0.0.0" on every line.

bgavrilMS:
Thx for looking folks

PR #1808, opened 2020-05-07
Enable retry-after throttling for all grants

bgavrilMS:


src/client/Microsoft.Identity.Client/OAuth2/Throttling/HttpStatusProvider.cs
```diff
@@ -66,5 +66,10 @@ public void TryThrottle(AuthenticationRequestParameters requestParams, IReadOnly
                 ThrottleCommon.TryThrowServiceException(strictThumbprint, ThrottlingCache, logger, nameof(HttpStatusProvider));
             }
         }
+
+        private static bool IsRequestSupported(AuthenticationRequestParameters requestParameters)
+        {
+            return !requestParameters.IsConfidentialClient;
```
bgavrilMS, line 72: Throttling based on HTTP 5xx responses is still only for public clients.

tests/Microsoft.Identity.Test.Unit.net45/Throttling/ThrottlingAcceptanceTests.cs
```diff
@@ -163,6 +163,48 @@ public async Task RetryAfter_LargeTimeoutHeader_DefaultTimeout_IsUsed_Async()
                 Assert.AreEqual(actualExpiration, RetryAfterProvider.MaxRetryAfter);
             }
         }
+
+        [TestMethod]
+        public async Task RetryAfter_ConfidentialClient_Async()
```
bgavrilMS, line 168: Unit test proving that this works for client_credentials grant

PR #1810, opened 2020-05-12
use previous clientId w/mobile apps

jennyf19:
This default clientId will not work w/any non-broker flow, as it doesn't have the correct redirect uri registered and the android manifest was not updated. I have added the android broker redirect uri to this client Id in the portal, so everything should work with this one client id.


jennyf19:
fyi @neha-bhargava

Issue #1750, opened 2020-04-07
[Feature Request] OIDC metadata cache should ignore tenants


bgavrilMS:
**Which Version of MSAL are you using ?**
4.11

**Platform**
all

**Scenario**

1. AcquireToken* using tenant1
2. AcquireToken* using tenant2

**Expected behavior**
A single call to OIDC metadata should be performed

**Actual behavior**
2 calls are performed (one for each tenant)

**Possible Solution**
We have discussed with the server-side folks and they have agreed that we can cache by environment only.

It might be possible to not call OIDC at all as we only use the /token and /authorize endpoint URIs, which have a standard format. But if an STS does not support /authorize, we would not know about it in a nice way (i.e. some STS-es return "unsupported"). Not a big deal though.

Still undecided: do we want to allow developers to bypass this behaviour and get the per-tenant metadata? Perhaps this can be a separate task?

bgavrilMS:
@jmprieur @henrik-me - I think I'm going to go one step further and not call OIDC at all? This is what Java does.  I talked to Jason about possible impact on cross-cloud scenarios, and there shouldn't be any. Thoughts?

We can add an option to toggle this later on, if needed, smth like: 

`.WithOIDCDiscovery(Discovery.PerTenant)` with possible values `None` (default), `PerTenant` and `PerAuthority`.

jmprieur:
@bgavrilMS  In which cases would we skip it? for Azure AD?
Do we care if we don't get the issuer aliases? (I though they were needed to know the preferred one?

henrik-me:
@jmprieur we would always skip it. Suggesion from Bogdan is though to potentially add functionality that would enable calling the OIDCDiscovery endpoint. Not having to make this call is a huge performance gain. We don't really need anything from there any way... will be a benefit in many scenarios. Not sure if there will be any sort of impact with the work for the CloudInstanceAware feature? @bgavrilMS Which end-point do we talk to for that info?

henrik-me:
Aside: ADAL does this per Authority today.

bgavrilMS:
I am proposing to skip the call to get the OIDC metadata: 

https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration

We only use the `authorization_endpoint` and `token_endpoint` from there

The logic around instance metadata (i.e. aliases and authority validation) remains the same.

jmprieur:
[...]

PR #1816, opened 2020-05-13
Fix for #1750 - Hardcode tenant and authorize endpoints instead of OI…

bgavrilMS:
…DC discovery


src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -5,9 +5,9 @@
 
 namespace Microsoft.Identity.Client.Instance
 {
-    /// <summary>
-    ///
-    /// </summary>
+    /// <remaks>
```
jennyf19, line 8:
```suggestion
    /// <remarks>
```

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -5,9 +5,9 @@
...
+    /// Must be kept immutable
+    /// </remaks>
```
jennyf19, line 10:
```suggestion
    /// </remarks>
```

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -196,14 +196,19 @@ internal static AuthorityType GetAuthorityType(string authority)
...
         /// <summary>
         /// Gets a tenanted authority if the current authority is tenantless.
         /// Returns the original authority on B2C and ADFS
         /// </summary>
         internal abstract string GetTenantedAuthority(string tenantId);
 
+        /// <summary>
+        /// Gets the authority endpoints based on the host and tenant id. Does not rely on OIDC discovery.
+        /// </summary>
+        internal abstract AuthorityEndpoints GetHardcodedEndpoints();
```
jennyf19, line 210: nit: just for discussion...these are actually the authorization server endpoints...Does it make sense to call them what they are, verse authority endpoints? just seems confusing. really just a discussion point here.

src/client/Microsoft.Identity.Client/Instance/AuthorityEndpoints.cs
```diff
@@ -9,11 +9,11 @@ namespace Microsoft.Identity.Client.Instance
 {
     internal class AuthorityEndpoints
     {
-        public AuthorityEndpoints(string authorizationEndpoint, string tokenEndpoint, string selfSignedJwtAudience)
+        public AuthorityEndpoints(string authorizationEndpoint, string tokenEndpoint)
         {
             AuthorizationEndpoint = authorizationEndpoint;
             TokenEndpoint = tokenEndpoint;
-            SelfSignedJwtAudience = selfSignedJwtAudience;
+            SelfSignedJwtAudience = tokenEndpoint;
```
jennyf19, line 16: this is the tokenEndpoint?

src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
```diff
@@ -14,6 +12,9 @@ internal class AadAuthority : Authority
         public const string DefaultTrustedHost = "login.microsoftonline.com";
         public const string AADCanonicalAuthorityTemplate = "https://{0}/{1}/";
 
+        private const string TokenEndpointTemplate = "{0}oauth2/v2.0/token";
+        private const string AuthorizationEndpointTemplate = "{0}oauth2/v2.0/authorize";
```
jennyf19, line 16: since these are the same for both AAD and ADFS, can they be moved to a constants class?

src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
```diff
@@ -56,5 +53,19 @@ internal override string GetTenantedAuthority(string tenantId)
...
         }
+
+        internal override AuthorityEndpoints GetHardcodedEndpoints()
+        {
+            string tokenEndpoint = string.Format(
+                    CultureInfo.InvariantCulture,
+                    TokenEndpointTemplate,
+                    AuthorityInfo.CanonicalAuthority);
+
+            string authEndpoint = string.Format(CultureInfo.InvariantCulture,
```
jennyf19, line 64: >authEndpoint [](start = 19, length = 12)

nit: call this `authorizationEndpoint`

src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
bgavrilMS: These are slightly different between AAD and ADFS, notice the "v2.0" part.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
bgavrilMS: Happy to rename them, but maybe not Authorization endpoints because "authoirzation endpoint" is one of them. Myabe OAuth2Endpoints?

jennyf19: OAuth2Endpoints make sense.

Issue #1791, opened 2020-04-29
[Bug] When ValidateAuthority is false, bypass instance discovery call


neha-bhargava:
**Which Version of MSAL are you using ?**
Note that to get help, you need to run the latest version. Preview version are also ok. 
For ADAL, please log issues to https://github.com/AzureAD/azure-activedirectory-library-for-dotnet

Latest

**Platform**

**What authentication flow has the issue?**
* Desktop / Mobile
    * [X] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**

**Repro**

```csharp
IPublicClientApplication pca = PublicClientApplicationBuilder
                            .Create(s_clientIdForPublicApp)
                            .WithAuthority(new Uri("https://login.fake.com/common/"), false)
                            .WithLogging(Log, LogLevel.Verbose, true)
                            .WithRedirectUri("http://localhost") // required for DefaultOsBrowser
                            .Build();
```

**Expected behavior**
Should bypass the instance discovery call when flag is passed as false

**Actual behavior**
Still validates the authority and errors out with "invalid_instance" error code.

**Possible Solution**

**Additional context/ Logs / Screenshots**
Add any other context about the problem here, such as logs and screebshots. Logging is described at https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging

henrik-me:
When fixing this, utests should be improved to cover these scenarios, possibly also an addition to e2e testing.

bgavrilMS:
We have an E2E test that verifies behaviour when validate=true, should be easy to repurpose for validate=false as well.

PR #1817, opened 2020-05-13
Nebharg/1791

neha-bhargava:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1791

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Authority.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Authority.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Identity.Client.Instance
         }
 
 
-        public static Authority CreateAuthority(string authority, bool validateAuthority = true)
+        public static Authority CreateAuthority(string authority, bool validateAuthority = false)
         {
             return CreateAuthority(AuthorityInfo.FromAuthorityUri(authority, validateAuthority));
         }
--- a/src/client/Microsoft.Identity.Client/Instance/Authority.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Authority.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Identity.Client.Instance
         }
 
 
-        public static Authority CreateAuthority(string authority, bool validateAuthority = false)
+        public static Authority CreateAuthority(string authority, bool validateAuthority = true)
         {
             return CreateAuthority(AuthorityInfo.FromAuthorityUri(authority, validateAuthority));
         }
--- a/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
@@ -109,10 +109,11 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
                      new NetworkCredential("", user.GetOrFetchPassword()).SecurePassword)
                      .ExecuteAsync())
                 .ConfigureAwait(false);
-
-
+#if NET_CORE
+            Assert.IsTrue(((System.Net.Sockets.SocketException)exception.InnerException).SocketErrorCode == System.Net.Sockets.SocketError.HostNotFound);
+#else
             Assert.IsTrue(((System.Net.WebException) exception.InnerException).Status == System.Net.WebExceptionStatus.NameResolutionFailure);
-
+#endif
         }
 
 
--- a/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
@@ -88,6 +88,33 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
             Assert.AreEqual("invalid_instance", exception.ErrorCode);
         }
 
+        [TestMethod]
+        public async Task AuthorityValidationTestWithFalseValidateAuthorityAsync()
+        {
+            LabResponse labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
+            LabUser user = labResponse.User;
+
+            IPublicClientApplication pca = PublicClientApplicationBuilder
+                .Create(labResponse.App.AppId)
+                .WithAuthority("https://bogus.microsoft.com/common", false)
+                .WithTestLogging()
+                .Build();
+
+            Trace.WriteLine("Acquire a token using a not so common authority alias");
+
+            HttpRequestException exception = await AssertException.TaskThrowsAsync<HttpRequestException>(() =>
+                 pca.AcquireTokenByUsernamePassword(
+                    s_scopes,
+                     user.Upn,
+                     new NetworkCredential("", user.GetOrFetchPassword()).SecurePassword)
+                     .ExecuteAsync())
+                .ConfigureAwait(false);
+
+
+            Assert.IsTrue(((System.Net.WebException) exception.InnerException).Status == System.Net.WebExceptionStatus.NameResolutionFailure);
+
+        }
+
 
         /// <summary>
         /// If this test fails, please update the <see cref="KnownMetadataProvider"/> to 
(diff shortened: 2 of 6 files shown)


src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -117,7 +117,7 @@ private static void ValidateTypeMismatch(AuthorityInfo configAuthorityInfo, Auth
         }
 
 
-        public static Authority CreateAuthority(string authority, bool validateAuthority = false)
+        public static Authority CreateAuthority(string authority, bool validateAuthority = true)
```
jennyf19, line 120: >validateAuthority = true [](start = 71, length = 24)

i think we want to leave the deafult value as false, otherwise it would be a breaking change. @jmprieur to confirm

neha-bhargava: I can revert it back, changed it while testing and was not intending to commit it :)

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
```diff
@@ -88,6 +88,33 @@ public async Task FailedAuthorityValidationTestAsync()
...
+            HttpRequestException exception = await AssertException.TaskThrowsAsync<HttpRequestException>(() =>
+                 pca.AcquireTokenByUsernamePassword(
+                    s_scopes,
+                     user.Upn,
+                     new NetworkCredential("", user.GetOrFetchPassword()).SecurePassword)
+                     .ExecuteAsync())
+                .ConfigureAwait(false);
+
+
+            Assert.IsTrue(((System.Net.WebException) exception.InnerException).Status == System.Net.WebExceptionStatus.NameResolutionFailure);
```
jennyf19, line 114: we don't get a better exception here?

neha-bhargava: No. What happens here is first there will be network discovery and if that fails it will check the validateAuthority flag. If the flag is false it will go directly to the uri. So, the authority here is a fake uri and that is what the exception is. In a normal scenario there will be no exception and a token will be generated.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
jennyf19: yes, it should stay as it was before. thanks. :)

jennyf19:
I'm not entirely sure, but I think the validate authority problem is happening [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/920e72c089eee861b290eeeb1295cddfede83e6b/src/client/Microsoft.Identity.Client/Instance/AuthorityEndpointResolutionManager.cs#L48_) and on line 54. Looking at Bogdan's PR, I think @bgavrilMS might have fixed this issue [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1816/files#diff-ae7149cb8388176de026ee4ff7dc5a7dR25) because we no longer need the OIDC .well-known endpoint, the /authorize /token endpoints will be hard coded. I would try pulling his branch and see the experience w/validateAuthority=false.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
jennyf19: we should catch that exception and through an MsalServiceException (as this is a service issue, but maybe a clientException... @bgavrilMS what do you think?). See my [this comment](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1817#issuecomment-628342946) first though :)

bgavrilMS: `MsalServiceException` is meant to describe AAD service errors. We have a contract with AAD and we know how to interpret these errors (i.e. read the JSON, with error_code, message etc.)

In this case, the error isn't coming from AAD. The request never reaches AAD. It goes into the great unknown that is the Internet and it dies a 404 Not Found death. MSAL is doing the right thing to not try to interpret this error - it just lets it bubble up to the developer. After all, it's the developer who used "bogus.microsoft.com" as authority, what do they expect? :D

So I think this behavior is ok. The test shows that instance discovery isn't blocking request.

As for disabling OIDC, you are right Jenny, I made changes. But I think my changes are much more extensive, so I will rebase them over Neha's change rather than the other way around.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
bgavrilMS: yes, needs to be reverted back :)

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
```diff
@@ -194,6 +194,8 @@ public void FailedValidationTest()
                 Authority instance = Authority.CreateAuthority("https://login.microsoft0nline.com/mytenant.com", true);
                 Assert.IsNotNull(instance);
                 Assert.AreEqual(instance.AuthorityInfo.AuthorityType, AuthorityType.Aad);
+
+                TestCommon.CreateServiceBundleWithCustomHttpManager(harness.HttpManager, authority: instance.AuthorityInfo.CanonicalAuthority, validateAuthority: true);
```
bgavrilMS, line 198: is this enough? I think you have to take the service bundle that you create here and use it to build that resolver object at line 199

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
bgavrilMS, line 114: @neha-bhargava - the PR build failed and it shows that a slightly different exception occurs on .NET Core. This is likely to do subtle differerences in the HTTP stack used by .NET classic (net45) and .NET core. I have a PR out to update the .netcore tests to 3.1 so that you can run them locally (well, it's a problem all of us are starting to have).

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/AadAuthorityTests.cs
bgavrilMS: This is only working because you changed the default ValidateAuthority in TestCommon from `false` to `true`. I think this new default is ok, but I don't think any other test changes are needed.

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -47,6 +48,7 @@ public override void TestInitialize()
             };
 
             _harness = base.CreateTestHarness();
+            TestCommon.CreateServiceBundleWithCustomHttpManager(_harness.HttpManager, validateAuthority: true);
```
bgavrilMS, line 51: I think you are still missing a unit test that should verify the behavior of the InstanceManager when validateAuthority is set to false.

bgavrilMS:
Please make sure that when you merge, the commit message has a short description, not just the issue number.

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
bgavrilMS: Example test that shows how to inject a custom authority: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/fe547af3f0fd0404f32ad38496467808f124d549/tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/B2cAuthorityTests.cs#L67

neha-bhargava: Added.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
neha-bhargava: reverted.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/AuthorityTests.cs
```diff
@@ -88,6 +88,34 @@ public async Task FailedAuthorityValidationTestAsync()
...
+            Trace.WriteLine("Acquire a token using a not so common authority alias");
+
+            HttpRequestException exception = await AssertException.TaskThrowsAsync<HttpRequestException>(() =>
+                 pca.AcquireTokenByUsernamePassword(
+                    s_scopes,
+                     user.Upn,
+                     new NetworkCredential("", user.GetOrFetchPassword()).SecurePassword)
+                     .ExecuteAsync())
+                .ConfigureAwait(false);
+#if NET_CORE
```
neha-bhargava, line 112: There is a difference in exceptions thrown by net standard and net core for unknown host. I have added a condition here. Not sure if this is the best practice.

bgavrilMS: This works fine, thanks Neha.

PR #1826, opened 2020-05-15
Stop hardcoding .net core SDK in builds

bgavrilMS:


bgavrilMS:
Getting this in as I just verified the release build is no longer broken.

Issue #1825, opened 2020-05-15
[Bug] Better error reporting when IWA is failing is needed


jabbera:
**Which Version of MSAL are you using ?**
4.13.0

**Platform**
net462

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [X] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**
This is a test app

**Repro**

```csharp
using Microsoft.Identity.Client;
using System;
using System.Linq;
using System.Net;
using System.Net.Http;
using System.Threading;
using System.Threading.Tasks;

namespace ConsoleApp12
{
    class Program
    {
        private const string CLIENT_ID = "<CLIENTID>"; // MAKE THIS YOURS
        private const string TENANT_ID = "<TENANTID>";
        private static IPublicClientApplication clientApplication; // Thread safe;

        static async Task Main(string[] args)
        {
            try
            {
                clientApplication = await CreatePublicClientWithCacheAsync();

                var token = await GetToken().ConfigureAwait(false);
                Console.WriteLine(token);
            }
            catch (Exception e)
            {
                Console.WriteLine(e);
                throw;
            }
            finally
            {
                Console.ReadLine();
            }
        }

        private static async Task<string> GetToken()
        {
            string[] scopes = new string[] { "https://database.windows.net//user_impersonation" };
            AuthenticationResult result;
            try
            {
                result = await clientApplication.AcquireTokenByIntegratedWindowsAuth(scopes).ExecuteAsync();
            }
            catch (MsalUiRequiredException)
            {
                result = await clientApplication.AcquireTokenInteractive(scopes).ExecuteAsync();
            }

            return result.AccessToken;
        }

        private static async Task<IPublicClientApplication> CreatePublicClientWithCacheAsync()
        {
            IPublicClientApplication pca = await CreatePublicClientAndBindCacheAsync()
                .ConfigureAwait(false);

            return pca;
        }

        private static async Task<IPublicClientApplication> CreatePublicClientAndBindCacheAsync()
        {

            var appBuilder = PublicClientApplicationBuilder.Create(CLIENT_ID)
                .WithTenantId(TENANT_ID)
                .WithHttpClientFactory(new StaticClientWithProxyFactory());

[...]

PR #1838, opened 2020-05-21
On WsTrust parse error, return body

jabbera:
This returns the body of the response if there is a parse error. It would have been really helpful to realize what this issue was. Instead I had to compile the library from source and debug it.

This drove me crazy. I kept getting an error attempting to Seamless SSO that spawned this issue here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1825

Turns out our firewall was intercepting the traffic and categorizing the SSSO site as "uncategorized" which requires a manual step to approve. It was basically impossible for me to determine this was an issue because the body was not being returned when there was a parse error.

### DIFF
--- a/src/client/Microsoft.Identity.Client/WsTrust/WsTrustWebRequestManager.cs
+++ b/src/client/Microsoft.Identity.Client/WsTrust/WsTrustWebRequestManager.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Identity.Client.WsTrust
             {
                 string message = string.Format(
                         CultureInfo.CurrentCulture,
-                        MsalErrorMessage.FederatedServiceParseErrorTemplate,
+                        MsalErrorMessage.ParsingWsTrustResponseFailedErrorTemplate,
                         wsTrustEndpoint.Uri,
                         resp.Body);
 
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/WsTrust/WsTrustWebRequestManager.cs
```diff
@@ -101,8 +101,14 @@ public async Task<MexDocument> GetMexDocumentAsync(string federationMetadataUrl,
...
             {
+                string message = string.Format(
+                        CultureInfo.CurrentCulture,
+                        MsalErrorMessage.FederatedServiceParseErrorTemplate,
+                        wsTrustEndpoint.Uri,
+                        resp.Body);
+
                 throw new MsalClientException(
-                    MsalError.ParsingWsTrustResponseFailed, MsalError.ParsingWsTrustResponseFailed, ex);
+                    MsalError.ParsingWsTrustResponseFailed, message, ex);
```
jennyf19, line 111: this is awesome. Thanks @jabbera for the contribution. Do you know if there would be PII data ever in the response body?
LGTM.

jabbera: Well, it's Microsoft's endpoint so out of the box maybe you can tell me since I haven't been able to get my firewall team to sort this yet:-)

As for what our firewall puts in, it's internal addresses and usernames, but I'd guess that's specific to each company.

jabbera: Renamed the error template to match the error being raised. (Seems to be your convention) and added a test. Happy to squash if you want.

jennyf19: Thanks! We can squash the commits when we merge. Appreciate the text coverage.

PR #1857, opened 2020-06-02
New dev app using Android 10

bgavrilMS:


tests/devapps/Xamarin Manual/Xamarin Manual/Xamarin Manual/MainPage.xaml.cs
```diff
@@ -0,0 +1,142 @@
...
+            try
+            {
+                var accounts = await _pca.GetAccountsAsync().ConfigureAwait(false);
+                var result = await _pca.AcquireTokenSilent(AuthConfig.Scopes, accounts.FirstOrDefault()).ExecuteAsync().ConfigureAwait(false);
+                UpdateStatus($"ATS Token! Valid for: {(result.ExpiresOn - DateTimeOffset.Now).TotalMinutes} min");
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(e);
+                UpdateStatus("Failed! " + e);
```
jennyf19, line 48: maybe mentioned which btn was hit to generate the exception in the error message?

tests/devapps/Xamarin Manual/Xamarin Manual/Xamarin Manual/MainPage.xaml.cs
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+
+        // this is for testing purposes only, product code should AcquireTokenSilent and if that fails
+        // it should fallback to AcquireTokenInteractive
+        private async void atiBtn_Clicked(object sender, EventArgs ea)
```
jennyf19, line 54: should you have a btn for ats and then ati?

bgavrilMS:
The customer facing samples are also upgraded to Android Q.

Samples are different than dev projects because they focus on simplicity and learning. Dev projects are focused on ease of testing. For example, I have AcquireTokenSilent and AcquireTOkenInteractive as 2 separate buttons here, but this is a bad idea for  sample.

PR #1866, opened 2020-06-05
Add logging with time measurments to broker code paths

bgavrilMS:


src/client/Microsoft.Identity.Client/Internal/Logger/DurationLogHelper.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Diagnostics;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.Internal.Logger
+{
+    internal sealed class DurationLogHelper : IDisposable
```
bgavrilMS, line 10: @neha-bhargava - the IDisposable pattern is very common in C# (somewhat similar to AutoCloseable in Java)... because it plays very nicely with the `using` statement - so code like is quite common.

`using (var dh = new DurationHelper())` is syntactic sugar for 

```csharp
try
{
var dh = new DurationHelper();
}
finally
{
  dh.Dispose();
}
```

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
```diff
@@ -71,51 +74,40 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
         {
             try
             {
-                if (brokerPayload.ContainsKey(BrokerParameter.BrokerInstallUrl))
```
bgavrilMS, line 74: Very minor refactoring - just moved some blocks of code into private methods.

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
```diff
@@ -233,5 +234,15 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
 
             return sb.ToString();
         }
+
+        public DurationLogHelper LogBlockDuration(string measuredBlockName, LogLevel logLevel = LogLevel.Verbose)
+        {
+            return new DurationLogHelper(this, measuredBlockName, logLevel);
+        }
+
+        public DurationLogHelper LogMethodDuration(LogLevel logLevel = LogLevel.Verbose, [CallerMemberName] string methodName = null)
```
bgavrilMS, line 243: Another C# trick ... [CallerMemberName] will populate the param (methodName) with the name of calling method. This is done by the compiler, not at runtime (reflection is very slow)

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=netcore-3.1

CC @jennyf19 in case you'll want to add some perf logging to Identity.Web

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
```diff
@@ -71,51 +74,40 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
...
+
+            using (_logger.LogBlockDuration("waiting for broker response"))
+            {
+                await s_readyForResponse.WaitAsync().ConfigureAwait(false);
+            }
+        }
 
+        private async Task AcquireTokenInteractiveViaBrokerAsync(IDictionary<string, string> brokerPayload)
+        {
+            using (_logger.LogMethodDuration())
```
bgavrilMS, line 105: So instead of doing smth like 

`_logger.LogMethodDuration("AcquireTokenInteractiveViaBrokerAsync")` 

we can just rely on the compiler to fill in the name ....

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
trwalke: very cool

---
In reply to: [436182007](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1866#discussion_r436182007) [](ancestors = 436182007)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
trwalke, line 105: I see, so it looks like you are creating a using block for the code that you want to test the time of and once the code execution is finished, the DurationLogHelper is disposed of and the elapsed timer will be logged when that happens right??

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
jennyf19: That's cool. asp.net core has something similar for logging.

---
In reply to: [436182007](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1866#discussion_r436182007) [](ancestors = 436182007)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
```diff
@@ -138,13 +130,43 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
...
+                string silentResult = await _brokerHelper.GetBrokerAuthTokenSilentlyAsync(brokerPayload, _activity).ConfigureAwait(false);
+                if (!string.IsNullOrEmpty(silentResult))
+                {
+                    return MsalTokenResponse.CreateFromAndroidBrokerResponse(silentResult, s_correlationId);
+                }
+
+                return new MsalTokenResponse
+                {
+                    Error = MsalError.BrokerResponseReturnedError,
+                    ErrorDescription = "Failed to acquire token silently from the broker." + MsalErrorMessage.AndroidBrokerCannotBeInvoked,
```
jennyf19, line 151:
```suggestion
                    ErrorDescription = "Android Broker: Failed to acquire token silently from the broker. " + MsalErrorMessage.AndroidBrokerCannotBeInvoked,
```

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
```diff
@@ -138,13 +130,43 @@ private async Task AcquireTokenInternalAsync(IDictionary<string, string> brokerP
...
+
+        private async Task<MsalTokenResponse> AcquireTokenSilentViaBrokerAsync(IDictionary<string, string> brokerPayload)
+        {
+            // Don't send silent background request if account information is not provided
+
+            using (_logger.LogMethodDuration())
             {
-                _logger.ErrorPiiWithPrefix(ex, "Broker invocation failed.");
-                throw;
+                _logger.Verbose("User is specified for silent token request. Starting silent broker request.");
```
jennyf19, line 141:
```suggestion
                _logger.Verbose("Android Broker: User is specified for silent token request. Starting silent broker request. ");
```

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
bgavrilMS: Yes exactly. At the end of the `using` block, the CLR will invoke the `Dispose()` method, which simply logs how many seconds have elapsed.

PR #1869, opened 2020-06-08
add  method to get accounts by userflow for B2C scenarios

jennyf19:
**Context:**
In AAD B2C, each policy is a unique and separate authorization server, which means that for each policy used, a new set of tokens will be in the cache. 
For example, user logs-in and goes to sign-up/sign-in policy (aka. authorization server 1) and gets tokens. Then user goes to edit profile policy (aka. authorization server 2) and gets tokens. MSAL now has 2 accounts in the cache one for sign-up/sign-in (auth server 1) and one for edit profile (auth server 2).

Currently, developers have to write the following code:
```csharp
private IAccount GetAccountByPolicy(IEnumerable<IAccount> accounts, string policy)	
{	
       foreach (var account in accounts)	
       {	
            string accountIdentifier = account.HomeAccountId.ObjectId.Split('.')[0];	
            if (accountIdentifier.EndsWith(policy.ToLower())) return account;	
       }	

        return null;	
}
```
in order to have the right account for the matching policy. For example, in B2C, to call a web api, generally you use the access token from the sign-up/sign-in policy as you want to ensure the user has already signed-in before accessing the api (but you could require any policy, this is done in the portal). 

So the developers need to filter the cache by policy and use the correct one.

The ObjectId contains the policy used: `HomeAccountId = {AccountId: fb61c8f4-9234-413e-bcbf-44c74ec450dc-b2c_1_susi.775527ff-9a37-4307-8b3d-cc311f58d925}`

In MS Identity Web, we need to incorporate this logic as well, so I figured it was time to move it to MSAL .NET.

If you need more information on this, let me know. Happy to provide more context. 
fyi: @henrik-me


jennyf19:
Updated sample w/the new api: https://github.com/Azure-Samples/active-directory-b2c-dotnet-desktop/pull/58

@jmprieur thoughts on this?

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -129,6 +129,31 @@ public async Task<IAccount> GetAccountAsync(string accountId)
             return accounts.FirstOrDefault(account => account.HomeAccountId.Identifier.Equals(accountId, StringComparison.OrdinalIgnoreCase));
         }
 
+        /// <summary>
+        /// Get the <see cref="IAccount"/> by its identifier among the accounts available in the token cache,
+        /// based on the user flow. This is for AAD B2C scenarios.
```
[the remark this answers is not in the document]
jennyf19: ok. thanks. i'll make those changes locally and push them up.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -129,6 +129,31 @@ public async Task<IAccount> GetAccountAsync(string accountId)
             return accounts.FirstOrDefault(account => account.HomeAccountId.Identifier.Equals(accountId, StringComparison.OrdinalIgnoreCase));
         }
 
+        /// <summary>
+        /// Get the <see cref="IAccount"/> by its identifier among the accounts available in the token cache,
+        /// based on the user flow. This is for Azure AD B2C scenarios.
+        /// </summary>
+        /// <param name="userFlow">User flow identifier. The identifier is the user flow being targeted by the specific B2C authority/>.
+        /// </param>
+        public async Task<IAccount> GetAccountByUserFlowAsync(string userFlow)
```
bgavrilMS, line 138: We don't have the "UserFlow" concept in MSALs as of now. Can we get agreement from other MSALs on introducing a new concept in the library? Or was this discussed and approved already by the B2C working group?

I am also unsure if filtering by objectId is a robust mechanism - was this discussed with server side folks?

jennyf19: we were told by b2c team to use "userFlow" in regards to the policy. this is to reduce the load on the developer, as msal should be handling this. When we designed the cache, B2C was overlooked, so we pushed this onto the developer. I don't want this in identity web. I cad hide it behind a feature flag if necessary.

> I am also unsure if filtering by objectId is a robust mechanism - was this discussed with server side folks?

Then why are we forcing customers to rely on this to filter the cache?

jennyf19: btw...msal .net doesn't have a concept of `KnownAuthorities` but other libs do, like obj-c and js.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -129,6 +129,31 @@ public async Task<IAccount> GetAccountAsync(string accountId)
...
+                throw new ArgumentException($"{nameof(userFlow)} should not be null or whitespace", nameof(userFlow));
+            }
+
+            var accounts = await GetAccountsAsync().ConfigureAwait(false);
+
+            foreach (var account in accounts)
+            {
+                string accountIdentifier = account.HomeAccountId.ObjectId.Split('.')[0];
+                if (accountIdentifier.EndsWith(userFlow.ToLowerInvariant()))
+                    return account;
```
bgavrilMS, line 151: A few more comments: 

- Is there only 1 account per userflow? Can't we have multiple?
- You can use `.EndsWith("stuff", System.StringComparison.OrdinalIgnoreCase);` for a more robust comparison. 
- nit, but it's cleaner to use Linq: 

```csharp
return accounts.Where(acc => acc.HomeAccountId.ObjectId.Split('.').EndsWith(...));
```

jennyf19: addressed this and moved to an extension method.

tests/Microsoft.Identity.Test.Unit.net45/CoreTests/InstanceTests/B2cAuthorityTests.cs
```diff
@@ -141,10 +141,10 @@ public void CanonicalAuthorityInitTest()
             string UriNoPort = TestConstants.B2CAuthority;
             string UriNoPortTailSlash = TestConstants.B2CAuthority;
 
-            const string UriDefaultPort = "https://login.microsoftonline.in:443/tfp/tenant/policy";
+            const string UriDefaultPort = "https://login.microsoftonline.in:443/tfp/tenant/b2c_1_susi";
 
-            const string UriCustomPort = "https://login.microsoftonline.in:444/tfp/tenant/policy";
-            const string UriCustomPortTailSlash = "https://login.microsoftonline.in:444/tfp/tenant/policy/";
+            const string UriCustomPort = "https://login.microsoftonline.in:444/tfp/tenant/b2c_1_susi";
```
henrik-me, line 146: >b2c_1_susi" [](start = 90, length = 11)

constants exists for the userflows

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
henrik-me: KnownAuthorities is a known and accepted concept which all libraries are supposed to implement as part of sdk-consistency.

The current implementation makes the GetAccounts method is internal, didn't you mean to make it public?

We should discuss whether this functionality should be added to MSAL or not.

bgavrilMS:
Are we sufficiently aligned to merge this so as to get it in next MSAL release?

jennyf19:
@bgavrilMS thanks! this is awesome! i'll help update samples

Issue #636, opened 2018-10-12
Handle null intent in SetAuthenticationContinuationEventArgs (MSAL 2.1.0)


nhsimarsden:
**Is your feature request related to a problem? Please describe.**
_SetAuthenticationContinuationEventArgs_ is the method used in when implementing MSAL for Xamarin Andorid, which handles the return from an interactive sign-in. It takes three arguments, requestCode (int), resultCode (Result) and data (Intent).

I am working with Xamarin.Forms and experiencing an issue on Android where the user triggers interactive sign-in (PublicClientApplication.AcquireTokenAsync), taps the home button, then launches the app again from the apps menu (rather than resuming from the overview). This results in the Intent being null and thus SetAuthenticationContinuationEventArgs fails.

Consider the example provided in the wiki:

```
protected override void OnActivityResult(int requestCode, Result resultCode, Intent data)
{
    base.OnActivityResult(requestCode, resultCode, data);
    AuthenticationContinuationHelper.SetAuthenticationContinuationEventArgs(requestCode, resultCode, data);
}
```

Following the described steps, **`data`** will be null and, if unhandled, the app will crash.

**Describe the solution you'd like**
It would be good to have MSAL throw an `MsalException` (or variant) so that I can handle this scenario gracefully within my sign in logic flow.

**Describe alternatives you've considered**
Right now I am checking for a null intent within MainActivity.OnActivityResult and simply not calling SetAuthenticationContinuationEventArgs.

**Additional context**
Steps to reproduce the scenario:
1) Launch app and start interactive signin (where the browser is shown)
2) Tap the device home button
3) Open the apps menu and open the app from the menu

nhsimarsden:
Rather than not calling `SetAuthenticationContinuationEventArgs`, another workaround is to create a dummy intent:

```
if (data == null)
{
    data = new Intent("ReturnFromEmbeddedWebview");
}
```

This is better than not calling SetAuthenticationContinuationEventArgs (which is breaking my sign-in logic). The only problem I can see here is that the interaction doesn't continue (for example, a B2C password reset which requires the user to enter a verification code that they receive via email).

bgavrilMS:
@nhsimarsden - When you say that the app crashes when the intent is null, do you mean that MSAL throws a null reference exception or some other exception that causes the app to crash? If that is the case, then we should treat this as a bug.

nhsimarsden:
[...]

PR #1870, opened 2020-06-08
Add check for null intent

neha-bhargava:
#636

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
@@ -33,6 +33,12 @@ namespace Microsoft.Identity.Client
             var logger = MsalLogger.Create(Guid.Empty, null);
             logger.Info(string.Format(CultureInfo.InvariantCulture, "Received Activity Result({0})", (int)resultCode));
 
+            if (data == null)
+            {
+                logger.Info("Null intent was received.");
+                return;
+            }
+
             AuthorizationResult authorizationResult;
             if (data.Action != null && data.Action.Equals("ReturnFromEmbeddedWebview", StringComparison.OrdinalIgnoreCase))
             {


src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -33,6 +33,12 @@ public static void SetAuthenticationContinuationEventArgs(int requestCode, Resul
             var logger = MsalLogger.Create(Guid.Empty, null);
             logger.Info(string.Format(CultureInfo.InvariantCulture, "Received Activity Result({0})", (int)resultCode));
 
+            if (data == null)
+            {
+                logger.Info("Null intent was received.");
```
bgavrilMS, line 38:
```suggestion
                logger.Info("SetAuthenticationContinuationEventArgs - ignoring intercepted null intent.");
```

src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -33,6 +33,12 @@ public static void SetAuthenticationContinuationEventArgs(int requestCode, Resul
             var logger = MsalLogger.Create(Guid.Empty, null);
             logger.Info(string.Format(CultureInfo.InvariantCulture, "Received Activity Result({0})", (int)resultCode));
 
+            if (data == null)
```
bgavrilMS, line 36: Having looked more at this, I am not sure this is the correct solution. There are cases where we can interpret the result exclusively from `Result resultCode`, for example for user cancelled.

trwalke: @neha-bhargava Im assuming that line 43 is actually throwing the null ref right? if that is the case, then the code that handles the Result resultCode exclusively (i.e. ProcessFromEmbeddedWebview) will never be reached if there is a null returned. Meaning, there must be some value in data when user cancels otherwise that scenario would throw an exception every time if data is null. 

with that being said I think it might be ok to leave it as is. 

Thoughts @bgavrilMS

bgavrilMS: I taled to @neha-bhargava offline. We can be more defensive and check for data != null. But there are cases where can do better, for example for authentication cancelled we only need the Result.

neha-bhargava: Added logic to include scenarios where data is not required.

src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -34,20 +35,25 @@ public static void SetAuthenticationContinuationEventArgs(int requestCode, Resul
...
                 // that indicates that the response returned to this class is for the broker.
             {
                 AndroidBroker.SetBrokerResult(data, (int)resultCode);
                 return;
             }
-            else
+            else if (data != null || AndroidConstants.AuthCodeReceived != (int)resultCode)
             {
                 authorizationResult = ProcessFromSystemWebview(requestCode, resultCode, data);
+            } else
```
bgavrilMS, line 53: Please have the else on a separate line.

src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -34,20 +35,25 @@ public static void SetAuthenticationContinuationEventArgs(int requestCode, Resul
...
 
             AuthorizationResult authorizationResult;
-            if (data.Action != null && data.Action.Equals("ReturnFromEmbeddedWebview", StringComparison.OrdinalIgnoreCase))
+
+            if (data != null && data.Action != null && data.Action.Equals("ReturnFromEmbeddedWebview", StringComparison.OrdinalIgnoreCase))
             {
                 authorizationResult = ProcessFromEmbeddedWebview(requestCode, resultCode, data);
             }
-            else if (!String.IsNullOrEmpty(data.GetStringExtra(BrokerConstants.BrokerResultV2)) || requestCode == BrokerConstants.BrokerRequestId) 
+            else if (data != null && (!String.IsNullOrEmpty(data.GetStringExtra(BrokerConstants.BrokerResultV2)) || requestCode == BrokerConstants.BrokerRequestId))
```
bgavrilMS, line 43: nit: You can simplify the logic here if you remove the `else` clauses and just set the result

neha-bhargava: But we do not want the subsequent if clauses to run if it passes this condition right?

src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
neha-bhargava: Java habit :)

src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -5,6 +5,7 @@
 using System.Globalization;
 using Android.App;
 using Android.Content;
+using Java.Sql;
```
bgavrilMS, line 8: What is this "using" statement?

src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
bgavrilMS: Yes, anyway, not important, let's leave it.

Issue #1820, opened 2020-05-14
[Bug] AcquireTokenSilent / GetAccounts / RemoveAccounts + Broker is not consistent with MSAL.Android behaviour


bgavrilMS:
**Which Version of MSAL are you using ?**
4.13

**Platform**
Android

**Expected behavior**

GetAccounts should fetch local accounts from MSAL cache and the broker accounts and merge them by home_account_id.
RemoveAccount should remove the account and associted tokens from both local cache and from the broker
AcquireTokenSilent should first try to get a token from local cache and, if none is available, from the broker.

Note: the local cache should not contain an RT because the broker does not give MSAL the RT. But if it does, MSAL can still use it to get back a token. If there are any failures, MSAL will fallback to the broker. I have a concern here that we will silently ignore a bunch of exceptions which can lead to perf degradation...

**Actual behavior**

GetAccounts does not merge local and broker accounts.
RemoveAccoutns does not remove local accoutns and tokens
ATS does not search the local cache

**Impact**
By not merging local state with broker state, developer does not have an overview of what accounts exist and thefore not be able to use AcquireTokenSilent, but have to rely on AcquireTokenInteractive.

This affects especially users who use an app before and after the broker is installed or who uninstall the broker.

bgavrilMS:
CC @rpdome for confirmation

rpdome:
The way we implemented this is that we have to controllers, [LocalMsalController](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/44b9414f3be233053f441435440db7fbcee312b0/msal/src/main/java/com/microsoft/identity/client/internal/controllers/LocalMSALController.java) and [BrokerMsalController](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/37a3d41840f7ff76b061c7754e69a5911f76dd5d/msal/src/main/java/com/microsoft/identity/client/internal/controllers/BrokerMsalController.java).

[...]

PR #1873, opened 2020-06-09
Silent request refactoring and Broker bug fix

trwalke:
MSAL will now check local cache for an access token instead of going directly to broker first.
Refactoring.

Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1820


src/client/Microsoft.Identity.Client/Internal/Requests/SilentAuthStrategyFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.Identity.Client.Internal.Requests
+{
+    internal static class SilentAuthStrategyFactory
+    {
+        public static ISilentAuthStrategy GetSilentAuthStrategy(SilentRequest request, IServiceBundle servicebundle, AuthenticationRequestParameters parameters, AcquireTokenSilentParameters silentParameters)
+        {
+            if (parameters.IsBrokerConfigured && servicebundle.PlatformProxy.CanBrokerSupportSilentAuth())
+            {
+                return new SilentBrokerAuthStretegy(request, servicebundle, parameters, silentParameters);
```
bgavrilMS, line 17: typo: `strategy`

bgavrilMS: also below

src/client/Microsoft.Identity.Client/Internal/Requests/SilentBrokerAuthStretegy.cs
```diff
@@ -1,48 +1,77 @@
...
+            _silentRequest = request;
+            _broker = _serviceBundle.PlatformProxy.CreateBroker(null);
+            _logger = authenticationRequestParameters.RequestContext.Logger;
+        }
+
+        public async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
+        {
+            MsalAccessTokenCacheItem cachedAccessTokenItem = null;
+
+            if (!_silentParameters.ForceRefresh && !_authenticationRequestParameters.HasClaims)
```
bgavrilMS, line 46: No, you don't need to skip invoking the borker if "force refresh" is set to true, because MSAL does not have a RT so the whole thing will fail. 

Instead, you have to set the "force_refresh" and "claims" parameters on the broker request itself - as these are supported by the broker - see https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/48cb9cd0f53ed6fa402f8bdcb7e7fb200a4f59f9/common/src/main/java/com/microsoft/identity/common/internal/broker/BrokerRequest.java#L52

src/client/Microsoft.Identity.Client/Internal/Requests/SilentBrokerAuthStretegy.cs
```diff
@@ -1,48 +1,77 @@
...
+            {
+                cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
```
bgavrilMS, line 48: You need to invoke the broker here, this is the `SilentBrokerAuth` strategy, not the `client` / `cache` strategy.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -29,294 +29,17 @@ internal class SilentRequest : RequestBase
...
-
-        private async Task<IAccount> GetAccountFromParamsOrLoginHintAsync(AcquireTokenSilentParameters silentParameters)
-        {
-            if (silentParameters.Account != null)
-            {
-                return silentParameters.Account;
-            }
-
-            return await GetSingleAccountForLoginHintAsync(silentParameters.LoginHint).ConfigureAwait(false);
+            strategy = SilentAuthStrategyFactory.GetSilentAuthStrategy(this, serviceBundle, authenticationRequestParameters, silentParameters);
```
bgavrilMS, line 32: According to the spec, MSAL should: 

- try the client (cache) strategy first
- if that fails, try the broker strategy second (if the broker is configured and enabled etc.)

bgavrilMS:
This is right direction, nice stuff. Please see inline comments. 

- For AcquireTokenSilent, MSAL should first try the "client" strategy and then the "broker" strategy
- The broker strategy needs to be simplified - there is no "force refresh" and "claims" logic -the broker takes care of that logic. 

I assume there will be a follow up for GetAccounts and RemoveAccounts as well, in a separate commit: 

- GetAccounts gets accounts using BOTH get_account strategies. Merges happen based on home_account_id
- RemoveAccounts also removes accounts using BOTH strategies, i.e. removal occurs from both places

I think both of these

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -110,10 +107,27 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
                         new AcquireTokenCommonParameters(), 
                         requestContext));
 
-                accounts = await cacheSessionManager.GetAccountsAsync(Authority).ConfigureAwait(false);
+                localAccounts = await cacheSessionManager.GetAccountsAsync(Authority).ConfigureAwait(false);
+            }
+
+            if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
+            {
+                //Broker is available so accounts will be merged using home account ID with Broker taking priority
```
trwalke: confirmed with android that they prioritize local accounts when there are duplicates in the broker and local caches. once the local tokens expire, MSAL will use the broker account instead. We can do the same here.

jennyf19:
> using System;

license info

---
Refers to: src/client/Microsoft.Identity.Client/Internal/Requests/ISilentAuthStrategy.cs:1 in 1faa97c. [](commit_id = 1faa97c5e23fc40cdb5f37323c2a0b7b9de812ef, deletion_comment = False)

src/client/Microsoft.Identity.Client/Internal/Requests/ISilentAuthStrategy.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Internal.Requests
+{
+    internal interface ISilentAuthStrategy
```
jennyf19, line 10: nit: maybe `ISilentAuthRequest` or `ISilentAuthRequestStrategy`?

jennyf19:
> using System;

readd license info

---
Refers to: src/client/Microsoft.Identity.Client/Internal/Requests/SilentBrokerAuthStretegy.cs:1 in 1faa97c. [](commit_id = 1faa97c5e23fc40cdb5f37323c2a0b7b9de812ef, deletion_comment = False)

src/client/Microsoft.Identity.Client/Internal/Requests/SilentBrokerAuthStretegy.cs
```diff
@@ -1,48 +1,59 @@
...
-using Microsoft.Identity.Client.Internal.Requests;
+using Microsoft.Identity.Client.Internal.Broker;
 using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.Utils;
 
-namespace Microsoft.Identity.Client.Internal.Broker
+namespace Microsoft.Identity.Client.Internal.Requests
 {
-    internal class BrokerSilentRequest
+    internal class SilentBrokerAuthStretegy : ISilentAuthStrategy
```
jennyf19, line 17: >e [](start = 38, length = 1)

sp: "strategy"

src/client/Microsoft.Identity.Client/Internal/Requests/SilentBrokerAuthStretegy.cs
```diff
@@ -1,48 +1,59 @@
...
+        public async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
+        {
+            var response = await SendTokenRequestToBrokerAsync().ConfigureAwait(false);
+            return await _silentRequest.CacheTokenResponseAndCreateAuthenticationResultAsync(response).ConfigureAwait(false);
         }
 
         public async Task<MsalTokenResponse> SendTokenRequestToBrokerAsync()
         {
-            if(!Broker.IsBrokerInstalledAndInvokable())
+            if (!Broker.IsBrokerInstalledAndInvokable())
```
jennyf19, line 51: there won't be a null ref here right? with `Broker.`

jennyf19:
> using System;

license info

---
Refers to: src/client/Microsoft.Identity.Client/Internal/Requests/SilentClientAuthStretegy.cs:1 in 1faa97c. [](commit_id = 1faa97c5e23fc40cdb5f37323c2a0b7b9de812ef, deletion_comment = False)

src/client/Microsoft.Identity.Client/Internal/Requests/SilentClientAuthStretegy.cs
```diff
@@ -0,0 +1,299 @@
...
+using Microsoft.Identity.Client.ApiConfig.Parameters;
+using Microsoft.Identity.Client.Cache;
+using Microsoft.Identity.Client.Cache.Items;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Instance;
+using Microsoft.Identity.Client.OAuth2;
+
+namespace Microsoft.Identity.Client.Internal.Requests
+{
+    internal class SilentClientAuthStretegy : ISilentAuthStrategy
```
jennyf19, line 16: >e [](start = 38, length = 1)

sp.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentClientAuthStretegy.cs
```diff
@@ -0,0 +1,299 @@
...
+    {
+        private AuthenticationRequestParameters AuthenticationRequestParameters { get; }
+        private ICacheSessionManager CacheManager => AuthenticationRequestParameters.CacheSessionManager;
+        protected IServiceBundle ServiceBundle { get; }
+        private readonly AcquireTokenSilentParameters _silentParameters;
+        private const string TheOnlyFamilyId = "1";
```
jennyf19, line 22: don't we already have a const for this from the work bogdan did awhile ago?

src/client/Microsoft.Identity.Client/Internal/Requests/SilentClientAuthStretegy.cs
```diff
@@ -0,0 +1,299 @@
...
+            IAccount account = await GetAccountFromParamsOrLoginHintAsync(_silentParameters).ConfigureAwait(false);
+            AuthenticationRequestParameters.Account = account;
+
+            AuthenticationRequestParameters.Authority = Authority.CreateAuthorityForRequest(
+                ServiceBundle.Config.AuthorityInfo,
+                AuthenticationRequestParameters.AuthorityOverride,
+                account?.HomeAccountId?.TenantId);
+        }
+
+        public async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
```
jennyf19, line 48: this class is mainly the same code as before right?

src/client/Microsoft.Identity.Client/Internal/Requests/SilentClientAuthStretegy.cs
```diff
@@ -0,0 +1,299 @@
...
+            return new AuthenticationResult(
+                cachedAccessTokenItem,
+                msalIdTokenItem,
+                AuthenticationRequestParameters.AuthenticationScheme,
+                AuthenticationRequestParameters.RequestContext.CorrelationId);
+        }
+
+        private void ThrowIfNoScopesOnB2C()
+        {
+            // B2C will not issue an access token if no scopes are requested
```
jennyf19, line 133: I think the comment is misleading...maybe something like:
// During AT Silent with no scopes, unlike AAD, B2C will not issue an access token
...then the rest of the comment.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -29,294 +30,56 @@ internal class SilentRequest : RequestBase
...
-            // Normal, non-FOCI flow
-            if (msalTokenResponse == null)
-            {
-                // Look for a refresh token
-                MsalRefreshTokenCacheItem appRefreshToken = await FindRefreshTokenOrFailAsync()
-                    .ConfigureAwait(false);
-
-                msalTokenResponse = await RefreshAccessTokenAsync(appRefreshToken, cancellationToken)
-                    .ConfigureAwait(false);
+                _logger.Info("Attempting to acquire token using client auth strategy...");
```
jennyf19, line 63: >client auth strategy [](start = 64, length = 20)

nit: if i were a customer i would have no idea what this means. lol. can we be more specific?

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -29,294 +30,56 @@ internal class SilentRequest : RequestBase
...
-#else
-                    if (MsalError.InvalidGrantError.Equals(ex?.ErrorCode, StringComparison.OrdinalIgnoreCase) &&
-                        MsalError.ClientMismatch.Equals(ex?.SubError, StringComparison.OrdinalIgnoreCase))
+                    if (errorCode == MsalError.InvalidGrantError
+                     || errorCode == MsalError.NoTokensFoundError
+                     || errorCode == MsalError.NoAccountForLoginHint)
                     {
-                        logger.Error("[FOCI] FRT refresh failed - client mismatch");
-                        return null;
+                        _logger.Info("client auth strategy failed to acquire a token. Attempting to use broker strategy");
```
jennyf19, line 76: >client [](start = 38, length = 6)

same comment as above. Maybe start the logger with the name of the class instead, or like bogdan did w/the PR on logging, where the name of the class is automatically included. then you'll know where we are in the code.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AccountTests.cs
```diff
@@ -298,6 +299,56 @@ public void TestGetAccounts()
...
+                                                                            .WithHttpManager(httpManager)
+                                                                            .WithTelemetry(new TraceTelemetryConfig())
+                                                                            .BuildConcrete();
+                //Adding local account to local cache
+                tokenCacheHelper.PopulateCache(app.UserTokenCacheInternal.Accessor,
+                                                TestConstants.Uid,
+                                                TestConstants.Utid,
+                                                TestConstants.ClientId,
+                                                TestConstants.ProductionPrefCacheEnvironment,
+                                                "LocalAccount@microsoft.com");
```
jennyf19, line 319: nit: make these constants

jennyf19:
I think this looks much better, thanks travis. I approve, but let BG give the final sign-off. lol. If you want me to look at it again later, lmk. 

---
In reply to: [641295740](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1873#issuecomment-641295740) [](ancestors = 641295740)

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -110,10 +107,27 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
                         new AcquireTokenCommonParameters(), 
                         requestContext));
 
-                accounts = await cacheSessionManager.GetAccountsAsync(Authority).ConfigureAwait(false);
+                localAccounts = await cacheSessionManager.GetAccountsAsync(Authority).ConfigureAwait(false);
+            }
+
+            if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
```
bgavrilMS, line 113: I don't think the CanBrokerSupportSilentAuth flag is needed for GetAccounts and RemoveAccounts. iOS broker can simply do nothing...

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -110,10 +107,27 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
...
+            {
+                //Broker is available so accounts will be merged using home account ID with local accounts taking priority
+                var broker = ServiceBundle.PlatformProxy.CreateBroker(null);
+                brokerAccounts = await broker.GetAccountsAsync(AppConfig.ClientId).ConfigureAwait(false);
+
+                foreach(IAccount account in brokerAccounts)
+                {
+                    if (localAccounts.Where(x => x.HomeAccountId == account.HomeAccountId).Count() == 0)
```
bgavrilMS, line 121: Use `Any(x => ...)` instead of Where + Count
Also, the string comparison needs to be done using String.Equals

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
bgavrilMS: This will be a cleaner design. 

You can probably eliminate it from AcquireTokenSilent as well, but it will be more complicated.

src/client/Microsoft.Identity.Client/Internal/Requests/ISilentAuthStrategy.cs
bgavrilMS: I like the word strategy in it because it shows that the pattern with the same name is used https://www.dofactory.com/net/strategy-design-pattern

src/client/Microsoft.Identity.Client/Internal/Requests/SilentBrokerAuthStretegy.cs
bgavrilMS: It's passed in the ctor, but we don't check for nulls ...

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -29,294 +30,56 @@ internal class SilentRequest : RequestBase
...
-            }
 
-            if (accounts.Count() > 1)
-            {
-                throw new MsalUiRequiredException(
-                    MsalError.MultipleAccountsForLoginHint,
-                    MsalErrorMessage.MultipleAccountsForLoginHint,
-                    null,
-                    UiRequiredExceptionClassification.AcquireTokenSilentFailed);
+            _brokerStrategy = new Lazy<ISilentAuthStrategy>(() => new SilentBrokerAuthStretegy(this,
```
bgavrilMS, line 34: Nice use of Lazy. 

nit: variable name should be `_brokerStrategyLazy`

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequest.cs
```diff
@@ -29,294 +30,56 @@ internal class SilentRequest : RequestBase
...
 
-            }
-
-            return accounts.First();
+            _logger = authenticationRequestParameters.RequestContext.Logger;
         }
 
-        private async Task<IAccount> GetAccountFromParamsOrLoginHintAsync(AcquireTokenSilentParameters silentParameters)
+        //For test
+        internal void SetStrategiesForTest(ISilentAuthStrategy clientStrategy, ISilentAuthStrategy brokerStrategy)
```
bgavrilMS, line 45: This is good, but the recommended pattern is to pass test implementations from the ctor, smth like 

```csharp
public SilentRequest(exisisting_params, ISilentAuthStrategy clientStrategyOverride, ISilentAuthStrategy brokerStrategyOverride)
{
_clientStrategy = clientStrategyOverride ?? new SilentClientAuthStretegy(...);
}
```

bgavrilMS: The advantage of this is that you can keep the class immutable, i.e. _clientStrategy  and _brokerStrategy variables will be readonly and once the SilentRequest object is created, it cannot be changed. Immutable classes are a dream - they are simpler to work with and are thread safe.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AccountTests.cs
bgavrilMS: Personally I like to see constants in the test, makes it more readable.

jennyf19: i know you do. ;) but since they are in a few different places, makes changing them easier if that needs to happen. we tend to have more constants than not anyway. my 2 cents

src/client/Microsoft.Identity.Client/Internal/Requests/SilentBrokerAuthStretegy.cs
jennyf19: right, that's what I meant. Are we checking for null anywhere before we get this far.

src/client/Microsoft.Identity.Client/Internal/Requests/ISilentAuthStrategy.cs
jennyf19: right, but what about `ISilentAuthRequestStrategy`?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AccountTests.cs
bgavrilMS: It's a good point.

jennyf19: plus i knew you would comment if i commented about it. lol. ;)

Issue #1851, opened 2020-05-30
Google IDP throwing "This browser or app may not be secure." (External Identities)


pottabathini:
**Which Version of MSAL are you using ?**
Downloaded MSAL Desktop app from QuickStarts inside Portal AAD Applications blade. 
(Quickstart: Acquire a token and call Microsoft Graph API from a Windows desktop app)

**Platform**
net45

**What authentication flow has the issue?**
*  Desktop
    * [ Yes] Interactive

**Is this a new or existing app?**
New App. Created for Testing

**Repro**

1. Follow below doc to enable self-service signup and add Google IDP
2. https://docs.microsoft.com/en-in/azure/active-directory/b2b/self-service-sign-up-user-flow
3. Download the MSAL desktop app from quick start (Inside AAD -> Application) and start running the application. 
4. Choose Google IDP and it will throw exception

**Expected behavior**
Desktop App should allow to login with Google IDP.

**Actual behavior**
![image](https://user-images.githubusercontent.com/11867309/83260029-b0d45480-a1d6-11ea-9504-6bc0337968bc.png)

bgavrilMS:
Root cause for this is that MSAL uses an embedded IE browser on .net desktop. 

I think the main out of this today is to try to use the use the system web browser (details [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/System-Browser-on-.Net-Core#system-browser-experience))

Otherwise, we need to find a way to use a more modern embedded browser. I don't think this is possible for .net classic 4.5, but we should be able to make it work on .net core (Tracking issue [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1398)) or you can make it work from you app by providing your own ICustomWebUi implementation.

jmprieur:
Closing as this question is answered

agrabhi:
@jmprieur - shouldnt the solution work out of box? It seems customer needs to do extra steps to be able to use MSAL for desktop

jmprieur:
@agrabhi 
Indeed, the system web browser is used by default for mobile platforms (Xamarin.iOS / Android), but the embedded Web view is used for desktop as using the system web view does not give a very good user experience.

Adding @brandwe to comment as he owns the desktop scenarios.

[...]

Issue #1815, opened 2020-05-13
Cannot migrate ADALV3 cachetokens to MSALV3 when multiple resourceId's are used


HaroldH76:
**Which Version of MSAL are you using ?**
MSAL 4.13.0
ADAL 5.2.7

**Platform**
Net 4.7

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [x] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Is this a new or existing app?**
The app is in production, and I am working on upgrading from ADALV5 to MSALV4. 

**Repro**
I have a confidential application which stores ADALV3 tokens in a custom SQL TokenCache storage. I try to upgrade to MSALV4.
But I'm having issues migrating the stored ADALV3 tokens to MSALV3.

The ADALV3 contains 2 tokens for the same audience but different resourceId's, one for https://graph.microsoft.com and one for https://outlook.office.com. 

I have this code to migrate the token:
```cs
public void BeforeAccessNotification(TokenCacheNotificationArgs args)
{
    var state = GetFromSQL();
    try
    {
        // new version
        args.TokenCache.DeserializeMsalV3(state, true);
    }
    catch 
    {
        // old version, will be migrated to new version 
        args.TokenCache.DeserializeAdalV3(state);
    }
}
```

This works without errors but when I retrieve an access token with this:
```cs
var account = (await app.GetAccountsAsync()).FirstOrDefault();
if (account != null)
{
    var result = await app.AcquireTokenSilent(defaultScope, account)
        .ExecuteAsync();
}
```

Then I get this error:
```
MSAL.Desktop.4.13.0.0.MsalClientException: 	ErrorCode: invalid_adal_cache
Microsoft.Identity.Client.MsalClientException: The ADAL cache is invalid as it contains multiple refresh token entries for one user. Mitigation: Delete the ADAL cache. If you do not maintain an ADAL cache, this may be a bug in MSAL.
```

**Expected behavior**
A possible migration path of stored ADALV3 tokens to MSALV3 tokens.

**Actual behavior**
Error after ADALV3 token is migrated to MSALV3 token:
```
MSAL.Desktop.4.13.0.0.MsalClientException: 	ErrorCode: invalid_adal_cache
[...]

PR #1884, opened 2020-06-12
Fix for #1851 - ignore ADAL resource string when fetching RTs

bgavrilMS:
Fix for #1815 1815


src/client/Microsoft.Identity.Client/Cache/AdalTokenCacheKey.cs
```diff
@@ -104,27 +107,23 @@ public override int GetHashCode()
         {
             const string Delimiter = ":::";
             var hashString = Authority + Delimiter
-                           + Resource.ToLowerInvariant() + Delimiter
+                          // + Resource.ToLowerInvariant() + Delimiter
```
jennyf19, line 110: why not delete this?

bgavrilMS: Yes good point

Issue #1902, opened 2020-06-25
[Feature Request] Expose a new string SuggestedCacheKey in TokenCacheNotificationArgs  


jmprieur:
**Is your feature request related to a problem? Please describe.**
Today, in confidential client applications, there needs to be one cache per user/account. And the difficulty is that the token cache custom serialization needs to decide where to serialize the cache based on some information which is not available today in the `TokenCacheNotificationArgs`. 
This was solved in Microsoft.Identity.Web by having the application compute a cache key (then used by the customer serializers to get the blob for the account in the case of the user token cache, or for the app token cache)

Unfortunately, we just learnt from the ASP.NET core team that the way we do it, by using the HttpContext is risky because, in particular, HttpContext is not thread safe, and should be avoided.

**Describe the solution you'd like**
The proposal is to have MSAL.NET propose a hint for a cache key (that the token cache customer serializers would be free to use or not) based on the flow. We would add a new property to TokenCacheNotificationArgs, of type string, and named **SuggestedCacheKey** computed as follows:

- when the flow is ClientCredentials, return the ClientId (or App_{ClientId}
- when the flow is OBO, return the hash of the access token passed-in to call the OBO
- for the other flows (including AcquireTokenSilent), use the home account ID when it's available.

**Describe alternatives you've considered**

Another possibility would be to add to AcquireTokensilent builders for confidential client apps the .WithTokenCacheKey method, but this is not my prefered option.

**Additional context**
See how things are done today in Microsoft.Identity.Web

https://github.com/AzureAD/microsoft-identity-web/blob/b9457a826dc4f6d7fbafc67d37f06b2d6cd2acd3/src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs#L60-L75

and what issue this provokes, as described in issue https://github.com/AzureAD/microsoft-identity-web/issues/235

jmprieur:
See https://github.com/Azure-Samples/ms-identity-aspnet-webapp-openidconnect/pull/38 for how the usage of SuggestedCacheKey and HasToken simplifies the token cache serialization in confidential client applications

PR #1908, opened 2020-06-27
Add suggested cache key for confidential client scenarios

jennyf19:
This is for #1902 for confidential client scenarios, and has more info on the issue.

[ ] OBO should use the AT hash as the cache key, because the claims in the AT can change, so we use the AT directly to determine the cache key so the right AT is used.

[ ] Client credentials uses `{client_id} + "_AppTokenCache"`, as it's using the app token cache, so it needs the token based on the client id. MS Identity Web suffixes it w/`"_AppTokenCache"`, so we'd like to keep this pattern to not break current customers.

[ ] all other confidential client flows (including silent) use the home account id, as the key, because there is a user account. 

In order to make the changes in MS Identity Web, we'd need to have this addition included in a nuget package. We have a deadline of July 6 on our end. cc: @jmprieur @henrik-me

### DIFF
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -76,6 +76,18 @@ namespace Microsoft.Identity.Client
         /// </summary>
         public async Task<IEnumerable<IAccount>> GetAccountsAsync()
         {
+            return await GetAccountsAndSetCacheKeyAsync(null).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Returns all the available <see cref="IAccount">accounts</see> in the user token cache for the application.
+        /// Also sets the cache key based on a given home account id, which is the account id of the home account for the user. 
+        /// This uniquely identifies the user across AAD tenants.
+        /// </summary>
+        /// <param name="homeAccountId">The identifier is the home account id of the account being targetted in the cache./>
+        /// </param>
+        private async Task<IEnumerable<IAccount>> GetAccountsAndSetCacheKeyAsync(string homeAccountId)
+        {
             RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
             IEnumerable<IAccount> localAccounts = Enumerable.Empty<IAccount>();
             IEnumerable<IAccount> brokerAccounts = Enumerable.Empty<IAccount>();
@@ -89,14 +101,17 @@ namespace Microsoft.Identity.Client
             }
             else
             {
+                var authParameters = new AuthenticationRequestParameters(
+                        ServiceBundle,
+                        UserTokenCacheInternal,
+                        new AcquireTokenCommonParameters(),
+                        requestContext);
+
+                authParameters.Account = new Account(homeAccountId, null, Authority);
                 // a simple session consisting of a single call
                 CacheSessionManager cacheSessionManager = new CacheSessionManager(
                     UserTokenCacheInternal,
-                    new AuthenticationRequestParameters(
-                        ServiceBundle, 
-                        UserTokenCacheInternal, 
-                        new AcquireTokenCommonParameters(), 
-                        requestContext));
+                     authParameters);
 
                 localAccounts = await cacheSessionManager.GetAccountsAsync(Authority).ConfigureAwait(false);
             }
@@ -150,7 +165,7 @@ namespace Microsoft.Identity.Client
         /// </param>
         public async Task<IAccount> GetAccountAsync(string accountId)
         {
-            var accounts = await GetAccountsAsync().ConfigureAwait(false);
+            var accounts = await GetAccountsAndSetCacheKeyAsync(accountId).ConfigureAwait(false);
             return accounts.FirstOrDefault(account => account.HomeAccountId.Identifier.Equals(accountId, StringComparison.OrdinalIgnoreCase));
         }
 
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.77
 // Licensed under the MIT License.
 
 using System;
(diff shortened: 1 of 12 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -86,6 +86,8 @@ internal class AuthenticationRequestParameters
 
         public string ClaimsAndClientCapabilities { get; private set; }
 
+        public string SuggestedCacheKey { get; set; }
```
bgavrilMS, line 89: Logically this does not belong in this class. Let me have if we can put it somewhere else...

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -89,14 +101,17 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
             }
             else
             {
+                var authParameters = new AuthenticationRequestParameters(
+                        ServiceBundle,
+                        UserTokenCacheInternal,
+                        new AcquireTokenCommonParameters(),
+                        requestContext);
+
+                authParameters.Account = new Account(homeAccountId, null, Authority);
```
bgavrilMS, line 110: We need to find a better design for this. It's too contrived that "GetAccount(id)" internally creates an Account object that you pass as a hint for the `GetAccounts` logic.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -141,6 +143,7 @@ public async Task ConfidentialClientWithCertificateTestAsync()
             MsalAssert.AssertAuthResult(authResult);
             appCacheRecorder.AssertAccessCounts(2, 1);
             Assert.IsTrue(appCacheRecorder.LastNotificationArgs.IsApplicationCache);
+            Assert.AreEqual(AppCacheKey, appCacheRecorder.LastNotificationArgs.SuggestedCacheKey);
```
bgavrilMS, line 146: Integration tests seem to only cover `AcquireTokenForClient` and `OBO`, but not `AcquireTokenSilent`. 
We need some unit test coverage as well - should be easy to update some existing tests for Client Creds and OBO. For AcquireTokenSilent as well. For `GetAccount` I think you need a dedicated unit test.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -75,50 +75,70 @@ internal ClientApplicationBase(ApplicationConfiguration config)
...
-                // a simple session consisting of a single call
-                CacheSessionManager cacheSessionManager = new CacheSessionManager(
-                    UserTokenCacheInternal,
-                    new AuthenticationRequestParameters(
-                        ServiceBundle, 
-                        UserTokenCacheInternal, 
-                        new AcquireTokenCommonParameters(), 
-                        requestContext));
+            var accountsFromCache = await GetAccountsFromCacheAsync(homeAccountId, requestContext).ConfigureAwait(false);
+            IEnumerable<IAccount> cacheAndBrokerAccounts =
```
bgavrilMS, line 94: Small refactoring of this method, not related to your work.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -75,50 +75,70 @@ internal ClientApplicationBase(ApplicationConfiguration config)
...
+            string homeAccountId,
+            RequestContext requestContext)
+        {
+            var authParameters = new AuthenticationRequestParameters(
+                    ServiceBundle,
+                    UserTokenCacheInternal,
+                    new AcquireTokenCommonParameters(),
+                    requestContext);
+
+            authParameters.SuggestedWebAppCacheKey = homeAccountId;
```
bgavrilMS, line 134: This is the only change in this file related to your work.

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -55,5 +57,21 @@ public sealed partial class TokenCacheNotificationArgs
...
+        /// The value is: 
+        /// 
+        /// <list type="bullet">
+        /// <item>the homeAccountId for AcquireTokenSilent and GetAccount(homeAccountId)</item>
+        /// <item>clientID + "_AppTokenCache" for AcquireTokenForClient</item>
+        /// <item>the hash of the original token for AcquireTokenOnBehalfOf</item>
+        /// <item>null for all other calls, such as PubliClientApplication calls, which should persist the token cache in a single location</item>
+        /// </list>
+        /// </summary>
+        public string SuggestedCacheKey { get; }
```
bgavrilMS, line 75: I added comments, please review. (CC @jmprieur )

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
jennyf19: thanks @bgavrilMS ...I will get back to work on this later today. appreciate the feedback.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
```diff
@@ -749,6 +751,56 @@ public void AcquireTokenSilentServiceErrorTest()
...
+
+                // Act
+                await app.AcquireTokenSilent(TestConstants.s_scope, accountById).ExecuteAsync().ConfigureAwait(false);
+
+                // Assert
+                cacheAccess.AssertAccessCounts(3, 0);
+                Assert.AreEqual(homeAccId, cacheAccess.LastNotificationArgs.SuggestedCacheKey);
+
+                // Act
+                await app.AcquireTokenSilent(TestConstants.s_scope, accountById.Username).ExecuteAsync().ConfigureAwait(false);
```
bgavrilMS, line 795: @jmprieur @jennyf19 - please see this test case - ATS with login hint is not able to suggest a token cache key, because MSAL needs to load all accounts and filter by username. Loading all accounts triggers a cache read. 

If this is not acceptable, we should discuss.

jmprieur: Yes, @bgavrilMS : we need to discuss.
this API, as well as GetAccountsAsync() don't make sense for confidential client applications.
but this is a different problem

jennyf19:
@bgavrilMS the tests don't seem to pass. i notice the same for your PR...its that expected?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/AcquireTokenSilentTests.cs
bgavrilMS: We can duplicate `GetAccountsAsync` methods so that they exist on both PCA and CCA. And then Obsolete them on CCA. But why would I not want to do `GetAccountsAsync` on a cca?

bgavrilMS:
@jennyf19 - on my PR I have a unit test failing, I'll take care of this. On this PR some unrelated integration tests are failing. Let me rerun the build.

jennyf19:
@bgavrilMS sounds good. go ahead and merge whenever you want. ty

PR #1911, opened 2020-07-02
Add a "HasTokens" flag on the cache notification args

bgavrilMS:


src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -55,5 +57,15 @@ public sealed partial class TokenCacheNotificationArgs
...
+
+        /// <summary>
+        /// Indicates whether MSAL's current in-memory token cache contains any tokens. The in-memory cache is        
+        /// kept in sync with the persisted version through OnBefore / OnAfter callbacks. 
+        /// </summary>
+        /// <remarks>
+        /// If this flag is false in the OnAfterAccessAsync notification, the token cache can be deleted.        
+        /// MSAL takes into consideration access tokens expiration when computing this flag, but not refresh token expiration, which is not known to MSAL.7
+        /// </remarks>
+        public bool HasTokens { get; }
```
bgavrilMS, line 69: @jmprieur - please review the comments on this public API.

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/TokenCacheNotificationTests.cs
```diff
@@ -212,5 +213,61 @@ public async Task TestSerializationViaAsync()
...
+                Assert.IsTrue(accounts.Any());
+
+                Trace.WriteLine("Step 4 - call RemoveAccounts - this will delete all the tokens");
+                var cacheAccessRecorder4 = app.UserTokenCache.RecordAccess();
+                await app.RemoveAsync(accounts.Single()).ConfigureAwait(false);
+
+                Assert.IsTrue(cacheAccessRecorder4.LastBeforeAccessNotificationArgs.HasTokens);
+                Assert.IsTrue(cacheAccessRecorder4.LastBeforeWriteNotificationArgs.HasTokens);
+                Assert.IsFalse(cacheAccessRecorder4.LastAfterAccessNotificationArgs.HasTokens);
+            }
```
bgavrilMS, line 270: @jmprieur @TiagoBrenck - please let me know if this behavior of the flag is intuitive enough. The flag tells you if the in-memory cache has tokens, so the Before* callbacks are not really interesting because it does not matter if the current in-memory cache has tokens or not, tokens can be injected from the persisted cache.

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -55,5 +57,15 @@ public sealed partial class TokenCacheNotificationArgs
...
         /// </remarks>
         public bool IsApplicationCache { get; }
+
+        /// <summary>
+        /// Indicates whether MSAL's current in-memory token cache contains any tokens. The in-memory cache is        
+        /// kept in sync with the persisted version through OnBefore / OnAfter callbacks. 
+        /// </summary>
+        /// <remarks>
+        /// If this flag is false in the OnAfterAccessAsync notification, the token cache can be deleted.        
+        /// MSAL takes into consideration access tokens expiration when computing this flag, but not refresh token expiration, which is not known to MSAL.7
```
jmprieur, line 67: Proposing to remove the 7 :)

```suggestion
        /// MSAL takes into consideration access tokens expiration when computing this flag, but not refresh token expiration, which is not known to MSAL.
```

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -55,5 +57,15 @@ public sealed partial class TokenCacheNotificationArgs
         /// See https://aka.ms/msal-net-app-cache-serialization for details.
         /// </remarks>
         public bool IsApplicationCache { get; }
+
+        /// <summary>
+        /// Indicates whether MSAL's current in-memory token cache contains any tokens. The in-memory cache is        
+        /// kept in sync with the persisted version through OnBefore / OnAfter callbacks. 
+        /// </summary>
+        /// <remarks>
+        /// If this flag is false in the OnAfterAccessAsync notification, the token cache can be deleted.
```
jmprieur, line 66:
```suggestion
        /// If this flag is false in the OnAfterAccessAsync notification, the serialized token cache can be deleted,
        /// for instance to free-up space in some storage.
```

tests/Microsoft.Identity.Test.Unit.net45/CacheTests/TokenCacheNotificationTests.cs
jmprieur: @bgavrilMS : I think it's well explained in the XML comment. We'll use it in Microsoft.Identity.Web and that will do (this is mostly for Confidential client apps)

jennyf19:
This is awesome @Bogdan Gavril ....nice improvements

---
In reply to: [441821866](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1911#issuecomment-441821866) [](ancestors = 441821866)

PR #1913, opened 2020-07-06
Fix SuggestedCacheKey

bgavrilMS:


src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
```diff
@@ -0,0 +1,65 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Identity.Client.Internal.Requests;
+
+namespace Microsoft.Identity.Client.Cache
+{
+    internal static class SuggestedWebCacheKeyFactory
```
bgavrilMS, line 8: I decided to centralize the calculation of the key even though we have 1 class for each request type. In the end, the design is cleaner.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentClientAuthStrategy.cs
```diff
@@ -258,7 +257,7 @@ private async Task<MsalRefreshTokenCacheItem> FindRefreshTokenOrFailAsync()
 
         private async Task<IAccount> GetSingleAccountForLoginHintAsync(string loginHint)
         {
-            var accounts = await CacheManager.GetAccountsAsync(ServiceBundle.Config.AuthorityInfo.CanonicalAuthority)
+            var accounts = await CacheManager.GetAccountsAsync()
```
bgavrilMS, line 260: Small refactoring - there's no need to pass the authority here, the cache manager has a reference to the service bundle.

PR #1919, opened 2020-07-08
UI Automation Updates.

trwalke:
Removing dependency on IsAppcenterBuild.
Updating mac os image

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
@@ -53,12 +53,12 @@ namespace Microsoft.Identity.Client.Platforms.iOS.EmbeddedWebview
         protected WKWebView PrepareWKWebView()
         {
             WKWebViewConfiguration wkconfg = new WKWebViewConfiguration() { };
-#if !IS_APPCENTER_BUILD
+
             if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
             {
                 wkconfg.DefaultWebpagePreferences.PreferredContentMode = WKContentMode.Mobile;
             }
-#endif
+
             _wkWebView = new WKWebView(View.Bounds, wkconfg)
             {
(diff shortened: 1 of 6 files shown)


src/client/Microsoft.Identity.Client/Platforms/iOS/SystemWebview/SystemWebUI.cs
```diff
@@ -97,25 +94,6 @@ public void Authenticate(Uri authorizationUri, Uri redirectUri, RequestContext r
 
                     sfAuthenticationSession.Start();
                 }
-#else
-                if (UIDevice.CurrentDevice.CheckSystemVersion(11, 0))
-                {
```
jennyf19, line 102: don't we need this still? for iOS 11?

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
```diff
@@ -53,12 +53,12 @@ public override void ViewDidLoad()
         protected WKWebView PrepareWKWebView()
         {
             WKWebViewConfiguration wkconfg = new WKWebViewConfiguration() { };
-#if !IS_APPCENTER_BUILD
```
bgavrilMS, line 56: yay!

src/client/Microsoft.Identity.Client/Platforms/iOS/SystemWebview/SystemWebUI.cs
trwalke: you are correct. adding back in

src/client/Microsoft.Identity.Client/Platforms/iOS/EmbeddedWebview/MsalAuthenticationAgentUIViewController.cs
bgavrilMS, line 56: I think this flag also appears in the some of the YAML build defintions..

trwalke: yes, it is still needed for the broker test app that we create for the broker team to test for us. However, these are no longer needed in any of the shipped product code.

Issue #1620, opened 2020-02-10
[Enhancement] iOS broker logic should verify redirect URI and provide good error message


bgavrilMS:
**Which Version of MSAL are you using ?**
4.7.1

**Platform**
xamarin.ios

**What authentication flow has the issue?**
interactive  + broker 

**Repro**
According to the MSAL.ObjC [docs](https://docs.microsoft.com/en-us/azure/active-directory/develop/redirect-uris-ios): 

`MSAL will verify if your redirect URI registers correctly, and return an error if it's not.`

**Actual**: MSAL.NET does not perform such a verification. If redirect URI is not set correctly, broker simply remains on screen and developers do not understand what needs to be done.

Note: I think that you can check at runtime the URL scheme configured in Info.plist and cross reference it to the RedirectUri.

neha-bhargava:
Fixed in [MSAL 4.17.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.17.0)

PR #1929, opened 2020-07-13
Fail if redirect uri is not correct on iOS + broker

bgavrilMS:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1620 plus add more logging

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
@@ -91,7 +91,9 @@ namespace Microsoft.Identity.Client.Platforms.iOS
             AuthenticationRequestParameters authenticationRequestParameters,
             AcquireTokenInteractiveParameters acquireTokenInteractiveParameters)
         {
+            ValidateRedirectUri(authenticationRequestParameters.RedirectUri);
             AuthenticationContinuationHelper.UnreliableLogger = _logger;
+
             using (_logger.LogMethodDuration())
             {
                 Dictionary<string, string> brokerRequest = CreateBrokerRequestDictionary(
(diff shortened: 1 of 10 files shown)


src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
```diff
@@ -104,6 +106,19 @@ public bool IsBrokerInstalledAndInvokable()
             }
         }
 
+        private void ValidateRedirectUri(Uri redirectUri)
+        {
+            string bundle = NSBundle.MainBundle.BundleIdentifier;
+            string expectedRedirectUri = $"msauth.{bundle}://auth";
+            if (!string.Equals(expectedRedirectUri, redirectUri.AbsoluteUri, StringComparison.OrdinalIgnoreCase))
```
bgavrilMS, line 113: @antrix1989 - could you please confirm the logic here? I want to throw if the user sets a redirect uri that does not follow this rule, although I've seen that in practice this rule is not set in stone, i.e. ADAL style redirect URIs continue to work. Note that we do not use SSO extensions yet.

src/client/Microsoft.Identity.Client/Platforms/iOS/WebviewBase.cs
```diff
@@ -37,15 +40,28 @@ public WebviewBase()
             RequestContext requestContext,
             CancellationToken cancellationToken);
 
-        public static bool ContinueAuthentication(string url)
+        public static bool ContinueAuthentication(string url, Core.ICoreLogger unreliableLogger)
```
trwalke, line 43: What exactly is unreliable referring to in this context?

bgavrilMS: Our logger is tied to a request - we create a new logger for each request. This way we always log the correlation ID. 
But since these authentication callbacks are static, we cannot inject the logger into them via the constructor. Instead, I made it so that each request sets the logger static property. 

This is potentially unreliable if multiple requests are fired at the same time - the logger will simply be the one from the last fire request. So we cannot necessarily trust the correlation IDs. That's why I was thinking `unreliable`. But maybe a better name can be found ... suggestions? (naming is hard)

trwalke: I think latestRequestLogger or lastRequestLogger might be better at explaining that this is only logging the last request. otherwise add a comment to explain so that people are aware of this during debugging

PR #1931, opened 2020-07-13
Add a new SLN for convenience

bgavrilMS:


bgavrilMS:
CC @jmprieur ... just libs and unit tests

jennyf19:
@bgavrilMS should we do this in identity web? is it auto-generated?

bgavrilMS:
@jennyf19 - yeah, sln files are generated by VS. The main reason for this small change is to give contributors an easier time with MSAL, as this solution does not have all the mobile apps and the integration test stuff which are a pain to build. Not sure you need anything like this on Identity.Web since that solution builds fast anyway.

Issue #1728, opened 2020-03-25
[Feature Request] Add an enum that indicates the source of a token - cache, IdP or broker.


bgavrilMS:
**Is your feature request related to a problem? Please describe.**
Services need to monitor and alarm around cache hits, including the token cache.

**Describe the solution you'd like**
`AuthenticationResult.TokenSource` 

**Describe alternatives you've considered**
Logs and ExpiresOn property can also be used but logs require lots of post-processing and ExpiresOn cannot be used reliably.

jmprieur:
@bgavrilMS : would we want it to be a public property?

bgavrilMS:
Yes, this would benefit our customers in Web Api / Web Api and Daemon scenarios at scale. Some of hte issues they are facing are described below. This improvement hope to address problem 4.


Problem 1: if a customer brings online a large number of machines at once (e.g. during a big scale up), then each of these machines will go to AAD to fetch a token, even though the token is identical
Solution: persist the token cache to Redis or some other distributed cache

Problem 2: It takes a few seconds to acquire a token from AAD (mostly network I/O). If during these few seconds other machines require the token, they would go to AAD to get it. Especially problematic if AAD is under high load.
Problem 3: If AAD is down when the token expires, your service will become unavailable. 
Solution: pro-active token renewal as described [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/High-availability#pro-active-token-renewal)

Problem 4: You do not have a mechanism to monitor and alarm for when MSAL fails to hit the token cache.
Solution: flag on AuthResult

henrik-me:
Docs and samples should be updated to showcase usage of this. Similar to logging to be on by default.

bgavrilMS:
I would like to get this in sooner rather than later. I see a pattern among our customers to create new CCA objects for each call, and to not use any cache serialisation. `AcquireTokenForClient` is the main problem here

If we had this flag, we could guide customers to writing integration tests / monitoring to check that they don't hit AAD on every request.

jmprieur:
LGTM., @bgavrilMS.

bgavrilMS:
Also please see https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/1682

neha-bhargava:
[...]

PR #1933, opened 2020-07-13
Add token source enum

neha-bhargava:
#1728

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
@@ -11,13 +11,6 @@ namespace Microsoft.Identity.Client
     /// </summary>
     public class AuthenticationResultMetadata
     {
-        /// <summary>
-        /// Default constructor for the class AuthenticationResultMetadata
-        /// </summary>
-        public AuthenticationResultMetadata()
-        {
-            
-        }
 
         /// <summary>
         /// Constructor for the class AuthenticationResultMetadata
@@ -31,6 +24,6 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// The source of the token in the result.
         /// </summary>
-        public TokenSource TokenSource { get; set; }
+        public TokenSource TokenSource { get; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Contains metadata of the authentication result.
+    /// </summary>
+    public class AuthenticationResultMetadata
+    {
+        /// <summary>
+        /// Default constructor for the class AuthenticationResultMetadata
+        /// </summary>
+        public AuthenticationResultMetadata()
+        {
+            
+        }
+
+        /// <summary>
+        /// Constructor for the class AuthenticationResultMetadata
+        /// <param name="tokenSource">The token source.</param>
+        /// </summary>
+        public AuthenticationResultMetadata(TokenSource tokenSource)
+        {
+            TokenSource = tokenSource;
+        }
+
+        /// <summary>
+        /// The source of the token in the result.
+        /// </summary>
+        public TokenSource TokenSource { get; set; }
+    }
+}
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -211,7 +211,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 idtItem, 
                 AuthenticationRequestParameters.AuthenticationScheme,
                 AuthenticationRequestParameters.RequestContext.CorrelationId,
-                TokenSource.IdentityProvider);
+                msalTokenResponse.TokenSource == TokenSource.None? TokenSource.IdentityProvider : msalTokenResponse.TokenSource);
         }
 
         private void ValidateAccountIdentifiers(ClientInfo fromServer)
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -210,7 +210,8 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 atItem, 
                 idtItem, 
                 AuthenticationRequestParameters.AuthenticationScheme,
-                AuthenticationRequestParameters.RequestContext.CorrelationId);
+                AuthenticationRequestParameters.RequestContext.CorrelationId,
+                TokenSource.IdentityProvider);
         }
 
         private void ValidateAccountIdentifiers(ClientInfo fromServer)
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -211,7 +211,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 idtItem, 
                 AuthenticationRequestParameters.AuthenticationScheme,
                 AuthenticationRequestParameters.RequestContext.CorrelationId,
-                msalTokenResponse.TokenSource == TokenSource.None? TokenSource.IdentityProvider : msalTokenResponse.TokenSource);
+                msalTokenResponse.TokenSource);
         }
 
         private void ValidateAccountIdentifiers(ClientInfo fromServer)
--- a/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
+++ b/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
@@ -27,5 +27,11 @@
   <ItemGroup>
     <ReferenceAssemblyProjectReference Include="..\Microsoft.Identity.Client.Ref\Microsoft.Identity.Client.Ref.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <None Remove="AuthenticationResultMetadata.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="AuthenticationResultMetadata.cs" />
+  </ItemGroup>
 
 </Project>
--- a/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
+++ b/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
@@ -27,11 +27,5 @@
   <ItemGroup>
     <ReferenceAssemblyProjectReference Include="..\Microsoft.Identity.Client.Ref\Microsoft.Identity.Client.Ref.csproj" />
   </ItemGroup>
-  <ItemGroup>
-    <None Remove="AuthenticationResultMetadata.cs" />
-  </ItemGroup>
-  <ItemGroup>
-    <Compile Include="AuthenticationResultMetadata.cs" />
-  </ItemGroup>
 
 </Project>
--- a/src/client/Microsoft.Identity.Client/TokenSource.cs
+++ b/src/client/Microsoft.Identity.Client/TokenSource.cs
@@ -12,17 +12,13 @@ namespace Microsoft.Identity.Client
     public enum TokenSource
     {
         /// <summary>
-        /// The source of access and Id token is not set.
+        /// The source of the access and Id token is Identity Provider - Azure Active Directory (AAD), ADFS or AAD B2C.
         /// </summary>
-        None = 0,
+        IdentityProvider,
         /// <summary>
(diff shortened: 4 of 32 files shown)


src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// Contains metadata of the authentication result.
+    /// </summary>
+    public class AuthenticationResultMetadata
+    {
+        /// <summary>
+        /// Default constructor for the class AuthenticationResultMetadata
+        /// </summary>
+        public AuthenticationResultMetadata()
```
bgavrilMS, line 17: I don't see any reason for this object to be mutable (i.e. have a default ctor and a property setter). Immutable objects are  much easier to work with - see https://www.linkedin.com/pulse/20140528113353-16837833-6-benefits-of-programming-with-immutable-objects-in-java/ ... Mutability for this particular object is not very important, but it's good practice to be immutable by default.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -27,5 +27,11 @@
   <ItemGroup>
     <ReferenceAssemblyProjectReference Include="..\Microsoft.Identity.Client.Ref\Microsoft.Identity.Client.Ref.csproj" />
   </ItemGroup>
+  <ItemGroup>
```
bgavrilMS, line 30: Please undo the changes to this file. Visual Studio gets a little confused sometimes when adding files, but the general pattern is that we include all .cs files by default, so this is not needed.

src/client/Microsoft.Identity.Client/TokenSource.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Specifies the source of the access and Id tokens in the authentication result.
```
bgavrilMS, line 10: I made some changes to the comments - please have a look. I though GitHub would "propose changes"... I guess I'm not a GitHub power user...

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -133,6 +133,7 @@ public async Task ConfidentialClientWithCertificateTestAsync()
             appCacheRecorder.AssertAccessCounts(1, 1);
             Assert.IsTrue(appCacheRecorder.LastAfterAccessNotificationArgs.IsApplicationCache);
             Assert.IsTrue(appCacheRecorder.LastAfterAccessNotificationArgs.HasTokens);
+            Assert.AreEqual(authResult.AuthenticationResultMetadata.TokenSource, TokenSource.IdentityProvider);
```
bgavrilMS, line 136: nit: in the Assert methods, the "expected" value is the first value.

jennyf19:
> using System;

license info.

---
Refers to: src/client/Microsoft.Identity.Client/TokenSource.cs:1 in f6dd77f. [](commit_id = f6dd77f5f45237e0f31cab33808b29056b0ab67a, deletion_comment = False)

src/client/Microsoft.Identity.Client/TokenSource.cs
jennyf19: lol. 

---
In reply to: [453897525](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1933#discussion_r453897525) [](ancestors = 453897525)

src/client/Microsoft.Identity.Client/TokenSource.cs
```diff
@@ -0,0 +1,31 @@
...
+        None = 0,
+        /// <summary>
+        /// The source of access and Id token is MSAL's cache.
+        /// </summary>
+        Cache,
+        /// <summary>
+        /// The source of the access and Id token is Identity Provider - Azure Active Directory (AAD), ADFS or AAD B2C.
+        /// </summary>
+        IdentityProvider,
+        /// <summary>
```
jennyf19, line 26: nit: would you want to separate b2c from AAD and ADFS or does it matter? Should it be called `AuthorizationServer`? When i think of an Identity provider, I think of Google or FB or Microsoft...really just a question

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/SilentAuthTests.cs
```diff
@@ -155,6 +155,7 @@ public async Task SilentAuth_TokenCacheRemainsPersistent_Async()
             Assert.IsFalse(at1.Equals(at2, System.StringComparison.InvariantCultureIgnoreCase));
             Assert.IsFalse(at1.Equals(at3, System.StringComparison.InvariantCultureIgnoreCase));
             Assert.IsFalse(at2.Equals(at3, System.StringComparison.InvariantCultureIgnoreCase));
+            Assert.AreEqual(TokenSource.IdentityProvider, authResult.AuthenticationResultMetadata.TokenSource);
         }
```
jennyf19, line 159: can you assert earlier, like after line 133, that the TokenSource is `TokenSource.Cache`...I guess finding one that does AT silent and pulls from the cache. I would just add this check in all the tests, what about the interactive tests and broker tests?

src/client/Microsoft.Identity.Client/TokenSource.cs
jmprieur: You need to use the +/- icon to proposed suggestions, @bgavrilMS 
image

src/client/Microsoft.Identity.Client/TokenSource.cs
bgavrilMS: The goal of this small improvement is for developers, especially on Identity.Web, to monitor their token cache hit ratio. There are other proposals around providing more details such as "which broker (name and version) was used?", but will treat them separately. 

Not sure about the naming, I proposed the OAuth2 term. @jmprieur - what do you think?

src/client/Microsoft.Identity.Client/TokenSource.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Specifies the source of the access and Id tokens in the authentication result.
+    /// </summary>
+    public enum TokenSource
+    {
+        /// <summary>
+        /// The source of access and Id token is not set.
+        /// </summary>
+        None = 0,
```
bgavrilMS, line 17: It looks like you are using this value for internal logic, but it does not have any meaning in the public API. There is always a source for the token. Let's not expose this.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -210,7 +210,8 @@ protected async Task<AuthenticationResult> CacheTokenResponseAndCreateAuthentica
                 atItem, 
                 idtItem, 
                 AuthenticationRequestParameters.AuthenticationScheme,
-                AuthenticationRequestParameters.RequestContext.CorrelationId);
+                AuthenticationRequestParameters.RequestContext.CorrelationId,
+                msalTokenResponse.TokenSource == TokenSource.None? TokenSource.IdentityProvider : msalTokenResponse.TokenSource);
```
bgavrilMS, line 214: Why can't msalTokenResponse.TokenSource always have the correct value?

src/client/Microsoft.Identity.Client/TokenSource.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// </summary>
+        Cache,
+        /// <summary>
+        /// The source of the access and Id token is Identity Provider - Azure Active Directory (AAD), ADFS or AAD B2C.
+        /// </summary>
+        IdentityProvider,
+        /// <summary>
+        /// The source of the access and Id token is a broker application - Authenticator or Company Portal. Brokers are supported only on Android and iOS.
+        /// </summary>
+        Broker
```
bgavrilMS, line 29: There are no unit tests that cover this. While the brokers are really only implemented in mobile code, which is not testable, it would be valuable to have a test that proves that the TokenSource value is indeed copied from MsalTokenSource to AuthenticationResult. I believe there are (a rather few) tests that mock IBroker, such as https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ad7970c3904d25662e5d756d076f777b0ed1867a/tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/BrokerRequestTests.cs#L183 - you can use this approach to unit test this.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: There are methods in the MsalTokenResponse class to build response for broker. So I am setting the value of token source to broker there. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1933/files#diff-ea5df3781a2dc2119f53390344fbfbd7R136

This method will either have the token source as Broker or Identity Provider. So instead I can check if Broker is set and otherwise set Identity Provider.

src/client/Microsoft.Identity.Client/TokenSource.cs
neha-bhargava: I was using it for internal logic but also I wanted to keep default to some other value in case the token source is not set or missed in some flow. Maybe rename None to Unidentified to make it more relevant?

src/client/Microsoft.Identity.Client/TokenSource.cs
bgavrilMS: Actually this test is a better fit: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/029e6031a1bccf88c5e31154712fcaa60a1fb4dd/tests/Microsoft.Identity.Test.Unit.net45/RequestsTests/InteractiveRequestOrchestrationTests.cs#L106

trwalke:
>             return await _silentRequest.CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);

If the access token expires the refresh token is used to refresh the token and acquire a new access token. Thus, the token is actually coming from AAD in this case. are you accounting for this scenario somewhere?

---
Refers to: src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentClientAuthStrategy.cs:119 in 0e4447b. [](commit_id = 0e4447bc28955687d648b04dccfe7d05293b59bb, deletion_comment = False)

trwalke:
>             return await _silentRequest.CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);

Also, is force refresh being handled? it should also be from AAD during silent calls

---
In reply to: [659063735](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1933#issuecomment-659063735) [](ancestors = 659063735)

---
Refers to: src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentClientAuthStrategy.cs:119 in 0e4447b. [](commit_id = 0e4447bc28955687d648b04dccfe7d05293b59bb, deletion_comment = False)

trwalke:
> > ```
> >         return await _silentRequest.CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
> > ```
> 
> If the access token expires the refresh token is used to refresh the token and acquire a new access token. Thus, the token is actually coming from AAD in this case. are you accounting for this scenario somewhere?
> 
> Refers to: src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentClientAuthStrategy.cs:119 in [0e4447b](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/0e4447bc28955687d648b04dccfe7d05293b59bb). [](commit_id = [0e4447b](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/0e4447bc28955687d648b04dccfe7d05293b59bb), deletion_comment = False)

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/SilentAuthTests.cs
bgavrilMS: Or after line 145, that AcquireTokenSilent gets a token from teh cache.

neha-bhargava: That one is withForceRefresh(True) so it would not get the token from the cache. Right?

src/client/Microsoft.Identity.Client/TokenSource.cs
neha-bhargava: Removed.

tests/Microsoft.Identity.Test.Integration.net45/HeadlessTests/SilentAuthTests.cs
neha-bhargava: Added in the other method where force refresh was false.

bgavrilMS: the assertion at line 158 is correct - force_refresh means "do not look for an AT in the cache".

Issue #1925, opened 2020-07-10
[Bug] MSAL is failing to migrate users from ADAL with "The ADAL cache is invalid"


trwalke:
**Which Version of MSAL are you using ?**
4.14+

**Platform**
xamarin android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

a. The app is in production, and I have upgraded to a new version of MSAL

The customer is experiencing issues signing in after upgrading to MSAL.NET related to ADAL caching. 

"Microsoft.Identity.Client.MsalClientException: The ADAL cache is invalid as it contains multiple refresh token entries for one user. Mitigation: Delete the ADAL cache. If you do not maintain an ADAL cache, this may be a bug in MSAL."

Over the course of 28 days, the customer has seen about 270 users with this issue

**Expected behavior**
User sign in after updating from ADAL to MSAL

**Actual behavior**
sign in fails with the following error "Microsoft.Identity.Client.MsalClientException: The ADAL cache is invalid as it contains multiple refresh token entries for one user. Mitigation: Delete the ADAL cache. If you do not maintain an ADAL cache, this may be a bug in MSAL."

What is the guidance to customers on resolving issues that come up like this?
how can the ADAL cache be deleted from MSAL?

**Stack Trace**
at Microsoft.Identity.Client.Cache.CacheFallbackOperations.GetAdalEntryForMsal (Microsoft.Identity.Client.Core.ICoreLogger logger, Microsoft.Identity.Client.Cache.ILegacyCachePersistence legacyCachePersistence, System.Collections.Generic.IEnumerable`1[T] environmentAliases, System.String clientId, System.String upn, System.String uniqueId) [0x00067] in <62c5f81aa08b40f6a5628154f09a4f0a>:0 
  at Microsoft.Identity.Client.TokenCache.Microsoft.Identity.Client.ITokenCacheInternal.FindRefreshTokenAsync (Microsoft.Identity.Client.Internal.Requests.AuthenticationRequestParameters requestParams, System.String familyId) [0x0027a] in <62c5f81aa08b40f6a5628154f09a4f0a>:0 
[...]

PR #1935, opened 2020-07-14
WIP Clear ADAL cache during error

trwalke:
Potential fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1925

This error occurs when that ADAL cache is invalid due to multiple RTs for one account. This change will clear the ADAL cache to allow the user to continue

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
@@ -326,7 +326,7 @@ namespace Microsoft.Identity.Client.Cache
                 adalCache.Clear();
 
                 throw new MsalClientException(
-                    MsalError.InvalidAdalCacheMultipleRTs, 
+                    MsalError.InvalidAdalCacheMultipleRTs,
                     MsalErrorMessage.InvalidAdalCacheMultipleRTs);
             }
 
--- a/src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
@@ -324,10 +324,9 @@ namespace Microsoft.Identity.Client.Cache
                     AdalCacheOperations.Deserialize(logger, legacyCachePersistence.LoadCache());
 
                 adalCache.Clear();
-
-                throw new MsalClientException(
-                    MsalError.InvalidAdalCacheMultipleRTs,
-                    MsalErrorMessage.InvalidAdalCacheMultipleRTs);
+                logger.Error(MsalErrorMessage.InvalidAdalCacheMultipleRTs);
+                legacyCachePersistence.WriteCache(AdalCacheOperations.Serialize(logger, adalCache));
+                return null;
             }
 
             return adalRts.FirstOrDefault();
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -318,6 +318,13 @@ internal static class CacheFallbackOperations
...
             if (rtGroupsByEnv.Any(g => g.Count() > 1))
             {
+                //Due to the fact that there is a problem with the ADAL cache that causes this exception, The ADAL cache will be removed when 
+                //this exception is triggered so that the can sign in interactivly and repopulate the cache.
+                IDictionary<AdalTokenCacheKey, AdalResultWrapper> adalCache =
+                    AdalCacheOperations.Deserialize(logger, legacyCachePersistence.LoadCache());
+
+                adalCache.Clear();
+
                 throw new MsalClientException(
```
trwalke, line 328: I know we discussed throwing a UI required exception but this is already throwing this so it would result in a breaking change right? @bgavrilMS

bgavrilMS: It is theoretically a breaking change indeed. But in practice, most apps already implement the pattern: 

try 
  - AquireTokenSiletn
catch UiRequiredEx
  - AcquireTokenInteractive

So I think throwing UIRequiredEx is still the right thing to do.

bgavrilMS: Most ADAL operations are implemented inside a `try / catch all `, but this one is an exception. So here's my suggestion

1. Instead of throwing, just log the error
2. return null. This will cause AcquireTokenSilent to not find any RTs, which makes it throw an MsalUiRequiredException, which is smth apps know how to handle

bgavrilMS: And also clear the ADAL cache, as it's broken.

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -318,8 +318,15 @@ internal static class CacheFallbackOperations
             // if we have more than 1 RT per env, there is smth wrong with the ADAL cache
             if (rtGroupsByEnv.Any(g => g.Count() > 1))
             {
+                //Due to the fact that there is a problem with the ADAL cache that causes this exception, The ADAL cache will be removed when 
+                //this exception is triggered so that the can sign in interactivly and repopulate the cache.
+                IDictionary<AdalTokenCacheKey, AdalResultWrapper> adalCache =
+                    AdalCacheOperations.Deserialize(logger, legacyCachePersistence.LoadCache());
+
+                adalCache.Clear();
```
bgavrilMS, line 326: by doing this you load the cache from persistence into memory and then clear it. But you need to write back the empty cache to persistence. Look at line 162 of this file for how to do it https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f032a4e3c6dde02f0c282a7e560ba49657a2f696/src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs#L162

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -318,8 +318,15 @@ internal static class CacheFallbackOperations
             // if we have more than 1 RT per env, there is smth wrong with the ADAL cache
             if (rtGroupsByEnv.Any(g => g.Count() > 1))
             {
+                //Due to the fact that there is a problem with the ADAL cache that causes this exception, The ADAL cache will be removed when
```
bgavrilMS, line 321: Try to write a unit test for this, there are already a bunch of tests here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f032a4e3c6dde02f0c282a7e560ba49657a2f696/tests/Microsoft.Identity.Test.Unit.net45/CacheTests/CacheFallbackOperationsTests.cs#L104

Issue #1912, opened 2020-07-02
[Bug] GetDeviceId sometimes takes 1.5s 


bgavrilMS:
**Which Version of MSAL are you using ?**
4.15

**Platform**
.net classic 

**Repro**

**Expected behavior**
AcquireTokenSilent normally takes ~300ms

**Actual behavior**
AcquireTokenSilent sometimes takes ~2s, whith a lot of time spent in GetDeviceId

**Possible Solution**
GetDeviceId on .net classic tries to enumerate the system NICs and use a network address as id. This can take a very long time when a VPN is configured on the system.

We should not call "get device id" because it is used in conjuction with MATS Telemetry. We do not use MATS telemetry at the moment.

neha-bhargava:
Fixed in [MSAL 4.17.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.17.0)

PR #1941, opened 2020-07-15
Disable all MATS telemetry if not configured

bgavrilMS:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1912

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/TelemetryTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/TelemetryTests.cs
@@ -122,6 +122,7 @@ namespace Microsoft.Identity.Test.Unit.PublicApiTests
 
         }
 
+#if DESKTOP
         [TestMethod]
         public async Task DoNotCallPlatformProxyAsync()
         {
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/TelemetryTests.cs
```diff
@@ -93,6 +102,71 @@ public void TelemetryInternalApiSample()
...
+
+            _telemetryManager = new TelemetryManager(
+                _serviceBundle.Config,
+                _platformProxy,
+                null,
+                true);
+
+            var correlationId = Guid.NewGuid().AsMatsCorrelationId();
+
+            var e1 = new ApiEvent(_logger, _crypto, correlationId) { Authority = new Uri("https://login.microsoftonline.com"), AuthorityType = "Aad" };
```
jennyf19, line 117: nit: I think we already have constant values for these

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/TelemetryTests.cs
```diff
@@ -93,6 +102,71 @@ public void TelemetryInternalApiSample()
...
+
+                harness.HttpManager.AddSuccessTokenResponseMockHandlerForPost(TestConstants.AuthorityCommonTenant);
+
+                // Act
+                await app
+                    .AcquireTokenInteractive(TestConstants.s_scope)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+
+                // Assert - NoDeviceIdProxy would fail if InternalGetDeviceId is called
```
jennyf19, line 152: there is no assert here?

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/TelemetryTests.cs
bgavrilMS: I'm sure we do, but the tests in this file do not use them. It's just an URI, has no meaning to telemetry.

tests/Microsoft.Identity.Test.Unit.net45/PublicApiTests/TelemetryTests.cs
bgavrilMS: Refactored to make assert more clear

Issue #1947, opened 2020-07-18
MsalExtensionException constructor always throws ArgumentNullException


FreddyD-GH:
**Which Version of MSAL are you using ?**
4.16.0

**Platform**
.NET Core 3.1

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Is this a new or existing app?**
This is a new app or experiment

**Repro**
Try to create a new MsalExtensionException by using the following constructor:
```csharp
public MsalExtensionException(string message, Exception innerException)
            : base(message, string.Empty, innerException)
```
No matter what you pass for arguments to the constructor, an ArgumentNullException will be thrown, because it is passing `string.Empty` as the message parameter to the base constructor, which causes any new MsalExtensionException to throw an ArgumentNullException, as you can see in the constructor for the parent class:
```csharp
public MsalException(string errorCode, string errorMessage, Exception innerException)
            : base(errorMessage, innerException)
        {
            if (string.IsNullOrWhiteSpace(Message))
            {
                throw new ArgumentNullException(nameof(errorMessage));
            }

            ErrorCode = errorCode;
        }
```
**Expected behavior**
There should be a way to not get an ArgumentNullException when using this constructor.

**Actual behavior**
An ArgumentNullException is thrown.

**Possible Solution**
Perhaps the order of the arguments being passed to the base constructor was inadvertently mixed up.  I'm not sure what the intent was here, but I would think that `base(string.Empty, message, innerException)` would probably make more sense than `base(message, string.Empty, innerException)`.

jennyf19:
@FreddyD-GH thanks for all the details on the issue. 
cc: @trwalke @neha-bhargava

bgavrilMS:
@FreddyD-GH - what are you trying to do ? It looks like this exception was made public by mistake, but it is not actually used by any of our extesnion points. 

[...]

PR #1949, opened 2020-07-20
Remove unsused MsalExtensionException

bgavrilMS:
Fix for #1947


bgavrilMS:
This exception type is public but was exposed by mistake. It is not used in any code path. And it is not usable (see #1947). 

So will delete without major version bump.

PR #1959, opened 2020-07-24
Update Java cache tests to always use the latest released version

bgavrilMS:


tests/CacheCompat/CommonCache.Test.MsalJava/JavaLanguageExecutor.cs
```diff
@@ -33,7 +33,7 @@ public JavaLanguageExecutor(string className)
 
             try
             {
-                string compileArguments = $"-f \"{pomFilePath}\" compile";
+                string compileArguments = $"-f \"{pomFilePath}\" compile versions:use-latest-releases";
```
bgavrilMS, line 36: Looks like this can be used to pick up the latest dependencies. It forces ALL dependecies to latest ver, but this project has very few so it's ok. I don't know how to make it use only MSAL latest.

henrik-me: What needs to be ensured is that we use the latest MSAL4J, dependencies not so much, we should just use what ever the Java team tested with imo

---
In reply to: [460200295](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1959#discussion_r460200295) [](ancestors = 460200295)

tests/CacheCompat/CommonCache.Test.MsalJava/pom.xml
```diff
@@ -24,12 +22,12 @@
         <dependency>
             <groupId>com.microsoft.azure</groupId>
             <artifactId>msal4j</artifactId>
-            <version>0.6.0-preview</version>
+            <version>1.6.1</version>
```
henrik-me, line 25: >1.6.1 [](start = 21, length = 5)

@petro, @avery : please ensure this is updated as part of Java releases.

Avery-Dunn: If we want this to always be updated as part of a release, couldn't we just have this dependency in the .pom to be ```<version>[1.6)</version>```

SomkaPe: Nice idea about grabbing latest version. 
<version>LATEST</version> is deprecated in maven 3
Recommended approach is using  "Versions Maven Plugin" - http://www.mojohaus.org/versions-maven-plugin/ 
+ passing versions:update-properties argument

tests/CacheCompat/CommonCache.Test.MsalJava/JavaLanguageExecutor.cs
SomkaPe: in order to pick up latest version of only one dependency "versions:update-properties" can be used -> "versions:update-properties updates properties defined in a project so that they correspond to the latest available version of specific dependencies. This can be useful if a suite of dependencies must all be locked to one version."

SomkaPe: also versions-maven-plugin should be added to pom.xml to use "versions:"

tests/CacheCompat/CommonCache.Test.MsalJava/pom.xml
bgavrilMS: Thanks @Avery-Dunn , I think your suggestion is the simplest and will go with it!

tests/CacheCompat/CommonCache.Test.MsalJava/JavaLanguageExecutor.cs
bgavrilMS: Will go with Avery's suggestion and use version interval.

Issue #1950, opened 2020-07-20
ID token related information is lost in the second call of AcquireTokenOnBehalfOf 


lnhzd:
To reproduce the issue, make 2 consecutive OBO calls with the same scope and same user assertion:

var result1 = Await this.confidentialApp.AcquireTokenOnBehalfOf(scopes, userAssertion).ExecuteAsync().ConfigureAwait(false);
var result2 = Await this.confidentialApp.AcquireTokenOnBehalfOf(scopes, userAssertion).ExecuteAsync().ConfigureAwait(false);

result1 has ID token, tenantId, uniqueId, accountName, etc populated
result2 has none of them populated.

2nd call seems to be getting token from cache, but ID token related information not populated on autehtnicationResult.
Is this by intention?

Thanks.

bgavrilMS:
No, this is a bug, will get it fixed asap. 

OBO call actually looks in the token cache because it tries to find access tokens associated with the same assertion. It seems that when looking into the cache, we find the access token but we don't look for the idtoken.

jmprieur:
@bgavrilMS is there enough information in the ID token to get the home tenant id ?

bgavrilMS:
The IDToken contains `oid` and `tid`, but I am not sure if the `tid` is the "home tenant id" ... it probably is though.

Why do you ask?

jmprieur:
to understand if we could surface the IAccount.

bgavrilMS:
I am afraid there isn't enough information for this. Imagine you do `AcquireTokenOBO(scope, assertion)` ... MSAL has to first read the cache and try to find an access token and ID token. But we don't have enough information to figure out `home_tenant_id` at this point

jmprieur:
@lnhzd why do you need the ID token? for a service to service call? do you mind explaining your scenario (out of curiosity)

lnhzd:
> @lnhzd why do you need the ID token? for a service to service call? do you mind explaining your scenario (out of curiosity)

not exactly our scenario but to simplify it: for logging purpose - so logging is associated with the user context.
I am aware that the information we need (e.g.: preferred_username claim, oid) can be extracted from the access token as well - but the way we USED to do it is to get an access token for calling SQL Azure (down stream service) whenever we need the display name - a bit lazy but the theory is we need to get access token for the downstream service anyway, it's easy just to get them from the typed object rather than going through raw claims.

[...]

PR #1968, opened 2020-08-03
Pull id token along AT when looking in the cache for OBO

bgavrilMS:
Fix for #1950

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
@@ -44,11 +44,15 @@ namespace Microsoft.Identity.Client.Internal.Requests
             MsalAccessTokenCacheItem msalAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
             if (msalAccessTokenItem != null)
             {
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -44,11 +44,15 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             MsalAccessTokenCacheItem msalAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
             if (msalAccessTokenItem != null)
             {
+                var msalIdTokenItem = await CacheManager.GetIdTokenCacheItemAsync(msalAccessTokenItem.GetIdTokenItemKey()).ConfigureAwait(false);
```
bgavrilMS, line 47: We already do this on SilentRequest: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d36d5867d1af9f969b6aed805d746c650f9106f0/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentClientAuthStrategy.cs#L124

jennyf19:
there shouldn't be a need to return the idToken in an OBO call, as there is no user.

Issue #1956, opened 2020-07-23
Support for regional ESTS.


bgavrilMS:
To increase the reliability, availability and perf of Azure, regionalization aims to keep all trafic inside a geographical area. For example, if an app needs to fetch data from Key Vault in WestUs2, all the traffic this entails - including MSAL generated traffic - should stay in WestUs2.

Proposal
The following proposal is an algorithm composed of several steps. Step 1 introduces new public API and has a discussion point.

Step 1
MSALs must provide the ability to configure regional
- cca.AcquireTokenForClient().WithAzureRegion(false).Execute()

Step 2
MSAL discovers the region
- by reading ENV variable REGION_NAME (works for Azure Functions)
- by making a call to the [local IMDS endpoint](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#instance-api)  (works for Azure VMs)
- otherwise fail with a meaningful exception

Step 3
MSAL computes the regional authority based on original authority.
- For the public cloud, the env is region + login.microsoft.com. Sovereign clouds are not yet supported, and MSAL should just throw.
- Regional authority: https://westus2.login.microsoft.com/12345-6789

Step 4
MSAL performs an instance discovery call to https://westus2.login.microsoft.com/common/discovery/instance 
- If the call succeeds, then proceed with using the preferred_network, preffered_cache and aliases.
- If the call fails, then proceed with using https://westus2.login.microsoft.com/12345-6789/  as a known authority, i.e. as if the instance metadata is:
`  "metadata": [
    {
      
      "preferred_network": "westus2.login.microsoft.com",
      "preferred_cache": "westus2.login.microsoft.com", 
      "aliases": [ " westus2.login.microsoft.com" ]
    },`

Additional details can be found here: [AAD SDK Proposal to Pin Auth to region.md](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FPinAuthToRegion%2FAAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md&_a=preview&version=GBdev)

trwalke:
fixed in [Release 4.18.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.18.0)

PR #1972, opened 2020-08-03
Add regional support

neha-bhargava:
#1956


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -69,6 +70,20 @@ public AcquireTokenForClientParameterBuilder WithSendX5C(bool withSendX5C)
...
         }
 
+        /// <summary>
+        /// Specifies if the token request should be sent to regional ESTS.
+        /// By default the value for this variable is false.
+        /// </summary>
+        /// <param name="withAzureRegion"><c>true</c> if the token request should be sent to regional ESTS. Otherwise <c>false</c>.
+        /// The default is <c>false</c></param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenForClientParameterBuilder WithAzureRegion(bool withAzureRegion)
```
bgavrilMS, line 80: maybe `bool autoDetectRegion` ?

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
```diff
@@ -16,12 +16,17 @@ internal class AcquireTokenForClientParameters : IAcquireTokenParameters
         /// </summary>
         public bool SendX5C { get; set; }
 
+        /// <summary>
+        /// </summary>
+        public bool WithAzureRegion { get; set; }
```
bgavrilMS, line 21: `AutoDetectRegion`?

src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
```diff
@@ -28,9 +30,17 @@ void AddToKnownCache(InstanceDiscoveryMetadataEntry entry)
...
+                foreach (string alias in entry.Aliases)
+                {
+                    s_knownPublicEnvironments.Add(alias);
+                }
+            }
+
             InstanceDiscoveryMetadataEntry publicCloudEntry = new InstanceDiscoveryMetadataEntry()
             {
-                Aliases = new[] { "login.microsoftonline.com", "login.windows.net", "login.microsoft.com", "sts.windows.net" },
+                Aliases = new[] { "login.microsoftonline.com", "login.windows.net", "login.microsoft.com", "sts.windows.net", "login.windows-ppe.net" },
```
bgavrilMS, line 43: I think PPE would be a different entry, not an alias of prod

src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
```diff
@@ -28,9 +30,17 @@ void AddToKnownCache(InstanceDiscoveryMetadataEntry entry)
                 }
             }
 
+            void AddToPublicEnvironment(InstanceDiscoveryMetadataEntry entry)
```
bgavrilMS, line 33: We tend to always specify the modifier, i.e. "private"

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -86,6 +83,24 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+        {
+            if (AuthenticationRequestParameters.Scope == null || !AuthenticationRequestParameters.Scope.Any())
+            {
+                throw new MsalClientException(
+                    MsalError.ScopesRequired,
+                    MsalErrorMessage.ScopesRequired);
+            }
+
+            // Throw exception if WithAzureRegion is set to true and the cloud is not public.
+            if (_clientParameters.WithAzureRegion && !KnownMetadataProvider.isKnownPublicEnvironment(AuthenticationRequestParameters.AuthorityInfo.Host))
```
bgavrilMS, line 96: I don't understand this validation: 

1. we should not enforce the fact that this feature is only supported in public cloud in MSAL, it's not our job. They may support it in other clouds.
2. what would happen if the authority would not be known to MSAL (since MSAL does not have an authoritative list of known authorities)? Can't we still regionalize it?

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -114,7 +114,8 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
 
         public async Task<InstanceDiscoveryMetadataEntry> GetMetadataEntryAsync(
             string authority, 
-            RequestContext requestContext)
+            RequestContext requestContext,
+            bool withAzureRegion = false)
```
bgavrilMS, line 118: isn't this flag available on the RequestContext? Similar to requestContext.ServiceBundle.Config.AuthorityInfo ?

src/client/Microsoft.Identity.Client/Instance/AuthorityEndpoints.cs
```diff
@@ -40,13 +42,26 @@ private async static Task UpdateAuthorityWithPreferredNetworkHostAsync(Authentic
...
+                    requestParameters.RequestContext,
+                    requestParameters.WithAzureRegion)
                 .ConfigureAwait(false);
 
             requestParameters.Authority = Authority.CreateAuthorityWithEnvironment(
                     requestParameters.AuthorityInfo,
                     metadata.PreferredNetwork);
         }
+
+        private async static Task<string> buildAuthorityWithRegionAsync(AuthorityInfo authorityInfo)
```
bgavrilMS, line 57: Did you explore the possibility of putting this logic in the DIscoveryManager itself? It could be a DiscoveryProvider on its own.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -86,6 +83,24 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             }
         }
 
+        private void validateRequest()
```
bgavrilMS, line 86: Please use PascalCasing, i.e. `ValidateRequest` for all methods, private and public.

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
```diff
@@ -0,0 +1,88 @@
...
+        private static RegionDiscovery Instance = null;
+        private HttpManager HttpManager;
+
+        private RegionDiscovery()
+        {
+            ImdsUri = new Uri("http://169.254.169.254/metadata/instance/compute");
+            Headers = new Dictionary<string, string>();
+            Headers.Add("api-version", "2019-06-01");
+            Logger = new NullLogger();
+            HttpManager = new HttpManager(new SimpleHttpClientFactory());
```
bgavrilMS, line 31: Please don't new up any of these objects. They are all avaialble on the service bundle, and will have been potentially configured by the user.

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        // For Unit tests only.
```
bgavrilMS, line 34: For unit testing, we have the ability to mock all services from the bundle.

bgavrilMS: But in general I prefer to have 2 constructors - one for product code and one for test code, smth like 

```csharp
public ClassUnderTest(IFoo fooDependnecyForTest, IBar barDepedencyForTest = null)
{
if (fooDependnecyForTest==null) { fooDependnecyForTest = new ... };
}
```

This approach allows you to keep the object immutable. And we like immutable, because [immutable is thread safe](http://tutorials.jenkov.com/java-concurrency/thread-safety-and-immutability.html) and keeps complexity down as well.

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
```diff
@@ -0,0 +1,88 @@
...
+        public void setHttpManager(HttpManager httpManager)
+        {
+            HttpManager = httpManager;
+        }
+
+        internal static RegionDiscovery GetInstance
```
bgavrilMS, line 40: Please avoid singletons if possible, they are hard to test. In this case I would put all this logic inside the DiscoveryManager as a provider.

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            get
+            {
+                if (Instance == null)
+                {
+                    lock (_lock)
```
bgavrilMS, line 46: Just as an observation, this singleton is not thread safe. 2 threads could reach the line 46 at the same time, one wait after  the other, and you'll have to 2 instances of your object. You need to acquire the lock and then check if Instance == null. 

This is great thread to read about singletons in C#: https://csharpindepth.com/articles/singleton

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
```diff
@@ -0,0 +1,88 @@
...
+                    }
+                }
+
+                return Instance;
+            }
+        }
+
+        internal async Task<string> getRegionAsync()
+        {
+            if (!Environment.GetEnvironmentVariable(RegionName).IsNullOrEmpty())
```
bgavrilMS, line 58: possible null ref.

neha-bhargava: No, its a hash table it either returns a value or null. Which is captured with IsNullOrEmpty().

bgavrilMS: I was trying to say that `null.IsNullOrEmpty()` will result in a null ref. You can use `string.IsNullOrEmptu(myString)` instead

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -69,6 +70,20 @@ public AcquireTokenForClientParameterBuilder WithSendX5C(bool withSendX5C)
             return this;
         }
 
+        /// <summary>
```
bgavrilMS, line 73: CC @jmprieur for new public API and comments

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -69,6 +70,20 @@ public AcquireTokenForClientParameterBuilder WithSendX5C(bool withSendX5C)
...
+        /// Specifies if the token request should be sent to regional ESTS.
+        /// By default the value for this variable is false.
```
bgavrilMS, line 75: Please add an aka.ms link

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ConfidentialClientExecutor.cs
```diff
@@ -57,6 +57,8 @@ public ConfidentialClientExecutor(IServiceBundle serviceBundle, ConfidentialClie
                 _confidentialClientApplication.AppTokenCacheInternal);
 
             requestParams.SendX5C = clientParameters.SendX5C;
+            requestParams.AutoDetectRegion = clientParameters.AutoDetectRegion;
```
bgavrilMS, line 60: Do we need this flag in 2 places?

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -85,6 +85,7 @@ private AuthorityInfo(string host, string canonicalAuthority, AuthorityType auth
         public AuthorityType AuthorityType { get; }
         public string UserRealmUriPrefix { get; }
         public bool ValidateAuthority { get; }
+        public bool AutoDetectRegion { get; set; }
```
bgavrilMS, line 88: I'm wondering a setter can be avoided here to keep this object immutable.

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ConfidentialClientExecutor.cs
```diff
@@ -57,6 +57,8 @@ public ConfidentialClientExecutor(IServiceBundle serviceBundle, ConfidentialClie
...
+            requestContext.ServiceBundle.Config.AuthorityInfo.AutoDetectRegion = clientParameters.AutoDetectRegion;
```
bgavrilMS, line 61: We should throw a nice exception if user tries to use this flag with ADFS authrority.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+        private readonly ICoreLogger _logger;
+        private readonly IHttpManager _httpManager;
+
+        public RegionDiscoveryProvider(IHttpManager httpManager, ICoreLogger logger = null)
+        {
+            _httpManager = httpManager;
+            _ImdsUri = new Uri("http://169.254.169.254/metadata/instance/compute/api-version=2019-06-01");
+            Headers = new Dictionary<string, string>();
+            Headers.Add("Metadata", "true");
+            _logger = logger ?? new NullLogger();
```
bgavrilMS, line 29: We always have a logger at this point, no need to check.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    internal sealed class RegionDiscoveryProvider : IRegionDiscoveryProvider
+    {
+        private const string RegionName = "REGION_NAME";
+        private readonly Uri _ImdsUri;
+        private IDictionary<string, string> Headers;
+        private readonly ICoreLogger _logger;
+        private readonly IHttpManager _httpManager;
+
+        public RegionDiscoveryProvider(IHttpManager httpManager, ICoreLogger logger = null)
```
bgavrilMS, line 23: logger should be mandatory.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+            _ImdsUri = new Uri("http://169.254.169.254/metadata/instance/compute/api-version=2019-06-01");
+            Headers = new Dictionary<string, string>();
+            Headers.Add("Metadata", "true");
+            _logger = logger ?? new NullLogger();
+
+        }
+
+        public async Task<string> getRegionAsync()
+        {
+            if (!Environment.GetEnvironmentVariable(RegionName).IsNullOrEmpty())
```
bgavrilMS, line 35: You'll get a null ref here because if the env variable does not exist, Env.GetEnvVariable return null and `null.IsNullOrEmpty()` throw a null ref 

Please use `!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(RegionName))` ... and if the env variable is not found, then this evalues to `!string.IsNullOrEmpty(null)` which evalues to `false`

src/client/Microsoft.Identity.Client/Region/IRegionDiscoveryProvider.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Region
+{
+    internal interface IRegionDiscoveryProvider
```
bgavrilMS, line 9: This provider does not behave like the other providers and it does too little, and so you are moving a lot of the logic to the manager. (In hindsight, I should have created an IProvider interface to enforce this...)

Please have a look at `NetworkMetadataProvider`, as the region provider is almost identical in functionality to it.

- does some stuff (env variable / network call) too figure out the region
- once you figure out the region, inject it into the static cache
- return an `InstanceDiscoveryMetadataEntry`

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -116,7 +122,14 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
             string authority, 
             RequestContext requestContext)
         {
+            var autoDetectRegion = requestContext.ServiceBundle.Config.AuthorityInfo.AutoDetectRegion;
```
bgavrilMS, line 125: the provider itself should make this check, not the manager

src/client/Microsoft.Identity.Client/Region/IRegionDiscoveryProvider.cs
bgavrilMS: After looking at the code in more detail, I think your solution is simpler and it is ok,

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+                }
+
+                LocalImdsResponse localImdsResponse = JsonHelper.DeserializeFromJson<LocalImdsResponse>(response.Body);
+
+                _logger.Info($"[Region discovery] Call to local IMDS returned region: {localImdsResponse.location}");
+                return localImdsResponse.location;
+            }
+            catch (Exception e)
+            {
+                _logger.Info("[Region discovery] Call to local imds failed." + e.Message);
```
bgavrilMS, line 59: maybe use .Error here.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -86,6 +83,24 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            }
+
+            // Throw exception if WithAzureRegion is set to true and the cloud is not public.
+            if (_clientParameters.AutoDetectRegion && !KnownMetadataProvider.isKnownPublicEnvironment(AuthenticationRequestParameters.AuthorityInfo.Host))
+            {
+                throw new MsalClientException(
+                    MsalError.RegionDiscoveryNotEnabled, 
+                    MsalErrorMessage.RegionDiscoveryNotAvailable);
+            }
+        }
```
bgavrilMS, line 102: Consider adding an extra check against ADFS authority type.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -86,6 +83,24 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+        {
+            if (AuthenticationRequestParameters.Scope == null || !AuthenticationRequestParameters.Scope.Any())
+            {
+                throw new MsalClientException(
+                    MsalError.ScopesRequired,
+                    MsalErrorMessage.ScopesRequired);
+            }
+
+            // Throw exception if WithAzureRegion is set to true and the cloud is not public.
+            if (_clientParameters.AutoDetectRegion && !KnownMetadataProvider.isKnownPublicEnvironment(AuthenticationRequestParameters.AuthorityInfo.Host))
```
bgavrilMS, line 96: As discussed, let's not block the users on this.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            _httpManager = httpManager;
+            _ImdsUri = new Uri("http://169.254.169.254/metadata/instance/compute/api-version=2019-06-01");
+            Headers = new Dictionary<string, string>();
+            Headers.Add("Metadata", "true");
+            _logger = logger ?? new NullLogger();
+
+        }
+
+        public async Task<string> getRegionAsync()
```
bgavrilMS, line 33: naming - PascalCase

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -195,5 +217,15 @@ private static InstanceDiscoveryMetadataEntry CreateEntryForSingleAuthority(Uri
                 PreferredNetwork = authority.Host
             };
         }
+
+        private async Task<string> BuildAuthorityWithRegionAsync(string canonicalAuthority)
```
bgavrilMS, line 221: Can you move this logic to the RegionProvider class?

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -195,5 +217,15 @@ private static InstanceDiscoveryMetadataEntry CreateEntryForSingleAuthority(Uri
...
+        {
+            string region = await _regionDiscoveryProvider.getRegionAsync().ConfigureAwait(false);
+
+            var builder = new UriBuilder(canonicalAuthority);
+            builder.Host = $"{region}.{builder.Host}";
```
bgavrilMS, line 226: I believe that we discussed the following rules: 

- if authority is a known public cloud, then regional authority will be `{region}.login.microsoft.com`
- otherwise, the regional authority will be `{region}.{authoirty}`

src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
```diff
@@ -68,6 +85,12 @@ void AddToKnownCache(InstanceDiscoveryMetadataEntry entry)
             AddToKnownCache(cloudEntryGermanay);
             AddToKnownCache(usGovCloudEntry);
             AddToKnownCache(usCloudEntry);
+            AddToPublicEnvironment(publicCloudEntry);
+        }
+
+        public static bool isKnownPublicEnvironment(string environment)
```
bgavrilMS, line 91: PascalCasing and naming (`IsPublicEnvironment` since "known" is already part of the class name)

src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
```diff
@@ -28,13 +30,28 @@ void AddToKnownCache(InstanceDiscoveryMetadataEntry entry)
...
+            }
+
             InstanceDiscoveryMetadataEntry publicCloudEntry = new InstanceDiscoveryMetadataEntry()
             {
                 Aliases = new[] { "login.microsoftonline.com", "login.windows.net", "login.microsoft.com", "sts.windows.net" },
                 PreferredNetwork = "login.microsoftonline.com",
                 PreferredCache = "login.windows.net"
             };
 
+            InstanceDiscoveryMetadataEntry ppePublicCloudEntry = new InstanceDiscoveryMetadataEntry()
```
bgavrilMS, line 48: not needed anymore?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
neha-bhargava: Changed.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
neha-bhargava: Discussed offline. Removed the check.

src/client/Microsoft.Identity.Client/Instance/AuthorityEndpoints.cs
neha-bhargava: Done.

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
neha-bhargava: Changed

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
neha-bhargava: This Object is created at app level and the AutoDetectRegion is set at request level. And hence the setter is required. I need the value in request context to retrieve it in Instance discovery manager. Could not figure out a better way.

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ConfidentialClientExecutor.cs
neha-bhargava: Right we dont need it in request params. Removing it from there.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Resolved.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
neha-bhargava: Removed,

src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
neha-bhargava: removed.

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
neha-bhargava: Changed.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Done

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Done

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
neha-bhargava: Sounds good.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
neha-bhargava: Added

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
neha-bhargava: removed.

src/client/Microsoft.Identity.Client/Region/RegionDiscovery.cs
neha-bhargava: changed

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
bgavrilMS: Ok, please note that AuthorityInfo has a copy constructor that needs to be updated : https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/241666c6756e30ca58bd88ed1e43c58203654f62/src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs#L73

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
bgavrilMS: Let's add a bit more context here, maybe:

```
If set, MSAL tries to auto-detect and use a regional Azure authority. This helps keep the authentication traffic inside the Azure region. If the region cannot be determined (e.g. not running on Azure), an exception is thrown.
```

src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
neha-bhargava: This is inside the static constructor and access modifier cannot be applied here.

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
neha-bhargava: Moved it into request context.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: There is no .Error variable here.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
bgavrilMS: As per latest discussions, comments must be kept minimal and should indicate that this feature requires a server side configuration (which won't be available for a while)

neha-bhargava: Updated.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
neha-bhargava: Added to the constructor.

Issue #1965, opened 2020-07-31
[Bug] AcquireTokenOnBehalfOf does not respect the WithAuthority modifer


jmprieur:
**Which Version of MSAL are you using ?**
4.17.0

**Platform**

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
* Web API
    * [x ] OBO

**Repro**

See also customer bug in Microsoft.Identity.Web: [Bug] Token acquisition from different tenant not working In progress P1 bug #[344](https://github.com/AzureAD/microsoft-identity-web/issues/344)

- Adele is a user of tenant contosodev, which is also a guest user of msidentitysamplestesting
- Developer has a web API that wants to acquire a token for Graph (user.read) on behalf of Adele but for each of the tenants (as the display name of Adele is not the same depending on the tenant). The admin consent was done for the web api in both tenants.

The Microsoft.Identity.Web code in the controller action of the Web API is:

```CSharp
string token1 = await _tokenAcquisition.GetAccessTokenForUserAsync(new string[] { "user.read" }, 
                                   "7f58f645-c190-4ce5-9de4-e2b7acd2a6ab").ConfigureAwait(false);
string token2 = await _tokenAcquisition.GetAccessTokenForUserAsync(new string[] { "user.read" }, 
                                   "3ebb7dbb-24a5-4083-b60c-5a5977aabf3d").ConfigureAwait(false); 
```

Which translates into 

```CSharp
var result1 = await _application
                 .AcquireTokenOnBehalfOf(scopes.Except(_scopesRequestedByMsal), 
                                        new UserAssertion(tokenUsedToCallTheWebApi))
                                        .WithAuthority(authorityOfTenant1)
                                        .ExecuteAsync()
                                        .ConfigureAwait(false);

var result2 = await _application
                 .AcquireTokenOnBehalfOf(scopes.Except(_scopesRequestedByMsal), 
                                        new UserAssertion(tokenUsedToCallTheWebApi))
                                        .WithAuthority(authorityOfTenant2)
                                        .ExecuteAsync()
                                        .ConfigureAwait(false);

token1 = result1.AccessToken;
token2 = result2.AccessToken;
```

**Expected behavior**
token1 and token2 should be different (in particular have a different issuer)

**Actual behavior**
[...]

PR #1973, opened 2020-08-03
Fix for #1965 - filter by tid on OBO if possible

bgavrilMS:
Fix for #1965


src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -312,30 +312,47 @@ private static IEnumerable<MsalAccessTokenCacheItem> FilterByTokenType(Authentic
...
+                    string.Equals(item.TenantId ?? string.Empty, requestTenantId ?? string.Empty, StringComparison.OrdinalIgnoreCase),
+                    requestParams.RequestContext.Logger,
+                    "Filtering by tenant id");
+            }
+            else
+            {
+                requestParams.RequestContext.Logger.Warning("Have not filtered by tenant ID. " +
+                    "This can happen in OBO scenario where authority is /common or /organizations. " +
+                    "Please use tenanted authority.");
+            }
```
jennyf19, line 351: @jmprieur should we have an aka.ms link or anything for this?

jennyf19:
sweet...thanks Bogdan! 

---
In reply to: [460447316](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1973#issuecomment-460447316) [](ancestors = 460447316)

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: A link would be good if we decide to throw an exception in this corner case, i.e. not allow users to call OBO with /common, which makes sense to me, but it is a breaking change.

Issue #1946, opened 2020-07-17
Implement POP (SHR) for confidential client


bgavrilMS:
What MSAL implements today: 

- [WithProofOfPossession](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/67325c7cabb9e4dbccc32c6764c865f088b8f3ad/src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs#L43) API
- this is available on .NET desktop for public client only
- there exists an internal interface `IPoPCryptoProvider` for key management
- key management is done by storing key on disk in a container (similar to DPAPI protection)

Tasks to complete: 

### First deliverable

1. Add an implementation of `IPoPCryptoProvider` that stores key in memory statically. Keys must be recycled every 8h. This should be the default implementation.
2. Expose public API on .net core and on confidentuil client

bgavrilMS:
CC @henrik-me - a spec for the PoP work

trwalke:
Continuing in Part 2 (https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2013)

trwalke:
fixed in [Release 4.18.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.18.0)

PR #1986, opened 2020-08-13
Pop implementation for Confidential Client

trwalke:
PR for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1946


henrik-me:
did you consider refactoring for re-use across net desktop and net core?

trwalke:
> did you consider refactoring for re-use across net desktop and net core?

@henrik-me  yes. Netcore is implemented here. one more commit to push

src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
```diff
@@ -33,6 +36,54 @@ public override Task<AuthenticationResult> ExecuteAsync(CancellationToken cancel
...
+        /// <param name="httpRequestMessage">An HTTP request to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request.</param>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</item>
+        /// <item> An Authentication header is automatically added to the request</item>
+        /// <item> The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters). </item>
+        /// <item> MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</item>
+        /// </list>
+        /// </remarks>
+        public T WithProofOfPosession(HttpRequestMessage httpRequestMessage)
```
bgavrilMS, line 55: Since this method already exists on the public client, it would be better to add a single public API to the base class `AbstractAcquireTokenParameterBuilder`. This way PoP will be available on both public and confidential client.

henrik-me:
I still see potential for refactoring's as there is code duplication which can be avoided.

---
In reply to: [673187877](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1986#issuecomment-673187877) [](ancestors = 673187877)

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Security.Cryptography;
+using Microsoft.Identity.Client.AuthScheme.PoP;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.Platforms.netcore
```
bgavrilMS, line 10: As @henrik-me  points out, you should be able to reuse this code between .net core and .net classic (and .net standard?!). We have a directory - Platforms/Shared with common implementations that you can put it.

trwalke: ah, ok. that makes more sense. Ill refactor

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                return GetOrCreatePersistededKey(ContainerName);
+            }
+        }
+
+        // This is a singleton because the key is the same on a device
+        private static readonly Lazy<NetCorePoPCryptoProvider> lazyInstance =
+            new Lazy<NetCorePoPCryptoProvider>(() => new NetCorePoPCryptoProvider());
+
+        public static NetCorePoPCryptoProvider Instance { get { return lazyInstance.Value; } }
```
bgavrilMS, line 53: Please be aware that by using the singleton pattern here, only the constructor is thread safe. All other methods are not thread safe.

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// </summary>
+    /// <remarks>
+    /// Key creation and storage only works on Windows. See https://stackoverflow.com/questions/41986995/implement-rsa-in-net-core/42006084 for more details.
+    /// </remarks>
+    internal class NetCorePoPCryptoProvider : IPoPCryptoProvider, IDisposable
+    {
+        internal /* internal for test only */ const int RsaKeySize = 2048;
+        internal /* internal for test only */ const string ContainerName = "com.microsoft.msal";
+        private static RSA s_SigningKey;
+        private bool s_PersistKey;
```
bgavrilMS, line 24: _persistKey (the `s` prefix means static, which this variable is not)

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public string CannonicalPublicKeyJwk { get; }
+
+        private NetCorePoPCryptoProvider(bool persistKey = false)
+        {
+            s_PersistKey = persistKey;
+
+            RSAParameters publicKeyInfo = s_SigningKey.ExportParameters(false);
+
+            CannonicalPublicKeyJwk = ComputeCannonicalJwk(publicKeyInfo);
```
bgavrilMS, line 63: If the key expires, you create a new one, but you don't update the CannonicalPublicKeyJwk

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            get
+            {
+                if (s_PersistKey)
+                {
+                    if (CheckKeyExpiration())
+                    {
+                        s_SigningKey = RSA.Create();
+                        s_SigningKey.KeySize = RsaKeySize;
+                        s_KeyTimeValidTo = DateTime.Now + TimeSpan.FromSeconds(s_DefaultKeyExpirationTime);
```
bgavrilMS, line 38: Please use only UTC times, i.e. DateTime.UtcNow

bgavrilMS: Or use DateTimeOffset...

 In any case, DateTime represents regional time, but here you only care about instantenous time. Solution is to use DateTime.UtcNow or DateTimeOffset 

See https://stackoverflow.com/questions/4331189/datetime-vs-datetimeoffset for details

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace Microsoft.Identity.Client.Platforms.netcore
+{
+    /// <summary>
+    /// This provider is based on the RSACryptoServiceProvider container concept. 
+    /// It will create and store a pair of keys in the context of the current user. Subsequent calls to this provider will reuse this key.
+    /// </summary>
+    /// <remarks>
+    /// Key creation and storage only works on Windows. See https://stackoverflow.com/questions/41986995/implement-rsa-in-net-core/42006084 for more details.
+    /// </remarks>
+    internal class NetCorePoPCryptoProvider : IPoPCryptoProvider, IDisposable
```
bgavrilMS, line 19: We discussed offline about persisting the key to disk or not. I believe now that is a bit too complex as there are too many issues in the code below (e.g. you persist the key even s_PersistKey is set to false). Since we do not need this functionality now, let's focus on quality and speed and abandon this persistence layer for now (which should not be used anyway) and favour the in-memory storage with expiration.

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+
+
+        public byte[] Sign(byte[] payload)
+        {
+            return RsaKey.SignData(payload, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+        }
+
+        private bool CheckKeyExpiration()
+        {
+            return ConvertToTimeT(s_KeyTimeValidTo) < ConvertToTimeT(DateTime.Now);
```
bgavrilMS, line 74: no need to convert, date times are comparable. Just make sure you use UTC

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// It will create and store a pair of keys in the context of the current user. Subsequent calls to this provider will reuse this key.
+    /// </summary>
+    /// <remarks>
+    /// Key creation and storage only works on Windows. See https://stackoverflow.com/questions/41986995/implement-rsa-in-net-core/42006084 for more details.
+    /// </remarks>
+    internal class NetCorePoPCryptoProvider : IPoPCryptoProvider, IDisposable
+    {
+        internal /* internal for test only */ const int RsaKeySize = 2048;
+        internal /* internal for test only */ const string ContainerName = "com.microsoft.msal";
+        private static RSA s_SigningKey;
```
bgavrilMS, line 23: If the class is a singleton, it acts like everything is static. Consider removing the singleton pattern and just create a new instance everytime and rely on these statics.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
trwalke: This class did not have access to the serviceBundle so I wasnt sure how to access it without dramatically changing the class. any thoughts

bgavrilMS: Yeah, those objects are complex ... So your derived class has the ServiceBundle, then your base class should also be able to get a hold of it. In an inheritance hierarchy (e.g. Animal > Cat > PersianCat), the order in which each of the underlying objects are created is from base class to most derived class (i.e. Animal() gets build, then Cat(), then PersianCat()). However, you do have the ability to pass around params through each of these constructors, smth like: 

```csharp
        internal AbstractPublicClientAcquireTokenParameterBuilder(
            IPublicClientApplicationExecutor publicClientApplicationExecutor)
            : base(publicClientApplicationExecutor.ServiceBundle) // like this
        {
            PublicClientApplicationExecutor = publicClientApplicationExecutor;
        }

```

and in the base class:

```csharp
         internal IServiceBundle ServiceBundle { get; }

        internal AbstractAcquireTokenParameterBuilder(IServiceBundle serviceBundle)
        {
            ServiceBundle = serviceBundle;
        }
```

bgavrilMS: This comment still stands.

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </summary>
+    internal class NetCorePoPCryptoProvider : IPoPCryptoProvider, IDisposable
+    {
+        internal /* internal for test only */ const int RsaKeySize = 2048;
+        internal /* internal for test only */ const string ContainerName = "com.microsoft.msal";
+        private static RSA s_SigningKey;
+        private DateTime _KeyTimeValidTo;
+        private readonly uint _DefaultKeyExpirationTime = 60 * 60 * 8; // Eight Hours
+
+        IClock Clock { get; set; }
```
bgavrilMS, line 25: Could you initialize this property via the ctor? It would make the class immutable, which has several advantages: 

- it is easier to maintain and understand
- it has improved thread safety

src/client/Microsoft.Identity.Client/PlatformsCommon/IClock.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.PlatformsCommon
+{
+    internal interface IClock
```
bgavrilMS, line 9: We already have an interface for this - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9df321e3b71a9c673cda812710147fe48a907005/src/client/Microsoft.Identity.Client/Utils/ITimeService.cs#L8

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.Platforms.netcore
+{
+    /// <summary>
+    /// The default implementation will store a key in memory for a period of 8 hours and recreate it to ensure the key is kept fresh.
+    /// Subsequent calls to this provider will reuse this key.
+    /// </summary>
+    internal class NetCorePoPCryptoProvider : IPoPCryptoProvider, IDisposable
```
bgavrilMS, line 17: You moved a bunch of code outside, but I am wondering if there is any difference between NetCorePoPProvider and NetDesktopPopProvider?

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        internal /* internal for test only */ const int RsaKeySize = 2048;
+        internal /* internal for test only */ const string ContainerName = "com.microsoft.msal";
```
bgavrilMS, line 20: no longer needed?

tests/Microsoft.Identity.Test.Unit.net45/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -82,7 +87,69 @@ public void ValidatePopRequestAndToken()
             Assert.IsTrue(jwkFromPopAssertion["jwk"].DeepEquals(initialJwk));
         }
 
-       
+        [TestMethod]
+        public void ValidateKeyExpiration()
```
bgavrilMS, line 91: Awesome!

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            Clock = new SystemClock();
+        }
+
+        public byte[] Sign(byte[] payload)
+        {
+            return PopCryptoHelper.Sign(RsaKey, payload);
+        }
+
+        public void Dispose()
```
bgavrilMS, line 63: I would not dispose of this, since the key can be shared by many instances

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        IClock Clock { get; set; }
+
+        public string CannonicalPublicKeyJwk { get; private set; }
+
+        private RSA RsaKey
+        {
+            get
+            {
+                lock (s_SigningKey)
```
bgavrilMS, line 33: yes, this is now thread safe!

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePoPCryptoProvider.cs
bgavrilMS: I looked at this in more detail, I think you can merge these 2 classes toghter and remove the POPHelper

src/client/Microsoft.Identity.Client/PlatformsCommon/IClock.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+       DateTime Now { get; }
```
bgavrilMS, line 11: Class not needed anymore?

src/client/Microsoft.Identity.Client/PlatformsCommon/NetSharedPoPCryptoProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+                    }
+
+                    return s_SigningKey;
+                }
+            }
+        }
+
+        public NetSharedPoPCryptoProvider(/*for testing only*/ITimeService timer = null)
+        {
+            if (timer == null)
```
bgavrilMS, line 65: nit: 

`_timer = timer ?? new TimeService()`

src/client/Microsoft.Identity.Client/PlatformsCommon/NetSharedPoPCryptoProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+                _timer = timer;
+            }
+        }
+
+        public byte[] Sign(byte[] payload)
+        {
+            return Sign(RsaKey, payload);
+        }
+
+        internal static bool CheckKeyExpiration(DateTime KeyTimeValidTo, ITimeService timer)
```
bgavrilMS, line 80: private

src/client/Microsoft.Identity.Client/PlatformsCommon/NetSharedPoPCryptoProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <summary>
+        /// Creates the cannonical representation of the JWK.  See https://tools.ietf.org/html/rfc7638#section-3
+        /// The number of parameters as well as the lexicographic order is important, as this string will be hashed to get a thumbprint
+        /// </summary>
+        internal static string ComputeCannonicalJwk(RSAParameters rsaPublicKey)
+        {
+            return $@"{{""{JsonWebKeyParameterNames.E}"":""{Base64UrlHelpers.Encode(rsaPublicKey.Exponent)}"",""{JsonWebKeyParameterNames.Kty}"":""{JsonWebAlgorithmsKeyTypes.RSA}"",""{JsonWebKeyParameterNames.N}"":""{Base64UrlHelpers.Encode(rsaPublicKey.Modulus)}""}}";
+        }
+
+        internal void SetTimer(ITimeService timer)
```
bgavrilMS, line 94: not needed

tests/Microsoft.Identity.Test.Unit.net45/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -82,7 +87,69 @@ public void ValidatePopRequestAndToken()
...
+                                                                            .WithAuthority(new Uri(ClientApplicationBase.DefaultAuthority), true)
+                                                                            .WithHttpManager(harness.HttpManager)
+                                                                            .WithTelemetry(new TraceTelemetryConfig())
+                                                                            .BuildConcrete();
+
+                harness.HttpManager.AddSuccessTokenResponseMockHandlerForPost(TestConstants.AuthorityCommonTenant);
+
+                Guid correlationId = Guid.NewGuid();
+
+                var provider = new NetSharedPoPCryptoProvider();
```
bgavrilMS, line 108: pass in your mock timeservice via the ctor and not via the setter

bgavrilMS (approved):
minor comments remaining, please resolve and commit

bgavrilMS (approved):

Issue #1995, opened 2020-08-17
[Bug] iOS broker redirect uri validation


last1here:
**Which Version of MSAL are you using ?**
MSAL 4.17.1

**Platform**
Xamarin Forms (iOS)

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Is this a new or existing app?**
Existing app upgrading MSAL from 4.16.1 to 4.17.1 and adding broker support.

**Repro**

```csharp
PCA = PublicClientApplicationBuilder.Create(authClientId)
                .WithTenantId(authTenantId)
                .WithIosKeychainSecurityGroup(authKeychain)
                .WithBroker()
                .WithRedirectUri($"msauth.{bundleId}://auth")
                .Build();
```

**Expected behavior**
The brokered flow completes correctly.

**Actual behavior**
The `MsalError.CannotInvokeBroker` exception thows. Works as expected in 4.16.1.

**Possible Solution**
From a quick check I think that the validation in the iOSBroker class is failing, specifically [this line](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs#L115) added as part of #1929 between 4.16.1 and 4.17.0. The `redirectUri.AbsoluteUri` returns the Uri with a trailing `/` which is then compared to expectedRedirectUri which is missing the trailing `/`. Adding a trailing `/` to expectedRedirectUri should resolve the issue.

bgavrilMS:
Thank you for reporting and investigating @last1here , I'll get this fixed ASAP.

markbeij:
We have this issue as well. This should be patched to 4.17.2 as this makes MSAL for iOS quite useless

trwalke:
fixed in [Release 4.18.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.18.0)

PR #1997, opened 2020-08-17
Fix for #1995 - bad redirect uri validation ios broker

bgavrilMS:
Fix for #1995

### DIFF
--- a/Libs.sln
+++ b/Libs.sln
@@ -62,6 +62,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "pipelines", "pipelines", "{
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NetFxConsoleApp", "tests\devapps\NetFxConsoleTestApp\NetFxConsoleApp.csproj", "{D02BB914-D0E3-45F5-A68B-59D079FA742E}"
 EndProject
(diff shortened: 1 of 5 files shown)


Libs.sln
```diff
@@ -248,6 +250,34 @@ Global
 		{D02BB914-D0E3-45F5-A68B-59D079FA742E}.Release|x64.Build.0 = Release|Any CPU
 		{D02BB914-D0E3-45F5-A68B-59D079FA742E}.Release|x86.ActiveCfg = Release|Any CPU
 		{D02BB914-D0E3-45F5-A68B-59D079FA742E}.Release|x86.Build.0 = Release|Any CPU
+		{0CE6E6B7-B914-4198-9C32-3B85CA4C2567}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
bgavrilMS, line 253: I tried to leave the "ref" project out of the SLN, but I'm getting all sort of errors :(

src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
```diff
@@ -109,26 +110,7 @@ public bool IsBrokerInstalledAndInvokable()
         private void ValidateRedirectUri(Uri redirectUri)
         {
             string bundleId = NSBundle.MainBundle.BundleIdentifier;
-            string expectedRedirectUri = $"msauth.{bundleId}://auth";
```
bgavrilMS, line 112: I moved this logic somewhere we can unit test it.

Issue #1948, opened 2020-07-18
[Bug] MSAL Suggested cache key can create performance issues to due to over serialization


RamjotSingh:
**Which Version of MSAL are you using ?**
4.16.1

**Platform**
netcore

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [x] OBO

Other? - please describe;

**Is this a new or existing app?**
The app is in production, and I have started using MSAL instead of ADAL

**Issue**
MSAL today in it's suggested cache key makes following suggestions

- App context - ClientId+_AppTokenCache
- User context
    - Silent acquire - User account Id
    - OnBehalfOf exchange - Original token hash

However there are a couple of issues I see with this

##### Issue with App context cache
We recently onboarded MSAL (shifting away from ADAL) and during rollout we saw a trace
`(False) MSAL 4.16.1.0 MSAL.NetCore Microsoft Windows 10.0.14393  [07/18/2020 13:52:56] Serializing token cache with 4247 items.`

This is happening because all app context tokens are being pushed into one single cache and since the tokens are tenantized, the moment a significant number of tenant hit the call the token cache just keeps on growing.

##### Issue with User context cache
For user context caches, since the cache key for OnBehalfOf exchange is the original token, as lots of callers make the call using unique clients (we have thousands of clients sending us potentially 100s of unique tokens per user) the token for same user gets cached over and over.

##### Solution which I propose
Based on the issues above, here is what I recommend as the token caches

- App context - ClientID+TenantId+_AppTokenCache
- User Context - ClientId + HomeAccountIdentifier (i.e. TenantID+UserId)

Additionally TokenCacheNotificationArgs should expose additional information like resourceId since for services which hit a lot of external services even tenant level cache might be too big and they might need to go granular.

bgavrilMS:
CC @jmprieur @jennyf19

bgavrilMS:
@jmprieur - would it not be enough to use home account id as the cache key? MSAL would still do the right thing and try to acquire OBO token with initial token hash...

RamjotSingh:
[...]

PR #2003, opened 2020-08-19
SuggestedCacheKey to use client ID and tenant ID

bgavrilMS:
Fix for #1948

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
@@ -16,16 +16,19 @@ namespace Microsoft.Identity.Client.Cache
 
             if (requestParameters.ApiId == TelemetryCore.Internal.Events.ApiEvent.ApiIds.AcquireTokenSilent)
             {
-                return requestParameters.Account?.HomeAccountId?.Identifier;
+                return requestParameters.Account?.HomeAccountId?.Identifier == null ? 
+                    null :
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
```diff
@@ -16,16 +16,19 @@ public static string GetKeyFromRequest(AuthenticationRequestParameters requestPa
 
             if (requestParameters.ApiId == TelemetryCore.Internal.Events.ApiEvent.ApiIds.AcquireTokenSilent)
             {
-                return requestParameters.Account?.HomeAccountId?.Identifier;
+                return requestParameters.Account?.HomeAccountId?.Identifier == null ?
```
bgavrilMS, line 19: @jmprieur - if you could review just this file, it captures all the actual changes, the rest are tests ...

jmprieur, line 19: Do we have the risk that the first time the pre-access event is called for OBO, we get a cache key, and the second time we get a different cache key? [we'd want it to be invariant and the same in all calls to before/after access]

jmprieur (commented):
I'm not sure about the tests.
We want all the calls to both before access and after access to have the same cache key

I think that testing if the home account id is null we induce a different cache key in the first call for OBO (as we don't know the user yet) ?

bgavrilMS:
Discussed offline with @jmprieur and decided to not take this for now as Microsoft.Identity.Web is in "ask mode"

Issue #1981, opened 2020-08-06
MSAL logging evaluates expressions even if logging is disabled


neha-bhargava:
Right now MSAL does string concatenation / string formatting and only then checks if the logs are enabled or not. This is a performance issue and we should be only creating logs when enabled.

How we do logging today:

```csharp
_logger.Info("Some log message that prints number of tokens " + accessTokenDictonary.Count);
```
This evaluates the Count even if logging is disabled.

Possible solution: 

- Modify the logging methods to use function pointers and lazily evaluate the strings

```csharp
class Program
    {
        static async Task Main(string[] args)
        {
            Logger logger = new Logger(1 /* INFO only */ );
            Dictionary d = new Dictionary();

            // bad - it evalues d.Count even though the log level is INFO
            logger.Debug("Count is: " + d.Count);

            // good - it does not evaluate d.Count 
            logger.Debug(() => "count is: " + d.Count);
        }
    }

    public class Logger
    {
        private int _logLevel;
        public Logger(int logLevel)
        {
            _logLevel = logLevel;
        }
        

        public void Info(Func<string> msgProducer)
        {
            if (_logLevel >= 1)
            {
                Console.WriteLine("INFO " + msgProducer());
            }
        }

        public void Debug(Func<string> msgProducer)
        {
            if (_logLevel >= 2)
            {
                Console.WriteLine("DEBUG " + msgProducer());
            }
        }

        public void Info(string msg)
        {
            if (_logLevel >= 1)
            {
                Console.WriteLine("INFO " + msg);
            }
        }

        public void Debug(string msg)
        {
            if (_logLevel >= 2)
            {
                Console.WriteLine("DEBUG " + msg);
            }
        }
    }

    public class Dictionary
    {
        public string Count
        {
            get
            {
                Console.WriteLine("Count evaluated (expensive!)");
                return "42";
            }
        }
    }
```

PR #2010, opened 2020-08-24
Improve logging performance around token cache

bgavrilMS:


src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -641,7 +657,8 @@ async Task<IEnumerable<IAccount>> ITokenCacheInternal.GetAccountsAsync(Authentic
                     requestParameters.HomeAccountId,
                     StringComparison.OrdinalIgnoreCase));
 
-                logger.Verbose($"Filtered by home account id. Remaning accounts {accounts.Count()}");
+                if (logger.IsLoggingEnabled(LogLevel.Verbose))
+                    logger.Verbose($"Filtered by home account id. Remaning accounts {accounts.Count()}");
```
jennyf19, line 661:
```suggestion
                    logger.Verbose($"Filtered by home account id. Remaining accounts {accounts.Count()} ");
```

src/client/Microsoft.Identity.Client/Utils/EnumerableExtensions.cs
```diff
@@ -37,11 +37,17 @@ internal static bool ContainsOrdinalIgnoreCase(this IEnumerable<string> set, str
...
+                logger.Verbose($"{logPrefix} - item count before: {list.Count()} ");
+            }
+
+            list = list.Where(predicate);
 
-            logger.Info($"{logPrefix} item count before {numberBefore} after {numberAfter}");
+            if (logger.IsLoggingEnabled(LogLevel.Verbose))
+            {
+                logger.Verbose($"{logPrefix} - item count after: {list.Count()} ");
+            }
```
bgavrilMS, line 50: This is the actual perf issue fix.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -264,19 +265,22 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
             // no match found after initial filtering
             if (!tokenCacheItems.Any())
             {
-                requestParams.RequestContext.Logger.Info("No matching entry found for user or assertion");
+                logger.Info("No matching entry found for user or assertion");
                 return null;
             }
 
-            requestParams.RequestContext.Logger.Info("Matching entry count - " + tokenCacheItems.Count());
+            if (logger.IsLoggingEnabled(LogLevel.Info))
```
trwalke, line 272: Is it possible for the logger to check for this internally?

bgavrilMS: Very good question. It's tricky and I wanted to provide a localized fix without changing "everything". We have a work item to look at this in more detail - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1981 which has a prototype for achieving what you are suggesting. 

Essentially in order to delay evaluation of the string to be logged, I was thinking to use a function pointer that generates the string. So instead of giving the logger a string, you give it a function that returns a string. If the log level is not high enough, the logger does not evaluate the function. 

But the problem with this solution is that it needs to be perf tested. Function pointers with closures are complicated and I am not sure that we won't create a different perf problem :). If you want to read more about closures, Jon Skeet has a great article: https://csharpindepth.com/Articles/Closures

PR #2022, opened 2020-08-31
Add the default constructor

neha-bhargava:


src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -25,6 +25,11 @@ public abstract class AbstractAcquireTokenParameterBuilder<T>
     {
         internal IServiceBundle ServiceBundle { get; }
 
+        /// <summary>
+        /// Default constructor for AbstractAcquireTokenParameterBuilder
+        /// </summary>
```
trwalke, line 30: should add [Obsolete("Please use the derived classes to construct this object.")]

trwalke: @jmprieur thoughts?

jennyf19: we don't need obsolete anymore.

---
In reply to: [480299373](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2022#discussion_r480299373) [](ancestors = 480299373)

neha-bhargava: Ohh we dont?

jennyf19: i don't think so. that causes the same issue. @jmprieur to make a final decision though.

jmprieur: In this case, we don't need it as there are other constructors.
the reason why we read this constructor is to keep the binary compatibility.

jmprieur (changes requested):
I propose to remove the obsolete attribute

neha-bhargava:
> I propose to remove the obsolete attribute

Removed.

jmprieur (approved):

PR #2052, opened 2020-09-23
Adding PKey Auth bypass to interactive auth flows

trwalke:
Testing workaround for PKey Auth bypass

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/EmbeddedWebview/AuthenticationAgentActivity.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/EmbeddedWebview/AuthenticationAgentActivity.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using Android.App;
 using Android.Content;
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/EmbeddedWebview/AuthenticationAgentActivity.cs
```diff
@@ -111,6 +113,24 @@ public override bool ShouldOverrideUrlLoading(WebView view, string url)
                     return true;
                 }
 
+                if (url.StartsWith(BrokerConstants.ClientTlsRedirect, StringComparison.OrdinalIgnoreCase))
```
henrik-me, line 116: >if (url.StartsWith(BrokerConstants.ClientTlsRedirect, StringComparison.OrdinalIgnoreCase)) [](start = 16, length = 90)

Before moving ahead with this I would like to get the CA team perspective on us doing this. Ideally the service should not send this. 
Also not sure if the https piece would be another part of the issue or not, also to be explored.

trwalke: Synced with CA team and they approved this fix.

Issue #1988, opened 2020-08-13
[Bug] Issue with MSAL 4.16.0 library when using Integrated Windows Authentication


hiteshdasikamw:
**Which Version of MSAL are you using ?**
MSAL 4.16.0

**Platform**
netcore

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [X] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**
c. This is a new app or experiment

**Repro**

```csharp
            app = PublicClientApplicationBuilder.Create(clientId).WithTenantId(tenant).WithAuthority(authority).Build();
            try
            {
                Task<AuthenticationResult> result = app.AcquireTokenByIntegratedWindowsAuth(scopes).WithUsername(username).ExecuteAsync();
                Console.Write(result.Result.AccessToken);
                Console.ReadLine();
            }
            catch (Exception exc)
            {
                Console.WriteLine(exc);
                MsalException x = exc.InnerException as MsalException;
                Console.WriteLine(x);
                MsalClientException ex = exc.InnerException as MsalClientException;
                Console.WriteLine(ex);
            }
```
The exception I get is shown below

**Expected behavior**
Expected to write access token onto the console.

**Actual behavior**
Below is the exception that occurs
```
System.AggregateException: One or more errors occurred. (Federated service at <Organization URL> returned error: ) ---> Microsoft.Identity.Client.MsalClientException: Federated service at <Organization URL> returned error:  ---> Microsoft.Identity.Client.MsalServiceException: Federated service at <Organization URL> returned error:
   at Microsoft.Identity.Client.WsTrust.WsTrustWebRequestManager.GetWsTrustResponseAsync(WsTrustEndpoint wsTrustEndpoint, String wsTrustRequest, RequestContext requestContext)
   at Microsoft.Identity.Client.WsTrust.CommonNonInteractiveHandler.GetWsTrustResponseAsync(UserAuthType userAuthType, String cloudAudienceUrn, WsTrustEndpoint endpoint, String username, SecureString securePassword)
   --- End of inner exception stack trace ---
[...]

PR #2064, opened 2020-09-30
Revert to a simple static HttpClient on .netcore #1988

bgavrilMS:


src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -203,13 +203,6 @@ public override Task StartDefaultOsBrowserAsync(string url)
 
         public override bool UseEmbeddedWebViewDefault => false;
 
-        public override IMsalHttpClientFactory CreateDefaultHttpClientFactory()
```
bgavrilMS, line 206: the base class defaults to SimpleHttpFactory

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -12,25 +12,26 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
...
     internal class SimpleHttpClientFactory : IMsalHttpClientFactory
     {
-        private static readonly Lazy<HttpClient> _httpClient = new Lazy<HttpClient>(() => InitializeClient());
+        private static readonly Lazy<HttpClient> s_httpClient = new Lazy<HttpClient>(() => InitializeClient());
 
         private static HttpClient InitializeClient()
         {
-            var httpClient = new HttpClient(new HttpClientHandler() { UseDefaultCredentials = true });
+            var httpClient = new HttpClient(new HttpClientHandler() { 
+                /* important for IWA */ UseDefaultCredentials = true });
```
bgavrilMS, line 26: This was missing in the NetCore client factory. Unfortunately it cannot be added because of a bug on .net core 2.1 (which was fixed on .net core 3.1 only). So we have to revert to the simple factory. This should be fine because on .NET core, our main customers are using Microsoft.Identity.Web which injects a performant httpclient factory that is part of asp.net core.

trwalke (approved):
Took a look and this makes sense to me. Thanks

Issue #2062, opened 2020-09-29
[Bug] Sometimes the status code returned in broker error messages is not parsed in the right format


trwalke:
**Which Version of MSAL are you using ?**
4.19

**Platform**
xamarin android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Repro**
**Expected behavior**
MSAL should properly parse http status codes broker error messages 

**Actual behavior**
MSAL will sometimes fail to parse these properly. 

**Possible Solution**
Refactor broker code to better parse error messages.

PR #2067, opened 2020-09-30
Resolving parsing issue with broker error

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2062

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
@@ -224,14 +224,8 @@ namespace Microsoft.Identity.Client.Platforms.Android.Broker
                         }
 
                         var httpResponse = new HttpResponse();
-                        httpResponse.Body = errorResult[BrokerResponseConst.BrokerHttpBody];
-                        httpResponse.Headers = errorResult[BrokerResponseConst.BrokerHttpHeaders];
-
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
```diff
@@ -225,7 +226,12 @@ internal static void SetBrokerResult(Intent data, int resultCode, ICoreLogger un
                         var httpResponse = new HttpResponse();
                         httpResponse.Body = errorResult[BrokerResponseConst.BrokerHttpBody];
                         httpResponse.Headers = errorResult[BrokerResponseConst.BrokerHttpHeaders];
-                        httpResponse.StatusCode = errorResult[BrokerResponseConst.BrokerHttpStatusCode];
+
+                        HttpStatusCode status;
+                        if (Enum.TryParse(errorResult[BrokerResponseConst.BrokerHttpStatusCode], out status))
```
bgavrilMS, line 231: Not sure what's happening here really. This is why "dynamic" should be avoided - no way of knowing what the objects really are. Maybe we should look again at why normal JSON parsing is failing in this case and avoid using dynamic.

Ultimately, I would only expect there to be strings when parsing json, not sure where there is an enum...

bgavrilMS (approved):

jennyf19 (approved):
:shipit:

Issue #1914, opened 2020-07-07
[Bug] Java.Lang exception from the broker


bgavrilMS:
**Which Version of MSAL are you using ?**
4.14

**Platform**
android + broker 

**Repro**

Sometimes, when calling the broker, the following exception is thrown: 

Java.Lang.Exception: Exception of type 'Java.Lang.Exception' was thrown.

JniEnvironment+InstanceMethods.CallObjectMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args)
JNIEnv.CallObjectMethod (System.IntPtr jobject, System.IntPtr jmethod, Android.Runtime.JValue* parms)
**IAccountManagerFutureInvoker.GetResult (System.Int64 timeout, Java.Util.Concurrent.TimeUnit unit)**
IAccountManagerFutureExtensions+<>c__DisplayClass0_0.<GetResultAsync>b__0 ()
Task`1[TResult].InnerInvoke ()
Task.Execute ()
AndroidBrokerHelper.GetBrokerAuthTokenSilentlyAsync (System.Collections.Generic.IDictionary`2[TKey,TValue] brokerPayload, Android.App.Activity callerActivity)
AndroidBroker.AcquireTokenInternalAsync (System.Collections.Generic.IDictionary`2[TKey,TValue] brokerPayload)
AndroidBroker.AcquireTokenUsingBrokerAsync (System.Collections.Generic.Dictionary`2[TKey,TValue] brokerPayload)
android.accounts.OperationCanceledException
android.accounts.AccountManager$AmsTask.internalGetResult AccountManager.java:2328
android.accounts.AccountManager$AmsTask.getResult AccountManager.java:2340
android.accounts.AccountManager$AmsTask.getResult AccountManager.java:2252

Remark: this is probably a Xamarin bug.

bgavrilMS:
Related Xamarin issue: https://github.com/xamarin/xamarin-android/issues/4898

bgavrilMS:
CC @trwalke

henrik-me:
planning: Travis to follow up with the more-info

trwalke:
What needs to be done here is the handling of the potential exceptions thrown when using GetAccountsAsync()

ForExample:
```CSharp
try {
    var r = result.GetAccountsAsync(…);
}
catch (Android.Accounts.OperationCanceledException e) {
    // …
}
```
see this for more details
[...]

PR #2068, opened 2020-09-30
WIP Handling java error in broker operations

trwalke:
potential fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1914

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -757,7 +757,7 @@ namespace Microsoft.Identity.Client
         public const string ExperimentalFeature = "experimental_feature";
 
         /// <summary>
-        /// <para>What happens?</para>A broker application is required, but one could not be found.
+        /// <para>What happens?</para>A broker application is required, but one could not be found or communicated with.
         /// <para>Mitigation</para>Install a broker application, such as Authenticator, from the application store
         /// </summary>
         public const string BrokerApplicationRequired = "broker_application_required";
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -763,6 +763,12 @@ namespace Microsoft.Identity.Client
         public const string BrokerApplicationRequired = "broker_application_required";
 
         /// <summary>
+        /// <para>What happens?</para>You are trying to authenticate with the broker but MSAL is unable to read the response from the broker.
+        /// <para>Mitigation</para>The currently installed broker may not support MSAL.Xamarin, you need to ensure that you have installed either Intune Company Portal (5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater). see https://aka.ms/Brokered-Authentication-for-Android"
+        /// </summary>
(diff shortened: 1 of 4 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -91,10 +92,19 @@ public async Task<Intent> GetIntentForInteractiveBrokerRequestAsync(BrokerReques
...
+                try
+                {
+                    bundleResult = (Bundle)await result.GetResultAsync(
+                         AccountManagerTimeoutSeconds,
+                         TimeUnit.Seconds)
+                         .ConfigureAwait(false);
+                }
+                catch (OperationCanceledException ex)
+                {
+                    _logger.Error("An error occured when trying to communicate with account manager: " + ex.Message);
```
bgavrilMS, line 106: Not sure what happens if you ignore this exception? Does the whole thing eventually fallback to using the browser?

I think we should repackage the exception to explain what is really happening...

`OperationCanceledException if the operation was canceled for any reason, including the user canceling the creation process or adding accounts (of this type) has been disabled by policy`

I bet that in 99% of the reported case, the actual issue comes from "disabled by policy".

I believe that for this issue to be resolved, we would have to properly support a different way of intercting with the broker, i.e. bounded service.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -762,6 +762,12 @@ public static class MsalError
         /// </summary>
         public const string BrokerApplicationRequired = "broker_application_required";
 
+        /// <summary>
+        /// <para>What happens?</para>You are trying to authenticate with the broker but MSAL is unable to read the response from the broker.
+        /// <para>Mitigation</para>The currently installed broker may not support MSAL.Xamarin, you need to ensure that you have installed either Intune Company Portal (5.0.4689.0 or greater) or Microsoft Authenticator (6.2001.0140 or greater). see https://aka.ms/Brokered-Authentication-for-Android"
+        /// </summary>
+        public const string FailedToGetBrokerResponse = "failed_to_get_broker_response";
```
jennyf19, line 769: nit: since this is just for android, maybe include "android" in the response string "failed_to_get_android_broker_response"

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -305,10 +332,23 @@ public async Task InitiateBrokerHandshakeAsync(Activity callerActivity)
...
+                        try
+                        {
+                            bundleResult = (Bundle)await result.GetResultAsync(
+                                AccountManagerTimeoutSeconds,
+                                TimeUnit.Seconds)
+                                .ConfigureAwait(false);
+                        }
+                        catch (OperationCanceledException ex)
+                        {
+                            _logger.Error("An error occured when trying to communicate with account manager: " + ex.Message);
```
jennyf19, line 346:
```suggestion
                            _logger.Error("An error occurred when trying to communicate with the account manager: " + ex.Message);
```

Issue #2013, opened 2020-08-26
[Feature Request] Implement POP (SHR) for confidential client bringing your own keys (#2)


trwalke:
What MSAL implements today: 

- [WithProofOfPossession](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/67325c7cabb9e4dbccc32c6764c865f088b8f3ad/src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs#L43) API
- this is available on .NET desktop for public client only
- there exists an internal interface `IPoPCryptoProvider` for key management
- key management is done by storing key on disk in a container (similar to DPAPI protection)

Tasks to complete: 

### First deliverable
1. Refactor WithProofOfPossession public API to be more like MSAL Android's implementation ([api](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/PoPAuthenticationScheme.java) and [example](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/testapps/testapp/src/main/java/com/microsoft/identity/client/testapp/MsalWrapper.java#L122-L125)) and take into consideration the fact that **Method** and **Url** are optional.

### Second deliverable

2. [Needs discussion] Refactor `IPoPCryptoProvider` interface to provide a higher level abstraction. The current `CannonicalPublicKeyJwk` is pretty low level. Or maybe it's ok as Wilson can plugin here. 
3. Make IPoPCryptoProvider class public and expose a new overload of `WithProofOfPossession` that takes in a `IPoPCryptoProvider` so that customers can define their own key management.

trwalke:
Continuation of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1946

jmprieur:
@trwalke do we need to do more ?

bgavrilMS:
@jmprieur - the second deliverable is around exposing an interface that enables users to bring their own POP keys. 
I'd say that a 3rd deliverable is to align our public API to the Android API, which has been approved in Api Reviews.

bgavrilMS:
Blocked on spec.

PR #2070, opened 2020-09-30
WIP POP improvements

trwalke:
for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2013

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
@@ -371,7 +371,6 @@ namespace Microsoft.Identity.Client
         /// <param name="httpRequestMessage">An HTTP request to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request.</param>
         /// <remarks>
         /// <list type="bullet">
-        /// <item>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</item>
         /// <item> An Authentication header is automatically added to the request</item>
         /// <item> The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters). </item>
         /// <item>MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</item>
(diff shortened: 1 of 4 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -406,7 +412,67 @@ internal T WithProofOfPosession(HttpRequestMessage httpRequestMessage, IPoPCrypt
...
+            return (T)this;
+        }
+
+        /// <summary>
+        /// Specifies the URL of the HTTP request to the protected resource which requires a PoP token. 
+        /// The PoP token will be cryptographically bound to the request.
+        /// </summary>
+        /// <param name="uri">Protected resource URL.</param>
+        /// <returns></returns>
+        public T WithProofOfPosessionUri(Uri uri)
```
bgavrilMS, line 454: I don't like this strategy of piling on options on the AcquireToken* method because it "polutes" . There will be at least 4 params on SHR -> 4 methods on each AcqurieToken* entry point. 

Instead, you can add a simple config object, similar to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/530d9f08b2c5f59edc2ad1dd82ca0729114e271e/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs#L125

So we could have a single method:

`WithPoP(PoPOptions options = null)`

bgavrilMS: The main question here is - what should the default be here? I would argue for a secure default, where POP takes into account the method (GET / POST) and the url ...

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -423,6 +489,25 @@ internal void ValidateAndCalculateApiId()
             CommonParameters.ApiId = CalculateApiEventId();
             CommonParameters.ApiTelemId = ApiTelemetryId;
             CommonParameters.CorrelationId = CommonParameters.UseCorrelationIdFromUser ? CommonParameters.UserProvidedCorrelationId : Guid.NewGuid();
+
+#if DESKTOP || NET_CORE
+            if (CommonParameters.UsingProofOfPossesion)
+            {
+                if (CommonParameters.PopUri == null )
+                {
+                    throw new MsalClientException(MsalError.PopUriCannotBeNull, "Proof of possesion endpoint is null.");
```
bgavrilMS, line 498: Uri is not mandatory as far as I understand. Please check with Brian / George.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -423,6 +489,25 @@ internal void ValidateAndCalculateApiId()
...
+                }
+
+                HttpRequestMessage message = new HttpRequestMessage(CommonParameters.PopMethod != null ? CommonParameters.PopMethod : HttpMethod.Get,
```
bgavrilMS, line 501: No, this isn't how it's supposed to work. The user has the ability to say "don't bind the method to the pop token".

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -406,7 +412,67 @@ internal T WithProofOfPosession(HttpRequestMessage httpRequestMessage, IPoPCrypt
...
+        }
+
+        /// <summary>
+        /// Specifies a provider that can handle the asymmetric key operations needed by POP, that encapsulates a pair of 
+        /// public and private keys and some typical crypto operations.
+        /// All symetric operations are SHA256
+        /// </summary>
+        /// <param name="popCryptoProvider"> Proof of posession cryptography provider</param>
+        /// <returns></returns>
+        public T WithPopCryptoProvider(IPoPCryptoProvider popCryptoProvider)
```
bgavrilMS, line 467: I think you are trying to tackle 2 stories at the same time, which makes it difficult to implement and review: 

1. Expose the crypto details to the user
2. Make the method and uri optional

I don't think you should tackle #2 right now, nobody asked for it and it will complicate things.

bgavrilMS, line 467: I think `IPoPCryptoProvider` is too low level and a higher level abstraction should be exposed to the user.

bgavrilMS (changes requested):
"bring your own encryption" public API needs more work.

Issue #1872, opened 2020-06-09
Facebook IDP throwing state mismatch error (External Identities) 


pottabathini:
**Which Version of MSAL are you using ?**
Downloaded MSAL Desktop app from QuickStarts inside Portal AAD Applications blade.
(Quickstart: Acquire a token and call Microsoft Graph API from a Windows desktop app)

**Platform**
net45

**What authentication flow has the issue?**
* Desktop
    * [Yes ] Interactive

**Is this a new or existing app?**
New App. Created for testing

**Repro**
  * Follow below doc to enable self-service signup and add Facebook IDP
  * https://docs.microsoft.com/en-in/azure/active-directory/b2b/self-service-sign-up-user-flow
  * Download the MSAL desktop app from quick start (Inside AAD -> Application) and start running the application.
 * Choose Facebook IDP and complete the Signup/ Signin process

**Fiddler Trances**
[StateMismatch FB Trace.zip](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/files/4751354/StateMismatch.FB.Trace.zip)

**Expected behavior**
MSAL.Net app should allow Facebook users to login. We tried with less number of characters in email address and it is working fine, if we have more characters in email address (ex: thisismyemailaddressformsal@gmail.com), MSAL.Net start throwing state mismatch exception.

**Actual behavior**
If email have more characters, MSAL throwing state mismatch exception. we tried with less number of characters it is working fine.

**Possible Solution**

**Additional context/ Logs / Screenshots**
![image](https://user-images.githubusercontent.com/11867309/84135360-818bd600-aa67-11ea-9a14-76d02f61b32f.png)

neha-bhargava:
I was able to repro the scenario. For some reason when the username is long there is a mismatch in the original state and state returned in the auth result. Which is validated [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/271cfbb35667525cd18287e0bf887206dea0ab44/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeRequestComponent.cs#L209). I debugged and found that the state returned in the auth result is truncated when the username is long.

neha-bhargava:
[...]

PR #2080, opened 2020-10-06
Add link for details for state mismatch exception

neha-bhargava:
#1872

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
+++ b/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
@@ -97,6 +97,8 @@ namespace Microsoft.Identity.Client
             +
             "2. An error occurred during a silent web authentication that prevented the http authentication flow from completing in a short enough time frame";
 
+        public const string StateMismatchErrorMessage = "Returned state({0}) from authorize endpoint is not the same as the one sent({1}). See https://aka.ms/msal-statemismatcherror for more details.";
+
         public const string UserRealmDiscoveryFailed = "User realm discovery failed";
 
         public const string WsTrustEndpointNotFoundInMetadataDocument =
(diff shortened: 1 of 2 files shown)


jennyf19 (approved):

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -97,6 +97,8 @@ public static string iOSBrokerKeySaveFailed(string keyChainResult)
             +
             "2. An error occurred during a silent web authentication that prevented the http authentication flow from completing in a short enough time frame";
 
+        public const string StateMismatchErrorMessage = "Returned state({0}) from authorize endpoint is not the same as the one sent({1}). See https://aka.ms/msal-statemismatcherror for more details.";
```
jmprieur, line 100: Is it good from a security point of view to display the state that was sent?

jmprieur (commented):
LGTM.
Thanks @neha-bhargava

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
neha-bhargava: I don't think it would ,matter as it has already failed. @bgavrilMS Suggestions?

jennyf19: that's a good point, plus what would you do w/those values? and they are pretty long. i think just having the error message w/the link is enough.

neha-bhargava: @jennyf19 Those were already there in the message. I just added the aka.ms link and moved the message to this class. I don't know what they were added for.

henrik-me:
Looks great! Provided a comment in the issue about where the aka.ms link and the information should be placed

---
In reply to: [503660297](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2080#issuecomment-503660297) [](ancestors = 503660297)

henrik-me (approved):
:shipit:

neha-bhargava:
> Looks great! Provided a comment in the issue about where the aka.ms link and the information should be placed
> 
> In reply to: [503660297](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2080#issuecomment-503660297) [](ancestors = 503660297)

Updated the Link.

Issue #2090, opened 2020-10-12
[Bug][Regression] Get NoTokensFoundError when migrating ADAL v3 token for user who has multi tenants


erich-wang:
**Which Version of MSAL are you using ?**
Note that to get help, you need to run the latest version. Preview version are also ok. 
For ADAL, please log issues to https://github.com/AzureAD/azure-activedirectory-library-for-dotnet
MSAL 4.19.0, the issue doesn't exist in MSAL 4.10.0

**Platform**
netcore/.Net 4.5

**What authentication flow has the issue?**
* Desktop / Mobile
For users associated with multi tenants, there're multiple RT in ADAL token cache because there is one corresponding RT for each tenant. When migrating ADAL v3 token, [the change](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/bafa2f06a9d4459dbb224052237520ba767e765b#diff-673aecdcc2a1fcef67de94c00ac55600R320) results in NoTokensFoundError

**Repro**

```csharp
            var builder = PublicClientApplicationBuilder.Create("1950a258-227b-4e31-a9cf-717495945fc2");
            var clientApplication = builder.Build();
            clientApplication.UserTokenCache.SetBeforeAccess((TokenCacheNotificationArgs args) =>
            {
                  args.TokenCache.DeserializeAdalV3(AdalToken)
            }
            var accounts = clientApplication.GetAccountsAsync().ConfigureAwait(false).GetAwaiter().GetResult();
            foreach (var account in accounts)
            {
                 var token = clientApplication.AcquireTokenSilent(new []{"https://management.core.windows.net/.default"}, account).ExecuteAsync().ConfigureAwait(false).GetAwaiter().GetResult();
                //Get NoTokensFoundError
           }

```

**Expected behavior**
Token should be returned successfully

**Actual behavior**
Get NoTokensFoundError

**Possible Solution**

**Additional context/ Logs / Screenshots**
Add any other context about the problem here, such as logs and screebshots. Logging is described at https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging

jmprieur:
@bgavrilMS : did it go in 4.20 ?

PR #2091, opened 2020-10-12
Fix for #2090 - allow multiple RTs per user from ADAL cache

bgavrilMS:


src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -231,105 +231,79 @@ internal static class CacheFallbackOperations
...
-
-                    list.Add(new MsalRefreshTokenCacheItem(
-                      new Uri(pair.Key.Authority).Host,
-                      pair.Key.ClientId,
-                      pair.Value.RefreshToken,
-                      pair.Value.RawClientInfo,
-                      familyId: null,
-                      homeAccountId: homeAccountId));
+                    logger.Warning("Could not filter ADAL entries by either UPN or unique ID, skipping.");
+                    return null;
```
bgavrilMS, line 277: First part of fix - if no filter is presented to ADAL legacy cache (upn or object ID), do not return an RT

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -231,105 +231,79 @@ internal static class CacheFallbackOperations
...
 
-                return list;
+                return listToProcess.Select(adalEntry => new MsalRefreshTokenCacheItem(
+                      new Uri(adalEntry.Key.Authority).Host,
+                      adalEntry.Key.ClientId,
+                      adalEntry.Value.RefreshToken,
+                      adalEntry.Value.RawClientInfo,
+                      familyId: null,
+                      homeAccountId: GetHomeAccountId(adalEntry.Value)))
+                .FirstOrDefault();
```
bgavrilMS, line 287: Second part of fix: it is acceptable to find multiple RTs for the same user in the ADAL cache, because ADAL stores all tokens (AT, RT, IdT) on a per-tenant basis, even though RTs are tenantless.

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -231,105 +231,79 @@ internal static class CacheFallbackOperations
...
+                    listToProcess =
+                        listToProcess.Where(p => account.Username.Equals(
+                            p.Key.DisplayableId, StringComparison.OrdinalIgnoreCase));                        
+                    
+                    filtered = true;
                 }
 
                 // if uniqueId is provided then use it to filter
-                if (!string.IsNullOrEmpty(uniqueId))
+                if (!string.IsNullOrEmpty(account?.HomeAccountId?.ObjectId))
```
jennyf19, line 264: nit: same as above, modify comment to reflect object Id filter

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
jennyf19: now you're filtering by username and objectid right? 

---
In reply to: [503286041](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2091#discussion_r503286041) [](ancestors = 503286041)

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
bgavrilMS: Yes, this is what we were doing before as well. Although it is very unlikely you can pass in an Account without username and without ID, theoretically you can do that, so I stopped this scenario. 

The actual fix is more about acknowledging the fact that ADAL saves one RT for each tenant (even though RTs are tenant-less), which was confusing. Sometime ago we started treating that as an error. It is not.

src/client/Microsoft.Identity.Client/Cache/CacheFallbackOperations.cs
```diff
@@ -231,105 +231,79 @@ internal static class CacheFallbackOperations
...
                     dictionary.Where(p =>
                         p.Key.ClientId.Equals(clientId, StringComparison.OrdinalIgnoreCase) &&
-                        environmentAliases.Contains(new Uri(p.Key.Authority).Host)).ToList();
+                        environmentAliases.Contains(new Uri(p.Key.Authority).Host));
+
+                bool filtered = false;
 
                 // if upn is provided then use it to filter
-                if (!string.IsNullOrEmpty(upn))
+                if (!string.IsNullOrEmpty(account?.Username))
```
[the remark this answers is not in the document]
bgavrilMS: Agreed, those comments are superflous.

PR #2094, opened 2020-10-13
Disable WAM integration

bgavrilMS:
This is the easiest way to disable WAM. I had to comment a bit of code, but this should not impact productivity when looking at WAM again.


src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -201,10 +201,12 @@ public override IPoPCryptoProvider GetDefaultPoPCryptoProvider()
         {
             return PoPProviderFactory.GetOrCreateProvider();
         }
-        public override IBroker CreateBroker(CoreUIParent uiParent)
```
bgavrilMS, line 204: There is a base class that provides an implementation for this, which simply throws a not supported exceptions. Customers do not see this exception, because they cannot enable the broker on .NET, as we do not expose `.WithBroker`

jennyf19 (approved):
:shipit:

PR #2098, opened 2020-10-14
Add a default null value to the parameter authenticationResultMetadata

neha-bhargava:
Fix for the breaking change in 4.17.0


neha-bhargava:
There is another PR for this

PR #2105, opened 2020-10-15
Pop updated

trwalke:
Updates for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2013


src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -71,19 +73,19 @@ public string FormatAccessToken(MsalAccessTokenCacheItem msalAccessTokenCacheIte
...
                 new JProperty(PoPClaimTypes.Cnf, new JObject(
                     new JProperty(PoPClaimTypes.JWK, popAssertion))));
 
             string popToken =  CreateJWS(payload.ToString(Json.Formatting.None), header.ToString(Json.Formatting.None));
 
             // For POP, we can also update the HttpRequest with the authentication header
-            _httpRequestMessage.Headers.Authorization = new AuthenticationHeaderValue(
-                AuthorizationHeaderPrefix,
-                popToken);
+            //_httpRequestMessage.Headers.Authorization = new AuthenticationHeaderValue(
```
trwalke, line 86: I dont see where this object is being used after this is set. How are these headers used in the pop binding?

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -404,9 +411,36 @@ internal T WithProofOfPosession(HttpRequestMessage httpRequestMessage, IPoPCrypt
...
+        /// </summary>
+        /// <param name="popAuthenticationConfiguration"> </param>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item> An Authentication header is automatically added to the request</item>
+        /// <item> The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters). </item>
+        /// <item>MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</item>
+        /// </list>
+        /// </remarks>
+        public T WithProofOfPosession(PopAuthenticationConfiguration popAuthenticationConfiguration)
```
bgavrilMS, line 434: IMO this should be the only POP method. Let's delete the old one.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// Configuration properties used to construct a Proof of possesion request
+    /// </summary>
+    public class PopAuthenticationConfiguration
+    {
+        /// <summary>
+        /// An HTTP uri to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request.
```
bgavrilMS, line 16: "The Uri will be cryptographically bound to the token". I guess, bounding is 2 way :)

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -368,45 +369,24 @@ public T WithCorrelationId(Guid correlationId)
...
             {
-                throw new ArgumentNullException(nameof(popCryptoProvider));
+                throw new ArgumentNullException(nameof(popAuthenticationConfiguration));
             }
 
-            CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithPoPScheme);
-            CommonParameters.AuthenticationScheme = new PoPAuthenticationScheme(httpRequestMessage, popCryptoProvider);
-
-            return this as T;
+            CommonParameters.UsingProofOfPossesion = true;
```
bgavrilMS, line 387: Is this really needed? If we have a POP config, we must be doing POP.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -368,45 +369,24 @@ public T WithCorrelationId(Guid correlationId)
...
-                    MsalErrorMessage.ExperimentalFeature(nameof(WithProofOfPosession)));
-            }
-
-            if (httpRequestMessage is null)
-            {
-                throw new ArgumentNullException(nameof(httpRequestMessage));
-            }
-
-            if (popCryptoProvider == null)
+            if (popAuthenticationConfiguration is null)
```
bgavrilMS, line 382: nit; 

A cleaner way of writing this: 

```csharp
CommonParameters.PopAuthenticationConfiguration = popAuthenticationConfiguration ?? throw new ArgumentNullException(...));
```

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -423,6 +403,19 @@ internal void ValidateAndCalculateApiId()
             CommonParameters.ApiId = CalculateApiEventId();
             CommonParameters.ApiTelemId = ApiTelemetryId;
             CommonParameters.CorrelationId = CommonParameters.UseCorrelationIdFromUser ? CommonParameters.UserProvidedCorrelationId : Guid.NewGuid();
+
+#if DESKTOP || NET_CORE
```
bgavrilMS, line 407: Let's not block anything on platform. Trust in .NET

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
bgavrilMS: In fact, why don't you just create a new `PoPAuthenticationScheme` here directly?

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -423,6 +403,19 @@ internal void ValidateAndCalculateApiId()
...
+            if (CommonParameters.UsingProofOfPossesion && CommonParameters.PopAuthenticationConfiguration != null)
+            {
+                if (CommonParameters.PopAuthenticationConfiguration.PopCryptoProvider == null)
```
bgavrilMS, line 410: It isn't the job of Validate() function or of this object to set defaults. This object should not have to understand anything related to POP.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
```diff
@@ -23,6 +26,8 @@ internal class AcquireTokenCommonParameters
         public string Claims { get; set; }
         public AuthorityInfo AuthorityOverride { get; set; }
         public ApiTelemetryId ApiTelemId { get; set; } = ApiTelemetryId.Unknown;
+        public bool UsingProofOfPossesion { get; set; }
```
bgavrilMS, line 29: As discussed above, try to move everything in the auth scheme.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// </summary>
+        public Uri RequestUri { get; set; }
```
bgavrilMS, line 19: I'd remove the `set` here since the URI is required, as denoted by your ctor.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// The HttP method to be bound to the request
```
bgavrilMS, line 22: The HTTP method ("GET", "POST" etc.) method that will be bound to the token. Leave null and the POP token will not be bound to the method.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// </summary>
+        public HttpMethod PopHttpMethod { get; set; }
+
+        /// <summary>
+        /// A provider that can handle the asymmetric key operations needed by POP, that encapsulates a pair of public and private keys
+        /// and some typical crypto operations
+        /// </summary>
+        public IPoPCryptoProvider PopCryptoProvider { get; set; }
```
bgavrilMS, line 30: An extensibility point that allows developers to define their own key management. Leave null and MSAL will use a default implementation, which generates an RSA key pair in memory and refreshes it every 8 hours.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// This will be populated with the POP authentication header.
+        /// </summary>
+        public AuthenticationHeaderValue PopAuthenticationRequestHeader {get; private set;}
```
bgavrilMS, line 35: I'm not sure what this is. If you're not sure either, delete it :)

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            if (requestUri == null || string.IsNullOrEmpty(requestUri.AbsoluteUri))
+            {
+                throw new ArgumentNullException(nameof(requestUri));
+            }
+
+            RequestUri = requestUri;
+        }
+
+        internal void SetPopHttpRequestHeaders(AuthenticationHeaderValue header)
```
bgavrilMS, line 51: Not needed.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -423,6 +403,19 @@ internal void ValidateAndCalculateApiId()
...
+                {
+                    CommonParameters.PopAuthenticationConfiguration.PopCryptoProvider = ServiceBundle.PlatformProxy.GetDefaultPoPCryptoProvider();
```
bgavrilMS, line 412: Hmm, this object (the config) belongs to the user, so it would be better if MSAL didn't modify it. I'd keep the pop crypto provider object as a member of the auth scheme.

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -30,12 +29,20 @@ internal class PoPAuthenticationScheme : IAuthenticationScheme
         /// Currently the signing credential algorithm is hard-coded to RSA with SHA256. Extensibility should be done
         /// by integrating Wilson's SigningCredentials
         /// </remarks>
-        public PoPAuthenticationScheme(HttpRequestMessage httpRequestMessage, IPoPCryptoProvider popCryptoProvider)
+        public PoPAuthenticationScheme(PopAuthenticationConfiguration popAuthenticationConfiguration)
         {
-            _httpRequestMessage = httpRequestMessage ?? throw new ArgumentNullException(nameof(httpRequestMessage));
-            _popCryptoProvider = popCryptoProvider ?? throw new ArgumentNullException(nameof(popCryptoProvider));
+            if (popAuthenticationConfiguration == null)
```
bgavrilMS, line 34: Use ?? it's cleaner

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        /// <summary>
+        /// An HTTP uri to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request.
+        /// </summary>
+        public Uri RequestUri { get; set; }
+
+        /// <summary>
+        /// The HttP method to be bound to the request
+        /// </summary>
+        public HttpMethod PopHttpMethod { get; set; }
```
bgavrilMS, line 24: `HttpMethod`

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -66,24 +73,24 @@ public string FormatAccessToken(MsalAccessTokenCacheItem msalAccessTokenCacheIte
...
-            JToken popAssertion = JToken.Parse(_popCryptoProvider.CannonicalPublicKeyJwk);
+            JToken popAssertion = JToken.Parse(_popAuthenticationConfiguration.PopCryptoProvider.CannonicalPublicKeyJwk);
 
             var payload = new JObject(
                 new JProperty(PoPClaimTypes.At, msalAccessTokenCacheItem.Secret),
                 new JProperty(PoPClaimTypes.Ts, (long)(DateTime.UtcNow - s_jwtBaselineTime).TotalSeconds ),
-                new JProperty(PoPClaimTypes.HttpMethod, _httpRequestMessage.Method.ToString()),
-                new JProperty(PoPClaimTypes.Host, _httpRequestMessage.RequestUri.Host),
-                new JProperty(PoPClaimTypes.Path, _httpRequestMessage.RequestUri.AbsolutePath),
+                new JProperty(PoPClaimTypes.HttpMethod, _popAuthenticationConfiguration.PopHttpMethod.ToString()),
```
bgavrilMS, line 81: null ref here, since HttpMethod is optional.
if HttpMethod is null, token should not be bound to it.

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -66,24 +73,24 @@ public string FormatAccessToken(MsalAccessTokenCacheItem msalAccessTokenCacheIte
...
+                new JProperty(PoPClaimTypes.Host, _popAuthenticationConfiguration.RequestUri.Host),
```
bgavrilMS, line 82: How does Android do this? Do they expose a separate config for Host and Path or simply Uri like here?

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -822,6 +822,11 @@ public static class MsalError
         /// </summary>
         public const string CryptoNet45 = "crypto_net45";
 
+        /// <summary>
+        /// Proof of posession authentication attempt is missing a an endpoint to bing to.
+        /// </summary>
+        public const string PopUriCannotBeNull = "Pop_Uri_Cannot_Be_Null";
```
bgavrilMS, line 828: The ArgumentNullException in the config should be enough.

tests/Microsoft.Identity.Test.Integration.net47/Microsoft.Identity.Test.Integration.net47.csproj
```diff
@@ -0,0 +1,76 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
bgavrilMS, line 1: Instead of creating a new project, please update the existing one (Integration.net45) to target .NET 4.7.2.
It already targets .NET 4.6.2, so we're not missing anything.

tests/Microsoft.Identity.Test.Integration.net47/PoPCryptoProviderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            var confidentialClientAuthority = PublicCloudTestAuthority;
+
+            var confidentialApp = ConfidentialClientApplicationBuilder
+                .Create(PublicCloudConfidentialClientID)
+                .WithAuthority(new Uri(confidentialClientAuthority), true)
+                .WithClientSecret(s_publicCloudCcaSecret)
+                .Build();
+
+            //RSA provider
```
bgavrilMS, line 81: Please separate this into 2 tests

tests/Microsoft.Identity.Test.Integration.net47/PoPCryptoProviderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+            // the method the POP token in bound to
+            request.Headers.Add("ShrMethod", popConfig.PopHttpMethod.ToString());
+
+            response = await httpClient.SendAsync(request).ConfigureAwait(false);
+
+            Assert.IsTrue(response.IsSuccessStatusCode);
+        }
+
+
+        private static Dictionary<string, string> GetTestSliceParams()
```
bgavrilMS, line 139: Probably not needed.

tests/Microsoft.Identity.Test.Integration.net47/PoPCryptoProviderTests.cs
```diff
@@ -0,0 +1,179 @@
...
+                {
+                    byte[] bytes = notificationArgs.TokenCache.SerializeMsalV3();
+                    _inMemoryCache = Encoding.UTF8.GetString(bytes);
+                }
+            });
+        }
+
+        private static X509Certificate2 GetCertificate()
+        {
+            X509Certificate2 cert = CertificateHelper.FindCertificateByThumbprint(TestConstants.AutomationTestThumbprint);
```
bgavrilMS, line 169: This did not work on my machine :(

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -30,12 +29,20 @@ internal class PoPAuthenticationScheme : IAuthenticationScheme
...
+            }
+            if (popAuthenticationConfiguration.PopCryptoProvider == null)
+            { 
+                throw new ArgumentNullException(nameof(popAuthenticationConfiguration.PopCryptoProvider));
+            }
+
+            _popAuthenticationConfiguration = popAuthenticationConfiguration;
 
-            var keyThumbprint = ComputeRsaThumbprint(_popCryptoProvider.CannonicalPublicKeyJwk);
+            var keyThumbprint = ComputeRsaThumbprint(popAuthenticationConfiguration.PopCryptoProvider.CannonicalPublicKeyJwk);
```
bgavrilMS, line 45: `ComputeRsaThumbprint` doesn't actually perform any RSA operations, it just does a hash. Please rename (I know it was named by me)

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
bgavrilMS, line 30: I wonder if we should actually put this crypto provider at the Application level instead of the Request level ? This way we can cache things like Thumbprint. 

I suggest that for now we go with this and we discuss this in the API Review.

bgavrilMS (changes requested):
Generally looks good, see inline comments.

tests/Microsoft.Identity.Test.Integration.net47/PoPCryptoProviderTests.cs
trwalke: I need to update the  default JWT algorithm to use "ES256" instead for the elliptic curve. 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/8595a3eabcc2f2f2f53f9d6f8215f64686b9543d/src/client/Microsoft.Identity.Client/Internal/JsonWebTokenConstants.cs#L13

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -368,45 +369,24 @@ public T WithCorrelationId(Guid correlationId)
         ///  PoP tokens are similar to Bearer tokens, but are bound to the HTTP request and to a cryptographic key, which MSAL can manage on Windows.
         ///  See https://aka.ms/msal-net-pop
         /// </summary>
-        /// <param name="httpRequestMessage">An HTTP request to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request.</param>
+        /// <param name="popAuthenticationConfiguration"> </param>
```
jmprieur, line 372: Don't we have a description of the parameter any longer?

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
jmprieur: that's a breaking change?
would we do that because it's an experimental feature?

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// </summary>
+        public AuthenticationHeaderValue PopAuthenticationRequestHeader {get; private set;}
+
+        /// <summary>
+        /// Constructs the configuration properties used to construct a proof of possesion request
+        /// </summary>
+        /// <param name="requestUri"></param>
+        public PopAuthenticationConfiguration(Uri requestUri)
+        {
+            if (requestUri == null || string.IsNullOrEmpty(requestUri.AbsoluteUri))
```
jennyf19, line 43: >requestUri.AbsoluteUri [](start = 59, length = 22)

why is this check needed? i don't see it used i this method...maybe remove?

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -19,9 +20,7 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
     internal class PoPAuthenticationScheme : IAuthenticationScheme
```
jennyf19, line 20: nit: some places it' has PoP and some have Pop. maybe go with Pop?

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -822,6 +822,11 @@ public static class MsalError
         /// </summary>
         public const string CryptoNet45 = "crypto_net45";
 
+        /// <summary>
+        /// Proof of posession authentication attempt is missing a an endpoint to bing to.
```
jmprieur, line 826:
```suggestion
        /// Proof of posession authentication attempt is missing an endpoint to bind to.
```

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
trwalke: We used to populate the user provided HTTP Request headers with the authentication headers calculated form the key. Im assuming this is no longer needed?

This value was already being used for tests.

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
trwalke: Uri like here

src/client/Microsoft.Identity.Client/Internal/JsonWebTokenConstants.cs
```diff
@@ -11,6 +11,7 @@ internal class JsonWebTokenConstants
         internal class Algorithms
         {
             public const string RsaSha256 = "RS256";
+            public const string EdcSha256 = "ES256";
```
trwalke, line 14: This is what I was referring to. Elliptic curve can be ES256/ES384/ES512 etc. This value will need to be updated based on the conical JWK created by the elliptic curve keys.

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -71,23 +73,36 @@ public string FormatAccessToken(MsalAccessTokenCacheItem msalAccessTokenCacheIte
...
         }
 
+        private string GetHeaderAlgorithm()
+        {
+            var JWK = JObject.Parse(_popAuthenticationConfiguration.PopCryptoProvider.CannonicalPublicKeyJwk);
+
+            //Need to update header to signal EC key is being used
+            if (JWK[JsonWebKeyParameterNames.Kty].ToString() == JsonWebKeyParameterNames.EC)
+            {
+                return JsonWebTokenConstants.Algorithms.EdcSha256;
```
trwalke, line 100: Need to add support for other elliptic curve key sizes here. currently only 256 is supported. 
@bgavrilMS

trwalke: supported values
https://docs.microsoft.com/en-us/azure/key-vault/keys/about-keys#signverify

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <summary>
+        ///An extensibility point that allows developers to define their own key management. 
+        ///Leave null and MSAL will use a default implementation, which generates an RSA key pair in memory and refreshes it every 8 hours.
+        ///Important note: developers wanting to change the key (e.g. rotate the key) should create a new instance of this object,
+        ///as MSAL will keep a thumbprint of keys in memory
+        /// </summary>
+        public IPoPCryptoProvider PopCryptoProvider { get; set; }
+
+        /// <summary>
+        /// This will be populated with the POP authentication header.
```
jmprieur, line 35: populated? you mean MSAL.NET will populate this property?

src/client/Microsoft.Identity.Client/AuthScheme/PoP/IPoPCryptoProvider.cs
```diff
@@ -19,13 +21,20 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
...
+    public interface IPoPCryptoProvider
     {
         /// <summary>
         /// The cannonical representation of the JWK.  See https://tools.ietf.org/html/rfc7638#section-3
         /// </summary>
         string CannonicalPublicKeyJwk { get; }
 
+        /// <summary>
+        /// Algorithm used to sign proof of possesion request. 
+        /// See https://docs.microsoft.com/en-us/azure/key-vault/keys/about-keys#signverify for ECD
```
jmprieur, line 33: Aweseome that you added the links to the algorithms, @trwalke ❤️

src/client/Microsoft.Identity.Client/AuthScheme/PoP/IPoPCryptoProvider.cs
```diff
@@ -19,13 +21,20 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
...
+        /// See https://docs.microsoft.com/en-us/azure/key-vault/keys/about-keys#signverify-1 for RSA
+        /// </summary>
+        string CryptographicAlgorithm { get; }
```
jmprieur, line 36: Are symmetric key algorithm supported? if not, do we want to mention it?

src/client/Microsoft.Identity.Client/Internal/JsonWebTokenConstants.cs
jmprieur: How does this relate to the list of algorithms in https://docs.microsoft.com/en-us/azure/key-vault/keys/about-keys#signverify and https://docs.microsoft.com/en-us/azure/key-vault/keys/about-keys#signverify1 ?

tests/Microsoft.Identity.Test.Integration.Win8/App.config
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
     <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2"/>
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.2"/>
```
jmprieur, line 4: Is it required because of crypto?
If yes:
- Do we need to document that this is a requirement for .NET FW apps to be updated to 4.7.2 ?
- What happens if they are not?

jmprieur (commented):
I've left a few comments and suggestions

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
bgavrilMS: Yes, experimental features can introduce breaking changes at any time.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
bgavrilMS: This is public API, developer can pass null and we need to protect against that. RequestUri is mandatory.

tests/Microsoft.Identity.Test.Integration.Win8/App.config
bgavrilMS: This is required to use Eliptic Curve crypto, which has decent support only in 4.7.2 

The way Travis is designing this interface is that he created 2 reference implementations of "bring your own key" to make sure he covers all scenarios: 

- bring your own RSA key via an X509Cert
- bring your own ECD key (this requires .net 4.7.2)

And the default strategy for when developers do not customize: 

- in-memory RSA key with 8h rotation

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
bgavrilMS: It doesn't really make that much sense since this is a config option, but you cannot set this property. 

The way consumers are supposed to retrieve the pop token is via `AuthenticationResult`. We can provide more covenient mechasism upon demand.

bgavrilMS:
```csharp
 var r2 = await confidentialApp.AcquireTokenForClient(s_keyvaultScope)
              .WithProofOfPosession(popConfig)
              .ConfigureAwait(false);
```

image

bgavrilMS (changes requested):
Please remove `popConfig.PopAuthenticationRequestHeader` and rely on AuthenticationResult instead. Otherwise LGTM!

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
trwalke: yes

PR #2112, opened 2020-10-16
Enable WAM on UWP and minor enhancements

bgavrilMS:


jennyf19 (approved):
most of this i already looked at before.

PR #2113, opened 2020-10-16
[WAM] AcquireTokenSilent(account=null) logs in the default Windows user

bgavrilMS:


src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
```diff
@@ -21,6 +21,10 @@ internal interface IBroker
             AuthenticationRequestParameters authenticationRequestParameters,
             AcquireTokenSilentParameters acquireTokenSilentParameters);
 
+        Task<MsalTokenResponse> AcquireTokenSilentDefaultUserAsync(
```
bgavrilMS, line 24: New method on the IBroker interface to make it explicit

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -28,49 +32,68 @@ internal class SilentRequest : RequestBase
...
+              AuthenticationRequestParameters.IsBrokerConfigured &&
+              ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
+
             try
             {
-                _logger.Info("Attempting to acquire token using using local cache...");
-                return await _clientStrategy.ExecuteAsync(cancellationToken).ConfigureAwait(false);
+                if (AuthenticationRequestParameters.Account != null)
+                {
+                    _logger.Verbose("Attempting to acquire token using using local cache...");
```
jennyf19, line 56:
```suggestion
                    _logger.Verbose("Attempting to acquire token using local cache...");
```

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -28,49 +32,68 @@ internal class SilentRequest : RequestBase
...
-                return await _clientStrategy.ExecuteAsync(cancellationToken).ConfigureAwait(false);
+                if (AuthenticationRequestParameters.Account != null)
+                {
+                    _logger.Verbose("Attempting to acquire token using local cache...");
+                    return await _clientStrategy.ExecuteAsync(cancellationToken).ConfigureAwait(false);
+                }
+
+                if (!isBrokerConfigured)
+                {
+                    _logger.Verbose("No account passed to AcquireTokenSilent and no broker configured.  ");
```
jennyf19, line 62: >no broker configured [](start = 81, length = 20)

idk about adding broker messaging here. how many customers are using a broker at this point? I think that makes the actually error more confusing.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -89,5 +112,62 @@ internal async Task<AuthenticationResult> ExecuteTestAsync(CancellationToken can
...
+        private async Task<IAccount> GetSingleAccountForLoginHintAsync(string loginHint)
+        {
+            if (!string.IsNullOrEmpty(loginHint))
+            {
+                IReadOnlyList<IAccount> accounts = (await CacheManager.GetAccountsAsync().ConfigureAwait(false))
+                    .Where(a => !string.IsNullOrWhiteSpace(a.Username) &&
+                           a.Username.Equals(loginHint, StringComparison.OrdinalIgnoreCase))
+                    .ToList();
+
+                if (accounts.Count() == 0)
```
jennyf19, line 139: >accounts [](start = 20, length = 8)

could accounts be null?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
```diff
@@ -312,5 +311,17 @@ private void HandleBrokerOperationError(Exception ex)
             else
                 throw new MsalClientException(MsalError.AndroidBrokerOperationFailed, ex.Message, ex);
         }
+
+        /// <summary>
+        /// Android Broker does not support loggin in the "default" user
+        /// </summary>
```
jennyf19, line 317: "...does not supporting logging in as a "default" user."

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
jennyf19: what exactly is a default user? what if i don't want to use the default (as the customer signing in), or how is the default set?

---
In reply to: [506400556](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2113#discussion_r506400556) [](ancestors = 506400556)

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -28,49 +32,68 @@ internal class SilentRequest : RequestBase
...
+                {
+                    _logger.Verbose("No account passed to AcquireTokenSilent and no broker configured.  ");
+                    throw new MsalUiRequiredException(
+                       MsalError.UserNullError,
+                       MsalErrorMessage.MsalUiRequiredMessage,
+                       null,
+                       UiRequiredExceptionClassification.AcquireTokenSilentFailed);
+                }
+
+                _logger.Verbose("No account passed to AcquireTokenSilent. Some brokers may be able to log in user with a default account...");
```
jennyf19, line 70: >Some brokers [](start = 74, length = 12)

looks like it's only WAM who can do this, so maybe just mention that so we don't get iOS/android devs asking about it.

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IPlatformProxy.cs
```diff
@@ -112,7 +112,7 @@ internal interface IPlatformProxy
         /// does not support silent auth and gives the RT back to MSAL.
         /// </summary>
         /// <returns></returns>
-        bool CanBrokerSupportSilentAuth();
+        bool CanBrokerSupportSilentAuth(); // TODO: get rid of this!
```
jennyf19, line 115: >// [](start = 43, length = 2)

lol. is there an issue tracking that work?

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
```diff
@@ -1,698 +1,755 @@
...
+
+        [TestMethod]
+        public void HandleInstallUrl_Throws()
+        {
+            AssertException.Throws<NotImplementedException>(() => _wamBroker.HandleInstallUrl("http://app"));
+        }
+
+        [TestMethod]
+        [Description("For AcquireTokenSilent, plugin selection occurs based on the authority only, as this is always tenanted.")]
+        public async Task Ats_PluginSelection_Async()
```
jennyf19, line 83: so if the dev passes in /common will you pull the tenantId from elsewhere?

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
```diff
@@ -1,698 +1,755 @@
...
+            string aadHomeAccId = $"{TestConstants.Uid}.{TestConstants.Utid}";
+            // Arrange
+            using (var harness = CreateTestHarness())
+            {
+                var wamAccountProvider = new WebAccountProvider("id", "user@contoso.com", null);
+                var requestParams = harness.CreateAuthenticationRequestParameters(TestConstants.AuthorityConsumerTidTenant); // MSA
+                requestParams.Account = new Account(
+                                    aadHomeAccId, // matching in on home acc id
+                                    "doesnt_matter@contoso.com", // matching is not on UPN
+                                    null); // account does not have wam_id, might be coming directly from WAM
```
jennyf19, line 553: when is the wam_id included and what does it mean?

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
trwalke, line 24: >AcquireTokenSilentDefaultUserAsync [](start = 32, length = 34)

Would this be the currently signed in user on windows?

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -28,49 +32,68 @@ internal class SilentRequest : RequestBase
...
-                    }
+                    _logger.Info("Attempting to use broker instead");
+                    return await _brokerStrategyLazy.Value.ExecuteAsync(cancellationToken).ConfigureAwait(false);
                 }
 
                 throw;
             }
         }
 
+        private static bool IsTryWithBrokerError(string errorCode)
```
trwalke, line 88: >IsTryWithBrokerError [](start = 28, length = 20)

Suggest changing to something like "ShouldTryWithBroker".

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -89,5 +112,62 @@ internal async Task<AuthenticationResult> ExecuteTestAsync(CancellationToken can
...
+                if (accounts.Count() == 0)
+                {
+                    throw new MsalUiRequiredException(
+                                MsalError.NoAccountForLoginHint,
+                                MsalErrorMessage.NoAccountForLoginHint,
+                                null,
+                                UiRequiredExceptionClassification.AcquireTokenSilentFailed);
+                }
+
+                if (accounts.Count() > 1)
```
trwalke, line 148: As jenny said. accounts may be null im assuming

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -28,49 +32,68 @@ internal class SilentRequest : RequestBase
...
         }
 
-        internal async override Task PreRunAsync()
-        {
-            await _clientStrategy.PreRunAsync().ConfigureAwait(false);
-        }
-
         protected override async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
         {
+            await UpdateRequestWithAccountAsync().ConfigureAwait(false);
```
trwalke, line 47: The broker silent auth did not do this before. Are you aware of any potential side effects?

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/ISilentAuthRequestStrategy.cs
```diff
@@ -3,13 +3,12 @@
 
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Identity.Client.OAuth2;
```
trwalke, line 6: needed?

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
jmprieur: It requires an aka.ms link;

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
bgavrilMS: The default user is decided by Windows. I'm trying to get the full details from WAM folks on the exact algorithm. In complex cases, Windows may not be able to determine the "default" user and interaction is required.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
bgavrilMS: No, log messages do not require aka.ms links. This is verbose log message. The MSAL team (us) are the consumers of logs, not developers.

jennyf19: I disagree on this one. Customers do consume the logs, as does css and anyone helping w/a support case or issue, so way more then the MSAL team. I would at least remove the "no broker configured" part.

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
bgavrilMS: I am not sure what the algorithm is, since you can be signed in with both a Work and School and an MSA account. But generally yes, currently signed in user.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
bgavrilMS: Ok no problem.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
bgavrilMS: No, Linq always returns empty collection instead of null (and so should we). But you did spot a problem actually, I should use `.Count` instead of `.Count()` since `.Count()` is more expensive.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
bgavrilMS: That is a very good question and probably one of the more complex areas of this integration. The code is [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/83acb1d345757b1df611a135105766b31c8892d6/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs#L630) 

In particular for AcquireTokenSilent we can use: input authority and account home tenant id

if input authority is `common`, and an account was passed in, then see if home tenant ID is MSA_tenantID. However, account==null, then MSAL will log in the current user, and we have a way of figuring out if the "current user" is MSA or not - [2 liner](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/83acb1d345757b1df611a135105766b31c8892d6/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WebAccountProviderFactory.cs#L28)

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
bgavrilMS: This is described in the technical aspects of the WAM integration [here](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/1972?_a=files&path=%2F%5BWindows%5D%20Wam%20Integration%2FWAM%20-%20MSA.NET%20integration%2FWAM%20integration%20-%20technical%20aspects.md)

As a summary, apps are not generally allowed to list WAM accounts, so `wam.GetAccounts()` typically returns null. But if the user logs in interactively, WAM returns an internal "wam_account_id" back to MSAL. We store this in our token cache and if the user performs `GetAccounts` we can now return this account and allow the app to perform `AcquireTokenSilent`

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
bgavrilMS: It did, this was part of the dreaded "PreExecute" which is now history.

trwalke: The PreRunAsync for the broker strategy does nothing today. it was only for the client strategy. Ive also seen some issues with broker coming from this code.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5bca60568023bf239a1f5f4d02579ba7cad7365c/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentBrokerAuthStrategy.cs#L93

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5bca60568023bf239a1f5f4d02579ba7cad7365c/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentClientAuthStrategy.cs#L38

bgavrilMS: Ok, I'll have a look

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
jennyf19: i see. thanks. appreciate the links

neha-bhargava (approved):
Looks good. Thanks

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
jmprieur: We could say to simplify "the user logged in in the Windows session" (the account that was used to sign-in to Windows in the morning when you had your coffee mug in the hand?)"?

jmprieur:
@bgavrilMS : just to make sure I understand:
- This is for .NET Framework and UWP
- No change of public API.
- but the behavior changes: if the account passed to AcquireTokenSilent is null, there is no longer an exception if .WithBroker is set to true, but WAM is involved?

bgavrilMS:
@jmprieur - 

`This is for .NET Framework and UWP`  No, just UWP. WAM on UWP does not require any extra dependencies.
`No change of public API.` WithBroker(true) is added to UWP public API. 
`but the behavior changes: if the account passed to AcquireTokenSilent is null, there is no longer an exception if .WithBroker is set to true, but WAM is involved?` yes

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IPlatformProxy.cs
bgavrilMS: I'll remove the comment for now. "No refactoring on the backlog" is what I heard from a knowledgeable PM.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
bgavrilMS: If you look at line 134 you'll see a Linq expression that does some filtering etc. Linq always returns an empty enumerable in case the underlying data has 0 elements.

PR #2114, opened 2020-10-17
Update regional IMDS endpoint

neha-bhargava:


jmprieur (approved):
This makes sense to me
Thanks @neha-bhargava

jennyf19 (approved):
Nit: maybe create a constant?

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -127,7 +127,7 @@ private void AddMockedResponse(HttpResponseMessage responseMessage)
                     new MockHttpMessageHandler
                     {
                         ExpectedMethod = HttpMethod.Get,
```
bgavrilMS, line 129: You can also add "ExpectedQueryParams" which would allow you to check the api-version

bgavrilMS (approved):
LGTM, but see minor test improvement idea

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -127,7 +127,7 @@ private void AddMockedResponse(HttpResponseMessage responseMessage)
...
-                        ExpectedUrl = "http://169.254.169.254/metadata/instance/compute/api-version=2020-06-01",
+                        ExpectedUrl = "http://169.254.169.254/metadata/instance/compute",
```
bgavrilMS, line 130: You have some tests failing which are easy to fix, just a small note to rename them `happyPathAsync` to a more meaningul name as it's hard to understand which test failed when lookinag the CI system. Also, please try to use PascalCase for all methods, private and public. I'll see about updating our code style to emit warnings.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
neha-bhargava: Added

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
neha-bhargava: Makes sense. Updated the names.

PR #2119, opened 2020-10-20
Add test app for regional auth

neha-bhargava:


RegionalTestApp/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            Console.WriteLine("CCA created");
+            
+            var result = await cca.AcquireTokenForClient(scopes)
+            .WithExtraQueryParameters(dict)
+            .WithAzureRegion(true)
+            .ExecuteAsync()
+            .ConfigureAwait(false);
+
+            Console.WriteLine("Token obtained");
```
henrik-me, line 49: >Console.WriteLine("Token obtained"); [](start = 12, length = 36)

There is no check on what the result contains... perhaps there should be some validation of the auth result, that also indicates the token was obtained from ests-r?

RegionalTestApp/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+                ["allowestsrnonmsi"] = "true"
+            };
+
+            string secret = "Lab_Secret"; //Lab secret from TestConstants.MsalCCAKeyVaultUri
+
+            //Uncomment below line to test in dev env.
+            //Environment.SetEnvironmentVariable("REGION_NAME", "centralus");
+            string[] scopes = new string[] { $"{clientId}/.default", };
+            var cca = ConfidentialClientApplicationBuilder.Create(clientId)
+                .WithAuthority("https://login.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47")
```
henrik-me, line 35: >login.windows.net/ [](start = 40, length = 18)

any reason the old windows.net is used instead of onmicrosoftonline.com?

RegionalTestApp/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            Console.WriteLine("Token obtained");
+            Console.ReadLine();
+        }
+
+        private static void Log(LogLevel level, string message, bool containsPii)
+        {
+            if (containsPii)
+            {
+                Console.ForegroundColor = ConsoleColor.Red;
```
henrik-me, line 57: aside: perhaps foreground and background color should be the same? for the PII data, then only by marking it will the data be visible?

henrik-me:
Great. Consider making this scenario part of our already existing console devapp.

henrik-me:
I'm missing the details of the validation confirming the scenario works as expected. Added a few suggestions.

RegionalTestApp/Program.cs
neha-bhargava: No, I can change it to onmicrosoftonline.com

RegionalTestApp/Program.cs
```diff
@@ -0,0 +1,107 @@
...
+                // From the collection of unexpired certificates, find the ones with the correct name.
+                X509Certificate2Collection signingCert = currentCerts.Find(X509FindType.FindByThumbprint, thumbprint, false);
+
+                // Return the first certificate in the collection, has the right name and is current.
+                cert = signingCert.OfType<X509Certificate2>().OrderByDescending(c => c.NotBefore).FirstOrDefault();
+            }
+            return cert;
+        }
+
+        private static async Task AcquireTokenAsync(bool withAzureRegion, X509Certificate2 certificate)
```
jennyf19, line 86: >bool withAzureRegion [](start = 52, length = 20)

move this parameter to the end and set it to false or true by default, so it's not needed, then you can remove either the true/false in line 43 or 47 and just pass the cert.

RegionalTestApp/Program.cs
```diff
@@ -0,0 +1,107 @@
...
+                .Build();
+
+            Console.WriteLine("CCA created");
+
+            var result = await cca.AcquireTokenForClient(scopes)
+            .WithAzureRegion(withAzureRegion)
+            .ExecuteAsync()
+            .ConfigureAwait(false);
+
+            Console.WriteLine("Access token:" + result.AccessToken);
```
jennyf19, line 104: is there anything else in the result that would be interesting to make assure the test worked as expected? I don't remember the specific values in the auth result, but maybe you can print more info to the console.

jennyf19 (approved):
:shipit:

RegionalTestApp/Program.cs
jennyf19: i still see login.windows.net, i think this is still open.

---
In reply to: [508175872](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2119#discussion_r508175872) [](ancestors = 508175872)

bgavrilMS: `login.windows.net` is a valid alias ... also testing that aliasing still works :)

RegionalTestApp/Program.cs
bgavrilMS: As far as I remember the AT token will have a special claim that indicates regionality. Not sure if this claim will also be present in the ID token.

neha-bhargava: Access tokens are same as mentioned in the spec. A token obtained for one region is valid for the non-regional endpoint (tokens for "westus2.login.microsoftonline.com " are the same as tokens for "login.microsotonline.com "). And vice-versa. It's the same token, minus one claim called rh

neha-bhargava: But then there is internal routing to global endpoint on ests side. I think all we should be verifying is that we get the access token and the token request is going to regional endpoint.

RegionalTestApp/Program.cs
neha-bhargava: I think that would make it confusing. It is just a test.

RegionalTestApp/Program.cs
neha-bhargava: There are logs that print on the console and it contains more info. I have added what to check from logs in the steps. I was checking other tests and did not find anything else that might fit in here. Let me know if you have suggestions for something specific you think could be added to the logs. I might not be aware of it.

trwalke (approved):
LGTM

Issue #2121, opened 2020-10-20
[Bug] TokenCache.SetIosKeychainSecurityGroup property wrongly exposed as public API


bgavrilMS:
**Which Version of MSAL are you using ?**
4.20

Property 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/929bef3d47a565848e71f80255ecca0cecaf5496/src/client/Microsoft.Identity.Client/TokenCache.cs#L107
is not supposed to be public / settable by developers. Developers have `.WithIosKeychain()` method

jmprieur:
Changing to be an enhancement, as this is not actionable immediately, but rather to a next major version.

Issue #2120, opened 2020-10-20
[Bug] Expose Certificate property on IConfidentialClient app


bgavrilMS:
**Which Version of MSAL are you using ?**
4.20

`Certificate { get; }` is a property on the ConfidentialClientAppliction but it is not found on the interface IConfidentialClientAppliction 

In general, public methods and properties should be exposed on the implementing interfaces.

PR #2122, opened 2020-10-20
Expose Certificate property on IConfidentialClient interface

bgavrilMS:
Fixes for #2120 (customer request) and #2121

### DIFF
--- a/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
+++ b/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig;
 
@@ -23,6 +24,11 @@ namespace Microsoft.Identity.Client
         /// See https://aka.ms/msal-net-token-cache-serialization. This is taken care of by MSAL.NET on other platforms.
         /// </remarks>
         ITokenCache AppTokenCache { get; }
+
+        /// <summary>
+        /// The certificate used to create this <see cref="ConfidentialClientApplication"/>, if any.
+        /// </summary>
+        X509Certificate2 Certificate { get; }
 
         /// <summary>
         /// [V3 API] Acquires a security token from the authority configured in the app using the authorization code
--- a/src/client/Microsoft.Identity.Client/TokenCache.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.cs
@@ -85,7 +85,10 @@ namespace Microsoft.Identity.Client
             LegacyCachePersistence = proxy.CreateLegacyCachePersistence();
 
 #if iOS
+
+#pragma warning disable CS0618 // This method was mistakenly made public, so it was obsoleted to prevent use
             SetIosKeychainSecurityGroup(serviceBundle.Config.IosKeychainSecurityGroup);
+#pragma warning restore CS0618 
 #endif // iOS
 
             IsAppTokenCache = isApplicationTokenCache;
--- a/src/client/Microsoft.Identity.Client/TokenCache.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
-        /// 
+        /// Please use PublicClientApplicationBuilder.WithIosKeychainSecurityGroup instead
         /// </summary>
         /// <param name="securityGroup"></param>
         [Obsolete("This method was made public by mistake and will be removed at the next major version. Please use WithIosKeychainSecurityGroup method on the PublicClientApplicationBuilder object instead.")]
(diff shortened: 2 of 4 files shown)


tests/Microsoft.Identity.Test.Unit/PublicApiTests/ObjectsAndInterfaces.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft.Identity.Test.Unit.PublicApiTests
+{
+    [TestClass]
+    public class ObjectsAndInterfaces
+    {
+        [TestMethod]
+        [Description(
+            "This test checks that we don't accidently expose a public property or method on" +
+            " an object and forget to expose it on the associated interface.")]
+        public void InterfaceExposesPublicSurfaceOfObjects()
```
bgavrilMS, line 17: This should protect against future similar bugs where we expose a public property / method on the obj but not on the interface.

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
```diff
@@ -24,6 +25,11 @@ public partial interface IConfidentialClientApplication : IClientApplicationBase
         /// </remarks>
         ITokenCache AppTokenCache { get; }
 
+        /// <summary>
+        /// The certificate used to create this <see cref="ConfidentialClientApplication"/>, if any.
+        /// </summary>
+        X509Certificate2 Certificate { get; }
```
jmprieur, line 31: Do we want to give an explicit implementation? (as this is a change of a public interface ?

src/client/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -103,7 +103,11 @@ internal TokenCache(IServiceBundle serviceBundle, ILegacyCachePersistence legacy
             LegacyCachePersistence = legacyCachePersistenceForTest;
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        ///
```
jmprieur, line 107:
```suggestion
        /// Please use PublicClientApplicationBuilder.WithIosKeychainSecurityGroup instead
```

tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
```diff
@@ -47,6 +49,68 @@ public static void AreEqual(DateTimeOffset expected, DateTimeOffset actual, Time
             Assert.IsTrue(DictionariesAreEqual(dict1, dict2, valueComparer));
         }
 
+        public static void InterfaceExposesObject(Type theObj, Type theInterface, string[] except = null)
```
jmprieur, line 52: Good idea :)
Of course sometimes a class implements 2 interfaces (See `TokenAcquisition` implements `ITokenAcquisition` and `ITokenAcquisitionInternal` in Microsoft.Identity.Web), but the private one can be implemented explicitly, and therefore the members don't need to be public.

jmprieur (commented):
Don't we have a breaking change by adding a member to the interface?
Do we want to use the [default implementation on interfaces](https://devblogs.microsoft.com/dotnet/default-implementations-in-interfaces/) C# 8.0 feature?
or have this as an extension method?

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
bgavrilMS: Sadly this isn't a possibility in MSAL, as this is a C# 8 feature, and C# 8 is not available with ancient .net 45

tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
bgavrilMS: Exactly, explicit interface implementations are different. `TokenCache` also implements `ITokenCache` and `ITokenCacheInternal`

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
bgavrilMS: I'd rather hide MSAL 4.20 - 4.22 from nuget :)

tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
jmprieur: So maybe an extension method?
Otherwise shall we bump-up the version to 5.0 ?

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
jmprieur: So maybe an extension method?
Otherwise shall we bump-up the version to 5.0 ?

bgavrilMS: IConfidentialClient2 ?

bgavrilMS: I think extension method is better...

jmprieur: I'm not a fan of IConfidentialClient2.

Issue #2133, opened 2020-10-23
[Bug] Interactive Brokered authentication with Android not working with versions 4.20.0 and above


bemartin:
**Which Version of MSAL are you using ?**
MSAL 4.20.0, MSAL 4.20.1, MSAL 4.21.0, MSAL 4.21.1

**Platform**
Xamarin Android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

Other? - please describe;

**Is this a new or existing app?**
The app is in production, and I am working on upgrading to a new version of MSAL

**Repro**

```csharp
return await ClientApplication
    .AcquireTokenInteractive(scopes)
    .WithAuthority(authority)
    .ExecuteAsync();
```

**Expected behavior**
The broker is launched, the user selects the account to use, control is returned to the calling application

**Actual behavior**
Our application is sent to the background and nothing else happens

**Additional context/ Logs / Screenshots**
Without any changes to our code, this works fine with MSAL 4.19.0 but fails with any version after that. We are currently running 4.19.0 in production.

The last logs we are seeing from our app are:
```
10-23 14:47:09.358: MSAL 4.21.1.0 MSAL.Xamarin.Android 29 [10/23/2020 21:47:09] ****** Starting D:\a\1\s\src\client\Microsoft.Identity.Client\Platforms\Android\Broker\AndroidBrokerHelper:ValidateBrokerRedirectURI
10-23 14:47:09.373: MSAL 4.21.1.0 MSAL.Xamarin.Android 29 [10/23/2020 21:47:09] ****** Finished D:\a\1\s\src\client\Microsoft.Identity.Client\Platforms\Android\Broker\AndroidBrokerHelper:ValidateBrokerRedirectURI in 14 ms
10-23 14:47:09.532: MSAL 4.21.1.0 MSAL.Xamarin.Android 29 [10/23/2020 21:47:09] ****** Finished D:\a\1\s\src\client\Microsoft.Identity.Client\Platforms\Android\Broker\AndroidBroker:AcquireTokenInteractiveViaBrokerAsync in 246 ms
[...]

PR #2154, opened 2020-10-30
Trwalke/broker prompt

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2133

Sent a preview package to customer and this fix has solved the issue for them.


src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerRequest.cs
```diff
@@ -65,7 +65,17 @@ public Uri RedirectUri
             AcquireTokenInteractiveParameters acquireTokenInteractiveParameters)
         {
             BrokerRequest br = FromAuthenticationParameters(authenticationRequestParameters);
-            br.Prompt = acquireTokenInteractiveParameters.Prompt.PromptValue.ToUpperInvariant();
+            var prompt = acquireTokenInteractiveParameters.Prompt.PromptValue;
```
bgavrilMS, line 68: nit: you can compare Prompt objects directly as they have `==` overridden. Pretty obscure, not sure who decided that :(

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerRequest.cs
```diff
@@ -65,7 +65,17 @@ public Uri RedirectUri
...
+
+            if (prompt == Client.Prompt.NoPrompt.PromptValue || prompt == Client.Prompt.NotSpecified.PromptValue)
```
bgavrilMS, line 70: Can you apply this fix to iOS as well? If unsure about the value needed, then use the old "select_account"

bgavrilMS (changes requested):
should apply to iOS broker as well

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerRequest.cs
henrik-me: why select account?

---
In reply to: [515239838](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2154#discussion_r515239838) [](ancestors = 515239838)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerRequest.cs
```diff
@@ -65,7 +65,17 @@ public Uri RedirectUri
             AcquireTokenInteractiveParameters acquireTokenInteractiveParameters)
         {
             BrokerRequest br = FromAuthenticationParameters(authenticationRequestParameters);
-            br.Prompt = acquireTokenInteractiveParameters.Prompt.PromptValue.ToUpperInvariant();
+            var prompt = acquireTokenInteractiveParameters.Prompt;
+
+            if (prompt == Client.Prompt.NoPrompt || prompt == Client.Prompt.NotSpecified)
+            {
+                br.Prompt = "NONE";
```
bgavrilMS, line 72: Isn't this a breaking change?

trwalke: we used to default to "none" for broker a long time ago but I think the behavior for "none" and "select_account" is ultimately the same. Ill switch to select account.

bgavrilMS (approved):

PR #2190, opened 2020-11-16
Update bug_report.md

bgavrilMS:
Make it clear that logs and network traces are mandatory. I also updated the wiki to explain how to get Network Traces with and without Fiddler. 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging


jmprieur (approved):
LGTM. Thanks @bgavrilMS

PR #2191, opened 2020-11-16
Fix PoP tests failing because of crypto

bgavrilMS:


tests/Microsoft.Identity.Test.Common/Core/Helpers/RSACertificatePopCryptoProvider.cs
```diff
@@ -10,49 +10,5 @@
...
-        private X509Certificate2 _cert;
-        private RSACryptoServiceProvider _signingKey;
-
-        public string CannonicalPublicKeyJwk { get; private set; }
-
-        public string CryptographicAlgorithm { get => "RS256"; }
-
-        private void InitializeSigningKey()
-        {
-            _signingKey = _cert.PrivateKey as RSACryptoServiceProvider;
```
bgavrilMS, line 36: The problem was here - the cast fails for new-style certs. RSACng is the "next-gen provider" for crypto, not RSACryptoServiceProvider. 

But in any case, the guideline is to use the `RSA` abstract class instead of casting to an actual implementation. This seems to work fine.

trwalke: ok, still seems odd. The cert used for this test is the same on my machine and on the cloud build but it works on my machine with RSACryptoServiceProvider so I'm not sure how this is able to resolve it. Also this was working on the pipeline for about a week with RSACryptoServiceProvider.

trwalke: ok. seems to be working now

trwalke (approved):
LGTM

Issue #2164, opened 2020-11-05
[Bug] ios http stack causing cancellation exceptions


bgavrilMS:
This is related to an ICM..

I think the issue might be related to https://github.com/xamarin/xamarin-macios/issues/9370 
 
Our SSO extension makes background requests as a result of which the requests got cancelled. 
I tried the workaround by using : 
`
var handler = new NSUrlSessionHandler() {
  BypassBackgroundSessionCheck = false, // readd the hack
};
var httpClient = new HttpClient (handler);
`

and that seemed to not cancel the request and I was able to SSO fine.

jmprieur:
@bgavrilMS let's take it for 4.23.0

henrik-me:
Should we allow a way for customers to opt. out of us setting this flag (besides from providing own handler)?

bgavrilMS:
I think this degree of control is too granular / complex and customers are not interested in obscure http stack bugs.

PR #2192, opened 2020-11-17
fix ios http stack bug

jennyf19:
#2164 
@bgavrilMS i couldn't test this as my provisioning profile is no longer working :( that lasted like 1 week.


bgavrilMS:
@jennyf19 - unfortunately our iOS build is broken so we cannot validate that way either. I suppose you could run a simple app on your Mac with this code to see if it works (automated profile etc.). But ideally we would test this via automation?

@neha-bhargava - any news on this? Sometimes you can guess the strings for the new Mac machines needed to run our tests.

jennyf19:
I can't use the Mac either because of the provisioning profile. so have to wait on that, unless someone else can test.

neha-bhargava (approved):

jennyf19:
build passed [here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=665612&view=results)...going to merge and see what happens.

PR #2197, opened 2020-11-17
Introduce PublicClientApplication.CurrentBrokerAccount for default WA…

bgavrilMS:
…M login

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -59,9 +59,9 @@ namespace Microsoft.Identity.Client
 
         /// <summary>
         /// This error code comes back from <see cref="IClientApplicationBase.AcquireTokenSilent(System.Collections.Generic.IEnumerable{string}, IAccount)"/> calls when the 
-        /// <see cref="PublicClientApplication.CurrentBrokerAccount"/> user is passed as the <c>account</c> parameter. Only some brokers (WAM) can login the current user.
+        /// <see cref="PublicClientApplication.OperatingSystemAccount"/> user is passed as the <c>account</c> parameter. Only some brokers (WAM) can login the current user.
         /// <para>Mitigation</para>
-        /// On Windows, use the broker .WithBroker(true) or use a different account, or otherwise call <see cref="IPublicClientApplication.AcquireTokenInteractive(System.Collections.Generic.IEnumerable{string})"/>
+        /// On Windows, use the broker via .WithBroker(true), use a different account, or otherwise call <see cref="IPublicClientApplication.AcquireTokenInteractive(System.Collections.Generic.IEnumerable{string})"/>
         /// </summary>
         public const string CurrentBrokerAccount = "current_broker_account";
 
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -56,6 +56,15 @@ namespace Microsoft.Identity.Client
         /// </summary>
         public const string UserNullError = "user_null";
 
+
+        /// <summary>
+        /// This error code comes back from <see cref="IClientApplicationBase.AcquireTokenSilent(System.Collections.Generic.IEnumerable{string}, IAccount)"/> calls when the 
+        /// <see cref="PublicClientApplication.CurrentBrokerAccount"/> user is passed as the <c>account</c> parameter. Only some brokers (WAM) can login the current user.
+        /// <para>Mitigation</para>
+        /// On Windows, use the broker .WithBroker(true) or use a different account, or otherwise call <see cref="IPublicClientApplication.AcquireTokenInteractive(System.Collections.Generic.IEnumerable{string})"/>
+        /// </summary>
+        public const string CurrentBrokerAccount = "current_broker_account";
+
         /// <summary>
         /// This error code denotes that no account was found having the given login hint.
         /// <para>What happens?</para>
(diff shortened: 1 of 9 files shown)


src/client/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -29,6 +29,30 @@ internal PublicClientApplication(ApplicationConfiguration configuration)
...
+            new Account(CurrentBrokerAccountDescriptor, null, null, null);
+
+        /// <summary>
+        /// A special account value that indicates that the current Operating System account should be used 
+        /// to login the user. Not all operating systems and authentication flows support this concept, in which 
+        /// case calling `AcquireTokenSilent` will throw an <see cref="MsalUiRequiredException"/>. 
+        /// 
+        /// Currently only the Windows broker is able to login the current Windows user, see https://aka.ms/msal-net-wam for details.
+        /// </summary>
+        public static IAccount CurrentBrokerAccount
```
bgavrilMS, line 43: @jmprieur  - how does this sound? Happy to take renaming suggestings. Currently this only works for WAM, and there's no plan to use "current account" concept in other brokers. Perhaps `CurrentWAMAccount` is better?

"Current account" concept also potentially makes sense in IWA (not covered here) and FLW scenarios (which we don't support)

bgavrilMS: Also please review the public comments.

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -56,6 +56,15 @@ public static class MsalError
         /// </summary>
         public const string UserNullError = "user_null";
 
+
```
jennyf19, line 59:
```suggestion
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -56,6 +56,15 @@ public static class MsalError
...
+        /// <summary>
+        /// This error code comes back from <see cref="IClientApplicationBase.AcquireTokenSilent(System.Collections.Generic.IEnumerable{string}, IAccount)"/> calls when the 
+        /// <see cref="PublicClientApplication.CurrentBrokerAccount"/> user is passed as the <c>account</c> parameter. Only some brokers (WAM) can login the current user.
+        /// <para>Mitigation</para>
+        /// On Windows, use the broker .WithBroker(true) or use a different account, or otherwise call <see cref="IPublicClientApplication.AcquireTokenInteractive(System.Collections.Generic.IEnumerable{string})"/>
```
jennyf19, line 64:
```suggestion
        /// On Windows, use the broker `.WithBroker(true)` , use a different account, or otherwise call <see cref="IPublicClientApplication.AcquireTokenInteractive(System.Collections.Generic.IEnumerable{string})"/>
```

src/client/Microsoft.Identity.Client/PublicClientApplication.cs
jennyf19: I think CurrentBrokerAccount is more flexible, as you mention there could be other use cases outside of WAM. My suggestion would be to move the comment on line 41 to the very beginning (like 37). A developer may not realize it's no available on iOS and not read the entire comment, so starting w/the limitation might be a good idea. also, is this the default user?

---
In reply to: [525507602](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2197#discussion_r525507602) [](ancestors = 525507602)

jennyf19: Also, i think the wiki page you made is really very well done. thanks.

---
In reply to: [525587965](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2197#discussion_r525587965) [](ancestors = 525587965,525507602)

tests/devapps/NetCoreTestApp/Program.cs
```diff
@@ -234,8 +234,12 @@ private static IConfidentialClientApplication CreateCca()
         {
             IConfidentialClientApplication cca = ConfidentialClientApplicationBuilder
                 .Create(s_clientIdForConfidentialApp)
+                //.WithAuthority("https://login.microsoftonline.com/12354-12345-123-123-23")
```
jennyf19, line 237: nit: remove the commented code? if it's needed maybe a comment as to what it can be used for?

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/MsalError.cs
jmprieur: Does ``` work? maybe rather <code>.WithBroker</code> in XML comments?

src/client/Microsoft.Identity.Client/PublicClientApplication.cs
jmprieur: From an experience point of view this is the account currently logged in in the OS session. But I guess CurrentBrokerAccount could work. I'm not sure people know necessarily what a broker is, and that's an implementation detail?. What about OperatingSystemCurrentAccount? Or OperatingSystemSessionAccount? or OperatingSystemAccount?

jmprieur (approved):
LGTM.
I left a couple of comments

src/client/Microsoft.Identity.Client/PublicClientApplication.cs
bgavrilMS: We can do `OperatingSystemCurrentAccount` but note that in the absence of WAM, MSAL will throw a UiRequiredException. Let's see if people get confused by this or not. At some point we may want to say AcquireTokenSilent(OperatingSystemCurrentAccount) = IWA but I'm rather cautious of that since IWA is such a pain to setup.

jmprieur: What do you mean, @bgavrilMS : it's a pain to setup?

Issue #2166, opened 2020-11-06
Add source of region discovery to telemetry


neha-bhargava:
To investigate the source of region discovery, add the source to telemetry. Which can be "ev" if the region is detected from environment variable or "imds" if it is detected from imds call.

PR #2199, opened 2020-11-17
Add region source to telemetry

neha-bhargava:
#2166


src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -53,19 +48,26 @@ public async Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority
...
+            ICoreLogger logger = requestContext.Logger;
+            string region;
+
             if (!Environment.GetEnvironmentVariable(RegionName).IsNullOrEmpty())
             {
-                logger.Info($"[Region discovery] Region found in environment variable: {Environment.GetEnvironmentVariable(RegionName)}");
-                return Environment.GetEnvironmentVariable(RegionName);
+                region = Environment.GetEnvironmentVariable(RegionName);
+                logger.Info($"[Region discovery] Region found in environment variable: {region}");
+
```
jennyf19, line 67: nit: add a "." and a space at the end of the log message.

src/client/Microsoft.Identity.Client/Region/RegionSource.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.;
+
+namespace Microsoft.Identity.Client.Region
```
jennyf19, line 4: nit: could this be added to the constants file instead? w/a comment under Region?

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
```diff
@@ -24,6 +25,27 @@ public class RegionalTelemetryTests : TestBase
         private MockHttpAndServiceBundle _harness;
         private ConfidentialClientApplication _app;
 
+        [TestInitialize]
+        public override void TestInitialize()
+        {
```
jennyf19, line 30: were the telemetry acceptance tests updated as well? i don't remember the class name atm.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -53,19 +48,26 @@ public async Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority
...
         }
 
 
-        private async Task<string> GetRegionAsync(ICoreLogger logger)
+        private async Task<string> GetRegionAsync(RequestContext requestContext)
         {
+            ICoreLogger logger = requestContext.Logger;
+            string region;
+
             if (!Environment.GetEnvironmentVariable(RegionName).IsNullOrEmpty())
```
jmprieur, line 63: Don't we want to extract the region in a variable before the test? isn't this a bit costly to call GetEnvironmentVariable(RegionName) twice ?

jmprieur (commented):
LGTM. Thanks @neha-bhargava

src/client/Microsoft.Identity.Client/Region/RegionSource.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    internal class RegionSource
```
bgavrilMS, line 6: Isn't it better to send numbers instead of strings? Those headers can get pretty large.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -49,7 +49,7 @@ public void RecordStoppedEvent(ApiEvent stoppedEvent)
         /// Csv expected format:
         ///      2|silent_successful_count|failed_requests|errors|platform_fields
         ///      failed_request is: api_id_1,correlation_id_1,api_id_2,correlation_id_2|error_1,error_2
-        ///      platform_fields: region_1,region_2
+        ///      platform_fields: region_1,region_source_1,region_2,region_source_2
```
bgavrilMS, line 52: Isn't this a breaking change? How will you tell the different between messages without region source and messages with region source?

src/client/Microsoft.Identity.Client/Region/RegionSource.cs
bgavrilMS, line 6: Please rename or add a comment to show that this is for telemetry purposes only.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
bgavrilMS: Or if you do not want to introduce version, add the new fields at the end. Because the number of fields is variable, it will be hard to distinguish versions.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
```diff
@@ -43,48 +65,39 @@ public class RegionalTelemetryTests : TestBase
...
-                AssertCurrentTelemetry(result.HttpRequest, ApiIds.AcquireTokenForClient, forceRefresh: true);
-                AssertPreviousTelemetry(
-                    result.HttpRequest,
-                    expectedSilentCount: 0,
-                    expectedFailedApiIds: new[] { ApiIds.AcquireTokenForClient },
-                    expectedCorrelationIds: new[] { step2CorrelationId },
-                    expectedErrors: new[] { "service_not_available" },
-                    expectedRegions: new[] { "centralus" });
-            }
+            Environment.SetEnvironmentVariable(TestConstants.RegionName, TestConstants.Region);
```
bgavrilMS, line 68: This enviroment variable is there to stay, i.e. this test will modify the global state of the machine. This will create problems when other regional tests run, because you do not have control over the order of the tests. 

Best practice is to cleanup or reset the global state. I would do both: 

1. All regional tests that set this env variable should have a "finally" clause where you clear the env variable (or you can do it via [TestCleanup])
2. All regional tests should clear this enviroment variable as part of their [TestSetup]

bgavrilMS (changes requested):
MSAL changes are ok. I'm a little concerned on the Kusto side of things.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
neha-bhargava: Done

src/client/Microsoft.Identity.Client/Region/RegionSource.cs
neha-bhargava: Added a comment.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
neha-bhargava: Leaving it as is for now as per discussion offline. Will have to use some strategy to add more fields to platform fields in future.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -78,12 +75,12 @@ public string GetLastRequestHeader()
...
                 {
-                    string region = ev[MsalTelemetryBlobEventNames.RegionDiscovered];
-
                     if (!firstFailure)
                         platformFields.Append(",");
 
-                    platformFields.Append(region);
+                    platformFields.Append(ev.RegionDiscovered);
+                    platformFields.Append(",");
+                    platformFields.Append(ev.RegionSource);
```
henrik-me, line 83: >platformFields.Append(ev.RegionSource); [](start = 20, length = 39)

seems like you are missing a check on whether regionsource is available?

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -120,13 +117,22 @@ public string GetCurrentRequestHeader(ApiEvent eventInProgress)
...
             eventInProgress.TryGetValue(MsalTelemetryBlobEventNames.RegionDiscovered, out string regionDiscovered);
+            eventInProgress.TryGetValue(MsalTelemetryBlobEventNames.RegionSource, out string regionSource);
+
+            // Since regional fields will only be logged in case it is opted.
+            var regionalFields = new StringBuilder();
+            if (!string.IsNullOrEmpty(regionDiscovered))
+            {
+                regionalFields.Append(regionDiscovered);
+                regionalFields.Append(",");
+                regionalFields.Append((regionSource));
```
henrik-me, line 128: is a check not needed for this?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
neha-bhargava: In case the region is discovered, the region source will be there and that is the reason I didn't add a check.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
neha-bhargava: If the region is discovered, the region source will be there and there is a check to see if the region is discovered.

---
In reply to: [526623194](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2199#discussion_r526623194) [](ancestors = 526623194)

Issue #2140, opened 2020-10-27
MSAL does not throw UiRequiredException when it should (Android Broker scenarios)


shristikatiyar:
**Which Version of MSAL are you using ?**
4.20

**Platform**
Xamrin.Android

**What authentication flow has the issue?**
Mobile

Other? - please describe;

**Is this a new or existing app?**
The app is in production and we have migrated from  ADAL to MSAL

**Repro**

                authResult = await pc.AcquireTokenSilent(scopes, account).ExecuteAsync();

                DataInsightsProvider.Instance.TrackEvent(DataInsightsEvents.FetchTokenSilentlySuccess);
            }
            catch (MsalUiRequiredException e)
            {
                // This exception is expected if there is no token in the cache.
                this.logger.Log($"{_logTagAuth}: Encountered error during silent authentication." +
                                                $" Error code = {e.ErrorCode}. Exception: {e}");
            }

**Expected behavior**
If a silent token acquisitaion call fails due to reason such as (no account found or broken resfresh toke is invalid) insted of MSAL  Service Exception we should get MSAL UI Required Exception so that we a interactive signin is performed.

**Actual behavior**
For some user in their session when a silent token acquistion is taking place MSAL Service Exception is thrown from broker
with error codes no_account_found or broker refresh token is invalid

**Possible Solution**
These error should throw a MSAL UI Required Exception

**Additional context/ Logs / Screenshots**
{\"Error Message\":\"Non-Graph Authentication Error:\",\"Exception Message\":\"Broker response returned error: Account does not exist in the broker, cannot proceed with silent request\"}
"Message\":\"nt.Internal.Requests.SilentBrokerAuthStrategy.SendTokenRequestToBrokerAsync () 

From the logs from telemetry error is returned from this function (while broker is trying to silently acquire token)

bgavrilMS:
@shristikatiyar - what exactly are the error codes (verbatim)? You mentioned `no_account_found` and smth related to refresh token.

bgavrilMS:
P1 since user impact is that they can't login or app crashes.

bgavrilMS:
We need to find out the exact error codes which need to be translated to `MsalUiRequiredException` from Android broker.

shristikatiyar:
> We need to find out the exact error codes which need to be translated to `MsalUiRequiredException` from Android broker.

Hi Bogdan , these are the two error codes we are getting. attaching the screenshot

Broker refresh token is invalid
[...]

PR #2200, opened 2020-11-18
process android broker error codes which should result in UI required…

jennyf19:
… exceptions

not sure if this is what you guys want to do here. #2140 also, did not test on a device as I only have my personal pixel and don't want to mess up the authenticator app. :/

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
@@ -42,9 +42,9 @@ namespace Microsoft.Identity.Client.Internal.Broker
 
         public const string TokenType = "token_type";
 
-        //Error codes returned from broker
-        public const string NoTokenFound = "no_tokens_found";
-        public const string NoAccountFound = "no_account_found";
-        public const string InvalidRefreshToken = "Broker refresh token is invalid";
+        //Error codes returned from Android broker
+        public const string AndroidNoTokenFound = "no_tokens_found";
+        public const string AndroidNoAccountFound = "no_account_found";
+        public const string AndroidInvalidRefreshToken = "Broker refresh token is invalid";
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
@@ -44,5 +44,7 @@ namespace Microsoft.Identity.Client.Internal.Broker
 
         //Error codes returned from broker
         public const string NoTokenFound = "no_tokens_found";
+        public const string NoAccountFound = "no_account_found";
+        public const string InvalidRefreshToken = "Broker refresh token is invalid";
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs
@@ -79,9 +79,9 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 _logger.Info(
                     LogMessages.ErrorReturnedInBrokerResponse(msalTokenResponse.Error));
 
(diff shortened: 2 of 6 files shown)


src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
```diff
@@ -44,5 +44,7 @@ internal static class BrokerResponseConst
 
         //Error codes returned from broker
         public const string NoTokenFound = "no_tokens_found";
+        public const string NoAccountFound = "no_account_found";
+        public const string InvalidRefreshToken = "Broker refresh token is invalid";
```
jennyf19, line 48: not sure if this is legit, i used the error codes from the customer's screen shot.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs
```diff
@@ -79,7 +79,9 @@ private async Task<MsalTokenResponse> SendTokenRequestToBrokerAsync()
                 _logger.Info(
                     LogMessages.ErrorReturnedInBrokerResponse(msalTokenResponse.Error));
 
-                if (msalTokenResponse.Error == BrokerResponseConst.NoTokenFound)
+                if (msalTokenResponse.Error == BrokerResponseConst.NoTokenFound ||
```
bgavrilMS, line 82: These error codes are Android specific and I think it would be better for the AndroidBroker to handle them, i.e. I would move this logic to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/93c3a6893477e0ee3ef3c0b63db94b8e1aca6e4c/src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs#L109

bgavrilMS, line 82: Could you prefix these constants with "Android"?

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
```diff
@@ -378,28 +378,68 @@ public void NoTokenFoundThrowsUIRequiredTest()
...
+                    return;
+                }
 
-                var response = new MsalTokenResponse
+                Assert.Fail("Wrong Exception thrown. ");
+            }
+        }
+
+        [TestMethod]
+        public void NoAccountFoundThrowsUIRequiredTest()
```
bgavrilMS, line 396: I was thinking that the best place to add this the logic you've added is actually in AndroidBroker.cs; However we can't test that. Having tests beats a marginally better architecture.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
bgavrilMS: It's legit :( Both Android broker and B2C miss the point of "error code" and send full blown messages :(

bgavrilMS (approved):
Minor comment about naming

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs
jennyf19: done.

---
In reply to: [526108373](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2200#discussion_r526108373) [](ancestors = 526108373)

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
jennyf19: +1 to the unit testing, especially since not sure how to manually test w/out messing up my authenticator app.

---
In reply to: [526108961](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2200#discussion_r526108961) [](ancestors = 526108961)

bgavrilMS (approved):

Issue #2083, opened 2020-10-07
[Documentation] Check aka.ms links


jennyf19:
As wiki info has moved to docs.ms some aka.ms links still point to the wiki, where the content has been removed. check & update aka.ms links

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs

jmprieur:
We can do a pass on MsalErrorMessages, but the problem is bigger than that: there are currently 6059 aka.ms links in 89 files in the MSAL.NET source code. 

This work needs to be spec-ed: how to do that efficiently and without causing more harm. Indeed, sometimes the wiki page was showing the end to end which is now split in the scenarios landing pages.

Also:
- P1: fix broken links
- P2: send to docs.ms when the information is there and the aka.ms link is used in only one context
- P2: update the links or the docs in docs.ms for more complex cases

jennyf19:
Docs.MS
- [x] https://aka.ms/msal-net-ios-13-broker
- [ ] https://aka.ms/msal-net-ios-broker
- [ ] https://aka.ms/msal-interactive-android
- [ ] https://aka.ms/msal-brokers
- [ ] http://aka.ms/valid-authorities
- [x] https://aka.ms/msal-net-enable-keychain-access
- [ ] https://aka.ms/msal-net-custom-web-ui
- [ ] https://aka.ms/msal-net-b2c
- [x] https://aka.ms/msal-net-signed-assertion

Goes to Wiki but content in Docs.
- [ ] https://aka.ms/msal-net-system-browsers
- [ ] https://aka.ms/Brokered-Authentication-for-Android

Goes to Wiki but content is in a sample? @jmprieur not sure what to do here:
- [ ] https://aka.ms/msal-net-application-configuration

Wiki
- [x]  https://aka.ms/msal-net-up
- [x] https://aka.ms/msal-statemismatcherror
- [X] https://aka.ms/msal-net-iwa
- [x] https://aka.ms/adal_token_cache_serialization
- [x] https://aka.ms/msal-client-apps
- [x] https://aka.ms/msal-net-client-credentials
- [x]  https://aka.ms/msal-net-3-breaking-changes
- [x] https://aka.ms/msal-net-device-code-flow
- [x] https://aka.ms/msal-net-3x-cache-breaking-change
- [x] https://aka.ms/msal-net-os-browser
[...]

PR #2202, opened 2020-11-18
spelling + spacing fixes to msal error messages

jennyf19:
done while starting #2083

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
+++ b/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
 using System.Globalization;
 
 namespace Microsoft.Identity.Client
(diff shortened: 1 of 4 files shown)


pmaytak (approved):

jmprieur (approved):
LGTM
Thanks @jennyf19.

Is, adding a space in the end, to make the logs clearer ?

bgavrilMS:
Looks like this breaks some unit tests. We should probably not assert on error messages, only on error codes.

jennyf19:
Yes. when all the logs are together it can be super hard to read. i'll maybe do another PR w/more fixes like that. 

---
In reply to: [533190058](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2202#issuecomment-533190058) [](ancestors = 533190058)

jennyf19:
I didn't even think to run the unit tests. lol. fixed.

---
In reply to: [729702492](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2202#issuecomment-729702492) [](ancestors = 729702492)

Issue #2207, opened 2020-11-18
[Internal] Device Auth tests are not running.


bgavrilMS:
See https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=666112&view=ms.vss-test-web.build-test-results-tab - not tests found

jennyf19:
@bgavrilMS the device auth tests seem to be passing now...maybe it was the build fix from earlier? idk. 
![image](https://user-images.githubusercontent.com/19942418/99588519-8dc57900-299f-11eb-844b-f30fc7c52810.png)
https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=666395&view=results

bgavrilMS:
They are passing because no tests are found.Test run takes 1s

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

bgavrilMS:
![image](https://user-images.githubusercontent.com/12273384/99592357-04866480-29e8-11eb-8c83-c5efd770cc78.png)

jennyf19:
oh i see.

trwalke:
ill take a look

jennyf19:
@trwalke i think i've got them working .i'll probably push up a PR tomorrow. have one more small issue to sort out w/publishing.

jennyf19:
@trwalke I'll leave these Todo's for you to follow up on, as you set up the VM w/the lab and are more familiar w/the feature. Thanks.
TODO:
- [ ] Check which version of VS is on the windows 8 VM
- [ ] Try running the test on the VM, as it's not running correctly currently, as noted in the build logs, but it runs locally. Maybe the VM is missing something.
It's failing before even starting the test:
![image](https://user-images.githubusercontent.com/19942418/100037009-256b0300-2db6-11eb-9c4f-8c7c365eb41a.png)
Thank you!

trwalke:
Made some progress on this. I am now able to get the test running on the agent by reconfiguring some of the parameters on the test task in YAML. There is an error to fix which I am investigating now but at least the test is actually being executed by ADO.

https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=707385&view=ms.vss-test-web.build-test-results-tab&runId=1422704&resultId=100001&paneView=debug

trwalke:
Tests are now passing. a little bit of clean up and ill submit pr very soon

PR #2235, opened 2020-11-24
fix device auth tests

jennyf19:
#2207 

TODO:
- Check which version of VS is on the windows 8 VM
- Try running the test on the VM, as it's not running correctly currently

When running locally, I now get a "this device is not compliant" error, which I think is correct, as I'm signing in with the win8 lab user on a win 10 device.

**Main fixes:**
- add `$` to file path
- Set `Microsoft.Identity.Test.Integration.Win8` to v4.6.2, as 4.7.2 is not supported on win8.
- Fix namespace (when going into properties in VS, it was showing a different default namespace)
- We need to use a specific lab user, but our lab queries are not set to filter by user. I just added the corrected user to the query, as there are no other parameters needed for this test. (before there was a mismatch in user -> password, as the default user response was returning).
- u/p does not work w/`common` so using `organizations` w/`.WithAuthority()`
- Add a reference to the MSTest.TestAdapter so the tests appear in the VS test runner
- due to newtonsoft error, the lab also needs to target net461, as 4.6.2 can only use newtonsoft v6 and lab is using v12, so there was an error popping up.


tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
```diff
@@ -21,42 +22,46 @@ public class DeviceAuthenticationTests
...
+                var (req, res) = factory.RequestsAndResponses
+                    .Where(x => x.Item1.RequestUri.AbsoluteUri == labResponse.Lab.Authority + "organizations/oauth2/v2.0/token"
+                             && x.Item2.StatusCode == HttpStatusCode.OK).ElementAt(1);
+
+                var AuthHeader = req.Headers.Single(h => h.Key == "Authorization").Value.FirstOrDefault();
 
+                Assert.IsTrue(!string.IsNullOrEmpty(AuthHeader));
+                Assert.IsTrue(AuthHeader.Contains("PKeyAuth"));
+            }
+            catch(Exception ex)
```
bgavrilMS, line 61: Is this needed? Exceptions fail the tests.

tests/Microsoft.Identity.Test.LabInfrastructure/Microsoft.Identity.Test.LabInfrastructure.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>net461; netstandard2.0</TargetFrameworks>
```
bgavrilMS, line 4: Could we keep .net471? According to the docs, https://docs.microsoft.com/en-us/dotnet/framework/install/on-windows-8-1, this is available for Win8.1 (not sure what this means, probably Win8 + Service Pack ?) but not for Win8

bgavrilMS (changes requested):
please remove try/catch assert and let's try to revert to .net471, as it is supported by win8.1 (not win8)

tests/Microsoft.Identity.Test.LabInfrastructure/Microsoft.Identity.Test.LabInfrastructure.csproj
jennyf19: I checked and the VM is 8.1, so i'll revert back on these.

tests/Microsoft.Identity.Test.Integration.Win8/DeviceAuthenticationTests.cs
jennyf19: it was for testing. have removed.

---
In reply to: [529523144](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2235#discussion_r529523144) [](ancestors = 529523144)

jennyf19:
done. thanks.

---
In reply to: [537466341](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2235#issuecomment-537466341) [](ancestors = 537466341)

bgavrilMS (approved):

Issue #2185, opened 2020-11-12
[Telemetry] Add telemetry that shows whether the implementation has implemented the token cache serialization or not.


henrik-me:

bgavrilMS:
Can be picked up by anybody?

neha-bhargava:
Fixed in [Msal release 4.24.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.24.0)

PR #2258, opened 2020-12-01
Nebharg/telemetry token cache serialized

neha-bhargava:
#2185


src/client/Microsoft.Identity.Client/ITokenCacheInternal.cs
```diff
@@ -75,6 +75,8 @@ internal interface ITokenCacheInternal : ITokenCache, ITokenCacheSerializer
         /// </summary>
         bool HasTokensNoLocks();
 
+        bool IsTokenCacheSerialized();
```
bgavrilMS, line 78: CC @pmaytak as you're also planning to add a similar method. Happy merging :)

bgavrilMS (approved):

pmaytak (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -29,8 +29,8 @@ public class UsernamePasswordIntegrationTests
 
         // Http Telemetry Constants
         private static Guid CorrelationId = new Guid("ad8c894a-557f-48c0-b045-c129590c344e");
-        private const string XClientCurrentTelemetryROPC = "2|1003,0|";
-        private const string XClientCurrentTelemetryROPCFailure = "2|1003,0|";
+        private const string XClientCurrentTelemetryROPC = "2|1003,0|0";
```
jennyf19, line 32: won't adding the new value first break pre-existing queries?

bgavrilMS: Good point, haven't caught that. New values should go at the end.

neha-bhargava: There are no existing queries for regional data currently. But good point to add new fields to end. I will create a PR to fix it. Thanks

Issue #2095, opened 2020-10-13
[Bug] Version 4.20.0 with dotnet sdk 5.0.100-rc.2.20479.15 results in NETSDK1130 error


fuocor:
**Which Version of MSAL are you using ?**
4.20.0

**Platform**
dotnet sdk 5.0.100-rc.2.20479.15

Other? - please describe;
using rc2 sdk and version 4.20.0 of MSAL results in:
`error NETSDK1130: Referencing a Windows Metadata component directly when targeting .NETCoreApp,Version=v5.0 is not supported. Use the C#/WinRT projection tool (https://aka.ms/cswinrt) or a provided projection for this target.`

**Is this a new or existing app?**
existing

**Expected behavior**
no build error

**Actual behavior**
crashes build

**Possible Solution**
revert to 4.0.19

jmprieur:
@fuocor 
This is now fixed in 4.20.1 released today
cc: @bgavrilMS

PR #2261, opened 2020-12-02
Perf improvements to AcquireTokenForClient flow

pmaytak:
Issue #2095 

Changes include:
- Update `GetAllAccessTokens` method to return `IEnumerable` instead of a `List`. 
- Use `IEnumerable.Where` instead of `FilterWithLogging`, in `ITokenCacheInternal.FindAccessTokenAsync`.
- Add checks if the serialization cache delegates were set, before calling them.

Added a BenchmarkDotNet performance test project to test the `AcquireTokenForClient`. Tested with cache size of 100, 1k, 10k, and 100k items. Also ran the test multiple times. There was some spread in mean data points between the runs (although I'm not sure why). (The spread was wider pre-fixes and much narrower post-fixes, except for .NET 4.8 which basically had no spread.) .NET 5 seems to be slightly faster than .NET Core 3.1 which is slightly faster than .NET 4.8. Below is the summary data.

**.NET Core 3.1.9** (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203)
| Items in cache | Mean (pre-fixes) | Mean (post-fixes) | Improvement |
|---------------:|--------------:|--------------:|--------------:|
|               100 |      71 - 91 μs |      64 - 73 μs | 10 - 20 %
|             1000 |     674 - 854 μs |     400 - 571 μs | 34 - 41 % 
|           10000 |   7.5 - 10 ms |   4.6 - 6.2 ms | 38 %
|         100000 | 122 - 176 ms | 83 - 117 ms | 32 - 34 %

Using in-memory token serialization cache yielded very similar results post-fix.
**With in-memory token serialization cache:**
| Items in cache | Mean (post-fixes) |
|---------------:|--------------:|
|               100 |      66 - 71 μs |
|             1000 |     407 - 530 μs |
|           10000 |  5.2 - 6 ms |
|         100000 | 85 - 114 ms |

**.NET Core 5.0.0** (CoreCLR 5.0.20.51904, CoreFX 5.0.20.51904)
| Items in cache | Mean (pre-fixes) | Mean (post-fixes) | Improvement |
|---------------:|--------------:|--------------:|--------------:|
|               100 |      67 - 89 μs |      54 - 62 μs | 20 - 31 %
|             1000 |     421 - 684 μs |     292 - 482 μs | 30 % 
|           10000 |   5.5 - 8.9 ms |   4.8 - 5.4 ms | 13 - 40 %
|         100000 | 111 - 165 ms | 69 - 98 ms | 38 - 41 %

**.NET Framework 4.8** (4.8.4250.0)
| Items in cache | Mean (pre-fixes) | Mean (post-fixes) | Improvement |
|---------------:|--------------:|--------------:|--------------:|
|               100 |      111 - 113 μs |      84 - 88 μs | 23 - 25 %
|             1000 |     593 - 614 μs |     515 - 531 μs | 14 % 
|           10000 |   12.3 ms |   7.3 ms | 41 %
|         100000 | 134 - 137 ms | 105 ms | 22 - 24 %

**Machine specs:**
``` ini
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-9850H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.100
```


src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
```diff
@@ -84,57 +84,67 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync()
...
+                                       _requestParams.Account,
+                                       hasStateChanged: false,
+                                       TokenCacheInternal.IsApplicationCache,
+                                       hasTokens: TokenCacheInternal.HasTokensNoLocks(),
+                                       suggestedCacheKey: key);
+                                    await TokenCacheInternal.OnBeforeAccessAsync(args).ConfigureAwait(false);
+                                }
+                                finally
+                                {
+                                    if (TokenCacheInternal.IsTokenCacheSerialized())
```
bgavrilMS, line 126: I don't think this value will change, no need to check twice.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -369,10 +351,8 @@ private static IEnumerable<MsalAccessTokenCacheItem> FilterByTokenType(Authentic
 
             if (filterByTenantId)
             {
-                tokenCacheItems = tokenCacheItems.FilterWithLogging(item =>
```
bgavrilMS, line 372: I was under the impression that developers need to set logging to VERBOSE for the filters to enumerate the underlying collection? I think we should simply tell people that in perf heavy apps they should not enable verbose logs (pretty standard ask) and leave the logging alone.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -448,23 +426,18 @@ private MsalAccessTokenCacheItem GetUnexpiredAccessTokenOrNull(AuthenticationReq
...
-            // if only one cached token found
-            if (filteredItems.Count() == 1)
+            try
             {
-                msalAccessTokenCacheItem = filteredItems.First();
-            }
-            else
+                return filteredItems.Single();
+            } 
+            catch (InvalidOperationException)
```
bgavrilMS, line 433: Let's not use exceptions for control flow. See https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -287,29 +295,9 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
...
-            if (logger.IsLoggingEnabled(LogLevel.Info))
-            {
-                logger.Info("Matching entry count - " + tokenCacheItems.Count());
-            }
-
             tokenCacheItems = FilterByScopes(requestParams, tokenCacheItems);
             tokenCacheItems = await FilterByEnvironmentAsync(requestParams, tokenCacheItems).ConfigureAwait(false);
 
-            // no match
-            if (!tokenCacheItems.Any())
```
bgavrilMS, line 307: You've introduced a bug here. In most cases, there will be 0 tokens remaining after all these filters. But instead of stopping, you check GetSingleResult, which will fail with an exception stating that there are 1+ tokens and MSAL cannot decide.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -769,35 +742,43 @@ async Task ITokenCacheInternal.RemoveAccountAsync(IAccount account, RequestConte
...
-                        account.HomeAccountId.Identifier);
+                    ITokenCacheInternal tokenCacheInternal = this;
 
                     try
                     {
-                        await (this as ITokenCacheInternal).OnBeforeAccessAsync(args).ConfigureAwait(false);
-                        await (this as ITokenCacheInternal).OnBeforeWriteAsync(args).ConfigureAwait(false);
-
-                        ((ITokenCacheInternal)this).RemoveMsalAccountWithNoLocks(account, requestContext);
+                        if (tokenCacheInternal.IsTokenCacheSerialized() || tokenCacheInternal.IsTokenCacheSerialized())
```
bgavrilMS, line 749: is this the same condition in the || ?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -769,35 +742,43 @@ async Task ITokenCacheInternal.RemoveAccountAsync(IAccount account, RequestConte
...
-                        this,
-                        ClientId,
-                        account,
-                        true,
-                        (this as ITokenCacheInternal).IsApplicationCache,
-                        (this as ITokenCacheInternal).HasTokensNoLocks(),
-                        account.HomeAccountId.Identifier);
+                    ITokenCacheInternal tokenCacheInternal = this;
 
                     try
```
bgavrilMS, line 747: Can you please review all these `try` blocks? They seem excessive. I can help.

tests/Microsoft.Identity.Client.Performance/GetAllAccessTokensTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Test.Common.Core.Mocks;
+using Microsoft.Identity.Test.Unit;
+
+namespace Microsoft.Identity.Test.Performance
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("AsyncUsage.CSharp.Usage", "UseConfigureAwait:Use ConfigureAwait", Justification = "Test project, inapplicable.")]
```
bgavrilMS, line 13: We tend to apply this rule in tests as well. It's good practice. You can always use `.ConfigureAwait(true)` which is the same as not using it.

tests/Microsoft.Identity.Client.Performance/GetAllAccessTokensTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class GetAllAccessTokensTests
```
bgavrilMS, line 14: Needs a better name, maybe `AcqruireTokenForClient_LargeCache` or smth

tests/Microsoft.Identity.Client.Performance/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+using BenchmarkDotNet.Jobs;
+using BenchmarkDotNet.Running;
+
+namespace Microsoft.Identity.Test.Performance
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            BenchmarkRunner.Run<GetAllAccessTokensTests>(
```
bgavrilMS, line 15: Looking forward to understanding better what the Benchmark project does.

henrik-me:
Would really like to see numbers for .net 4.6.1 to be able to compare with 4.8

henrik-me:
Wondering if we should add a hashmap to make lookup faster?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Good catch. This is why unit tests failed, but the error message was not accurate. 

If I add this check, could it happen that when line 307 executes, there are tokens in the cache, so we continue; then when we enter into `GetSingleResult`, another thread removes those tokens, and we end up with empty cache and the same error?

Similar point is that we call `.Any`, `.Count`, `.First` which apply all the predicates on the `IEnumerable` a few times in this search process. But other threads could update the cache in between these calls. Would it be better to have `IList<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true).ToList();` above, so that we query a stable list?

There would still be a general perf benefit, since `HasTokensNoLocks` would still get an `IEnumerable` from `GetAllAccessTokens`.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Combined the two outer try's. I think semaphore release should be in the separate outer try, in case there's exception in with OnAfterAccessAsync in the inner try.

tests/Microsoft.Identity.Test.Unit/CacheTests/TokenCacheNotificationTests.cs
```diff
@@ -269,5 +270,38 @@ await app
                 Assert.IsFalse(cacheAccessRecorder4.LastAfterAccessNotificationArgs.HasTokens);
             }
         }
+
+        [TestMethod]
+        public async Task TestNotSubscribedAsync()
```
pmaytak, line 275: @bgavrilMS I was thinking about a unit test something like this. Mock `AppTokenCacheInternal` and check if cache events are called. Adding serialization cache (L287-288) should break this test, but currently it does not. Mu guess is we need to add implementation to the mocked token cache.

Does this approach seem reasonable?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
jennyf19: @pmaytak in codeflow, the issues resolved via the Github UI don't appear as resolved, can you comment on the items from Bogdan as "done" or "fixed" so it's clear they've been addressed. thanks

---
In reply to: [534843438](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2261#discussion_r534843438) [](ancestors = 534843438)

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -19,6 +19,9 @@
 
 namespace Microsoft.Identity.Client
 {
+    /// <summary>
+    /// IMPORTANT: this class is perf critical; any changes must be benchmarked using Microsoft.Identity.Test.Performace
+    /// </summary>
     public sealed partial class TokenCache : ITokenCacheInternal
```
jennyf19, line 25: This is what JM was referring to, can we have an aka.ms link or something which directs us to how to run the tests, which information is needed before/after, and where to put it...etc... thank you.

bgavrilMS: We don't have a place to record historical info, @pmaytak recorded some data (see PR description). For a historical view, we need the same machine. We don't have a stable "golden config" machine. 

So the proposal is to just run the benchmark test (it's just a console app) manually - before any changes and after any changes.

tests/Microsoft.Identity.Test.Unit/CacheTests/TokenCacheNotificationTests.cs
bgavrilMS: Yes, it need a bit more mocking power, but it works now.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -446,25 +453,20 @@ private MsalAccessTokenCacheItem GetUnexpiredAccessTokenOrNull(AuthenticationReq
...
+        private static MsalAccessTokenCacheItem GetSingleResult(
+            AuthenticationRequestParameters requestParams,
+            IReadOnlyList<MsalAccessTokenCacheItem> filteredItems)
         {
-            MsalAccessTokenCacheItem msalAccessTokenCacheItem;
-
             // if only one cached token found
-            if (filteredItems.Count() == 1)
+            if (filteredItems.Count == 1)
             {
```
jennyf19, line 462: this is much better.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryTokenCacheAccessor.cs
```diff
@@ -154,62 +156,27 @@ public void DeleteAccount(MsalAccountCacheKey cacheKey)
...
         {
-            // perf: do not call ConcurrentDictionary.Values as it takes a lock
-            List<MsalAppMetadataCacheItem> metadata = new List<MsalAppMetadataCacheItem>();
-            foreach (var kvp in _appMetadataDictionary)
-            {
-                metadata.Add(kvp.Value);
-            }
-
-            return metadata;
+            return _appMetadataDictionary.Select(kv => kv.Value);
```
jennyf19, line 179: +1 to removing code.

tests/Microsoft.Identity.Test.Common/Core/Mocks/TokenCacheHelper.cs
```diff
@@ -15,22 +15,24 @@ internal class TokenCacheHelper
...
-            accessor.SaveAccessToken(atItem);
+            for (int i = 1; i <= tokensQuantity; i++)
+            {
+                var atItem = CreateAccessTokenItem(string.Format(System.Globalization.CultureInfo.InvariantCulture, TestConstants.ScopeStrFormat, i));
+                accessor.SaveAccessToken(atItem);
+            }
         }
 
-        internal static MsalAccessTokenCacheItem CreateAccessTokenItem()
+        internal static MsalAccessTokenCacheItem CreateAccessTokenItem(string scopes = "")
```
jennyf19, line 27: >"" [](start = 87, length = 2)

sting.Empty?

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
jennyf19: right, i understand that. but what JM is saying, is, for future reference, how do we run the performance runner? like someone new on the team, how do they run it? what are they looking for? We need to have this info somewhere, shouldn't be too hard to do. it's just a short doc on how to run it and what to look for, either in the wiki or in the onenote. we can add to it later, but this should be part of the PR, i think.

jennyf19: later, he can figure out how to handle the historical data, but that's separate.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Ah, ok. will do.
Done.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryTokenCacheAccessor.cs
```diff
@@ -154,62 +156,27 @@ public void DeleteAccount(MsalAccountCacheKey cacheKey)
...
         {
-            // perf: do not call ConcurrentDictionary.Values as it takes a lock
-            List<MsalRefreshTokenCacheItem> rts = new List<MsalRefreshTokenCacheItem>();
-            foreach (var kvp in _refreshTokenCacheDictionary)
-            {
-                rts.Add(kvp.Value);
-            }
-
-            return rts;
+            return _refreshTokenCacheDictionary.Select(kv => kv.Value).ToList();
```
pmaytak, line 164: @bgavrilMS As discussed, added `ToList` in these methods, since we need to investigate what the impact of returning an `IEnumerable` is.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Done.

src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
pmaytak: Done.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Done. Reverted back.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Done. Reverted back.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Done.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Done.

tests/Microsoft.Identity.Client.Performance/GetAllAccessTokensTests.cs
pmaytak: Done.

tests/Microsoft.Identity.Client.Performance/GetAllAccessTokensTests.cs
pmaytak: Done.

tests/Microsoft.Identity.Client.Performance/Program.cs
pmaytak: Done. Will add info here: https://aka.ms/msal-net-performance-testing.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Done.

tests/Microsoft.Identity.Test.Common/Core/Mocks/TokenCacheHelper.cs
pmaytak: Default parameter must be a compile-time constant. `string.Empty` is not.

PR #2264, opened 2020-12-02
change order of telem items

jennyf19:


neha-bhargava (approved):
Looks good Jenny, Thanks. I will wait for the checks to run and then merge it.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -127,15 +127,15 @@ public string GetCurrentRequestHeader(ApiEvent eventInProgress)
             var regionalFields = new StringBuilder();
             if (!string.IsNullOrEmpty(regionDiscovered))
```
bgavrilMS, line 128: It doesn't matter if regionDiscover is true or false. We should not have any `if` branches here. If a field is empty, we should still log it. Otherwise there is no way on the DB side of things to make queries. 

So we should log in our private fields: 

regionDiscovered, regionSource, isTokenCacheSerialized

Then when we make queries, we always know that at position 1 there is regionDiscovered, at position 2 there is regionSource and so on.

bgavrilMS: Same comment for `GetLastRequestHeader`

Issue #2263, opened 2020-12-02
[Bug] AcquireTokenOnBehalfOfAsync is still visible in the Intellisene


jmprieur:
**Which Version of MSAL are you using ?**
MSAL.NET 4.23.0

**Platform**
netcore

**Repro**
![image](https://user-images.githubusercontent.com/13203188/100887367-2fc38600-34b5-11eb-99d1-bf7450e0ac46.png)

**Possible Solution**
Apply the EditorBrowsable(false) attribute

neha-bhargava:
Fixed in [Msal release 4.24.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.24.0)

PR #2265, opened 2020-12-02
fix intellisense for oboasync

jennyf19:
#2263


jmprieur (approved):
LGTM.
Thanks for fixing this, @jennyf19 the IntelliSense is so much better, @bgavrilMS !

PR #2266, opened 2020-12-02
Fix telemetry, remove if block while adding telemetry logs

neha-bhargava:


src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -125,17 +125,13 @@ public string GetCurrentRequestHeader(ApiEvent eventInProgress)
...
             // Since regional fields will only be logged in case it is opted.
             var regionalFields = new StringBuilder();
-            if (!string.IsNullOrEmpty(regionDiscovered))
-            {
-                regionalFields.Append(regionDiscovered);
-                regionalFields.Append(",");
-                regionalFields.Append((regionSource));
-                regionalFields.Append(",");
-            }
+            regionalFields.Append(regionDiscovered);
```
bgavrilMS, line 128: Looks good, but can you please fix the `GetLastRequestHeader` as well?

neha-bhargava: Fixed that too.

jennyf19 (approved):
thanks @neha-bhargava

Issue #2189, opened 2020-11-16
[Bug] ArgumentNullException in RSACryptoServiceProvider under heavy load


alexzaitzev:
**Which Version of MSAL are you using ?**
Microsoft.Identity.Client 4.22.0 

**Platform**
net461

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App
    * [x] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Is this a new or existing app?**
The app was in production with the library `Microsoft.IdentityModel.Clients.ActiveDirectory 5.2.8`. I haven't upgraded MSAL, but sometimes started to observe an exception: `The parameter is incorrect. Exception type: CryptographicException`. I've updated to `Microsoft.Identity.Client 4.22.0` but still continue receiving exceptions with the similar stack trace: `Value cannot be null. Parameter name: keyBlob Exception type: ArgumentNullException`. 

**Repro**
It's hard to reproduce, issue began to appear when request rate was increased. It usually happens after swap production & staging slots during deployment in Azure `Cloud service (Classic) ` or after scaling out. Restarting problem VM fixes the problem.

This code initializes `KeyVaultClient(AuthenticationCallback authenticationCallback, params DelegatingHandler[] handlers)` in `AuthenticationCallback` delegate:
```csharp
private async Task<string> GetToken(string authority, string resource, string scope)
{
    IConfidentialClientApplication clientApplication = ConfidentialClientApplicationBuilder
        .Create(this.clientId)
        .WithAuthority(new Uri(authority), validateAuthority: true)
        .WithCertificate(this.clientCertificate)
        .Build();

    AuthenticationResult result = await clientApplication
        .AcquireTokenForClient(new[] { scope }) // exception throws here
        .ExecuteAsync()
        .ConfigureAwait(false);

    return result.AccessToken;
}
```

**Expected behavior**
Acquire authentication token without an exception.

**Actual behavior**
Sometimes exception is being thrown:
```
Value cannot be null.
Parameter name: keyBlob Exception type: ArgumentNullException 
Stack trace:
   at System.Security.Cryptography.RSACryptoServiceProvider.IsPublic(Byte[] keyBlob)
   at System.Security.Cryptography.X509Certificates.PublicKey.get_Key()
[...]

PR #2276, opened 2020-12-04
rsa mapping

bgavrilMS:
Issue #2189


src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -85,10 +92,21 @@ public byte[] SignWithCertificate(string message, X509Certificate2 certificate)
                 return rsa.SignData(messageBytes, sha);
             }
 #else
-            using (var key = certificate.GetRSAPrivateKey())
+            if (!CertificateToRsaMap.TryGetValue(certificate.Thumbprint, out RSA rsa))
```
bgavrilMS, line 95: I assume the same code needs to run on .NET Core ? I remember you saying it's not going to work on NetStandard1.3

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -13,11 +13,18 @@
...
+using System.Collections.Concurrent;
+using System.Threading;
 
 namespace Microsoft.Identity.Client.Platforms.net45
 {
     internal class NetDesktopCryptographyManager : ICryptographyManager
     {
+        private static ConcurrentDictionary<string, RSA> CertificateToRsaMap = new ConcurrentDictionary<string, RSA>();
+        private static ConcurrentDictionary<string, RSACryptoServiceProvider> CertificateToRsaCspMap = new ConcurrentDictionary<string, RSACryptoServiceProvider>();
+        private static int maximumCertificateToRsaMap = 1000;
```
bgavrilMS, line 25: Why does one dictionary have a max size while another does not?

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -13,11 +13,18 @@
...
 using Microsoft.Win32.SafeHandles;
 using Microsoft.Identity.Client.Platforms.net45.Native;
+using System.Collections.Concurrent;
+using System.Threading;
 
 namespace Microsoft.Identity.Client.Platforms.net45
 {
     internal class NetDesktopCryptographyManager : ICryptographyManager
     {
+        private static ConcurrentDictionary<string, RSA> CertificateToRsaMap = new ConcurrentDictionary<string, RSA>();
```
bgavrilMS, line 23: How large are these RSA  objects? Do we really need this limit?

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -85,10 +92,21 @@ public byte[] SignWithCertificate(string message, X509Certificate2 certificate)
                 return rsa.SignData(messageBytes, sha);
             }
 #else
-            using (var key = certificate.GetRSAPrivateKey())
+            if (!CertificateToRsaMap.TryGetValue(certificate.Thumbprint, out RSA rsa))
             {
-                return key.SignData(messageBytes, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+                if (CertificateToRsaMap.Count == maximumCertificateToRsaMap)
```
bgavrilMS, line 97: Could a thread bump the number of certs here to more than 1000 and skip this if? Probably safer if the condition were `>=`

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -13,11 +13,18 @@
...
+using System.Threading;
 
 namespace Microsoft.Identity.Client.Platforms.net45
 {
     internal class NetDesktopCryptographyManager : ICryptographyManager
     {
+        private static ConcurrentDictionary<string, RSA> CertificateToRsaMap = new ConcurrentDictionary<string, RSA>();
+        private static ConcurrentDictionary<string, RSACryptoServiceProvider> CertificateToRsaCspMap = new ConcurrentDictionary<string, RSACryptoServiceProvider>();
+        private static int maximumCertificateToRsaMap = 1000;
+        object clearMapLock = new object();
```
bgavrilMS, line 26: Shouldn't this lock also be static?

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -85,10 +92,21 @@ public byte[] SignWithCertificate(string message, X509Certificate2 certificate)
...
             }
 #else
-            using (var key = certificate.GetRSAPrivateKey())
+            if (!CertificateToRsaMap.TryGetValue(certificate.Thumbprint, out RSA rsa))
             {
-                return key.SignData(messageBytes, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+                if (CertificateToRsaMap.Count == maximumCertificateToRsaMap)
+                {
+                    ClearCertificateToRsaMap clearCertificateToRsaMap = new ClearCertificateToRsaMap(CertificateToRsaMap, clearMapLock);
+                    (new Thread(new ThreadStart(clearCertificateToRsaMap.Clear))).Start();
```
bgavrilMS, line 100: If we really need a ConcurrentDictionary with a max count, it would be better to create a separate class for it, so that we can unit test it easily. I'd think that using `GetOrAdd`  and `Clear` might simplify some of this logic.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
bgavrilMS, line 23: There is higher level caching of assertions in MSAL. I wonder why there is much contention here to create the bug. @trwalke added some caching.

henrik-me:
@brentschmaltz @GeoK @keegan-caruso : thoughts on @bgavrilMS comments to Brent's change?

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -143,7 +167,8 @@ private static RSACryptoServiceProvider GetCryptoProviderForSha256_Net45(X509Cer
                 // With this flag, a CryptographicException is thrown instead.
                 //
                 csp.Flags |= CspProviderFlags.UseExistingKey;
-                return new RSACryptoServiceProvider(csp);
+                CertificateToRsaCspMap[certificate.Thumbprint] = new RSACryptoServiceProvider(csp);
```
GeoK, line 170: Do we know if the RsaCryptoServiceProvider we just created is faulty (e.g. fails on Sign ops)?
If it is faulty, MSAL would throw each time when the RSA object is used.. until the pass-trough cache gets cleared.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
keegan-caruso: ConcurrentDictionary.GetOrAdd could simplify, but need to be careful, overload that takes a valueFactory doesn't have a guarantee that the value returned from the function will be inserted into the dictionary.

keegan-caruso, line 100: using a Task might be preferable here

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -202,5 +227,29 @@ internal static SafeCertContextHandle GetCertificateContext(X509Certificate cert
 
             return certContext;
         }
+
+        internal class ClearCertificateToRsaMap
+        {
+            ConcurrentDictionary<string, RSA> _rsaToCertificateMap;
```
keegan-caruso, line 233: these references should be readonly

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
pmaytak, line 26: Not used, can be removed, since using `ConcurrentDictionary.Clear()` instead.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: Oh, they should both have max size.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: When we designed other caches, we assumed that the count would be small. 
I turned out, as usual, some scenarios we hadn't thought of, filled the cache up with millions of object, services had to be restarted periodically.

This is sort of a fail safe.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: yep.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: @GeoK that is a good point, we should only add after a successful usage.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: This code pulled from a stale branch, disposing the RSA objects is fraught with issues.
Instead, just clear the dictionary.
If the number is reasonable, say 100 or 200, the memory pressure should be minimal.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: @pmaytak This code pulled from a stale branch, disposing the RSA objects is fraught with issues.
Instead, just clear the dictionary.
If the number is reasonable, say 100 or 200, the memory pressure should be minimal.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: @keegan-caruso This code pulled from a stale branch, disposing the RSA objects is fraught with issues.
Instead, just clear the dictionary.
If the number is reasonable, say 100 or 200, the memory pressure should be minimal.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: This code was pulled from a stale branch, disposing the RSA objects is fraught with issues.
Instead, just clear the dictionary.
If the number is reasonable, say 100 or 200, the memory pressure should be minimal.

henrik-me:
@brentschmaltz : expecting to address more comments? Can this be committed?
@bgavrilMS : thoughts?

henrik-me:
@brentschmaltz, what about netcore? any difference or is the same that needs to be done?

henrik-me:
@pmaytak : can you pls. help validate this fix for perf impact? Scenario is similar to that you tested recently for S2S tokens. Would be good to see before and after and with lots of tokens in the cache at the same time and lots of signing operations.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
brentschmaltz: I am not quite following....

brentschmaltz:
@henrik-me I have shared with @bgavrilMS a way to create certificates on the fly that will be used for creating a large number of certs.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
pmaytak: @bgavrilMS @brentschmaltz 
Should the same caching code be included in [NetCoreCryptographyManager](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/117256981411b9b0c017eb208f98c6259deaa52b/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCoreCryptographyManager.cs#L76) and [NetStandard13CryptographyManager](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/117256981411b9b0c017eb208f98c6259deaa52b/src/client/Microsoft.Identity.Client/Platforms/netstandard13/NetStandard13CryptographyManager.cs#L68)?

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
pmaytak: Updated. Resolved.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -79,16 +84,25 @@ public byte[] SignWithCertificate(string message, X509Certificate2 certificate)
...
 #if NET45
-            var rsa = GetCryptoProviderForSha256_Net45(certificate);
+            var rsaCryptoProvider = GetCryptoProviderForSha256_Net45(certificate);
             using (var sha = new SHA256Cng())
             {
-                return rsa.SignData(messageBytes, sha);
+                var signedData = rsaCryptoProvider.SignData(messageBytes, sha);
+                // Cache only valid RSA crypto providers
+                s_certificateToRsaCspMap[certificate.Thumbprint] = rsaCryptoProvider;
+                return signedData;
```
pmaytak, line 93: @GeoK @brentschmaltz 
Is this what you meant - to cache providers only when they can sign successfully?

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -79,16 +84,25 @@ public byte[] SignWithCertificate(string message, X509Certificate2 certificate)
...
             }
 #else
-            using (var key = certificate.GetRSAPrivateKey())
+            if (!s_certificateToRsaMap.TryGetValue(certificate.Thumbprint, out RSA rsa))
             {
-                return key.SignData(messageBytes, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
+                if (s_certificateToRsaMap.Count >= s_maximumMapSize)
+                    s_certificateToRsaMap.Clear();
+
+                s_certificateToRsaMap[certificate.Thumbprint] = certificate.GetRSAPrivateKey();
```
pmaytak, line 101: @GeoK @brentschmaltz
Do we need to check here if RSA can sign successfully before caching?

pmaytak:
@henrik-me 
I created a perf test [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2276/files#diff-9626b54f32887fed11bc3797e492b1e37ed4d8a52297e8527547c2a8f2b69095). It creates a confidential client app with certificate and a request builder, then benchmarks only the actual request.
Tested the signing code before and after fixes with 472 target, 48 runtime. Ran test for 1, 100, 990 apps/certs and for 1k and 20k iterations. 
Interestingly I didn't see any perf difference before and after fixes - all calls take ~300-330us.

@henrik-me @brentschmaltz @bgavrilMS 
Is there anything else that needs to be done for this work item?

henrik-me:
I think this is good. Power BI was looking for this fix, I think we should get it part of next build and work with them to get it tested in PROD, they can do safe rollout of the change

---
In reply to: [757096504](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2276#issuecomment-757096504) [](ancestors = 757096504)

henrik-me:
Also I think this change needs to be applied to netcore as well? 

---
In reply to: [757099735](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2276#issuecomment-757099735) [](ancestors = 757099735,757096504)

tests/Microsoft.Identity.Client.Performance/CryptoManagerTests.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        /// <summary>
+        /// Adds mocked HTTP response to the HTTP manager before each call.
+        /// Sets the index of the next app request to use.
+        /// </summary>
+        [IterationSetup]
+        public void IterationSetup()
+        {
+            _requestIdx = _requestIdx++ % AppsCount;
+            _httpManager.AddMockHandlerSuccessfulClientCredentialTokenResponseMessage();
```
pmaytak, line 60: @bgavrilMS Is there a way to setup the HttpManager to always return this mock response; so it doesn't have to be added before each request?

bgavrilMS: Not sure I understand, let's discuss offline.

bgavrilMS:
This looks good to me

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopCryptographyManager.cs
```diff
@@ -13,11 +13,16 @@
 using Microsoft.Identity.Client.Utils;
 using Microsoft.Win32.SafeHandles;
 using Microsoft.Identity.Client.Platforms.net45.Native;
+using System.Collections.Concurrent;
 
 namespace Microsoft.Identity.Client.Platforms.net45
 {
     internal class NetDesktopCryptographyManager : ICryptographyManager
     {
+        private static readonly ConcurrentDictionary<string, RSA> s_certificateToRsaMap = new ConcurrentDictionary<string, RSA>();
```
bgavrilMS, line 22: @pmaytak - please apply this fix to the other 2 crypto managers involved in confidential client apps: 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e917c1206b48d63fdc46cd8fe81ee20a9654ac4d/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCoreCryptographyManager.cs#L76 (you can also copy the test for this one)

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e917c1206b48d63fdc46cd8fe81ee20a9654ac4d/src/client/Microsoft.Identity.Client/Platforms/netstandard13/NetStandard13CryptographyManager.cs#L66

bgavrilMS: Or consolidate the common part of the code in a different file, somewhere in PlatformsCommon for example

tests/Microsoft.Identity.Client.Performance/CryptoManagerTests.cs
bgavrilMS: I don't think adding it before each request is such a bad solution.

Issue #2177, opened 2020-11-10
[Feature Request] Optimize and harden IMDS endpoint calling logic


neha-bhargava:
1. Instead of getting the json response from the imds endpoint for the compute node, there is an option to just get the region as text. Update the imds call to get the region as text.
2. Imds call does not need to retry if it fails. Update the HttpManager to add a method for imds calls.

jmprieur:
@neha-bhargava :I believe you've been working on this ?

neha-bhargava:
> @neha-bhargava :I believe you've been working on this ?

@jmprieur Just saw this comment. Yes I investigated this and will work on this.

pmaytak:
This is included in [MSAL 4.25.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0).

PR #2283, opened 2020-12-08
Optimize imds call

neha-bhargava:
#2177


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -76,6 +77,78 @@ protected virtual HttpClient GetHttpClient()
             return await ExecuteWithRetryAsync(uri, headers, body, HttpMethod.Post, logger, doNotThrow: true).ConfigureAwait(false);
         }
 
+        public async Task<HttpResponse> SendGetWithTimeoutAsync(
+            Uri endpoint,
+            IDictionary<string, string> headers,
+            ICoreLogger logger,
+            int timeoutInMs)
```
bgavrilMS, line 84: Use CancellationToken here instead of timeout.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -76,6 +77,78 @@ protected virtual HttpClient GetHttpClient()
             return await ExecuteWithRetryAsync(uri, headers, body, HttpMethod.Post, logger, doNotThrow: true).ConfigureAwait(false);
         }
 
+        public async Task<HttpResponse> SendGetWithTimeoutAsync(
```
bgavrilMS, line 80: Can you try to modify the existing methods to take in a CancellationToken instead ? There is too much duplicate code that you've written.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -76,6 +77,78 @@ protected virtual HttpClient GetHttpClient()
...
+                    MsalErrorMessage.HttpRequestUnsuccessful,
+                    (int)response.StatusCode, response.StatusCode));
+
+
+                is5xxError = (int)response.StatusCode >= 500 && (int)response.StatusCode < 600;
+            }
+            catch (TaskCanceledException exception)
+            {
+                logger.Error("The HTTP request failed or it was canceled. " + exception.Message);
+                throw new MsalServiceException(
```
bgavrilMS, line 111: It's better to let the cancellation exception propagate to the are of the code where you use it instead of catching and rethrowing a new exception. `MsalServiceExceptions` are signals for the user, not for your own internal logic.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -75,27 +76,19 @@ private async Task<string> GetRegionAsync(RequestContext requestContext)
                 {
                     { "Metadata", "true" }
                 };
-                
-                HttpResponse response = await _httpManager.SendGetAsync(BuildImdsUri(DefaultApiVersion), headers, logger).ConfigureAwait(false);
+
+                HttpResponse response = await _httpManager.SendGetWithTimeoutAsync(BuildImdsUri(DefaultApiVersion), headers, logger, TimeoutInMs).ConfigureAwait(false);
```
bgavrilMS, line 80: I don't understand what's happening here. I see that you've eliminated the retry and that's ok, but I don't get the whole timeout. If the 2 seconds run out, this method will throw. Is this expected? 

Also, the timeout logic is wrong. Every AcquireToken* methdo, including `AcquireTokenForClient` have the option of using a CancellationToken, which the user provides. See `ExecuteAsync(CancellationToken)`. 

So if the user sets a CancellationToken of 1 second, but your internal logic has a timeout of 2 seconds, you need to obey the user's timeout. This sounds complicated but luckly .NET has mechanism for this, see 

```csharp
 CancellationTokenSource.CreateLinkedTokenSource(
                    internalCancellationToken, 
                    externalCancellationToken);
```

bgavrilMS (changes requested):
Too much code duplication in HttpManager
Cancellation logic needs to take into account user's CancellationToken

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Discussed offline and updated

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -19,16 +20,21 @@ internal sealed class RegionDiscoveryProvider : IRegionDiscoveryProvider
         private const string RegionName = "REGION_NAME";
 
         // For information of the current api-version refer: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#versioning
-        private const string ImdsEndpoint = "http://169.254.169.254/metadata/instance/compute";
+        private const string ImdsEndpoint = "http://169.254.169.254/metadata/instance/compute/location";
         private const string DefaultApiVersion = "2020-06-01";
 
         private readonly IHttpManager _httpManager;
         private readonly INetworkCacheMetadataProvider _networkCacheMetadataProvider;
+        private readonly int _imdsCallTimeout;
```
pmaytak, line 28: I would specify time `_imdsCallTimeoutInMs`.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -75,47 +81,52 @@ private async Task<string> GetRegionAsync(RequestContext requestContext)
...
+                    MsalError.RegionDiscoveryFailed,
+                    MsalErrorMessage.RegionDiscoveryFailed,
+                    response);
+            }  
+            catch (MsalServiceException e)
             {
-                throw;
+                if (MsalError.RequestTimeout.Equals(e.ErrorCode))
+                {
+                    throw new MsalServiceException("region_discovery_failed", "app cancelled or timeout expired", e);
```
pmaytak, line 110: These should be `MsalError` constants?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -75,47 +81,52 @@ private async Task<string> GetRegionAsync(RequestContext requestContext)
...
-                throw new MsalServiceException(MsalError.RegionDiscoveryFailed, MsalErrorMessage.RegionDiscoveryFailed);
+                logger.Info("[Region discovery] Call to local imds failed. " + e);
+                throw new MsalServiceException(MsalError.RegionDiscoveryFailed, MsalErrorMessage.RegionDiscoveryFailed, e);
             }
         }
 
+        private CancellationToken GetCancellationToken(CancellationToken userCancellationToken)
+        {
+            CancellationTokenSource tokenSource = new CancellationTokenSource(_imdsCallTimeout);
+            CancellationTokenSource linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(userCancellationToken);
```
pmaytak, line 125:
```suggestion
            CancellationTokenSource linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(userCancellationToken, tokenSource.Token);
```

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -69,24 +76,100 @@ public async Task SuccessfulResponseFromEnvironmentVariableAsync ()
...
+                new NetworkCacheMetadataProvider());
+
+            _userCancellationTokenSource.Cancel();
+
+            var ex = await AssertException.TaskThrowsAsync<MsalServiceException>(() => regionDiscoveryProvider.GetMetadataAsync(
+                new Uri("https://login.microsoftonline.com/common/"),
+                _testRequestContext))
+                .ConfigureAwait(false);
+
+            Assert.AreEqual("region_discovery_failed", ex.ErrorCode);
```
pmaytak, line 152: Use `MsalError` constant?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -207,6 +141,12 @@ private HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string
             {
                 logger.Error("The HTTP request failed or it was canceled. " + exception.Message);
                 isRetryable = true;
+
+                if (cancellationToken.IsCancellationRequested)
```
bgavrilMS, line 145: nice

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
jennyf19: yes. and MsalErrorMessage. nice catch

---
In reply to: [541345262](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2283#discussion_r541345262) [](ancestors = 541345262)

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -215,27 +294,28 @@ private void AddMockedResponse(HttpResponseMessage responseMessage, string apiVe
...
             {
                 queryParams.Add("api-version", apiVersion);
+                queryParams.Add("format", "text");
 
                 _httpManager.AddMockHandler(
                    new MockHttpMessageHandler
                    {
                        ExpectedMethod = HttpMethod.Get,
-                       ExpectedUrl = "http://169.254.169.254/metadata/instance/compute",
+                       ExpectedUrl = "http://169.254.169.254/metadata/instance/compute/location",
```
jennyf19, line 303: should this be a constant value? Then we don't have to update everywhere if there are future changes.

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Updated.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Good point. Updated.

pmaytak (approved):
👍

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -298,6 +298,7 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
         public const string SSHCertUsedAsHttpHeader = "MSAL was configured to request SSH certificates from AAD, and these cannot be used as an HTTP authentication header. Developers are responsible for transporting the SSH certificates to the target machines. ";
         public const string BrokerApplicationRequired = "Installation of broker failed. The broker application must be installed to continue authentication. ";
         public const string RegionDiscoveryFailed = "Region discovery for the instance failed. Region discovery can only be made if the service resides in Azure function or Azure VM. See https://aka.ms/msal-net-region-discovery for more details. ";
+        public const string RegionDiscoveryFailedWithTimeout = "Region discovery failed due to app cancellation or timeout.";
```
jennyf19, line 301:
```suggestion
        public const string RegionDiscoveryFailedWithTimeout = "Region discovery failed due to app cancellation or timeout. ";
```

jennyf19: makes the logs easier to read :)

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -75,47 +81,52 @@ private async Task<string> GetRegionAsync(RequestContext requestContext)
...
+                    throw new MsalServiceException(MsalError.RegionDiscoveryFailed, MsalErrorMessage.RegionDiscoveryFailedWithTimeout, e);
+                }
+
+                throw e;
             }
             catch (Exception e)
             {
-                logger.Info("[Region discovery] Call to local imds failed." + e.Message);
-                throw new MsalServiceException(MsalError.RegionDiscoveryFailed, MsalErrorMessage.RegionDiscoveryFailed);
+                logger.Info("[Region discovery] Call to local imds failed. " + e);
```
jennyf19, line 117: Actually this should go through the Pii logs, as the exception could have pii, as we don't know what's in it.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
neha-bhargava: Added as a constant.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: This exception would mostly be either timeout exception or any other exception where we were unable to get a response from imds. There is no Pii data involved here.

Issue #2278, opened 2020-12-05
[Fundamentals] Ensure all tests runs on a release build including the UI tests


jennyf19:
**Goal:** It must be simple to answer the question, which tests have run as part of the bits we are shipping, and are the tests passing. Additionally it must be easy to detect if any tests which is supposed to run as part of the test sheet was run or not.

- [ ] Add unit test/integration tests to release yaml. 
- [ ] Ensure the UI tests are started with a release build
- [ ] Document which tests have been ignored 
- [ ] For a release ensure that it's easy to point to the overview of all test runs in the test sheet (new row added to the test sheet)

bgavrilMS:
I removed the unit tests and integration tests from the release build, because we already run them as part of CI build and they take a long time and are flaky. I've also noticed that sometimes we include some tests in CI build but forget to include them in the Release build (i.e. need to refactor the templates etc.), for example the cache compat tests never ran on release builds.

I'm ok with running tests on the Release build, but I argue this is not a P1. Today, a release build must be associated with a passing CI build.

trwalke:
I know I have needed to run a quick release build in the past to create a preview package for a customer so it is nice to have the build be quick. I think the best of both worlds is to have the release build trigger a set of test runs from our CI and UI Auto pipelines.

henrik-me:
It's was not easy to validate that all tests had run. These tests needs to run as part of the release build. Happy to get more details on what is flaky. Moving back to P1. It took me almost 2 hours to get the information I needed before being able to give sign-off and still I wasn't sure everything had run on the actual bits we shipped.

jmprieur:
We could also have an environment variable (settable in the build properties) to skip the test in case you want to have  a quick build?, but the tests would run by default?

bgavrilMS:
A setting (not necessarily env variable) would be my preference as well. We also to reorganize the existing templates a bit to ensure all tests run on both builds with minimal yaml duplication.

pmaytak:
This is included in [MSAL 4.25.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0).

PR #2286, opened 2020-12-08
Adding test runs to release build

trwalke:
Making changes to pipeline to allow tests to run on release build based on the value of RunTests.

RunTests can be set in the beginning of the test run.

true to run tests, false not to.

This will be true by default so tests will always run.

This will also trigger ui automation on the master branch

Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2278

### DIFF
--- a/build/pipeline-releasebuild.yaml
+++ b/build/pipeline-releasebuild.yaml
@@ -46,7 +46,7 @@ steps:
     BuildPlatform: '$(BuildPlatform)'
     BuildConfiguration: '$(BuildConfiguration)'
     MsalClientSemVer: $(MsalClientSemVer)
-    Solution: 'Libs.sln'
+    Solution: 'LibsAndSamples.sln'
 
 # Run Unit Tests
 - template: template-run-unit-tests.yaml
--- a/build/pipeline-releasebuild.yaml
+++ b/build/pipeline-releasebuild.yaml
@@ -63,13 +63,13 @@ steps:
 - template: template-run-unit-tests.yaml
   parameters:
     BuildConfiguration: '$(BuildConfiguration)'
-    condition: and(succeeded(), eq(variables['RunTests'], 'true'))
+    condition: eq(variables['RunTests'], 'true')
 
 # Run Integration Tests
 - template: template-run-integration-tests.yaml
   parameters:
     BuildConfiguration: '$(BuildConfiguration)'
-    condition: and(succeeded(), eq(variables['RunTests'], 'true'))
+    condition: eq(variables['RunTests'], 'true')
 
 # Publish nuget packages and symbols to VSTS package manager.
 - template: template-publish-packages-and-symbols.yaml
--- a/build/pipeline-releasebuild.yaml
+++ b/build/pipeline-releasebuild.yaml
@@ -64,13 +64,11 @@ steps:
 - template: template-run-unit-tests.yaml
   parameters:
     BuildConfiguration: '$(BuildConfiguration)'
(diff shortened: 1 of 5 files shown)


build/pipeline-releasebuild.yaml
```diff
@@ -46,14 +47,29 @@ steps:
...
-    Solution: 'Libs.sln'
+    Solution: 'LibsAndSamples.sln'
 
 # Run Post-build code analysis (e.g. Roslyn)
 - template: template-postbuild-code-analysis.yaml
 
 # Pack and sign all of the nuget packages
 - template: template-pack-and-sign-all-nugets.yaml
 
+- task: UseDotNet@2
```
bgavrilMS, line 58: Shouldn't this run first?

build/pipeline-releasebuild.yaml
```diff
@@ -46,14 +47,29 @@ steps:
...
+  displayName: 'Use .Net Core SDK 3.x'
+  inputs:
+    version: 3.x
+
+# Run Unit Tests
```
bgavrilMS, line 63: Could you add a template that runs all tests? I don't think we ever run just unit tests or just integration tests. Should include cache compat tests as well if not included in integration tests already.

build/pipeline-releasebuild.yaml
```diff
@@ -46,14 +47,29 @@ steps:
...
+- template: template-run-unit-tests.yaml
+  parameters:
+    BuildConfiguration: '$(BuildConfiguration)'
+
+# Run Integration Tests
+- template: template-run-integration-tests.yaml
```
bgavrilMS, line 69: What about UI automation?

jennyf19: The Win8 tests are not working atm, but should be included.

build/pipeline-releasebuild.yaml
trwalke: I was under the impression that they are already there. Ill double check

build/pipeline-releasebuild.yaml
trwalke: Win 8 and UI auto will be in a separate build that is automatically triggered when this runs. The trigger settings do not show up in the PR but they are visible in the pipeline settings

build/pipeline-releasebuild.yaml
trwalke: ill reconfigure the test template

build/pipeline-releasebuild.yaml
jennyf19: right, but the device auth tests are not running in the VM. there is a separate issue for that.

trwalke: Understood. I am working on fixing it

build/pipeline-releasebuild.yaml
trwalke: There was a note that I read that stated the signing tasks cannot run on .core 3 only 2, so I reset the version to 3 here after the signing. I am going to refactor all of these into a template so this will be removed

bgavrilMS: That makes sense, I think @jennyf19 also encountered this problem.

jennyf19: yes. it has to be version 2.x, you guys have it right, we had 2.0, and the error said to use 2.1, which doesn't work, but 2.x does.

trwalke:
Test run with disabled tests
image

bgavrilMS (approved):

PR #2306, opened 2020-12-15
Move WAM into separate package (.net fwk only)

bgavrilMS:


src/client/Microsoft.Identity.Client.Desktop/Microsoft.Identity.Client.Desktop.csproj
```diff
@@ -0,0 +1,71 @@
...
+    <TargetFrameworkNetDesktop>net462</TargetFrameworkNetDesktop>
+    <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
+    <TargetFrameworkNetCore>netcoreapp3.1</TargetFrameworkNetCore>
+    <TargetFrameworks>$(TargetFrameworkNetDesktop);$(TargetFrameworkNetStandard);$(TargetFrameworkNetCore)</TargetFrameworks>
+
+    <PathToMsalSources>$(MSBuildThisFileDirectory)../Microsoft.Identity.Client/</PathToMsalSources>
+
+  </PropertyGroup>
+
+  <PropertyGroup Label="NuGet and AssemblyInfo metadata">
```
bgavrilMS, line 13: @henrik-me  @jmprieur please review the new package description

src/client/Microsoft.Identity.Client.Desktop/Microsoft.Identity.Client.Desktop.csproj
```diff
@@ -0,0 +1,71 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+
+    <TargetFrameworkNetDesktop>net462</TargetFrameworkNetDesktop>
```
bgavrilMS, line 4: We could go with net461 as well, but I figured that we'll likely move WebView2 in here as well and WebView2 requires net462

henrik-me: Most will likely ask us for 4.6.1, however given 462 is a serving release of 4.6 I suppose this will be acceptable for most, and more than likely the runtime will be higher on most machines. 

Any reason to not build a netcore package?

---
In reply to: [543456605](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2306#discussion_r543456605) [](ancestors = 543456605)

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -120,20 +120,33 @@ public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keycha
         }
 
         /// <summary>
-        /// On Android and iOS, brokers enable Single-Sign-On, device identification,
+        /// Brokers enable Single-Sign-On, device identification,
         /// and application identification verification. To enable one of these features,
         /// you need to set the WithBroker() parameters to true. See https://aka.ms/msal-net-brokers 
         /// for more information on platform specific settings required to enable the broker.
+        /// 
+        /// On iOS and Android, Autheticator and Company Portal serve as brokers.
```
henrik-me, line 128: >Autheticator [](start = 32, length = 12)

nit:  naming

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -120,20 +120,33 @@ public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keycha
...
+#if NET461 || NET_CORE
+            if (Config.BrokerCreatorFunc == null)
+            {
+                throw new PlatformNotSupportedException(
+                    "The Windows broker is not directly available on MSAL for .NET Framework or .NET Core 3.x" +
+                    " To use it, please install the nuget package named Microsoft.Identity.Client.Desktop " +
+                    "and call the extension method .WithWindowsBroker(true)");
+            }
+#endif
+#if NET461 || WINDOWS_APP || NET5_WIN
```
henrik-me, line 149: Perhaps we should have a DESKTOP_BROKER flag?

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -120,20 +120,33 @@ public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keycha
...
-        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]  // hide confidential client on mobile
-#endif
         public PublicClientApplicationBuilder WithBroker(bool enableBroker = true)
         {
-#if DESKTOP || WINDOWS_APP || NET5_WIN
+#if NET45
+            throw new PlatformNotSupportedException("The Windows broker is not available on .NET Framework 4.5, please use at least .NET Framework 4.6.2");
+#endif
+
+#if NET461 || NET_CORE
```
henrik-me, line 140: Shouldn't this be specified with all versions of Windows Desktop broker? Same flag as below?

src/client/Microsoft.Identity.Client.Desktop/Microsoft.Identity.Client.Desktop.csproj
bgavrilMS: NetCore 3.1 and NetStandard 2.0 targets are defined on lines 5 and 6

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
bgavrilMS: The main difference between NET fwk and NET Core is the WinForms dependency, which defines the type `IWin32Window`. That's why all these #if flags are needed.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -239,7 +254,9 @@ public PublicClientApplicationBuilder WithParentActivityOrWindow(Func<IWin32Wind
 
             return WithParentFunc(() => (object)windowFunc());
         }
+#endif
 
+#if DESKTOP || NET5_WIN || NET_CORE
```
henrik-me, line 259: Is this  only applicable on Windows desktop? Perhaps we should have a flag for that?

henrik-me:
still reviewing.....

src/client/Microsoft.Identity.Client.Desktop/WamExtension.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            if (!builder.Config.ExperimentalFeaturesEnabled)
+            {
+                throw new MsalClientException(
+                    MsalError.ExperimentalFeature,
+                    MsalErrorMessage.ExperimentalFeature(nameof(WithWindowsBroker)));
+            }
+            builder.Config.IsBrokerEnabled = enableBroker;
+
+#if !NETSTANDARD
```
henrik-me, line 28: >NETSTANDARD [](start = 5, length = 11)

did we want to start having netstandard behave like netcore?

henrik-me:
>             }

Doesn't seem like a great thing for customers to set. Should this be a property to flip instead or will this go away with passthrough improvements in MSAL? 

---
Refers to: tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs:76 in 268bdd6. [](commit_id = 268bdd6fa0af827bb215c0dd383a4b3d52d2df19, deletion_comment = False)

tests/devapps/WAM/NetFrameworkWam/Form1.cs
```diff
@@ -75,7 +76,12 @@ private IPublicClientApplication CreatePca()
                 .Create(clientId)
                 .WithAuthority(this.authorityCbx.Text)
                 .WithExperimentalFeatures(true)
-                .WithBroker(this.useBrokerChk.Checked)
+
+                //.WithWindowsBroker(true)
```
henrik-me, line 80: >//.WithWindowsBroker(true) [](start = 16, length = 26)

remove

tests/devapps/WAM/NetFrameworkWam/Form1.cs
```diff
@@ -75,7 +76,12 @@ private IPublicClientApplication CreatePca()
...
+                .WithWindowsBroker(true)
+
+                //.WithBroker(true)
```
henrik-me, line 83: >//.WithBroker(true) [](start = 16, length = 19)

naming: would have been better if we could stick to this terminology, what will we do on Linux/Mac when we have a broker there?
Seems unfortunate that we can't have both names the same?

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
```diff
@@ -78,7 +79,8 @@ private IPublicClientApplication CreatePca()
                 .Create(clientId)
                 .WithAuthority(this.authorityCbx.Text)
                 .WithExperimentalFeatures(true)
-                .WithBroker(this.useBrokerChk.Checked)
+                //.WithBroker(this.useBrokerChk.Checked)
+                .WithWindowsBroker(this.useBrokerChk.Checked)
```
henrik-me, line 83: See naming conversation in the netcore ....

src/client/Microsoft.Identity.Client.Desktop/WamExtension.cs
bgavrilMS: We still support only netstandard 1.3 which does not support COM / WAM

tests/devapps/WAM/NetFrameworkWam/Form1.cs
bgavrilMS: We should discuss about this in more detail, I only discussed it with @jmprieur . There are 2 aspects: 

1. discoverability of API - if we hide `WithBroker` in MSAL and only make it available on the extension, it is harder to discover. It would also be a (small) breaking change, as we have WithBroker API today, even if it simply throws.
2. netstandard support - from a netstandard perspective, we already have WithBroker (to support Xamarin). 

It's an interesting question this - how will we support Mac and Linux brokers? I imagine we'll only be able to PInvoke into such brokers, so they will live in MSAL package. 



```csharp
// net5 and UWP
.WithBroker(true);

// net core 3.x and net classic
.WithWindowsBroker(true)
.WithBroker(true)
```

jmprieur:
@bgavrilMS 
For the discoverability of features I propose (as discussed with @henrik-me), that we build Roslyn analyzers to give advices based on the context. This would work here, and also for disabling the ADAL cache, and more (the redirect URI)
I'll write a 1 pager with this proposal.

bgavrilMS:
Roslyn analyzer would be amazing. I imagine we can install one with MSAL automatically.

jmprieur:
yes absolutely. They were designed to be companions of libraries.

bgavrilMS:
@henrik-me - .WithBroker already exists today on every platform, so we cannot remove it. If I define a `WithBroker` extension method, VS automatically picks the non-extension method, so a developer would have to fully qualify the extension method to get it work. I don't like it...

image

bgavrilMS:
@henrik-me - please let me know if you have any more concerns here. This is still "exprimental" so we can change things later. I'll fix the small naming issues before commiting.

henrik-me:
Same concern applies, but lets get this shipped and get some developer feedback. I still don't like Windows in the naming (perhaps we should call it WithWAM to be explicit about the usage reason).

For discoverability, I recommend also updating the nuget description to mention the package, and what it's needed for. E.g.: Please see also, xyz package for UI and Windows broker (WAM) extensions.

---
In reply to: [758711196](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2306#issuecomment-758711196) [](ancestors = 758711196)

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
```diff
@@ -78,7 +79,8 @@ private IPublicClientApplication CreatePca()
...
                 // there is no need to construct the PCA with this redirect URI, 
                 // but WAM uses it. We could enforce it.
                 .WithRedirectUri($"ms-appx-web://microsoft.aad.brokerplugin/{clientId}")
```
henrik-me, line 86: > .WithRedirectUri($"ms-appx-web://microsoft.aad.brokerplugin/{clientId}") [](start = 15, length = 73)

are we doing this by default with WAM and uwp?, I think we should

henrik-me:
> using System;

thought: there is a lot of shared code in the two test apps, perhaps a shared project between the two (then we can also verify the netstandard behavior?

---
Refers to: tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs:1 in b4ace2b. [](commit_id = b4ace2b56c1843874e0a5633ddc099ef6a981d18, deletion_comment = False)

tests/Microsoft.Identity.Test.Unit/CoreTests/PlatformProxyFactoryTests.cs
```diff
@@ -153,12 +153,9 @@ public void FlagDefaultsTest()
         {
             var proxy = PlatformProxyFactory.CreatePlatformProxy(null);
 
-#if SUPPORTS_BROKER
+#if NET5_WIN
             Assert.IsTrue(proxy.BrokerSupportsWamAccounts);
             Assert.IsTrue(proxy.CanBrokerSupportSilentAuth());
```
henrik-me, line 158: not sure why these are not available for all supported platforms?

henrik-me (approved):
:shipit:

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
bgavrilMS: This itnerface is part of WinForms, and we don't have a dependency on WinForms from NET_CORE

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
bgavrilMS: No, we decided to keep the old model where the redirect uri is used by fallback browser (e.g. in case same code runs on Linux)

Issue #2260, opened 2020-12-01
[Bug] Once the region discovery is done calls with WithAzureRegion set to false should not hit the regional token endpoint


neha-bhargava:
**Which Version of MSAL are you using ?**

Latest

**Repro**

```csharp
// 1st call with regional auth set to true
await cca.AcquireTokenForClient(scopes) .WithAzureRegion(true);

// 2nd call to global
await cca.AcquireTokenForClient(scopes) .WithAzureRegion(false);
```

**Expected behavior**
The 2nd call should go to global token endpoint.

**Actual behavior**
The 2nd call is going to regional token endpoint.

**Possible Solution**
This is happening because we are caching the network metadata with preferred network as the regional endpoint once the region is discovered for both the original authority and the authority created with the region.

Either cache the regional metadata separate, or cache it only for the regional authority.

neha-bhargava:
Marking it as P1 since the network metadata cache is a static cache and once the region is discovered we will be hitting the regional endpoint irrespective of the value for WithAzureRegion. This could cause problems if someone wants to fallback to global in case ests-r endpoint fails or if someone wants to rollout traffic to ests-r in phases.

jmprieur:
Reopenining the issue closed by the bot (until we release it)

pmaytak:
This is included in [MSAL 4.25.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0).

PR #2308, opened 2020-12-15
remove original authority from metadata aliases

neha-bhargava:
Proposed fix for issue https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2260

### DIFF
--- a/src/client/Microsoft.Identity.Client/Region/IRegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/IRegionDiscoveryProvider.cs
@@ -15,6 +15,6 @@ namespace Microsoft.Identity.Client.Region
     {
         Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority, RequestContext requestContext);
 
-        void clearStaticCache();
+        void Clear();
     }
 }
--- a/src/client/Microsoft.Identity.Client/Region/IRegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/IRegionDiscoveryProvider.cs
@@ -14,5 +14,7 @@ namespace Microsoft.Identity.Client.Region
     internal interface IRegionDiscoveryProvider
     {
         Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority, RequestContext requestContext);
+
+        void clearStaticCache();
     }
 }
--- a/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Identity.Client.Region
         /*
          * To be used only for tests
          */
-        public void clearStaticCache()
+        public void Clear()
         {
             s_region = null;
         }
--- a/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
@@ -50,11 +50,11 @@ namespace Microsoft.Identity.Client.Region
         {
             ICoreLogger logger = requestContext.Logger;
             string environment = authority.Host;
-            InstanceDiscoveryMetadataEntry cachedEntry = _networkCacheMetadataProvider.GetMetadata(environment, logger);
+            Uri regionalizedAuthority = await BuildAuthorityWithRegionAsync(authority, requestContext).ConfigureAwait(false);
+            InstanceDiscoveryMetadataEntry cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, logger);
 
             if (cachedEntry == null)
             {
-                Uri regionalizedAuthority = await BuildAuthorityWithRegionAsync(authority, requestContext).ConfigureAwait(false);
                 CacheInstanceDiscoveryMetadata(CreateEntry(authority, regionalizedAuthority));
 
                 cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, logger);
@@ -68,7 +68,6 @@ namespace Microsoft.Identity.Client.Region
             
             return cachedEntry;
         }
-
 
         private async Task<string> GetRegionAsync(RequestContext requestContext)
         {
--- a/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
@@ -48,21 +48,19 @@ namespace Microsoft.Identity.Client.Region
 
         public async Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority, RequestContext requestContext)
         {
-            ICoreLogger logger = requestContext.Logger;
-            string environment = authority.Host;
             Uri regionalizedAuthority = await BuildAuthorityWithRegionAsync(authority, requestContext).ConfigureAwait(false);
-            InstanceDiscoveryMetadataEntry cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, logger);
+            InstanceDiscoveryMetadataEntry cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, requestContext.Logger);
 
             if (cachedEntry == null)
             {
                 CacheInstanceDiscoveryMetadata(CreateEntry(authority, regionalizedAuthority));
 
-                cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, logger);
-                logger.Verbose($"[Region Discovery] Created metadata for the regional environment {environment} ? {cachedEntry != null}");
+                cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, requestContext.Logger);
+                requestContext.Logger.Verbose($"[Region Discovery] Created metadata for the regional environment {regionalizedAuthority.Host} ? {cachedEntry != null}");
             }
             else
             {
-                logger.Verbose($"[Region Discovery] The network provider found an entry for {environment}");
+                requestContext.Logger.Verbose($"[Region Discovery] The network provider found an entry for {regionalizedAuthority.Host}");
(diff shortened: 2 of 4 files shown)


src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -37,6 +38,14 @@ internal sealed class RegionDiscoveryProvider : IRegionDiscoveryProvider
             _imdsCallTimeoutMs = imdsCallTimeout;
         }
 
+        /*
+         * To be used only for tests
+         */
+        public void clearStaticCache()
+        {
```
jennyf19, line 45: is there a reason to not implement the `INetworkCacheMetadataProvider`, which also has `Clear()`...then you can do `s_region.Clear()`, to make sure the cache is cleared. right now, it doesn't seem to be actually cleared. also, not sure why to pass the request context, when you have the logger, you can just pass the logger instead.

henrik-me:
I'm not seeing a test that validates the scenario which is failing. Am I missing something?

src/client/Microsoft.Identity.Client/Region/IRegionDiscoveryProvider.cs
```diff
@@ -14,5 +14,7 @@ namespace Microsoft.Identity.Client.Region
     internal interface IRegionDiscoveryProvider
     {
         Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority, RequestContext requestContext);
+
+        void clearStaticCache();
     }
```
trwalke, line 19: nit: uppercase C

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -148,6 +144,7 @@ public async Task NoImdsCancellation_UserCancelled_Async()
                 .ConfigureAwait(false);
 
             Assert.AreEqual(MsalError.RegionDiscoveryFailed, ex.ErrorCode);
+            regionDiscoveryProvider.clearStaticCache();
```
trwalke, line 147: Should there be some additional validation to confirm that the cache is clear?
Is there a separate test for that?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: @jennyf19 I am not sure what you mean when you say implement `INetworkCacheMetadataProvider`. I have an instance of `INetworkCacheMetadataProvider` in this class. Do you mean adding a static variable there to store the region and use it there?

Also, I am passing request context since it is passed on to GetRegionAsync method further to log telemetry data and get the cancellation token from the request context which I added earlier. Hence passing the requestContext object.

jennyf19: I mean implement the `INetworkCacheMetadataProvider` instead of creating a new one.

neha-bhargava: Ideally this would be passed from `InstanceDiscoveryManager`. The constructor creates a new instance only if it is not passed from the calling class.

neha-bhargava: I think the method name is confusing, by static cache I just mean to clear the value in the static variable region. I think I will rename it to just Clear

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
neha-bhargava: I added that method specifically to clear the s_region static variable that holds the value for region. The value for that variable was persisting between tests and causing the tests to fail. No validation is required here.

neha-bhargava:
> I'm not seeing a test that validates the scenario which is failing. Am I missing something?

Added integration test to validate the scenario.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -193,16 +202,20 @@ private void CacheInstanceDiscoveryMetadata(InstanceDiscoveryMetadataEntry metad
...
+                s_region = await GetRegionAsync(requestContext).ConfigureAwait(false);
+            }
+            
             var builder = new UriBuilder(canonicalAuthority);
 
             if (KnownMetadataProvider.IsPublicEnvironment(canonicalAuthority.Host))
             {
-                builder.Host = $"{region}.login.microsoft.com";
+                builder.Host = $"{s_region}.login.microsoft.com";
             }
```
jennyf19, line 215: do we have to worry about sovereign clouds in this scenario?

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
jennyf19: is there a test which tests the following:
- create one cca.AquireTokenForClient(scopes).WithAzureRegion(true);
and then same cca....
- cca.AcquireTokenForClient(scopes).WithAzureRegion(false)?
because that should show the 2nd call going to the global endpoint, correct?

---
In reply to: [544648991](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2308#discussion_r544648991) [](ancestors = 544648991)

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: No, currently ests-r support only public clouds and in case it is anything other than public cloud we are appending the region to the authority. This is confirmed with the ests team.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -37,18 +38,26 @@ internal sealed class RegionDiscoveryProvider : IRegionDiscoveryProvider
...
+         */
+        public void Clear()
+        {
+            s_region = null;
+        }
+
         public async Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority, RequestContext requestContext)
         {
             ICoreLogger logger = requestContext.Logger;
             string environment = authority.Host;
```
jennyf19, line 52: >string environment = authority.Host; [](start = 12, length = 36)

can be removed

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -37,18 +38,26 @@ internal sealed class RegionDiscoveryProvider : IRegionDiscoveryProvider
...
+            InstanceDiscoveryMetadataEntry cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, logger);
 
             if (cachedEntry == null)
             {
-                Uri regionalizedAuthority = await BuildAuthorityWithRegionAsync(authority, requestContext).ConfigureAwait(false);
                 CacheInstanceDiscoveryMetadata(CreateEntry(authority, regionalizedAuthority));
 
-                cachedEntry = _networkCacheMetadataProvider.GetMetadata(environment, logger);
+                cachedEntry = _networkCacheMetadataProvider.GetMetadata(regionalizedAuthority.Host, logger);
                 logger.Verbose($"[Region Discovery] Created metadata for the regional environment {environment} ? {cachedEntry != null}");
```
jennyf19, line 61: >logger [](start = 16, length = 6)

nit: for all the `logger` instances, you can just use the rc directly `requestContext.Logger` and same w/ the methods, just pass in the rc.logger

jennyf19:
In `InstanceDiscoveryManger`, i don't htink you need to create a new NetworkCacheMetadataProvider(), as it will be created if it's null, so you can just pass it to the RegionDiscoveryProvider:
`new RegionDiscoveryProvider(_httpManager, _networkCacheMetadataProvider);` and same class, do you need to pass the IRegionDiscoveryProivder? I think you could just create it in the GetMetadataEntryAsyn() method if autoDetectREgion is true.
Something like this:
```csharp
 public /* public for test */ InstanceDiscoveryManager(
            IHttpManager httpManager,
            IMatsTelemetryManager telemetryManager,
            bool shouldClearCaches,
            IUserMetadataProvider userMetadataProvider = null,
            Uri userProvidedInstanceDiscoveryUri = null,
            IKnownMetadataProvider knownMetadataProvider = null,
            INetworkCacheMetadataProvider networkCacheMetadataProvider = null,
            INetworkMetadataProvider networkMetadataProvider = null)
        {
            _httpManager = httpManager ?? throw new ArgumentNullException(nameof(httpManager));
            _telemetryManager = telemetryManager ?? throw new ArgumentNullException(nameof(telemetryManager));

            _userMetadataProvider = userMetadataProvider;
            _knownMetadataProvider = knownMetadataProvider ?? new KnownMetadataProvider();
            _networkCacheMetadataProvider = networkCacheMetadataProvider ?? new NetworkCacheMetadataProvider();
            
            _networkMetadataProvider = networkMetadataProvider ?? 
                new NetworkMetadataProvider(
                    _httpManager, 
                    _telemetryManager, 
                    _networkCacheMetadataProvider, 
                    userProvidedInstanceDiscoveryUri);

            if (shouldClearCaches)
            {
                _networkCacheMetadataProvider.Clear();
            }
        }

...
 if (autoDetectRegion)
            {
                var regionDiscoveryProvider = new RegionDiscoveryProvider(_httpManager, _networkCacheMetadataProvider);
                return await regionDiscoveryProvider.GetMetadataAsync(new Uri(authority), requestContext).ConfigureAwait(false);
            }
```


---
In reply to: [747085504](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2308#issuecomment-747085504) [](ancestors = 747085504)

jennyf19 (approved):
looks great. just a few additional comments.
:shipit:

neha-bhargava:
I am passing the `_networkCacheMetadataProvider` from the `InstanceDiscoveryManger`. The constructor first checks if the object `_networkCacheMetadataProvider` is null and only then creates a new object. That is a basic null check to avoid null pointer exception if it is null.

I can create the instance of `RegionDiscoveryProvider` in the if block but then it will be created more than once for every request as this flow is called multiple times

---
In reply to: [747786041](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2308#issuecomment-747786041) [](ancestors = 747786041,747085504)

Issue #2302, opened 2020-12-14
[Feature Request] Review sending PKAuth on unsupported platforms


bgavrilMS:
Today, on pretty much all platforms, MSAL advertises that it is PKAuth capable, and when ESTS challenges the app, MSAL simply states "oh, I can't actually do PKAuth". This seems very wasteful for ESTS (extra network call). Behaviour was copied from ADAL but no actual confirmation from ESTS that it is the right thing to do. 

Affected platforms: 

- .NET (FWK and Core) on Win10
- UWP
- mobile

bgavrilMS:
Related to: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1849

trwalke:
Issue has been identified on the UAP target. MSAL.UAP should be replaying the challenge with the bypass response but is instead throwing an exception by mistake. Will be resolved by next release

bgavrilMS:
I still think we should not send the PKAuth capability to ESTS unless OS = win10 and tfm = .net fwk or .net core

jmprieur:
@bgavrilMS do you mean OS = win10? or != win10 ?

jmprieur:
See also https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2026

bgavrilMS:
Let's continue to track the work on #1849

pmaytak:
This is included in [MSAL 4.25.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0).

Issue #1849, opened 2020-05-27
[Enhancement] Even on Win10 MSAL tries to extract a PKAuth key, which is a perf hit


bgavrilMS:
**Which Version of MSAL are you using ?**
4.14

**Platform**
net45 only

**Repro**

This is related to bug #1848 BUT is not the root cause of #1848 !

MSAL tries to determine if it's running on Win 7, 8 or 8.1, but the verification is wrong - see screenshot from my Win 10 machine:

![image](https://user-images.githubusercontent.com/12273384/83013774-7a9ba700-a015-11ea-8fdd-a08d4366599d.png)

You can read why `Enviroment.GetVersion()` is actually deprecated here: https://stackoverflow.com/questions/31885302/how-can-i-detect-if-my-app-is-running-on-windows-10

I think we need to make a call to read the registry to get the version of Windows.

**Expected behavior**
Exit the PKAuth logic faster.

**Actual behavior**
1. MSAL checks if the OS is supported. It should say "no" but it says "yes" on Win 10
2. MSAL tries to pull out the device key by reading certs etc. 
3. MSAL fails to pull out device key
4. MSAL reponds with "I can't perform PKAUth"

Step 2-3 can be avoided

**Possible Solution**

bgavrilMS:
@trwalke

bgavrilMS:
Probably as designed.

pmaytak:
This is included in [MSAL 4.25.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0).

Issue #1824, opened 2020-05-14
Provide more accurate Windows OS information in client info headers


onsubmit:
**Is your feature request related to a problem? Please describe.**
This feature request is in response to a Microsoft internal customer-reported [incident](https://icm.ad.msft.net/imp/v3/incidents/details/185949148/home). The customer noticed that when making requests on Windows Server 2012, in the sign-in activity logs, they would see OS = Windows 8 instead. After internal discussion, it was determined this was due to MSAL.NET using `Environment.OSVersion.ToString();` to report the OS version, which returns the exact same value for Win8 and Win2K12. [Source](https://docs.microsoft.com/en-us/windows/win32/sysinfo/operating-system-version).

`Environment.OSVersion` is also not a reliable way to accurately determine the real OS. The recommended approach is to use the [version helper APIs](https://docs.microsoft.com/en-us/windows/win32/sysinfo/version-helper-apis).

**Describe the solution you'd like**
MSAL.NET should follow the recommendations in the docs to more accurately report the OS to avoid confusing downstream effects.

**Describe alternatives you've considered**
None

**Additional context**
ping anyoung@microsoft for more context

jennyf19:
Thanks for the info @onsubmit

jmprieur:
Thanks @onsubmit !

henrik-me:
@onsubmit : can you pls. specify exactly what you would like MSAL to return for each of the platforms of interest?

onsubmit:
@henrik-me sent you mail internally

bgavrilMS:
We also use this logic wrongly in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1849
Fix should not be difficult.

jmprieur:
See also information in the duplicate https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2303

bgavrilMS:
This has been fixed in MSAL 4.30

PR #2312, opened 2020-12-16
PKey Auth Updates

trwalke:
Preventing MSAL from advertising PKey Auth on unsupported platforms.
Resolving OS version check bug.

Fixes for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1849
and https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2302


src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -76,8 +78,20 @@ public static bool CanOSPerformPKeyAuth()
         {
             //PKeyAuth can only be performed on operating systems with a major OS version of 6. 
             //This corresponds to windows 7, 8, 8.1 and their server equivilents.
-#if DESKTOP
-            if (Environment.OSVersion.Version.Major == 6)
```
trwalke, line 80: Sense Environment.OSVersion is not accurate, I used slightly different methods to properly detect the OS version. These are slightly different between .net fwk and .net core but produce the same results in testing

trwalke: @pmaytak We used to advertise that we are PKey Auth capable on every call in the token client but now we are looking to only do this on older versions of windows. (win 8 and below).

Thus I am adding these checks here to make sure we are only adding this header on supported platforms. Do you think these checks will have any impact on performance?

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -157,8 +158,11 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
                     _serviceBundle.HttpTelemetryManager.GetLastRequestHeader());
             }
 
-            //Signaling that the client can perform PKey Auth
-            _oAuth2Client.AddHeader(PKeyAuthConstants.DeviceAuthHeaderName, PKeyAuthConstants.DeviceAuthHeaderValue);
+            //Signaling that the client can perform PKey Auth on supported platforms
+            if (DeviceAuthHelper.CanOSPerformPKeyAuth())
+            {
+                _oAuth2Client.AddHeader(PKeyAuthConstants.DeviceAuthHeaderName, PKeyAuthConstants.DeviceAuthHeaderValue);
+            }
```
trwalke, line 165: @pmaytak We used to advertise that we are PKey Auth capable on every call in the token client but now we are looking to only do this on older versions of windows. (win 8 and below).

Thus I am adding these checks here to make sure we are only adding this header on supported platforms. Do you think these checks will have any impact on performance?

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
trwalke: @pmaytak We used to advertise that we are PKey Auth capable on every call in the token client but now we are looking to only do this on older versions of windows. (win 8 and below).

Thus I am adding these checks here to make sure we are only adding this header on supported platforms. Do you think these checks will have any impact on performance?

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -76,8 +78,20 @@ public static bool CanOSPerformPKeyAuth()
...
+            //Environment.OSVersion as it will return incorrect information on some operating systems
+#if NET_CORE
```
jmprieur, line 82: @trwalke 
A partner recommended in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1824, the use of the [version helper APIs](https://docs.microsoft.com/en-us/windows/win32/sysinfo/version-helper-apis), but I suppose for PKAuth this is not needed ?

jmprieur (commented):
LGTM.
I've asked a question (with a related issue)

jennyf19:
i think we need to get the device auth tests working. [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2207](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2207)

---
In reply to: [554694167](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2312#issuecomment-554694167) [](ancestors = 554694167)

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
jennyf19: correct. in the linked issue in that issue, bogdan includes a SO post w/more info.

---
In reply to: [545162569](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2312#discussion_r545162569) [](ancestors = 545162569)

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
pmaytak: @trwalke Seems like `CanOSPerformPKeyAuth` ends up being called for each request, right? Also, the OS version shouldn't change. So I would probably call this method once (at app startup or with a first request) and cache the OS version in a variable somewhere in the app.

Also according to [this](https://stackoverflow.com/questions/1097363/what-are-the-performance-implications-of-accessing-a-windows-registry-value) and [this](https://devblogs.microsoft.com/oldnewthing/20060222-11/?p=32193), registry access should be fast and the recent values are cached, but there're still extra those cycles spent, compared to if we cache the value in the app.

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
pmaytak: I would probably just use the registry once and cache the OS version in the app. Seems like a simpler approach.

Interestingly the [version helper APIs](https://docs.microsoft.com/en-us/windows/win32/sysinfo/version-helper-apis) are recommened for this. But then they use VerifyVersionInfo function which is [deprecated](https://docs.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1) in Windows 10 but I guess, can still work if the app has app.manifest.

pmaytak (approved):

trwalke:
Currently waiting for customer to test the preview package for this fix

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -257,4 +257,8 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="System.Management" />
```
bgavrilMS, line 261: is this for the registry stuff? if so, please add it to the desktop-only ItemGroup

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -7,14 +7,18 @@
...
 using Microsoft.Identity.Client.Http;
 using Microsoft.Identity.Client.PlatformsCommon.Shared;
 using Microsoft.Identity.Client.Utils;
+using Microsoft.Win32;
 
 namespace Microsoft.Identity.Client.PlatformsCommon
 {
     internal class DeviceAuthHelper
     {
+        private static bool? s_canOSPerformPKeyAuth;
```
bgavrilMS, line 20: maybe use the Lazy<T> initalization as it's thread safe (although this is probably ok as well)

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -74,14 +78,34 @@ public static string GetBypassChallengeResponse(Dictionary<string, string> respo
 
         public static bool CanOSPerformPKeyAuth()
         {
+            if (s_canOSPerformPKeyAuth != null)
+            {
+                return (bool)s_canOSPerformPKeyAuth;
```
bgavrilMS, line 83: s_canOSPerformPKeyAuth.Value

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
bgavrilMS: Please add the link where you copied the registry solution to the code, it's helpful to have a reference

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -74,14 +78,34 @@ public static string GetBypassChallengeResponse(Dictionary<string, string> respo
...
+            }
+
             //PKeyAuth can only be performed on operating systems with a major OS version of 6. 
             //This corresponds to windows 7, 8, 8.1 and their server equivilents.
-#if DESKTOP
-            if (Environment.OSVersion.Version.Major == 6)
+            //Environment.OSVersion as it will return incorrect information on some operating systems
+#if NET_CORE
```
bgavrilMS, line 89: #if NET_CORE || NET5_WIN

src/client/Microsoft.Identity.Client/PlatformsCommon/DeviceAuthHelper.cs
```diff
@@ -74,14 +78,34 @@ public static string GetBypassChallengeResponse(Dictionary<string, string> respo
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                && !RuntimeInformation.OSDescription.Contains("Windows 10"))
```
bgavrilMS, line 91: please use a case-insensitive Contains

bgavrilMS (commented):
LGTM with some small changes, particularly the new reference

trwalke:
Currently waiting for customer to test the preview package for this fix. The customer says that they are encountering build issues when trying to test fix.

However, I am able to test this locally and it works.

henrik-me:
>     It should not be required after the december release of .NET 5 SDK -->

@bogdan, do you have a work item tracking this?

---
Refers to: src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj:160 in 6344d7c. [](commit_id = 6344d7cbe502e289f74003fddca95bc9533ce344, deletion_comment = False)

bgavrilMS:
@trwalke - can we merge this? It's been sitting here for 28 days. I think this is a reasonable fix and @pmaytak reviewed it.

bgavrilMS (approved):

Issue #2259, opened 2020-12-01
[Feature Request] Add optional api to provide region for regional auth scenarios


neha-bhargava:
Add an optional api to provide region by the service. 

``` cs
.WithAzureRegion(true, UseRegion: <region>)
```

Refer to [this](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/2363?path=%2FPinAuthToRegion%2FAAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md&_a=files&anchor=add-api-to-provide-an-option-to-specify-the-region-when-opting-for-regional-auth.) document for details.

pmaytak:
This is included in [MSAL 4.25.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0).

PR #2315, opened 2020-12-18
Nebharg/add api use region

neha-bhargava:
#2259


src/client/Microsoft.Identity.Client/Region/RegionSource.cs
```diff
@@ -9,6 +9,7 @@ internal enum RegionSource
         None = 0,
         EnvVariable = 1,
         Imds = 2,
-        Cache = 3
+        Cache = 3,
+        Userprovided = 4
```
pmaytak, line 13: `UserProvided`

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -169,6 +170,12 @@ public int RegionSource
             set => this[RegionSourceKey] = (value).ToString(CultureInfo.InvariantCulture);
         }
 
+        public string ValidateUseRegion
```
pmaytak, line 173: Hmm maybe rename to something like `IsValidUserProvidedRegion` and maybe make it a nullable bool?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RegionalAuthIntegrationTests.cs
```diff
@@ -205,10 +278,11 @@ public async Task AcquireTokenToGlobalEndpointThenRegionalEndpoint_UseTokenFromC
 
         private async Task<AuthenticationResult> CreateAuthenticationResultAsync(
```
pmaytak, line 279: Not part of this PR, but I would maybe rename this to `GetAuthenticationResultAsync`, so that it's clearer that we're calling a method on CCA and not creating a mock auth result object?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RegionalAuthIntegrationTests.cs
```diff
@@ -103,6 +103,73 @@ public async Task AcquireTokenToRegionalEndpointAsync()
                 AuthenticationResult result = await CreateAuthenticationResultAsync().ConfigureAwait(false); // regional endpoint
                 AssertTokenSource_IsIdP(result);
                 AssertValidHost(true, factory);
+                AssertTelemetry(factory, "2|1004,0|centralus,1,0,");
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(TestConstants.RegionName, null);
```
pmaytak, line 110: Would it be appropriate to put this line into a separate test cleanup method, so it doesn't have to be repeated?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -79,10 +79,13 @@ public AcquireTokenForClientParameterBuilder WithSendX5C(bool withSendX5C)
         /// By default the value for this variable is false.
         /// See https://aka.ms/msal-net-region-discovery for more details.
         /// </summary>
-        /// <param name="autoDetectRegion"><c>true</c> if the token request should be sent to regional ESTS. Otherwise <c>false</c>.
-        /// The default is <c>false</c></param>
+        /// <param name="autoDetectRegion"><c>true</c> if the token request should be sent to regional ESTS. The default is <c>false</c>.
+        /// </param>
+        /// <param name="useRegion"> optional parameter to provide region to MSAL. This parameter will be used along with auto detection of region.
+        /// If the region is auto detected, the provided region will be compared with the detected region and used in telemetry to do analysis on correctness of the region provided.
+        /// If auto region detection fails, the provided region will be used for instance metadata.</param>
```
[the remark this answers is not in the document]
neha-bhargava: The auto detect region is set to true if the user wants to opt for regional endpoint. If set to false or not set, the requests will go to the global endpoint as usual.
UseRegion is an optional api where a user can provide the region in case they have it set in their config or they know the region. In case we are able to detect the region we will use the auto detected region and compare it with the region provided by the user to analyse the correctness of the region provided. In case we fail to detect the region and we have a value in useRegion, we will go ahead and use this region to hit the regional endpoint. And if in case user provided some garbage value in that place and we form a regional endpoint using it, ests is working on a way to handle those cases.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -79,10 +79,13 @@ public AcquireTokenForClientParameterBuilder WithSendX5C(bool withSendX5C)
...
         /// <returns>The builder to chain the .With methods</returns>
-        public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegion)
+        public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegion, string useRegion = "")
```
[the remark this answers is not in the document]
neha-bhargava: Makes sense, regionToUse sounds good. I will update it.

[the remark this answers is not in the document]
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
```diff
@@ -20,13 +20,18 @@ internal class AcquireTokenForClientParameters : IAcquireTokenParameters
         /// </summary>
         public bool AutoDetectRegion { get; set; }
 
+        /// <summary>
+        /// </summary>
+        public string UseRegion { get; set; }
```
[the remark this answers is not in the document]
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/Region/RegionSource.cs
neha-bhargava: Updated

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RegionalAuthIntegrationTests.cs
neha-bhargava: Updated

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RegionalAuthIntegrationTests.cs
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
neha-bhargava: Good point. Updated.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
jennyf19: Why not just use the region, if it's provided, and skip the auto detect region call? Then, if the the call fails, I guess try to autoDetect the region and try the request again? Is this specific guidance from eSTS on how to handle this?

---
In reply to: [549887878](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r549887878) [](ancestors = 549887878)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
jennyf19: i'm a little confused...so sorry for the weird questions...so if I do `.WithAzureRegion()` does that mean I am expecting to use the regional endpoint? If there a reason to even have the boolean? If a customer includes `.WithAzureRegion()` can we assume they want the regional endpoint? so if a customer uses this, it's always going to regional, and then they can specify the actual region they want, if they know it. It just seems counterintuitive to say I want to use the regional endpoint, and then also have to set a boolean to true. interested in discussing offline if you want.

---
In reply to: [549874926](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r549874926) [](ancestors = 549874926)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -93,6 +96,7 @@ public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegi
 
             CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithAzureRegion, autoDetectRegion);
             Parameters.AutoDetectRegion = autoDetectRegion;
+            Parameters.RegionToUse = regionToUse;
```
jennyf19, line 99: if it's an empty string do we still want to set it, as that's the default?

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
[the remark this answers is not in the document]
jennyf19: or maybe `RegionSpecifiedInConfig` 

---
In reply to: [550358060](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r550358060) [](ancestors = 550358060)

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -204,11 +206,31 @@ private void CacheInstanceDiscoveryMetadata(InstanceDiscoveryMetadataEntry metad
...
+
             if (s_region.IsNullOrEmpty())
             {
-                s_region = await GetRegionAsync(requestContext).ConfigureAwait(false);
+                try
+                {
+                    s_region = await GetRegionAsync(requestContext).ConfigureAwait(false);
+
+                    if (!regionToUse.IsNullOrEmpty())
+                    {
```
jennyf19, line 218: i would think the customer would know the region they want to target, and we should just use the provided value. what does ests suggest we do or have they provided input on that?

jennyf19: why do we care if the region they provided is valid or not? like you mentioned earlier, they could provide a garbage region and then we are logging all these errors, but we have no telemetry dashboard set up to get the data on this. Why does ests care if the region provided by the customer matches the region we discover? seems like a test they should have on their side, but not necessarily something we need in MSAL.

---
In reply to: [551518052](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r551518052) [](ancestors = 551518052)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
neha-bhargava: This is required to analyze the validity of the value that user provides. We know that there are azure services where we cannot detect the region and there can be more platforms where auto detection fails. For this reason it was decided to allow the user to provide the region. But this is something we cannot completely rely on as it could be hardcoded or the value of region they provide might not be the actual azure region. So, we will analyze the validity of the region provided by the user and in the initial phases of adoption we can let them know that there is a mismatch in the region they provide and region that was detected. Also, in case we fail to compare we are logging the region and region source which can help to ensure the region provided is correct and the calls are going to regional endpoint. If user opts to use regional endpoint goal is to hit the regional endpoint and use global only as a last option.

---
In reply to: [551511449](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r551511449) [](ancestors = 551511449,549887878)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
neha-bhargava: Initially, there was no plan to ask user to provide the region. But since now adopting regional is a resilience feature, we are adding things like allow the user to pass region as currently we cannot detect region on all platforms. So, the bool `autoDetectRegion` is what currently exists and people are testing it on their test environments. `regionToUse` is an optional field that someone may choose to provide and we will use it in case we are unable to auto detect the region to give more confidence to people adopting it.

---
In reply to: [551514796](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r551514796) [](ancestors = 551514796,549874926)

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: It is highly likely that people use a value for region that is different from the actual azure region and with allowing user to provide the region we are also open to the possibility of someone hardcoding this value and hitting the wrong regional endpoint. Happy to discuss this offline.

---
In reply to: [551519474](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r551519474) [](ancestors = 551519474,551518052)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
jennyf19: right, that makes sense. i guess what i'm suggesting is simplifying the API to `public AcquireTokenForClientParameterBuilder WithAzureRegion(string regionToUse = "")` because if they use this builder, they are automatically opting in to regional detection, but if they provide a value, we 1) try that region 2) if that fails try to detect the region 3) use global endpoint

---
In reply to: [551575530](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r551575530) [](ancestors = 551575530,551514796,549874926)

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
jennyf19: highly likely or highly unlikely? if highly likely, then sounds like we should do something else to detect this sooner. also, customers won't have the telemetry logs to see the azure regions, is this also getting logged in the regular logs as well? I guess, how will they know which region they are hitting? +1 to discussing. maybe at stand-up tomorrow.

---
In reply to: [551576965](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r551576965) [](ancestors = 551576965,551519474,551518052)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
jmprieur: Thanks for the explanation, @neha-bhargava, I had a similar question as @jennyf19 

In that case, what about having the following public API: `WithAzureRegion(string regionToUse="autoDetect")`:
- If the developer provides the region, then for sure the boolean would be true ? 
- and if they don't provide it (then the region would become) "autoDetect", which you'd then internally set boolean `autoDetectRegion` to true (but that would not be needed to be surfaced to the developers)

Note: you don't need to change your current method. Just make it internal, and wrap it into the API I propose?

I was also wondering if there is a value of passing `WithAzureRegion(false)`? What is the use case? would it be that the developer used the regional endpoint and then want to go to Global? but given this is on the AcquireTokenForClientParameterBuilder, the developer could just omit the `.WithAzureRegion()` to do that?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
jmprieur: Do we want to have an API on the IConfidentialClientApplication which would be string GetAzureRegion() ?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
pmaytak: Side note: We could just check for a default empty string, `WithAzureRegion(string regionToUse="")`. Would perhaps changing the parameter name to `fallbackRegion` be clearer?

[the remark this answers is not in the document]
henrik-me: @jmprieur : the reason to have the boolean is to avoid developers having to write if statements e.g.
if (boolUseRegion)
  builder.withRegion(...)

instead they can write withyRegion(boolUseRegion, "valueFromMyConfig")

#CleanerCode

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
henrik-me: I think you should revert back to what you had before for the reason mentioned above. the developer experience is better that way as it avoids the need for if statements

---
In reply to: [550357996](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2315#discussion_r550357996) [](ancestors = 550357996)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -93,6 +96,7 @@ public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegi
 
             CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithAzureRegion, autoDetectRegion);
```
henrik-me, line 97: >CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithAzureRegion, autoDetectRegion); [](start = 12, length = 95)

Wondering if you should set the regionToUse value in telemetry here as well

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
```diff
@@ -20,13 +20,18 @@ internal class AcquireTokenForClientParameters : IAcquireTokenParameters
         /// </summary>
         public bool AutoDetectRegion { get; set; }
 
+        /// <summary>
+        /// </summary>
+        public string RegionToUse { get; set; }
```
henrik-me, line 25: Even though this is an internal class, I believe the comments made on what the data is warrents that the summary field is specified.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -204,11 +206,31 @@ private void CacheInstanceDiscoveryMetadata(InstanceDiscoveryMetadataEntry metad
...
         private async Task<Uri> BuildAuthorityWithRegionAsync(Uri canonicalAuthority, RequestContext requestContext)
         {
+            string regionToUse = requestContext.ServiceBundle.Config.AuthorityInfo.RegionToUse;
+
             if (s_region.IsNullOrEmpty())
             {
-                s_region = await GetRegionAsync(requestContext).ConfigureAwait(false);
+                try
+                {
+                    s_region = await GetRegionAsync(requestContext).ConfigureAwait(false);
```
henrik-me, line 215: >s_region [](start = 20, length = 8)

what happens if this is null?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -204,11 +206,31 @@ private void CacheInstanceDiscoveryMetadata(InstanceDiscoveryMetadataEntry metad
...
+
+                    if (!regionToUse.IsNullOrEmpty())
+                    {
+                        requestContext.ApiEvent.IsValidUserProvidedRegion = s_region.Equals(regionToUse);
+                    }
+                }
+                catch (MsalServiceException e)
```
henrik-me, line 222: >e [](start = 44, length = 1)

you are not using this, remove and simply throw instead

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -204,11 +206,31 @@ private void CacheInstanceDiscoveryMetadata(InstanceDiscoveryMetadataEntry metad
...
+                {
+                    if (regionToUse.IsNullOrEmpty())
+                    {
+                        throw e;
```
henrik-me, line 226: >e [](start = 30, length = 1)

remove

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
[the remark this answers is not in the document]
jennyf19: @henrik-me how would that lead to not using if statements? if they don't want to use the region, then they just won't use .withRegion(). in the AT call, right?

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -120,16 +120,18 @@ public string GetCurrentRequestHeader(ApiEvent eventInProgress)
...
+            var platformConfig = new StringBuilder();
+            platformConfig.Append(regionDiscovered + ",");
+            platformConfig.Append(regionSource + ",");
+            platformConfig.Append(ConvertFromStringToBitwise(isTokenCacheSerialized) + ",");
+            platformConfig.Append(string.IsNullOrEmpty(isValidUserProvidedRegion) ? isValidUserProvidedRegion : ConvertFromStringToBitwise(isValidUserProvidedRegion)); //The value for this will be 1 if the region provided is valid, 0 if invalid and "" in case it could not be validated.
 
             return $"{TelemetryConstants.HttpTelemetrySchemaVersion2}" +
                 $"|{apiId},{ConvertFromStringToBitwise(forceRefresh)}" +
-                $"|{regionalFields},{ConvertFromStringToBitwise(isTokenCacheSerialized)}";
+                $"|{platformConfig}";
```
henrik-me, line 134: > $"|{platformConfig}"; [](start = 15, length = 22)

can you pls. show a few examples of what this looks like from your testing? with region, without region specified, where region can be discovered, where it can't be discovered, etc?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -169,6 +170,24 @@ public int RegionSource
             set => this[RegionSourceKey] = (value).ToString(CultureInfo.InvariantCulture);
         }
 
+        public bool? IsValidUserProvidedRegion
+        {
+#pragma warning disable CA1305 // .net standard does not have an overload for ToString() with Culture
```
henrik-me, line 175: >.net standard does not have an overload for ToString() with Culture [](start = 34, length = 67)

aside: [https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings?redirectedfrom=MSDN#stringtoupper-and-stringtolower](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings?redirectedfrom=MSDN#stringtoupper-and-stringtolower)

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
```diff
@@ -138,9 +137,86 @@ public async Task TelemetrySerializedTokenCacheTestAsync()
...
+                    ApiIds.AcquireTokenForClient,
+                    forceRefresh: false,
+                    "1",
+                    isCacheSerialized: false,
+                    isvalidUserProvidedRegion: "1");
+                AssertPreviousTelemetry(result.HttpRequest, expectedSilentCount: 0);
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(TestConstants.RegionName, "");
```
henrik-me, line 183: >Environment.SetEnvironmentVariable(TestConstants.RegionName, ""); [](start = 16, length = 65)

move to test cleanup? similar to what you did in the other test class?

henrik-me (approved):
:shipit:

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Yes, in case the region is not detected an exception is thrown

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: This is to catch the exception if the region was not detected. If in this case we don't even have a user provided region we throw the error and let the fallback to global logic handle it later. In case we have a user provided region, we don't throw an error but use that region.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: This ensures that we only throw an exception when we fail to auto detect and we don't have a user provided region.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
neha-bhargava: If I use toUpper, it messes up with code [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/c98af5f29b76077a9bf596b3e0c0dc5d78244755/src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs#L132) when it converts the bool to bitwise. So leaving it as is.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
neha-bhargava: Done

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
neha-bhargava: Added

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
neha-bhargava: Added this to telemetry

neha-bhargava:
Kusto query and results for different scenarios with user provided region.

Format in MsalCur: 2|api_id,force_refresh|region,region_source,is_token_cache_serialized,user_provided_region,validate_use_region

AllPerRequestTable
| where CorrelationId == "994f2694-f88f-40da-baa4-5cf887106fa7"
| where env_time between(datetime(2021-01-07 18:00:00) .. datetime(2021-01-07 20:55:35))
| project ApplicationId, env_time, CorrelationId, ClientInfo, MsalCur, MsalLast

Scenario 1: No user provided region, auto detection successful
Correlation id: 994f2694-f88f-40da-baa4-5cf887106fa7
image

Scenario 2: Valid User provided region, auto detection successful
Correlation id: 95eb45ca-0b95-42e2-b161-fda7371e4909
image

Scenario 3: Invalid user provided region, auto detection successful
Correlation id: 8339cf22-e597-4c53-b8e4-011d4b8d373c
image

Scenario 4: User provided region, auto detection fails
Correlation id: 8ea300e6-12b0-40fd-81cc-f7f9a6f95dda
image

Global
Correlation id: a4f35dab-a957-4505-b29e-f89ee616ff5b
image

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
neha-bhargava: changed the default value to null.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
neha-bhargava: Added a comment with scenarios and kusto query results.

PR #2316, opened 2020-12-18
Fix tests

neha-bhargava:
Fix the tests failing because of the value of region not getting cleared.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
@@ -80,7 +80,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
                     userProvidedInstanceDiscoveryUri);
 
             _regionDiscoveryProvider = regionDiscovery ??
-                new RegionDiscoveryProvider(_httpManager, _networkCacheMetadataProvider);
+                new RegionDiscoveryProvider(_httpManager, _networkCacheMetadataProvider, shouldClearCaches: shouldClearCaches);
 
             if (shouldClearCaches)
             {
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -80,7 +80,7 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
                     userProvidedInstanceDiscoveryUri);
 
             _regionDiscoveryProvider = regionDiscovery ??
-                new RegionDiscoveryProvider(_httpManager, _networkCacheMetadataProvider);
+                new RegionDiscoveryProvider(_httpManager, _networkCacheMetadataProvider, shouldClearCaches: shouldClearCaches);
```
trwalke, line 83: It may be that I am just ignorant of this but what does the "shouldClearCaches: " do? Can you just pass in the "shouldClearCaches" value by itself?

trwalke (approved):
Just one question

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
neha-bhargava: When we do test initialize this value is passed from there. I added a variable in the RegionDiscoveryProvider constructor to take this value in. 
```
public RegionDiscoveryProvider(
            IHttpManager httpManager, 
            INetworkCacheMetadataProvider networkCacheMetadataProvider = null, 
            int imdsCallTimeout = 2000,
            /* for tests only */ bool shouldClearCaches = false)
```
This is the constructor. Since I am not passing the imdsCallTimeout, I need to mention the variable I am passing

Issue #2287, opened 2020-12-08
Add logic to fallback to global endpoint if region detection through environment variable and imds fails.


neha-bhargava:
In case MSAL fails to detect the region, fallback to global endpoint.

Please refer to the [design doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/2363?path=%2FPinAuthToRegion%2FAAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md&discussionId=22632&_a=files&anchor=proposed-changes-to-add-resiliency-to-auto-region-detection) for details.

Currently we will only fallback to global when region detection fails and not when regional ests fail. For regional ests failures we are waiting on the ests-r team to provide us with the allow list of errors for which we need to fallback to global.

* Add an api FallbackToGlobal in the confidential client WithAzureRegion(AutoDetectRegion: true, FallbackToGlobal: true)
* Fallback logic in case we fail to detect the region -- first request decides this and then cache
* Add to telemetry if we are falling back to global for the request

bgavrilMS:
@neha-bhargava @trwalke  - isn't this done?

pmaytak:
This is included in [MSAL 4.25.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0).

PR #2317, opened 2020-12-23
Adding api to allow users to fallback to global if regional lookup fails

trwalke:
fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2287

@neha-bhargava  Can you take a look here and make sure I am doing everything correctly?


trwalke:
test will be added soon

neha-bhargava:
The changes look good. I have a question, in case the fallbackToGlobal is set to true but region is found lets set the value in telemetry as 0 and leave it blank in case the fallbackToGlobal is false.

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -126,7 +126,12 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
 
             if (autoDetectRegion)
             {
-                return await _regionDiscoveryProvider.GetMetadataAsync(new Uri(authority), requestContext).ConfigureAwait(false);
+                var autoDetectedRegion =  await _regionDiscoveryProvider.GetMetadataAsync(new Uri(authority), requestContext).ConfigureAwait(false);
```
neha-bhargava, line 129: I will suggest renaming autoDetectedRegion to something like regionalizedMetadata.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -221,6 +227,13 @@ private async Task<Uri> BuildAuthorityWithRegionAsync(Uri canonicalAuthority, Re
                 }
                 catch (MsalServiceException e)
                 {
+                    if (requestContext.ServiceBundle.Config.AuthorityInfo.FallbackToGlobal)
+                    {
+                        requestContext.Logger.Info($"Attempting to fall back to global region");
```
neha-bhargava, line 232:
```suggestion
                        requestContext.Logger.Info($"Attempting to fall back to global endpoint");
```

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -100,6 +100,37 @@ public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegi
...
+        /// See https://aka.ms/msal-net-region-discovery for more details.
+        /// </summary>
+        /// <param name="autoDetectRegion"><c>true</c> if the token request should be sent to regional ESTS. The default is <c>false</c>.
+        /// </param>
+        /// <param name="fallbackToGlobal"></param>
+        /// <param name="useRegion"> optional parameter to provide region to MSAL. This parameter will be used along with auto detection of region.
+        /// If the region is auto detected, the provided region will be compared with the detected region and used in telemetry to do analysis on correctness of the region provided.
+        /// If auto region detection fails, the provided region will be used for instance metadata.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegion, bool fallbackToGlobal, string useRegion = "")
```
neha-bhargava, line 118: I was wondering if we should default the fallbackToGlobal to false. Thoughts @henrik-me?

pmaytak: Should fallback be defaulted to true? My understanding is that regional ESTS is a complementary feature to improve request performance. So we'd rather send request to global, than to error and not send it at all?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -100,6 +100,37 @@ public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegi
...
+        /// Specifies if the token request should be sent to regional ESTS.
+        /// If set, MSAL tries to auto-detect and use a regional Azure authority. This helps keep the authentication traffic inside the Azure region. 
+        /// If the region cannot be determined (e.g. not running on Azure), MSALClientException is thrown with error code region_discovery_failed.
+        /// This feature requires configuration at tenant level.
+        /// By default the value for this variable is false.
+        /// See https://aka.ms/msal-net-region-discovery for more details.
+        /// </summary>
+        /// <param name="autoDetectRegion"><c>true</c> if the token request should be sent to regional ESTS. The default is <c>false</c>.
+        /// </param>
+        /// <param name="fallbackToGlobal"></param>
```
pmaytak, line 113:
```suggestion
        /// <param name="fallbackToGlobal"><c>true</c> to fallback to global ESTS endpoint when calls to regional ESTS fail.</param>
```

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -176,6 +177,12 @@ public string ValidateUseRegion
             set => this[ValidateUseRegionKey] = value;
         }
 
+        public string FallBackToGlobal
```
pmaytak, line 180: Is nullable boolean appropriate for this?

trwalke:
Need to block reviews for the time being, i need to fix the git merge issues first

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -100,6 +100,37 @@ public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegi
...
+        /// See https://aka.ms/msal-net-region-discovery for more details.
+        /// </summary>
+        /// <param name="autoDetectRegion"><c>true</c> if the token request should be sent to regional ESTS. The default is <c>false</c>.
+        /// </param>
+        /// <param name="fallbackToGlobal"><c>true</c> to fallback to global ESTS endpoint when calls to regional ESTS fail.</param>
+        /// <param name="regionToUse"> optional parameter to provide region to MSAL. This parameter will be used along with auto detection of region.
+        /// If the region is auto detected, the provided region will be compared with the detected region and used in telemetry to do analysis on correctness of the region provided.
+        /// If auto region detection fails, the provided region will be used for instance metadata.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegion, bool fallbackToGlobal, string regionToUse = null)
```
henrik-me, line 118: >public AcquireTokenForClientParameterBuilder WithAzureRegio [](start = 8, length = 59)

considering taking the name changes discussed separately

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -100,6 +100,37 @@ public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegi
...
+        {
+            if (!ServiceBundle.Config.ExperimentalFeaturesEnabled)
+            {
+                throw new MsalClientException(
+                    MsalError.ExperimentalFeature,
+                    MsalErrorMessage.ExperimentalFeature(nameof(WithAzureRegion)));
```
henrik-me, line 124: Should probably be a function by now, it has been copy pasted many times by now

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
henrik-me, line 118: >bool autoDetectRegion, bool fallbackToGlobal [](start = 69, length = 44)

provide defaults, and remove other instances of WithAzureRegion ?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -100,6 +100,37 @@ public AcquireTokenForClientParameterBuilder WithAzureRegion(bool autoDetectRegi
...
+        /// Specifies if the token request should be sent to regional ESTS.
+        /// If set, MSAL tries to auto-detect and use a regional Azure authority. This helps keep the authentication traffic inside the Azure region. 
+        /// If the region cannot be determined (e.g. not running on Azure), MSALClientException is thrown with error code region_discovery_failed.
+        /// This feature requires configuration at tenant level.
+        /// By default the value for this variable is false.
+        /// See https://aka.ms/msal-net-region-discovery for more details.
+        /// </summary>
+        /// <param name="autoDetectRegion"><c>true</c> if the token request should be sent to regional ESTS. The default is <c>false</c>.
+        /// </param>
+        /// <param name="fallbackToGlobal"><c>true</c> to fallback to global ESTS endpoint when calls to regional ESTS fail.</param>
```
henrik-me, line 113: >when calls to regional ESTS fail [](start = 91, length = 32)

this should only happen when we are not able to detect a region or if there is no provided region

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
henrik-me, line 118: >fallbackToGlobal [](start = 97, length = 16)

this parameter should go last

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -126,7 +126,12 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
 
             if (autoDetectRegion)
             {
-                return await _regionDiscoveryProvider.GetMetadataAsync(new Uri(authority), requestContext).ConfigureAwait(false);
+                var regionalMetadata =  await _regionDiscoveryProvider.GetMetadataAsync(new Uri(authority), requestContext).ConfigureAwait(false);
+
+                //If the auto detected region is null, this indicates that the FallbackToGlobal parameter was set and the GetMetadataAsync failed to get the region.
```
henrik-me, line 131: >/If the auto detected region is null, this indicates that the FallbackToGlobal parameter was set an [](start = 17, length = 99)

really? that sounds confusing. Are we keeping the right logic in the right place? when we need long explanations like this, naming and code location is often not the best choice.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -227,6 +235,14 @@ private async Task<Uri> BuildAuthorityWithRegionAsync(Uri canonicalAuthority, Re
                 }
                 catch (MsalServiceException e)
```
henrik-me, line 236: >catch (MsalServiceException e) [](start = 16, length = 30)

e is not used? simply throw instead of thowing e

henrik-me:
>                     LogTelemetryData(s_region, RegionSource.UserProvided, requestContext);

We  should log what the user provided in the entry point, not needed here ideally. here we can log what is used, we should log both what we detected (if we detected anything), what the user provided and what was used.

---
Refers to: src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs:253 in e5d5c35. [](commit_id = e5d5c3599df9cbd2e3c59e2e8c7e1320ca379312, deletion_comment = False)

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -195,6 +196,12 @@ public string UserProvidedRegion
 #pragma warning restore CA1305 // Specify IFormatProvider
         }
 
+        public string FallBackToGlobal
```
henrik-me, line 199: >FallBackToGlobal [](start = 22, length = 16)

why is this not following the pattern of lower casing?

henrik-me (changes requested):
:clock1:

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
trwalke: I am not entirely sure. This entire file seems to be following this pattern so I used what is already established here.

trwalke:
> > ```
> >                 LogTelemetryData(s_region, RegionSource.UserProvided, requestContext);
> > ```
> 
> We should log what the user provided in the entry point, not needed here ideally. here we can log what is used, we should log both what we detected (if we detected anything), what the user provided and what was used.
> 
> Refers to: src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs:253 in [e5d5c35](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/e5d5c3599df9cbd2e3c59e2e8c7e1320ca379312). [](commit_id = [e5d5c35](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/e5d5c3599df9cbd2e3c59e2e8c7e1320ca379312), deletion_comment = False)

I can confirm that this is is recording what was detected and used in the "s_region" parameter. What the user has provided is pulled out of the request context param.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
neha-bhargava, line 199:
```suggestion
        public string FallbackToGlobal
```

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -52,9 +52,15 @@ public void Clear()
...
         }
 
-        public async Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority, RequestContext requestContext)
+        public async Task<InstanceDiscoveryMetadataEntry> TryGetMetadataAsync(Uri authority, RequestContext requestContext)
         {
             Uri regionalizedAuthority = await BuildAuthorityWithRegionAsync(authority, requestContext).ConfigureAwait(false);
+
+            if (regionalizedAuthority == null && requestContext.ServiceBundle.Config.AuthorityInfo.FallbackToGlobal)
+            {
+                return null;
```
neha-bhargava, line 61: I would suggest to add a log here if fallback to global will happen.

neha-bhargava:
Yes it is all logged. What was provided and what was detected and which value was used as region.

---
In reply to: [758302818](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2317#issuecomment-758302818) [](ancestors = 758302818)

neha-bhargava (approved):
Looks good. Some minor comments.

henrik-me:
Generally looks good. There are still a few open comments which should be addressed. Please ack all comments (mark resolve, won't fix, another pr, or similar)

henrik-me (approved):
:shipit:

PR #2322, opened 2020-12-31
Disabling tests

trwalke:
Disabling device auth tests


trwalke:
see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2323 for details

PR #2337, opened 2021-01-13
Fix for WAM + console apps

bgavrilMS:
Experience (see the grey splash screen..). I could add a progress bar ... thoughts?

image

Implication of the splash screen - the splash screen means Win.Forms is used. So we cannot offer WAM from pure NetStandard.

**Tested**: 

  |   | ui thread | non ui thread
-- | -- | -- | --
net classic | wpf | x | x
  | win forms | x | x
  | console | x | x
net core 3 | win forms | x | x
  | console | x | x
net 5 | win forms | x | x
  | console | x | x
uwp |   | x | x

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AccountPicker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AccountPicker.cs
@@ -8,6 +8,8 @@ using Microsoft.Identity.Client.Instance;
 using Windows.Security.Authentication.Web.Core;
 using Windows.Security.Credentials;
 using Windows.UI.ApplicationSettings;
+using System.Runtime.InteropServices;
+using Microsoft.Identity.Client.PlatformsCommon.Shared;
 
 #if NET5_WIN
 using Microsoft.Identity.Client.Platforms.net5win;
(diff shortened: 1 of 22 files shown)


src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AccountPicker.cs
```diff
@@ -36,37 +38,173 @@ internal class AccountPicker : IAccountPicker
...
+        {
+            // if there is a sync context, move to it (go to ui thread)
+            if (_synchronizationContext != null)
+            {
+                await _synchronizationContext;
+            }
+
+            if (UseSplashScreen())
+            {
+                await ShowPickerWithSplashScreenAsync().ConfigureAwait(false);
```
bgavrilMS, line 67: Console apps do not have ui thraed / synchronization context

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AccountPicker.cs
```diff
@@ -36,37 +38,173 @@ internal class AccountPicker : IAccountPicker
...
+
+        /// <summary>
+        /// The account picker API has bug that prevent correct usage from console apps. 
+        /// To workaround, show a splash screen and attach to it.
+        /// </summary>
+        /// <returns></returns>
+        private Task<bool> ShowPickerWithSplashScreenAsync()
+        {
+
+            if (Thread.CurrentThread.GetApartmentState() == ApartmentState.MTA)
```
bgavrilMS, line 106: Console apps do not have UI thread, current thread will always be MTA for console apps. This will cause the Account Picker to crash. To fix this, we will create a STA thread (i.e. UI thread). 

Note that a WinForms / WPF app can have multiple STA threads, but invoking the Account Picker from an STA thread other than the UI thread will cause a crash. 

So... to sum up

- console app - create an STA thread
- non-console apps - move to the existing STA thread (or throw a meaningful exception if AcquireTokenInteractice is not called from UI thread).

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
```diff
@@ -179,10 +183,6 @@ internal class WamBroker : IBroker
 
             try
             {
-#if WINDOWS_APP
-                // UWP requires being on the UI thread
-                await _synchronizationContext;
```
bgavrilMS, line 184: Moved this to wamProxy

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AccountPicker.cs
```diff
@@ -36,37 +38,172 @@ internal class AccountPicker : IAccountPicker
...
         }
 
-        private async Task<WebAccountProvider> ShowPickerAsync()
+        private async Task ShowPickerForWin32WindowAsync(IntPtr windowHandle)
         {
             AccountsSettingsPane retaccountPane = null;
             try
             {
+                retaccountPane = AccountsSettingsPaneInterop.GetForWindow(windowHandle);
+                retaccountPane.AccountCommandsRequested += Authenticator_AccountCommandsRequested;
```
jennyf19, line 172: can `retaccountPane` be null here, I see you check for it later.

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/win32/Splash.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.	// Licensed under the MIT License.

```

jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.	// Licensed under the MIT License.

﻿using System;
```

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/win32/Win32Window.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.	// Licensed under the MIT License.

﻿using System;
```

jennyf19 (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/win32/Splash.Designer.cs
```diff
@@ -0,0 +1,67 @@
...
+            this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
+            this.ClientSize = new System.Drawing.Size(231, 63);
+            this.ControlBox = false;
+            this.Controls.Add(this.label1);
+            this.FormBorderStyle = System.Windows.Forms.FormBorderStyle.None;
+            this.MaximizeBox = false;
+            this.MinimizeBox = false;
+            this.Name = "Splash";
+            this.ShowIcon = false;
+            this.ShowInTaskbar = false;
```
pmaytak, line 55: Could the splash screen size be set to something like 1px x 1px (my guess is 0px is probably not possible) to completely hide it from the user? If not, I would update the label to something more descriptive like 'Please select an account to sign in'.

pmaytak (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AccountPicker.cs
```diff
@@ -36,37 +38,172 @@ internal class AccountPicker : IAccountPicker
...
                 await AccountsSettingsPane.ShowAddAccountAsync();
-#else
-                retaccountPane = AccountsSettingsPaneInterop.GetForWindow(_parentHandle);
-                retaccountPane.AccountCommandsRequested += Authenticator_AccountCommandsRequested;
-                await AccountsSettingsPaneInterop.ShowAddAccountForWindowAsync(_parentHandle);
-#endif
+
                 return _provider;
             }
             catch (Exception e)
```
neha-bhargava, line 209: Does the retaccountPane needs to be set to null here too in the finally?

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AccountPicker.cs
bgavrilMS: well, this is a trick to get the GC to dispose of that object sooner rather than later. So it does not NEED to be set to null, but it can't hurt.

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/win32/Splash.Designer.cs
bgavrilMS: It can be made 1px x 1px but discussing with Windows guys they are against it. It's a hack, let's not hide it. It will be seen. 
I am against text in the label because of localization issues.

PR #2341, opened 2021-01-14
Fix spelling and mark default to true for FallbackToGlobal

neha-bhargava:
Update the api as discussed in the meeting with default value for FallbackToGlobal as true.


pmaytak (approved):
👍

PR #2347, opened 2021-01-15
Add fallbackToGlobal to telemetry

neha-bhargava:


src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -106,7 +106,7 @@ public string GetLastRequestHeader()
 
         /// <summary>
         /// Expected format: 2|api_id,force_refresh|platform_config
-        /// platform_config: region,region_source,is_token_cache_serialized,user_provided_region,validate_use_region,is_legacy_cache_enabled
+        /// platform_config: region,region_source,is_token_cache_serialized,user_provided_region,validate_use_region,is_legacy_cache_enabled,fallback_to_global
```
jennyf19, line 109: we haven't shipped yet, do we want to put all the regional stuff together? so switch it w/the legacy cache?

pmaytak (approved):

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
pmaytak: btw would have to move `is_token_cache_serialized` too, it's in the middle of regional as well (unless that was already released previously?)

neha-bhargava: `is_token_cache_serialized` was released previously. I too had a thought if I should move it together. Let me make that change real quick -- switch is_legacy_cache_enabled with fallback_to_global

trwalke (approved):
Thanks Neha. this looks good

PR #2348, opened 2021-01-15
Fix reading of WSTrust13 policies

CrustyJew:
Currently `MexDocument` will never return WSTrust13 policies. This is caused by WSTrust13 using `XmlNamespace.Sp` security policy on its `UsernameToken` nodes.

This pull request fixes that by simplifying the code and searching first for the node with `XmlNamespace.Sp` followed by searching for the node with `XmlNamespace.Sp2005` if it wasn't found.

I can provide some sample XML files if needed, but I basically just used ADFS 3.0's /trust/mex end point to generate a file.

### DIFF
--- a/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
+++ b/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
@@ -91,28 +91,25 @@ namespace Microsoft.Identity.Client.WsTrust
                 IEnumerable<XElement> all = exactlyOnePolicy.Descendants(XmlNamespace.Wsp + "All");
                 foreach (XElement element in all)
                 {
-                    XNamespace securityPolicy = XmlNamespace.Sp;
                     XElement auth = element.Elements(XmlNamespace.Http + "NegotiateAuthentication").FirstOrDefault();
                     if (auth != null)
                     {
                         AddPolicy(policy, UserAuthType.IntegratedAuth);
                     }
 
-                    auth = element.Elements(securityPolicy + "SignedEncryptedSupportingTokens").FirstOrDefault();
-                    if (auth == null && ((auth = element.Elements(XmlNamespace.Sp2005 + "SignedSupportingTokens").FirstOrDefault()) ==
-                                         null))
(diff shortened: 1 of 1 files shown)


jmprieur:
Thanks @CrustyJew. Yes, it would be useful to provide sample XML files.

CrustyJew:
To test it I created a simple console app that reads the XML file and creates an instance of `MexDocument` then outputs the results of `.GetWsTrustUsernamePasswordEndpoint().Uri.ToString()`. I've attached an example xml file that I trimmed down to highlight the issue. 
[example mex.zip](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/files/5830730/example.mex.zip)

Running this file with the existing code base gets a result of  `https://fake.auth0.com/trust/2005`

Running this file with this pull requests gets a result of `https://fake.auth0.com/trust13`

henrik-me:
I'm not sure about this change. @brentschmaltz @GeoK would be good with your perspective.

CrustyJew:
@henrik-me do you have any specific concerns I can address? 

The existing code for `GetWsTrustUsernamePasswordEndpoint()` indicates it should prefer the WSTrust13 binding but it would never get picked up with the current code not ever adding that policy to the list due to namespace issues.

henrik-me:
Backwards compatibility, possibility of injecting invalid xml, also considering if we are making such a change if there are elements we should include for other wstrust versions. Also not sure how to have a full set of test cases ensuring we are compliant with the spec.

brentschmaltz:
@CrustyJew is this because the call graph only see WsTrust2005?

CrustyJew:
@brentschmaltz Correct, and judging by the comment in the code on [this line](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/cc044f6da9f7a6e7268531d1f4c225cfcba64124/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs#L69) `//try ws-trust 1.3 first` and the code, it SHOULD be using WSTrust13 if it is available, it's just the code is not working correctly to pick it up into the available list of policies.

If you look at the original code (prior to my pull request) it kinda sorta starts to parse the WSTrust13 version with the Security Policy namespace for the first node it parses out, just not for the later ones.

bgavrilMS:
I do not think we have the bandwidth for testing this change. It is possible that WAM implements the protocol better, I suggest using WAM as described here: httsp://aka.ms/msal-net-wam 
When calling `AcquireTokenSilent` with WAM enabled, WAM will attempt to perform IWA protocol. 
ROPC via WAM is not supported.

Issue #2353, opened 2021-01-19
[Bug] App Gets stuck in "Client Request Loop" due to User.ReadBasic.All scope 


TheJoeFin:
**Logs and Network traces**
I will work on getting logs.

**Which Version of MSAL are you using ?**
MSAL 4.18 works without issue, but upgrading beyond 4.18 causes this exception loop to occur. 

**Platform**
UWP

**What authentication flow has the issue?**
* Desktop / Mobile
    * [X] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)

Other? - please describe;

**Is this a new or existing app?**
The app (Ink Calendar) is in production, I haven't upgraded MSAL, but started seeing this issue when upgrading beyond 4.18

**Repro**

```csharp

try
{
    authResult = await PublicClientApp.AcquireTokenSilent(scopes, firstAccount)
                                                     .ExecuteAsync();
}
catch (MsalUiRequiredException ex)
{
    Debug.WriteLine($"MsalUiRequiredException: {ex.Message}");

    // This is where the looping failure occurs: ↓
    authResult = await PublicClientApp.AcquireTokenInteractive(scopes)
                                                  .ExecuteAsync()
                                                  .ConfigureAwait(false);
}
return authResult.AccessToken;



```

**Expected behavior**
I expect versions beyond 4.18 to work as well as 4.18 does. 

**Actual behavior**
When the app tries to upload the ink file which is being saved there is an exception thrown "the file is in use by another process" then stops and tries again later. When upgrading MSAL beyond 4.18 this exception throws the MS Graph identity into a loop which asks the user to login again, over and over and over. Closing the app is the only way to stop the loop. 

**Possible Solution**
Stay on 4.18

bgavrilMS:
Please try to get logs. In the meantime, could you tell me what scopes you ask for?

TheJoeFin:
Scopes
` User.Read User.ReadBasic.All Calendars.Read Files.ReadWrite.AppFolder`

For logging do you want logs on the good package (4.18) and the latest package I'm seeing the exception loop? Or just when I'm seeing the exception loop?

bgavrilMS:
Latest MSAL version to see the actual bug

TheJoeFin:
Here is the end of the debug output. Let me know if you need more or something else

```
[...]

PR #2354, opened 2021-01-20
More verbose logging around FindAT and SaveMsalResponse

bgavrilMS:
No perf impact on FindAccessToken logic...

Before
|                         Method | TokenCacheSize |     Mean |   Error |  StdDev |
|------------------------------- |--------------- |---------:|--------:|--------:|
| AcquireTokenForClientTestAsync |         100000 | 104.9 ms | 2.06 ms | 2.37 ms |

After
|                         Method | TokenCacheSize |     Mean |   Error |  StdDev |
|------------------------------- |--------------- |---------:|--------:|--------:|
| AcquireTokenForClientTestAsync |         100000 | 104.4 ms | 2.05 ms | 2.52 ms |

### DIFF
--- a/src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
+++ b/src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
@@ -9,6 +9,8 @@ using Microsoft.Identity.Client.Internal.Broker;
 using Microsoft.Identity.Json;
 using Microsoft.Identity.Json.Linq;
 using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Core;
+using System.Text;
 
 namespace Microsoft.Identity.Client.OAuth2
 {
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -181,5 +183,43 @@ internal static MsalTokenResponse CreateFromAndroidBrokerResponse(string jsonRes
...
+                withoutPii.AppendLine("==MsalTokenResponse==");
+
+                withPii.AppendLine($"Error: {Error} ErrorDescription: {ErrorDescription}");
+                withoutPii.AppendLine($"Error: {Error} ErrorDescription: {ErrorDescription}");
+                withPii.AppendLine($"Scopes: {Scope} ");
+                withoutPii.AppendLine($"Scopes: {Scope} ");
+                withPii.AppendLine($"ExpiresIn: {ExpiresIn} RefreshIn {RefreshIn}");
+                withoutPii.AppendLine($"ExpiresIn: {ExpiresIn} RefreshIn {RefreshIn}");
+
+                withoutPii.AppendLine(
```
jennyf19, line 204: I don't think we should log any token info or client info, and definitely not RTs, in either pii or non-pii logs. Has the guidance on this changed? 

what problem are we trying to solve? which issue is this linked to?

bgavrilMS: There is a linked issue associated with the PR, it is https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2353

I am no logging any tokens, except for ClientInfo which is not a token. I'm only logging true or false if the token is present or not.

jennyf19: got it. my mistake, sorry about the false alarm.

bgavrilMS: No prob, I'm happy that someone is looking at this important aspect in detail. Would not want to be the root cause of some security incident :p

trwalke (approved):

Issue #2367, opened 2021-01-27
MCG : warning MCG0007: Unresolved P/Invoke method


myokeeh:
I'm now getting an MCG0007 warning for "Release" on my UWP app using version 4.25.0. I'm sure it has something to do with "Compile with .NET native tool chain" also. I don't think it's a bug per se since I haven't encountered any problems with authentication.

>  C:\App\.nuget\packages\microsoft.net.native.compiler\2.2.9-rel-29512-01\tools\Microsoft.NetNative.targets(805,5): warning : 
> MCG : warning MCG0007: Unresolved P/Invoke method 'user32.dll!GetWindowThreadProcessId' for method System.UInt32Microsoft.Identity.Client.PlatformsCommon.Shared.WindowsNativeMethods.GetWindowThreadProcessId(System.IntPtr, System.UInt32)'. Calling this method would throw exception at runtime. Please make sure the P/Invoke either points to a Windows API allowed in UWP applications, or a native DLL that is part of the package. If for some reason your P/Invoke does not satisfy those requirements, please use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP APIs.
> [C:\App\Source\Repos\App\AppWTS\AppWTS.csproj]
> 
> C:\App\.nuget\packages\microsoft.net.native.compiler\2.2.9-rel-29512-01\tools\Microsoft.NetNative.targets(805,5): warning :       
> MCG : warning MCG0007: Unresolved P/Invoke method 'secur32.dll!GetUserNameEx' for method 'System.Boolean Microsoft.Identity.Client.PlatformsCommon.Shared.WindowsNativeMethods.GetUserNameEx(System.Int32, System.Text.StringBuilder, System.UInt32)'. Calling this method would throw exception at runtime. Please make sure the P/Invoke either points to a Windows API allowed in UWP applications, or a native DLL that is part of the package. If for some reason your P/Invoke does not satisfy those requirements, please use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP APIs. [C:\App\Source\Repos\App\AppWTS\AppWTS.csproj]
> 
> C:\App\.nuget\packages\microsoft.net.native.compiler\2.2.9-rel-29512-01\tools\Microsoft.NetNative.targets(805,5): warning :       
[...]

PR #2374, opened 2021-02-01
Fix for #2367

bgavrilMS:
Do not include some Win32Native methods in UWP as they cannot be invoked anyway and they cause warnings (e.g. GetForegroundWindow() etc.)


jennyf19 (approved):
:shipit:

pmaytak (approved):

PR #2375, opened 2021-02-01
[WAM] AADPlugin.GetAccounts logic for finding true home account id

bgavrilMS:
AADPlugin.GetAccounts does not reliably give us the home account id for the accounts it knows about, because WAM's AAD plugin does not deal with guest accounts. 

To overcome this problem, we'll use the following logic: 

1. get all accounts from WAM and foreach WAM account ...
1. try to find an account in MSAL's token cache with the same UPN. If found, use that account's ID.
1. if you can't find one, fire a silent web request via WAM, asking for just "profile" and "openid" for that account
1. if successful, use the account ID that is embedded in the response (i.e. the client_info stuff).
1. if not successful, then skip the account (it's not a big deal, the user will just have to interactively login with it to discover it).

This solves the problem of MSA-pt accounts, which are guest accounts and are reported incorrectly by AAD.


src/client/Microsoft.Identity.Client/Account.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Identity.Client
     /// Contains information of a single account. A user can be present in multiple directories and thus have multiple accounts.
     /// This information is used for token cache lookup and enforcing the user session on the STS authorize endpoint.
     /// </summary>
-    internal sealed class Account : IAccountInternal
+    internal sealed class Account : IAccount
```
bgavrilMS, line 14: small refactoring - use Account object more instead of internal interface.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -155,13 +169,21 @@ private async Task<IEnumerable<IAccount>> GetAccountsInternalAsync(ApiIds apiId,
...
+            string homeAccountIdFilter,
+            ICacheSessionManager cacheSessionManager,
+            RequestContext requestContext)
         {
             if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
                 var brokerAccounts =
-                    (await broker.GetAccountsAsync(AppConfig.ClientId, AppConfig.RedirectUri).ConfigureAwait(false))
+                    (await broker.GetAccountsAsync(
```
bgavrilMS, line 181: The broker interface is much more complex now, as it needs to be able to fetch accounts from the cache, save a token response etc.

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AadPlugin.cs
```diff
@@ -54,18 +112,103 @@ private Account ConvertToMsalAccountOrNull(WebAccount webAccount)
...
+            return ParseSuccessfullWamResponse(wamResult.ResponseData[0], out _);
+        }
+
+        private bool MatchCacheAccount(
+            WebAccount webAccount,
+            IEnumerable<IAccount> accountsFromCache,
+            out AccountId homeAccountId)
+        {
+
+            // TODO: a match can also be done on local account id
```
bgavrilMS, line 195: not doing this.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -145,8 +145,22 @@ public async Task RemoveAsync(IAccount account)
 
         private async Task<IEnumerable<IAccount>> GetAccountsInternalAsync(ApiIds apiId, string homeAccountIdFilter = null)
         {
-            var accountsFromCache = await GetAccountsFromCacheAsync(apiId, homeAccountIdFilter).ConfigureAwait(false);
-            var accountsFromBroker = await GetAccountsFromBrokerAsync(homeAccountIdFilter).ConfigureAwait(false);
+            RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
```
jennyf19, line 148: can this be made from the ServiceBundle? I don't remember exactly.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -145,8 +145,22 @@ public async Task RemoveAsync(IAccount account)
...
+                   new AcquireTokenCommonParameters() { ApiId = apiId },
+                   requestContext,
+                   homeAccountIdFilter);
+
+            // a simple session consisting of a single call
+            var cacheSessionManager = new CacheSessionManager(
+                UserTokenCacheInternal,
+                authParameters);
+
+            var accountsFromCache = await cacheSessionManager.GetAccountsAsync().ConfigureAwait(false);
```
jennyf19, line 162: >cacheSessionManager [](start = 42, length = 19)

null check?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -42,9 +42,9 @@ public sealed partial class TokenCache : ITokenCacheInternal
                 requestParams.RequestContext.Logger.Info("ID Token not present in response. ");
             }
 
-            var tenantId = Authority
-                .CreateAuthority(requestParams.TenantUpdatedCanonicalAuthority.AuthorityInfo.CanonicalAuthority)
-                .TenantId;
+            //Authority.CreateAuthorityWithTenant(requestParams.Authority.AuthorityInfo, idToken?.TenantId);
```
jennyf19, line 45: remove?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -42,9 +42,9 @@ public sealed partial class TokenCache : ITokenCacheInternal
...
+
+            var tenantId = GetTenantId(idToken, requestParams);
```
jennyf19, line 47: >tenantId [](start = 16, length = 8)

nit: does this need a variable? it can be passed inline on line 71.

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AadPlugin.cs
```diff
@@ -27,25 +30,80 @@ public AadPlugin(IWamProxy wamProxy, IWebAccountProviderFactory webAccountProvid
...
+        /// handling of guest accounts: 
+        /// 
+        /// 1. Read the accounts from WAM.AADPlugin
+        /// 2. For each account, we need to find its home_account_id as the one from WAM may not be correct
+        /// 3. If we can find a cached account with the same LocalAccountId or UPN, use it
+        /// 4. If not, make a simple silent token request and use the client info provided
+        /// </summary>
+        public async Task<IReadOnlyList<IAccount>> GetAccountsAsync(
+            string clientId,
+            string authority,
```
jennyf19, line 44: nit: change to `clientID` to match interface

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AadPlugin.cs
```diff
@@ -27,25 +30,80 @@ public AadPlugin(IWamProxy wamProxy, IWebAccountProviderFactory webAccountProvid
...
+                            await ConvertToMsalAccountOrNullAsync(
+                                clientId,
+                                webAcc,
+                                instanceMetadata,
+                                cacheSessionManager,
+                                accountsFromCache).ConfigureAwait(false));
+
+                var msalAccounts = (await Task.WhenAll(msalAccountTasks).ConfigureAwait(false)).Where(a => a != null).ToList();
+
+                _logger.Info($"[WAM AAD Provider] GetAccountsAsync converted {wamAccounts.Count()} MSAL accounts");
```
jennyf19, line 88: assuming all the wamAccounts were converted...would it be helpful to know how many msalAccounts were created?

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamAadPluginTests.cs
```diff
@@ -0,0 +1,354 @@
+﻿using System;
```
jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿using System;
```

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamAadPluginTests.cs
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+        [TestMethod]
+        public async Task GetAccounts_WamAccounts_NoAuthority_Async()
+        {
+            // Arrange
+            using (MockHttpAndServiceBundle harness = CreateTestHarness())
+            {
+                var wamAccountProvider = new WebAccountProvider("id", "user1@contoso.com", null);
+                _webAccountProviderFactory.GetAccountProviderAsync("organizations").Returns(wamAccountProvider);
```
jennyf19, line 88: would we ever create the webaccountProviderFactory w/out the authority or tenant?

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
bgavrilMS: Yes, I think so. But GetAccounts is a bit strange, because it's not a request per se. Anyway, this is just some code copied around.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -155,13 +169,21 @@ private async Task<IEnumerable<IAccount>> GetAccountsInternalAsync(ApiIds apiId,
             return cacheAndBrokerAccounts;
         }
 
-        private async Task<IEnumerable<IAccount>> GetAccountsFromBrokerAsync(string homeAccountIdFilter)
+        private async Task<IEnumerable<IAccount>> GetAccountsFromBrokerAsync(
+            string homeAccountIdFilter,
+            ICacheSessionManager cacheSessionManager,
+            RequestContext requestContext)
```
jennyf19, line 175: >RequestContext requestContext [](start = 12, length = 29)

can be removed, as it's not used.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamAadPluginTests.cs
bgavrilMS: There's a special string you can pass in to get the web account provider of the "OS default account" - `https://login.windows.local`

This corresponds to the user from the "Your Info" settings page in Windows

image


https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/cb1361f5c235fdf893957ea145e299cf4686f0df/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WebAccountProviderFactory.cs#L24

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -145,8 +145,22 @@ public async Task RemoveAsync(IAccount account)
 
         private async Task<IEnumerable<IAccount>> GetAccountsInternalAsync(ApiIds apiId, string homeAccountIdFilter = null)
         {
-            var accountsFromCache = await GetAccountsFromCacheAsync(apiId, homeAccountIdFilter).ConfigureAwait(false);
-            var accountsFromBroker = await GetAccountsFromBrokerAsync(homeAccountIdFilter).ConfigureAwait(false);
+            RequestContext requestContext = CreateRequestContext(Guid.NewGuid());
+
+            var authParameters = new AuthenticationRequestParameters(
+                   ServiceBundle,
```
jennyf19, line 151: could you just call `CreateRequestParameters(new AcquireTokenCommonParameters() { ApiId = apiId }, CreateRequestContext(Guid.NewGuid()), UserTokenCacheInternal);` and pass it directly to the CacheSessionManager?

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
jennyf19: actually, you can remove it, as it's not used in GetAccountsFromBrokerAsync.

---
In reply to: [568137667](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2375#discussion_r568137667) [](ancestors = 568137667)

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AadPlugin.cs
```diff
@@ -27,25 +30,80 @@ public AadPlugin(IWamProxy wamProxy, IWebAccountProviderFactory webAccountProvid
...
+                                $"[WAM AAD Provider] Could not convert the WAM account {w.UserName} (id: {w.Id}) to an MSAL account because the Authority could not be found",
+                                $"[WAM AAD Provider] Could not convert the WAM account {w.Id} to an MSAL account because the Authority could not be found");
+
+                            return null;
+                        }
+                    })
+                    .Where(a => a != null);
+
+                var instanceMetadata = await instanceDiscoveryManager.GetMetadataEntryTryAvoidNetworkAsync(authority, webAccountEnvs, cacheSessionManager.RequestContext)
+                    .ConfigureAwait(false);
```
jennyf19, line 73: I think you can do this:
`cacheSessionManager.RequestContext.ServiceBundle.InstanceDiscoveryManager.GetMetadataEntryTryAvoidNetworkAsync(authority, webAccountEnvs, cacheSessionManager.RequestContext);`

so you don't have to pass the instanceDiscoveryManager around, you can get it from the RC.

jennyf19 (approved):
:shipit:

jennyf19:
few comments, and haven't gone through the WamAadPluginTests yet, but will later today. Looks good.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamAadPluginTests.cs
jennyf19: that's cool. 

i guess i'm asking more is this test an actual scenario, where we wouldn't pass the authority. just curious.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamAadPluginTests.cs
```diff
@@ -0,0 +1,354 @@
...
+                    harness.ServiceBundle.Config.AuthorityInfo.CanonicalAuthority,
+                    Arg.Any<IEnumerable<string>>(), rq)
+                    .Returns(CreateEntryForSingleAuthority(new Uri(TestConstants.AuthorityCommonTenant))); // user set this authority in config
+
+                // assume there is a cache account with the same UPN 
+                _cacheSessionManager.GetAccountsAsync().Returns(new[] {
+                    new Account(TestConstants.HomeAccountId, User1Upn, null, null )});
+
+                // Act
+                var accounts = await _aadPlugin.GetAccountsAsync(
```
jennyf19, line 173: >GetAccountsAsync [](start = 48, length = 16)

how does GetAccountsAsync know the right UPN via the aadPlugin?

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamAadPluginTests.cs
```diff
@@ -0,0 +1,354 @@
...
+        {
+            return new InstanceDiscoveryMetadataEntry()
+            {
+                Aliases = new[] { authority.Host },
+                PreferredCache = authority.Host,
+                PreferredNetwork = authority.Host
+            };
+        }
+    }
+
```
jennyf19, line 351: is there value in having a test w/multiple accounts?

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
bgavrilMS: Good practice for methdos that return lists or the sort is to return empty list instead of null, to avoid all these null checks. I will add them for defensive purposes.

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AadPlugin.cs
bgavrilMS: I agree, but I like this a bit more because of separation of concerns. RequestContext has a LOT of things inside, so whenever a method requires the RequestContext it's actually very difficult to understand what that method actually needs. This way, I can immediately tell that to GetAccounts, the broker needs to manipulate the cache and to get cloud aliases.

Issue #2150, opened 2020-10-29
[Feature Request] [L] Implement content provider for communication with the Android broker.


trwalke:
**Is your feature request related to a problem? Please describe.**
The current communication method that MSAL uses (Account Manager) with the broker is one of the broker's fallback mechanisms that uses IBinder  for IPC communication. One limitation here is that the Android OS allocates a maximum of 1MB buffer on RAM per process for IBinder transactions. This buffer size includes all the transactions within the app process like intents between activities, fragments and broadcast receivers etc. Given this, Android recommends to limit the transaction size to absolute minimum while using bound service. If the buffer size goes beyond 1MB either on the app process or on the Broker's background process, the OS throws a TransactionTooLargeException  or a DeadObjectException  by killing the remote process (broker's :auth process). 

The content provider implementation is a very recent update in broker applications so MSAL will need a way to fall back to the old implementation incase the user is on an older version of the broker.

**Describe the solution you'd like**
Implement the content provider in MSAL.NET so that MSAL can communicate with the broker without running into IBinder errors. 

Deliverables: 

- Implement Content provider in MSAL for broker communication.

- Enable MSAL to fallback on the account manager communication strategy if the content provider fails.

**Additional context**
[Xamarin Content Provider Documentation](https://docs.microsoft.com/en-us/xamarin/android/platform/content-providers/)
[Broker Content Provider Spec](https://identitydivision.visualstudio.com/DefaultCollection/DevEx/_git/AuthLibrariesApiReview/pullrequest/1554?_a=files&path=%2FBrokerContract%2FContentProviderAndroidBroker.md&discussionId=15009)

bgavrilMS:
Another support incident related to AddAccount API being banned in enterprises via Android 4 Work.

bgavrilMS:
@trwalke - are we implementing this ? Should we schedule it for 4.26?

jmprieur:
Outstanding work:
- [ ] address PR review
- [ ] re-test in the test app
- [ ] update the manual test
- [ ] test with the Xamarin sample (with and without broker)

trwalke:
Released in MSAL 4.27.0
https://www.nuget.org/packages/Microsoft.Identity.Client/

PR #2380, opened 2021-02-02
Content Provider for Android Broker

trwalke:
This is the implementation of the content provider communication interface in android broker.

The AndroidBroker.CS is the old broker that has the same functionality overall so I suggest starting with the AndroidContentProviderBroker.cs and the AndroidBrokerOrchestrator.cs classes.
 
Will be adding additional refactoring to clean up code further where possible.

fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2150


src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBroker.cs
```diff
@@ -19,17 +19,17 @@
...
 using System.Linq;
+using Android.Accounts;
+using Java.Util.Concurrent;
+using OperationCanceledException = Android.Accounts.OperationCanceledException;
+using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Platforms.Android.Broker
 {
     [AndroidNative.Runtime.Preserve(AllMembers = true)]
     internal class AndroidBroker : IBroker
```
bgavrilMS, line 30: This type should be renamed to AndroidAccountManagerBroker

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerOrchestrator.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Platforms.Android.Broker
+{
+    internal class AndroidBrokerOrchestrator
+    {
+        CoreUIParent _uIParent;
```
bgavrilMS, line 23: Please add visibility modifiers on all of members. `private readonly`

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerOrchestrator.cs
```diff
@@ -0,0 +1,84 @@
...
+            // Return a default broker in case no broker is installed to handle install url
+            return new AndroidContentProviderBroker(_uIParent, _logger);
+        }
+
+        internal IBroker GetBroker()
+        {
+            return Task.Run(async () => await GetInstalledBrokerAsync().ConfigureAwait(false)).Result;
+        }
+
+        public bool IsBrokerInstalledAndInvokable()
```
bgavrilMS, line 74: Please review all your public / internal methods. Most do not have to be internal. Try to use private methods as much as possible. 

In this case, it's not even worth having a separate method...

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerOrchestrator.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            return _brokerHelper.IsBrokerInstalledAndInvokable();
+        }
+
+        internal static void SetBrokerResult(Intent data, int resultCode, ICoreLogger unreliableLogger)
```
bgavrilMS, line 79: is it really needed?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+
+        public async void InitiateBrokerHandshakeAsync()
+        {
+            using (_logger.LogMethodDuration())
+            {
+                try
+                {
+                    Bundle HandshakeBundleResult = await GetHandshakeBundleResultFromBrokerAsync().ConfigureAwait(false);
+
+                    _negotiatedBrokerProtocalKey = _brokerHelper.GetProtocalKeyFromHandshakeResult(HandshakeBundleResult);
```
bgavrilMS, line 51: Please install an extension like 
image

It will help auto-correct these quickly typed names. I have this problem as well.

`Protocol`

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+            return _brokerHelper.IsBrokerInstalledAndInvokable();
+        }
+
+        public async void InitiateBrokerHandshakeAsync()
+        {
+            using (_logger.LogMethodDuration())
+            {
+                try
+                {
+                    Bundle HandshakeBundleResult = await GetHandshakeBundleResultFromBrokerAsync().ConfigureAwait(false);
```
bgavrilMS, line 49: naming style - camelCase

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+
+                    _negotiatedBrokerProtocalKey = _brokerHelper.GetProtocalKeyFromHandshakeResult(HandshakeBundleResult);
+                }
+                catch
+                {
+                    throw;
```
bgavrilMS, line 55: don't catch and throw.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerOrchestrator.cs
```diff
@@ -0,0 +1,84 @@
...
+                    _logger.Info("Unable to handshake with Content Provider Broker. Attempting handshake with Account Manager Broker.");
+
+                    try
+                    {
+                        var broker = new AndroidBroker(_uIParent, _logger);
+                        await broker.InitiateBrokerHandshakeAsync(_uIParent.CallerActivity).ConfigureAwait(false);
+                    }
+                    catch
+                    {
+                        _logger.Info("Unable to connect to any of the brokers.");
```
bgavrilMS, line 60: it's good practice to catch the actual exception and to log it.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerOrchestrator.cs
```diff
@@ -0,0 +1,84 @@
...
+                {
+                    var broker = new AndroidContentProviderBroker(_uIParent, _logger);
+
+                    if (!s_contentProviderIsAvailable)
+                    {
+                        broker.InitiateBrokerHandshakeAsync();
+                        s_contentProviderIsAvailable = true;
+                    }
+                }
+                catch
```
bgavrilMS, line 49: catch and log

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+            var bundle = _brokerHelper.GetHandshakeOperationBundle();
+            var OperationBundleJSON = _brokerHelper.SearializeBundleToJSON(bundle);
+            return await PerformContentResolverOperationAsync(ContentResolverOperation.hello, OperationBundleJSON).ConfigureAwait(false);
+        }
+
+        public async Task<MsalTokenResponse> AcquireTokenInteractiveAsync(
+            AuthenticationRequestParameters authenticationRequestParameters,
+            AcquireTokenInteractiveParameters acquireTokenInteractiveParameters)
+        {
+            _brokerHelper.CheckPowerOptimizationStatus();
```
bgavrilMS, line 71: Is this needed?! I thought not being affected by power opt was one of the main reasons for using content prov

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+
+            InitiateBrokerHandshakeAsync();
+
+            var brokerRequest = PrepareInteractiveBrokerRequest(authenticationRequestParameters, acquireTokenInteractiveParameters);
```
bgavrilMS, line 75: CreateInteractiveRequest

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+
+            return await PerformAcquireTokenInteractiveAsync(brokerRequest).ConfigureAwait(false);
+        }
+
+        private BrokerRequest PrepareInteractiveBrokerRequest(AuthenticationRequestParameters authenticationRequestParameters,
```
bgavrilMS, line 80: I would avoid these very small 1-2 lines of code methods. They fragment the code too much. Methods should also not be longer than ~50 lines or so (best description I know around this is `you should not have to scroll when reading the code of a method`)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+                                                              AcquireTokenInteractiveParameters acquireTokenInteractiveParameters)
+        {
+            BrokerRequest brokerRequest = BrokerRequest.FromInteractiveParameters(
+                authenticationRequestParameters, acquireTokenInteractiveParameters);
+
+            // There can only be 1 broker request at a time so keep track of the correlation id
+            AndroidBrokerHelper.CorrelationId = brokerRequest.CorrelationId;
```
bgavrilMS, line 87: I'd actually move all the statics to a separate class to simplify this helper a bit.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+                string silentResult = await PerformAcquireTokenSilentFromBrokerAsync(brokerRequest).ConfigureAwait(false);
+                return _brokerHelper.HandleSilentAuthenticationResult(silentResult, brokerRequest.CorrelationId);
+            }
+        }
+
+        /// <summary>
+        /// This method is only used for Silent authentication requests so that we can check to see if an account exists in the account manager before
+        /// sending the silent request to the broker. 
+        /// </summary>
+        public async Task<BrokerRequest> UpdateRequestWithAccountInfoAsync(BrokerRequest brokerRequest)
```
bgavrilMS, line 197: not public, also probably not needed.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+        {
+            var accountData = await GetBrokerAccountDataAsync(brokerRequest).ConfigureAwait(false);
+
+            return _brokerHelper.UpdateBrokerRequestWithAccountData(accountData, brokerRequest);
+        }
+
+        public async Task<string> PerformAcquireTokenSilentFromBrokerAsync(BrokerRequest brokerRequest)
+        {
+            Bundle silentOperationBundle = _brokerHelper.CreateSilentBrokerBundle(brokerRequest);
+            var OperationBundleJSON = _brokerHelper.SearializeBundleToJSON(silentOperationBundle);
```
bgavrilMS, line 207: camelCase vars

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        public async Task<IEnumerable<IAccount>> GetAccountsAsync(string clientID, string redirectUri)
+        {
+            using (_logger.LogMethodDuration())
+            {
+                if (!IsBrokerInstalledAndInvokable())
+                {
+                    _logger.Warning("Android broker is either not installed or is not reachable so no accounts will be returned. ");
+                    return new List<IAccount>();
```
bgavrilMS, line 219: nit: Enumerable.Empty<IAccount>()

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+
+        private async Task<IEnumerable<IAccount>> PerformGetAccountsAsync(string clientID, string redirectUri)
+        {
+            BrokerRequest brokerRequest = new BrokerRequest() { ClientId = clientID, RedirectUri = new Uri(redirectUri) };
+
+            try
+            {
+                InitiateBrokerHandshakeAsync();
```
bgavrilMS, line 232: Should this be cached? If it worked once, should we do it over and over?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,349 @@
...
+                {
+                    _logger.Warning("Android broker is either not installed or not reachable so no accounts will be removed. ");
+                    return;
+                }
+
+                await PerformRemoveAccountsAsync(applicationConfiguration, account).ConfigureAwait(false);
+            }
+        }
+
+        private async Task PerformRemoveAccountsAsync(IApplicationConfiguration applicationConfiguration, IAccount account)
```
bgavrilMS, line 267: nit: generally I prefer the naming `RemoveAccountsInternalAsync()` or `RemoveAccountsImplAsync()` - the reasoning is that it keeps this drop down nicely grouped. I saw this naming patter in the C++ team code.

image
image

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -33,22 +41,42 @@ internal class AndroidBrokerHelper
...
         private readonly Context _androidContext;
 
+        // When the broker responds, we cannot correlate back to a started task. 
+        // So we make a simplifying assumption - only one broker open session can exist at a time
+        // This semaphore is static to enforce this
+        public static SemaphoreSlim ReadyForResponse { get; set; } = new SemaphoreSlim(0);
+
         // Important: this object MUST be accessed on a background thread. Android will check this and throw otherwise.
-        private readonly AccountManager _androidAccountManager;
+        public AccountManager AndroidAccountManager { get; }
```
bgavrilMS, line 50: Violates the separation of concerns principle - Only the AccountManagerBroker should know about this. This helper class should encapsulate all the common operations instead.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -33,22 +41,42 @@ internal class AndroidBrokerHelper
...
         private readonly ICoreLogger _logger;
 
-        private const long AccountManagerTimeoutSeconds = 5 * 60;
+        public static MsalTokenResponse InteractiveBrokerTokenResponse { get; set; } = null;
+        //Since the correlation ID is not returned from the broker response, it must be stored at the beginning of the authentication call and re-injected into the response at the end.
+        public static string CorrelationId { get; set; }
+
+        public string NegotiatedBrokerProtocalKey { get; set; }
```
bgavrilMS, line 57: Should not be here.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
bgavrilMS: I think you need to review this class entirely to make it more SRP (single resp principle) friendly

bgavrilMS (changes requested):
Generally looks good, needs some cleanup.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
trwalke: Definitely need this

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
trwalke: You are correct. not needed.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
trwalke: It is possible to cache. i spoke with the android team adn they provided guidance on how to do this. I created a seperate issue here for this
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2390

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
trwalke: So I did this and named it AndroidBrokerStaticHelper. is that name ok?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,455 @@
...
+        {
+            _parentActivity = uiParent?.Activity;
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            AuthenticationContinuationHelper.LastRequestLogger = _logger;
+            _brokerHelper = new AndroidBrokerHelper(Application.Context, logger);
+        }
+
+        public bool IsBrokerInstalledAndInvokable()
+        {
+            return _brokerHelper.IsBrokerInstalledAndInvokable();
```
jennyf19, line 51: >_brokerHelper [](start = 19, length = 13)

null check here?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,455 @@
...
+
+            CheckPowerOptimizationStatus();
+
+            AndroidBrokerStaticHelper.InteractiveBrokerTokenResponse = null;
+
+            BrokerRequest brokerRequest = BrokerRequest.FromInteractiveParameters(
+                authenticationRequestParameters, acquireTokenInteractiveParameters);
+
+            // There can only be 1 broker request at a time so keep track of the correlation id
+            s_correlationId = brokerRequest.CorrelationId;
```
jennyf19, line 67: >s_correlationId [](start = 12, length = 15)

how is this keeping track of the request? i don't see it used in the class after being set here.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,455 @@
...
+            CheckPowerOptimizationStatus();
+
+            BrokerRequest brokerRequest = BrokerRequest.FromSilentParameters(
+                authenticationRequestParameters, acquireTokenSilentParameters);
+
+            try
+            {
+                await InitiateBrokerHandshakeAsync().ConfigureAwait(false);
+                var androidBrokerTokenResponse = await AcquireTokenSilentViaBrokerAsync(brokerRequest).ConfigureAwait(false);
+                return androidBrokerTokenResponse;
```
jennyf19, line 100: >androidBrokerTokenResponse [](start = 23, length = 26)

nit: just return the AcquireTokenSilentBrokerAsync(), no need for the variable

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,455 @@
...
+        }
+
+        private async Task AcquireTokenInteractiveViaBrokerAsync(BrokerRequest brokerRequest)
+        {
+            using (_logger.LogMethodDuration())
+            {
+                // onActivityResult will receive the response for this activity.
+                // Launching this activity will switch to the broker app.
+
+                _logger.Verbose("Starting Android Broker interactive authentication. ");
```
jennyf19, line 117: like how neha did [Region discovery] or something at the start of the log messages for regional, i think you did have something similar for android, but might be useful on these log messages to have a short title like [Android Broker]...idk.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
jennyf19: also, you no need for the broker request variable either.

---
In reply to: [570491051](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r570491051) [](ancestors = 570491051)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,455 @@
...
+            }
+            catch (Exception ex)
+            {
+                _logger.ErrorPiiWithPrefix(ex, "Android broker silent invocation failed. ");
+                _brokerHelper.HandleBrokerOperationError(ex);
+                throw;
+            }
+        }
+
+        private async Task AcquireTokenInteractiveViaBrokerAsync(BrokerRequest brokerRequest)
```
jennyf19, line 110: >BrokerRequest brokerRequest [](start = 65, length = 27)

for this and the silent request, you might not need to pass in the request, just create it in the method, it's not used in the outer method anyway.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,455 @@
...
+            else
+            {
+                _logger.Info("Intent created from BundleResult is null. ");
+                throw new MsalClientException(MsalError.NullIntentReturnedFromAndroidBroker, MsalErrorMessage.NullIntentReturnedFromBroker);
+            }
+
+            return CreateInteractiveBrokerIntent(brokerRequest, interactiventent);
+        }
+
+        private Intent CreateInteractiveBrokerIntent(BrokerRequest brokerRequest, Intent brokerIntent)
```
jennyf19, line 174: >brokerRequest [](start = 67, length = 13)

i think brokerRequest is only used here, and as it's static, can you just create it in this class and use it? no point in pass it around.

jennyf19: this is assuming you don't need that correlation_id, which isn't used anywhere

---
In reply to: [570495116](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r570495116) [](ancestors = 570495116)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,455 @@
...
+        }
+
+        private Intent CreateIntentFromInteractiveBundle(Bundle bundle, BrokerRequest brokerRequest)
+        {
+            if (bundle == null)
+            {
+                _logger.Info("Android account manager didn't return any results for interactive broker request. ");
+            }
+
+            Intent interactiventent = (Intent)bundle?.GetParcelable(AccountManager.KeyIntent);
```
jennyf19, line 156: >interactiventen [](start = 19, length = 15)

spelling

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
jennyf19: can you get rid of this method and just put the logic above?

---
In reply to: [570496371](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r570496371) [](ancestors = 570496371,570495116)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
jennyf19: I agree with Bogdan on this.

---
In reply to: [569703588](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r569703588) [](ancestors = 569703588)

tests/devapps/XamarinDev/XamarinDev.Android/XamarinDev.Android.csproj
```diff
@@ -61,7 +61,12 @@
     <PackageReference Include="Xamarin.Android.Support.v7.MediaRouter" Version="28.0.0.3" />
     <PackageReference Include="Xamarin.Android.Support.Core.Utils" Version="28.0.0.3" />
     <PackageReference Include="Xamarin.Android.Support.CustomTabs" Version="28.0.0.3" />
-    <PackageReference Include="Xamarin.Essentials" Version="1.0.1" />
+    <PackageReference Include="Xamarin.Essentials" Version="1.6.0" />
+	<PackageReference Include="Xamarin.AndroidX.Lifecycle.LiveData" Version="2.1.0" />
+    <PackageReference Include="Xamarin.Google.Android.Material" Version="1.0.0" />
+    <PackageReference Include="Xamarin.AndroidX.Legacy.Support.V4" Version="1.0.0" />
+    <PackageReference Include="Xamarin.AndroidX.MediaRouter" Version="1.1.0" />
```
jennyf19, line 68: do we need all these additional packages?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
trwalke: I dont think it is needed as the broker helper is always instantiated in the broker's constructor.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
trwalke: this is using the wrong variable, will update

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs
```diff
@@ -210,7 +210,9 @@ private static bool IsChromeEnabled()
 
         public override IBroker CreateBroker(ApplicationConfiguration appConfig, CoreUIParent uiParent)
         {
-            return new AndroidBroker(uiParent, Logger);
+            AndroidBrokerOrchestrator androidBrokerOrchestrator = new AndroidBrokerOrchestrator(uiParent, Logger);
```
bgavrilMS, line 213: 1. `orchestrator` is a type that manages other types and knows a lot of the high level flow, but doesn't know low level details. Your class isn't an orchestrator. It's a factory. I'd name it as such.
2. It's ok for factories to be static..
```AndroidBrokerFactory.CreateBroker(uiParent, logger)```

File: src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs (hunk 25 above)
Comment (trwalke) [reply src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs:110]: If I create it here then I would need to pass in 2 variables to this function in order to create the brokerRequest: "(authenticationRequestParameters, acquireTokenSilentParameters)"

so I figured it was better to create the request initially and only pass it around.

File: src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs [hunk 30]
```diff
@@ -227,6 +229,11 @@ public override bool CanBrokerSupportSilentAuth()
             }
         }
 
+        public override bool IsBrokerInstalledAndInvokable()
+        {
+            return base.IsBrokerInstalledAndInvokable();
```
bgavrilMS, line 234: base.isBrokerInstalledAndInvokable just returns false no?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
trwalke: It is possible for there to be multiple silent requests happening at once so it is better to keep it as a per request basis. Also, the reason I moved this to its own method is to make the code more readable. Moving low level logic to its own function makes it easier to reuse later and makes reading/understanding the method using it a little bit easier

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs
trwalke: ah, yes, this is dead code. removing

tests/devapps/XamarinDev/XamarinDev.Android/XamarinDev.Android.csproj
trwalke: These are to update the test app to androidX. Figured it is worth updating to the latest to test out the flows.
However, not needed

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -29,26 +36,34 @@ internal class AndroidBrokerHelper
...
-        private const long AccountManagerTimeoutSeconds = 5 * 60;
-
         public AndroidBrokerHelper(Context androidContext, ICoreLogger logger)
         {
             _androidContext = androidContext ?? throw new ArgumentNullException(nameof(androidContext));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
             _logger.Verbose("Getting the Android context for broker request. ");
-            _androidAccountManager = AccountManager.Get(_androidContext);
+            AndroidAccountManager = AccountManager.Get(_androidContext);
```
jennyf19, line 52: >AccountManager [](start = 36, length = 14)

I get a warning that this `AccountManager` is only available on Android 9, not 10.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+using Microsoft.Identity.Json.Linq;
+using Microsoft.Identity.Client.Internal.Requests;
+using Microsoft.Identity.Client.ApiConfig.Parameters;
+using Microsoft.Identity.Client.Http;
+using System.Net;
+using Android.OS;
+using System.Linq;
+using Android.Accounts;
+using Java.Util.Concurrent;
+using OperationCanceledException = Android.Accounts.OperationCanceledException;
```
jennyf19, line 24: remove. this doesn't seem to be used

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+using Microsoft.Identity.Client.Instance.Discovery;
+
+namespace Microsoft.Identity.Client.Platforms.Android.Broker
+{
+    [AndroidNative.Runtime.Preserve(AllMembers = true)]
+    internal class AndroidAccountManagerBroker : IBroker
+    {
+        private long AccountManagerTimeoutSeconds { get; } = 5 * 60;
+        //Since the correlation ID is not returned from the broker response, it must be stored at the beginning of the authentication call and re-injected into the response at the end.
+        private readonly AndroidBrokerHelper _brokerHelper;
```
jennyf19, line 36: this comment doesn't make sense here..maybe move to wear the logic for this happens?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
trwalke: Ill double check the availability

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+            {
+                try
+                {
+                    Bundle helloRequestBundle = _brokerHelper.CreateHandShakeOperationBundle();
+
+                    Bundle helloRequestResult = await PerformAccountManagerOperationAsync(helloRequestBundle).ConfigureAwait(false);
+
+                    if (helloRequestResult != null)
+                    {
+                        var bpKey = helloRequestResult?.GetString(BrokerConstants.NegotiatedBPVersionKey);
```
jennyf19, line 341: >? [](start = 54, length = 1)

can remove, already checking for null above.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+                Intent brokerIntent = await GetIntentForInteractiveBrokerRequestAsync(brokerRequest, _parentActivity).ConfigureAwait(false);
+
+                if (brokerIntent != null)
+                {
+                    _brokerHelper.LaunchInteractiveActivity(_parentActivity, brokerIntent);
+                }
+            }
+        }
+
+        private async Task<Intent> GetIntentForInteractiveBrokerRequestAsync(BrokerRequest brokerRequest, Activity callerActivity)
```
jennyf19, line 125: >, Activity callerActivity [](start = 104, length = 25)

remove. unused param

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+            try
+            {
+                // Callback is not passed since it is making a blocking call to get
+                // intent. Activity needs to be launched from calling app
+                // to get the calling app's meta data if needed at BrokerActivity.
+
+                Bundle addAccountOptions = new Bundle();
+                addAccountOptions.PutString(BrokerConstants.BrokerAccountManagerOperationKey, BrokerConstants.GetIntentForInteractiveRequest);
+
+                Bundle accountManagerResult = await PerformAccountManagerOperationAsync(addAccountOptions).ConfigureAwait(false);
```
jennyf19, line 136: >erformAccountManagerOperationAsync [](start = 53, length = 34)

we call this a lot, in the handshake and then again in the interactive call...is it needed in both places? i guess yes, but seems costly.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
jennyf19: true. didn't see that. thanks.

---
In reply to: [570590223](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r570590223) [](ancestors = 570590223)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+            if (string.IsNullOrEmpty(accounts))
+            {
+                _logger.Info("Android account manager didn't return any accounts. ");
+                throw new MsalUiRequiredException(MsalError.NoAndroidBrokerAccountFound, MsalErrorMessage.NoAndroidBrokerAccountFound);
+            }
+
+            return _brokerHelper.UpdateBrokerRequestWithAccountData(accounts, brokerRequest);
+        }
+
+        private string GetBrokerAccounts(BrokerRequest brokerRequest, Activity callerActivity)
```
jennyf19, line 232: > Activity callerActivity [](start = 69, length = 24)

remove. unused param

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+
+            _logger.Info("Android Broker didn't return any results. ");
+            return null;
+        }
+
+        /// <summary>
+        /// This method is only used for Silent authentication requests so that we can check to see if an account exists in the account manager before
+        /// sending the silent request to the broker. 
+        /// </summary>
+        private BrokerRequest UpdateRequestWithAccountInfo(BrokerRequest brokerRequest, Activity callerActivity)
```
jennyf19, line 219: >Activity callerActivity [](start = 88, length = 23)

remove. unused param.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+        }
+
+        private string GetBrokerAccounts(BrokerRequest brokerRequest, Activity callerActivity)
+        {
+            Bundle getAccountsBundle = _brokerHelper.CreateBrokerAccountBundle(brokerRequest);
+            getAccountsBundle.PutString(BrokerConstants.BrokerAccountManagerOperationKey, BrokerConstants.GetAccounts);
+
+            //This operation will acquire all of the accounts in the account manager for the given client ID
+            Bundle bundleResult = PerformAccountManagerOperationAsync(getAccountsBundle).Result;
+            return bundleResult?.GetString(BrokerConstants.BrokerAccounts);
```
jennyf19, line 239: we already did this in `InitiateBrokerHandshakeAysnc()` on line 96ish...is there a way to only need to call `PerformAccountManagerOpAsync()` once?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformProxy.cs
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -208,31 +120,30 @@ public void CheckForBrokerAccountInfoInAccountManager(BrokerRequest brokerReques
             string homeAccountId = brokerRequest.HomeAccountId;
             string localAccountId = brokerRequest.LocalAccountId;
 
-            if (!string.IsNullOrEmpty(accounts))
+            if (!string.IsNullOrEmpty(accountData))
```
jennyf19, line 123: I don't think this is needed as you already check above.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -208,31 +120,30 @@ public void CheckForBrokerAccountInfoInAccountManager(BrokerRequest brokerReques
...
+                    var accountInfoLocalAccountID = accountInfo[BrokerResponseConst.LocalAccountId]?.ToString();
 
-                    if (string.Equals(accountData[BrokerResponseConst.UserName].ToString(), username, StringComparison.OrdinalIgnoreCase))
+                    if (string.Equals(accountInfo[BrokerResponseConst.UserName].ToString(), username, StringComparison.OrdinalIgnoreCase))
                     {
                         // TODO: broker request should be immutable!
-                        brokerRequest.HomeAccountId = accountDataHomeAccountID;
-                        brokerRequest.LocalAccountId = accountDataLocalAccountID;
+                        brokerRequest.HomeAccountId = accountInfoHomeAccountID;
+                        brokerRequest.LocalAccountId = accountInfoLocalAccountID;
```
jennyf19, line 137: why not use the variables created before and is it safe to override the value here?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -208,31 +120,30 @@ public void CheckForBrokerAccountInfoInAccountManager(BrokerRequest brokerReques
...
                         _logger.Info("Found broker account in Android account manager using the provided login hint. ");
-                        return;
+                        return brokerRequest;
                     }
 
-                    if (string.Equals(accountDataHomeAccountID, homeAccountId, StringComparison.Ordinal) &&
-                         string.Equals(accountDataLocalAccountID, localAccountId, StringComparison.Ordinal))
+                    if (string.Equals(accountInfoHomeAccountID, homeAccountId, StringComparison.Ordinal) &&
+                         string.Equals(accountInfoLocalAccountID, localAccountId, StringComparison.Ordinal))
```
jennyf19, line 143: when would homeacount and localacocunt id be the same, and what does that mean? is there a reason to check for the user or just use home/local account?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -241,50 +152,25 @@ public void CheckForBrokerAccountInfoInAccountManager(BrokerRequest brokerReques
...
-                _logger.Info("Android account manager didn't return any accounts. ");
-            }
-
             List<IAccount> brokerAccounts = new List<IAccount>();
 
-            if (!string.IsNullOrEmpty(accounts))
+            if (!string.IsNullOrEmpty(accountData))
             {
-                dynamic authResult = JArray.Parse(accounts);
+                dynamic authResult = JArray.Parse(accountData);
```
jennyf19, line 164: >uthResult [](start = 25, length = 9)

nit: is there a better name for this? we use authResult a lot, and i don't think this is an authResult as we typically use it.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -241,50 +152,25 @@ public void CheckForBrokerAccountInfoInAccountManager(BrokerRequest brokerReques
...
                 foreach (JObject account in authResult)
                 {
                     if (account.ContainsKey(BrokerResponseConst.Account))
                     {
-                        var accountData = account[BrokerResponseConst.Account];
-                        var homeAccountID = accountData.Value<string>(BrokerResponseConst.HomeAccountId) ?? "";
-                        var userName = accountData.Value<string>(BrokerResponseConst.UserName) ?? "";
-                        var environment = accountData.Value<string>(BrokerResponseConst.Environment) ?? "";
+                        var accountInfo = account[BrokerResponseConst.Account];
+                        var homeAccountID = accountInfo.Value<string>(BrokerResponseConst.HomeAccountId) ?? "";
```
jennyf19, line 171: nit; remove variables and move directly to the method

jennyf19, line 171: > "" [](start = 107, length = 3)

nit: string.Empty

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerStaticHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+using Android.Content;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Internal.Broker;
+using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Json.Linq;
+
+namespace Microsoft.Identity.Client.Platforms.Android.Broker
+{
+    internal class AndroidBrokerStaticHelper
```
jennyf19, line 19: add `static` keyword

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerStaticHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+                            Error = MsalError.AuthenticationCanceledError,
+                            ErrorDescription = MsalErrorMessage.AuthenticationCanceled,
+                        };
+                        break;
+                    case (int)BrokerResponseCode.BrowserCodeError:
+                        unreliableLogger?.Info("Response received - error. ");
+
+                        dynamic errorResult = JObject.Parse(data.GetStringExtra(BrokerConstants.BrokerResultV2));
+                        string error = null;
+                        string errorDescription = null;
```
jennyf19, line 63: nit: remove null assignment on both.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerStaticHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+                        else
+                        {
+                            error = BrokerConstants.BrokerUnknownErrorCode;
+                            errorDescription = "Error Code received, but no error could be extracted. ";
+                            unreliableLogger?.Error("Error response received, but not error could be extracted. ");
+                        }
+
+                        var httpResponse = new HttpResponse();
+                        //TODO: figure out how to get status code properly deserialized from JObject
+                        httpResponse.Body = errorResult[BrokerResponseConst.BrokerHttpBody]?.ToString();
```
jennyf19, line 81: >errorResult [](start = 44, length = 11)

this will through a null ref, as it's outside the block.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+using AndroidUri = Android.Net.Uri;
+
+namespace Microsoft.Identity.Client.Platforms.Android.Broker
+{
+    [AndroidNative.Runtime.Preserve(AllMembers = true)]
+    internal class AndroidContentProviderBroker : IBroker
+    {
+        private readonly AndroidBrokerHelper _brokerHelper;
+        private readonly ICoreLogger _logger;
+        private Activity _parentActivity;
```
jennyf19, line 32: readonly?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+        private string _negotiatedBrokerProtocolKey = String.Empty;
```
jennyf19, line 33: >S [](start = 54, length = 1)

small s

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+            AuthenticationContinuationHelper.LastRequestLogger = _logger;
+            _brokerHelper = new AndroidBrokerHelper(Application.Context, logger);
+        }
+
+        public bool IsBrokerInstalledAndInvokable()
+        {
+            return _brokerHelper.IsBrokerInstalledAndInvokable();
+        }
+
+        public async void InitiateBrokerHandShakeAsync()
```
jennyf19, line 48: >void [](start = 21, length = 4)

return `Task`

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+
+            _logger.Error(errorDescription);
+            throw new MsalClientException(errorCode, errorDescription);
+        }
+
+        public async Task<MsalTokenResponse> AcquireTokenInteractiveAsync(
+            AuthenticationRequestParameters authenticationRequestParameters,
+            AcquireTokenInteractiveParameters acquireTokenInteractiveParameters)
+        {
+            InitiateBrokerHandShakeAsync();
```
jennyf19, line 99: await this call

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+
+                _brokerHelper.LaunchInteractiveActivity(_parentActivity, interactiveIntent);
+            }
+        }
+
+        private Intent CreateInteractiveBrokerIntent(BrokerRequest brokerRequest, Bundle bundleResult)
+        {
+            string packageName = bundleResult.GetString("broker.package.name");
+            string className = bundleResult.GetString("broker.activity.name");
+            string uid = bundleResult.GetString("caller.info.uid");
```
jennyf19, line 146: >string uid [](start = 12, length = 10)

remove? not used.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+            _logger.InfoPii("GetInteractiveBrokerBundle: " + brokerRequestJson, "Enable PII to see the broker request. ");
+            return bundle;
+        }
+
+        public async Task<MsalTokenResponse> AcquireTokenSilentAsync(
+            AuthenticationRequestParameters authenticationRequestParameters,
+            AcquireTokenSilentParameters acquireTokenSilentParameters)
+        {
+            InitiateBrokerHandShakeAsync();
+
```
jennyf19, line 181: await this call.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+            BrokerRequest brokerRequest = BrokerRequest.FromSilentParameters(
```
jennyf19, line 182: this seems the exact same as the AndroidAccoutnManagerBroker class...oh wait...are these two different 'android broker" flavors? i have lots of quesitons for tomorrow. :)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+            {
+                return _brokerHelper.GetSilentResultFromBundle(silentOperationBundleResult);
+            }
+
+            _logger.Info("Android Broker didn't return any results. ");
+            return null;
+        }
+
+        public async Task<IReadOnlyList<IAccount>> GetAccountsAsync(
+            string clientID,
```
jennyf19, line 233: >clientID [](start = 19, length = 8)

nit: rename to `clientId` to match interface

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+        private async Task<string> GetBrokerAccountDataAsync(BrokerRequest brokerRequest)
+        {
+            var getAccountsBundle = _brokerHelper.CreateBrokerAccountBundle(brokerRequest);
+            var operationBundleJSON = SearializeBundleToJSON(getAccountsBundle);
+            var bundleResult = await PerformContentResolverOperationAsync(ContentResolverOperation.getAccounts, operationBundleJSON).ConfigureAwait(false);
+
+            return bundleResult?.GetString(BrokerConstants.BrokerAccounts);
+        }
+
+        public async Task RemoveAccountAsync(ApplicationConfiguration applicationConfiguration, IAccount account)
```
jennyf19, line 278: >applicationConfiguration [](start = 70, length = 24)

rename to appConfig to match interface

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
jennyf19: i asked this in the accountManager class...we definitely need to cache it. i would say that should happen before we ship this. let's discuss.

---
In reply to: [569855972](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r569855972) [](ancestors = 569855972)

jennyf19: also needs to be awaited.

---
In reply to: [570655066](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r570655066) [](ancestors = 570655066,569855972)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
jennyf19: agree w/bogdan, also i would remove this method, it's not really doing anything and it's not being test separately, so just move the try/catch into RemoveAccountAsync

---
In reply to: [569710387](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r569710387) [](ancestors = 569710387)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerConstants.cs
```diff
@@ -182,5 +182,19 @@ internal static class BrokerConstants
...
+
+        public const string CompanyPortalPackageName = "com.microsoft.windowsintune.companyportal";
+        public const string CompanyPortalContentProviderUri = "content://com.microsoft.windowsintune.companyportal.microsoft.identity.broker";
+        public const string MsAuthenticatorContentProviderUri = "content://com.azure.authenticator.microsoft.identity.broker";
+    }
+
+    //Content provider operation
+    internal enum ContentResolverOperation
+    {
+        hello,
```
jennyf19, line 194: nit: add a small comment on what "hello" is, as it's specific to the android broker.

tests/devapps/XamarinDev/XamarinDev/App.xaml.cs
```diff
@@ -14,7 +14,7 @@ public partial class App : Application
 
         public static object RootViewController { get; set; }
 #if !IS_APPCENTER_BUILD
-        public const string DefaultClientId = "4a1aa1d5-c567-49d0-ad0b-cd957a47f842"; // in msidentity-samples-testing tenant -> PublicClientSample
+        public const string DefaultClientId = "16dab2ba-145d-4b1b-8569-bf4b9aed4dc8"; // in msidentity-samples-testing tenant -> PublicClientSample
```
jennyf19, line 17: >16dab2ba-145d-4b1b-8569-bf4b9aed4dc8" [](start = 47, length = 37)

is this okay to change? will everything still work for all the tests, including iOS? If not, maybe include a separate android broker id.

tests/devapps/XamarinDev/XamarinDev.Android/XamarinDev.Android.csproj
jennyf19: I'm asking more because if customers were to update MSAL, would they also have to install these packages. if they are not part of the broker updates, then i think it's okay. 

---
In reply to: [570607603](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2380#discussion_r570607603) [](ancestors = 570607603)

jennyf19: @neha-bhargava can we double check what gets added when updating the Xamarin samples? thanks.

neha-bhargava: @jennyf19 I updated the Xamarin samples sometime back for testing and these updates looks fine for androidX.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
trwalke: yes it is. However caching for the handshake will happen later

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
trwalke: Not yet. Once we have handshake caching we can remove

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
trwalke: Yes it is safe to overwrite. for the silent request, these values start out as null because we need to fetch them from the broker first. Then we populate

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
trwalke: Discussed in meeting

tests/devapps/XamarinDev/XamarinDev/App.xaml.cs
trwalke: not needed. will revert

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
trwalke: I checked this out and can confirm that this is accessible when targeting android 10. Might be a VS problem on your end.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+        private ContentResolver GetContentResolver()
+        {
+            if (_parentActivity == null)
+            {
+                return Application.Context.ContentResolver;
+            }
+            else
+            {
+                return _parentActivity.ContentResolver;
+            }
```
pmaytak, line 341: nit: could be turned into: `return _parentActivity?.ContentResolver ?? Application.Context.ContentResolver`

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -0,0 +1,393 @@
...
+
+            return BrokerConstants.MsAuthenticatorContentProviderUri + "/" + operation;
+        }
+
+        private string SearializeBundleToJSON(Bundle bundle)
+        {
+            return Base64UrlHelpers.Encode(marshall(bundle));
+        }
+
+        private static byte[] marshall(Bundle parcelable)
```
pmaytak, line 360: Capitalize `Marshall`

src/client/Microsoft.Identity.Client/Platforms/Android/AuthenticationContinuationHelper.cs
```diff
@@ -51,7 +51,7 @@ public static void SetAuthenticationContinuationEventArgs(int requestCode, Resul
                 //The BrokerRequestId is an ID that is attached to the activity launch during brokered authentication
                 // that indicates that the response returned to this class is for the broker.
                 LastRequestLogger?.Info("Processing result from broker.");
-                AndroidBroker.SetBrokerResult(data, (int)resultCode, LastRequestLogger);
+                AndroidBrokerStaticHelper.SetBrokerResult(data, (int)resultCode, LastRequestLogger);
```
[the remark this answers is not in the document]
bgavrilMS: I think that Result enum is misleading because of Xamarin translations. You'd think it can only have values -1, 0 and 1 - 
image

But this is not the case, Authenticator can set it to other values. Custom activities (which Android broker is) can set it to different results. 

In fact, the Java API uses an int:

https://developer.android.com/reference/android/app/Activity#onActivityResult(int,%20int,%20android.content.Intent)

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -0,0 +1,452 @@
...
+        {
+            using (_logger.LogMethodDuration())
+            {
+                // onActivityResult will receive the response for this activity.
+                // Launching this activity will switch to the broker app.
+
+                _logger.Verbose("Starting Android Broker interactive authentication. ");
+                Intent brokerIntent = await GetIntentForInteractiveBrokerRequestAsync(brokerRequest, _parentActivity).ConfigureAwait(false);
+
+                if (brokerIntent != null)
```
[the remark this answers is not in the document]
bgavrilMS: As far as I can tell, GetIntentForInteractiveBrokerRequestAsync will either return an intent object or it will throw. So a null check is not needed here?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerFactory.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            // Return a default broker in case no broker is installed to handle install URL
+            return new AndroidContentProviderBroker(uIParent, logger);
+        }
+
+        public static IBroker CreateBroker(CoreUIParent uIParent, ICoreLogger logger)
+        {
+            if (s_installedBroker == BrokerType.None)
+            {
+                return Task.Run(async () => await GetInstalledBrokerAsync(uIParent, logger).ConfigureAwait(false)).Result;
```
bgavrilMS, line 73: Use `GetAwaiter().GetResult()` insteaed of `.Result`

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerFactory.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Platforms.Android.Broker
+{
+    internal class AndroidBrokerFactory
+    {
+        private static BrokerType s_installedBroker;
```
bgavrilMS, line 23: Are you sure this is ok? This implies that if user upgrades Authenticator but does not restart the app, it will use AccountManager mechanism. Not saying it's wrong, just wanted to double check that this is an acceptable experience.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerFactory.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private enum BrokerType 
+        {
+            None,
```
bgavrilMS, line 27: Not really "None", but "Unknown"

bgavrilMS (approved):
Generally OK, please take your time to resolve the other comments.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerFactory.cs
neha-bhargava: None since this will be the case when no broker is installed. But then initially it will be Unknown.

bgavrilMS: `NoneOrUnknown` ?

bgavrilMS: Or code both for clarity.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
```diff
@@ -208,31 +120,30 @@ public void CheckForBrokerAccountInfoInAccountManager(BrokerRequest brokerReques
             string homeAccountId = brokerRequest.HomeAccountId;
             string localAccountId = brokerRequest.LocalAccountId;
 
-            if (!string.IsNullOrEmpty(accounts))
+            if (!string.IsNullOrEmpty(accountData))
             {
-                dynamic authResult = JArray.Parse(accounts);
+                dynamic authResult = JArray.Parse(accountData);
```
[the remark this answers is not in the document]
trwalke: it is used because this is the only object that seems to be able to get the data from the parsed account data. I tried using standard .NET apis for this but they do not always work on Android. Sometimes they throw an error or return null

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerConstants.cs
trwalke: I added it above the enum because the ContentResolverOperation is only used by android

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerHelper.cs
trwalke: it may be possible that the user passes in a request with the localacocunt  id and the homeacount id. in that case we are just checking to see if the account data matches between the request data and the data in the account manager before executing the silent call

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidBrokerFactory.cs
trwalke: discussed offline.
This allows us to avoid having an extra handshake for each request and can be reset by restarting the app

jennyf19 (approved):
looks good. tested the v2 sample w/and w/out broker. no issue. nice work @trwalke

Issue #2356, opened 2021-01-21
[Feature Request] Add to telemetry the field for cache_refresh status


neha-bhargava:
As a resiliency request, we need to add a value to telemetry to indicate what caused cache refresh. For this we need to add a field cache_refresh to telemetry which can have following values:
0 - if token request goes to ESTS because no cached AT exists
1 - if token request goes to ESTS becasue cached AT expired 
2 - if token request goes to ESTS because refresh_in was used and the existing token needs to be refreshed
3 - if token request goes to ESTS because force_refresh was set to true
*blank* - if no cache is used

This change will have the schema version for telemetry changed to 3. The current telemetry header will look like 
``` schema_version | api_id, force_refresh, cache_refresh | platform_fields ``` after the modification.

Reference: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/2506?_a=files&anchor=current-request

trwalke:
fixed in [4.26.0 Release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.26.0)

PR #2383, opened 2021-02-02
Add telemetry for cache refresh

neha-bhargava:
#2356 

As per spec:

0 if token request goes to ESTS because no cached AT exists, 
1 if token request goes to ESTS becasue cached AT expired, 
2 if token request goes to ESTS because refresh_in was used and the existing token needs to be refreshed, 
3 if token request goes to ESTS because force_refresh was set to true, blank if the cache is not hit to make the request

@neha-bhargava  - if you're using a broker, please confirm that the value will be ""

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -38,14 +38,7 @@ namespace Microsoft.Identity.Client.Cache
         public async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync()
         {
             await RefreshCacheForReadOperationsAsync(CacheEvent.TokenTypes.AT).ConfigureAwait(false);
-            MsalAccessTokenCacheItem msalAccessToken = await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
-
-            if (msalAccessToken == null && _requestParams.RequestContext.ApiEvent.CacheRefresh == null)
-            {
-                _requestParams.RequestContext.ApiEvent.CacheRefresh = "0";
-            }
-
-            return msalAccessToken;
+            return await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
         }
 
         public async Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem>> SaveTokenResponseAsync(MsalTokenResponse tokenResponse)
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -38,7 +38,14 @@ namespace Microsoft.Identity.Client.Cache
         public async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync()
         {
             await RefreshCacheForReadOperationsAsync(CacheEvent.TokenTypes.AT).ConfigureAwait(false);
-            return await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
+            MsalAccessTokenCacheItem msalAccessToken = await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
+
+            if (msalAccessToken == null && _requestParams.RequestContext.ApiEvent.CacheRefresh == null)
+            {
+                _requestParams.RequestContext.ApiEvent.CacheRefresh = "0";
+            }
+
+            return msalAccessToken;
         }
 
         public async Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem>> SaveTokenResponseAsync(MsalTokenResponse tokenResponse)
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
@@ -71,7 +71,10 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 cacheRefresh = CacheRefresh.ForceRefresh;
             }
 
-            AuthenticationRequestParameters.RequestContext.ApiEvent.CacheRefresh = (int) cacheRefresh;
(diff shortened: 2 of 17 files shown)


src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
```diff
@@ -38,7 +38,14 @@ internal class CacheSessionManager : ICacheSessionManager
         public async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync()
         {
             await RefreshCacheForReadOperationsAsync(CacheEvent.TokenTypes.AT).ConfigureAwait(false);
-            return await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
+            MsalAccessTokenCacheItem msalAccessToken = await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
+
+            if (msalAccessToken == null && _requestParams.RequestContext.ApiEvent.CacheRefresh == null)
+            {
+                _requestParams.RequestContext.ApiEvent.CacheRefresh = "0";
```
bgavrilMS, line 45: Magic numbers are not readable. Please use an enum and transform it later.

src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
```diff
@@ -38,7 +38,14 @@ internal class CacheSessionManager : ICacheSessionManager
         public async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync()
         {
             await RefreshCacheForReadOperationsAsync(CacheEvent.TokenTypes.AT).ConfigureAwait(false);
-            return await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
+            MsalAccessTokenCacheItem msalAccessToken = await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
+
+            if (msalAccessToken == null && _requestParams.RequestContext.ApiEvent.CacheRefresh == null)
```
bgavrilMS, line 43: I don't think this logic belongs here. CacheManager does not know about this. Logic should go to request classes.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -60,6 +60,10 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                         TokenSource.Cache);
                 }
             }
+            else
```
bgavrilMS, line 63: Please raise an issue that "refresh_in" logic is not performed for OBO.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
```diff
@@ -54,10 +54,20 @@ public async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellat
                     AuthenticationRequestParameters.RequestContext.ApiEvent.IsAccessTokenCacheHit = true;
```
bgavrilMS, line 54: Is `IsAccessTokenCacheHit` still needed? Can you consolidate the telemetry, i.e. still send that flag, but based on your more complex API event?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -462,6 +462,8 @@ private MsalAccessTokenCacheItem FilterByExpiry(MsalAccessTokenCacheItem msalAcc
                         "Access token has expired or about to expire. " +
                         GetAccessTokenExpireLogMessageContent(msalAccessTokenCacheItem));
                 }
+
+                requestParams.RequestContext.ApiEvent.CacheRefresh = "1";
```
bgavrilMS, line 466: I think the logic for values "0" and "1" should be done here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ccbf33200ba4ddb606f5a3e13b05c9c8e43e8335/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L328

bgavrilMS: Smth like

```csharp

  // no match
            if (finalList.Count == 0)
            {
                logger.Verbose("No tokens found for matching authority, client_id, user and scopes.");
                API_EVENT = "0"
                return null;
            }
  
            msalAccessTokenCacheItem = FilterByExpiry(msalAccessTokenCacheItem, requestParams);
            if (at==null) { API_EVENT = "1" }
```

bgavrilMS: Telemetry (and logging) are cross cutting concerns, meaning that it's ok for all parts of the product to contribute to telemetry. This violates Single Responsability Principle, but it's acceptable. Still,  let's try to concentrate as much of this logic to one place rather than spreading it around.

bgavrilMS (changes requested):
Minor changes needed. Main concern is about using "magic numbers" instead of constants or a an enum, to improve code readability.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
neha-bhargava: Created issue: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2389

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
neha-bhargava: This is something that we still need and maybe we can consolidate it with CacheRefresh but those are 2 different things. There won't be a big change in the code here, we will reduce the use of this variable but it will increase the logic that will set this value as true and will complicate in case we have some extra values to add to CacheRefresh enum. IMO we should leave it like this but maybe I am thinking in a different direction than you. We can discuss this offline.

trwalke (approved):
I agree with the comments Bogdan posted. 
LGTM

bgavrilMS (approved):

Issue #2381, opened 2021-02-02
[Feature Request] Specify tenant id for application token cache key


yohanb:
Looking at the current implementation of the token cache, I noticed when caching app tokens, the key is _clientId_AppTokenCache_.  In a multi-tenant application, this cache key will continuously get overridden for each tenant the token is requested.  Why not specify the tenant ID in the token cache key like so _clientId_tenantId_AppTokenCache_?
Thanks.

bgavrilMS:
On it

yohanb:
@bgavrilMS wasn't expecting that! 😂
let me know how I can help.

yohanb:
took the initiative to create a PR for the requested enhancement.

trwalke:
fixed in [4.26.0 Release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.26.0)

PR #2384, opened 2021-02-03
Added tenantID to suggested app token cache key

yohanb:
@bgavrilMS here's a proposed fix to adding the tenantID to the suggested cache key on app tokens (see #2381).

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
 using Microsoft.Identity.Client.Internal.Requests;
 
 namespace Microsoft.Identity.Client.Cache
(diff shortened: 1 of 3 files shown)


bgavrilMS:
@yohanb - I want to a make a small change and fix some other tests where we do look at this. So I'm merging to a temporary branch and I'll a commit on top. 

Much appreciated.

src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
```diff
@@ -54,6 +55,13 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
 
             if (requestParameters.ApiId == TelemetryCore.Internal.Events.ApiEvent.ApiIds.AcquireTokenForClient)
             {
+                if (!string.IsNullOrEmpty(requestParameters.Authority.TenantId) &&
+                    Guid.TryParse(requestParameters.Authority.TenantId, out var tenantId))
```
bgavrilMS, line 59: Change is fine with one small issue: tenantId may not be a GUID, for example "bgavril.onmicrosoft.com" is a valid tenant ID. 
We only care if tenatnID is not null or empty.

yohanb: Yeah I wasn't sure about that.
It also means that tenantid can be "common" or "organizations, etc. 
I'm guessing that's ok? It'll result in a cache miss for those who already have those tokens cached.

PR #2385, opened 2021-02-03
Add logging in HttpManager plus minor logging updates in RegionDiscoveryProvider.

pmaytak:
Add logging in HttpManager plus minor logging updates in RegionDiscoveryProvider.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -212,11 +212,13 @@ namespace Microsoft.Identity.Client.Http
                 requestMessage.Method = method;
                 requestMessage.Content = body;
 
-                logger.Verbose($"Sending request from HttpManager. Method: {method}. URI path: {endpoint.Scheme}://{endpoint.Authority}{endpoint.AbsolutePath}. ");
+                logger.Verbose($"[HttpManager] Sending request. Method: {method}. URI path: {(endpoint == null ? "NULL" : $"{endpoint.Scheme}://{endpoint.Authority}{endpoint.AbsolutePath}")}. ");
 
                 using (HttpResponseMessage responseMessage =
                     await client.SendAsync(requestMessage, cancellationToken).ConfigureAwait(false))
                 {
+                    logger.Verbose($"[HttpManager] Received response. Status code: {responseMessage.StatusCode}. ");
+
                     HttpResponse returnValue = await CreateResponseAsync(responseMessage).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -86,7 +86,7 @@ private async Task<string> GetRegionAsync(RequestContext requestContext)
 
             if (!string.IsNullOrEmpty(region))
             {
-                requestContext.Logger.Info($"[Region discovery] Region found in environment variable: {region}.");
+                requestContext.Logger.Info($"[Region Discovery] Region found in environment variable: {region}.");
```
jmprieur, line 89: why a capitalized D? (this is a sentence?). Microsoft is advising to use sentence casing in public facing docs, logs etc ...

jmprieur (commented):
LGTM for the wording.
I think we should use sentence casing (according to [Microsoft voice](https://docs.microsoft.com/en-us/style-guide/top-10-tips-style-voice))

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
pmaytak: Capitalized for consistency. We used `[Region Discovery]` and `[Instance Discovery]` else where. I wasn't aware of that guidance, so I can update to use sentence casing.

bgavrilMS: @jmprieur - these are MSAL logs, not docs.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -211,6 +212,8 @@ private static bool HasRetryAfterHeader(HttpResponse response)
                 requestMessage.Method = method;
                 requestMessage.Content = body;
 
+                logger.Verbose($"Sending request from HttpManager. Method: {method}. URI path: {endpoint.Scheme}://{endpoint.Authority}{endpoint.AbsolutePath}. ");
```
bgavrilMS, line 215: Do you want to log the HTTP status of the response as well?

bgavrilMS (approved):

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
pmaytak: Done. We also have additional logging for response errors in the outer method.

PR #2386, opened 2021-02-03
Add tenantID to suggested app token cache key

bgavrilMS:


src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
```diff
@@ -54,8 +55,8 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
 
             if (requestParameters.ApiId == TelemetryCore.Internal.Events.ApiEvent.ApiIds.AcquireTokenForClient)
             {
-                key = requestParameters.AppConfig.ClientId + "_AppTokenCache";
-                return true;
+                string tenantId = requestParameters.Authority.TenantId ?? "";
+                key = $"{requestParameters.AppConfig.ClientId}_{tenantId}_AppTokenCache";
```
bgavrilMS, line 59: Test scenarios: 
- authority is "l.m.o/common" -> tenant id is common
- authority is tenanted -> tenant id
- authority does not support tenants (ADFS or B2C) -> ""

bgavrilMS: Thanks @yohanb for proposing the fix!

src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
```diff
@@ -54,8 +55,8 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
 
             if (requestParameters.ApiId == TelemetryCore.Internal.Events.ApiEvent.ApiIds.AcquireTokenForClient)
             {
-                key = requestParameters.AppConfig.ClientId + "_AppTokenCache";
-                return true;
+                string tenantId = requestParameters.Authority.TenantId ?? "";
```
yohanb, line 58: you're not using the tenantId var in your returned result

yohanb: also if tenantId is an empty string, key will have double underscores, i.e. clientId__AppTokenCache

bgavrilMS: This is fine, I think it's more readable this way, to show that a tenantId does not exist (i.e. ADFS authority)

yohanb (approved):

yohanb:
awesome, thanks @bgavrilMS !  What is the release cycle like and when can I expect to be able to use this?

bgavrilMS:
We expect to release MSAL 4.26 by mid next-week. Sooner if testing goes well.

PR #2388, opened 2021-02-03
MSAL.Desktop to use .NET FWK 4.6.1

bgavrilMS:


bgavrilMS:
CC @mjcheetham

jmprieur (approved):
Thanks @bgavrilMS 
LGTM
- Is it going to break any customer (i'd think not)? 
- isn't there an impact on usable crypto algorithms?

PR #2393, opened 2021-02-06
Fix instance discovery caching.

pmaytak:
Fixes for two bugs with regional instance discovery.

After the app startup, during the instance discovery flow, there are at least two calls made to the `TryGetMetadataAsync` in `RegionDiscoveryProvider`. First call passes in a known authority (like login.microsoft.com). Second call passes in a regionalized authority (like region.login.microsoft.com).

1
In `BuildAuthorityWithRegionAsync` method there's [code at the end that transforms the passed-in URL by adding a region](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/018c34ef0efbd7902aa71a1370609998911c345d/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs#L263). During the first call, the known authority is correctly transformed into a regionalized authority. However, during the second call, a region is prepended to an already regionalized authority. Result - this adds an extra item into the instance discovery cache; this always ends up calling the `TryGetMetadataAsync` an extra time; and the token cache uses regionalized host instead of a known host in the token cache item key. Fix - check if URL is already regionalized before transforming.
image
image
image

2
[When an `InstanceDiscoveryMetadataEntry` is created](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/018c34ef0efbd7902aa71a1370609998911c345d/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs#L195), the original known authority host entry is not cached. Result - Every time there's a request for a token, a known authority is passed in, checked against the cache, not found, goes through regionalization code, then the regionalized authority is checked against the cache. Fix - add the known authority host as a key to the instance discovery cache with properties pointing to the regionalized host to skip the regionalization each time.
image
image

Todo: The telemetry unit tests need to be updated because previously both calls to `TryGetMetadataAsync` returned regions source as environment or IMDS; with this change, the second call is always returns cache source and overwrites the first telemetry.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
@@ -252,13 +252,18 @@ namespace Microsoft.Identity.Client.Region
                 requestContext.ApiEvent.FallbackToGlobal = false;
             }
 
+            if (canonicalAuthority.Host.StartsWith($"{s_region}."))
+            {
+                return canonicalAuthority;
+            }
+
             var builder = new UriBuilder(canonicalAuthority);
 
             if (KnownMetadataProvider.IsPublicEnvironment(canonicalAuthority.Host))
             {
                 builder.Host = $"{s_region}.login.microsoft.com";
             }
-            else if (!builder.Host.Equals($"{s_region}.login.microsoft.com"))
+            else
             {
                 builder.Host = $"{s_region}.{builder.Host}";
             }
--- a/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
@@ -151,7 +151,7 @@ namespace Microsoft.Identity.Client.Region
         {
             requestContext.ApiEvent.RegionDiscovered = region;
 
-            if (requestContext.ApiEvent.RegionSource == 0)
+            if (requestContext.ApiEvent.RegionSource == (int)RegionSource.None)
             {
                 requestContext.ApiEvent.RegionSource = (int)regionSource;
             }
@@ -199,7 +199,7 @@ namespace Microsoft.Identity.Client.Region
         {
             return new InstanceDiscoveryMetadataEntry()
             {
-                Aliases = new[] { regionalizedAuthority.Host, orginalAuthority.Host },
+                Aliases = new[] { regionalizedAuthority.Host },
                 PreferredCache = orginalAuthority.Host,
                 PreferredNetwork = regionalizedAuthority.Host
             };
--- a/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
@@ -134,7 +134,7 @@ namespace Microsoft.Identity.Client.Region
             }
             catch (Exception e)
             {
-                requestContext.Logger.Info("[Region discovery] Call to local imds failed. " + e);
+                requestContext.Logger.Info("[Region discovery] Call to local IMDS failed. " + e);
                 throw new MsalServiceException(MsalError.RegionDiscoveryFailed, MsalErrorMessage.RegionDiscoveryFailed, e);
             }
         }
@@ -194,7 +194,7 @@ namespace Microsoft.Identity.Client.Region
         {
             return new InstanceDiscoveryMetadataEntry()
             {
-                Aliases = new[] { regionalizedAuthority.Host },
+                Aliases = new[] { regionalizedAuthority.Host, orginalAuthority.Host },
                 PreferredCache = orginalAuthority.Host,
                 PreferredNetwork = regionalizedAuthority.Host
             };
@@ -243,7 +243,7 @@ namespace Microsoft.Identity.Client.Region
 
                     s_region = regionToUse;
                     requestContext.ApiEvent.FallbackToGlobal = false;
-                    requestContext.Logger.Info($"[Region discovery] Region auto detection failed. Region provided by the user will be used: ${regionToUse}.");
+                    requestContext.Logger.Info($"[Region discovery] Region auto detection failed. Region provided by the user will be used: {regionToUse}.");
                     LogTelemetryData(s_region, RegionSource.UserProvided, requestContext);
                 }
             }
@@ -258,7 +258,7 @@ namespace Microsoft.Identity.Client.Region
             {
                 builder.Host = $"{s_region}.login.microsoft.com";
             }
-            else
+            else if (!builder.Host.Equals($"{s_region}.login.microsoft.com"))
             {
                 builder.Host = $"{s_region}.{builder.Host}";
             }
--- a/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
@@ -150,7 +150,12 @@ namespace Microsoft.Identity.Client.Region
         private void LogTelemetryData(string region, RegionSource regionSource, RequestContext requestContext)
         {
             requestContext.ApiEvent.RegionDiscovered = region;
-            requestContext.ApiEvent.RegionSource = (int)regionSource;
+
+            if (requestContext.ApiEvent.RegionSource == 0)
+            {
+                requestContext.ApiEvent.RegionSource = (int)regionSource;
+            }
+
             requestContext.ApiEvent.UserProvidedRegion = requestContext.ServiceBundle.Config.AuthorityInfo.RegionToUse;
         }
 
--- a/tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
@@ -57,6 +57,7 @@ namespace Microsoft.Identity.Test.Unit.CoreTests
             Environment.SetEnvironmentVariable(TestConstants.RegionName, "");
             _testRequestContext.ServiceBundle.Config.AuthorityInfo.RegionToUse = "";
             _harness?.Dispose();
+            _regionDiscoveryProvider.Clear();
             base.TestCleanup();
         }
 
@@ -145,7 +146,7 @@ namespace Microsoft.Identity.Test.Unit.CoreTests
             Environment.SetEnvironmentVariable(TestConstants.RegionName, TestConstants.Region);
             
             // In the instance discovery flow, TryGetMetadataAsync is always called with a known authority first, then with regionalized.
-            var t = await _regionDiscoveryProvider.TryGetMetadataAsync(new Uri("https://login.microsoftonline.com/common/"), _testRequestContext).ConfigureAwait(false);
+            await _regionDiscoveryProvider.TryGetMetadataAsync(new Uri("https://login.microsoftonline.com/common/"), _testRequestContext).ConfigureAwait(false);
             InstanceDiscoveryMetadataEntry regionalMetadata = await _regionDiscoveryProvider.TryGetMetadataAsync(regionalizedAuthority, _testRequestContext).ConfigureAwait(false);
 
             ValidateInstanceMetadata(regionalMetadata);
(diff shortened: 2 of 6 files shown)


src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -258,7 +258,7 @@ private async Task<Uri> BuildAuthorityWithRegionAsync(Uri canonicalAuthority, Re
             {
                 builder.Host = $"{s_region}.login.microsoft.com";
             }
-            else
+            else if (!builder.Host.Equals($"{s_region}.login.microsoft.com"))
```
jennyf19, line 261: i would do something like this:
```csharp
 else if (!builder.Host.StartsWith($"{s_region}."))
 {
                builder.Host = $"{s_region}.{builder.Host}";
 }
```

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -137,6 +137,20 @@ public async Task ResponseFromUserProvidedRegionDifferentFromRegionDetectedAsync
...
+
+            Environment.SetEnvironmentVariable(TestConstants.RegionName, TestConstants.Region);
+            
+            // In the instance discovery flow, TryGetMetadataAsync is always called with a known authority first, then with regionalized.
+            var t = await _regionDiscoveryProvider.TryGetMetadataAsync(new Uri("https://login.microsoftonline.com/common/"), _testRequestContext).ConfigureAwait(false);
+            InstanceDiscoveryMetadataEntry regionalMetadata = await _regionDiscoveryProvider.TryGetMetadataAsync(regionalizedAuthority, _testRequestContext).ConfigureAwait(false);
+
+            ValidateInstanceMetadata(regionalMetadata);
+        }
+
```
jennyf19, line 153: what's the e2e scenario to reproduce this, from a customer perspective? this is a good test, but how does a customer get into this situation? would be interesting to have a test that covers that scenario, without forcing the 2 calls to the instance discovery endpoint separately.

also, clear the discoveryProviders at the end. remove unused variable `t` or assert something with it.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -194,7 +194,7 @@ private static InstanceDiscoveryMetadataEntry CreateEntry(Uri orginalAuthority,
         {
             return new InstanceDiscoveryMetadataEntry()
             {
-                Aliases = new[] { regionalizedAuthority.Host },
+                Aliases = new[] { regionalizedAuthority.Host, orginalAuthority.Host },
                 PreferredCache = orginalAuthority.Host,
```
jennyf19, line 198: i don't think this is needed once the actual issue on line 261 is fixed.

jmprieur: Do we think that this is the right fix? If we add the `originalAuthorityHost` to this new `InstanceDiscoveryMetadataEntry` we'll have several  `InstanceDiscoveryMetadataEntry` which `Aliases` property will contain the same value, therefore the  `InstanceDiscoveryMetadataEntry` won't be a partition any longer?
Wouldn't this bring some random behavior?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
jmprieur: @jennyf19's proposal is clearer to me, and this would also work for national clouds: @neha-bhargava said in a previous PR that they were not supported yet, but if they were we'd just prepend the region to the regular hostname, which would work this the suggestion.

jmprieur (commented):
I've asked a couple of questions.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
bgavrilMS: @neha-bhargava to look at this.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
bgavrilMS: National clouds are not supported yet, but the spec agreed with estsr mentions this corner case.

bgavrilMS:
These are very good bugs @pmaytak, excellent testing!

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -243,7 +243,7 @@ private async Task<Uri> BuildAuthorityWithRegionAsync(Uri canonicalAuthority, Re
 
                     s_region = regionToUse;
                     requestContext.ApiEvent.FallbackToGlobal = false;
-                    requestContext.Logger.Info($"[Region discovery] Region auto detection failed. Region provided by the user will be used: ${regionToUse}.");
+                    requestContext.Logger.Info($"[Region discovery] Region auto detection failed. Region provided by the user will be used: {regionToUse}.");
```
bgavrilMS, line 246: There's potentially a null ref issue at line 215 - 

```if (string.IsNullOrEmpty(str))``` is the correct null check.

File: src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs (hunk 1 above)
Comment (bgavrilMS) [reply src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs:261]: Anyway, both suggestions seem fine.

File: tests/devapps/RegionalTestApp/Program.cs [hunk 5]
```diff
@@ -45,7 +45,7 @@ static async Task Main(string[] args)
 
             Console.WriteLine("=== Acquire token global ===");
             await AcquireTokenAsync(certificate, false).ConfigureAwait(false);
-
+            Console.ReadLine();
```
trwalke, line 48: What is the reason for this?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: I had this earlier, the alias with the original authority, removed it because then it will cache the instance discovery with the original alias and if regional is set to false for some percentage, then still it will go to the regional endpoint. So, I don't think this is needed.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: I agree with what @jmprieur said. For PPE env we just prepend the region to the original authority. Makes sense to implement what @jennyf19 has mentioned.

trwalke:
FYI: failed tests here 
image

tests/devapps/RegionalTestApp/Program.cs
pmaytak: This was to prevent the console window from closing when running this dev app executable (like on a test VM).

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
pmaytak: Updated this check to use `StartsWith` and moved it a bit higher, per Neha's suggestion.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
pmaytak: The scenario is when the app is run the first time and regionalization is turned on and the instance cache is empty. In end-to-end flow, `TryGetMetadataAsync` ends up being called twice - once for known authority and once for regionalized. Although, this might make more sense to convert this into an integration test.

In MS Test, the constructor and instance variables are run before each test method. I verified that instance cache is clear before each test. I did add code to clear the cache in test cleanup anyway, though.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
pmaytak: Reverted this after discussion. Not the right fix.

jennyf19: what's the right fix then?

pmaytak:
@neha-bhargava @jmprieur @jennyf19 @trwalke @bgavrilMS 

We kept the fix for double appending the region. Reverted the changes related to original host caching. Tests pass locally.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
pmaytak: Not sure, more investigation is needed. Will create a separate issue for this.

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
```diff
@@ -150,7 +150,12 @@ private CancellationToken GetCancellationToken(CancellationToken userCancellatio
         private void LogTelemetryData(string region, RegionSource regionSource, RequestContext requestContext)
         {
             requestContext.ApiEvent.RegionDiscovered = region;
-            requestContext.ApiEvent.RegionSource = (int)regionSource;
+
+            if (requestContext.ApiEvent.RegionSource == (int)RegionSource.None)
```
henrik-me, line 154: >if (requestContext.ApiEvent.RegionSource == (int)RegionSource.None) [](start = 12, length = 67)

why?

henrik-me:
I'm not seeing a test that exercises what was fixed? Am I missing something?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
pmaytak: @neha-bhargava can explain better then me.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -137,6 +138,20 @@ public async Task ResponseFromUserProvidedRegionDifferentFromRegionDetectedAsync
             regionDiscoveryProvider.Clear();
         }
 
+        [TestMethod]
+        public async Task SuccessfulResponseFromRegionalizedAuthorityAsync()
```
pmaytak, line 142: @henrik-me This is the test method that tests the scenario. Calling `TryGetMetadataAsync` twice because it gets called twice in the end-to-end flow.

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
henrik-me: I thought the idea was that someone calls aquiretokenforclient 2 times and gets into this situation? Or are you saying that our code does that before returning the token? .. in which case why is that?

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
henrik-me: have a link to the issue?

---
In reply to: [572444297](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2393#discussion_r572444297) [](ancestors = 572444297)

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: Since we call this method multiple times in the flow, this value can get overwritten when this method is called the second time and indicate that the value was cached. So, to maintain the original value we have to check if we are not overwriting the value we had.

henrik-me: what is the problem of overwriting it? Why would the first value be better than the second value?

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
pmaytak: The latter - our code does it. Not sure why, investigation is needed.

src/client/Microsoft.Identity.Client/Region/RegionDiscoveryProvider.cs
neha-bhargava: For example, for a single request the flow goes like following:
1. Region is not discovered yet and env variable is the source of region discovery, region_source will set to 1.
2. The method is called again and in the previous iteration the value gets cached, region_source will set to 3.
This will cause the region_source value to always be 3 which means the region was read from the cache. I am not sure why it was working fine before this fix and the tests and telemetry was showing the correct results.

Issue #2411, opened 2021-02-12
Refreshing token using client application ID URI


usdruscia:
**Which Version of MSAL are you using ?**
MSAL.NET 4.25.0

**Platform**
.NET Framework 4.5

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App 
    * [ ] Authorization code
    * [ ] OBO
* Daemon App 
    * [ ] Service to Service calls

**Is this a new or existing app?**
This is a new app or experiment

**Repro**

```csharp
// PublicClientApplication instantiation
var builder = PublicClientApplicationBuilder
                .Create("https://my.app") // I would like to use the URI instead of the ID
                .WithAuthority(AzureCloudInstance.AzurePublic, tenantId)
                .WithDefaultRedirectUri();

m_PublicClientApplication = builder.Build();

// Token acquisition
var accounts = await m_PublicClientApplication.GetAccountsAsync();
var result = await m_PublicClientApplication.AcquireTokenSilent("https://my.api/user_impersonation", accounts.FirstOrDefault()).ExecuteAsync();
```

**Expected behavior**
Get a refreshed access_token when it expires (which works if I use the `app ID` instead of the `app ID URI`)

**Actual behavior**
MsalUiRequiredException - AADSTS70000: Provided grant is invalid or malformed

bgavrilMS:
Apologies, I accidentally disabled the nice exception that you normally get when you try to use a client_id that is not a GUID. It says: 

```xml
 <para>What happens?</para>You've specified a client ID that is not a <see cref="Guid"/>
 <para>Mitigation</para>Use the application ID (a GUID) from the application portal as client ID in this SDK
```

Will get this validation back in with the next MSAL release. 
You can use the non-guid form of the tenant ID by the way, e.g. `contoso.onmicrosoft.com`

usdruscia:
Will the URI form will work at some point for the client ID?

bgavrilMS:
Not that I'm aware of. May I ask what you are trying to do that you need the URI instead of the GUID?

jmprieur:
@usdruscia. This is not going to happen and here is why:
[...]

PR #2414, opened 2021-02-16
Fix for #2411 - validation of client ID

bgavrilMS:


tests/Microsoft.Identity.Test.Unit/AppConfigTests/PublicClientApplicationBuilderTests.cs
```diff
@@ -49,6 +49,30 @@ public void TestWithDifferentClientId()
...
+                      .WithAuthority(TestConstants.B2CAuthority)
+                      .Build());
+
+            Assert.AreEqual(MsalError.ClientIdMustBeAGuid, ex.ErrorCode);
+
+            // ADFS does not have this constraint
+            PublicClientApplicationBuilder.Create("http://my.app")
+                        .WithAuthority(new Uri(TestConstants.ADFSAuthority))
+                        .Build();
+
```
jennyf19, line 73: assert something here? maybe that the client id is the same as one passed in?

jennyf19 (approved):

trwalke (approved):

PR #2415, opened 2021-02-16
Fix #2412 - do not assume all authorities are AAD

### LINKED ISSUE #2412: [Bug] WithAuthority(string) and WithAuthority(Uri) have different behaviours
MSAL 4.26

`WithAuthority(Uri)` is parsing the authority and flow uses AAD, B2C or ADFS authority
`WithAuthority(string)` assumes this is an AAD authority

bgavrilMS:


jennyf19 (approved):

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -582,7 +582,33 @@ public T WithAuthority(Uri authorityUri, bool validateAuthority = true)
...
+        /// <param name="validateAuthority">Whether the authority should be validated against the server metadata.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public T WithAuthority(string authorityUri, bool validateAuthority = true)
+        {
+            if (string.IsNullOrWhiteSpace(authorityUri))
+            {
+                throw new ArgumentNullException(authorityUri);
+            }
+
+            Config.AuthorityInfo = AuthorityInfo.FromAuthorityUri(authorityUri, validateAuthority);
```
trwalke, line 611: Will this result in a change in behavior for those using this overload if they are using an adfs or B2C authority?

bgavrilMS: Before the fix we were using 

` AuthorityInfo.FromAadAuthority(authorityUri, validateAuthority)`

which, as the name suggests, will create an AAD authority. So B2C and ADFS would encounter issues (Null Reference probably because we'd try to extract the tenant).

The fix ensures we parse the string and resolve it to either AAD, b2C or ADFS

Issue #2419, opened 2021-02-18
Add private security nupkg analyzer to IDE


bgavrilMS:
- [x] Discuss solutions with 1es team
- [x] Investigate running analyzers from IDE
- [x] Convert ruleset to editorconfig format

PR #2429, opened 2021-02-23
Add Internal SDL rulesets to VS build

neha-bhargava:
#2419 
Add private security nupkg analyzer to build and IDE

- [x]  Discuss solutions with 1es team
- [ ]  Consume package from build and fail build if rule is broken
- [ ]  Ensure latest version of the package is always consumed, not a fixed version
- [x]  Investigate running analyzers from IDE
- [ ]  Investigate a way to not break non-MSFT contributors from building MSAL.


Directory.Build.props
```diff
@@ -1,24 +1,31 @@
...
-    <TreatWarningsAsErrors>False</TreatWarningsAsErrors>
 
     <LangVersion>7.2</LangVersion>
     <!-- This is strong naming, not signing-->
     <SignAssembly>true</SignAssembly>
     <!-- The MSAL.snk has both private and public keys -->
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)/build/MSAL.snk</AssemblyOriginatorKeyFile>
     <DelaySign>false</DelaySign>
-    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)/build/SolutionWideAnalyzerConfig.ruleset</CodeAnalysisRuleSet>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
bgavrilMS, line 10: This is pretty heavy. We have a bunch of Xamarin warnings that are hard to ignore. Do you manage to get a clean build?

src/client/Microsoft.Identity.Client.Desktop/Microsoft.Identity.Client.Desktop.csproj
```diff
@@ -56,6 +56,20 @@
      <Reference Include="System.Windows.Forms" />
 
    </ItemGroup>
+   <ItemGroup>
+     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.2">
+       <PrivateAssets>all</PrivateAssets>
+       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+     </PackageReference>
+     <PackageReference Include="Microsoft.Internal.Analyzers" Version="2.6.9">
```
bgavrilMS, line 64: Try to use floating versions like 

image

I'd actually always download the latest version if possible.

Directory.Build.props
neha-bhargava: No. There are few errors here that are not generated by analyzers:
image

Directory.Build.props
```diff
@@ -1,24 +1,31 @@
...
+    <AdditionalAnalyzerRuleSet>$(MSBuildThisFileDirectory)/build/SolutionWideAnalyzerConfig.ruleset</AdditionalAnalyzerRuleSet>
   </PropertyGroup>
   <ItemGroup>
     <!-- PrivateAssets="All" is on these packages to ensure they're development dependencies
          and aren't included as transitionary dependencies to package consumers.
          For clarity, without PrivateAssets marked here, anyone consuming Microsoft.Identity.Client
          would also be forced to install these dependencies.  PrivateAssets avoids this problem. -->
     <PackageReference Include="AsyncUsageAnalyzers" Version="1.0.0-alpha003" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" PrivateAssets="All" />    
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" PrivateAssets="All" />
```
bgavrilMS, line 26: We generally add analyzers for all projects in the solution, i.e. tests and samples too. Good code practices need to be followed everywhere. 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/84329e60dd751ffa0de9342bbc152e09a354017b/Directory.Build.props#L26

neha-bhargava: Yeah those got included with VS. I will update in all the places

Directory.Build.props
bgavrilMS: You can do:   `<WarningsAsErrors>CS4014</WarningsAsErrors>` instead

tests/Microsoft.Identity.Test.Core.UIAutomation/Microsoft.Identity.Test.UIAutomation.Infrastructure.csproj
```diff
@@ -30,6 +30,21 @@
     <Reference Include="System" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers">
```
bgavrilMS, line 33: these should be removed as well.

Directory.Build.props
```diff
@@ -1,24 +1,31 @@
...
   <ItemGroup>
     <!-- PrivateAssets="All" is on these packages to ensure they're development dependencies
          and aren't included as transitionary dependencies to package consumers.
          For clarity, without PrivateAssets marked here, anyone consuming Microsoft.Identity.Client
          would also be forced to install these dependencies.  PrivateAssets avoids this problem. -->
     <PackageReference Include="AsyncUsageAnalyzers" Version="1.0.0-alpha003" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" PrivateAssets="All" />    
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.*" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.*" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Internal.Analyzers" Version="2.6.*" PrivateAssets="All" />
```
bgavrilMS, line 28: Add a link to 1es wiki about this

Directory.Build.props
neha-bhargava: Ideally we should not have to do this as we are getting the SDL recommended rules as errors without setting this as true. But the security bugs that we have been getting are the ones appearing in the warnings. So either we keep on adding the warnings that we are getting as security bugs as and when they come or we can have this flag as true.
We can also ask the TSA team generating the security bugs, why we are getting non recommended SDL rules as bugs as they are supposed to be warnings according to the analyzers too.

NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear/>
+    <add key="Internal Analyzers" value="https://securitytools.pkgs.visualstudio.com/_packaging/SecureDevelopmentTools@Local/nuget/v3/index.json" />
+  </packageSources>
```
trwalke, line 6: From the error I see in the build here:
"Unable to find package MSBuild.Sdk.Extras. No packages exist with this id in source(s): Internal Analyzers"
it looks like the nuget task is only going to Internal Analyzers to source the nuget packages. I think you may want to add an entry to the default nuget feed because that is where the MSBuild.Sdk.Extras lives.

"https://api.nuget.org/v3/index.json" is the source

neha-bhargava: Yes, I added that before but as Jean-Marc mentioned in the standup, we cannot include internal and external feeds in the nuget.config. We need to create an upstream for nuget.org and add it to the artifacts.

neha-bhargava: Refer to: https://docs.opensource.microsoft.com/tools/cg/security-alerts/nuget-multifeed-configuration.html#internal-projects-eg-non-oss

neha-bhargava: Basically it suggests using nuget.org as upstream: https://docs.microsoft.com/en-us/azure/devops/artifacts/nuget/upstream-sources?view=azure-devops

NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="nuget.orgFeed" value="https://identitydivision.pkgs.visualstudio.com/IDDP/_packaging/nuget.orgFeed/nuget/v3/index.json" />
```
neha-bhargava, line 5: Created a feed with nuget.org and upstream feed for internal analyzers: https://identitydivision.visualstudio.com/IDDP/_packaging?_a=feed&feed=nuget.orgFeed. This is one way I found to create a feed to include in nuget.config where we can have public and private sources.

trwalke:
Did some investigation on this and discovered an issue with how the upstream sourced packages work. There are a number of errors on the build in this PR that are a result of missing packages:
"Unable to find package MSBuild.Sdk.Extras. No packages exist with this id in source(s): Internal Analyzers"

This is because of the usage of the upstream source which will not allow packages to be consumed unless someone manually installs the package in the package manager console. Thus, for each of these errors and all nuget packages that we would want to use going forward, we will need to manually go to the nuget website, check for the latest version and use the nuget install command to add it to the upstream feed. just so that we can add an internal feed. Also, we cannot add the separate feed manually because multifeed nuget configs are banned now.
https://docs.microsoft.com/en-us/azure/devops/artifacts/nuget/consume?view=azure-devops#consume-packages

I am testing out adding the ruleset file manually for this now. it will reduce complexity significantly and we can subscribe to the RoslynSecAnnounce DL to be notified of new SDL ruleset updates

jmprieur:
Do we still need MsBuild.SdkExtras now that the project system fully supports `<TargetFrameworks>` ?

bgavrilMS:
Yes, MSBUild.SDK.Extras are needed for the Xamarin stuff and I think for UWP stuff as well.

trwalke:
Closing this as the work has been done here
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2453

Issue #2389, opened 2021-02-03
[L] refresh_in logic is not performed for OBO


neha-bhargava:
Refresh_in logic needs to be implemented for OBO flow: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d8b8d3747ab7defd76e5b6b553c70e199a8a877f/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs#L49

Refer to: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d8b8d3747ab7defd76e5b6b553c70e199a8a877f/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs#L51

bgavrilMS:
If you have time, have a look at this one @trwalke. It should be fairly simple, but take your time to understand how refresh_in works.

PR #2439, opened 2021-03-01
Adding refresh_in logic for OBO flow

trwalke:
Refresh in logic for OBO
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2389

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                         AuthenticationRequestParameters.RequestContext.CorrelationId,
                         TokenSource.Cache);
                 }
-                else if(msalAccessTokenItem != null)
+                else if(msalAccessTokenItem == null)
                 {
                     cacheRefresh = CacheRefresh.NoCachedAT;
                 }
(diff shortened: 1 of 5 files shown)


trwalke:
Adding test soon

pmaytak (commented):
LGTM.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -61,13 +63,18 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                         AuthenticationRequestParameters.RequestContext.CorrelationId,
                         TokenSource.Cache);
                 }
-                else
+                else if(msalAccessTokenItem == null)
```
bgavrilMS, line 66: Don't need the `else` clause because the previous clause returns, i.e. this can be just 

`if (msalAccessTokenItem != null)` ... 

Or simpler: `cacheRefresh = (at == null) ? CacheRefresh.NoCacheAt  : CacheRefresh.RefreshIn`

tests/Microsoft.Identity.Test.Common/Core/Mocks/TokenCacheHelper.cs
```diff
@@ -51,7 +52,8 @@ internal static MsalAccessTokenCacheItem CreateAccessTokenItem(string scopes = "
             string environment = TestConstants.ProductionPrefCacheEnvironment,
             string displayableId = TestConstants.DisplayableId,
             string rtSecret = TestConstants.RTSecret,
-            string overridenScopes = null, 
+            string overridenScopes = null,
+            string userAssertion = TestConstants.UserAssertion,
```
bgavrilMS, line 56: Since this is general purpose method, let's not add user assertion to every AT

bgavrilMS (approved):

PR #2448, opened 2021-03-03
Bogavril/wv2 options

bgavrilMS:
Expose the ability for developers to set a fixed Title for their WebView2 panel and to use a fixed version WV2, as the installer for the global one needs admin permissions.

Missing unit tests.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
@@ -115,6 +115,25 @@ namespace Microsoft.Identity.Client
             return this;
         }
 
+        // Remark: Default browser WebUI is not available on mobile (Android, iOS, UWP), but allow it at runtime
+        // to avoid MissingMethodException
+        /// <summary>
+        /// Specifies options for using the system OS browser handle interactive authentication.
+        /// </summary>
+        /// <param name="options">Data object with options</param>
+        /// <returns>The builder to chain the .With methods</returns>
+#if !SUPPORTS_WEBVIEW2
+        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)] // hide everywhere but NetStandard
+#endif
+        public AcquireTokenInteractiveParameterBuilder WithWebView2Options(WebView2Options options)
+        {
+            WebView2Options.ValidatePlatformAvailability();
+
+            CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithSystemBrowserOptions);
+            Parameters.UiParent.WebView2Options = options;
+            return this;
+        }
+
         /// <summary>
         /// Sets the <paramref name="loginHint"/>, in order to avoid select account
         /// dialogs in the case the user is signed-in with several identities. This method is mutually exclusive
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Identity.Client
         #if !SUPPORTS_WEBVIEW2 // currently only WebView2 allows customization
         [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
         #endif
-        public AcquireTokenInteractiveParameterBuilder WithEmbeddedWebviewOptions(
+        public AcquireTokenInteractiveParameterBuilder WithEmbeddedWebViewOptions(
             EmbeddedWebViewOptions options)
         {
             EmbeddedWebViewOptions.ValidatePlatformAvailability();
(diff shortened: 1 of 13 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
```diff
@@ -115,6 +115,25 @@ public AcquireTokenInteractiveParameterBuilder WithSystemWebViewOptions(SystemWe
...
+        // to avoid MissingMethodException
+        /// <summary>
+        /// Specifies options for using the system OS browser handle interactive authentication.
+        /// </summary>
+        /// <param name="options">Data object with options</param>
+        /// <returns>The builder to chain the .With methods</returns>
+#if !SUPPORTS_WEBVIEW2
+        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)] // hide everywhere but NetStandard
+#endif
+        public AcquireTokenInteractiveParameterBuilder WithWebView2Options(WebView2Options options)
```
bgavrilMS, line 128: @jmprieur for new API surface and naming.

src/client/Microsoft.Identity.Client/ApiConfig/WebView2Options.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        public WebView2Options()
+        {
+            ValidatePlatformAvailability();
+        }
+
+        /// <summary>
+        /// Forces a static title to be set on the window hosting the browser. If not configured, the widow's title is set to the web page title.
```
jmprieur, line 32: What is it for? is it for branding the application?
What is the title of the web page? does it vary as a function of the action dialog happening with Azure AD? 

In general for branding the best is to do it at the application registration level: https://docs.microsoft.com/azure/active-directory/develop/howto-add-branding-in-azure-ad-apps

src/client/Microsoft.Identity.Client/ApiConfig/WebView2Options.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        public string Title { get; set; }
+
+        /// <summary>
+        /// Allows use of a static WebView2 runtime. If unset, the WebView2 API tries to locate the globally available version.
```
jmprieur, line 37: I don't understand why and what this is?
Do we want to write:

It is possible for applications to bundle a fixed version of the runtime, and ship it side-by-side.
For this you need to tell MSAL (so it can tell WebView2) where to find the runtime bits by setting this property. If you don't set it, MSAL will attempt to use a system-wide "evergreen" installation of the runtime."

jmprieur (commented):
I proposed a few changes

src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
```diff
@@ -32,6 +33,13 @@ internal AadAuthority(AuthorityInfo authorityInfo) : base(authorityInfo)
 
         internal override string TenantId { get; }
 
+        internal bool IsWorkAndSchoolOnly()
```
jennyf19, line 36: >IsWorkAndSchoolOnly [](start = 22, length = 19)

is this used anywhere? do we have a test for it?

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/ApiConfig/WebView2Options.cs
bgavrilMS: It is less for branding of the application and more about having less confusion around where this window comes from. End-users report being confused about it. It has been asked several times from the mobile front as well.

Especially useful for GCM which shares the client_id of VS (for many reasons) and cannot easily move off of it.

src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
bgavrilMS: Oops, bad cherry pick, will undo it!

mjcheetham (commented):
This looks great!

Just a question however.. does this mean that only the WebView2-based embedded browser will support setting a custom title?

How much more work would it be to have this be an `EmbeddedWebViewOptions` (akin to the `SystemWebViewOptions`) where the `Title` property works for both the IE and WV2 windows, and the `BrowserExecutableFolder` property only be respected if a `UseWebView2` boolean property was set to `true`?

I'm not sure what your plans are in the future (near or far) about supporting the legacy web view, or if "Embedded == WebView2" only in that future.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
```diff
@@ -115,6 +115,24 @@ public AcquireTokenInteractiveParameterBuilder WithSystemWebViewOptions(SystemWe
...
 
+        /// <summary>
+        /// Specifies options for using the embedded wevbiew for interactive authentication.
+        /// </summary>
+        /// <param name="options">Data object with options</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        #if !SUPPORTS_WEBVIEW2 // currently only WebView2 allows customization
+        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+        #endif
+        public AcquireTokenInteractiveParameterBuilder WithEmbeddedWebviewOptions(
```
bgavrilMS, line 126: @jmprieur @jennyf19 - based on @mjcheetham 's feedback, I decided to rename this to `WithEmbeddedWebviewOptions` to match the existing `WithSystemWebViewOptions`

The actual options only affect WV2 and document this.

PR #2453, opened 2021-03-04
Adding SDL editorconfigs for IDE

trwalke:
fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2419
Adding editor config for SDL rules
converting old solution wide ruleset to editorconfig


trwalke:
the src/SdlReccomendedError.editorconfig overwright the config in the top level solution wide editor config

src/.editorconfig
```diff
@@ -0,0 +1,242 @@
+# editorconfig.org
```
trwalke, line 1: Since this config is down one level in the folder hierarchy, it will overwrite the rules in the SDL editorconfig, allowing us to merge the two rulesets together

LibsAndSamples.sln
```diff
@@ -78,12 +78,11 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MultiCloudTestApp", "tests\
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "all", "all", "{C44AADC0-1E9B-427F-AC02-642BDF8D936B}"
 	ProjectSection(SolutionItems) = preProject
-		.editorconfig = .editorconfig
+		src\.editorconfig = src\.editorconfig
```
bgavrilMS, line 81: AFAIK .editorconfig is hierarchical, so this ensures that MSAL and MSAL.Desktop (i.e. our products) are covered. However, we want all the code to be covered. Does that produce too many errors?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -263,6 +263,10 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.RulesetToEditorconfigConverter" Version="3.0.0" />
```
bgavrilMS, line 267: Please this package reference under "build tools" group above and add "PrivateAssets=All". This makes the dependency as "build only", i.e. our runtime package must not depend on it.

src/.editorconfig
```diff
@@ -0,0 +1,242 @@
...
+# Copied from the .NET Core repo https://github.com/dotnet/corefx/blob/master/.editorconfig
```
bgavrilMS, line 2: Doesn't this produce a ton of warnings / errors? We should get them fixed (possibly in a different PR).

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
bgavrilMS, line 267: I'd like to better understand this solution, let's sync up on this, as me asking tons of questions on the PR is not practical :)

.editorconfig
```diff
@@ -1,161 +1,798 @@
-# editorconfig.org
-# Copied from the .NET Core repo https://github.com/dotnet/corefx/blob/master/.editorconfig
+# NOTE: Requires **VS2019 16.3** or later
 
-# top-most EditorConfig file
+# Microsoft SDL Roslyn Rules - Required and Recommended (Error) PreRelease version
+# Description: This rule set contains all Microsoft SDL required and recommended rules for Microsoft.CodeAnalysis.NetAnalyzers v5.0.3 (and Microsoft.CodeAnalysis.FxCopAnalyzers), configured to error rather than warn.  Generated 2021-02-08T19:40:15Z.
```
trwalke, line 4: This ruleset actually intended for Microsoft.CodeAnalysis.NetAnalyzers v5.0.3 as this one is designed for public consumption according to the Roslyn team. The "IAXXXX" rules are only used when the internal package is used but we do not need those at the moment. They plan on making that internal package public for opensource projects in the future. 

@bgavrilMS

src/.editorconfig
trwalke: Yes, I added a few additional rules to remove a bunch of warnings. More can be removed in a separate PR

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: This is primarily used to convert the ruleset to .editorconfig.  Will explain in instructions

tests/devapps/DesktopTestApp/Program.cs
```diff
@@ -18,7 +18,7 @@ static void Main()
             Application.EnableVisualStyles();
             Application.SetCompatibleTextRenderingDefault(false);
 
-            ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Ssl3;
```
trwalke, line 21: @bgavrilMS new analyzer complains about using this as it is outdated. Is this still needed?

Directory.Build.props
```diff
@@ -19,6 +19,7 @@
          For clarity, without PrivateAssets marked here, anyone consuming Microsoft.Identity.Client
          would also be forced to install these dependencies.  PrivateAssets avoids this problem. -->
     <PackageReference Include="AsyncUsageAnalyzers" Version="1.0.0-alpha003" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" PrivateAssets="All" />    
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" PrivateAssets="All" />
```
bgavrilMS, line 22: This analyzer has been deprecated, see https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019
It has been replaced with Microsoft.CodeAnalysis.NetAnalyzers, which you just referneced. Do we need to keep the old analyzer? 

IMO, the old ruleset is useless and should be fully replaced by the work you are doing. Keep nothing, except maybe the async analyzer (the one that warns if you don't use ConfigureAwait).

src/.editorconfig
bgavrilMS: We should have 1 .editorconfig for all code.

Directory.Build.props
bgavrilMS: i.e. <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)/build/SolutionWideAnalyzerConfig.ruleset</CodeAnalysisRuleSet>  can go.

bgavrilMS (changes requested):
1. Let's try to have a single editorconfig. 
2. FxCop analyzer package is deprecated, can it be removed?
3. It's not clear which ruleset you are using. the recommended SDL one (which is packaged in the fwk) seems ok, but I do not see it being referenced?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -261,6 +261,7 @@
 
   <ItemGroup Label="Build Tools" Condition="$([MSBuild]::IsOsPlatform('Windows'))">
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.CodeAnalysis.RulesetToEditorconfigConverter" Version="3.0.0" PrivateAssets="All" />
```
bgavrilMS, line 264: This is cool, but how does it work? Does it pick up ANY ruleset file associated with the projects and merge it with editorconfig? 

If so, where are you referincing `Sdl.Recommended.Error.ruleset` ?

Directory.Build.props
pmaytak: If I remember correctly, the async analyzer is also obsolete, new Roslyn analyzers have that rule (https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007?view=vs-2019)

trwalke: This is one of the errors @pmaytak but there is another one that we need from the async analyzer that I cant find. seems like they rules may have been moved to a package called Microsoft.VisualStudio.Threading but I cant find too much info on it yet

bgavrilMS (approved):
LGTM, please make sure we don't have warnings in our projects because of these changes before commiting. If there are 100s of warnings, let's talk.

PR #2458, opened 2021-03-08
Add IsBrokerAvailable, IsEmbeddedWebviewAvailable and IsUserInteractive

bgavrilMS:
IsDesktopSession returns false is the OS is headless, e.g. SSH into Linux. Cross-plat implementation for this taken from GCM core repo. 

IsSystemWebviewAvailable / IsEmbeddedWebViewAvailable depend on IsDesktopSessiopn.

Minimal unit tests since this is highly environmental, i.e. build agents can run either with UI or without depending on configuration or default settings. Tests would not have value / i.e. not catch anything.

Fixes #2458 and https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2120

### DIFF
--- a/src/client/Microsoft.Identity.Client.Desktop/MsalDesktopWebUiFactory.cs
+++ b/src/client/Microsoft.Identity.Client.Desktop/MsalDesktopWebUiFactory.cs
@@ -25,11 +25,11 @@ namespace Microsoft.Identity.Client.Desktop
             _isWebView2AvailableFunc = isWebView2AvailableForTest ?? IsWebView2Available;
         }
 
-        public bool IsSystemWebViewAvailable => IsDesktopSession;
+        public bool IsSystemWebViewAvailable => IsUserInteractive;
 
-        public bool IsDesktopSession => DesktopOsHelper.IsDesktopSession();
+        public bool IsUserInteractive => DesktopOsHelper.IsUserInteractive();
 
-        public bool IsEmbeddedWebviewAvailable => IsDesktopSession && IsWebView2Available();
+        public bool IsEmbeddedWebViewAvailable => IsUserInteractive && IsWebView2Available();
 
         public IWebUI CreateAuthenticationDialog(
             CoreUIParent coreUIParent, 
--- a/src/client/Microsoft.Identity.Client.Desktop/MsalDesktopWebUiFactory.cs
+++ b/src/client/Microsoft.Identity.Client.Desktop/MsalDesktopWebUiFactory.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Identity.Client.Desktop
 
         public bool IsDesktopSession => DesktopOsHelper.IsDesktopSession();
 
-        public bool IsEmbeddedWebviewAvailable => throw new NotImplementedException();
+        public bool IsEmbeddedWebviewAvailable => IsDesktopSession && IsWebView2Available();
 
         public IWebUI CreateAuthenticationDialog(
             CoreUIParent coreUIParent, 
@@ -66,7 +66,7 @@ namespace Microsoft.Identity.Client.Desktop
         }
 
 
-        private bool IsWebView2Available(string browserExecutableFolder)
+        private bool IsWebView2Available(string browserExecutableFolder = null)
         {
             try
             {
--- a/src/client/Microsoft.Identity.Client.Desktop/MsalDesktopWebUiFactory.cs
+++ b/src/client/Microsoft.Identity.Client.Desktop/MsalDesktopWebUiFactory.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 using Microsoft.Identity.Client.Internal;
+using Microsoft.Identity.Client.Platforms.Features.DesktopOs;
 using Microsoft.Identity.Client.Platforms.Features.WebView2WebUi;
 using Microsoft.Identity.Client.Platforms.Shared.Desktop.OsBrowser;
 using Microsoft.Identity.Client.UI;
@@ -24,7 +25,11 @@ namespace Microsoft.Identity.Client.Desktop
             _isWebView2AvailableFunc = isWebView2AvailableForTest ?? IsWebView2Available;
         }
 
-        public bool IsSystemWebViewAvailable => true;
+        public bool IsSystemWebViewAvailable => IsDesktopSession;
+
+        public bool IsDesktopSession => DesktopOsHelper.IsDesktopSession();
+
+        public bool IsEmbeddedWebviewAvailable => throw new NotImplementedException();
 
         public IWebUI CreateAuthenticationDialog(
             CoreUIParent coreUIParent, 
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Identity.Client
         /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
         /// parameters, and to create a public client application instance</returns>
         /// <remarks>If your app uses .NET classic or .NET Core 3.x, and you wish to use the Windows broker, 
-        /// please install the nuget package Microsoft.Identity.Client.Desktop and call .WithWindowsBroker(true)</remarks>
+        /// please install the nuget package Microsoft.Identity.Client.Desktop and call .WithDesktopFeatures()</remarks>
         public PublicClientApplicationBuilder WithBroker(bool enableBroker = true)
         {
 #if NET45
@@ -276,6 +276,23 @@ namespace Microsoft.Identity.Client
             return WithParentFunc(() => (object)windowFunc());
         }
 #endif
+
+        /// <summary>
+        /// Returns true if a broker can be used. 
+        /// </summary>
+        /// <remarks>
+        /// On Windows, the broker (WAM) can be used on Win10 and is always installed. See https://aka.ms/msal-net-wam
+        /// On Mac, Linux and older versions of Windows a broker is not available.
+        /// If your application is .NET5, please use the target .net5.0-windows10.0.17763.0 for all Windows versions and target net5.0 to target Linux and Mac.
+        /// If your application is .NET classic or .NET Core 3.1 and you wish to use the Windows Broker, please install Microsoft.Identity.Client.Desktop first and call WithDesktopFeatures().
+        /// 
+        /// On mobile, the device must be Intune joined and Authenticator or Company Portal must be installed. See https://aka.ms/msal-brokers
+        /// </remarks>
+        public bool IsBrokerAvailable()
+        {            
+            return PlatformProxyFactory.CreatePlatformProxy(null)
+                .CreateBroker(base.Config, null).IsBrokerInstalledAndInvokable();
+        }
 
         /// <summary>
         /// </summary>
--- a/src/client/Microsoft.Identity.Client/Extensions.cs
+++ b/src/client/Microsoft.Identity.Client/Extensions.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Extension methods
+    /// </summary>
+    public static class Extensions // TODO MSAL5: move these to their parent types
+    {
+        /// <summary>
+        /// Returns true if MSAL can use a system browser.
+        /// </summary>
+        /// <remarks>
+        /// On Windows, Mac and Linux a system browser can always be used, except in cases where there is no UI, e.g. SSH connection.
+        /// On Android, the browser must support tabs.
+        /// </remarks>
+        public static bool IsSystemWebViewAvailable(this IPublicClientApplication publicClientApplication)
+        {
+            if (publicClientApplication is PublicClientApplication pca)
+            {
+                return pca.IsSystemWebViewAvailable;
+            }
+
+            throw new ArgumentException("This extension method is only available for the PublicClientApplication implementation " +
+                "of the IPublicClientApplication interface.");
+
+        }
+
+        /// <summary>
+        /// Returns true if MSAL can use an embedded webview (browser). 
+        /// </summary>
+        /// <remarks>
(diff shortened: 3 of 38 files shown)


src/client/Microsoft.Identity.Client.Desktop/MsalDesktopWebUiFactory.cs
```diff
@@ -24,7 +25,11 @@ internal class MsalDesktopWebUiFactory : IWebUIFactory
             _isWebView2AvailableFunc = isWebView2AvailableForTest ?? IsWebView2Available;
         }
 
-        public bool IsSystemWebViewAvailable => true;
+        public bool IsSystemWebViewAvailable => IsDesktopSession;
```
bgavrilMS, line 28: Can't have a system / embedded webview if there is no UI ...

src/client/Microsoft.Identity.Client/Extensions.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Extension methods
+    /// </summary>
+    public static class Extensions // TODO MSAL5: move these to their parent types
```
jmprieur, line 13: What about `OsCapabilitiesExtensions`?
I know that when using the IntelliSense, the class won't appear. However when looking at the reference documentation, customers will see the extension class name, and it's better if it's meaninful.

src/client/Microsoft.Identity.Client/Extensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            throw new ArgumentException("This extension method is only available for the PublicClientApplication implementation " +
+                "of the IPublicClientApplication interface.");
+
+        }
+
+        /// <summary>
+        /// Returns true if MSAL can use an embedded webview (browser). 
+        /// </summary>
+        /// <remarks>
+        /// Currently there are no embedded webviews on Mac and Linux. On Windows, app developers or users should install
```
jmprieur, line 38: On Windows? depending on the Windows platform the old embedded web view is still available?
Would it be on Windows 10? is installing WebView2 a requirement?

src/client/Microsoft.Identity.Client/Extensions.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            throw new ArgumentException("This extension method is only available for the PublicClientApplication implementation " +
+                "of the IPublicClientApplication interface.");
+        }
+
+        /// <summary>
+        /// Returns false when the program runs in headless OS, for example when SSH-ed into a Linux machine.
+        /// Browsers (webviews) and brokers cannot be used if there is no UI support. Instead, please use <see cref="PublicClientApplication.AcquireTokenWithDeviceCode(IEnumerable{string}, Func{DeviceCodeResult, Task})"/>
+        /// </summary>
+        public static bool IsDesktopSession(this IPublicClientApplication publicClientApplication)
```
jmprieur, line 56: Should it be `IsInDesktopSession`?

src/client/Microsoft.Identity.Client/Extensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            throw new ArgumentException("This extension method is only available for the PublicClientApplication implementation " +
+                "of the IPublicClientApplication interface.");
+        }
+
+        /// <summary>
+        /// Returns false when the program runs in headless OS, for example when SSH-ed into a Linux machine.
+        /// Browsers (webviews) and brokers cannot be used if there is no UI support. Instead, please use <see cref="PublicClientApplication.AcquireTokenWithDeviceCode(IEnumerable{string}, Func{DeviceCodeResult, Task})"/>
+        /// </summary>
```
jmprieur, line 55: And AcquireTokenSilent when WAM? AcquireTokenByIntegratedWindowsAuthentication

src/client/Microsoft.Identity.Client/Extensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            throw new ArgumentException("This extension method is only available for the PublicClientApplication implementation " +
+                "of the IPublicClientApplication interface.");
+        }
+
+        /// <summary>
+        /// Returns the certificate used to create this <see cref="ConfidentialClientApplication"/>, if any.
+        /// </summary>
+        public static X509Certificate2 GetCertificate(this IConfidentialClientApplication confidentialClientApplication)
```
jmprieur, line 89: Fixes https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2120 without a breaking change

src/client/Microsoft.Identity.Client/Extensions.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            throw new ArgumentException("This extension method is only available for the PublicClientApplication implementation " +
+                "of the IPublicClientApplication interface.");
+        }
+
+        /// <summary>
+        /// Returns true if a broker can be used. 
+        /// </summary>
+        /// <remarks>
+        /// On Windows, the broker (WAM) can be used on Win10 and is always installed.
```
jmprieur, line 71: we probably want to provide aka.ms links for each of these?
- https://aka.ms/msal-net-wam
- https://aka.ms/msal-brokers
- https://aka.ms/msal-net-ios-broker, https://aka.ms/msal-net-ios-13-broker
- https://aka.ms/Brokered-Authentication-for-Android

jmprieur (commented):
LGTM. Great addition. Thanks @bgavrilMS 

Also thanks @mjcheetham for pushing this requirement. I had not realized the code to produce these capability flags was so painful to write.

I've left a few comments on naming of the public API and XML comments.

src/client/Microsoft.Identity.Client/Extensions.cs
```diff
@@ -0,0 +1,101 @@
+﻿using System;
```
jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿﻿using System;
```

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/DesktopOsHelper.cs
```diff
@@ -0,0 +1,141 @@
+﻿using System;
```
jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿using System;
```

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/MacNativeMethods.cs
```diff
@@ -0,0 +1,210 @@
+﻿using System;
```
jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿using System;
```

src/client/Microsoft.Identity.Client/Extensions.cs
jennyf19: +1

---
In reply to: [589412204](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2458#discussion_r589412204) [](ancestors = 589412204)

src/client/Microsoft.Identity.Client/Extensions.cs
jennyf19: any value in having extensions for pca and extensions for cca separately?

---
In reply to: [589415189](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2458#discussion_r589415189) [](ancestors = 589415189)

src/client/Microsoft.Identity.Client/UI/IWebUIFactory.cs
```diff
@@ -15,6 +15,7 @@ internal interface IWebUIFactory
             RequestContext requestContext);
 
         bool IsSystemWebViewAvailable { get; }
-        
+        bool IsDesktopSession { get; }
```
pmaytak, line 18: Would it be more accurate to name this `IsUiSession` since it's also used in Android and iOS factories?

src/client/Microsoft.Identity.Client/UI/IWebUIFactory.cs
```diff
@@ -15,6 +15,7 @@ internal interface IWebUIFactory
...
+        bool IsEmbeddedWebviewAvailable { get; }
```
pmaytak, line 19: Capitalize V, `IsEmbeddedWebviewAvailable` -> `IsEmbeddedWebViewAvailable`

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopWebUIFactory.cs
```diff
@@ -12,11 +13,21 @@ namespace Microsoft.Identity.Client.Platforms.net45
...
-        public bool IsSystemWebViewAvailable => true;
+        public bool IsSystemWebViewAvailable => IsDesktopSession;
+
+        public bool IsDesktopSession
+        {
+            get
+            {
+                return DesktopOsHelper.IsDesktopSession();
+            }
+        }
```
pmaytak, line 24:
```suggestion
        public bool IsDesktopSession => DesktopOsHelper.IsDesktopSession();
```

pmaytak (approved):

src/client/Microsoft.Identity.Client/Extensions.cs
bgavrilMS: It is only available on .NET FWK, I did not add it to .NET Core or .NET 5. I believe we discussed about this, but we can rediscuss if needed. 

WV2 needs to be installed. It is avaiable on Win7, 8, 8.1 and 10. Not sure about server versions, most likely it works.

src/client/Microsoft.Identity.Client/Extensions.cs
bgavrilMS: Updated.

src/client/Microsoft.Identity.Client/Extensions.cs
bgavrilMS: Done.

src/client/Microsoft.Identity.Client/Extensions.cs
bgavrilMS: Thank you for looking up the work item @jmprieur , apreciate it :). I remembered the problem, but forgot we were still tracking it....

bgavrilMS: @jennyf19 - I can't think of any.

src/client/Microsoft.Identity.Client/Extensions.cs
bgavrilMS: Naming is hard, so I don't know how it "should" be. Personally, I don't like the "IsIn". This is how @mjcheetham named the helper in GCM (granted GCM is not a library, it's an app).

src/client/Microsoft.Identity.Client/UI/IWebUIFactory.cs
bgavrilMS: I will name it similarly to .NET5, i.e. https://github.com/dotnet/runtime/blob/4f67869210c4d1014255bd5ed907e78f795f9bef/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L145-L165

src/client/Microsoft.Identity.Client/Extensions.cs
jennyf19: @bgavrilMS i meant more of if we ever try to split the library again, cca/pca, then we'd have separate extensions for both already.

src/client/Microsoft.Identity.Client/OsCapabilitiesExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+                "of the IPublicClientApplication interface.");
+        }
+
+        /// <summary>
+        /// Returns false when the program runs in headless OS, for example when SSH-ed into a Linux machine.
+        /// Browsers (webviews) and brokers cannot be used if there is no UI support. 
+        /// Instead, please use <see cref="PublicClientApplication.AcquireTokenWithDeviceCode(IEnumerable{string}, Func{DeviceCodeResult, Task})"/>
+        /// or <see cref="PublicClientApplication.AcquireTokenByIntegratedWindowsAuth(IEnumerable{string})"/>
+        /// </summary>
+        public static bool IsUserInteractive(this IPublicClientApplication publicClientApplication)
```
bgavrilMS, line 60: @jmprieur @jennyf19 @pmaytak - I renamed "IsDesktopSession" to "IsUserInteractive". This is the wording used by .NET5

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -277,6 +277,23 @@ public PublicClientApplicationBuilder WithParentActivityOrWindow(Func<IntPtr> wi
...
+        /// </summary>
+        /// <remarks>
+        /// On Windows, the broker (WAM) can be used on Win10 and is always installed. See https://aka.ms/msal-net-wam
+        /// On Mac, Linux and older versions of Windows a broker is not available.
+        /// If your application is .NET5, please use the target .net5.0-windows10.0.17763.0 for all Windows versions and target net5.0 to target Linux and Mac.
+        /// If your application is .NET classic or .NET Core 3.1 and you wish to use the Windows Broker, please install Microsoft.Identity.Client.Desktop first and call WithDesktopFeatures().
+        /// 
+        /// On mobile, the device must be Intune joined and Authenticator or Company Portal must be installed. See https://aka.ms/msal-brokers
+        /// </remarks>
+        public bool IsBrokerAvailable()
```
bgavrilMS, line 291: @jmprieur @jennyf19 @pmaytak - I moved `IsBrokerAvaiable()` to the builder, instead of on the app. The reason is that `WithBroker` is an app property, so you'd have to create the app via `app = builder.WithBroker()` and then call `app.IsBrokerAvailable()`

The more natural way is to have it on the app builder: 

```csharp
var builder = PublicClientApplicationBuilder().Create("client_id");
if (builder.IsBrokerAvailable()) 
{
 Console.WriteLine("No broker! Install one!"); 
} 
else 
{ 
     builder = builder.WithBroker();
}
}

```

Thx @mjcheetham for pointing out this problem!

jennyf19: makes sense. thanks for the heads-up

PR #2460, opened 2021-03-08
Add support for Win 7, 8 10 (<17663) to NET5

bgavrilMS:
Please see https://docs.microsoft.com/en-us/dotnet/standard/analyzers/platform-compat-analyzer and 
    https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -157,7 +157,7 @@ namespace Microsoft.Identity.Client
             {
                 throw new PlatformNotSupportedException(
                     "If you have a Windows application which targets net5 or net5-windows, please change the target to net5-windows10.0.17763.0, which provides support from Win7 to Win10. For details see https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md" +
-                    "If you have a cross-platform (Windows, Mac, Linux) application which targets net5, please dual target net5 and net5-windows. Your installer should deploy the net5 version on Mac and Linux and the net5-window10.0.17763.0 on Win7-Win10. For details see https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md" +
+                    "If you have a cross-platform (Windows, Mac, Linux) application which targets net5, please dual target net5 and net5-windows10.0.17763.0. Your installer should deploy the net5 version on Mac and Linux and the net5-window10.0.17763.0 on Win7-Win10. For details see https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md" +
                     "If you have a .NET Core 3.1 application, please install the nuget package named Microsoft.Identity.Client.Desktop and call the extension method .WithDesktopFeatures() first.");
             }
 #endif
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Identity.Client
         /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
         /// parameters, and to create a public client application instance</returns>
 #if !iOS
-        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)] 
+        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
 #endif
         public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keychainSecurityGroup)
         {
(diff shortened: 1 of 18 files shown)


build/SolutionWideAnalyzerConfig.ruleset
```diff
@@ -79,7 +79,7 @@
   </Rules>
   <Rules AnalyzerId="Microsoft.NetCore.Analyzers" RuleNamespace="Microsoft.NetCore.Analyzers">
     <Rule Id="CA1304" Action="Error" />
-    <Rule Id="CA1305" Action="Warning" />
+    <Rule Id="CA1305" Action="None" />
```
bgavrilMS, line 82: unrelated..

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -71,9 +71,13 @@
...
-    <!-- Remove after Dec Release of .NET SDK -->
-    <WindowsSDKVersion>10.0.17763.0</WindowsSDKVersion>
     <UseWindowsForms>true</UseWindowsForms>
+    <!-- 
+    MSAL will run on Windows 7 and 8 but requires to be built against Win10 to use WinRT APIs for WAM 
+    See https://docs.microsoft.com/en-us/dotnet/standard/analyzers/platform-compat-analyzer and 
+    https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md for details
+    -->
+    <SupportedOSPlatformVersion>7</SupportedOSPlatformVersion>
```
bgavrilMS, line 80: Functionally, this is all that is needed.

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/StaTaskScheduler.cs
```diff
@@ -13,6 +13,9 @@ namespace Microsoft.Identity.Client.Platforms.Features.DesktopOs
     // This IDisposable class doe not need to implement Dispose method in standard way, because it is sealed.
     // If it ever needs to become inheritable, it should follow the standard pattern as desribed in http://msdn.microsoft.com/en-us/library/fs2xkftw(v=vs.110).aspx.
     /// <summary>Provides a scheduler that uses STA threads.</summary>
+#if NET5_WIN
+    [System.Runtime.Versioning.SupportedOSPlatform("windows")]
```
bgavrilMS, line 17: Don't know why the analyzer complains as all these types are internal... Logged a bug on .net team ... https://github.com/dotnet/roslyn-analyzers/issues/4933 but for now, annotating the warnings.

bgavrilMS:
CC @mjcheetham

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -135,19 +135,33 @@ public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keycha
...
                 throw new PlatformNotSupportedException(
-                    "The Windows broker is not directly available on MSAL for .NET Framework or .NET Core 3.x" +
+                    "The Windows broker is not directly available on MSAL for .NET Framework" +
                     " To use it, please install the nuget package named Microsoft.Identity.Client.Desktop " +
-                    "and call the extension method .WithWindowsBroker(true)");
+                    "and call the extension method .WithDesktopFeatures() first.");
+            }
+#endif
+
+#if NET_CORE
```
bgavrilMS, line 155: @mjcheetham - does this look right to you? When you target .net5 on Mac and Linux, it will NO-OP for now (until we have broker on Mac / Linux). On Windows, it will throw this exception.

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/StaTaskScheduler.cs
pmaytak, line 17: Why do we need to specify this? Is the default support 17663+ because of our target?

My understanding if I run MSAL on Win 7, then all these classes are available, unless they are marked with `SupportedOSPlatform("windows10.0.17763.0")`, which makes them available on 17763+.

Or is this to specify that this is available only on windows and not linux and such?

pmaytak (commented):
LGTM (although I admit it takes a bit time to get used to these different version combinations).

Is my understanding correct?
- We target net5 17663 to use WAM APIs.
- `<SupportedOSPlatformVersion>7</SupportedOSPlatformVersion>` allows MSAL to run on Windows 7+.
- But now we have to use `SupportedOSPlatform` to hide the 17663+ APIs from older runtimes.

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/StaTaskScheduler.cs
bgavrilMS: It looks like 

`thread.SetApartmentState(ApartmentState.STA);`

is available only on Windows. 

I don't know why the analyzer is complaining about this, since net5-windows10.xyz is only available on Windows :) 
I'll log another bug on them I think.

Issue #1967, opened 2020-07-31
[Enhancement] GetAccountsAsync should no longer be exposed in confidential client applications (via deprecation)


jmprieur:
**Which Version of MSAL are you using ?**
4.17.0

**Platform**
net45, netcore

**What authentication flow has the issue?**
* Web App
    * [ ] Authorization code
    * [ ] OBO
* Web API
    * [ ] OBO

**Repro**

```csharp
var accounts = await confidentialClientApplications.GetAccountsAsync();
```

**Expected behavior**
`GetAccountsAsync()` does not make sense in confidential client applications as there should be one cache per user and GetAccountsAsync() does not know which cache key to use.

We'd want to have a warning at build time: Use GetAccountAsync in web apps and web APIs, and use a token cache serializer for better security and performance. See https://aka.ms/msal-net-cca-token-cache-serialization.

**Proposed spec**
In `IConfidentialClientApplication`, add a new method (with the same signature as in the base interface):

```CSharp
        /// <inheritdoc/>
        [Obsolete("Use GetAccountAsync in web apps and web APIs, and use a token cache serializer for better security and performance. See https://aka.ms/msal-net-cca-token-cache-serialization.")]
        new Task<IEnumerable<IAccount>> GetAccountsAsync()
```

Given the signature is the same as in the base interface this is not really adding a new method to the interface, and therefore, not a breaking change.

**Actual behavior**
It's possible to call `GetAccountsAsync()`, but it always returns 0 accounts (unless developers don't override serialization, but this should not be done in confidential client applications).

jmprieur:
Proposing, in 4.28.0 to just:
- add a new method GetAccountsAsync on IConfidentialClientApplication (at the moment it's only on ApplicationBase)
- add an obsolete attribute, with warning? The message should be: `Use GetAccountAsync in web apps and a token cache serializer for better security and performance. See  https://aka.ms/msal-net-cca-token-cache-serialization`

PR #2464, opened 2021-03-09
Obsolete attribute to GetAccounts for CCA.

pmaytak:
Issue #1967

`GetAccountsAsync` in `IConfidentialClientApplication` overwrites one in `IClientApplicationBase`. Adds `obsolete` attribute.


jmprieur (approved):
LGTM. Thanks @pmaytak
I guess this does not make sense to think of units tests in this case?

pmaytak:
@jmprieur Yea, I only tested manually with PCA and CCA apps and the warning shows correctly.

Issue #2461, opened 2021-03-09
[M] CCA cache - Log error when cache is not serialized


pmaytak:
For confidential client apps, log error when cache is not serialized.

jmprieur:
@pmaytak, , 
- we should log an error (but not throw)
- the error message should be: `The default token cache provided by MSAL is not designed to be performant when used in confidential client applications. Please use token cache serialization. See https://aka.ms/msal-net-cca-token-cache-serialization.`

BobBonser:
I am getting this error in my log even though I do have AddInMemoryCaches in my startup

            var initialScopes = Configuration.GetValue<string>("DownstreamApi:Scopes")?.Split(' ');

            services.AddMicrosoftIdentityWebAppAuthentication(Configuration, "AzureAd")
                .EnableTokenAcquisitionToCallDownstreamApi(initialScopes)
                    .AddMicrosoftGraph(Configuration.GetSection("DownstreamApi"))
                    .AddInMemoryTokenCaches();

pmaytak:
Thanks @BobBonser, Is this the same issue as https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2990?

jmprieur:
Which version of Id.Web do you use. @BobBonser ?

BobBonser:
It does appear to be the same issue.

BobBonser:
Not sure what I'd.Web means.

bgavrilMS:
@BobBonser - Microsoft.Identity.Web ... But let's continue the conversation of #2990, as this issue is closed.

PR #2466, opened 2021-03-09
Log error when CCA doesn't use cache serialization.

pmaytak:
Issue #2461

As discussed with Bogdan, logging this in the `RequestBase` seemed the most appropriate because during CCA/PCA app instance creation/building we don't know yet if the cache is serialized, and we don't have existing way elsewhere to check if app instance is PCA or CCA.


pmaytak:
> Do I understand that this will be logged at each token request?

@jmprieur Yes, at each request.

jmprieur:
That's going to be a good incentive to use a serializer.
@henrik-me do you think this would be too much? (and we'd want to reach threashold before logging the errors?)

PR #2472, opened 2021-03-11
Fix some warnings and treat compiler warnings as errors

bgavrilMS:


Directory.Build.props
```diff
@@ -11,6 +11,7 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <IsPackable>false</IsPackable>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
bgavrilMS, line 14: This is for compiler warnings, which are different than Roslyn analyzers. 

@jennyf19 - consider enabling this in M.I.W as well.

jennyf19: How is line 14 different from line 3? should there be a condition included or one removed?

trwalke: So we are now treating all warnings as errors for the entire project? Will the Xamarin projects also be affected by this?

bgavrilMS: I'm struggling with this, there are too many warnings from NuGet / msbuild that get treated as errors :( in the CI build, but not locally!

But if you don't have any MSB or NU warning, I think it's the way to go.

Issue #2476, opened 2021-03-15
[Bug] WAM - if authority ends in /organizations, logging in interactively with a login hint fails


bgavrilMS:
MSAL 4.27

**Repro Steps**

1.	Set the authority to https://login.microsoftonline.com/organizations
2.	Sign-In Interactively with an AAD user that has an account in the OS (for example, via login hint). The sign in must be interactive (i.e. force auth or the user has sign-in frequency CA, or consent is required etc.)

**Actual**: 

![image](https://user-images.githubusercontent.com/12273384/111176208-0977b100-85a1-11eb-81e4-ce2814f805ce.png)

Note: using /common works

**Logs**

--> ORG (FAILS)

 === WebTokenRequest ===
PromptType: ForceAuthentication
webTokenRequest.Property: authority: https://login.microsoftonline.com/organizations/
WebAccountProvider: 
=== WebAccountProvider ===
Authority organizations
DisplayName Work or school account
Id https://login.microsoft.com

UserName subzero@bogavrilltd.onmicrosoft.com
webAccount.Property: UPN: subzero@bogavrilltd.onmicrosoft.com
webAccount.Property: TenantId: 49f548d0-12b7-4169-a390-bb5304d24462
webAccount.Property: UserName: subzero@bogavrilltd.onmicrosoft.com
WebAccountProvider: 
=== WebAccountProvider ===
Authority organizations
DisplayName Work or school account
Id https://login.microsoft.com

--> COMMON (WORKS)

=== WebTokenRequest ===
PromptType: ForceAuthentication
webTokenRequest.Property: authority: https://login.microsoftonline.com/common/
WebAccountProvider: 

=== WebAccountProvider ===
Authority common
DisplayName Work or school account
Id https://login.microsoft.com

== WAM ACCOUNT == 
webAccount.Property: UPN: subzero@bogavrilltd.onmicrosoft.com
webAccount.Property: Authority: https://login.microsoftonline.com/common
WebAccountProvider: === WebAccountProvider ===
Authority organizations
DisplayName Work or school account
Id https://login.microsoft.com

Impact: 

|                     | Works ?|
| ------------- |:-------------:|
| AAD User + normal app    | no |
| MSA User + normal app  | yes |
| AAD User + MSA-PT app    | no |
| MSA User + MSA-PT app  | yes |

bgavrilMS:
Possible solution:

```csharp
[...]

PR #2479, opened 2021-03-16
Partial fix for #2476 - WAM + /organizations + login hint

bgavrilMS:
Please see #2476 for details 
Fix works for: 

|                     | Works ?|
| ------------- |:-------------:|
| AAD User + normal app    | yes |
| MSA User + normal app  | yes |
| AAD User + MSA-PT app    | yes |
| MSA User + MSA-PT app  | no |

which is better than current status, where AAD users cannot login.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
             _msaPassthroughHandler = msaPassthroughHandler ??
                 new MsaPassthroughHandler(_logger, _msaPlugin, _wamProxy, _parentHandle);
 
-            _wamOptions = appConfig.WindowsBrokerOptions ?? 
+            _wamOptions = appConfig.WindowsBrokerOptions ??
                 WindowsBrokerOptions.CreateDefault();
 
         }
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
```diff
@@ -235,12 +235,20 @@ private bool IsAadOnlyAuthority(Authority authority)
...
                 provider,
                 authenticationRequestParameters,
-                isForceLoginPrompt: false,
+                isForceLoginPrompt: false, 
                 isInteractive: true,
                 isAccountInWam: true)
            .ConfigureAwait(false);
 
-            WamAdapters.AddMsalParamsToRequest(authenticationRequestParameters, webTokenRequest);          
+            // because of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2476
```
jennyf19, line 243: this seems like a work around, shouldn't the actual fix be in WAM? maybe create a work time so we come back to this later?

bgavrilMS: I'm keeping the work item open, since this does not fully fix the problem. Discussion with WAM is ongoing.

bgavrilMS: Agreed it's a hack.

pmaytak (approved):

PR #2490, opened 2021-03-22
clear cache when deserializing msal v3 when shouldClearExistingCache is set to true

jennyf19:


jmprieur (approved):
LGTM:
Thanks @jennyf19 

BTW, this will also solve the real issue with the singleton (which we closed in the past as non-repro because the repro was not exhibiting this specific behavior)

bgavrilMS (approved):

henrik-me (approved):
:shipit:

Issue #1845, opened 2020-05-25
[Enhancement][L] Enable OBO for SP (ADAL to MSAL Migration Issue)


niruatweb:
**Work**

- [x] See https://github.com/Penguinwizzard/microsoft-authentication-library-for-dotnet/commit/de00f6162e85536ff043bf5b0e87dc0c35c23b02 for a solution
- [x] Evaluate if solution will work, e.g. token cache 
- [x] Create an E2E test using PPE env where this is allowed.

**Which Version of MSAL are you using ?**
4.13.0. I have described the issue in stack overflow. https://stackoverflow.microsoft.com/questions/198842. Generating OBO for SPN fails with "client info is null"

**Platform**
net45

**What authentication flow has the issue?**
* Web API
    * [x] OBO for SPNs

Other? - please describe;
OBO for users work fine but not for groups.

**Is this a new or existing app?**
YES

**Repro**

```await _confidentialApp.AcquireTokenOnBehalfOf(_userScopes, 
  new UserAssertion(userBearerToken, "urn:ietf:params:oauth:grant-type:jwt-bearer"))
    .WithAuthority(aadAuthenticationAuthority)
    .ExecuteAsync();
```
```client info is null

==== Complete Stack Trace ===
Microsoft.Identity.Client.MsalClientException: client info is null
   at Microsoft.Identity.Client.Core.ClientInfo.CreateFromJson(String clientInfo)
   at Microsoft.Identity.Client.Internal.Requests.RequestBase.<CacheTokenResponseAndCreateAuthenticationResultAsync>d__17.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Identity.Client.Internal.Requests.OnBehalfOfRequest.<ExecuteAsync>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Identity.Client.Internal.Requests.RequestBase.<RunAsync>d__14.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #2506, opened 2021-03-24
Adding support for Service principasl on OBO

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1845

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -190,7 +190,8 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
             if (!AuthenticationRequestParameters.IsClientCredentialRequest &&
                 AuthenticationRequestParameters.ApiId != ApiEvent.ApiIds.AcquireTokenByRefreshToken &&
-                AuthenticationRequestParameters.AuthorityInfo.AuthorityType != AuthorityType.Adfs)
+                AuthenticationRequestParameters.AuthorityInfo.AuthorityType != AuthorityType.Adfs &&
+                !(msalTokenResponse.ClientInfo is null))
             {
                 //client_info is not returned from client credential flows because there is no user present.
                 fromServer = ClientInfo.CreateFromJson(msalTokenResponse.ClientInfo);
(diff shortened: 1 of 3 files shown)


jmprieur (approved):
Thanks @trwalke 
LGTM

Given the very specifics of the app registration don't we want to add a comment in the test, pointing to an internal page explaining the end to end ?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -190,7 +190,8 @@ protected async Task<AuthenticationResult> CacheTokenResponseAndCreateAuthentica
 
             if (!AuthenticationRequestParameters.IsClientCredentialRequest &&
                 AuthenticationRequestParameters.ApiId != ApiEvent.ApiIds.AcquireTokenByRefreshToken &&
-                AuthenticationRequestParameters.AuthorityInfo.AuthorityType != AuthorityType.Adfs)
+                AuthenticationRequestParameters.AuthorityInfo.AuthorityType != AuthorityType.Adfs &&
+                !(msalTokenResponse.ClientInfo is null))
```
bgavrilMS, line 194: We need to add some logging statements to make sure we understand we're dealing with SP OBO and not normal OBO, if we read the MSAL's logs. Perhaps in the OBO request class ...

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -693,6 +698,51 @@ public async Task ClientCreds_ClientAssertion_AAD_NoWilson_Async()
...
+                                                         .ExecuteAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(authenticationResult);
+            Assert.IsNotNull(authenticationResult.AccessToken);
+
+            authenticationResult = await _confidentialApp.AcquireTokenOnBehalfOf(scopes2, userAssertion)
+                                                         .WithAuthority(PPEAuthenticationAuthority)
+                                                         .ExecuteAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(authenticationResult);
```
bgavrilMS, line 739: Please add an Assert on authenticationResult.ResultMetadata.TokenSource - it should be from the cache.

bgavrilMS (approved):

Issue #2418, opened 2021-02-17
[Bug] Cannot use Custom Tabs with Android 29+


trwalke:
**Logs and Network traces**
Without logs or traces, it is unlikely that the team can investigate your issue. Capturing logs and network traces is described at https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging

**Which Version of MSAL are you using ?**
MSAL 4.24.0 +

**Platform**
xamarin android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App 
    * [ ] Authorization code
    * [ ] OBO
* Daemon App 
    * [ ] Service to Service calls

**Repro**
Ensure the app is targeting android 10 (API 29) in the project properties.
Ensure that the system browser is enabled in MSAL.NET (Should be on by default)
Try to acquire a token interactively

**Expected behavior**
MSAL should launch the custom tab browser for authentication

**Actual behavior**
The system browser is launched without custom tabs.

**Possible Solution**
Try modifying the browser tab packages that are used in the interactive auth.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d451b16a9ae66742047b0ba37fb74d875a69056d/src/client/Microsoft.Identity.Client/Platforms/Android/SystemWebview/AuthenticationActivity.cs#L37-L45

jmprieur:
@trwalke. is it a duplicate of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2139 ?

bgavrilMS:
Closing as duplicate of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2139

trwalke:
Not a duplicate. this is the second issue identified in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2139

trwalke:
@tipa it appears that there is a new way to implement the custom tabs in android 29. Ive been playing around with it to see if there is a way to adapt our code with it to make it work. I managed to get the custom tab to show up but the responses do not return to our code at the moment. I was also able to get the legacy support for custom tabs to work so I think it will be ok for us to use that for now.

tipa:
[...]

PR #2510, opened 2021-03-26
Using android.support.customtabs for android api 29+

trwalke:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2418
The proper way to support custom tabs on android 29+ is to use the new custom tab intent as described [here](https://developer.android.com/reference/androidx/browser/customtabs/CustomTabsIntent).

While trying out the new CustomTabsIntent, the custom tab shows up but our code does not capture the auth response. It will take a bit of work in MSAL to get this to happen so in the meantime we can use the android.support.customtabs implemented in this PR

Tested on andoridX and 11


pmaytak (commented):
Is there a reason to use the androidx custom tabs besides that androidx are newer libraries? Can we just keep classic support custom tabs if it works on all Android versions?

trwalke:
Not entirely sure yet. I will need to sync with the android team for more info. However, this is working with classic support for now

pmaytak (approved):

jennyf19 (approved):
:shipit:

PR #2513, opened 2021-03-29
ValidateAuthority must be false when using regional

bgavrilMS:
When using regional authorities, they are not validated and as such, `validateAuthority` should be set to false explicitly to drive expectations. 

A regionalized authority is seen as invalid by ESTS, e.g. `westus.microsoft.com` is not valid. So there is no support for validating the regionalized authority.

There is however support to validate the original authority, i.e. `login.microsoftonline.com` or whatever is passed in. And we could assume that if original authority is valid, then `region + original env` is also valid. This would be an improvement over the existing implemenattion and should be treated separately.

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -232,6 +232,15 @@ namespace Microsoft.Identity.Client
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (!string.IsNullOrEmpty(Config.AzureRegion) && Config.AuthorityInfo.ValidateAuthority == true)
+            {
+                throw new MsalClientException(
+                    MsalError.RegionalAuthorityValidation,
+                    "You configured both Regional Authority and Authority Validation. Authority validation is not currently supported for regional Azure authorities." +
+                    "You can set the validateAuthority flag to false to use Azure Regional authority. Do not disable authority validation if you read the authority from a potentially unstrusted source, " +
+                    "for example from the WWWAuthenticate header of an HTTP request that resulted in a 401 response." );                    
+            }
         }
 
         /// <summary>
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -950,5 +950,14 @@ namespace Microsoft.Identity.Client
         ///  or the app developer can install the WebView2 runtime https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution
         /// </summary>
         public const string WebView2NotInstalled = "webview2_runtime_not_installed";
+
+
+        /// <summary>
+        /// <para>What happens?</para>You configured both Regional Authority and Authority Validation. Authority validation is not currently supported for regional authorities.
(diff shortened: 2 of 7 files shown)


trwalke (approved):
Makes sense to be more upfront about it for this scenario.
LGTM

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -232,6 +232,15 @@ internal override void Validate()
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (!string.IsNullOrEmpty(Config.AzureRegion) && Config.AuthorityInfo.ValidateAuthority == true)
```
jmprieur, line 236: Given that the default is true, this makes the developer experience more complicated? We might want to have an aka.ms link to explain everything?

jmprieur (commented):
I've left a suggestion

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -232,6 +232,15 @@ internal override void Validate()
...
+            {
+                throw new MsalClientException(
+                    MsalError.RegionalAuthorityValidation,
+                    "You configured both Regional Authority and Authority Validation. Authority validation is not currently supported for regional Azure authorities." +
+                    "You can set the validateAuthority flag to false to use Azure Regional authority. Do not disable authority validation if you read the authority from a potentially unstrusted source, " +
```
jennyf19, line 241:
```suggestion
                    "You can set the validateAuthority flag to false to use Azure Regional authority. Do not disable authority validation if you read the authority from a potentially untrusted source, " +
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -950,5 +950,14 @@ public static class MsalError
         ///  or the app developer can install the WebView2 runtime https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution
         /// </summary>
         public const string WebView2NotInstalled = "webview2_runtime_not_installed";
+
+
+        /// <summary>
+        /// <para>What happens?</para>You configured both Regional Authority and Authority Validation. Authority validation is not currently supported for regional authorities.
+        /// <para>Mitigation</para>Set the validateAuthority flag to false to use Azure Regional authority. Do not disable authority validation if you read the authority from a potentially unstrusted source,
```
jennyf19, line 957:
```suggestion
        /// <para>Mitigation</para>Set the validateAuthority flag to false to use Azure Regional authority. Do not disable authority validation if you read the authority from a potentially untrusted source, 
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -950,5 +950,14 @@ public static class MsalError
...
         /// </summary>
         public const string WebView2NotInstalled = "webview2_runtime_not_installed";
+
+
+        /// <summary>
+        /// <para>What happens?</para>You configured both Regional Authority and Authority Validation. Authority validation is not currently supported for regional authorities.
+        /// <para>Mitigation</para>Set the validateAuthority flag to false to use Azure Regional authority. Do not disable authority validation if you read the authority from a potentially unstrusted source, 
+        /// for example from the WWWAuthenticate header of an HTTP request that resulted in a 401 response. 
+        ///  </summary>
+        public const string RegionalAuthorityValidation = "regional_autority_validation";
```
jennyf19, line 960:
```suggestion
        public const string RegionalAuthorityValidation = "regional_authority_validation";
```

jennyf19 (approved):

PR #2516, opened 2021-03-30
Bogavril/x client os

### LINKED ISSUE #2517: [Bug] MSAL not sending correct OS version string on .NET classic
msal 4.29

MSAL sends x-client-os header with version string. THis is used by the server for device auth.
On .NET classic, this functionality is broken, as we mostly just send "6.2.5900" due to a bug in .NET fwk (see https://stackoverflow.com/questions/33328739/system-environment-osversion-returns-wrong-version)

This fixes the issue by send the correct OS description, e.g. "Windows 10", "Windows Server 2016" etc.

bgavrilMS:
Send correct x-client-os OS info on .NET in header

Fixes #2517


src/client/Microsoft.Identity.Client/Platforms/uap/UapPlatformProxy.cs
```diff
@@ -107,7 +107,7 @@ protected override string InternalGetOperatingSystem()
         {
             // In WinRT, there is no way to reliably get OS version. All can be done reliably is to check
             // for existence of specific features which does not help in this case, so we do not emit OS in WinRT.
-            return null;
+            return "Windows 10";
```
pmaytak, line 110: What is the reason for this change? Maybe should update the comment to match.

pmaytak (approved):

src/client/Microsoft.Identity.Client/Platforms/uap/UapPlatformProxy.cs
bgavrilMS: Removed the comment. We should always send the OS as ESTS makes some decisions based on it,.

PR #2529, opened 2021-04-04
Add Unity UWP dev app with README.

pmaytak:
Adds a Unity dev app that can be built for UWP and that uses MSAL to do silent and device code flows. README talks about the project plus how to build and run.


tests/devapps/Unity/README.md
```diff
@@ -0,0 +1,82 @@
+### Unity UWP dev app
+
```
jennyf19, line 2: nit: i would think this should go in the wiki, no one will find it here. there is nothing really in the [MSAL with Unity](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/hololens2) section

jennyf19:
my two cents, I'm not a fan of adding another dev app which draws more runtime, etc...maybe this could be moved to a sample or given to CxP?

tests/devapps/Unity/Assets/Scripts/MSALScript.cs
```diff
@@ -0,0 +1,118 @@
...
+        IPublicClientApplication app = PublicClientApplicationBuilder.Create(clientId)
+            .WithAuthority(authority)
+            .WithRedirectUri(redirectUrl)
+            .WithLogging((level, message, pii) => {
+                Debug.Log($"MSAL [{level}] {pii} - {message}");
+            }, LogLevel.Verbose, true)
+            .Build();
+
+        await GetToken(); // Acquires token with device code
+        await GetToken(); // Acquires token silently
```
jennyf19, line 47: this is confusing. we need to have fairly good examples of how to use the code in case customers come across it. what if there is a token in the cache? why call it twice?

tests/devapps/Unity/Assets/Scripts/MSALScript.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        async Task GetToken()
+        {
+            AuthenticationResult authResult = null;
+
+            try
+            {
+                try
+                {
```
jennyf19, line 56: this can be one try/catch w/catching of different exceptions

tests/devapps/Unity/Assets/Scripts/MSALScript.cs
```diff
@@ -0,0 +1,118 @@
...
+                    Log("MSAL acquired token with device code.");
+                }
+            }
+            catch (Exception ex)
+            {
+                Log($"MSAL Exception acquiring token: {ex}");
+            }
+
+            if (authResult is null)
+            {
```
jennyf19, line 84: I would do something similar to the device code sample -> [https://github.com/Azure-Samples/active-directory-dotnetcore-devicecodeflow-v2/blob/master/device-code-flow-console/PublicAppUsingDeviceCodeFlow.cs#L54](https://github.com/Azure-Samples/active-directory-dotnetcore-devicecodeflow-v2/blob/master/device-code-flow-console/PublicAppUsingDeviceCodeFlow.cs#L54) or do a PR to fix this up and move it to that sample w/another folder.

tests/devapps/Unity/Assets/Scripts/MSALScript.cs
```diff
@@ -0,0 +1,118 @@
...
+                Log("MSAL auth result is null.");
+            }
+            else
+            {
+                var sb = new StringBuilder();
+                sb.AppendLine($"MSAL Username={authResult.Account.Username}");
+                sb.AppendLine($"MSAL AccessToken={authResult.AccessToken}");
+                sb.AppendLine($"MSAL IdToken={authResult.IdToken}");
```
jennyf19, line 92: don't print the AT.

tests/devapps/Unity/Assets/Scripts/MSALScript.cs
```diff
@@ -0,0 +1,118 @@
...
+                try
+                {
+                    var accounts = await app.GetAccountsAsync();
+                    Log("MSAL acquiring token silently.");
+                    authResult = await app.AcquireTokenSilent(scopes, accounts.FirstOrDefault()).ExecuteAsync(CancellationToken.None);
+                    Log("MSAL acquired token silently.");
+                }
+                catch (MsalUiRequiredException)
+                {
+                    Log("MSAL acquiring token with device code.");
```
jennyf19, line 64: do we need all the extra log messages? we print out a good amount already.

pmaytak:
> my two cents, I'm not a fan of adding another dev app which draws more runtime, etc...maybe this could be moved to a sample or given to CxP? I think this could easily go into the already existing device code sample as another folder or something, but we'd have to fix up the code more as it would be more customer facing, and showing our recommended call patterns.

Yea, I think I could've been more clear - well, I wasn't intending this to be a public sample or have this in any of our automated builds (since there's also a manual build step too). I was thinking this is more for us and debugging Unity purposes. (Also wanted to have a place to document my learnings). We can discuss and, if it's worthwhile to make a sample, I could clean up the code.

bgavrilMS:
If you don't include it in the build, you can add as many dev samples you need. I have at least 6 samples that I use exclusively.

tests/devapps/Unity/Assets/Scripts/MSALScript.cs
bgavrilMS: I think this applies to public samples? It's helpful to print ATs to decode them later in dev apps.

bgavrilMS (approved):

pmaytak:
As discussed moved README to [Unity UWP dev app](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Unity-UWP-dev-app) wiki. Merging as a dev app for debugging (**not** included in the solution).

PR #2536, opened 2021-04-08
Add NuGet.Config.

pmaytak:


trwalke:
What is the reason we need this? Just to be clear

pmaytak:
> What is the reason we need this? Just to be 100% sure

@trwalke It makes sure we're using only one and correct NuGet feed. Plus I think it's the correct/better way to fix the issue we had with the NuGet restore step in builds.

trwalke (approved):

Issue #2514, opened 2021-03-29
[Feature Request] Regional authority validation


bgavrilMS:
Regional authorities cannot be validated, however the we could validate the non-regional authority. 

For example, if the input env is `hacked.com` and the regionalized env is `westus.hacked.com` 

- we cannot validate `westus.hacked.com`
- but we can validate `hacked.com` and assume `westus.hacked.com` belongs to the same threat domain

bgavrilMS:
CC @jmprieur

pmaytak:
This is included in [MSAL 4.30.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.30.0).

PR #2545, opened 2021-04-14
Regional authority validation

bgavrilMS:
Lots of files due to renames and small changes, actual changes are not large

Issue #2514


src/client/Microsoft.Identity.Client/Instance/AuthorityManager.cs
```diff
@@ -0,0 +1,75 @@
...
+                _instanceDiscoveryAndValidationExecuted = true;
+            }
+        }
+
+        public AuthorityEndpoints GetEndpoints(string loginHint)
+        {
+            if (!_instanceDiscoveryAndValidationExecuted)
+            {
+                throw new InvalidOperationException("RunInstanceDiscoveryAndValidationAsync must be called first");
+            }
```
jennyf19, line 65: won't this be an MSAL issue? nothing the developer can do, right? this should not happen, correct? It just doesn't seem actionable if I were a dev and got this message.

bgavrilMS: yes

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/AdfsAuthorityTests.cs
```diff
@@ -1,406 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-using System;
```
jennyf19, line 4: were these moved?

src/client/Microsoft.Identity.Client/Instance/AuthorityResolutionManager.cs
```diff
@@ -3,54 +3,79 @@
...
 
+        private static readonly ConcurrentHashSet<string> s_validatedEnvironments =
+            new ConcurrentHashSet<string>();
+
         private readonly IServiceBundle _serviceBundle;
+        private readonly SemaphoreSlim _semaphoreSlim = new SemaphoreSlim(1, 1);
 
-        public AuthorityEndpointResolutionManager(IServiceBundle serviceBundle, bool shouldClearCache = true)
+        public AuthorityResolutionManager(IServiceBundle serviceBundle, bool shouldClearCache = true)
         {
```
jennyf19, line 30: i don't think we need the serviceBundle, we don't use it and we can use the RC to get the service bundle anyway, so if we pass the RC instead of the service bundle in IAuthorityValidator, then we don't have to pass the RC later, as we'll have it in the contructor. I made a draft PR w/those changes, as it was easier to review this with the branch anyway. [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2546](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2546)

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/Instance/AuthorityResolutionManager.cs
```diff
@@ -3,54 +3,79 @@
...
+
+        public async Task ValidateAuthorityAsync(Authority authority, RequestContext context)
+        {
+            if (!s_validatedEnvironments.Contains(authority.AuthorityInfo.Host))
+            {
+                await _semaphoreSlim.WaitAsync().ConfigureAwait(false);
+                {
+                    try
+                    {
+                        if (!s_validatedEnvironments.Contains(authority.AuthorityInfo.CanonicalAuthority))
```
pmaytak, line 47: This should be `authority.AuthorityInfo.Host` right?

pmaytak (approved):
Tested a bit with regional test app and it works. Jenny's PR #2546 also seems appropriate.

src/client/Microsoft.Identity.Client/Instance/AuthorityManager.cs
jennyf19: maybe just do the instancediscoveryAndValidation check again here to avoid throwing an error?

PR #2546, opened 2021-04-14
remove service bundle and use RC

jennyf19:


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -253,9 +253,9 @@ private void ValidateAccountIdentifiers(ClientInfo fromServer)
             throw new MsalClientException(MsalError.UserMismatch, MsalErrorMessage.UserMismatchSaveToken);
         }
 
-        protected async Task ResolveAuthorityAsync()
+        protected Task ResolveAuthorityAsync()
         {      
-            await AuthenticationRequestParameters.AuthorityManager.RunInstanceDiscoveryAndValidationAsync().ConfigureAwait(false);            
+            return AuthenticationRequestParameters.AuthorityManager.RunInstanceDiscoveryAndValidationAsync();
```
jennyf19, line 258: this is a place were await can be avoided and we can return the task directly. when await is used, it adds a case statement to the state machine, which then has to be context switched to. this will save around 15% of cpu time in multithreaded environments. If we don't want to make this change now though, i can remove it.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -268,20 +268,19 @@ protected async Task ResolveAuthorityAsync()
...
+        protected Task<MsalTokenResponse> SendTokenRequestAsync(
             string tokenEndpoint,
             IDictionary<string, string> additionalBodyParameters,
             CancellationToken cancellationToken)
         {
             string scopes = GetOverridenScopes(AuthenticationRequestParameters.Scope).AsSingleString();
             var tokenClient = new TokenClient(AuthenticationRequestParameters);
 
-            return await tokenClient.SendTokenRequestAsync(
+            return tokenClient.SendTokenRequestAsync(
```
jennyf19, line 279: returning the task directly instead of awaiting. see reason above.

Issue #2219, opened 2020-11-19
[Bug][M] Interactive + System Browser fails if listening to localhost is forbidden - better error required


Steve0212a:
**Which Version of MSAL are you using ?**
Microsoft.Identity.Client 4.22.0

**Platform**
netcore

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App 
    * [ ] Authorization code
    * [ ] OBO
* Daemon App 
    * [ ] Service to Service calls

Other? - please describe;

**Is this a new or existing app?**
Existing app that works on most machines, but fails on others

**Repro**
No special code to reproduce as this appears to be a Windows configuration issue.  When I called AcquireTokenInteractive(), everything worked on my development machine.  However, I had an end user that it would not work for.  He could authenticate using the software on my machine without issue, so I knew it was specific to his machine.  He did have VS 2019 on his machine, so I began debugging and it would throw the exception below when AcquireTokenInteractive() was called.

	System.Net.HttpListenerException (1214): The format of the specified network name is invalid.
	   at System.Net.HttpListener.AddPrefixCore(String registeredPrefix)
	   at System.Net.HttpListener.AddAllPrefixes()
	   at System.Net.HttpListener.Start()
	   at Microsoft.Identity.Client.Platforms.Shared.DefaultOSBrowser.HttpListenerInterceptor.ListenToSingleRequestAndRespondAsync(Int32 port, Func`2 responseProducer, CancellationToken cancellationToken)
	   at Microsoft.Identity.Client.Platforms.Shared.Desktop.OsBrowser.DefaultOsBrowserWebUi.InterceptAuthorizationUriAsync(Uri authorizationUri, Uri redirectUri, CancellationToken cancellationToken)
	   at Microsoft.Identity.Client.Platforms.Shared.Desktop.OsBrowser.DefaultOsBrowserWebUi.AcquireAuthorizationAsync(Uri authorizationUri, Uri redirectUri, RequestContext requestContext, CancellationToken cancellationToken)
	   at Microsoft.Identity.Client.Internal.AuthCodeRequestComponent.FetchAuthCodeAndPkceInternalAsync(IWebUI webUi, CancellationToken cancellationToken)
[...]

PR #2547, opened 2021-04-14
More descriptive error message when MSAL is listening for OS browser.

pmaytak:
Issue #2219

We don't really know why a HttpListenerException will be thrown, so just adding a more descriptive error message.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
@@ -89,6 +89,13 @@ namespace Microsoft.Identity.Client.Platforms.Shared.DefaultOSBrowser
                 cancellationToken.ThrowIfCancellationRequested();
 
                 // if cancellation was not requested, propagate original ex
+                if (ex is HttpListenerException)
+                {
+                    throw new Exception("HttpListenerException occurred when listening for the response from the browser. " + 
+                        "One cause might be the app is unable to listen on the specified URL. " + 
+                        "One solution might be to run 'netsh http add iplisten 127.0.0.1' from the Admin command prompt.", 
+                        ex);
+                }
                 throw;
             }
             finally
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
@@ -35,6 +35,7 @@ namespace Microsoft.Identity.Client.Platforms.Shared.DefaultOSBrowser
             cancellationToken.ThrowIfCancellationRequested();
 
             HttpListener httpListener = null;
+            string urlToListenTo = string.Empty;
             try
             {
                 if(string.IsNullOrEmpty(path))
@@ -46,7 +47,7 @@ namespace Microsoft.Identity.Client.Platforms.Shared.DefaultOSBrowser
                     path = (path.StartsWith("/") ? path : "/" + path);
                 }
 
-                string urlToListenTo = "http://localhost:" + port + path;
+                urlToListenTo = "http://localhost:" + port + path;
 
                 if (!urlToListenTo.EndsWith("/"))
                 {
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
```diff
@@ -89,6 +89,13 @@ public HttpListenerInterceptor(ICoreLogger logger)
                 cancellationToken.ThrowIfCancellationRequested();
 
                 // if cancellation was not requested, propagate original ex
+                if (ex is HttpListenerException)
+                {
+                    throw new Exception("HttpListenerException occurred when listening for the response from the browser. " +
```
jennyf19, line 94: >Exception [](start = 30, length = 9)

nit: should this be an `MsalClientException`?

src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
```diff
@@ -89,6 +89,13 @@ public HttpListenerInterceptor(ICoreLogger logger)
                 cancellationToken.ThrowIfCancellationRequested();
 
                 // if cancellation was not requested, propagate original ex
```
jennyf19, line 91: >// if cancellation was not requested, propagate original ex [](start = 16, length = 59)

i would move this comment to line 99 before the throw.

src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
pmaytak, line 94: Alternatively, instead of throwing, I can just log error here.

bgavrilMS: Yes, `MsalClientException`. There should even be an analyzer that prevents you from raising simple Exception, as it is too "general".

bgavrilMS: No, let's rethrow an MsalClientException and set this exception as inner exception.

src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
```diff
@@ -89,6 +89,13 @@ public HttpListenerInterceptor(ICoreLogger logger)
...
+                if (ex is HttpListenerException)
+                {
+                    throw new Exception("HttpListenerException occurred when listening for the response from the browser. " + 
+                        "One cause might be the app is unable to listen on the specified URL. " + 
+                        "One solution might be to run 'netsh http add iplisten 127.0.0.1' from the Admin command prompt.",
```
bgavrilMS, line 96: I would word it more like 

`An HttpListenerException occurred while listening on http:\\localhost:{port} for the system browser to complete the login. Possible cause and mitigation: .... "

bgavrilMS (changes requested):
Don't raise plain old Exception, raise an MsalClientException and reword it a bit.

src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
pmaytak: Done.

src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
pmaytak: Done.

src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
pmaytak: Done.

src/client/Microsoft.Identity.Client/Platforms/Features/DefaultOSBrowser/HttpListenerInterceptor.cs
pmaytak: Done.

bgavrilMS (approved):

Issue #2544, opened 2021-04-13
[Bug] AcquireTokenForClient should use a partitioned token cache by default and not log errors


rabenshm:
**Which Version of MSAL are you using ?**
MSAL 4.29

**Platform**
net472

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App 
    * [ ] Authorization code
    * [ ] OBO
* Daemon App 
    * [x] Service to Service calls

**Is this a new or existing app?**
The app is in production, and I have upgraded to a new version of MSAL

**Repro**

Use the following approach for token caching: 

- maintain a dictionary of [client_id] -> [ConfidentialClientApp]
- not a multi-tenant app, there is always 1 tenant 
- do not serialize the cache in any way
- for each request, try to reuse an existing CCA. If it does not exit, create one.

**Expected behavior**
Token and no error log messages

**Actual behavior**
Token. But we get the following **error** log message:

2021-04-12 15:34:11Z KD.MSAL    (False) MSAL 4.29.0.0 MSAL.Desktop Microsoft Windows NT 10.0.19042.0 [04/12/2021 15:34:11 - ] The default token cache provided by MSAL is not designed to be performant when used in confidential client applications. Please use token cache serialization. See https://aka.ms/msal-net-cca-token-cache-serialization.

**Impact**
Kusto offer this as part of their SDK, and pipe the MSAL errors into their logging pipeline, which customers monitor. These errors will create noise for customers monitoring the logs.

bgavrilMS:
CC @jmprieur

jmprieur:
@bgavrilMS 
I'm happy to remove the log error, but only if we provide an in-memory partitionned cache by default (which we could register with the normal pattern). Maybe something like the following? (and we could move / copy MsalAbstractTokenCacheProvider to MSAL.NET)

```CSharp
   /// <summary>
    /// Msal token cache serialize based on a concurrent dictionary.
    /// </summary>
    public class MsalDictionaryBasedTokenCacheSerializer : MsalAbstractTokenCacheProvider
    {
        private ConcurrentDictionary<string, byte[]> partitionnedCache = new ConcurrentDictionary<string, byte[]>();

        /// <inheritdoc/>
        protected override Task<byte[]> ReadCacheBytesAsync(string cacheKey)
        {
[...]

PR #2550, opened 2021-04-15
MSAL to use a partioned cache for S2S (client_credential cache)

bgavrilMS:
Benchmark tests: 

|                     Method | TokenCacheSize |         Mean |      Error |     StdDev |
|--------------------------- |--------------- |-------------:|-----------:|-----------:|
| 'Different tenants - O(1)' |           1000 |     60.89 us |   0.660 us |   0.551 us |
|  'Different scopes - O(n)' |           1000 | 44,379.64 us | 415.110 us | 367.984 us |

And a note on the correct calling pattern for multi-tenant apps (@jmprieur): 

```csharp
// this object can be a singleton
var cca = ConfidentialClientApplicationBuilder()
               .WithAuthority("https://login.microsoft.com/common") // we don't know the tenant at this point
              .Build();

// later, when we know the tenant
await cca.AcquireTokenForClient().WithAuthority("https://login.microsoft.com/TENANT_ID").ExecuteAsync();

```

Issue #2544


src/client/Microsoft.Identity.Client/ITokenCacheInternal.cs
```diff
@@ -71,8 +71,15 @@ internal interface ITokenCacheInternal : ITokenCache, ITokenCacheSerializer
         /// </summary>
         bool HasTokensNoLocks();
 
+
         bool IsTokenCacheSerialized();
 
+        /// <summary>
+        /// MSAL adds serialziation for UWP and also for ConfidentiaClient app token cache.
```
jennyf19, line 78:
```suggestion
        /// MSAL adds serialization for UWP and also for ConfidentialClient app token cache. 
```

src/client/Microsoft.Identity.Client/Cache/CacheImpl/InMemoryPartitionedCacheSerializer.cs
```diff
@@ -0,0 +1,53 @@
...
+            }
+
+            _logger.Verbose($"[InMemoryPartitionedTokenCache] ReadCacheBytes did not find cacheKey {cacheKey}");
+
+            return null;
+        }
+
+        protected override void RemoveKey(string cacheKey)
+        {
+            bool removed = CachePartition.TryRemove(cacheKey, out _);
```
jennyf19, line 39: is it valuable to include the cache size in all the logs messages, if it's available?

pmaytak (commented):
LGTM so far, will review after tests added.

src/client/Microsoft.Identity.Client/Cache/CacheImpl/InMemoryPartitionedCacheSerializer.cs
bgavrilMS: Added it on Write only, as `Count` is expensive on a ConcurrentDictionary

Issue #2554, opened 2021-04-16
[Bug] DatabaseUtils: java.lang.IllegalArgumentException: bad base-64


timdriever:
**Logs and Network traces**
Via the Authenticator app I sent feedback with logging, incident id: RH9JA8G3.

```
04-16 16:15:13.868 11565 11790 E DatabaseUtils: Writing exception to parcel
04-16 16:15:13.868 11565 11790 E DatabaseUtils: java.lang.IllegalArgumentException: bad base-64
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.util.Base64.decode(Base64.java:162)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.util.Base64.decode(Base64.java:137)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.util.Base64.decode(Base64.java:119)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at com.microsoft.identity.client.BrokerContentProvider.query(BrokerContentProvider.java:126)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.content.ContentProvider.query(ContentProvider.java:1235)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.content.ContentProvider.query(ContentProvider.java:1328)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.content.ContentProvider$Transport.query(ContentProvider.java:267)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:104)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.os.Binder.execTransactInternal(Binder.java:1056)
04-16 16:15:13.868 11565 11790 E DatabaseUtils: 	at android.os.Binder.execTransact(Binder.java:1029)
04-16 16:15:13.869 10514 11474 I mono-stdout: MSAL: Error (False) MSAL 4.29.0.0 MSAL.Xamarin.Android 29 [04/16/2021 14:15:13] [Android broker] Failed to get accounts from the broker.  Sender=AzureAdAuthTokenProvider
 ```

**Which Version of MSAL are you using ?**
Msal 4.29.0

**Platform**
Xamarin.Android

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
[...]

Issue #2467, opened 2021-03-10
[Bug] Bad magic number exception when signing in


timdriever:
**Logs and Network traces**
```
Info	1-3-2021 10:35:06 +01:00	AzureAdAuthService	Attempting to sign in silently
Debug	1-3-2021 10:35:06 +01:00	AzureAdAuthService	Starting RequestTokenSilentWithTimeoutAsync
Info	1-3-2021 10:35:06 +01:00	AzureAdAuthTokenProvider	RequestTokenSilentAsync with scopes: api://****.default
Info	1-3-2021 10:35:06 +01:00	AzureAdAuthTokenProvider	RequestTokenSilentAsync with account: ***@***.nl
Error	1-3-2021 10:35:08 +01:00	AzureAdAuthTokenProvider	MsalException: MSAL.Xamarin.Android.4.27.0.0.MsalClientException: 
	ErrorCode: android_broker_operation_failed
Microsoft.Identity.Client.MsalClientException: Bad magic number for Bundle: 0x139193 ---> Java.Lang.IllegalStateException: Bad magic number for Bundle: 0x139193
  at Java.Interop.JniEnvironment+InstanceMethods.CallNonvirtualObjectMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniObjectReference type, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args) [0x0008e] in <8b3b636835d84984ba4604c1f57b1983>:0 
  at Java.Interop.JniPeerMembers+JniInstanceMethods.InvokeNonvirtualObjectMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters) [0x0001f] in <8b3b636835d84984ba4604c1f57b1983>:0 
  at Android.Content.ContentResolver.Query (Android.Net.Uri uri, System.String[] projection, System.String selection, System.String[] selectionArgs, System.String sortOrder) [0x000aa] in <44e54a86dea24313a2bdb807df77c27a>:0 
  at Microsoft.Identity.Client.Platforms.Android.Broker.AndroidContentProviderBroker+<>c__DisplayClass23_0.<PerformContentResolverOperationAsync>b__0 () [0x00041] in <0e26de906a8248dd9fddc4acb51c8aca>:0 
[...]

PR #2578, opened 2021-04-26
Fix for querying Android content provider

pmaytak:
Fix for querying Android content provider plus nit renamings. Fix is to do normal Base64 encoding instead of Base64 URL encoding. Repro is to provide the bundle (basically just the client Id and redirect URI) that would result in Base64 and Base64Url encoded values being different. Android then throws an exception trying to Base64 decode a Base64Url encoded value.

Issue #2554 
Also (unverified) possible fix for #2467 as Android seemed to throw bad magic exception if operation parameters passed to query method where malformed.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
 using Android.App;
 using Android.Content;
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
```diff
@@ -345,9 +344,9 @@ private string GetContentProviderURIForOperation(string operation)
             return BrokerConstants.MsAuthenticatorContentProviderUri + "/" + operation;
         }
 
-        private string SearializeBundleToJSON(Bundle bundle)
+        private string SerializeBundle(Bundle bundle)
         {
-            return Base64UrlHelpers.Encode(Marshall(bundle));
+            return Convert.ToBase64String(Marshall(bundle));
```
pmaytak, line 349: This is the fix. Everything else is renamings.

bgavrilMS: Very cool. How did you figure it out?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidContentProviderBroker.cs
pmaytak: Basically looked at all the code where we encode the values that we pass in the content resolver query. Then repro'd by faking the broker request with the client Id and redirect URI of the app that was breaking.

Issue #2525, opened 2021-04-01
[Feature Request] MSAL libraries provide routing information to CCS


henrik-me:
[Feature 1333541: [MSAL.NET] MSAL libraries provide routing information to CCS](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1333541)

[CCS MSAL routing hint one pager - info available in epic](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1331220)

Stories: 

- [x] ROPC and IWA - send username [M]
- [x] AcquireTokenSilent - send client_info [S]
- [x] AcquireTokenInteractive - send login_hint [S]
- [x] Expose a `WithExtraHttpHeaders(IDictionary<string, string>)` that allows confidential clients to send client_info for `AcquireTokenByAuthorizationCode` and `OBO` [M]
- [x] [MS Id web](https://github.com/AzureAD/microsoft-identity-web/issues/1146) work [S] to use `WithExtraHttpHeaders` - included in 1.14 release
- [x] Confirm that headers sent are in the right format etc. 
- [x] Make sure to cover B2C and ADFS in E2E testing.

bgavrilMS:
The gist of it is to add 2 headers: 

`X-AnchorMailbox:”Oid:<userObjectId>@<userTenantID>” or  
X-AnchorMailbox: “UPN: <userUpn>” `

# We need to understand

- [ ] What auth flows should this cover?
- [ ] Token endpoint and authorize endpoint ? How can CCS cache the authorization code, as it is protected by PKCE and cannot be redeemed multiple times?
- [ ] Client credentials flow (daemon scenario) ? What should we send? We don't have OID, only tenant ID if the user didn't use `common`
- [ ] Does this need to be done by M.I.W in the first part of the auth flow which is handled by ASP.NET Core? I.e. when getting the auth code.
- [ ] How can we test this feature?

jennyf19:
- going to the /authorize endpoint for this doesn't make sense. We can't store the auth code, not only because of PKCE, but also because it has state, so holding for a period of time will make the auth code unusable. Also, we don't have that information about the signed in user at that point.. so only makes sense on the /token endpoint. 
- also, privacy concerns w/sending the UPN on the header? 
- Does B2C support this or be impacted by the extra header info?
- I would think the spec writers would need to come up w/the s2s scenarios, as there is no user, as you mention.

bgavrilMS:
[...]

PR #2590, opened 2021-04-28
Adding CCS Routing information to MSAL.

trwalke:
Adding CCS Routing information to MSAL
Adding WithExtraHttpHeaders to Auth code and OBO flow for CCA

Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2525


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
```diff
@@ -101,5 +101,16 @@ public AcquireTokenByAuthorizationCodeParameterBuilder WithPkceCodeVerifier(stri
             Parameters.PkceCodeVerifier = pkceCodeVerifier;
             return this;
         }
+
+        /// <summary>
+        ///
```
trwalke, line 106: Need to add info

trwalke:
Waiting for E2E test with ESTS and Identity Web

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
bgavrilMS: @jmprieur - thoughts on adding this method on all AcquireToken* methods?

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -161,6 +161,54 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
...
+            }
+        }
+
+        private void AddExtraHttpHeaders()
+        {
+            if (_requestParams.ExtraHttpHeaders != null)
+            {
+                foreach (KeyValuePair<string, string> pair in _requestParams.ExtraHttpHeaders)
+                {
+                    _oAuth2Client.AddHeader(pair.Key, pair.Value);
```
bgavrilMS, line 203: What happens if the user adds a header which already exists, for example x-client-sku ? IMO the user value should overwrite the lib value. We should have a test for this.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -161,6 +161,54 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
...
             }
+
+            AddCCSRoutingInformation(additionalBodyParameters);
+
+            AddExtraHttpHeaders();
+        }
+
+        private void AddCCSRoutingInformation(IDictionary<string, string> additionalBodyParameters)
+        {
+            if (_requestParams.Account != null && _requestParams.Account.HomeAccountId != null)
```
bgavrilMS, line 172: nit: 

```csharp
if (_requestParams?.Account?.HomeAccountId != null)

```

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -161,6 +161,54 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
             {
                 _oAuth2Client.AddHeader(PKeyAuthConstants.DeviceAuthHeaderName, PKeyAuthConstants.DeviceAuthHeaderValue);
             }
+
+            AddCCSRoutingInformation(additionalBodyParameters);
+
+            AddExtraHttpHeaders();
+        }
+
+        private void AddCCSRoutingInformation(IDictionary<string, string> additionalBodyParameters)
```
bgavrilMS, line 170: Needs some comments as we'll forget what CCS even is.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -229,23 +232,60 @@ private async Task<AuthenticationResult> RunTestForUserAsync(LabResponse labResp
...
+                .ConfigureAwait(false);
+
             await MsalAssert.AssertSingleAccountAsync(labResponse, pca, result).ConfigureAwait(false);
             Assert.IsFalse(userCacheAccess.LastAfterAccessNotificationArgs.IsApplicationCache);
+            AssertCCSRoutingInformationIsSent(factory, labResponse);
 
             return result;
         }
 
+        private void AssertCCSRoutingInformationIsSent(HttpSnifferClientFactory factory, LabResponse labResponse)
```
bgavrilMS, line 262: This is fine, but we should also focus on unit testing. In particular, we don't have integrated tests for IWA.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -161,6 +161,54 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
             {
                 _oAuth2Client.AddHeader(PKeyAuthConstants.DeviceAuthHeaderName, PKeyAuthConstants.DeviceAuthHeaderValue);
             }
+
+            AddCCSRoutingInformation(additionalBodyParameters);
```
bgavrilMS, line 165: I'm in 2 minds about having this functionality here. On one side, the code is pretty compact. On the other, it violates the single reponsability principle, i.e. the the tokenclient class neesd to format requests to the token endpoint. It shouldn't need to know how to extract CCS information. The request classes themselves are a better place for this kind of knowledge, smth like; 


```csharp
//in RequestBase.cs define a GetCCSHeader() which every request that can provide this must implement
protected string GetCCSHeader() { return null; }

// modify SendTokenRequestAsync to inject this extra header into the tokenClient it creates.

```

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -161,6 +161,54 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
...
+
+            AddExtraHttpHeaders();
```
bgavrilMS, line 167: This is fine.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
bgavrilMS: It's also a slightly more performant solution, as `client_credentials` request will simply return `null` instead of going though all those if conditions and string comparisons.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
trwalke: The code in the OAuth client will overwrite the previous value if the key for the header already exists.
see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e6c3d8cb66232dd69a7f0229b24ffe171a8edf8c/src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs#L62

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
trwalke: Since all of the request classes simply call the base SendTokenRequestAsync and the base class has all of the data necessary to the header most of the time I only implemented the method in the classes that could not provide any data or if the data was not available to the requestBase

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
trwalke: Do you know of a good way to validate headers in the Unit tests? The existing techniques that I used seem not to work with the mocks. @bgavrilMS

bgavrilMS: There are 2 ways of doing this. The first way is to use the `ExpectedRequestHeaders` from the `MockHttpMessageHandler` object. 

The second way is to examine the HttpRequestMessage which the `MockHttpMessageHandler` exposes through the `ActualRequestMessage` property. It will have the actual headers sent.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
bgavrilMS: @trwalke - I suggest we add this to all AcquireToken* methods, just like we do for `WithExtraQueryParameters`. Also, I would add it to the namespace `microsoft.identity.client.advanced`, to not polute the existing API, which is already too complicated.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
```diff
@@ -101,5 +101,16 @@ public AcquireTokenByAuthorizationCodeParameterBuilder WithPkceCodeVerifier(stri
             Parameters.PkceCodeVerifier = pkceCodeVerifier;
             return this;
         }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="extraHttpHeaders"></param>
+        /// <returns></returns>
+        public AcquireTokenByAuthorizationCodeParameterBuilder WithExtraHttpHeaders(IDictionary<string, string> extraHttpHeaders)
```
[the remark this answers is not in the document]
jmprieur: @bgavrilMS : I think that was already what I proposed: have it on all the AcquireToken* methods ^^^
So yes, I agree (and ok for a different namespace)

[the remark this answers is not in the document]
trwalke: @jmprieur @bgavrilMS what is the best way to add this to a different namespace? a partial class of the base acquireTokenParameterBuilder class?

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenParameterBuilderExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// </summary>
+    public static class AcquireTokenParameterBuilderExtensions
+    {
+        /// <summary>
+        /// Adds additional Http Headers to the token request.
+        /// </summary>
+        /// <param name="builder">Parameter builder for a acquiring tokens.</param>
+        /// <param name="extraHttpHeaders">additional Http Headers to add to the token request.</param>
+        /// <returns></returns>
+        public static AcquireTokenInteractiveParameterBuilder WithExtraHttpHeaders(
```
trwalke, line 18: I tried to use the base parameter builder classes so that I only needed one method here but they are generics and caused a lot of strange build errors. I ended up adding an overload for each of the child parameter builder types.
@bgavrilMS

jmprieur: that's not the right solution, @trwalke.

bgavrilMS: Yes, the use generics here is very tricky. In hindsight, we should have opted for a simpler construct that does not rely on abstract generic classes with recursive generics. It also has the downside that it cannot be unit tested :(


Anyway, here's the magic formula, which I happen to stumble upon

```csharp
 /// <summary>
    /// Extensions for <see cref="AbstractAcquireTokenParameterBuilder{T}"/>
    /// </summary>
    public static class AbstractAcquireTokenParameterBuilderExtension
    {
        /// <summary>
        /// 
        /// </summary>
        /// <param name="builder"></param>
        /// <param name="extraHttpHeaders"></param>
        public static AbstractAcquireTokenParameterBuilder<T> WithExtraHttpHeader<T>(this AbstractAcquireTokenParameterBuilder<T> builder, IDictionary<string, string> extraHttpHeaders)
            where T : AbstractAcquireTokenParameterBuilder<T>
        {
            builder.CommonParameters.ExtraHttpHeaders = extraHttpHeaders;
            return builder;
        }
    }
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -278,13 +278,57 @@ protected Task ResolveAuthorityAsync()
...
+            {
+                if (!String.IsNullOrEmpty(AuthenticationRequestParameters.Account.HomeAccountId.Identifier))
+                {
+                    var userObjectId = AuthenticationRequestParameters.Account.HomeAccountId.ObjectId;
+                    var userTenantID = AuthenticationRequestParameters.Account.HomeAccountId.TenantId;
+                    string OidCCSHeader = CoreHelpers.GetCCSClientInfoheader(userObjectId, userTenantID);
+
+                    return new Tuple<string, string>(Constants.OidCCSHeader, OidCCSHeader);
+                }
+                else if (!String.IsNullOrEmpty(AuthenticationRequestParameters.Account.Username))
```
bgavrilMS, line 309: Please simplify these if / else clauses ... for example since you return at line 307, you don't need the `else`.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -278,13 +278,57 @@ protected Task ResolveAuthorityAsync()
...
                 additionalBodyParameters,
                 scopes,
                 tokenEndpoint,
                 cancellationToken);
         }
 
+        //The CCS header is used by the CCS service to help route requests to resources in Azure during requests to speed up authentication.
+        //It consists of either the ObjectId.TenantId or the upn of the account signign in.
+        //See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2525
+        protected virtual Tuple<string, string> GetCCSHeader(IDictionary<string, string> additionalBodyParameters)
```
bgavrilMS, line 297: nit: use `KeyValuePair<string, string>` instead of the tuple. The KeyValuePair is the natural structure that goes into a dictionary. 

nit: tuples are very cool, but sadly we need to use C# 8 to use them. We can't because .NET 4.5 support, so our use of Tuples is not great, t.Item1, t.Item2 etc. Hence KeyValuePair struct here might be a better fit.

bgavrilMS (approved):
Minor comment

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -161,6 +161,24 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
...
+        }
+
+        private void AddExtraHttpHeaders()
+        {
+            if (_requestParams.ExtraHttpHeaders != null)
+            {
+                foreach (KeyValuePair<string, string> pair in _requestParams.ExtraHttpHeaders)
+                {
+                    _oAuth2Client.AddHeader(pair.Key, pair.Value);
```
pmaytak, line 174: Can `Value` be null? Do we need a check? Maybe a check in the `AddHeader` method?

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
```diff
@@ -237,5 +237,15 @@ private static void AddKeyValueString(StringBuilder messageBuilder, string key,
             messageBuilder.AppendFormat(CultureInfo.InvariantCulture, "{0}{1}=", delimiter, key);
             messageBuilder.Append(value);
         }
+
+        internal static string GetCCSClientInfoheader(string userObjectId, string userTenantID)
```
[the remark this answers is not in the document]
trwalke: see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2590#discussion_r643680488

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
```diff
@@ -237,5 +237,15 @@ private static void AddKeyValueString(StringBuilder messageBuilder, string key,
...
+        {
+            return (string.IsNullOrEmpty(userObjectId) || string.IsNullOrEmpty(userTenantID)) ? string.Empty : $@"oid:{userObjectId}@{userTenantID}";
+        }
+
+        internal static string GetCCSUpnHeader(string upn)
+        {
+            return string.IsNullOrEmpty(upn)? string.Empty : $@"upn:{upn}";
```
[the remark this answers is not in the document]
trwalke: Good catch

trwalke:
I am running into this error and I cannot repro locally. Any idea? @bgavrilMS 
https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=780019&view=ms.vss-test-web.build-test-results-tab&runId=1578500&resultId=100045&paneView=debug

bgavrilMS:
@trwalke - fixed the test

Issue #2573, opened 2021-04-22
[Bug][L] ConfidentialClient does not support ECDsa client application certificates.


fuocor:
**Which Version of MSAL are you using ?**
4.30

**Platform**
APSNETCORE 5.0

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web App 
    * [x] Authorization code
    * [ ] OBO
* Daemon App 
    * [ ] Service to Service calls

Other? - please describe;

**Is this a new or existing app?**
existing
**Repro**
1. Register an AzureAD confidential client application with a certificate using ECDsa
2. Create an instance on the ConfidentialClient
3. Try and acquire an OBO token

```csharp
var onBehalfOfClientApplication = await this.oauthTokenProvider.BindAsync(accountId, this.TenantId).ConfigureAwait(false);

var oboResult = await onBehalfOfClientApplication
    .AcquireTokenOnBehalfOf(oauthScopes, userAssertion)
    .ExecuteAsync().ConfigureAwait(false);
```

**Expected behavior**
A token.

**Actual behavior**
```
      System.NotSupportedException: The certificate key algorithm is not supported.
         at System.Security.Cryptography.X509Certificates.PublicKey.get_Key()
         at Microsoft.Identity.Client.Platforms.netcore.NetCoreCryptographyManager.SignWithCertificate(String message, X509Certificate2 certificate)
         at Microsoft.Identity.Client.Internal.ClientCredentialWrapper.Sign(ICryptographyManager cryptographyManager, String message)
         at Microsoft.Identity.Client.Internal.JsonWebToken.Sign(ClientCredentialWrapper credential, Boolean sendCertificate)
         at Microsoft.Identity.Client.Internal.Requests.ClientCredentialHelper.CreateClientCredentialBodyParameters(ICoreLogger logger, ICryptographyManager cryptographyManager, ClientCredentialWrapper clientCredential, String clientId, AuthorityEndpoints endpoints, Boolean sendX5C)
         at Microsoft.Identity.Client.OAuth2.TokenClient.AddBodyParamsAndHeaders(IDictionary`2 additionalBodyParameters, String scopes)
[...]

PR #2601, opened 2021-05-04
Add support for ECDsa and DSA

neha-bhargava:
#2573 Adding support for ECDSA and DSA certs


bgavrilMS:
Should we close this @neha-bhargava ?

PR #2602, opened 2021-05-04
Adding the Azure AD Kerberos Helper

yoko-ms:
Adding the Azure AD Kerberos Feature into MSAL:

* [Design Document](https://nam06.safelinks.protection.outlook.com/ap/p-59584e83/?url=https%3A%2F%2Fmicrosoft.sharepoint-df.com%2F%3Ap%3A%2Ft%2FAzureADKerberos%2FEUKpEJAOFqtDg9lJfOyZV_sBvrw6Hz5StFcOcjhJTYXZEA%3Fe%3DnwWkso&data=04%7C01%7CKo.Young%40microsoft.com%7C426a5ad400484d1cb93b08d8dde661ed%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637503328854636146%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=uIxuDnQ0F36SzYTKEdeepAIfPnsSemb3qpaOZRp%2BLRI%3D&reserved=0)
* [AAD Kerberos for MSAL WIKI](https://identitydivision.visualstudio.com/IdentityWiki/_wiki/wikis/IdentityWiki.wiki/20601/AAD-Kerberos-for-MSAL)

Many customers want to migrate to modern authentication protocols but are stuck because their existing apps rely on Kerberos. The AAD Kerberos Feature enables the user to create an application with MSAL getting Kerberos Ticket from Azure AD using modern user authentication method and interface with existing Kerberos-based authentication services.

**Usage Scenario 1**

- Client applications want to access any traditional Kerberos Service directly.
- Client application gets Kerberos Ticket using Id Token.
- Client application read out Kerberos Ticket from received Id Token using MSAL Kerberos Module.
- Client application can cache the Kerberos Ticket in the current user’s Ticket Cache or use it directory to access any traditional Kerberos Service.

**Usage Scenario 2**
Client application interfaces with a Proxy service supporting Kerberos Service based on web token as the Kerberos Ticket container.

- Proxy service is a general modern web service supporting all types of web token and can handle it without MSAL library.
- Client application gets Kerberos Ticket using an access token.
- Client application post received access token to the Proxy service as a web token (full access token).
- MSAL Kerberos Module will not provide any processing for received access token.

**Supported Features**

1. Kerberos Ticket Claim generation
2. Read out Kerberos Ticket from received Id Token
3. Save the received Kerberos Ticket into current user's Windows Ticket Cache.
4. Find a cached Kerberos Ticket matched with given service principal name from current user's Windows Ticket Cache.


src/client/Microsoft.Identity.Client/Internal/ClaimsHelper.cs
```diff
@@ -28,7 +28,7 @@ internal static class ClaimsHelper
             return claims;
         }
 
-        private static JObject MergeClaimsIntoCapabilityJson(string claims, JObject capabilitiesJson)
+        internal static JObject MergeClaimsIntoCapabilityJson(string claims, JObject capabilitiesJson)
```
yoko-ms, line 31: Changed as internal to use from the KerberosSupplementalTicketManager.AddKerberosTicketClaim() to merge the Kerberos Supplemental Ticket claims into the token request claims body parameter.

build/platform_and_feature_flags.props
```diff
@@ -26,4 +26,9 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <DefineConstants>$(DefineConstants);NETSTANDARD;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_BROKER;SUPPORTS_CUSTOM_CACHE;SUPPORTS_OS_SYSTEM_BROWSER;SUPPORTS_WEBVIEW2</DefineConstants>
   </PropertyGroup>
+
+  <PropertyGroup Condition="$([MSBuild]::IsOsPlatform('Windows'))">
```
yoko-ms, line 30: NOTE: The AAD Kerberos Feature is targeting only the Windows platform now. We added the SUPPORT_KERBEROS flag to count on this limitation.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        /// <summary>
+        /// Save current Kerberos Ticket to current user's Ticket Cache.
+        /// </summary>
+        /// <param name="ticket">Kerberos ticket object to save.</param>
+        /// <remarks>Can throws <see cref="ArgumentException"/> when given ticket parameter is not a valid Kerberos Supplemental Ticket.
+        /// Can throws <see cref="Win32Exception"/> if error occurs while saving ticket information into Ticket Cache.
+        /// </remarks>
+        public static void SaveToCache(KerberosSupplementalTicket ticket)
```
jmprieur, line 61: We might want to be more precise than "Cache", as this is a different cache from the token cache for instance. What about `SaveToWindowsTicketCache` ?

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// <summary>
+         /// Save current Kerberos Ticket to current user's Ticket Cache.
+         /// </summary>
+         /// <param name="ticket">Kerberos ticket object to save.</param>
+         /// <param name="logonId">The Logon Id of the user owning the ticket cache.
+         /// The default of 0 represents the currently logged on user.</param>
+         /// <remarks>Can throws <see cref="ArgumentException"/> when given ticket parameter is not a valid Kerberos Supplemental Ticket.
+         /// Can throws <see cref="Win32Exception"/> if error occurs while saving ticket information into Ticket Cache.
+         /// </remarks>
+        public static void SaveToCache(KerberosSupplementalTicket ticket, long logonId)
```
jmprieur, line 75: Same thing: `SaveToWindowsTicketCache` ?

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// Checks a Kerberos Service Ticket associated with given service principal name exists
+        /// in current user's Ticket Cache.
+        /// </summary>
+        /// <param name="servicePrincipalName">Service principal name to find associated Kerberos Ticket.</param>
+        /// <returns>True if Kerberos Ticket exists. False, otherwise.</returns>
+        /// <returns>True if Ticket save to Ticket Cache successfully. False, otherwise.</returns>
+        /// <remarks>
+        /// Can throws <see cref="Win32Exception"/> if error occurs while searching ticket information from Ticket Cache.
+        /// </remarks>
+        public static bool IsTKerberosTicketExistsInCache(string servicePrincipalName)
```
jmprieur, line 101: Is looks funny. I guess "Does" would be better, but I understand the conversion "Is" for methods returning bools ...

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// <summary>
+        /// Reads a Kerberos Service Ticket associated with given service principal name from
+        /// current user's Ticket Cache.
+        /// </summary>
+        /// <param name="servicePrincipalName">Service principal name to find associated Kerberos Ticket.</param>
+        /// <returns>Byte stream of searched Kerberos Ticket information if exists. Null, otherwise.</returns>
+        /// <remarks>
+        /// Can throws <see cref="Win32Exception"/> if error occurs while searching ticket information from Ticket Cache.
+        /// </remarks>
+        public static byte[] GetKerberosTicketFromCache(string servicePrincipalName)
```
jmprieur, line 132: `GetKerberosTicketFromWindowsTicketCache` ?

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// current user's Ticket Cache.
+        /// </summary>
+        /// <param name="servicePrincipalName">Service principal name to find associated Kerberos Ticket.</param>
+        /// <param name="logonId">The Logon Id of the user owning the ticket cache.
+        /// The default of 0 represents the currently logged on user.</param>
+        /// <returns>Byte stream of searched Kerberos Ticket information if exists. Null, otherwise.</returns>
+        /// <remarks>
+        /// Can throws <see cref="Win32Exception"/> if error occurs while searching ticket information from Ticket Cache.
+        /// </remarks>
+        public static byte[] GetKerberosTicketFromCache(string servicePrincipalName, long logonId)
```
jmprieur, line 148: `GetKerberosTicketFromWindowsTicketCache` ?

jmprieur (commented):
Thanks @yoko-ms 
I've only looked at the public API and proposed a few suggestions for renamings
Is it supposed to work on UWP? Desktop apps? .NET Core on Windows?

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Renamed.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Renamed.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: This interface is not required - removed.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Renamed.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Renamed.

yoko-ms:
> Thanks @yoko-ms
> I've only looked at the public API and proposed a few suggestions for renamings
> Is it supposed to work on UWP? Desktop apps? .NET Core on Windows?

It should support all platforms on Windows. If there's anything missed, please let me know.

src/client/Microsoft.Identity.Client/Kerberos/KerberosIdTokenParser.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+
+        /// <summary>
+        /// RAW token string from Azure AD is base64 url encoded.
+        /// Decodes the given base64 url encoded raw token string.
+        /// </summary>
+        /// <param name="arg">Base64 url encoded token string.</param>
+        /// <returns>Returns decoded 8-bit unsigned byte array.</returns>
+        private static byte[] DecodeBytes(string arg)
```
trwalke, line 82: We have this functionality in a utility class "Base64UrlHelpers.DecodeToBytes()". Is there a reason to recreate these methods here instead of using the one we already have?

yoko-ms: Useful information. I didn't have context for the Base64UrlHelpers.DecodeToBytes() method. Replaced to use the existing Base64UrlHelpers class.

trwalke: Just saw the update. looks good @yoko-ms.

build/platform_and_feature_flags.props
trwalke: Since this is ultimatly a windows target, maybe we can simply call it "WINDOWS". Similar to what we have for "iOS", "MAC" and "ANDROID". Then we can use it for other features that may come in the future.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,220 @@
...
+            }
+
+            if (ticketContainer == KerberosTicketContainer.IdToken)
+            {
+                return string.Format(
+                    CultureInfo.InvariantCulture,
+                    IdTokenAsRepTemplate,
+                    servicePrincipalName);
+            }
+            else
```
trwalke, line 211: Since you are returning in the if statement above, you probably dont need the else.

src/client/Microsoft.Identity.Client/Kerberos/Win32/Credential.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// such as a password, or a Kerberos protocol ticket.
+    /// </summary>
+    public abstract class Credential
+    {
+        internal abstract CredentialHandle Structify();
+
+        /// <summary>
+        /// Create a new <see cref="Credential"/> object.
+        /// </summary>
+        /// <returns>Creted new <see cref="Credential"/> object.</returns>
```
trwalke, line 17:
```suggestion
        /// <returns>Newly created <see cref="Credential"/> object.</returns>
```

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,220 @@
...
+
+            // decodes the second section containing the Kerberos Ticket claim if exists.
+            byte[] payloadBytes = Base64UrlHelpers.DecodeToBytes(sections[1]);
+            string payload = Encoding.UTF8.GetString(payloadBytes);
+            if (string.IsNullOrEmpty(payload))
+            {
+                return null;
+            }
+
+            // parse the JSON data anf find the included Kerberos Ticket claim.
```
trwalke, line 56:
```suggestion
            // parse the JSON data and find the included Kerberos Ticket claim.

```

src/client/Microsoft.Identity.Client/Kerberos/Win32/InitContextFlag.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Identity.Client.Kerberos.Win32
+{
+    /// <summary>
+    /// Bit flags that indicate requests for the context for InitializeSecurityContext API call.
+    /// 
+    /// https://docs.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-initializesecuritycontexta
```
trwalke, line 11: is the extra space needed?

tests/devapps/KerberosConsole/AADKerberosLogger.cs
```diff
@@ -0,0 +1,201 @@
...
+            {
+                if (string.IsNullOrEmpty(_logFileName))
+                {
+                    do
+                    {
+                        _lastFileIndex++;
+
+                        _logFileName = String.Format(
+                            CultureInfo.InvariantCulture,
+                            @"C:\Users\yoko\Documents\AzureAD-MSAL-{0}-{1}.log",
```
trwalke, line 41: Should probably point to the current windows user

tests/Microsoft.Identity.Test.Unit/Kerberos/KerberosSupplementalTicketManagerTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        [TestMethod]
+        public void FromIdToken_WithKerberosTicket()
+        {
+            KerberosSupplementalTicket ticket = KerberosSupplementalTicketManager.FromIdToken(TestTokenWithTicket);
+
+            Assert.IsNotNull(ticket);
+            Assert.IsTrue(string.IsNullOrEmpty(ticket.ErrorMessage));
+            Assert.IsFalse(string.IsNullOrEmpty(ticket.KerberosMessageBuffer));
+            Assert.AreEqual(TestSpn, ticket.ServicePrincipalName, "Service principal name is not matched.");
+            Assert.AreEqual(TestCn, ticket.ClientName, "Clien name is not matched.");
```
trwalke, line 104:
```suggestion
            Assert.AreEqual(TestCn, ticket.ClientName, "Client name is not matched.");

```

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,220 @@
...
+        private const int DefaultLogonId = 0;
+        private const string KerberosClaimType = "xms_as_rep";
+        private const string IdTokenAsRepTemplate = @"{{""id_token"": {{ ""xms_as_rep"":{{""essential"":""false"",""value"":""{0}""}} }} }}";
+        private const string AccessTokenAsRepTemplate = @"{{""access_token"": {{ ""xms_as_rep"":{{""essential"":""false"",""value"":""{0}""}} }} }}";
+
+        /// <summary>
+        /// Creates a <see cref="KerberosSupplementalTicket"/> object from given ID token string..
+        /// </summary>
+        /// <param name="idToken">ID token string.</param>
+        /// <returns>A <see cref="KerberosSupplementalTicket"/> object if exists and parsed correctly. Null, otherwise.</returns>
```
trwalke, line 32:
```suggestion
        /// <returns>A <see cref="KerberosSupplementalTicket"/> object if exists in the <see cref="idToken"/> and is parsed correctly. Null, otherwise.</returns>

```

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,220 @@
...
+
+namespace Microsoft.Identity.Client.Kerberos
+{
+    /// <summary>
+    /// Helper class to manage Kerberos Ticket Claims.
+    /// </summary>
+    public static class KerberosSupplementalTicketManager
+    {
+        private const int DefaultLogonId = 0;
+        private const string KerberosClaimType = "xms_as_rep";
```
trwalke, line 24: We try to follow the pattern of _(lowercase char)****
for example "_defaultLogonId"

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Updated.

src/client/Microsoft.Identity.Client/Kerberos/Win32/Credential.cs
yoko-ms: Updated.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Updated.

src/client/Microsoft.Identity.Client/Kerberos/Win32/InitContextFlag.cs
yoko-ms: Removed.

tests/devapps/KerberosConsole/AADKerberosLogger.cs
yoko-ms: Nice catch. Test code replaced.

tests/Microsoft.Identity.Test.Unit/Kerberos/KerberosSupplementalTicketManagerTests.cs
yoko-ms: Updated.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Updated.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Changed.

src/client/Microsoft.Identity.Client/Kerberos/Win32/SspiSecurityContext.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        public ContextStatus InitializeSecurityContext(string targetName, out byte[] clientRequest)
+        {
+            var targetNameNormalized = targetName.ToLowerInvariant();
+
+            clientRequest = null;
+
+            // 1. acquire
+            // 2. initialize
+            // 3. ??
```
trwalke, line 70: What are these comments trying to say?

src/client/Microsoft.Identity.Client/Kerberos/Win32/LsaTokenSafeHandle.cs
```diff
@@ -0,0 +1,67 @@
...
+                var error = Marshal.GetLastWin32Error();
+
+                throw new Win32Exception(error);
+            }
+
+            return true;
+        }
+
+        public void Impersonate()
+        {
```
trwalke, line 37: Is this needed? Seems like it isnt used anywhere?

src/client/Microsoft.Identity.Client/Kerberos/Win32/SspiSecurityContext.cs
yoko-ms: This code part just copied from Kerberos.NET public package (https://www.nuget.org/packages/Kerberos.NET/) to support Windows Ticket Cache interface. I want to keep the code as it is.

src/client/Microsoft.Identity.Client/Kerberos/Win32/LsaTokenSafeHandle.cs
yoko-ms: This code part just copied from Kerberos.NET public package (https://www.nuget.org/packages/Kerberos.NET/) to support Windows Ticket Cache interface. I want to keep the code as it is.

trwalke:
I think this is a good candidate for E2E automation tests. @yoko-ms once you have a test environment we can create some tests for it.

build/platform_and_feature_flags.props
bgavrilMS, line 30: This is a build time check. We always build on Windows, so the condition is always true and pretty much useless :), i.e. the flag will be available even on Xamarin etc. 

image

bgavrilMS: To enable this Kerberos on desktop OS supporting Windows, we need to add this flag to .NET Fx, .NET Core, .NET 5, .NET Standard and UWP. @yoko-ms  - do you expect this to work on UWP ?

Secondly, for .NET Core and .NETStandard targets, we need to further check the OS using a helper like this [one](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1770654e7a2985dc06849366e5ae8e229b22faa7/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOs/DesktopOsHelper.cs#L23)

@yoko-ms - I can help make this changes.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -462,6 +465,19 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
             return (T)this;
         }
 
+        /// <summary>
+        /// Sets the parameters required to get a Kerberos Ticket from Azure AD service.
+        /// </summary>
+        /// <param name="servicePrincipalName">Service principal name to get Kerberos Service Ticket.</param>
+        /// <param name="ticketContainer">Container to use for Kerberos Ticket.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public T WithKerberosTicketClaim(string servicePrincipalName, KerberosTicketContainer ticketContainer)
```
bgavrilMS, line 474: With this, Kerberos will be made available to all flows of public client applications and confidential client applications. Do we really support this for confidential client apps?

bgavrilMS: Actually from your wiki it seems this is only available for public client flows.... So it should go into `PublicClientApplicationBuilder`

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -274,4 +275,9 @@
     </None>
   </ItemGroup>
 
+  <!-- Configurations for Azure AD Kerberos. -->
+  <ItemGroup Condition="$([MSBuild]::IsOsPlatform('Windows'))">
```
bgavrilMS, line 279: Not really needed as we only build on Windows. I will fix this for you.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -130,7 +133,15 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
             }
 
             _oAuth2Client.AddBodyParameter(OAuth2Parameter.Scope, scopes);
+
+#if SUPPORT_KERBEROS
```
bgavrilMS, line 137: My main question is here - what is MSAL expected to do on non-Windows operating systems? We have 2 options: 

- throw a meaningful exception 
- do nothing

I generally prefer to throw a meaningful exception as it is more obvious. But if the flow can be made to work by providing a degraded experience, then maybe we can go down this route. How did you fix this problem for MSAL C++?

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -130,7 +133,15 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
...
+            // Add Kerberos Ticket claims if there's valid service principal name in Configuration.
+            // Kerberos Ticket claim is only allowed at token request due to security issue.
+            // It should not be included for authorize request.
+            KerberosSupplementalTicketManager.AddKerberosTicketClaim(_oAuth2Client, _requestParams);
+#else
```
bgavrilMS, line 142: I wonder if we should merge these...

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,219 @@
...
+        /// <param name="requestParams"><see cref="AuthenticationRequestParameters"/> containing request parameters.</param>
+        internal static void AddKerberosTicketClaim(
+            OAuth2Client oAuth2Client,
+            AuthenticationRequestParameters requestParams)
+        {
+            string kerberosClaim = GetKerberosTicketClaim(
+                requestParams.RequestContext.ServiceBundle.Config.KerberosServicePrincipalName,
+                requestParams.RequestContext.ServiceBundle.Config.TicketContainer);
+
+            if (string.IsNullOrEmpty(kerberosClaim))
```
bgavrilMS, line 174: In `AcquireTokenSilent`, if claims are defined, then we bypass the token cache. Should we bypass the token cache if the kerberos claim is defined?

build/platform_and_feature_flags.props
yoko-ms: Thanks for pointing. The AAD Kerberos feature can be supported in all platform/OS excepts the Ticket Cache support feature which has dependency on Windows. Based on this, I figured out that I can use (iOS || MAC || ANDROID) combination as a replacement of SUPPORT_KERBEROS.
- Removed and replaced to use existing (iOS || MAC || ANDROID) definition.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
yoko-ms: The AAD Kerberos feature can be supported in all platform/OS excepts the Ticket Cache support feature which only has dependency on Windows. 
- Removed this condition.
- Changed all Windows OS dependent files so that they are not included in build time using (iOS || MAC || ANDROID) condition.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
yoko-ms: For the AAD Kerberos Feature,
 - It can be supported for all OS and platforms.
 - One exception is for the Ticket Cache interface which is only supported in Windows OS platform now.

I'd make a change to conditionally compile the Windows Ticket Cache interface components using the (iOS || MAC || ANDROID) flag combination. 
- I'd changed the KerberosSupplementalTicketManager to throw exception when the user  call's Ticket Cache associated function in non-Windows OS platform.

With these changes, the whole AAD Kerberos feature  can be supported for all OS platforms.


In C++, there's a flag indicating the Windows OS platform, and all AAD Kerberos Features are only included for Windows target.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
yoko-ms: Yes. Kerberos Ticket claim should be merged into existing claims.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: Yes. Since the Kerberos Ticket can be updated, it's better not to use cached token if possible.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -91,7 +91,9 @@ public static void SaveToWindowsTicketCache(KerberosSupplementalTicket ticket)
          /// </remarks>
         public static void SaveToWindowsTicketCache(KerberosSupplementalTicket ticket, long logonId)
         {
-#if SUPPORT_KERBEROS
+#if (iOS || MAC || ANDROID)
```
bgavrilMS, line 94: This is better. 

But it does not take into account code that runs on Mac and Linux on .NET Core. There are a bunch of helpers in `DesktopOsHelper` class like `IsWindows`.

yoko-ms: Updated.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -267,6 +267,12 @@
     <PackageReference Include="Microsoft.CodeAnalysis.RulesetToEditorconfigConverter" Version="3.0.0" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="Microsoft.VisualBasic">
+      <HintPath>..\..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\Microsoft.VisualBasic.dll</HintPath>
```
trwalke, line 272: What is this used for?

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
```diff
@@ -0,0 +1,240 @@
...
+            }
+
+            return null;
+        }
+
+        /// <summary>
+        ///  Checks the current application is running under Windows OS or not.
+        /// </summary>
+        /// <returns>True if running under Windows OS. False, otherwise.</returns>
+        internal static bool IsWindows()
```
trwalke, line 170: I just realized that we already have this functionality here
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/DesktopOsHelper.cs#L23

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
yoko-ms: Nice catch. Test code was pushed without removal.
Removed.

src/client/Microsoft.Identity.Client/Kerberos/KerberosSupplementalTicketManager.cs
yoko-ms: I'd tried DesktopOsHelper.IsWindows() at the first time, but it is not working for net45 environment and compiler complains there's missed assembly reference.  Checking all used cases, it's only working under NET_CORE or DESKTOP condition. With this reason, I'd implemented this function with reference for it.

bgavrilMS:
Thx for the follow up @yoko-ms . I am looking at this now and am making a few changes.

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/Kerberos/AcceptContextFlag.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.Identity.Client.Kerberos.Win32
+namespace Microsoft.Identity.Client.Platforms.Features.DesktopOs.Kerberos
```
bgavrilMS, line 6: @yoko-ms - I moved all the low level pinvoke stuff into a directory that is available only on Win32. This will avoid having all the `#if !ANDROID && ...` flags.

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/Kerberos/LsaBufferSafeHandle.cs
```diff
@@ -1,31 +1,28 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-using System;
-using System.Runtime.InteropServices;
-
-namespace Microsoft.Identity.Client.Kerberos.Win32
-{
-#if !(iOS || MAC || ANDROID)
```
bgavrilMS: Not sure what the spacing problem is ... but there are no changes except removal of #if and namepsace rename

bgavrilMS (approved):
I tested using the console app added and it works fine.

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/Kerberos/AcceptContextFlag.cs
yoko-ms: Thanks for all your works.

bgavrilMS:
@trwalke - when you get an hour or so, can you please review as well ? I think this is ready to be committed. We can discuss E2E tewsting later, there is a manual test app that we can use for now.

tests/Microsoft.Identity.Test.Unit/Kerberos/KerberosSupplementalTicketManagerTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// <summary>
+        /// Username within the ID token.
+        /// </summary>
+        private static readonly string _testClientName = "localAdmin@aadktest.onmicrosoft.com";
+
+        /// <summary>
+        /// Sample ID Token without Kerbero Service Ticket.
+        /// </summary>
+        private static readonly string _testIdToken =
+            "eyJ0eXAiOiJKV1QiLCJyaCI6IjAuQWdBQXI0R0lRckdhczBDQldEWVJOWV9fYUlLMElWSlJKck5NbXRqQW1uamszcDRzQU5NLiIsImFsZyI6IlJTMjU2"
```
jmprieur, line 34: do we think that this is OK? / will be accepted by compliance?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -1,129 +1,140 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
jmprieur, line 1: what happened to this file? space issues?

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -1,357 +1,374 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
jmprieur, line 1: Same thing. Why does GitHub think that the files are that different? spaces/tab issues?

jmprieur (commented):
Thanks @yoko-ms 
LGTM

I'm wondering about space formatting in 2 big files, though?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
yoko-ms: It's due to merge from master branch.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
yoko-ms: Same - It's due to git merge from master branch.

tests/Microsoft.Identity.Test.Unit/Kerberos/KerberosSupplementalTicketManagerTests.cs
yoko-ms: It's a IdToken sample and doesn't contains any sensitive information. Is that a matter?

trwalke:
I took a look and everything seems fine with the latest updates. Thanks @yoko-ms

trwalke (approved):
LGTM

Issue #2131, opened 2020-10-23
[Feature Request] Support prefersEphemeralWebBrowserSession to hide iOS security prompt for system browser


ckrempp91:
## Why?
This mandatory popup is very annoying when using the system browser and can cause confusion to users of the application.

![iOS_Prompt](https://user-images.githubusercontent.com/15944499/97018406-bc714080-1514-11eb-8865-37787b1657cf.png)

I got the image from this issue: #512 (Wants to Use "Azure.com" to Sign In - iOS 11)

## How?
If the `prefersEphemeralWebBrowserSession` property is set to `true` on the `ASWebAuthenticationSession` then the system browser would not attempt to display the prompt. The trade off is that SSO can not happen because the cookies/browser data will be locked to the single session as shown by the Apple documentation:

> Set prefersEphemeralWebBrowserSession to true to request that the browser doesn’t share cookies or other browsing data between the authentication session and the user’s normal browser session. Whether the request is honored depends on the user’s default web browser. Safari always honors the request.

One thing to note is that this is only supported on iOS 13 (and above) so for the cases of iOS 11 & 12 the popup would still be required.

We can have a new method added to the PublicClientApplicationBuilder similar to the `withIosKeychainSecurityGroup`. The new method could be named something like: 
* `withSupportOfSso(bool supportSso)` 
* `withSsoSupport(bool supportSso)` 
* `withIosSsoSupport(bool supportSso)`

The default for the bool can be `false` to match the current behavior, in the implementation code. For the method exposed through the client builder it can be `true` to have clarity of the functionality to the developer.

Apple documentation on the [`prefersEphemeralWebBrowserSession`](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/3237231-prefersephemeralwebbrowsersessio)

jennyf19:
@ckrempp91 we don't plan to make this change. We'll keep our eye on it in case more customers ask for this. Thanks!

ckrempp91:
@jennyf19 If there is a technical limitation for not being able to include this that is fine, but if this is technically possible and just won't be included could you please explain the reason?

maQus123:
@jennyf19 What is the reason not implementing this? We're using MSAL/Azure B2C in an consumer facing app and believe it would improve UX. Thanks for info!

jennyf19:
Thanks @maQus123, will re-open the issue and let the team prioritize

jennyf19:
[...]

PR #2603, opened 2021-05-05
Option to hide prompt in iOS + disposed semaphore.

SameerK-MSFT:
Added option to hide prompt in iOS. Disposed SemaphoreSlim after use to release resource.
#2131

### DIFF
--- a/build/platform_and_feature_flags.props
+++ b/build/platform_and_feature_flags.props
@@ -1,29 +1,29 @@
 <Project>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetCore)' ">
-    <DefineConstants>$(DefineConstants);NET_CORE;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_OS_SYSTEM_BROWSER;SUPPORTS_BROKER;SUPPORTS_WEBVIEW2</DefineConstants>
+    <DefineConstants>$(DefineConstants);NET_CORE;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_BROKER;SUPPORTS_WEBVIEW2</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet5Win)' ">
-    <DefineConstants>$(DefineConstants);NET5_WIN;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_OS_SYSTEM_BROWSER;SUPPORTS_BROKER;SUPPORTS_WEBVIEW2</DefineConstants>
+    <DefineConstants>$(DefineConstants);NET5_WIN;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_BROKER;SUPPORTS_WEBVIEW2</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkUap)' ">
-    <DefineConstants>$(DefineConstants);WINDOWS_APP;SUPPORTS_BROKER;SUPPORTS_OS_SYSTEM_BROWSER;SUPPORTS_CUSTOM_CACHE</DefineConstants>
+    <DefineConstants>$(DefineConstants);WINDOWS_APP;SUPPORTS_BROKER;SUPPORTS_CUSTOM_CACHE</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" $(TargetFramework.StartsWith('MonoAndroid')) ">
-    <DefineConstants>$(DefineConstants);ANDROID;SUPPORTS_BROKER;SUPPORTS_OS_SYSTEM_BROWSER</DefineConstants>
+    <DefineConstants>$(DefineConstants);ANDROID;SUPPORTS_BROKER</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(TargetFramework)' == '$(TargetFrameworkNetDesktop45)' Or '$(TargetFramework)' == '$(TargetFrameworkNetDesktop461)' ">
-    <DefineConstants>$(DefineConstants);DESKTOP;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_OS_SYSTEM_BROWSER;SUPPORTS_WEBVIEW2</DefineConstants>
+    <DefineConstants>$(DefineConstants);DESKTOP;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_WEBVIEW2</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop461)'">
-    <DefineConstants>$(DefineConstants);SUPPORTS_BROKER;SUPPORTS_OS_SYSTEM_BROWSER</DefineConstants>
+    <DefineConstants>$(DefineConstants);SUPPORTS_BROKER</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkIos)'">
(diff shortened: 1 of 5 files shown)


trwalke (dismissed):
Minor comment. 
LGTM otherwise.

henrik-me:
no tests/utests? ... would be great to have system browser enabled for e2e testing....
imo: at minimum an existing manual test should be expanded to cover this scenario

build/platform_and_feature_flags.props
```diff
@@ -6,19 +6,19 @@
     <DefineConstants>$(DefineConstants);NET5_WIN;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_OS_SYSTEM_BROWSER;SUPPORTS_BROKER;SUPPORTS_WEBVIEW2</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkUap)' ">
-    <DefineConstants>$(DefineConstants);WINDOWS_APP;SUPPORTS_BROKER;SUPPORTS_CUSTOM_CACHE</DefineConstants>
+    <DefineConstants>$(DefineConstants);WINDOWS_APP;SUPPORTS_BROKER;SUPPORTS_OS_SYSTEM_BROWSER;SUPPORTS_CUSTOM_CACHE</DefineConstants>
```
bgavrilMS, line 9: UAP / UWP (Universal Windows Platform) does not support SYSTEM_BROWSER

src/client/Microsoft.Identity.Client/ApiConfig/SystemWebViewOptions.cs
```diff
@@ -1,169 +1,176 @@
...
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        public SystemWebViewOptions()
+        {
+            ValidatePlatformAvailability();
+        }
+
+        /// <summary>
+        /// When the user finishes authenticating, MSAL will respond with a 200 OK message,
```
bgavrilMS, line 37: Please add comments that these existing options are exclusive for desktop operating systems (so .NET Clasic and .NET Core). And if the user tries to `set` one of these on Android, iOS or UWP, we should throw a nice exception.

src/client/Microsoft.Identity.Client/ApiConfig/SystemWebViewOptions.cs
```diff
@@ -1,169 +1,176 @@
...
+        /// When the user finishes authenticating, but an error occurred, MSAL will redirect the browser to the given Uri
+        /// </summary>
+        /// <remarks>Takes precedence over <see cref="HtmlMessageError"/></remarks>
+        public Uri BrowserRedirectError { get; set; }
+
+        /// <summary>
+        /// This hides privacy prompt displayed on the iOS Devices (ver 13.0+) when set to true.
+        /// By default, it is false and displays the prompt.
+        /// </summary>
+        public bool HidePrivacyPrompt { get; set; } = false;
```
bgavrilMS, line 65: This one should only be visible in iOS and NetStandard. Trying to set this from another platform should result in an exception and the API should be hidden as well.

bgavrilMS, line 65: For naming purposes ... I would name it `iOSHidePrivacyPrompt` to make it clear it only affects iOS. And the comments should mention prefersEphemeralWebBrowserSession and the fact that the browser will not remember previous logged-in users.

src/client/Microsoft.Identity.Client/Platforms/iOS/IosWebUIFactory.cs
```diff
@@ -1,40 +1,41 @@
...
-            
-            return new SystemWebUI()
-            {
-                RequestContext = requestContext
-            };
-        }
-
-    }
-}
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bgavrilMS, line 1: There's smth wrong with your spacing settings. The diffs are all screwed up :)

bgavrilMS: We use the VS defaults around tabs and there is an .editorconfig file that sets some styles for you... 
image

bgavrilMS (changes requested):
ios logic looks good, but needs some work to better hide some of the APIs

src/client/Microsoft.Identity.Client/ApiConfig/SystemWebViewOptions.cs
```diff
@@ -1,169 +1,176 @@
...
+#if DESKTOP
+            url = url.Replace("&", "^&");
+            Process.Start(new ProcessStartInfo("cmd", $"/c start microsoft-edge:{url}") { CreateNoWindow = true });
+            await Task.FromResult(0).ConfigureAwait(false);
+#else
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                url = url.Replace("&", "^&");
+                Process.Start(new ProcessStartInfo("cmd", $"/c start microsoft-edge:{url}") { CreateNoWindow = true });
```
bgavrilMS, line 122: Becuase you included this in the Android and iOS, it causes a crash. Process is not available - see the PR build failure ... `src\client\Microsoft.Identity.Client\ApiConfig\SystemWebViewOptions.cs(122,17): Error CS0103: The name 'Process' does not exist in the current context`

build/platform_and_feature_flags.props
SameerK-MSFT: Updated to removed SUPPORTS_OS_SYSTEM_BROWSER from all configurations

src/client/Microsoft.Identity.Client/ApiConfig/SystemWebViewOptions.cs
SameerK-MSFT: Hides for UWP i.e. WINDOWS_APP

src/client/Microsoft.Identity.Client/ApiConfig/SystemWebViewOptions.cs
SameerK-MSFT: Updated the flags

src/client/Microsoft.Identity.Client/ApiConfig/SystemWebViewOptions.cs
SameerK-MSFT: Added the hide attribute

src/client/Microsoft.Identity.Client/Platforms/iOS/IosWebUIFactory.cs
SameerK-MSFT: I have those settings. Not sure what went wrong. Hope it goes fine in the second push.

src/client/Microsoft.Identity.Client/ApiConfig/SystemWebViewOptions.cs
SameerK-MSFT: Good suggestion. Done.

bgavrilMS (approved):

SameerK-MSFT:
> no tests/utests? ... would be great to have system browser enabled for e2e testing....
> imo: at minimum an existing manual test should be expanded to cover this scenario

I have a test ready in the sample app. That can be pushed after the new NuGet gets published and consumed by App.

jennyf19:
@SameerK-MSFT the current iOS tests are using the [embedded webview](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/tests/devapps/XForms/XForms/TestPage.xaml.cs#L196), i don't remember exactly, but we don't have system browser for the UI automated tests. You should be able to make changes to the other Xamarin dev app now, which would apply the changes you made. Also, did you change the spaces to tabs because the diff still looks odd to me.

bgavrilMS:
The iOS builds seem to be failing with an env error: 

`Xamarin.iOS task failed with error Failed to find iOS signing identity. Verify the signing and provisioning information provided.. For guidance on setting up the build definition, see https://go.microsoft.com/fwlink/?LinkId=760847.`

Is it possible that the test provisioning profile expired? @trwalke - is this the same error happening a few weeks ago?

trwalke:
> The iOS builds seem to be failing with an env error:
> 
> `Xamarin.iOS task failed with error Failed to find iOS signing identity. Verify the signing and provisioning information provided.. For guidance on setting up the build definition, see https://go.microsoft.com/fwlink/?LinkId=760847.`
> 
> Is it possible that the test provisioning profile expired? @trwalke - is this the same error happening a few weeks ago?

Yes, However, this time is seems to be related to an actual issue with the cert and not just a service issue, im requesting a new certificate for the build.
@bgavrilMS

trwalke:
restarting builds

SameerK-MSFT:
I will update sample app after the package is released

PR #2588, opened 2021-04-28
Remove the HAVE_CAS define so that Newtonsoft.Json never uses dynamic code generation as its not supported in Unity.

najadojo:
Remove the HAVE_CAS define so that Newtonsoft.Json never uses dynamic code generation as its not supported in Unity.
Solves #2586

PR #2605, opened 2021-05-06
Remove HAVE_CAS flag.

pmaytak:
Don't review. Duplicate of #2588. Only created this PR to run the PR builds since we don't run them on forked repos.


jennyf19:
you can also set the forked branch to merge to a different branch (not master), so you can merge the changes there and then update or make further changes as needed before merging to master. i've found that helpful, esp w/larger PRs where there may be additional work needed.

pmaytak:
Thanks @jennyf19, that's helpful, yea, I'll do that from now on.

Issue #2584, opened 2021-04-27
[Feature Request][M] Add AuthenticationResult.GetIdTokenClaims()


bgavrilMS:
See #2583 for details

neha-bhargava:
Closing this as we plan to expose the id token claims with #2583 in the tenant profiles

Issue #2583, opened 2021-04-27
[Feature Request] [2L] Expose tenant profiles in IAccount and ID token claims in the AuthenticationResult


jmprieur:
**Is your feature request related to a problem? Please describe.**
Several customers want to get the ID token claims in the authentication result. See for instance #2580.

**Describe the solution you'd like**
- The AuthenticationResult could expose a new method `.GetIdTokenClaims()` returning a `ClaimsPrincipal` as this is the data structure which is commonly used in .NET - see #2584 
- More generally, we'd want that `IAccount` exposes `TenantProfiles`
- Update API review with .NET and Java design

```CSharp
interface IAccount
{
 // current members

 /// <summary>
 /// Get the information about the account in each tenant, keyed by tenant Id
 /// </summary>
 IDictionnary<string, ITenantProfile> TenantProfiles { get; }
}

interface ITenantProfile
{
  /// <summary>
  /// Get object if of the account in this tenant
  /// </summary>
  public string Oid {get;}

  /// <summary>
  /// Get the tenant ID
  /// </summary>
  public string TenantId {get;}

  /// <summary>
  /// Get the claims of the ID token for the account in this tenant
  /// </summary>
  public ClaimsPrincipal Claims {get;}

  /// <summary>
  /// Is the tenant the home tenant of the user?
  /// </summary>
  public bool IsHomeTenant {get;}
}
```

**Additional context**
- Tenant profile is a consistency item ... See the API review spec / design for Android: [Account & Tenant Profiles
](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FAccountDomainModel%2FOverview.md&_a=preview&version=GBdev&anchor=account-%26-tenant-profiles)
- This is mostly for public client applications, or applications that don't use ASP.NET / ASP.NET Core as these framework already populate the .User.

jmprieur:
See also https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2571

bgavrilMS:
Not sure this can be done for Confidential Client due to not checking signatures, continuing discussions offline.

neha-bhargava:
FYI: PR for Java's implementation of tenant profiles in api review: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/3515?path=%2FAccountDomainModel%2FOverview.md

bgavrilMS:
[...]

PR #2621, opened 2021-05-11
Add method to get the id token claims

neha-bhargava:
#2584 
Add method to return claims from id token

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -5,9 +5,14 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
+#if !NETSTANDARD1_3
+using System.Security.Claims;
+#endif
 using Microsoft.Identity.Client.AuthScheme;
 using Microsoft.Identity.Client.Cache.Items;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
+using Microsoft.Identity.Client.Utils;
+using Microsoft.Identity.Json;
 
 namespace Microsoft.Identity.Client
 {
@@ -263,5 +268,47 @@ namespace Microsoft.Identity.Client
                 _authenticationScheme?.AuthorizationHeaderPrefix ?? TokenType,
                 AccessToken);
         }
+
+#if !NETSTANDARD1_3
+        /// <summary>
+        /// Creates an object to hold the claims from an ID Token.
+        /// </summary>
+        /// <returns>ClaimsPrincipal object containing the claims from the ID Token.</returns>
+        public ClaimsPrincipal GetIdTokenClaims()
+        {
+            if (string.IsNullOrEmpty(IdToken))
+                return null;
+
+            string[] idTokenSegments = IdToken.Split(new[] { '.' });
+
+            if (idTokenSegments.Length < 2)
+            {
+                throw new MsalClientException(
+                    MsalError.InvalidJwtError,
+                    MsalErrorMessage.IDTokenMustHaveTwoParts);
+            }
+
+            try
+            {
+                IDictionary<string, string> idTokenClaims = JsonConvert.DeserializeObject<Dictionary<string, string>>(
+                    Base64UrlHelpers.DecodeToString(idTokenSegments[1]));
+
+                IList<Claim> claims = new List<Claim>();
+                foreach (KeyValuePair<string, string> claim in idTokenClaims)
+                {
+                    claims.Add(new Claim(claim.Key, claim.Value));
+                }
+
+                return new ClaimsPrincipal(new ClaimsIdentity(claims));
+            }
+            catch (JsonException exc)
+            {
+                throw new MsalClientException(
+                    MsalError.JsonParseError,
+                    MsalErrorMessage.FailedToParseIDToken,
+                    exc);
+            }
+        }
+#endif
     }
 }
--- a/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
+++ b/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
@@ -303,6 +303,30 @@ namespace Microsoft.Identity.Test.Common.Core.Mocks
             return string.Format(CultureInfo.InvariantCulture, "someheader.{0}.somesignature", Base64UrlHelpers.Encode(id));
         }
 
+        public static string CreateIdTokenWithExtraClaim(string uniqueId, string displayableId, string tenantId)
+        {
+            string id = "{\"aud\": \"e854a4a7-6c34-449c-b237-fc7a28093d84\"," +
+                        "\"iss\": \"https://login.microsoftonline.com/6c3d51dd-f0e5-4959-b4ea-a80c4e36fe5e/v2.0/\"," +
+                        "\"iat\": 1455833828," +
+                        "\"nbf\": 1455833828," +
+                        "\"exp\": 1455837728," +
+                        "\"ipaddr\": \"131.107.159.117\"," +
+                        "\"name\": \"Marrrrrio Bossy\"," +
+                        "\"oid\": \"" + uniqueId + "\"," +
+                        "\"preferred_username\": \"" + displayableId + "\"," +
+                        "\"some_claim\": \"value\"," +
+                        "\"sub\": \"K4_SGGxKqW1SxUAmhg6C1F6VPiFzcx-Qd80ehIEdFus\"," +
+                        "\"tid\": \"" + tenantId + "\"," +
+                        "\"ver\": \"2.0\"}";
+            return string.Format(CultureInfo.InvariantCulture, "someheader.{0}.somesignature", Base64UrlHelpers.Encode(id));
+        }
+
+        public static string CreateIdTokenWithInvalidJson()
+        {
+            string id = "{\"aud\"}";
+            return string.Format(CultureInfo.InvariantCulture, "someheader.{0}.somesignature", Base64UrlHelpers.Encode(id));
+        }
+
         private static string CreateIdTokenForB2C(string uniqueId, string tenantId, string policy)
         {
             string id = "{" + 
--- a/tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
@@ -2,10 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Security.Claims;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Test.Common;
+using Microsoft.Identity.Test.Common.Core.Mocks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Identity.Test.Unit.PublicApiTests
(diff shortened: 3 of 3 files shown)


tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
```diff
@@ -303,6 +303,30 @@ public static string CreateIdToken(string uniqueId, string displayableId, string
             return string.Format(CultureInfo.InvariantCulture, "someheader.{0}.somesignature", Base64UrlHelpers.Encode(id));
         }
 
+        public static string CreateIdTokenWithExtraClaim(string uniqueId, string displayableId, string tenantId)
+        {
+            string id = "{\"aud\": \"e854a4a7-6c34-449c-b237-fc7a28093d84\"," +
```
jmprieur, line 308: For the tests, you can probably use Wilson (https://www.nuget.org/packages/Microsoft.IdentityModel.JsonWebTokens) to create the raw IDToken from claims ...

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -263,5 +268,47 @@ public string CreateAuthorizationHeader()
                 _authenticationScheme?.AuthorizationHeaderPrefix ?? TokenType,
                 AccessToken);
         }
+
+#if !NETSTANDARD1_3
+        /// <summary>
+        /// Creates an object to hold the claims from an ID Token.
```
jmprieur, line 274: `Returns a <See ClaimsPrincipal/> containing the claims extracted from the ID Token.`

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -263,5 +268,47 @@ public string CreateAuthorizationHeader()
...
+        /// </summary>
+        /// <returns>ClaimsPrincipal object containing the claims from the ID Token.</returns>
+        public ClaimsPrincipal GetIdTokenClaims()
```
jmprieur, line 277: We should probably name the method `GetUser()`. This would be consistent with what .NET, ASP.NET, and ASP.NET Core do?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -263,5 +268,47 @@ public string CreateAuthorizationHeader()
...
+        {
+            if (string.IsNullOrEmpty(IdToken))
```
jmprieur, line 279: I guess that's the case for client credentials? right?

jmprieur: We don't verify the signature of the IDToken (for v2.0 ID Tokens). Maybe we want to indicate it? (not a big problem in practice as it comes from AAD)

jmprieur (commented):
This is a good start, @neha-bhargava
I proposed a few suggestions / things to discuss.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -5,9 +5,14 @@
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
+#if !NETSTANDARD1_3
```
bgavrilMS, line 8: That's a bummer.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: How can we indicate that we don't verify the signature?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
```diff
@@ -52,6 +55,111 @@ public void GetAuthorizationHeader()
             Assert.AreEqual("SomeTokenType at", ar.CreateAuthorizationHeader());
         }
 
+        [TestMethod]
+        public void GetIdTokenClaimsTest()
```
bgavrilMS, line 59: Maybe add a check on some of the integration tests.

bgavrilMS:
LGTM, ready for review.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
neha-bhargava: Yes, the ClaimsPrincipal class is not available for NetStandard1.3.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
neha-bhargava: Working on it

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
bgavrilMS: You can use Wilson to create "correct" tokens with signatures etc. Alternatively you can just add some assertions to integration tests, which will ensure you can parse actual ESTS tokens.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
neha-bhargava: I took a look at the api review doc: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FAccountDomainModel%2FOverview.md&_a=preview&version=GBdev&anchor=account-%26-tenant-profiles. Here the method is named account.getClaims(). But it is not exposed in authenticationResult  directly.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -263,5 +268,46 @@ public string CreateAuthorizationHeader()
...
+        /// Returns a <See cref= "ClaimsPrincipal"/> object containing the claims extracted from the ID Token.
+        /// </summary>
+        public ClaimsPrincipal GetIdTokenClaims()
+        {
+            if (string.IsNullOrEmpty(IdToken))
+                return null;
+
+            string[] idTokenSegments = IdToken.Split(new[] { '.' });
+
+            if (idTokenSegments.Length < 2)
```
jennyf19, line 283: >< 2 [](http://example.com/codeflow?start=39&length=3)

I think it should be 3, and the message should be that it is specifically a JWT format, which has three base64url-encoded strings, separated by dots.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -263,5 +268,46 @@ public string CreateAuthorizationHeader()
                 _authenticationScheme?.AuthorizationHeaderPrefix ?? TokenType,
                 AccessToken);
         }
+
+#if !NETSTANDARD1_3
+        /// <summary>
+        /// Returns a <See cref= "ClaimsPrincipal"/> object containing the claims extracted from the ID Token.
+        /// </summary>
+        public ClaimsPrincipal GetIdTokenClaims()
```
jennyf19, line 276: this logic is the same as [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/Internal/IdToken.cs#L9) shouldn't this go in the IdToken class and be exposed a different way?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
```diff
@@ -2,10 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
```
jennyf19, line 5: you can assert the exception intead of doing the try/catch

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
```diff
@@ -52,6 +55,111 @@ public void GetAuthorizationHeader()
...
+                MockHelpers.CreateIdTokenWithExtraClaim(TestConstants.UniqueId, TestConstants.DisplayableId, TestConstants.TenantId), 
+                new[] { "scope" }, Guid.NewGuid(),
+                "SomeTokenType",
+                new AuthenticationResultMetadata(TokenSource.Cache));
+            
+            ClaimsPrincipal claimsPrincipal = ar.GetIdTokenClaims();
+            Assert.IsNotNull(claimsPrincipal);
+            Assert.AreEqual(13, claimsPrincipal.Claims.Count());
+            var claims = claimsPrincipal.Claims;
+            IDictionary<string, string> claimsDictionary = new Dictionary<string, string>();
```
jennyf19, line 78: i'm not sure what this is doing that's different from the code? seems like the same code repeated again...but maybe i'm reading this wrong.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: I think it's somewhat different. Internall, we deserialize the id token to an object called IDToken. Neha's deserializing the idtoken to a dictionary and creating a ClaimsPrincipal. 

There may a bit of code duplication in how we parse that id token string (i.e. the splitting) - possibly worth have a static method somewhere to reduce duplication.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: @jmprieur @jennyf19 @neha-bhargava 

Today, MSAL parses the IDTOken internally and uses a bunch of claims from it: oid, tid, sub, upn, preferred_username and more. So I think we must have already decided that we don't need to validate the id token?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: This splitting logic can be consolidated. The code is there from day 1 of ADAL probably :). Possibly to take into account v1 IDTokens, which are not signed.

bgavrilMS (approved):

bgavrilMS (approved):
I think only the naming is pending.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
neha-bhargava: This is the same logic we use to parse the idtoken to create idtoken object

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
neha-bhargava: Agree, there is a small part of code that is duplicated. I will try to refactor it into a static method.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
neha-bhargava: Here I am reversing the code to extract the claims to assert the claim I have added.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs
```diff
@@ -52,6 +55,111 @@ public void GetAuthorizationHeader()
...
+            {
+                ar.GetIdTokenClaims();
+                Assert.Fail("Should have failed due to invalid token");
+            }
+            catch (MsalClientException e)
+            {
+                Assert.IsNotNull(e);
+                Assert.AreEqual(MsalError.JsonParseError, e.ErrorCode);
+            }
+        }
```
pmaytak, line 161: This can be refactored like https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/cc1ba2338f98153bc2bd28e00bf4c17e73058ae2/tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/InstanceProviderTests.cs#L106-L107

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
pmaytak: It is available in netstandard2.0+, if we move to that.

pmaytak (approved):
👍 (+ minor refactorings and what others mentioned).

neha-bhargava:
Closing this PR. Will expose the method with #2583

Issue #2637, opened 2021-05-17
[Feature Request] Expose AuthenticationResult.TokenType with MSAL.NET


hkuadithya:
We are migrating some legacy code in our projects from ADAL to MSAL.

Both ADAL and MSAL return an object of type `AuthenticationResult` after acquiring an access token.

* ADAL: [Microsoft.IdentityModel.Clients.ActiveDirectory.AuthenticationResult](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.clients.activedirectory.authenticationresult?view=azure-dotnet)
* MSAL: [Microsoft.Identity.Client.AuthenticationResult](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.authenticationresult?view=azure-dotnet)

In our older code path, we rely on the type of ADAL access token i.e. `AuthenticationResult.AccessTokenType` to set the scheme in [AuthenticationHeaderValue](https://docs.microsoft.com/en-us/previous-versions/visualstudio/hh193878(v=vs.118)) for making another request.

```
new AuthenticationHeaderValue(
    scheme: firstPartyAuthResult.AccessTokenType,
    parameter: firstPartyAuthResult.AccessToken);
```

MSAL's [AutenticationResult](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.authenticationresult?view=azure-dotnet&viewFallbackFrom=azure-dotnet) does have `AccessToken` but it **doesn't** seem to have `AccessTokenType` (correct me If I'm wrong). 

How do I determine the `AccessTokenType` for a MSAL auth result? Please don't suggest hardcoding the scheme to `Bearer`. Thanks in advance.

SO thread mirroring this issue: https://stackoverflow.microsoft.com/questions/258737

hkuadithya:
@bgavrilMS could you provide a timeline for the release of MSAL.NET `4.32`? Is there a link I can refer to for checking the release timelines?

jennyf19:
@hkuadithya should be in 1-2 weeks.

PR #2638, opened 2021-05-17
- small updates, see below.

jennyf19:
- make tokenType public.
- remove un-needed string.format
- spelling fixes
- stopwatch.startnew()

#2637

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -228,13 +228,13 @@ namespace Microsoft.Identity.Client
         /// </summary>
         public Guid CorrelationId { get; }
 
-        internal ApiEvent ApiEvent { get; }
-
         /// <summary>
         /// Identifies the type of access token. By default tokens returned by Azure Active Directory are Bearer tokens.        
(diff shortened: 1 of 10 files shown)


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -228,13 +228,13 @@ public partial class AuthenticationResult
...
         public Guid CorrelationId { get; }
 
-        internal ApiEvent ApiEvent { get; }
-
         /// <summary>
         /// Identifies the type of access token. By default tokens returned by Azure Active Directory are Bearer tokens.        
         /// <seealso cref="CreateAuthorizationHeader"/> for getting an HTTP authorization header from an AuthenticationResult.
         /// </summary>
-        internal string TokenType { get; }
+        public string TokenType { get; }
```
bgavrilMS, line 235: Since you made this property public, you must also update the ctor of AuthenticationResult, otherwise developers will not be able to test properly. We had a customer issue with this in the past, so I even wrote a test `PublicTestConstructorCoversAllProperties` (which is happily failing on this PR).

jennyf19: thx @bgavrilMS the comment really threw me off. LOL

PR #2640, opened 2021-05-18
Fix for exception when token_type is missing from token response.

### LINKED ISSUE #2636: [Bug] Get token for client with ADFS throws null ref due to TokenType
Throws a null ref  in `GetKey()` on the TokenType being null when trying to save the access token.

not sure if related:
Also, the tenantId when using ADFS is null, so the cache key is missing the tenantId, and looks like the following: 
`567794c8-a349-4ye3-672a-9947accbde95__AppTokenCache`

and this is the authority: `https://127.0.0.1/adfs/` 

See this [line](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs#L68).

pmaytak:
Issue #2636. The [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) says token_type parameter is required in token response. 
Added a null check for token type after receiving token response. Reorganized some logging in `MsalTokenResponse`.


bgavrilMS:
Well, ADFS has sometimes been a step behind and fixing ADFS bugs is difficult since it's a boxed product. I think this is the right fix (just needs a unit test). We can also follow up with ADFS team for them to do the right thing and be OAuth2 compliant.

jennyf19:
I think throwing the ArgumentNullException is the correct thing to do if TokenType is null, but the question is, why is it not getting set to any default value? this is specifically with [MsalAccessTokenCacheKey](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs#L45)
Debugging through the integration tests, ADFS is sending back a token type of "bearer" and we do ignore case on the string comparisons. could also be an issue with customer set-up, i believe logs were provided in the issue. I've asked which version of ADFS is being used.
image

jennyf19:
It's a customer set-up issue, which is returning a null token_type.

I think the fix should go in TokenClient line 82, where we check for a mismatch of TokenType, but can also add a null check as well, and if null, let's just throw w/a good message to configure the IDP correctly to return the token type

pmaytak:
@jennyf19 Sounds good. Agree. Will make that change.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -362,6 +362,8 @@ public static string TokenTypeMismatch(string requestTokenType, string responseT
                 requestTokenType, responseTokenType);
         }
 
+        public const string TokenTypeMissing = "The response from the token endpoint does not contain token_type parameter. This happens if the Identity Provider (AAD, B2C, ADFS, etc.) did not include the token type in the response. In case of ADFS, verify the configuration or consider upgrading to the latest version. ";
```
jmprieur, line 365: the latest version of ADFS? (not the library?)

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -191,35 +191,36 @@ public void Log(ICoreLogger logger, LogLevel logLevel)
...
-
-                withPii.AppendLine($"Error: {Error} ErrorDescription: {ErrorDescription}");
-                withoutPii.AppendLine($"Error: {Error} ErrorDescription: {ErrorDescription}");
+                withPii.AppendLine($"{Environment.NewLine}[MsalTokenResponse]");
+                withPii.AppendLine($"Error: {Error}");
+                withPii.AppendLine($"ErrorDescription: {ErrorDescription}");
                 withPii.AppendLine($"Scopes: {Scope} ");
+                withPii.AppendLine($"ExpiresIn: {ExpiresIn}");
+                withPii.AppendLine($"RefreshIn: {RefreshIn}");
+                withPii.AppendLine($"AccessToken: {!string.IsNullOrEmpty(AccessToken)}");
```
jmprieur, line 200: Maybe we should rather write "AccessTokenReturned" ? as this is just the presence of the AT?

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -191,35 +191,36 @@ public void Log(ICoreLogger logger, LogLevel logLevel)
...
+                withPii.AppendLine($"AccessToken Type: {TokenType}");
+                withPii.AppendLine($"RefreshToken: {!string.IsNullOrEmpty(RefreshToken)}");
```
jmprieur, line 202: Same here

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -191,35 +191,36 @@ public void Log(ICoreLogger logger, LogLevel logLevel)
...
+                withPii.AppendLine($"IdToken: {!string.IsNullOrEmpty(IdToken)}");
```
jmprieur, line 203: And here?

jmprieur (commented):
LGTM.
Thanks @pmaytak

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
pmaytak: I made this message similar to the `TokenTypeMismatch` above. Yes, in this case it's probably a ADFS configuration issue or a ADFS bug.

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -757,6 +757,12 @@ public static class MsalError
         /// </summary>
         public const string TokenTypeMismatch = "token_type_mismatch";
 
+        /// <summary>
+        /// <para>What happens?</para>The response from the token endpoint does not contain token_type parameter.
```
jennyf19, line 761:
```suggestion
        /// <para>What happens?</para>The response from the token endpoint does not contain the token_type parameter.
```

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
jennyf19, line 365:
```suggestion
        public const string TokenTypeMissing = "The response from the token endpoint does not contain the token_type parameter. This happens if the Identity Provider (AAD, B2C, ADFS, etc.) did not include the token type in the token response. Verify the configuration of the IdP. ";
```

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -191,35 +191,36 @@ public void Log(ICoreLogger logger, LogLevel logLevel)
...
-                withPii.AppendLine($"Error: {Error} ErrorDescription: {ErrorDescription}");
-                withoutPii.AppendLine($"Error: {Error} ErrorDescription: {ErrorDescription}");
+                withPii.AppendLine($"{Environment.NewLine}[MsalTokenResponse]");
+                withPii.AppendLine($"Error: {Error}");
+                withPii.AppendLine($"ErrorDescription: {ErrorDescription}");
                 withPii.AppendLine($"Scopes: {Scope} ");
+                withPii.AppendLine($"ExpiresIn: {ExpiresIn}");
+                withPii.AppendLine($"RefreshIn: {RefreshIn}");
+                withPii.AppendLine($"AccessToken returned: {!string.IsNullOrEmpty(AccessToken)}");
+                withPii.AppendLine($"AccessToken Type: {TokenType}");
```
jennyf19, line 201: TokenType is not Pii

jennyf19: sorry, it's hard to review w/the lines changed and matching up what's pii and what's not. i thought our logs didn't have to be repeated anymore, and the pii logs would include the non-pii.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -191,35 +191,36 @@ public void Log(ICoreLogger logger, LogLevel logLevel)
...
-                    $"AccessToken Type {TokenType} " +
-                    $"RefreshToken {!String.IsNullOrEmpty(RefreshToken)} " +
-                    $"IdToken {!String.IsNullOrEmpty(IdToken)} " +
-                    $"ClientInfo {ClientInfo} ");
-
-                withPii.AppendLine($"FamilyId: {FamilyId} WamAccountId {!string.IsNullOrEmpty(WamAccountId)}");
-                withoutPii.AppendLine($"FamilyId: {FamilyId} WamAccountId {!string.IsNullOrEmpty(WamAccountId)}");
+                withoutPii.AppendLine($"ExpiresIn: {ExpiresIn}");
+                withoutPii.AppendLine($"RefreshIn: {RefreshIn}");
+                withoutPii.AppendLine($"AccessToken returned: {!string.IsNullOrEmpty(AccessToken)}");
```
jennyf19, line 214: this is pii.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
jennyf19: nit: also might consider making it "AccessTokenType" to match the other one as well.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
jennyf19: Ignore this comment. Not applicable to the fix anyway.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
pmaytak: Which `the other one` are you referring to?

jennyf19: sorry, i'll work on being more clear -> like this `_requestParams.AuthenticationScheme.AccessTokenType` and actually call it `AccessTokenType` because that's what it is. so this would be `AccessTokenTypeMissing` - just a thought.

pmaytak: Thanks for clarifying. Updated.

pmaytak:
Also added a unit test.

tests/Microsoft.Identity.Test.Unit/TokenClientTests.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using OpenQA.Selenium.Remote;
+
+namespace Microsoft.Identity.Test.Unit
+{
+    [TestClass]
+    public class TokenClientTests : TestBase
+    {
+        [TestMethod]
+        public async Task SendTokenRequest_MissingAccessTokenTypeInResponse_Throws_Async()
```
jennyf19, line 19: I would move this to [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationSchemeTests.cs#L93)

jennyf19 (approved):
one comment on the test location. :shipit:

tests/Microsoft.Identity.Test.Unit/TokenClientTests.cs
pmaytak: 👍

neha-bhargava (approved):

PR #2641, opened 2021-05-18
 @bgavrilMS Extend /organizations workaround to try to fix #2615

### LINKED ISSUE #2615: [Bug] WAM doesn't properly work if multiple work accounts are configured on the same machine
**Logs and Network traces**
[App.log](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/files/6453108/App.log)

Microsoft.Identity.Client 4.30.1

**Platform**
UWP

* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Auth
    * [ ] Username Password
* Web App 
    * [ ] Authorization code
    * [ ] OBO
* Daemon App 

Other? - please describe;

The app is in production, and I have enabled WAM as authentication method

**Repro**

1. Setup a Windows 10 machine with multiple work accounts configured (AAD Registered device)
2. Use the following code to perform silent authentication using the work account configured in Windows 10.

```csharp
private async Task<AuthenticationResult> _AcquireTokenInteractive(IEnumerable<string> defaultScopes, IEnumerable<string> extraScopes, bool useOsAccount = false)
{
    AuthenticationResult authenticationResult;
    IAccount accountToLogin;
    var accounts = await _client.GetAccountsAsync();
    if (accounts.Any())
        accountToLogin = accounts.FirstOrDefault();
    else
        accountToLogin = Microsoft.Identity.Client.PublicClientApplication.OperatingSystemAccount;
 
    
    var acquireTokenInteractiveBuilder = _client.AcquireTokenInteractive(defaultScopes)
                                            .WithExtraScopesToConsent(extraScopes);
    if (useOsAccount)
        acquireTokenInteractiveBuilder.WithAccount(accountToLogin);
 
    authenticationResult = await acquireTokenInteractiveBuilder.ExecuteAsync();
    return authenticationResult;
}

```

**Expected behavior**
The application fails to silently login since there are multiple work accounts, so WAM uses the integrated Windows picker to enable the user to choose one of the available accounts.

image

**Actual behavior**
Instead of WAM, the web picker is displayed. However, the web picker doesn't work. It correctly shows the work accounts connected to Windows but, when you try to choose one of them, login always fails with the following error:

image

**Additional context/ Logs / Screenshots**
The problem happens only if multiple work accounts are configured. If a single one is present, the application can silently login with that account without issues.
Additionally, if you don't pass Microsoft.Identity.Client.PublicClientApplication.OperatingSystemAccount as account to use for login, WAM will properly kick in, prompting the account to choose with the native Windows picker (which works, unlike the web one). However, in my case, this isn't a feasible solution, because it would make lose the opportunity of enabling silent logging for all the other supported scenarios (single work account or device fully AAD joined).

bgavrilMS:


tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
```diff
@@ -601,6 +601,55 @@ public async Task ATI_WithDefaultUser_Async()
...
+                    .Returns(Task.FromResult(webTokenRequest));
+
+                var webTokenResponseWrapper = Substitute.For<IWebTokenRequestResultWrapper>();
+                webTokenResponseWrapper.ResponseStatus.Returns(WebTokenRequestStatus.Success);
+                var webTokenResponse = new WebTokenResponse();
+                webTokenResponseWrapper.ResponseData.Returns(new List<WebTokenResponse>() { webTokenResponse });
+
+                _wamProxy.RequestTokenForWindowAsync(Arg.Any<IntPtr>(), webTokenRequest).
+                    Returns(Task.FromResult(webTokenResponseWrapper));
+                _aadPlugin.ParseSuccessfullWamResponse(webTokenResponse, out _).Returns(_msalTokenResponse);
```
trwalke, line 635: what does (out underscore) mean? is it just a variable?

trwalke (approved):
Just one question, otherwise LGTM

trwalke:
Ah, didn't realize this was a draft at first

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
bgavrilMS: Good catch. It's called a Discard, a concept introduced in C# 7. https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards

It's simple - if you don't care about a variable, you can use underscore and the compiler will not complain about things like "you assign a variable but you never use it".

bgavrilMS:
@trwalke - thx for the review, it's now a PR.

neha-bhargava (approved):

pmaytak (approved):

pmaytak (approved):
👍

Issue #2444, opened 2021-03-03
[Bug] [M] MSAL.NET fails if B2C user flows/policies have a name containing a dot (.)


endintiers:
**Which Version of MSAL are you using ?**
MSAL 4.27

**Platform**
Xamarin/UWP

**What authentication flow has the issue?**
Other? - please describe;
AcquireTokenSilent/Identity Token Cache

**Is this a new or existing app?**
c. This is a new app or experiment

**Repro**
Create a B2C flow called **{a}.signupsignin**
AcquireTokenInteractive using that flow - works. 
... later ...
GetAccountsAsync, which returns a valid Account
call IPublicClientApplication.AcquireTokenSilent with that account

**Expected behavior**
A user token is acquired from the token cache for the account.

**Actual behavior**
throws Microsoft.Identity.Client.MsalClientException: 'Returned user identifier does not match the sent user identifier when saving the token to the cache. '

**Possible Solution**
I assume the problem is the extra '.' in the homeAccountId.Identifier which looks like: {uid}-{a}.signupsignin.{tenantId}

I notice that in various places in the library, for example AccountId.ParseFromString, there is code such as 
`string[] elements = str.Split('.');`
And then an assumption that there will only be 2 elements... 

Rather than changing the split character (in many places) - which would probably cause other problems, or trying to deal with an array of length 'n', I would suggest:
1) Documenting: "Don't use '.' in your B2C flow/Policy names"
2) Checking for extra '.'s in homeAccountId.Identifiers when first seeing an Account - probably only needed in 3(ish) places? and throwing an exception (referencing the documentation).

endintiers:
An alternative could be to unseal "Account".
In which case: `string userIdentifier = ((Microsoft.Identity.Client.Account)account).HomeAccountId.Identifier`.
would work, so no need to split the string?

bgavrilMS:
Looks more like a bug to me @jmprieur

bgavrilMS:
See if we can use better parsing on AccountId. Otherwise, we can disallow dots.

jmprieur:
@bgavrilMS : did we decide to close issues instead of adding the "Fixed" label?

bgavrilMS:
Yes, I did not hear any objections.

PR #2648, opened 2021-05-20
Handle case when b2c profile contains a '.'

neha-bhargava:
#2444 

Try to handle the scenario when the b2c profile have a '.' character.


src/client/Microsoft.Identity.Client/AccountId.cs
```diff
@@ -63,14 +63,14 @@ internal static AccountId ParseFromString(string str)
...
+            int lastIndexOfDot = str.LastIndexOf('.');
 
-            if (elements.Length == 1)
+            if (lastIndexOfDot == -1)
             {
                 return new AccountId(str); //Account id is from ADFS; no . in the string
             }
 
-            return new AccountId(str, elements[0], elements[1]);
+            return new AccountId(str, str.Substring(0, lastIndexOfDot), str.Substring(lastIndexOfDot + 1));
```
neha-bhargava, line 73: This is the fix, splitting the string based on the last `.`

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -117,7 +117,7 @@ public async Task<IEnumerable<IAccount>> GetAccountsAsync(string userFlow)
             var accounts = await GetAccountsInternalAsync(ApiIds.GetAccountsByUserFlow).ConfigureAwait(false);
 
             return accounts.Where(acc =>
-                acc.HomeAccountId.ObjectId.Split('.')[0].EndsWith(
+                acc.HomeAccountId.ObjectId.EndsWith(
```
neha-bhargava, line 120: I don't see the reason for adding this split here as the ObjectId is obtained after splitting the homeAccountId with the `.`

src/client/Microsoft.Identity.Client/AccountId.cs
bgavrilMS, line 73: Very cool!

Can you add more tests around these string operations? Lots of "off by 1" errors possible. This method is static so should be very easy to test.

bgavrilMS (approved):
LGTM, please add a few unit tests to cover the parsing

src/client/Microsoft.Identity.Client/AccountId.cs
neha-bhargava: Added tests for parsing

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AccountIdTest.cs
```diff
@@ -34,5 +34,75 @@ public void AccountIdPublicApi()
...
+
+        [TestMethod]
+        [Description("Test the parsing of HomeAccountId with no dot")]
+        public void ParseFromStringNoDotTest()
+        {
+            AccountId accountId = AccountId.ParseFromString("uid");
+
+            Assert.AreEqual("uid", accountId.ObjectId);
+            Assert.AreEqual(null, accountId.TenantId);
+        }
```
pmaytak, line 46: The tests can be condensed into one test with adding `[DataRow]` attribute like in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/cfa2b36b9b5732d182bf6e31bb8a9fbc1127ab3a/tests/Microsoft.Identity.Test.Unit/AppConfigTests/PublicClientApplicationBuilderTests.cs#L383

pmaytak (approved):
👍

Issue #2560, opened 2021-04-19
[Bug] WAM interactive dialog (the Account Picker) is immediately canceled when shown in an elevated app


BenPryor:
**Which Version of MSAL are you using ?**
MSAL 4.29.0

**Platform**
.NET 5

**What authentication flow has the issue?**
Desktop / Mobile: Interactive

**Is this a new or existing app?**
This is a new app or experiment

**Repro**
C# console app, .NET 5, target framework `net5-windows10.0.17763.0`.
```csharp
static async Task Main()
{
    IPublicClientApplication app = PublicClientApplicationBuilder
        .Create("my-client-id")
        .WithAuthority("https://login.microsoftonline.com/my-tenant-id")
        .WithExperimentalFeatures()
        .WithBroker()
        .Build();

    await app.AcquireTokenInteractive(new string[] { "my-resource-id/.default" })
        .ExecuteAsync();
}
```

**Expected behavior**
The WAM interactive dialog (account picker) should be shown.

**Actual behavior**
If the console app is run elevated, the WAM dialog briefly flashes and is immediately canceled (MSAL returns `authentication_canceled`). If the console app is run non-elevated, the WAM dialog is shown as expected.

**Additional context/ Logs / Screenshots**
I see this issue on .NET framework, .NET Core 3.1, and .NET 5.

In case this is a bug in WAM itself, I am reproing this on Win10 20H2.

bgavrilMS:
A workaround has been provided by the Windows team: 

```
This is one of the native APIs that is historically known to have problems calling from .Net. The issue is that this needs to be called before COM remoting is initialized in a process or it will happen implicitly, and there are things CLR may do early in the process lifetime that cause this. This is mentioned (with some details inaccurate) on the pinvoke.net page:
You shouldn't call CoInitializeSecurity from managed code.  http://pinvoke.net/default.aspx/ole32/CoInitializeSecurity.html 

Most of what you can do with CoInitializeSecurity, including access permissions as needed here, can also be done with an AppID key for the exe (see [here ](https://docs.microsoft.com/en-us/windows/win32/com/appid) and [here](https://docs.microsoft.com/en-us/windows/win32/com/appid-key)). Note that for COM to read this from an elevated process it must be in HKLM.
```

mjcheetham:
[...]

PR #2656, opened 2021-05-25
WAM admin process: better error + helper methods

pmaytak:
Related to issue #2560.

Added a descriptive error when Account Picker returns no accounts and the process is elevated.
Added related Windows native methods to check for elevated user and to set process security.
Added Troubleshooting section to [WAM wiki](https://aka.ms/msal-net-wam).

**Testing**
Tested workaround with NetCoreWinFormsWAM dev app in net5 and netcoreapp3.1 - works. Does not work with net472 dev app. Breaks when trying to set the process security, since it was already set by CLR (which is know issue with this workaround).


src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/WindowsNativeUtils.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            if (result != 0)
+            {
+                throw new MsalClientException(MsalError.InitializeProcessSecurityError, MsalErrorMessage.InitializeProcessSecurityError($"0x{result:x}"));
+            }
+        }
+
+        [DllImport("shell32.dll")]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        private static extern bool IsUserAnAdmin();
```
pmaytak, line 42: Another way to do this is `new WindowsPrincipal(WindowsIdentity.GetCurrent()).IsInRole(WindowsBuiltInRole.Administrator)`; however, it's [not available](https://apisof.net/catalog/7716df7f-41fa-ebc8-1c5a-bd4d110b7ed1) in netcoreapp2.1.

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
```diff
@@ -335,6 +333,12 @@ private static void AddPromptToRequest(Prompt prompt, bool isForceLoginPrompt, W
 
                 if (accountProvider == null)
                 {
+#if !WINDOWS_APP
+                    if (WindowsNativeUtils.IsElevatedUser())
+                    {
+                        throw new MsalClientException(MsalError.AuthenticationFailedWamElevatedProcess, MsalErrorMessage.AuthenticationFailedWamElevatedProcess);
```
pmaytak, line 339: There are two scenario's when this will be hit:
- App is elevated and workaround was not used - Account Picker crashes.
- App is elevated and workaround was used, but the user manually cancelled the dialog.

bgavrilMS: Hmm, this is not ok. "auth_cancelled" is a very common use case that apps handle. We cannot change this behavior. I think we can only change the error message to say "User cancelled auth or x y z"

bgavrilMS (changes requested):
We can't change the auth_cancelled behavior. It's one of the few error codes that app developers actually handle. We can change the error message only.

src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/WindowsNativeUtils.cs
```diff
@@ -0,0 +1,90 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Identity.Client.Platforms.Features.DesktopOs
```
pmaytak, line 7: @bgavrilMS  Hmm should I put these into a different namespace? Do we have one for public helpers?

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
pmaytak: Changed to keep the same error code, but different message.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -133,7 +133,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <Compile Include="$(PathToMsalSources)\Platforms\netstandard13\**\*.cs" LinkBase="Platforms\netstandard13" />
-    <Compile Include="$(PathToMsalSources)\Platforms\Features\DesktopOs\**\*.cs" />
+    <Compile Include="$(PathToMsalSources)\Platforms\Features\DesktopOs\**\*.cs" Exclude="$(PathToMsalSources)\Platforms\Features\DesktopOs\WindowsNativeUtils.cs" />
```
bgavrilMS, line 136: Why do you need to exclude this?

bgavrilMS (approved):

neha-bhargava (approved):

bgavrilMS (approved):

PR #2664, opened 2021-05-27
Fix some warnings

bgavrilMS:
Fix some warnings

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByIntegratedWindowsAuthParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByIntegratedWindowsAuthParameterBuilder.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -499,25 +498,38 @@ internal virtual void Validate()
...
         }
 
         #region Authority
-        private void CreateAuthorityInfoFromEnums()
+        private void ResolveAuthority()
         {
             if (Config.AuthorityInfo != null)
             {
-                return;
+                if (!string.IsNullOrEmpty(Config.TenantId) &&
```
trwalke, line 511: This looks like the same changes made in this PR here?
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2666/files

Should this be here?

SameerK-MSFT: +1

PR #2666, opened 2021-05-27
Fix for #2543 - take into account WithTenant()

### LINKED ISSUE #2543: [Bug] [M] WithTenant ignored, resulting in CA failures for multi-tenant apps
MSAL 4.29 with WAM on UWP

## Repro

```csharp
var options = new PublicClientApplicationOptions();
             options.ClientId = token.ClientId;                            // “110797d6-4a5e-4e58-a06d-f1bf3f3a8069”

            var app = PublicClientApplicationBuilder
                .CreateWithApplicationOptions(options)
                .WithTenantId(token.TenantId)                             //The tenantId passed here (“3cfc49f9-956e-4e4e-a1b6-e03368c2e448”)
                .WithAuthority(token.AuthorityEndpoint)          //Set to the “common” endpoint “https://login.microsoftonline.com/common”
               .WithBroker(true)
              
               .Build();
```

Expected: AcquireToken* should work for user 
Actual: AcquireToken* fails because the authority passed to WAM ends in "common", instead of the tenanted one, causing issues with Conditional Access

**Workaround**

Use a tenanted authority instead, i.e. `https://login.microsoftonline.com/3cfc49f9-956e-4e4e-a1b6-e03368c2e448`

bgavrilMS:
Fixes #2543  .

**Testing**

unit

**Performance impact**

Extra logic added only in corner-case scenario.

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Instance;
 using Microsoft.Identity.Client.Instance.Discovery;
 using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
 using Microsoft.Identity.Client.Utils;
@@ -467,9 +468,7 @@ namespace Microsoft.Identity.Client
             if (string.IsNullOrWhiteSpace(Config.ClientId))
             {
                 throw new MsalClientException(MsalError.NoClientId, MsalErrorMessage.NoClientIdWasSpecified);
-            }
-
-            CreateAuthorityInfoFromEnums();
+            }         
 
             //ADFS does not require client id to be in the form of a GUID.
             if (Config.AuthorityInfo?.AuthorityType != AuthorityType.Adfs && !Guid.TryParse(Config.ClientId, out _))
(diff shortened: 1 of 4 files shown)


src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -499,25 +498,38 @@ internal virtual void Validate()
...
         }
 
         #region Authority
-        private void CreateAuthorityInfoFromEnums()
+        private void ResolveAuthority()
         {
             if (Config.AuthorityInfo != null)
             {
-                return;
+                if (!string.IsNullOrEmpty(Config.TenantId) &&
```
trwalke, line 511: Does this also cover application configurations like "PublicClientApplicationOptions"?

trwalke:
Do we need a manual test to ensure WAM gets the proper authority?

trwalke (approved):

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -793,6 +793,11 @@ public static class MsalError
         /// </summary>
         public const string AuthorityTypeMismatch = "authority_type_mismatch";
 
+        /// <summary>
+        /// <para>What happens?</para>You specified a tenant twice - once in WithAuthority() and once using WithTenant()
+        /// <para>Mitigation</para>Specify the tenant once.
+        /// </summary>
+        public const string AuthorityTenantSpecifiedTwice = "authority_type_mismatch";
```
SameerK-MSFT, line 800: The value of the constant AuthorityTenantSpecifiedTwice is same as that of AuthorityTypeMismatch.

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/MsalError.cs
bgavrilMS: Good catch

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: Seems so. I added a test to it.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -499,25 +498,38 @@ internal virtual void Validate()
...
+                    Config.AuthorityInfo.AuthorityType == AuthorityType.Aad)
+                {
+                    AadAuthority aadAuthority = Authority.CreateAuthority(Config.AuthorityInfo) as AadAuthority;
+                    if (!aadAuthority.IsCommonOrganizationsOrConsumersTenant())
+                    {
+                        throw new MsalClientException(MsalError.AuthorityTenantSpecifiedTwice, "You specified a tenant twice - once in WithAuthority() and once using WithTenant().");
```
neha-bhargava, line 517: Do we also throw an exception when the tenant specified twice is same?

neha-bhargava (approved):

pmaytak (approved):

tests/Microsoft.Identity.Test.Unit/AppConfigTests/PublicClientApplicationBuilderTests.cs
```diff
@@ -594,6 +594,50 @@ public void WithClientCapabilitiesViaOptions()
...
+            var options = new PublicClientApplicationOptions();
+            options.ClientId = TestConstants.ClientId;            
+
+            var app1 = PublicClientApplicationBuilder
+                .CreateWithApplicationOptions(options)
+                .WithTenantId(TestConstants.TenantId)                            
+                .WithAuthority("https://login.microsoftonline.com/common")       
+               .Build();
+
+            Assert.AreEqual($"https://login.microsoftonline.com/{TestConstants.TenantId}/", app1.Authority);
```
pmaytak, line 610: nit: `$"https://login.microsoftonline.com/{TestConstants.TenantId}/"` can be changed to `TestConstants.AuthorityTenant`

Issue #2623, opened 2021-05-13
[Feature Request] [2L 5S] OBO to perform refresh_token flow


bgavrilMS:
Change OBO as per https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?version=GBdev&path=%2FConfidentialClient%2FOBO.md

- [x] add user_assertion_hash to RT as well as AT [S]
- [ ] change OBO logic to perform refresh_token flow [L]
- [ ] in refresh_token flow, only use RT associated with user_assertion_hash [S]
- [x] E2E Test: OBO with expired AT and valid RT in cache [L]
- [x] E2E Test: OBO with expired AT and missing RT in cache [S]
- [x] E2E Test: OBO with expired AT and RT not associated with user_assertion_hash in cache [S]
- [ ] Test that this is not a breaking change, i.e. GetAccounts / AcquireTokenSilent behave as expected [S]

See also: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2558 for the scenario

bgavrilMS:
Will resolve issues #2447 #2558

jmprieur:
See also https://github.com/AzureAD/microsoft-identity-web/issues/441

PR #2667, opened 2021-05-27
Add refresh token flow to OBO

neha-bhargava:
Fixes # .
#2623 

- [x] add user_assertion_hash to RT as well as AT [S]
- [x] change OBO logic to perform refresh_token flow [L]
- [x] in refresh_token flow, only use RT associated with user_assertion_hash [S]

**Testing**
- [x] Unit test: test request headers for OBO requests
- [x] E2E Test: OBO with expired AT and valid RT in cache [L]
- [x] E2E Test: OBO with expired AT and missing RT in cache [S]
- [x] E2E Test: OBO with expired AT and RT not associated with user_assertion_hash in cache [S]
- [x] Unit test to ensure the OBO flow is using the refresh token when the access token is expired. [S]

**Performance impact**
- [ ] Performance tests for OBO


tests/Microsoft.Identity.Test.Unit/Resources/ExpectedTokenCache.json
```diff
@@ -73,33 +73,37 @@
...
-      "environment": "env_1",
-      "client_id": "d3adb33f-c0de-ed0c-c0de-deadb33fc0d3",
-      "secret": "access_token_secret",
-      "credential_type": "RefreshToken",
+        "home_account_id": "my-uid.my-utid",
+        "environment": "env_1",
+        "client_id": "d3adb33f-c0de-ed0c-c0de-deadb33fc0d3",
+        "secret": "access_token_secret",
+        "credential_type": "RefreshToken",
+        "user_assertion_hash": "assertion hash"
```
bgavrilMS, line 81: This field is optional, can you have one of the RTs without it?

bgavrilMS:
LGTM so far

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -85,18 +85,63 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
-                }
+                return await RefreshRtOrFetchNewAccessTokenAsync(cancellationToken).ConfigureAwait(false);
             }
             catch (MsalServiceException e)
             {
                 return HandleTokenRefreshError(e, msalAccessTokenItem);
             }
         }
 
+        private async Task<AuthenticationResult> RefreshRtOrFetchNewAccessTokenAsync(CancellationToken cancellationToken)
```
bgavrilMS, line 96: if there is a lot code duplication with AcquireTokenSilent, let's try to find a way to reduce it

neha-bhargava: Yes, there are some things different here from the silent flow, in silent flow we fail if the rt cannot be refreshed with UIRequiredException. Here we don't fail. I was also thinking of how to reuse the code. Maybe expose this part of code in some util.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -342,7 +346,7 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
                 return null;
             }
 
-            MsalAccessTokenCacheItem msalAccessTokenCacheItem = GetSingleResult(requestParams, finalList);
+            MsalAccessTokenCacheItem msalAccessTokenCacheItem = (MsalAccessTokenCacheItem)GetSingleResult(requestParams, finalList);
```
trwalke, line 349: I think it might be better to create two seperate methods that call into GetSingleResult but with the correct cache item type to avoid having to cast each time the method is used. 
Example: MsalAccessTokenCacheItem GetSingleAccessTokenResult(IReadOnlyList<MsalAccessTokenCacheItem> filteredItems....)

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -581,23 +585,22 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
...
 
-            if (candidateRt != null)
-                return candidateRt;
+                return msalRefreshTokenCacheItem;
+            }
 
             requestParams.RequestContext.Logger.Info("Checking ADAL cache for matching RT. ");
+            
 
             // ADAL legacy cache does not store FRTs
```
trwalke, line 605: nit: extra space above

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: Added

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -402,7 +406,7 @@ private static IEnumerable<MsalAccessTokenCacheItem> FilterByTokenType(Authentic
             string requestTenantId = requestParams.Authority.TenantId;
             bool filterByTenantId = true;
 
-            if (requestParams.UserAssertion != null) // OBO
+            if (requestParams.ApiId == ApiEvent.ApiIds.AcquireTokenOnBehalfOf) // OBO
             {
```
jennyf19, line 410: nit: are we sure this is getting set? & if we check on the UserAssertion, we are sure it's not null

jennyf19:
>                                 item.UserAssertionHash.Equals(requestParams.UserAssertion.AssertionHash, StringComparison.OrdinalIgnoreCase),

this could be null now if the above logic changes for the if statement

---
Refers to: src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs:413 in 92ed27f. [](commit_id = 92ed27f24009f1b1574d4b5fdc5522be4819bfc8, deletion_comment = False)

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -44,9 +44,9 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             {
                 // look for access token in the cache first.
                 // no access token is found, then it means token does not exist
-                // or new assertion has been passed. We should not use Refresh Token
-                // for the user because the new incoming token may have updated claims
-                // like MFA etc.
+                // or new assertion has been passed. 
+                // Look for a refresh token, if refresh token is found perform refresh token flow.
```
jennyf19, line 48: so...how is the above comment getting addressed?

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Cache.Items;
+using Microsoft.Identity.Client.Internal.Requests;
+using Microsoft.Identity.Client.OAuth2;
+
+namespace Microsoft.Identity.Client.Internal
+{
+    internal static class SilentRequestHelper
```
jennyf19, line 13: how is this different from the one in the CacheSilentStrategy class?

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
```diff
@@ -221,24 +221,6 @@ private async Task<MsalTokenResponse> TryGetTokenUsingFociAsync(CancellationToke
             return null;
         }
 
-        private async Task<MsalTokenResponse> RefreshAccessTokenAsync(MsalRefreshTokenCacheItem msalRefreshTokenItem, CancellationToken cancellationToken)
```
jennyf19, line 224: i don't know why this class was changed. then, you don't need to pass a bunch of parameters around.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -628,7 +634,66 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
...
+
+            authResult = await confidentialApp.AcquireTokenOnBehalfOf(s_scopes, userAssertion)
+                .ExecuteAsync(CancellationToken.None)
+                .ConfigureAwait(false);
+
+            MsalAssert.AssertAuthResult(authResult, user);
+            Assert.AreEqual(atHash, userCacheRecorder.LastAfterAccessNotificationArgs.SuggestedCacheKey);
+
+#pragma warning disable CS0618 // Type or member is obsolete
+            await confidentialApp.GetAccountsAsync().ConfigureAwait(false);
```
jennyf19, line 653: >GetAccountsAsync [](http://example.com/codeflow?start=34&length=16)

this can't be used w/cca

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -628,7 +634,66 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
...
+#pragma warning restore CS0618 // Type or member is obsolete
+            Assert.IsNull(userCacheRecorder.LastAfterAccessNotificationArgs.SuggestedCacheKey);
```
jennyf19, line 655: what is this test testing?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -647,6 +712,71 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
...
+            Assert.IsTrue(userCacheRecorder.LastAfterAccessNotificationArgs.HasTokens);
+            Assert.AreEqual(atHash, userCacheRecorder.LastAfterAccessNotificationArgs.SuggestedCacheKey);
+            Assert.AreEqual(TokenSource.IdentityProvider, authResult.AuthenticationResultMetadata.TokenSource);
+
+            TokenCacheHelper.ExpireAccessTokens(confidentialApp2.UserTokenCacheInternal);
+            TokenCacheHelper.UpdateUserAssertions(confidentialApp2);
+
+            //Should perform OBO flow since the access token and the refresh token contains the wrong user assertion hash
+            authResult = await confidentialApp2.AcquireTokenOnBehalfOf(s_scopes, userAssertion)
+                .ExecuteAsync(CancellationToken.None)
```
jennyf19, line 771: if the hash is wrong it shouldn't work, this would mean i could use any userAssertion

tests/Microsoft.Identity.Test.Unit/CacheTests/CacheSerializationTests.cs
```diff
@@ -57,15 +57,28 @@ private MsalAccessTokenCacheItem CreateAccessTokenItem()
...
             return new MsalRefreshTokenCacheItem
             {
                 ClientId = TestConstants.ClientId,
                 Environment = "env",
                 HomeAccountId = TestConstants.HomeAccountId,
                 Secret = "access_token_secret",
-                RawClientInfo = string.Empty, 
+                RawClientInfo = string.Empty,
+                UserAssertionHash = "assertion hash"
             };
```
jennyf19, line 82: nit" ~"assertion_hash"`

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
jennyf19: because if you change the condition, and the assertion is null, you'll get a null ref later.

src/client/Microsoft.Identity.Client/Cache/Items/MsalRefreshTokenCacheItem.cs
```diff
@@ -52,6 +52,8 @@ internal MsalRefreshTokenCacheItem()
         /// </summary>
         public string FamilyId { get; set; }
 
+        internal string UserAssertionHash { get; set; }
```
trwalke, line 55: This should be moved to the base class and removed from the child classes. This will allow us to simply code elsewhere. There are only 3 child classes so we should be ok

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -621,6 +633,45 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
...
+                                "Filtering by home account id");
+            }
+
+            rtCacheItems = rtCacheItems.FilterWithLogging(item =>
+                            string.Equals(item.FamilyId ?? string.Empty,
+                            familyId ?? string.Empty, StringComparison.OrdinalIgnoreCase),
+                            requestParams.RequestContext.Logger,
+                            "Filtering by family id");
+
+            // If family id have a value that means we are looking for a refresh token for foci.
```
trwalke, line 663:
```suggestion
            // If family id has a value, it means we are looking for a refresh token for foci.
```

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
trwalke: This was done to remove code duplication as the silent request and the OBO request will use very similar code going forward. I agree that passing additional variables around is not ideal but it might be better than duplicate code. @neha-bhargava do you think you can figure out a way to reduce the number of parameters being passed around. If not then I guess we can take it as is.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -44,9 +44,9 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+                // If a refresh token is not found, fetch new access token for OBO
```
jmprieur, line 49: How do we do that?  If we don't have a refresh token? I'm a bit confused? do you mean if the access token is not found, fetch a new access token, @neha-bhargava ?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -647,6 +712,71 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
...
+            var confidentialApp2 = ConfidentialClientApplicationBuilder
+                .Create(confidentialClientID)
+                .WithAuthority(new Uri(oboHost + authResult.TenantId), true)
+                .WithClientSecret(secret)
+                .WithTestLogging()
+                .BuildConcrete();
+
+            TokenCacheHelper.ExpireAndSaveAccessToken(confidentialApp2.UserTokenCacheInternal, atItems.FirstOrDefault());
+
+            //Should perform OBO flow since the access token is expired and the refresh token does not exist
```
jmprieur, line 754: The bit I don't understand here is that if the access token is expired, then it's certainly because the token that was passed in the web API (that is therefore in the **userAssertion**) has itself expired. In other words, I don't know when this could happen in real life? Or is it testing "no cache" ?

What we'd want to test (and I don't know how difficult this is) is:
- We call OBO with a particular user assertion.
- we wait for 1h until the user assertion / OBO token expires
- we then call OBO again with this expired user assertion, which is, in this case only used as a key to find the refresh token, and we get a new OBO token. Note that given OBO is used in a web API, the user assertion would come necessarily from the web api itself part of a long lived process, as if it was coming from a client of the Web API (calling the web API with an expired token), the incoming token would never have been validated by the web API (as it was expired)

jmprieur (commented):
Thanks @neha-bhargava 
I left a couple of comments, I'm not sure about one of the integration tests (and therefore the behavior)

src/client/Microsoft.Identity.Client/Cache/Items/MsalRefreshTokenCacheItem.cs
neha-bhargava: Since `MsalIdTokenCacheItem` won't have `UserAssertion` I did not add it to the base class.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
neha-bhargava: Earlier we did not look for the refresh token for obo and instead if AT is not found would fetch a new AT. Now with the new [API review](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?version=GBdev&path=%2FConfidentialClient%2FOBO.md&_a=preview) we look for a refresh token filtered by the user assertion in the request. If a refresh token is found we use it. If a RT is not found we fetch the access token.

neha-bhargava: I will update the comment to make it more clear

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
neha-bhargava: We land here when AT is not found in the cache. Earlier we did not look for RT and just fetched a fresh AT. Now, we look for RT in the cache with same UserAssertion as request and use it if it is found. If not found we fetch a fresh AT

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
neha-bhargava: The methods are moved from here to the [SilentRequestHelper.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/92ed27f24009f1b1574d4b5fdc5522be4819bfc8/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs#L13). These methods were also used for OBO refresh flow so to reduce duplicate code I moved them.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs
neha-bhargava: This is a helper class I added to not introduce duplicate code. Since OBO refresh flow required these 2 methods which are similar to the silent flow.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: Same condition is used below [see](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/92ed27f24009f1b1574d4b5fdc5522be4819bfc8/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L440). `UserAssertion` here is used to check for OBO flow. Discussed it with @bgavrilMS. I can revert it if required.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
jennyf19: And what if the refresh token is no longer valid, ex. something on the resource side has changed, how do we revoke the RT to force a new AT/RT pair?

trwalke: @jmprieur the scenario you describe is tested here
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/161411bf059185d8b4e20d7cae0d6c3d29b2da13/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs#L728-L734

The AT is expired by updating the exp on the MsalAccessTokenCacheItem and then the RT is used to refresh the OBO tokens.

This specific section here is testing this scenario which is a requirement from the task
"E2E Test: OBO with expired AT and missing RT in cache [S]"

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/161411bf059185d8b4e20d7cae0d6c3d29b2da13/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs#L743-L757

If you are saying that it is not possible to reach a state where the AT is expired and no RT is available then perhaps the requirement is not needed.

@bgavrilMS @neha-bhargava

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
trwalke: If the hash is wrong I think it should ignore the tokens in the cache and try to fetch new tokens from the endpoint with the new userAssertion. If the userAssertion is determined to not be valid by AAD then no tokens will be acquired.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
trwalke: This was a check used by a different OBO test in same the file. is was accidently copied over. not need. I'm am not sure why it was placed in the other test however.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
trwalke: see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2667/files#r648732899

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -44,9 +44,10 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             {
```
bgavrilMS, line 44: The condition at line 43 should be probably more like: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/161411bf059185d8b4e20d7cae0d6c3d29b2da13/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs#L46

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -86,18 +87,34 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            // If a refresh token is not found, fetch a new access token
+            if (appRefreshToken == null)
+            {
+                AuthenticationRequestParameters.RequestContext.Logger.Verbose("[OBO request] No Refresh Token was found in the cache. Fetching OBO token from ESTS");
+
+                var result = await FetchNewAccessTokenAsync(cancellationToken).ConfigureAwait(false);
+                return result;
+            }
+
+            var msalTokenResponse = await SilentRequestHelper.RefreshAccessTokenAsync(appRefreshToken, cancellationToken, this, AuthenticationRequestParameters)
```
bgavrilMS, line 112: nit: by convention, CancellationToken is always last param.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -86,18 +87,34 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
             }
         }
 
+        private async Task<AuthenticationResult> RefreshRtOrFetchNewAccessTokenAsync(CancellationToken cancellationToken)
+        {
+            // Look for a refresh token
+            MsalRefreshTokenCacheItem appRefreshToken = await CacheManager.FindRefreshTokenAsync().ConfigureAwait(false);
+
+            // If a refresh token is not found, fetch a new access token
+            if (appRefreshToken == null)
```
bgavrilMS, line 104: nit: code is a bit more readable if you handle the case where you get a refresh token back first 

`if (appRefreshToken!=null) { ... } `

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
bgavrilMS: There are more params because the requests use class inheritance and some common code is in the base class. One of the reasons I don't like class inheritance.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: We are sure that this param (ApiId) is always is set and we make this kind of verification is many other places. 

I think the code is more readable like this. In fact, UserAssertion type is used in IWA and ROPC as well, but it is not set in the request params, so this code path is followed exclusively in OBO. It took me 5 min to figure this out :( since we are creating a UserAssertion in ~7 places.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -497,9 +501,19 @@ private MsalAccessTokenCacheItem FilterByExpiry(MsalAccessTokenCacheItem msalAcc
             return null;
         }
 
-        private static MsalAccessTokenCacheItem GetSingleResult(
+        private static MsalAccessTokenCacheItem GetSingleAccessTokenResult(AuthenticationRequestParameters requestParams, IReadOnlyList<MsalAccessTokenCacheItem> filteredItems)
+        {
+            return (MsalAccessTokenCacheItem) GetSingleResult(requestParams, filteredItems);
```
bgavrilMS, line 506: This code is perf sensitive, so let's avoid upcasting.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -497,9 +501,19 @@ private MsalAccessTokenCacheItem FilterByExpiry(MsalAccessTokenCacheItem msalAcc
...
+        }
+
+        private static MsalRefreshTokenCacheItem GetSingleRefreshTokenResult(AuthenticationRequestParameters requestParams, IReadOnlyList<MsalRefreshTokenCacheItem> filteredItems)
+        {
+            return (MsalRefreshTokenCacheItem) GetSingleResult(requestParams, filteredItems);
```
bgavrilMS, line 511: I think you're introducing a change here that should not be introduced. If 2 AT are found by filtering, we throw because there's smth wrong in MSAL. If 2 RTs are found, we just pick one, we don't throw.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: We are sure that this param (ApiId) is always is set and we make this kind of verification is many other places.

I think the code is a bit more readable like this. UserAssertion is also used in IWA and ROPC , but not for cache operations!! This code path is followed exclusively in OBO. It took me 5 min to figure this out :( since we are creating a UserAssertion object in ~7 places, but only OBO sets it in the requestParams. 

So either we say - if this request has a UserAssertion, then filter by it or we say "in case of OBO, filter by user assertion". IWA and ROPC use UserAssertiom, but do not fitler by it. Hence it makes a bit more sense to have the code as Neha changed it.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -621,6 +633,45 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
...
+            IEnumerable<MsalRefreshTokenCacheItem> rtCacheItems,
+            string familyId)
+        {
+            if (requestParams.ApiId == ApiEvent.ApiIds.AcquireTokenOnBehalfOf) // OBO
+            {
+                rtCacheItems = rtCacheItems.FilterWithLogging(item =>
+                                !string.IsNullOrEmpty(item.UserAssertionHash) &&
+                                item.UserAssertionHash.Equals(requestParams.UserAssertion.AssertionHash, StringComparison.OrdinalIgnoreCase),
+                                requestParams.RequestContext.Logger,
+                                "Filtering by user assertion id");
```
bgavrilMS, line 647: Add the `requestParams.UserAssertion.AssertionHash` to the log message

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -621,6 +633,45 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
...
+            {
+                rtCacheItems = rtCacheItems.FilterWithLogging(item => item.HomeAccountId.Equals(
+                                requestParams.Account.HomeAccountId?.Identifier, StringComparison.OrdinalIgnoreCase),
+                                requestParams.RequestContext.Logger,
+                                "Filtering by home account id");
+            }
+
+            rtCacheItems = rtCacheItems.FilterWithLogging(item =>
+                            string.Equals(item.FamilyId ?? string.Empty,
+                            familyId ?? string.Empty, StringComparison.OrdinalIgnoreCase),
```
bgavrilMS, line 659: nit: if at the beginning of this method you add `familyId = familyId ?? string.empty` then you wouldn't have to check so many times.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -621,6 +633,45 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
...
+            }
+            else
+            {
+                rtCacheItems = rtCacheItems.FilterWithLogging(item => item.HomeAccountId.Equals(
+                                requestParams.Account.HomeAccountId?.Identifier, StringComparison.OrdinalIgnoreCase),
+                                requestParams.RequestContext.Logger,
+                                "Filtering by home account id");
+            }
+
+            rtCacheItems = rtCacheItems.FilterWithLogging(item =>
```
bgavrilMS, line 657: nit

```csharp 
bool filterByFamilyId = string.IsNullOrEmtpy(familyId);
if (filterByFamilyId) 
{
      // filter by family ID
}
else
{
    // filter by client id
}

Comment (bgavrilMS) [reply src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs:657]: Appreciate how much more readable RT code filtering is! The fact that we should filter either by clientID or by familyID was very subtle / poorly done, this is much more clear.

File: src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs (hunk 22 above)
Comment (bgavrilMS) [reply src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs:647]: And maybe do that for the AT as well.

File: tests/Microsoft.Identity.Test.Common/TestCommon.cs [hunk 25]
```diff
@@ -111,7 +113,7 @@ public static IServiceBundle CreateDefaultAdfsServiceBundle()
                 commonParameters,
                 requestContext,
                 authority)
-            {                
+            {
```
bgavrilMS, line 116: no changes in this file?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -628,7 +634,7 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
                 .WithAuthority(new Uri(oboHost + authResult.TenantId), true)
                 .WithClientSecret(secret)
                 .WithTestLogging()
-                .Build();
+                .BuildConcrete();
```
bgavrilMS, line 637: needed?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
jennyf19: if the cache is partitioned, how can we have the wrong hash? I think either 1) the comment is not right or 2) the test is not right.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
jennyf19: we shouldn't have two RTs anyway, at least not in the partitioned cache. We have to be really careful just using one of them, as the claims could be different.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
jennyf19: that wasn't my point, sorry i wasn't clear. If we remove the check for UserAssertion being null, and then use the UserAssertion on line 413, that will throw a null ref if it's null: `requestParams.UserAssertion.AssertionHash`. so we need to check if it's null before we use it. that's all.

bgavrilMS: We don't allow OBO without a UserAssertion ... I don't think. We need to bail otherwise, because OBO does not have information about the user, i.e. our filtering would be completely screwed up if we had OBO and null UserAssertion.

Maybe log an error?

jennyf19: or we just check that it's done earlier in the code, i haven't checked and add a comment so we know for future reference that it's not null. just seems odd to be checking this late, i agree.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -649,6 +655,125 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
...
+
+            MsalAssert.AssertAuthResult(authResult, user);
+            Assert.AreEqual(atHash, userCacheRecorder.LastAfterAccessNotificationArgs.SuggestedCacheKey);
+
+            //Run OBO again. Should get token from cache
+            authResult = await confidentialApp.AcquireTokenOnBehalfOf(s_scopes, userAssertion)
+                .ExecuteAsync(CancellationToken.None)
+                .ConfigureAwait(false);
+
+            Assert.IsNotNull(authResult);
```
bgavrilMS, line 716: A vbetter assertion would be to verify the `grant_type` that was used when talking to the /token endpoint. You can do that by injecting `HttpSnifferClientFactory` which will record all HTTP traffic. To retrieve the request payload, you'll have to add smth like: 

```csharp
                    req.Content.LoadIntoBufferAsync().GetAwaiter().GetResult();
                    var payload = req.Content.ReadAsStringAsync().GetAwaiter().GetResult(); 
 
   //  payload has grant_type

```

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
bgavrilMS: @jmprieur - I don't think MSAL can't handle this scenario ? I think a middleware web api would need to hold on to the original assertion, even if it expires, and keep asking MSAL for a token. Are you suggesting an API like 

```AcquireTokenOnBehalfOf(s_scopes, userAssertionHashString)``` ?

Comment (bgavrilMS) [reply tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs:749]: @jennyf19 - MSAL would throw an MsalUiRequiredException, which the middleware web API needs to handle by sending a 401 or by stopping. 

If the upstream web api receives the 401, it should send a new user assertion A2. The middleware web api calls `AcquireTokenOBO(scope, A2)` which in turn: 

- looks for an access token associated with A2 (none is found)
- looks for a refresh token associated with A2 (none is found)
- calls ESTS with grant_type = OBO + A2
- ESTS responds with AT + RT
- MSAL stores the AT and RT in the cache and associates them with A2

If you use a partitioned cache, the old AT / RT are simply left to rot. If you'd be (wrongly) using a flat cache, the old AT / RT would be overriden.

File: tests/Microsoft.Identity.Test.Common/TestConstants.cs [hunk 28]
```diff
@@ -158,6 +158,43 @@ public static HashSet<string> s_scope
 y1NFZmB24rMoq8C+HPOpuVLzkwBr+qcCq7ry2326auogvVMGaxhHlwSLR4Q1OhRjKs8JctCk2+5Qs1NHfawa7jWHxdAK6cLm7Rv/c0ig2Jow7wRaI5ciAcEjX7
 m1t9gRT1mNeeluL4cZa6WyVXqXc6U2wfR5DY6GOMUubN5Nr1n8Czew8TPfab4OG37BuEMNmBpqoRrRgFnDzVtItOnhuFTa0=";
 
+        public const string FormattedAccessToken = @"eyJ0eXAiOiJKV1QiLCJub25jZSI6Im12OFZLdk11LUpiaThzRU14a20zWkVNaHoxRHcxUlFCdWowWnRzUWYtWGMiLCJhbGciOiJSUzI1NiIsIng1dCI6IkRxVXU4
```
bgavrilMS, line 161: What is this needed for?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
```diff
@@ -649,6 +655,125 @@ private async Task RunOnBehalfOfTestAsync(LabResponse labResponse)
...
+            TokenCacheHelper.ExpireAccessTokens(confidentialApp2.UserTokenCacheInternal);
+            TokenCacheHelper.UpdateUserAssertions(confidentialApp2);
+
+            //Should perform OBO flow since the access token and the refresh token contains the wrong user assertion hash
+            authResult = await confidentialApp2.AcquireTokenOnBehalfOf(s_scopes, userAssertion)
+                .ExecuteAsync(CancellationToken.None)
+                .ConfigureAwait(false);
+
+            Assert.IsNotNull(authResult);
+            Assert.IsNotNull(authResult.AccessToken);
```
bgavrilMS, line 770: Can you assert that we also have id token please?

tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs
```diff
@@ -0,0 +1,138 @@
...
+    [TestClass]
+    public class OboRequestTests
+    {
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            TestCommon.ResetInternalStaticCaches();
+        }
+
+        private MockHttpMessageHandler AddMockHandlerAadSuccess(MockHttpManager httpManager, string authority)
```
bgavrilMS, line 24: nit: probably exists already...

tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                AddMockHandlerAadSuccess(httpManager, TestConstants.AuthorityCommonTenant);
+
+                var cca = ConfidentialClientApplicationBuilder
+                                                         .Create(TestConstants.ClientId)
+                                                         .WithClientSecret(TestConstants.ClientSecret)
+                                                         .WithAuthority(TestConstants.AuthorityCommonTenant)
+                                                         .WithHttpManager(httpManager)
+                                                         .BuildConcrete();
+
+                UserAssertion userAssertion = new UserAssertion(TestConstants.FormattedAccessToken);
```
bgavrilMS, line 53: Do you really need an AT that looks like an AT? Do we do any sort of validation on it? I would think not (if we do, it's probably a bug)

bgavrilMS (commented):
Looks good, let's discuss during the meeting today if all scenarios are met. From coding perspective, I think it meets the spec. Minor comments around testing.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -86,18 +87,34 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+                return result;
+            }
+
+            var msalTokenResponse = await SilentRequestHelper.RefreshAccessTokenAsync(appRefreshToken, cancellationToken, this, AuthenticationRequestParameters)
+                .ConfigureAwait(false);
+            
+            return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
+        }
+
         private async Task<AuthenticationResult> FetchNewAccessTokenAsync(CancellationToken cancellationToken)
```
henrik-me, line 118: >FetchNewAccessTokenAsync [](http://example.com/codeflow?start=49&length=24)

This will also fetch the RT? Method name no longer accurate?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
jennyf19: sorry, what is it testing specifically? that link just goes to this PR.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
jennyf19: do we have a test for this:
- new UserAssertion(AT1) -> cache OBO_AT1/RT1 pair
- same UserAssertion(AT1) -> use cached OBO_RT1
- new UserAssertion(AT2 (just a claim change, same user)) -> cache OBO_AT2/RT2 pair
- same UserAssertion(AT2) -> cached OBO_RT2
- same UserAssertion(AT1) -> use cached OBO_RT1

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ConfidentialClientIntegrationTests.cs
trwalke: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2667#discussion_r648732899

Wrong link

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
neha-bhargava: There are no changes to this method. This is simply performing OBO to fetch a new token instead of using AT or RT from cache.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: That was the flow before where we returned the first or default. But I am not sure in which case will we have more than 1 RT with the filters. And if there are more than 1 RTs then will using any one of those be fine?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Took a look at the code path. We can add a null check here at the earliest. The parameter is required, but null can still be passed in, so we could get NPE in token cache. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f2866fc6884a8a4bb3a210dc1150b3f6088d3e96/src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs#L122-L129

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -587,21 +591,17 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
...
+            allRts = allRts.Where(
+                item => aliases.ContainsOrdinalIgnoreCase(item.Environment));
 
-            requestParams.RequestContext.Logger.Info("Refresh token found in the cache? - " + (candidateRt != null));
+            IReadOnlyList<MsalRefreshTokenCacheItem> finalList = allRts.ToList();
+            requestParams.RequestContext.Logger.Info("Refresh token found in the cache? - " + (finalList.Count() != 0));
 
-            if (candidateRt != null)
-                return candidateRt;
+            if (finalList.Count() > 0)
```
pmaytak, line 601: nit: `.Count` instead of `.Count()` (although since it's a list, in might not matter.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        internal static async Task<MsalTokenResponse> RefreshAccessTokenAsync(MsalRefreshTokenCacheItem msalRefreshTokenItem, RequestBase request, AuthenticationRequestParameters authenticationRequestParameters, CancellationToken cancellationToken)
+        {
+            authenticationRequestParameters.RequestContext.Logger.Verbose("Refreshing access token...");
+            await authenticationRequestParameters.AuthorityManager.RunInstanceDiscoveryAndValidationAsync().ConfigureAwait(false);
+
+            var msalTokenResponse = await request.SendTokenRequestAsync(GetBodyParameters(msalRefreshTokenItem.Secret), cancellationToken)
+                                    .ConfigureAwait(false);
+
+            if (msalTokenResponse.RefreshToken == null)
```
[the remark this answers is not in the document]
neha-bhargava: This is a method which is called for silent flows as well. I moved this method from there to reuse it for OBO as well.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: I don't think this will scenario will ever occur where there will not be a user assertion for an obo flow as it is a required param for AcquireTokenOnBehalfOf.

tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs
trwalke: There was a parsing error earlier in the development cycle but it is fixed now

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: Added check

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -123,6 +123,12 @@ public sealed partial class ConfidentialClientApplication
             IEnumerable<string> scopes,
             UserAssertion userAssertion)
         {
+            if (userAssertion == null)
+            {
+                ServiceBundle.ApplicationLogger.Error("User assertion for OBO request should not be null");
```
bgavrilMS, line 128: Cool

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -654,20 +642,21 @@ private MsalAccessTokenCacheItem FilterByKeyId(MsalAccessTokenCacheItem item, Au
                                 "Filtering by home account id");
             }
 
+            // This will also filter for the case when familyId is null and exclude RTs with familyId in filtered list
             rtCacheItems = rtCacheItems.FilterWithLogging(item =>
-                            string.Equals(item.FamilyId ?? string.Empty,
-                            familyId ?? string.Empty, StringComparison.OrdinalIgnoreCase),
-                            requestParams.RequestContext.Logger,
-                            "Filtering by family id");
+                    string.Equals(item.FamilyId ?? string.Empty,
```
bgavrilMS, line 647: Should this filter be in the `else` clause associated with `string.IsNullOrEmpty(familyId)` ? In the corner case that an FOCI app is used for OBO, all of the RTs from it will have familyId set to "1". OBO does not look for a family RT specifically, but when it looks for RT's for a specific client_id, it should not ignore an RT just because it also has familyId=1

tests/devapps/WebApi/readme_perf.txt
```diff
@@ -81,12 +90,24 @@ Static Wilson LRU cache with normal size limitation (max size ~2GB or 700k token
...
-
-
-
-To run the perf tests, start this webapi, install a tool like https://github.com/codesenberg/bombardier and hit it with:
-
-bombardier -d 60s -l https://localhost:44355/FlatCache
-bombardier -d 60s -l https://localhost:44355/StaticDictionary
-bombardier -d 60s -l https://localhost:44355/Singleton
-bombardier -d 60s -l https://localhost:44355/WilsonLruCache
+OBO performance tests for refresh flow. Added with 4.33 release.
```
bgavrilMS, line 93: Can you please describe the methodology used? I.e. how the cache is hit, how often the Refresh Token flow is performed (roughly), the timeout values used etc.

bgavrilMS (approved):
LGTM - please see my comment on the FRT filtering and more comments needed for the perf tests

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -86,18 +87,34 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+                var msalTokenResponse = await SilentRequestHelper.RefreshAccessTokenAsync(appRefreshToken, this, AuthenticationRequestParameters, cancellationToken)
+                .ConfigureAwait(false);
+
+                return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
+            }
+
+            AuthenticationRequestParameters.RequestContext.Logger.Verbose("[OBO request] No Refresh Token was found in the cache. Fetching OBO token from ESTS");
+
+            var result = await FetchNewAccessTokenAsync(cancellationToken).ConfigureAwait(false);
+            return result;
```
jennyf19, line 115: return directly w/out creating a variable

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -86,18 +87,34 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            // If a refresh token is not found, fetch a new access token
+            if (appRefreshToken != null)
+            {
+                var msalTokenResponse = await SilentRequestHelper.RefreshAccessTokenAsync(appRefreshToken, this, AuthenticationRequestParameters, cancellationToken)
+                .ConfigureAwait(false);
+
+                return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
+            }
+
+            AuthenticationRequestParameters.RequestContext.Logger.Verbose("[OBO request] No Refresh Token was found in the cache. Fetching OBO token from ESTS");
```
jennyf19, line 112: >Verbose [](http://example.com/codeflow?start=66&length=7)

change to info

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: No, this should not be placed in the else block since this also filters RT for the foci refresh token. For example if an app left a family then it should not return a refresh token for that app. I added this to the else block earlier and there is a test that failed for this case: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9d2c425e7b301580ee8355b6f8c6cf66f505384b/tests/Microsoft.Identity.Test.Unit/RequestsTests/FociTests.cs#L170

jennyf19:
have tested with JM's long running process branch.

---
In reply to: [851337117](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2667#issuecomment-851337117) [](http://example.com/codeflow?ancestors=851337117,851337117)

jennyf19 (approved):
:shipit:

bgavrilMS (approved):

Issue #2665, opened 2021-05-27
[Feature Request] [M] MSAL improves search of metadata in the federation metadata XML during WS-Trust flows.


trwalke:
**Is your feature request related to a problem? Please describe.**
Continuation of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2152

**Describe the solution you'd like**
MSAL should search for the "wsdl:definitions" node within the provided federation metadata document during WS-Trust instead of assuming it is always the top node. WHile ADFS provided by Microsoft will always follow this pattern, third party IDPs may not which can cause authentication errors.

MSAL should search for the node in the Federation metadata provided by either the endpoint or the WithFederationMetadata(string federationMetadata) api instead of assuming that the top node will always have the metadata.
This will enable third party IDPs to provide federation metadata in a slightly different format.

The location where this parsing occurs is here:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs#L82

here:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs#L138

and here:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs#L197

**Alternatives**
MSAL can provide an api that enables developers to provide the parent node of the XML node that contains the required "wsdl:definitions". If the required node is the top node this api should not be used as MSAL already searches the top node by default.

trwalke:
There is still an ongoing discussion on whether or not MSAL should provide the api for this or do it internally in Email

PR #2668, opened 2021-05-28
Enabling mexDocument to search for wsdl:definition node in federation metadata XML

trwalke:
Fixes # .
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2665

Enabling mexDocument to search for wsdl:definition node in federation metadata XML
Added proper exception for federation metadata parsing
Added unit test

**Testing**
Added unit test

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
+++ b/src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
@@ -79,20 +79,7 @@ namespace Microsoft.Identity.Client.WsTrust
 
         private void ReadPolicies(XContainer mexDocument)
         {
-            IEnumerable<XElement> policyElements = mexDocument.Elements().First().Elements(XmlNamespace.Wsp + "Policy");
-
-            //Unable to find policy information in root node. Attempting to search document for policy information
-            if (policyElements.Count() == 0)
-            {
-                policyElements = mexDocument.Elements().DescendantsAndSelf().Elements(XmlNamespace.Wsp + "Policy");
-
-                if (policyElements.Count() == 0)
-                {
-                    throw new MsalClientException(
-                    MsalError.ParsingWsMetadataExchangeFailed,
-                    MsalErrorMessage.ParsingMetadataDocumentFailed + " Could not parse policy data.");
-                }
-            }
+            IEnumerable<XElement> policyElements = FindElements(mexDocument, XmlNamespace.Wsp, "Policy");
 
             foreach (XElement policy in policyElements)
             {
@@ -149,20 +136,7 @@ namespace Microsoft.Identity.Client.WsTrust
 
         private void ReadPolicyBindings(XContainer mexDocument)
         {
-            IEnumerable<XElement> bindingElements = mexDocument.Elements().First().Elements(XmlNamespace.Wsdl + "binding");
-
-            //Unable to find binding information in root node. Attempting to search document for binding information
-            if (bindingElements.Count() == 0)
-            {
-                bindingElements = mexDocument.Elements().DescendantsAndSelf().Elements(XmlNamespace.Wsdl + "binding");
-
-                if (bindingElements.Count() == 0)
-                {
-                    throw new MsalClientException(
-                    MsalError.ParsingWsMetadataExchangeFailed,
-                    MsalErrorMessage.ParsingMetadataDocumentFailed + " Could not parse binding data.");
-                }
-            }
+            IEnumerable<XElement> bindingElements = FindElements(mexDocument, XmlNamespace.Wsdl, "binding");    
 
             foreach (XElement binding in bindingElements)
             {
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
```diff
@@ -136,6 +150,20 @@ private void ReadPolicies(XContainer mexDocument)
         private void ReadPolicyBindings(XContainer mexDocument)
         {
             IEnumerable<XElement> bindingElements = mexDocument.Elements().First().Elements(XmlNamespace.Wsdl + "binding");
+
+            //Unable to find binding information in root node. Attempting to search document for binding information
+            if (bindingElements.Count() == 0)
```
jmprieur, line 155: This seems very similar code, with a different parent to the collection (mexDocument, binding) and different value to search ("Policy", "binding", "service")
you might want to extract this code into a method ?

jmprieur (commented):
A suggestion (avoid copy/paste :))

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
trwalke: Sorry, this wasn't the right version of the branch. I was in a hurry yesterday and forgot to commit my latest refactoring.

Thanks for looking @jmprieur

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
```diff
@@ -225,6 +228,26 @@ private void SetPolicyEndpointAddresses(XContainer mexDocument)
             }
         }
 
+        private IEnumerable<XElement> FindElements(XContainer mexDocument, XNamespace xNamespace, string element)
+        {
+            IEnumerable<XElement> bindingElements = mexDocument.Elements().First().Elements(xNamespace + element);
```
bgavrilMS, line 233: nit: (naming) these are no longer `biding` elements, - they can be `service`, `biding` or `policy`

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
```diff
@@ -225,6 +228,26 @@ private void SetPolicyEndpointAddresses(XContainer mexDocument)
...
+
+            //Unable to find metadata information in root node. Attempting to search document.
+            if (bindingElements.Count() == 0)
```
bgavrilMS, line 236: I'm not familiar with XML parsing, so just a newbie question - can bindingElements be null? Probably not..

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
```diff
@@ -225,6 +228,26 @@ private void SetPolicyEndpointAddresses(XContainer mexDocument)
...
+            {
+                bindingElements = mexDocument.Elements().DescendantsAndSelf().Elements(xNamespace + element);
+
+                if (bindingElements.Count() == 0)
+                {
+                    throw new MsalClientException(
+                    MsalError.ParsingWsMetadataExchangeFailed,
+                    MsalErrorMessage.ParsingMetadataDocumentFailed + $" Could not parse {element} data.");
```
bgavrilMS, line 244: Could not find element XYZ

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
SameerK-MSFT: bindingElemts.Any() has better perf than .Count() == 0

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
```diff
@@ -225,6 +228,26 @@ private void SetPolicyEndpointAddresses(XContainer mexDocument)
...
         }
 
+        private IEnumerable<XElement> FindElements(XContainer mexDocument, XNamespace xNamespace, string element)
+        {
+            IEnumerable<XElement> bindingElements = mexDocument.Elements().First().Elements(xNamespace + element);
+
+            //Unable to find metadata information in root node. Attempting to search document.
+            if (bindingElements.Count() == 0)
+            {
+                bindingElements = mexDocument.Elements().DescendantsAndSelf().Elements(xNamespace + element);
```
pmaytak, line 238: Could this be changed to `mexDocument.Elements().DescendantsAndSelf(xNamespace + element);`?

tests/Microsoft.Identity.Test.Unit/RequestsTests/IntegratedWindowsAuthUsernamePasswordTests.cs
```diff
@@ -288,6 +288,51 @@ public async Task AcquireTokenByIntegratedWindowsAuthTestAsync()
             }
         }
 
+        [TestMethod]
+        [DeploymentItem(@"Resources\TestMex3rdParty.xml")]
+        [DeploymentItem(@"Resources\WsTrustResponse13.xml")]
+        public async Task AcquireTokenByIntegratedWindowsAuth3rdPartyIDPTestAsync()
```
pmaytak, line 294: This test seems to be almost the same as `AcquireTokenByIntegratedWindowsAuthTestAsync`. Can the XML file path can be passed into as a parameter to `AcquireTokenByIntegratedWindowsAuthTestAsync` as a data row?

pmaytak (approved):

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
pmaytak, line 238: Another thought. When processing a 3rd party IdP metadata, we always come to this line (3 times in this case) and end up enumerating all descendants each time. Perhaps it's better in the constructor to find the `wsdl:definitions` element and then just pass that around (to the `ReadPolicies`, `ReadPolicyBindings`, etc.) instead of the whole mex document.

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
trwalke: @bgavrilMS not entirely sure but now that we allow custom federation data ill add a null check.

src/client/Microsoft.Identity.Client/WsTrust/MexDocument.cs
trwalke: I tried this earlier and it did not work

PR #2675, opened 2021-06-02
Do not throw exception if WithTenant and WithAuthority mention the sa…

bgavrilMS:
…me tenant

Fixes # .

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
@@ -512,7 +512,8 @@ namespace Microsoft.Identity.Client
                     Config.AuthorityInfo.AuthorityType == AuthorityType.Aad)
                 {
                     AadAuthority aadAuthority = Authority.CreateAuthority(Config.AuthorityInfo) as AadAuthority;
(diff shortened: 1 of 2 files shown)


pmaytak (approved):

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -512,9 +512,12 @@ private void ResolveAuthority()
                     Config.AuthorityInfo.AuthorityType == AuthorityType.Aad)
                 {
                     AadAuthority aadAuthority = Authority.CreateAuthority(Config.AuthorityInfo) as AadAuthority;
-                    if (!aadAuthority.IsCommonOrganizationsOrConsumersTenant())
+                    if (!aadAuthority.IsCommonOrganizationsOrConsumersTenant() &&
```
jennyf19, line 515: B2C will work as well, correct?

bgavrilMS: yes, this whole logic is for when `Config.AuthorityInfo.AuthorityType == AuthorityType.Aad`

I guess we could throw a meaningful exception if authority is B2C and "WithTenant" is used. I think we do nothing right now.

PR #2676, opened 2021-06-03
Add metrics

### LINKED ISSUE #2646: Design and add monitoring flags for global stats
- [x] tokens obtained by MSAL (for entire process)
- [x] tokens obtained by MSAL via cache (for entire process)
- [x] time spent in MSAL (per process) - includes HTTP and cache times
- [x] <s>Azure region detected (expose in auth result metadata)</s> Separate issue for this.

Should be exposed in AuthenticationResult.AuthenticationResultMetadata and via telemetry.
For the entire process = for the whole life of MSAL, i.e. use static counters. 

- [x] Create API Review doc

pmaytak:
Fixes #2646.

Added additional MSAL-wide static metrics.

**Testing**
Tested manually with devapps. Added `MetricsTests.cs`, updated telemetry tests.


src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
```diff
@@ -65,6 +63,7 @@ internal class InteractiveRequest : RequestBase
             cancellationToken.ThrowIfCancellationRequested();
             MsalTokenResponse tokenResponse = await GetTokenResponseAsync(cancellationToken)
                 .ConfigureAwait(false);
+            Metrics.TotalAccessTokensFromIdP++;
```
bgavrilMS, line 66: Since you access a shared variable from multiple threads, the 2 operations (add and assign) might not happen at once. You can use smth like Interlocked.Increment to have "add" as an atomic operation. See https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=net-5.0

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -24,20 +24,35 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
...
         /// <summary>
-        /// Time (in ms) MSAL spent for HTTP communication.
+        /// Time, in milliseconds, MSAL spent for HTTP communication during this request.
         /// </summary>
         public long DurationInHttpInMs { get; set; }
+
+        /// <summary>
+        /// Total tokens obtained by MSAL from the identity provider.
+        /// </summary>
+        public long TotalAccessTokensFromIdP { get; set; }
```
bgavrilMS, line 46: I recommend that you start an API review as soon as possible, as we need to allow some time for reviewers.

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -39,20 +39,5 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
         /// Time, in milliseconds, MSAL spent for HTTP communication during this request.
         /// </summary>
         public long DurationInHttpInMs { get; set; }
-
-        /// <summary>
-        /// Total tokens obtained by MSAL from the identity provider.
-        /// </summary>
-        public long TotalAccessTokensFromIdP { get; set; }
```
bgavrilMS: Maybe add a comment in this class that more metrics are in the Metrics class.

bgavrilMS (approved):

bgavrilMS:
I think we should also expose the Azure Region in the metadatga (as a separate PR of course, but we can add it to the API revbiew)

jennyf19:
we also have telemetry tests, so you could add this new content there and assert that it's in the headers. would be nice to have a telemetry dashboard so we can easily read all these values, as the telemetry has really exploded. we should bring this up at a team meeting. no point in having telemetry if it's not easily accessible.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -114,7 +113,11 @@ public string GetCurrentRequestHeader(ApiEvent eventInProgress)
 
             var platformConfig = new StringBuilder();
             platformConfig.Append(ConvertFromStringToBitwise(isTokenCacheSerialized) + ",");
-            platformConfig.Append(ConvertFromStringToBitwise(isLegacyCacheEnabled));
+            platformConfig.Append(ConvertFromStringToBitwise(isLegacyCacheEnabled) + ",");
+            
+            platformConfig.Append(Metrics.TotalDurationInMs + ",");
+            platformConfig.Append(Metrics.TotalAccessTokensFromIdP + ",");
+            platformConfig.Append(Metrics.TotalAccessTokensFromCache);
```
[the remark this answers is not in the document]
jennyf19: i think it's specific to .NET, so not sure if we are required to track that. it would be in the telemetry section of the API review docs.

tests/Microsoft.Identity.Test.Unit/MetricsTests.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System;
```
jennyf19, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

﻿﻿using System;
```

tests/Microsoft.Identity.Test.Unit/MetricsTests.cs
```diff
@@ -0,0 +1,128 @@
...
+                Assert.AreEqual(1, Metrics.TotalAccessTokensFromIdP);
+                Assert.AreEqual(0, Metrics.TotalAccessTokensFromCache);
+                Assert.IsTrue(Metrics.TotalDurationInMs > 0);
+            }
+        }
+
+        [TestMethod]
+        public async Task MetricsUpdatedSucessfully_AcquireTokenSilent_Async()
+        {
+            using (var harness = CreateTestHarness())
```
jennyf19, line 99: nit: move silent w/interactive for more interesting results maybe?
can you add the metrics to other tests, like ones that are testing the TokenSource, you could also check the metrics there as well. just a thought.

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs
```diff
@@ -63,6 +63,7 @@ private async Task<MsalTokenResponse> SendTokenRequestToBrokerAsync()
             }
 
             ValidateResponseFromBroker(msalTokenResponse);
+            Metrics.IncrementTotalAccessTokensFromIdP();
```
trwalke, line 66: The android broker may return a result from its internal cache rather than the IDP. I am not sure if there is anything in the broker response that indicates whether the token came from the IDP or the internal cache. Maybe worth a check. Otherwise, you can consider all tokens from the broker the idp? Does WAM have something similar?

trwalke (approved):
Need to fix telemetry tests.
LGTM otherwise

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs
pmaytak: I think WAM is the same, we don't know if it's from IdP or cache. Since we have `TokenSource.Broker`, I could add separate token count for that for consistency?

neha-bhargava (approved):

tests/Microsoft.Identity.Test.Unit/MetricsTests.cs
pmaytak: #Resolved

tests/Microsoft.Identity.Test.Unit/MetricsTests.cs
pmaytak: #Resolved

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs
pmaytak: #Resolved

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -115,9 +115,10 @@ public string GetCurrentRequestHeader(ApiEvent eventInProgress)
             platformConfig.Append(ConvertFromStringToBitwise(isTokenCacheSerialized) + ",");
             platformConfig.Append(ConvertFromStringToBitwise(isLegacyCacheEnabled) + ",");
             
-            platformConfig.Append(Metrics.TotalDurationInMs + ",");
             platformConfig.Append(Metrics.TotalAccessTokensFromIdP + ",");
-            platformConfig.Append(Metrics.TotalAccessTokensFromCache);
+            platformConfig.Append(Metrics.TotalAccessTokensFromCache + ",");
```
bgavrilMS, line 119: I did not understand that we want to send these as telemetry. What is the point? Telemetry string is pretty limited in size, we should be selective with data points.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
```diff
@@ -105,16 +104,17 @@ private async Task<MsalTokenResponse> GetTokenResponseAsync(CancellationToken ca
...
                     cancellationToken)
                     .ConfigureAwait(false);
 
                 // if we don't get back a result, then continue with the WebUi 
-                if (tokenResponse != null)
+                if (brokerTokenResponse != null)
                 {
                     _logger.Info("Broker attempt completed successfully. ");
-                    return tokenResponse;
+                    Metrics.IncrementTotalAccessTokensFromBroker();
```
trwalke, line 116: The broker implementation looks good @pmaytak

pmaytak:
Discussed with Henrik and Bogdan - removed telemetry stuff, since it's a bit redundant as we send cache info and silent request count already.

PR #2677, opened 2021-06-03
improving the XML comment for GetAccountAsync

jmprieur:
See also https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/troubleshooting#getaccountasync-doesnt-return-any-account-in-clouds-scenarios

### DIFF
--- a/src/client/Microsoft.Identity.Client/IClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/IClientApplicationBase.cs
@@ -43,7 +43,8 @@ namespace Microsoft.Identity.Client
         Task<IEnumerable<IAccount>> GetAccountsAsync();
 
         /// <summary>
-        /// Get the <see cref="IAccount"/> by its identifier among the accounts available in the token cache.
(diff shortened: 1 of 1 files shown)


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/IClientApplicationBase.cs
```diff
@@ -43,7 +43,8 @@ public partial interface IClientApplicationBase
         Task<IEnumerable<IAccount>> GetAccountsAsync();
 
         /// <summary>
-        /// Get the <see cref="IAccount"/> by its identifier among the accounts available in the token cache.
+        /// Get the <see cref="IAccount"/> by its identifier among the accounts available in the token cache and of the same
+        /// instance (authority host) as <see cref="Authority"/>
```
bgavrilMS, line 47: We do not use the term `instance` in MSAL. We use `environment`. See IAccount interface.

```suggestion
        /// environment (authority host) as <see cref="Authority"/> 
```

PR #2678, opened 2021-06-04
Add extra filtering by env logs.

pmaytak:
Fixes # .

Add extra verbose logging when filtering access tokens by environment. We had logs for other filters, but not this one.

**Testing**
Manually with a dev app.

**Performance impact**
Only for Verbose logging, extra `.Count` on token collection.


bgavrilMS (approved):

neha-bhargava (approved):
👍

PR #2680, opened 2021-06-04
Add more logging and small enhacement to an int test

bgavrilMS:
Fixes - no work item, need more logging for investigation.

- log app config hash code, to be able to tell if requests belong to same app
- rename DefaultLogger to AppLogger
- enhance an integration test to use WithAuthority at request level

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -50,37 +50,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
             acquireTokenParameters.LogParameters(AuthenticationRequestParameters.RequestContext.Logger);
         }
 
-        private void LogRequestStarted(AuthenticationRequestParameters authenticationRequestParameters)
-        {
-            string messageWithPii = string.Format(
-                CultureInfo.InvariantCulture,
-                "=== Token Acquisition ({3}) started:\n\tAuthority: {0}\n\tScope: {1}\n\tClientId: {2}\n\t",
-                authenticationRequestParameters.AuthorityInfo?.CanonicalAuthority,
-                authenticationRequestParameters.Scope.AsSingleString(),
-                authenticationRequestParameters.AppConfig.ClientId,
-                GetType().Name);
-
-            string messageWithoutPii = string.Format(
-                CultureInfo.InvariantCulture,
-                "=== Token Acquisition ({0}) started:\n\t",
-                GetType().Name);
-
-            if (authenticationRequestParameters.AuthorityInfo != null &&
-                KnownMetadataProvider.IsKnownEnvironment(authenticationRequestParameters.AuthorityInfo?.Host))
-            {
-                messageWithoutPii += string.Format(
-                    CultureInfo.CurrentCulture,
-                    "\n\tAuthority Host: {0}",
-                    authenticationRequestParameters.AuthorityInfo?.Host);
-            }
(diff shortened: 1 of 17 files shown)


tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
```diff
@@ -55,16 +55,16 @@ public void TestInitialize()
...
-            await RunOnBehalfOfTestAsync(aadUser1, true).ConfigureAwait(false);
-            await RunOnBehalfOfTestAsync(aadUser2, false).ConfigureAwait(false);
-            await RunOnBehalfOfTestAsync(adfsUser, true).ConfigureAwait(false);
-            await RunOnBehalfOfTestAsync(aadUser2, true).ConfigureAwait(false);
-            await RunOnBehalfOfTestAsync(aadUser2, false, true).ConfigureAwait(false);
+            //await RunOnBehalfOfTestAsync(aadUser1, true).ConfigureAwait(false);
+            //await RunOnBehalfOfTestAsync(aadUser2, false).ConfigureAwait(false);
+            //await RunOnBehalfOfTestAsync(adfsUser, true).ConfigureAwait(false);
+            //await RunOnBehalfOfTestAsync(aadUser2, true).ConfigureAwait(false);
+            //await RunOnBehalfOfTestAsync(aadUser2, false, true).ConfigureAwait(false);
```
neha-bhargava, line 67: Any reason for commenting these?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -338,23 +308,65 @@ protected Task ResolveAuthorityAsync()
...
+                authenticationRequestParameters.RequestContext.Logger.Error("The default token cache provided by MSAL is not designed to be performant when used in confidential client applications. Please use token cache serialization. See https://aka.ms/msal-net-cca-token-cache-serialization.");
+            }
+        }
+
         private void LogReturnedToken(AuthenticationResult result)
         {
             if (result.AccessToken != null && 
                 AuthenticationRequestParameters.RequestContext.Logger.IsLoggingEnabled(LogLevel.Info))
             {
+                int appHashCode = AuthenticationRequestParameters.AppConfig.GetHashCode();
```
pmaytak, line 357: - Maybe rename here and in log below to appConfigHashCode?
- In which scenario would this hashCode be useful?

neha-bhargava: I think the hash code is useful to know the instance of confidential client being used to track how many CCA are created.

trwalke (approved):
I think this is a good idea.
LGTM

pmaytak (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
bgavrilMS: Good catch, reverting.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: Yes, just as Neha stated, it's used to differentiate between different CCA objects. It's to be able to tell if ppl are using 1 or more instances of CCA object, as it has cache implications.

PR #2684, opened 2021-06-08
Add auto-detected region to auth result metadata.

### LINKED ISSUE #2646: Design and add monitoring flags for global stats

pmaytak:
Fixes #2646.

Add the region which was auto-detected to `AuthenticationResult.AuthenticationResultMetadata`.

**Testing**
Tested manually with regional test app. Updated some unit tests.

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
@@ -39,5 +39,10 @@ namespace Microsoft.Identity.Client
         /// Time (in ms) MSAL spent for HTTP communication.
         /// </summary>
         public long DurationInHttpInMs { get; set; }
+
+        /// <summary>
+        /// Azure region which was auto-detected.
+        /// </summary>
+        public string AzureRegionAutoDetected { get; set; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Identity.Client.Region
         private readonly IHttpManager _httpManager;
         private readonly int _imdsCallTimeoutMs;
 
-        private static string s_autoDiscoveredRegion;
+        internal static string AutoDiscoveredRegion;
         private static bool s_failedAutoDiscovery = false;
 
         public RegionManager(
@@ -49,7 +49,7 @@ namespace Microsoft.Identity.Client.Region
             if (shouldClearStaticCache)
             {
                 s_failedAutoDiscovery = false;
-                s_autoDiscoveredRegion = null;
+                AutoDiscoveredRegion = null;
             }
         }
 
@@ -65,7 +65,7 @@ namespace Microsoft.Identity.Client.Region
 
             // MSAL always performs region auto-discovery, even if the user configured an actual region
             // in order to detect inconsistencies and report via telemetry
-            var discoveredRegion = await DiscoverAndCacheAsync(azureRegionConfig, logger, requestContext.UserCancellationToken).ConfigureAwait(false);
+            var discoveredRegion = await DiscoverAndCacheAsync(logger, requestContext.UserCancellationToken).ConfigureAwait(false);
 
             RecordTelemetry(requestContext.ApiEvent, azureRegionConfig, discoveredRegion);
 
@@ -135,7 +135,7 @@ namespace Microsoft.Identity.Client.Region
                  apiEvent.RegionOutcome == (int)(default(RegionOutcome)));
         }
 
-        private async Task<RegionInfo> DiscoverAndCacheAsync(string azureRegionConfig, ICoreLogger logger, CancellationToken requestCancellationToken)
+        private async Task<RegionInfo> DiscoverAndCacheAsync(ICoreLogger logger, CancellationToken requestCancellationToken)
         {
             if (s_failedAutoDiscovery == true)
             {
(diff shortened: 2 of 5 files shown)


src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -135,7 +135,7 @@ private bool IsTelemetryRecorded(ApiEvent apiEvent)
                  apiEvent.RegionOutcome == (int)(default(RegionOutcome)));
         }
 
-        private async Task<RegionInfo> DiscoverAndCacheAsync(string azureRegionConfig, ICoreLogger logger, CancellationToken requestCancellationToken)
```
pmaytak, line 138: Cleanup - unused variable.

neha-bhargava (approved):
Looks good

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -96,7 +96,9 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
 
                         authenticationResult.AuthenticationResultMetadata.DurationTotalInMs = sw.ElapsedMilliseconds;
                         authenticationResult.AuthenticationResultMetadata.DurationInHttpInMs = apiEvent.DurationInHttpInMs;
-                        authenticationResult.AuthenticationResultMetadata.DurationInCacheInMs = apiEvent.DurationInCacheInMs;                        
+                        authenticationResult.AuthenticationResultMetadata.DurationInCacheInMs = apiEvent.DurationInCacheInMs;
+
+                        authenticationResult.AuthenticationResultMetadata.AzureRegionAutoDetected = Region.RegionManager.AutoDiscoveredRegion;
```
jennyf19, line 101: question to discuss w/team: we had talked about doing a helper method which would return the region, so developers could call that upfront and use it directly, or call it and see what it returns. is there value in having it come after the fact in the auth result, or are we still considering the helper method? or both?

trwalke:
I think the link in the description is targeting the wrong issue?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke, line 101: We seem to have two different names for this value. autoDetected and autoDiscovered. Is this intentional?

pmaytak: They seem to be used interchangeably. Yea, we should be consistent. @neha-bhargava Do you know how it's mentioned in the specs, what's official term (auto-discover vs auto-detect)?

pmaytak: Additional thoughts. Since region auto discovery is done once, then cached, region in auth result will always be the same. So maybe a helper method or a public static property on like RegionManager might be more appropriate. Also the discovery is done when AcquireTokenX is called, so before that first call, region helper will return null, unless we add discovery logic to it (which I think we also discussed). I think the primary question is what is the end users' scenario for needing a region here.

jennyf19:
it's correct. Bogdan suggested this in the metrics PR, so it's linking to that PR. But looking at this now, we might want to have an issue if we want it as a helper. 

---
In reply to: [857383362](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2684#issuecomment-857383362) [](http://example.com/codeflow?ancestors=857383362,857383362)

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
jennyf19: @jmprieur thoughts on this?
I think @pmaytak found the right question:

> what is the end users' scenario for needing a region here.

neha-bhargava: This term is used interchangeably in the docs too. Good point to make it consistent. Since we use `TryAutoDetect` value to indicate that MSAL should discover the region we can use auto-detect to be consistent.

neha-bhargava: @pmaytak The end user scenario for needing this is to confirm the region that is being used to hit the regional endpoint. And I remember we discussed this and decided that we should expose the auto detected region but when I think again, I think we should actually expose the region used instead. @bgavrilMS to confirm

jmprieur: static string ConfidentialClientApplication.AutoDetecteRegion() ?

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -39,5 +39,10 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
         /// Time (in ms) MSAL spent for HTTP communication.
         /// </summary>
         public long DurationInHttpInMs { get; set; }
+
+        /// <summary>
+        /// Azure region which was auto-detected.
```
bgavrilMS, line 44:
```suggestion
        /// MSAL will try to auto-detect the region if .WithAzureRegion API is used. If the app developer
        /// configured a region, MSAL will always use it, but auto-detection will still be attempted for telemetry purposes. 
        /// The value of the region here is the auto-detected one.
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: @neha-bhargava - if the user configures a region, it is always used. Is there a way to detect if the region was ok or not? I don't think there is.

bgavrilMS (approved):

trwalke (approved):

pmaytak:
New issue: [[Feature Request] Add a public method to get auto-detected region #2698](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2698)

PR #2687, opened 2021-06-08
CCS Header updates for interactive request

trwalke:
Adding ccs header for interactive request using client info.

Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2525

Adding ccs header for interactive request using client info.
Requesting client info from authorize endpoint

**Testing**
existing integration tests cover scenario

**Performance impact**
Minor. For interactive flow only

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
@@ -12,6 +12,7 @@ using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.UI;
 using Microsoft.Identity.Client.Internal.Broker;
 using Microsoft.Identity.Client.Utils;
+using Microsoft.Identity.Json.Linq;
 
 namespace Microsoft.Identity.Client.Internal.Requests
 {
@@ -22,17 +23,20 @@ namespace Microsoft.Identity.Client.Internal.Requests
         private readonly string _authorizationCode;
         private readonly string _pkceCodeVerifier;
         private readonly TokenClient _tokenClient;
+        private readonly string _clientInfo;
 
         public AuthCodeExchangeComponent( 
             AuthenticationRequestParameters requestParams,
             AcquireTokenInteractiveParameters interactiveParameters,
             string authorizationCode,
-            string pkceCodeVerifier)
+            string pkceCodeVerifier,
+            string clientInfo)
         {
             _requestParams = requestParams ?? throw new ArgumentNullException(nameof(requestParams));
             _interactiveParameters = interactiveParameters ?? throw new ArgumentNullException(nameof(interactiveParameters));
             _authorizationCode = authorizationCode ?? throw new ArgumentNullException(nameof(authorizationCode));
             _pkceCodeVerifier = pkceCodeVerifier ?? throw new ArgumentNullException(nameof(pkceCodeVerifier));
+            _clientInfo = clientInfo;
 
             _tokenClient = new TokenClient(requestParams);
             _interactiveParameters.LogParameters(requestParams.RequestContext.Logger);
@@ -40,10 +44,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
         public Task<MsalTokenResponse> FetchTokensAsync(CancellationToken cancellationToken)
         {
-            if (!string.IsNullOrEmpty(_interactiveParameters.LoginHint))
-            {
-                _tokenClient.AddHeaderToClient(Constants.CCSRoutingHintHeader, CoreHelpers.GetCCSUpnHeader(_interactiveParameters.LoginHint));
-            }
+            AddCCSHeadersToTokenClient();
             return _tokenClient.SendTokenRequestAsync(GetBodyParameters());
         }
 
@@ -58,6 +59,22 @@ namespace Microsoft.Identity.Client.Internal.Requests
             };
 
             return dict;
+        }
+
+        private void AddCCSHeadersToTokenClient()
+        {
+            if (!string.IsNullOrEmpty(_clientInfo))
+            {
+                var decodedClientInfo = Base64UrlHelpers.DecodeToString(_clientInfo);
+
+                var clientInfo = JsonHelper.DeserializeFromJson<ClientInfo>(decodedClientInfo);
+
+                _tokenClient.AddHeaderToClient(Constants.CCSRoutingHintHeader, CoreHelpers.GetCCSClientInfoheader(clientInfo.UniqueObjectIdentifier, clientInfo.UniqueTenantIdentifier));
+            }
+            else if (!string.IsNullOrEmpty(_interactiveParameters.LoginHint))
+            {
+                _tokenClient.AddHeaderToClient(Constants.CCSRoutingHintHeader, CoreHelpers.GetCCSUpnHeader(_interactiveParameters.LoginHint));
+            }
         }
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 var clientInfo = ClientInfo.CreateFromJson(_clientInfo);
 
                 _tokenClient.AddHeaderToClient(Constants.CcsRoutingHintHeader,
-                                               CoreHelpers.GetCcsClientInfoheader(clientInfo.UniqueObjectIdentifier,
+                                               CoreHelpers.GetCcsClientInfoHeader(clientInfo.UniqueObjectIdentifier,
                                                                                   clientInfo.UniqueTenantIdentifier));
             }
             else if (!string.IsNullOrEmpty(_interactiveParameters.LoginHint))
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
@@ -65,16 +65,22 @@ namespace Microsoft.Identity.Client.Internal.Requests
         {
             if (!string.IsNullOrEmpty(_clientInfo))
             {
-                var decodedClientInfo = Base64UrlHelpers.DecodeToString(_clientInfo);
-
-                var clientInfo = JsonHelper.DeserializeFromJson<ClientInfo>(decodedClientInfo);
-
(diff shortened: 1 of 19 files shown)


src/client/Microsoft.Identity.Client/UI/AuthorizationResult.cs
```diff
@@ -138,6 +143,8 @@ public static AuthorizationResult FromStatus(AuthorizationStatus status, string
         [JsonProperty]
         public string CloudInstanceHost { get; set; }
 
+        public string Client_info { get; set; }
```
jennyf19, line 146: >Client_info [](http://example.com/codeflow?start=22&length=11)

`ClientInfo`

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
```diff
@@ -59,5 +60,27 @@ public Task<MsalTokenResponse> FetchTokensAsync(CancellationToken cancellationTo
 
             return dict;
         }
+
+        private void AddCCSHeadersToTokenClient()
```
jennyf19, line 64: nit: use the ClientInfo class, so you don't have to worry about error handling w/the json parsing. 
`ClientInfo.CreateFromJson(authResult.ClientInfo);`

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
```diff
@@ -240,7 +240,7 @@ private static void AddKeyValueString(StringBuilder messageBuilder, string key,
 
         internal static string GetCCSClientInfoheader(string userObjectId, string userTenantID)
         {
-            return (string.IsNullOrEmpty(userObjectId) || string.IsNullOrEmpty(userTenantID)) ? string.Empty : $@"oid:{userObjectId}@{userTenantID}";
+            return (string.IsNullOrEmpty(userObjectId) || string.IsNullOrEmpty(userTenantID)) ? string.Empty : $@"Oid:{userObjectId}@{userTenantID}";
         }
```
jennyf19, line 244: should Oid be captialized?

tests/Microsoft.Identity.Test.Common/TestCommon.cs
```diff
@@ -114,5 +116,18 @@ public static IServiceBundle CreateDefaultAdfsServiceBundle()
             {                
             };
         }
+
+        public static string GetCCSHeader(HttpSnifferClientFactory factory)
```
jennyf19, line 120: how do you know the header contains the correct values and that clientinfo is returned or not from the /authorize endpoint?
any impact on B2C or ADFS?

There are no tests for `AddCCSHeadersToTokenClient()`, unless you use the ClientInfo class directly.

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
```diff
@@ -131,8 +131,10 @@ private async Task<MsalTokenResponse> GetTokenResponseAsync(CancellationToken ca
                 .ConfigureAwait(false);
 
             _logger.Info("An authorization code was retrieved from the /authorize endpoint. ");
-            string authCode = result.Item1;
+            AuthorizationResult authResult = result.Item1;
+            string authCode = authResult.Code;
             string pkceCodeVerifier = result.Item2;
+            string clientInfo = authResult.Client_info;
```
jennyf19, line 137: >authResult.Client_info [](http://example.com/codeflow?start=32&length=22)

nit: remove variable

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
```diff
@@ -22,28 +23,28 @@ internal class AuthCodeExchangeComponent : ITokenRequestComponent
...
             string authorizationCode,
-            string pkceCodeVerifier)
+            string pkceCodeVerifier,
+            string clientInfo)
         {
             _requestParams = requestParams ?? throw new ArgumentNullException(nameof(requestParams));
             _interactiveParameters = interactiveParameters ?? throw new ArgumentNullException(nameof(interactiveParameters));
             _authorizationCode = authorizationCode ?? throw new ArgumentNullException(nameof(authorizationCode));
             _pkceCodeVerifier = pkceCodeVerifier ?? throw new ArgumentNullException(nameof(pkceCodeVerifier));
+            _clientInfo = clientInfo;
```
jennyf19, line 40: should we throw here if it's null? otherwise, if using ClientInfo, we will throw later. how often will this be null from the /authorize endpoint? do we know in which cases it will be null, and if so, can we send an emtpy string or something else to avoid an issue?
_clientInfo = clientInfo ?? string.Empty

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
trwalke: Only the O, It is based on the spec

tests/Microsoft.Identity.Test.Common/TestCommon.cs
trwalke: This was handled by a test that created previously. I decided to refactor things a bit to make it more clear though.

trwalke: ADFS and B2C tests are unaffected by the CCS routing info.

jennyf19: do they return client info from the /authorize endpoint?

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
trwalke: If no client info is sent from the authorize endpoint and this value is null MSAL will not send any CCS routing info. We are only using the client info here for CCS which is not mandatory for the token exchange so it is ok if we don't send any routing headers when it is not available. No need to throw in this case. 

I'm not sure if ADFS or B2C will send the client info but I do not believe those technologies will have CCS support.

jennyf19: i understand that, but i'm saying to use the ClientInfo class that we already have, and that class will throw if it's null, so that's something to work around, but I would use that class because it has the correct error handling and test coverage. the other methods do not. Should see if other authorization servers return client info from the /authorize endpoint. And in which cases will this be null, if ever. I don't think it's common for it to be null from the /token endpoint.

trwalke: I have this code here which will prevent the ClientInfo class from constructing the object and throwing the exception. 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/cb136556765bbccb285b2ae737e3a337dae9b477/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs#L66

I figured this will workaround that issue.

jennyf19: i think you should still use the clientInfo class because it has the correct error handle for json issues.

jennyf19 (changes requested):
:clock1:

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
trwalke: I am using it here
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/cb136556765bbccb285b2ae737e3a337dae9b477/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs#L64-L83

If the client info is null then I don't create the client info class but if it is not null, then it is fed into the `ClientInfo.CreateFromJsonClientInfo.CreateFromJson` If there is a JSON issue it will be caught.

tests/Microsoft.Identity.Test.Common/TestCommon.cs
jennyf19: is client info returned by B2C and ADFS when going to the /authorize endpoint?

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
trwalke: I think there are some missing words or typos in your last comment. can you please rephrase why I should remove the method?

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeExchangeComponent.cs
```diff
@@ -59,5 +60,26 @@ public Task<MsalTokenResponse> FetchTokensAsync(CancellationToken cancellationTo
 
             return dict;
         }
+
+        private void AddCCSHeadersToTokenClient()
```
jennyf19, line 64: >CCS [](http://example.com/codeflow?start=24&length=3)

should be `AddCcsHeadersToTokenClient()`

tests/Microsoft.Identity.Test.Common/TestCommon.cs
trwalke: It is returned for B2C but not ADFS.

pmaytak (approved):
+ a comment about serializing `ClientInfo` in `AuthenticationResult`.

tests/Microsoft.Identity.Test.Common/TestCommon.cs
jennyf19: so is there any issue w/sending the ccs header to b2c? have we tested to see how b2c handles the additional headers? I assume they are ignored, but would be good to check.

trwalke: Yes, I checked them out. No issues with sending the additional CCS headers when they are not supported. We also have integration tests that send the headers and they are still working. if any issues arise we will know about it.

jennyf19 (approved):

src/client/Microsoft.Identity.Client/UI/AuthorizationResult.cs
```diff
@@ -138,6 +143,8 @@ public static AuthorizationResult FromStatus(AuthorizationStatus status, string
         [JsonProperty]
         public string CloudInstanceHost { get; set; }
 
+        public string ClientInfo { get; set; }
```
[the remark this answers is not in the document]
trwalke: Yeah I'm not sure it needs it. I haven't seen anything in the code indicating that it is needed and it is not a requirement for auth normally. Ill leave it as is until we need to change.

PR #2690, opened 2021-06-09
Measure GetHttpClient call as well

bgavrilMS:
Fixes # .

Measure GetHttpClient call in "time spent in HTTP"

**Testing**

**Performance impact**


jennyf19 (approved):

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -223,6 +221,8 @@ private static bool HasRetryAfterHeader(HttpResponse response)
 
                 Stopwatch sw = Stopwatch.StartNew();
 
+                HttpClient client = GetHttpClient();
+
```
jmprieur, line 225: I was more thinking logging the time it takes to get it?

PR #2704, opened 2021-06-15
Fixing performance test project

trwalke:
Fixes # .
Fix for performance test

Error: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> Microsoft.Identity.Client.MsalClientException: You did not define an authority at the application level, so it defaults to the https://login.microsoftonline.com/common.

However, the request is for a different cloud login.microsoft.com. This is not supported - the app and the request must target the same cloud.

Correct Authority

**Testing**
Manual testing of performance app

**Performance impact**
None

### DIFF
--- a/tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientLargeCacheTests.cs
+++ b/tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientLargeCacheTests.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Identity.Test.Performance
 
             await _ccaTokensDifferByTenant.AcquireTokenForClient(new[] { "scope" })
               .WithForceRefresh(false)
-              .WithAuthority($"https://login.microsoft.com/{tenant}")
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientLargeCacheTests.cs
```diff
@@ -61,7 +61,7 @@ public async Task AcquireTokenForClient_DifferentTenants_TestAsync()
 
             await _ccaTokensDifferByTenant.AcquireTokenForClient(new[] { "scope" })
               .WithForceRefresh(false)
-              .WithAuthority($"https://login.microsoft.com/{tenant}")
+              .WithAuthority($"https://login.microsoftonline.com/{tenant}")
```
jennyf19, line 64: nit: you could also define the authority on the application level to keep the same host, but if it still works, this is fine.

jennyf19 (approved):

pmaytak (approved):

Issue #2486, opened 2021-03-19
[Feature Request] [L] TokenCacheNotificationArgs should expose the token expiry so that cache can optimize eviction (client_credentials only)


jmprieur:
**Is your feature request related to a problem? Please describe.**
In token cache serializers for confidential client applications:
- for user flows, a RT is present in the cache and expiration cannot be determined, because tokens can be refreshed indefinitely.
- for AcquireTokenForClient however, there is no RT and the expiration of the cache can be computed as the max over the expiration of each access token  in cache.

**Describe the solution you'd like**

In `TokenCacheNotificationArgs` add a new property of type `SuggestedCacheExpiry`, and of the same type as the expiry in the token cache times (I suppose this is DateTimeOffset, but this needs to be checked).

This property will be computed by items.Max(item => item.ExpiresOnUnixTimestamp), and then convert that to a UtcDateTime.

```
class TokenCacheNotificationArgs
{
 // Other members

 /// <Summary>
 /// Suggested value of the expiry, to help determining the cache eviction time. 
 // This value is <b>only</b> set on the <code>OnAfterAcces</code> delegate, on a cache write
 /// operation (that is when <code>args.HasStateChanged<code> is <code>true</code>) and when the cache write 
 /// is triggered from the <code>AcquireTokenForClient<code> method. In all other cases it's <code>null<code>, as there is a refresh token, and therefore the
 /// access tokens are refreshable.
 /// </Summary> 
 UtcDateTime? SuggestedCacheExpiry {get; private set;}
```

jmprieur:
The problem is that the cache corresponding to an incoming token into a web app contains several OBO tokens. What should be the expiry? How do the OBO token expire vs the incoming token ? (for which MSAL.NET does not know the expiry)

bgavrilMS:
Needs more spec and a better understanding of OBO being able to refresh using RT or not.

jmprieur:
Some things to consider:
- For client credential tokens don't have a refresh token, so they could expire. The expiry of the cache could be the max of the expiry of the tokens in the cache
- @jmprieur and @jennyf19 to propose an end to end (including what would happen in the app/Ms.Id.Web)

jmprieur:
After doing an investigation it appears that given the access tokens can be refreshed for 90d, this does not make sense to expose an expiry.
Closing this issue.

jmprieur:
[...]

PR #2710, opened 2021-06-16
Add token expiry to token cache notification args

trwalke:
Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2486

Adds SuggestedCacheExpiry to the TokenCacheNotificationArgs

**Testing**
Manual testing
Unit test

**Performance impact**
Negligible performance impact

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -210,11 +210,11 @@ namespace Microsoft.Identity.Client
                     ITokenCacheInternal tokenCacheInternal = this;
                     if (tokenCacheInternal.IsTokenCacheSerialized())
                     {
-                        DateTime? cacheExpiry = null;
+                        DateTimeOffset? cacheExpiry = null;
 
                         if (requestParams.IsClientCredentialRequest)
                         {
-                            cacheExpiry = CanculateSuggestedCacheKey();
+                            cacheExpiry = CanculateSuggestedCacheExpiry();
                         }
 
                         var args = new TokenCacheNotificationArgs(
@@ -246,15 +246,11 @@ namespace Microsoft.Identity.Client
             }
         }
 
-        private DateTime? CanculateSuggestedCacheKey()
+        private DateTimeOffset? CanculateSuggestedCacheExpiry()
         {
             IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true).ToList();
             var unixCacheExpiry = tokenCacheItems.Max(item => item.ExpiresOnUnixTimestamp);
-
-            DateTime suggestedCacheExpiry = new DateTime(1970, 1, 1, 0, 0, 0, 0, System.DateTimeKind.Utc);
-            suggestedCacheExpiry = suggestedCacheExpiry.AddSeconds(Convert.ToDouble(unixCacheExpiry)).ToLocalTime();
-
-            return (DateTime?)suggestedCacheExpiry;
+            return (DateTimeOffset?)CoreHelpers.UnixTimestampStringToDateTime(unixCacheExpiry);
         }
 
         private string GetTenantId(IdToken idToken, AuthenticationRequestParameters requestParams)
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -246,11 +246,11 @@ namespace Microsoft.Identity.Client
             }
         }
 
-        private DateTimeOffset? CalculateSuggestedCacheExpiry()
+        private DateTimeOffset CalculateSuggestedCacheExpiry()
         {
             IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true);
             var unixCacheExpiry = tokenCacheItems.Max(item => item.ExpiresOnUnixTimestamp);
-            return (DateTimeOffset?)CoreHelpers.UnixTimestampStringToDateTime(unixCacheExpiry);
+            return (DateTimeOffset)CoreHelpers.UnixTimestampStringToDateTime(unixCacheExpiry);
         }
 
         private string GetTenantId(IdToken idToken, AuthenticationRequestParameters requestParams)
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -212,9 +212,9 @@ namespace Microsoft.Identity.Client
                     {
                         DateTimeOffset? cacheExpiry = null;
 
-                        if (requestParams.IsClientCredentialRequest)
+                        if (!_accessor.GetAllRefreshTokens().Any())
                         {
-                            cacheExpiry = CanculateSuggestedCacheExpiry();
+                            cacheExpiry = CalculateSuggestedCacheExpiry();
                         }
 
                         var args = new TokenCacheNotificationArgs(
@@ -246,9 +246,9 @@ namespace Microsoft.Identity.Client
             }
         }
 
-        private DateTimeOffset? CanculateSuggestedCacheExpiry()
+        private DateTimeOffset? CalculateSuggestedCacheExpiry()
         {
-            IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true).ToList();
+            IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true);
             var unixCacheExpiry = tokenCacheItems.Max(item => item.ExpiresOnUnixTimestamp);
             return (DateTimeOffset?)CoreHelpers.UnixTimestampStringToDateTime(unixCacheExpiry);
         }
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -210,6 +210,13 @@ namespace Microsoft.Identity.Client
                     ITokenCacheInternal tokenCacheInternal = this;
                     if (tokenCacheInternal.IsTokenCacheSerialized())
                     {
+                        DateTime? cacheExpiry = null;
+
+                        if (requestParams.IsClientCredentialRequest)
+                        {
+                            cacheExpiry = CanculateSuggestedCacheKey();
+                        }
+
                         var args = new TokenCacheNotificationArgs(
                             this,
                             ClientId,
@@ -218,7 +225,8 @@ namespace Microsoft.Identity.Client
                             tokenCacheInternal.IsApplicationCache,
                             tokenCacheInternal.HasTokensNoLocks(),
                             requestParams.RequestContext.UserCancellationToken,
-                            suggestedCacheKey: suggestedWebCacheKey);
+                            suggestedCacheKey: suggestedWebCacheKey,
+                            suggestedCacheExpiry: cacheExpiry);
 
                         Stopwatch sw = Stopwatch.StartNew();
                         await tokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);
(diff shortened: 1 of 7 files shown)


tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1080,6 +1080,55 @@ public void EnsurePublicApiSurfaceExistsOnInterface()
...
+
+                long expectedTimeDiff = 1000;
+                httpManager.CreateSuccessfulClientCredentialTokenResponseWithExpiry(expectedTimeDiff.ToString());
+
+                var appCacheAccess = app.AppTokenCache.RecordAccess();
+                var userCacheAccess = app.UserTokenCache.RecordAccess();
+
+                app.AppTokenCache.SetAfterAccess((args) => 
+                {
+                    if (args.SuggestedCacheExpiry != null)
```
pmaytak, line 1105: `if` could be replaced with an Assert that `SuggestedCacheExpiry` is not null (when HasStateChanged true).

pmaytak (approved):

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -238,6 +246,13 @@ public sealed partial class TokenCache : ITokenCacheInternal
             }
         }
 
+        private DateTimeOffset? CanculateSuggestedCacheExpiry()
+        {
+            IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true).ToList();
+            var unixCacheExpiry = tokenCacheItems.Max(item => item.ExpiresOnUnixTimestamp);
```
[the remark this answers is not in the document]
jmprieur: good catch, @pmaytak !
Does the string comparison give the same result as the numeric comparison in this particular case?

jmprieur (commented):
An important question about the string comparison.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -238,6 +246,13 @@ public sealed partial class TokenCache : ITokenCacheInternal
             }
         }
 
+        private DateTimeOffset? CanculateSuggestedCacheExpiry()
+        {
+            IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true).ToList();
```
bgavrilMS, line 251: Why "ToList()" ?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
bgavrilMS: Yes, this is a good point. ExpiresOnUnixTimestamp  should be a long, not a string. Further downstream we cast to `long` to convert to `DateTimeOffset`. No point in these UnixTimestamps to be stored as strings.

tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientLargeCacheTests.cs
```diff
@@ -78,11 +90,20 @@ public async Task AcquireTokenForClient_DifferentScopes_TestAsync()
...
               .ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// This is a test to validate the performance impact of calculating the TokenCacheNotificationArgs.SuggestedCacheExpiry when saving an access token.
+        /// </summary>
+        [Benchmark(Description = "Token cache expiry - O(n)")]
+        public async Task CaclulateTokenCacheExpiryTestTestAsync()
+        {
+            await _ccaTokensDifferByScope.AppTokenCacheInternal.SaveTokenResponseAsync(_requestParams, _response).ConfigureAwait(false);
```
bgavrilMS, line 104: Was under the impression that this test is not good enough because `SaveTokenResponseAsync` executes in under 100ms, and so the measurements are not reliable.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
```diff
@@ -171,6 +171,21 @@ public static void AddMockHandlerContentNotFound(this MockHttpManager httpManage
             return handler;
         }
 
+        public static MockHttpMessageHandler CreateSuccessfulClientCredentialTokenResponseWithExpiry(
```
bgavrilMS, line 174: Code duplication - this is alraedy present in `CreateSuccessfulClientCredentialTokenResponseMessage`. Why not just enhance the existing one?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1080,6 +1080,55 @@ public void EnsurePublicApiSurfaceExistsOnInterface()
...
+                                                              .WithAuthority(new Uri(ClientApplicationBase.DefaultAuthority), true)
+                                                              .WithRedirectUri(TestConstants.RedirectUri)
+                                                              .WithClientSecret(TestConstants.ClientSecret)
+                                                              .WithHttpManager(httpManager)
+                                                              .BuildConcrete();
+
+                long expectedTimeDiff = 1000;
+                httpManager.CreateSuccessfulClientCredentialTokenResponseWithExpiry(expectedTimeDiff.ToString());
+
+                var appCacheAccess = app.AppTokenCache.RecordAccess();
```
bgavrilMS, line 1100: why do you need to record cache access?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1080,6 +1080,55 @@ public void EnsurePublicApiSurfaceExistsOnInterface()
             PublicClientApplicationTests.CheckBuilderCommonMethods(byRefreshTokenBuilder);
         }
 
+        [TestMethod]
+        public async Task ConfidentialClientSuggestedExpiryAsync()
```
bgavrilMS, line 1084: I'm not sure this tests the right thing. We should have 2 tests around expiry: 

Token cache has AT1 with exp 1h, AT2 with exp 2h and AT3 with exp 3h

1. new AT comes in, with expiry 1.5h. expectated expiry ~3h 
2. new AT comes in with expirty 0.5h -> ~3h
3. new AT with 5h -> ~5h
etc.

bgavrilMS (changes requested):
better unit testing

tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientLargeCacheTests.cs
pmaytak: Don't we expect the cache to be small, since we partition by tenant Id / client Id. We could just add extra sizes to `TokenCacheSize` (10k, 100k).

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
bgavrilMS: `ExpiresOn ` is calculated on the fly using this `CoreHelpers.UnixTimestampStringToDateTime`. If you have a flat cache with 1000s of tokens, you execute this operation over and over. Not great. Maybe let's leave the string comparison there?

tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientLargeCacheTests.cs
bgavrilMS: Worth a try. I remember that when I tried to populate a very large cache (100k), the operation did not finish in the 1h I had dinner :).

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
trwalke: I was thinking this was too specific to add to a general used by many tests but on second thought it is the better option to enhance the existing one

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
trwalke: Not needed

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: I was getting strange errors from intellisense that prompted this. The errors are gone now so this will be removed

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -210,6 +210,13 @@ public sealed partial class TokenCache : ITokenCacheInternal
                     ITokenCacheInternal tokenCacheInternal = this;
                     if (tokenCacheInternal.IsTokenCacheSerialized())
                     {
+                        DateTimeOffset? cacheExpiry = null;
+
+                        if (requestParams.IsClientCredentialRequest)
```
bgavrilMS, line 215: Let's try to make the code more robust here. This is hardcoding the fact that client_creds flow is currentlty the only one where RTs are not involved. 

I think the code would be more robust if the conditition was "there are no RTs in this cache". Hence you can call `_accessor.GetAllRefreshTokens().Any()`

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -238,6 +246,13 @@ public sealed partial class TokenCache : ITokenCacheInternal
             }
         }
 
+        private DateTimeOffset? CalculateSuggestedCacheExpiry()
+        {
+            IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true);
+            var unixCacheExpiry = tokenCacheItems.Max(item => item.ExpiresOnUnixTimestamp);
+            return (DateTimeOffset?)CoreHelpers.UnixTimestampStringToDateTime(unixCacheExpiry);
```
bgavrilMS, line 253: nit: You can cast to DateTimeOffset

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1080,6 +1080,59 @@ public void EnsurePublicApiSurfaceExistsOnInterface()
...
+                httpManager.AddMockHandlerSuccessfulClientCredentialTokenResponseMessage(TestConstants.DefaultAccessToken, expectedTimeDiff.ToString());
+
+                TokenCacheHelper.PopulateDefaultAppTokenCache(app);
+                app.AppTokenCache.SetAfterAccess((args) => 
+                {
+                    if (args.HasStateChanged == true)
+                    {
+                        Assert.IsNotNull(args.SuggestedCacheExpiry);
+                        var timeDiff = args.SuggestedCacheExpiry.Value.ToUnixTimeSeconds() - DateTimeOffset.Now.ToUnixTimeSeconds();
+                        Assert.IsTrue(Enumerable.Range((int)expectedTimeDiff - allowedTimeVariation, (int)expectedTimeDiff).Contains((int)expectedTimeDiff) );
```
bgavrilMS, line 1112: This looks wrong. Enumerable.Range will generate all int numbers between N and M. Whean dealing with time, it's easier to use `TimeSpan`. You already have a helper like: 

`AreEqual(DateTimeOffset expected, DateTimeOffset actual, TimeSpan delta)`

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/6a54e7df0f005bab9138f70399007e9d07d45473/tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs#L35

tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientLargeCacheTests.cs
bgavrilMS: @trwalke - since test is inconclusive, please delete it.

bgavrilMS (approved):
Looks good, but minor issues remain with tests

PR #2713, opened 2021-06-17
Jennyf/ccs api

jennyf19:
Fixes # .
- Remove `WithExtraHttpHeaders` -> why? was not clear on how to use it and put a lot of work onto the developer to send the right info. 
- Headers were getting added twice:
image
- Headers should be included in OBO and create auth URL request.

Proposing:
`public AcquireTokenSilentParameterBuilder WithCcsRoutingHint(string userObjectIdentifier, string tenantIdentifier)`

On ATSilentParameterBuilder, OBOParameterBuilder, AuthCodeParameterBuilder, & URLParameterBuilder. 

Everything else should use the login hint or username in the request, or client_info, if we have it.

Add routing [header once](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2713/files#diff-e32790539dabb8da60283eb2e6b58350dc9b6076be91b3ea05ec40da50abd19dR182) w/priority to the ccsRoutingHint.

**Testing**
Added unit tests, see comments in PR re: ADFS tests.

Tested w/ID Web & have send fiddler info to ccs to see if we are actually going to ccs.

```
POST /7f58f645-c190-4ce5-9de4-e2b7acd2a6ab/oauth2/v2.0/token HTTP/1.1

x-anchormailbox: oid:2f5a54aa-b033-4796-a492-d0cd4e0b3f0c@7f58f645-c190-4ce5-9de4-e2b7acd2a6ab

GET /7f58f645-c190-4ce5-9de4-e2b7acd2a6ab/oauth2/v2.0/authorize?.........&x-anchormailbox=upn%3ajenny%40msidentitysamplestesting.onmicrosoft.com&......

x-anchormailbox	upn:jenny@msidentitysamplestesting.onmicrosoft.com

```

**Performance impact**
have not had time.


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
```diff
@@ -101,5 +102,26 @@ public AcquireTokenByAuthorizationCodeParameterBuilder WithPkceCodeVerifier(stri
...
             return this;
         }
+
+        /// <summary>
+        /// To help with resiliency, AAD Cached Credential Service (CCS) operates as an AAD backup.
+        /// </summary>
+        /// <param name="userObjectIdentifier">GUID which is unique to the user, parsed from the client_info.</param>
+        /// <param name="tenantIdentifier">GUID format of the tenant ID, parsed from the client_info.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenByAuthorizationCodeParameterBuilder WithCcsRoutingHint(string userObjectIdentifier, string tenantIdentifier)
```
jennyf19, line 112: I can't add to the common parameters directly, as we don't want this on AT for client..

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -98,7 +98,7 @@ protected override void EnrichTelemetryApiEvent(ApiEvent apiEvent)
             apiEvent.IsConfidentialClient = true;
         }
 
-        protected override SortedSet<string> GetOverridenScopes(ISet<string> inputScopes)
+        protected override SortedSet<string> GetOverriddenScopes(ISet<string> inputScopes)
```
jennyf19, line 101: typo. not related.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -76,29 +76,29 @@ public async Task ARLINGTON_ROPC_AAD_Async()
         public async Task ARLINGTON_ROPC_ADFS_Async()
         {
             var labResponse = await LabUserHelper.GetArlingtonADFSUserAsync().ConfigureAwait(false);
-            await RunHappyPathTestAsync(labResponse).ConfigureAwait(false);
+            await RunHappyPathTestAsync(labResponse, isAad: true).ConfigureAwait(false);
```
jennyf19, line 79: when i run the adfs tests, we use an assertion, so we don't have a username to send, so i'm not sure how the adfs tests worked previously.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -169,31 +167,52 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
...
             {
                 _oAuth2Client.AddHeader(PKeyAuthConstants.DeviceAuthHeaderName, PKeyAuthConstants.DeviceAuthHeaderValue);
             }
 
-            AddExtraHttpHeaders();
+            AddCcsHeader(additionalBodyParameters);
         }
 
-        private void AddExtraHttpHeaders()
+        private void AddCcsHeader(IDictionary<string, string> additionalBodyParameters)
```
trwalke, line 179: This was my thought initially when implementing this but  this is not the best way to do this. This is introducing a lot of if statememnts to the token client for flows that do not have CCS headers at all such as the confidential client flows. The token client is perf sensitive code which is why we originally opted to place the CCSHeader creation in the token request classes.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
trwalke: Not sure what you mean. The ropc requires the user to pass in the username so we simply take that value and construct the header with it. Even for adfs accounts

image

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -290,12 +292,6 @@ private async Task RunHappyPathTestAsync(LabResponse labResponse, string federat
             // 4) After successful log-in, remove the password line you added in with step 1, and run the integration test again.
         }
 
-        private void AssertCcsRoutingInformationIsSent(HttpSnifferClientFactory factory, LabResponse labResponse)
```
trwalke, line 293: The intention here is to make the code more readable by adding a method with a name that clearly identifies what is happening here. Otherwise it can be difficult for those not familiar with CCS to figureout exactly what we are doing.

trwalke:
"Headers should be included in OBO and create auth URL request."

I think it is a good idea to add this to the create auth URL request but what do you you mean by "Headers should be included in OBO"? The WithExtraHttpHeaders api was adding the headers to any request that used it including OBO and we don't have any user data to add to the OBO flow from the api.

jennyf19:
it doesn't seem to be included with OBO:


---
In reply to: [863479139](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2713#issuecomment-863479139) [](http://example.com/codeflow?ancestors=863479139,863479139)

trwalke:
> it doesn't seem to be included with OBO:
> image
> 
> In reply to: [863479139](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2713#issuecomment-863479139)

Yes. That is for data coming in from the api where the dev is only passing in the users access token which does not contain any CCS data. However, now that I think about it, Neha and I have worked on the RT flow for OBO so I think we may be able to add it.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
jennyf19: this is what i see when i tested:
image

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
jennyf19: true. i didn't see the need for such a complicated method when i could just add a one-liner.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
jennyf19: can you share the perf baseline? I didn't find it in your PR.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
trwalke: Ok, I think I see what is happening here. We used to construct the CCS header in the request which allows for the use of the _UserNamePasswordParameters which is available at the request level. Since this PR has it in the tokenClient. those values are not available.

I remember running into this and had to make this tweak:

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d8e77fe072d4326027ed7f75c97179a03ba0897e/src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs#L136-L139

jennyf19: ccs is not supported with ADFS correct? so we don't even need it for ADFS.

trwalke: Right.

trwalke: However, Im not 100% sure about hybrid environments. Ill reach out to the CCS contact

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
trwalke: The PR was added before the template was added to our PRs where we ask for Perf so I do not have it. We just tried to keep the overall number of conditionals as low as possible.

@bgavrilMS for input

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
jennyf19: sounds good. thanks.

src/client/Microsoft.Identity.Client/Internal/Constants.cs
```diff
@@ -27,7 +27,7 @@ internal static class Constants
 
         public const string UserRealmMsaDomainName = "live.com";
 
-        public const string CcsRoutingHintHeader = "X-AnchorMailbox";
+        public const string CcsRoutingHintHeader = "x-anchormailbox";
```
trwalke, line 30: Need to ocnfirm spec with CCS contact

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeRequestComponent.cs
```diff
@@ -193,8 +193,12 @@ private UriBuilder CreateInteractiveRequestParameters(IDictionary<string, string
                 //The CCS header is used by the CCS service to help route requests to resources in Azure during requests to speed up authentication.
                 //It consists of either the ObjectId.TenantId or the upn of the account signign in.
                 //See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2525
```
trwalke, line 195: This comment should be on the main "GetCcsHeader()" method that was moved to the token client

jennyf19: done. tx.

trwalke (changes requested):
Requesting changes

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -76,29 +76,29 @@ public async Task ARLINGTON_ROPC_AAD_Async()
...
-            await RunHappyPathTestAsync(labResponse).ConfigureAwait(false);
+            await RunHappyPathTestAsync(labResponse, isAad: true).ConfigureAwait(false);
         }
 
         [TestMethod]
         public async Task ROPC_ADFSv4Federated_Async()
         {
             var labResponse = await LabUserHelper.GetAdfsUserAsync(FederationProvider.AdfsV4, true).ConfigureAwait(false);
-            await RunHappyPathTestAsync(labResponse).ConfigureAwait(false);
+            await RunHappyPathTestAsync(labResponse, isAad: true).ConfigureAwait(false);
```
trwalke, line 86: shouldn't this be isAdfs?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
trwalke: Just got word from the CCS guys. ADFS is supported with CCS

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -80,6 +81,27 @@ public AcquireTokenSilentParameterBuilder WithForceRefresh(bool forceRefresh)
             return this;
         }
 
+        /// <summary>
+        /// To help with resiliency, AAD Cached Credential Service (CCS) operates as an AAD backup.
+        /// </summary>
+        /// <param name="userObjectIdentifier">GUID which is unique to the user, parsed from the client_info.</param>
+        /// <param name="tenantIdentifier">GUID format of the tenant ID, parsed from the client_info.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenSilentParameterBuilder WithCcsRoutingHint(string userObjectIdentifier, string tenantIdentifier)
```
bgavrilMS, line 90: Is this needed on AcquireTokenSilent? MSAL already has the uid.tid from the account object, which is mandatory in this API.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
jennyf19: interesting, when did they start supporting it? i asked last week, and unless i understood the [discussion wrong](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2687#discussion_r649578911), i thought it was not supported with ADFS.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
jennyf19: but your PR has the template:
image
so i wasn't sure what "minor" meant and if you had numbers for that.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
bgavrilMS, line 79: I don't think this "isAad" param is needed. These are all AAD tests. The only ADFS ROPC test is here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d1d3a6a88031c7123efa21ad6e663cafcd9efb7e/tests/Microsoft.Identity.Test.Integration.NetFx/HeadlessTests/UsernamePasswordIntegrationTests.cs#L106

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
trwalke: Oh, that was for the second PR where I was adding the client info to the /authorize endpint. There are 2 PRs for CCS

The original one is where most of the changes occured
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2590

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
trwalke: I think that may have been a misunderstanding on our part. There is an email from the CCS team just now that says it is supported.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
bgavrilMS: Yes, it's less about perf and more about having future-proof code. I thought that each request class should take the responsibility for figuring out the routing hint. 

In the end, I think that only ROPC and IWA are slightly different (because they can use the username), and all the rest of the requests are more or less the same - check if they have an Account object or a LoginHint object. So most of the code was a bunch of if / else anyway https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2713/files#diff-0eecc4a74d3964af4ff79f87b1ca6336318ff7d6bac101461d707c26ab3f27beL272. So I'm not objecting to this anymore, there'll be a ton of if / else either way.

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -127,6 +127,16 @@ public string Claims
         // TODO: ideally, this can come from the particular request instance and not be in RequestBase since it's not valid for all requests.
         public bool SendX5C { get; set; }
 
+        private string _ccsRoutingHint;
+        /// <summary>
+        /// OID and Tenant ID of the signed-in user for CCS routing.
+        /// </summary>
+        public string CcsRoutingHint
```
bgavrilMS, line 134: This is a bit tricky. I think I understand what you are doing but it's not obvious. `_ccsRoutingHint` is the one from the app developer. `_commonParameters.CcsRoutingHint` is the one that gets calculated from the users right?

I'd maybe rename `_ccsRoutingHint ` with `_userCcsRoutingHint` or smth like that. And not use `commonParameters.CcsRoutingHint` here at all, since it's not technically a request param.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
pmaytak: Plus for confidential client, the `_requestParams.CcsRoutingHint` will be null, so this whole `if` will be skipped.

src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs
```diff
@@ -169,31 +167,55 @@ private void AddBodyParamsAndHeaders(IDictionary<string, string> additionalBodyP
...
+            string ccsHeaderHint = string.Empty;
+            if (!string.IsNullOrEmpty(_requestParams.CcsRoutingHint))
+            {
+                ccsHeaderHint = $"oid:{_requestParams.CcsRoutingHint}";
+            }
+
+            else if(_requestParams?.Account?.HomeAccountId != null)
             {
-                foreach (KeyValuePair<string, string> pair in _requestParams.ExtraHttpHeaders)
+                if (!string.IsNullOrEmpty(_requestParams.Account.HomeAccountId.Identifier))
```
pmaytak, line 192: Perhaps these two `ifs` can be combined into:
`else if (!string.IsNullOrEmpty(_requestParams?.Account?.HomeAccountId?.Identifier))`

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -691,6 +686,97 @@ public async Task GetAuthorizationRequestUrlNoRedirectUriTestAsync()
...
+        public async Task GetAuthorizationRequestUrl_IgnoreLoginHint_UseCcsRoutingHint_TestAsync()
+        {
+            using (var httpManager = new MockHttpManager())
+            {
+                ConfidentialClientApplication app = CreateCca(httpManager);
+
+                var uri = await app
+                    .GetAuthorizationRequestUrl(TestConstants.s_scope)
+                    .WithLoginHint(TestConstants.DisplayableId)
+                    .WithCcsRoutingHint("oid", "tid")
```
pmaytak, line 699: Should add to the method comment that routing hint will take precedence over login hint.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
bgavrilMS: Federated ADFS is not ADFS. It's AAD. All these tests are AAD. There is a "direct ADFS" test below, that does not use RunHappyPathTestAsync

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenParameterBuilderExtensions.cs
```diff
@@ -1,25 +0,0 @@
...
-    /// </summary>
-    public static class AcquireTokenParameterBuilderExtensions
-    {
-        /// <summary>
-        /// Adds additional Http Headers to the token request.
-        /// </summary>
-        /// <param name="builder">Parameter builder for a acquiring tokens.</param>
-        /// <param name="extraHttpHeaders">additional Http Headers to add to the token request.</param>
-        /// <returns></returns>
-        public static T WithExtraHttpHeaders<T>(this AbstractAcquireTokenParameterBuilder<T> builder, IDictionary<string, string> extraHttpHeaders)
```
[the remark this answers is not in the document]
bgavrilMS: No, we give full access to the HTTP client so customers wanting to manipulate the http flow can do it that way.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
```diff
@@ -85,6 +87,27 @@ public AcquireTokenOnBehalfOfParameterBuilder WithForceRefresh(bool forceRefresh
             return this;
         }
 
+        /// <summary>
+        /// To help with resiliency, AAD Cached Credential Service (CCS) operates as an AAD backup.
+        /// </summary>
+        /// <param name="userObjectIdentifier">GUID which is unique to the user, parsed from the client_info.</param>
+        /// <param name="tenantIdentifier">GUID format of the tenant ID, parsed from the client_info.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenOnBehalfOfParameterBuilder WithCcsRoutingHint(string userObjectIdentifier, string tenantIdentifier)
```
neha-bhargava, line 96: This code is same as [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2713/files#diff-5197c1f0bc80f2a410918064894a2a01956c5d6426f84d680ff1492a2204f508R112). Can this be reused?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
bgavrilMS, line 112: What about UPN routing hint? `WithCcsRoutingHint(string upn)`

jennyf19: the other ones would use .WithLoginHint() if it's passed in. no need to add UPN if LoginHint is not available on that builder.

bgavrilMS: Good point.

PR #2717, opened 2021-06-18
Bogavril/tc logs

### LINKED ISSUE #2716: [Bug] Log messages about token filters are recorded multiple times
4.32.1

Verbose log messages such as "Filtering by tenant_Id. Items before: 3" are logged several times.

bgavrilMS:
Fixes #2716 

Replace IEnumerable with IReadOnly list to avoid re-execution of filters 
Add more logging
Short-circuit some logic if token_count = 0

**Testing**
perf testing

**Performance impact**
FindAccessTokenAsync for client_creds flow (before and after) does not show any meaningful difference.

before
|                     Method | TokenCacheSize |         Mean |        Error |       StdDev |
|--------------------------- |--------------- |-------------:|-------------:|-------------:|
| 'Different tenants - O(1)' |           1000 |     73.24 us |     1.372 us |     1.468 us |
|  'Different scopes - O(n)' |           1000 | 58,334.41 us | 1,104.858 us | 2,647.169 us |

after
|                     Method | TokenCacheSize |         Mean |        Error |       StdDev |
|--------------------------- |--------------- |-------------:|-------------:|-------------:|
| 'Different tenants - O(1)' |           1000 |     76.41 us |     0.513 us |     0.429 us |
|  'Different scopes - O(n)' |           1000 | 58,034.85 us | 1,149.229 us | 1,951.479 us |

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/ITokenCacheAccessor.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/ITokenCacheAccessor.cs
@@ -37,15 +37,15 @@ namespace Microsoft.Identity.Client.Cache
 
         void DeleteAccount(MsalAccountCacheKey cacheKey);
 
-        IEnumerable<MsalAccessTokenCacheItem> GetAllAccessTokens();
+        IReadOnlyList<MsalAccessTokenCacheItem> GetAllAccessTokens();
 
-        IEnumerable<MsalRefreshTokenCacheItem> GetAllRefreshTokens();
+        IReadOnlyList<MsalRefreshTokenCacheItem> GetAllRefreshTokens();
 
-        IEnumerable<MsalIdTokenCacheItem> GetAllIdTokens();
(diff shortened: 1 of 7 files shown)


jennyf19 (approved):
:shipit:

SameerK-MSFT (approved):

pmaytak (changes requested):
Some thoughts/questions:
- Have you tested with more than 1k tokens in the cache? Wonder if 1k is too small, maybe we need larger cache to see variance better.
- The perf test only tests ATs in the cache, right? These changes affect RTs too. Should we have a test for RT search flow too?
- Currently the perf test differentiates tokens by tenant and scope. Does this test real-world scenario? In order to get a more realistic test - how would the tokens in real-world app be different? My guess the environment, tenant (because of partitioning), token type are probably the same for all ATs. What about filtering by user assertions? (If the assertion doesn't match, we would hit all the filters?)

pmaytak (approved):

PR #2718, opened 2021-06-18
Remove the cache penalty delay from code

neha-bhargava:
Remove the changes to implement cache delay for perf testing from MSAL code and add it to the perf tests.


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Cache/CacheImpl/AbstractDistribuitedCacheSerializer.cs
```diff
@@ -51,11 +51,11 @@ private void OnAfterAccess(TokenCacheNotificationArgs args)
             }
```
jmprieur, line 51: There is a typo in the name of the file.

PR #2721, opened 2021-06-21
Propagate cancellation to cache semaphore and remove app id

bgavrilMS:
Fixes # .

- propagate cancellation token to `waitAynnc`
- remove app hash ID since it is not reliable.

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Identity.Client.Cache
                     };
 
                     _requestParams.RequestContext.Logger.Verbose("[Cache Session Manager] Waiting for cache semaphore");
-                    await TokenCacheInternal.Semaphore.WaitAsync().ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
```diff
@@ -105,7 +105,7 @@ private async Task RefreshCacheForReadOperationsAsync(CacheEvent.TokenTypes cach
                     };
 
                     _requestParams.RequestContext.Logger.Verbose("[Cache Session Manager] Waiting for cache semaphore");
-                    await TokenCacheInternal.Semaphore.WaitAsync().ConfigureAwait(false);
+                    await TokenCacheInternal.Semaphore.WaitAsync(_requestParams.RequestContext.UserCancellationToken).ConfigureAwait(false);
```
bgavrilMS, line 108: @jennyf19 - this is already in the drop we gave MyApps

jennyf19: i think so.

jennyf19 (approved):
:shipit:

neha-bhargava (approved):

pmaytak (approved):

PR #2722, opened 2021-06-21
Update EditorConfig to add copyright file header for new files.

pmaytak:
Fixes # .

Updated EditorConfig to add copyright header to new files. 
Note: If the file doesn't have a header, set the rule to warning, but it doesn't seem to work for existing files.


bgavrilMS (approved):
Nice!

neha-bhargava (approved):
👍

PR #2724, opened 2021-06-22
Add framework runtime to logs

### LINKED ISSUE #2559: [Feature Request] [M] Add runtime info to logs
Add runtime information, e.g. .NET Core 3.4 to the logs. We currently only log the target framework, i.e. the tfm of MSAL used.

pmaytak:
Fixes #2559.

Add framework runtime to logs. `RuntimeInformation.FrameworkDescription` is available on all platforms except .NET Framework 4.7 and lower.
References:
[Determine which .NET Framework versions are installed](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#query-the-registry-using-code)
[.NET API Catalog](https://apisof.net/catalog/d6361ded-7f4c-a4fa-3049-c19c6bf1761c)

**Testing**
Manually tested targets below (except iOS, no device to test with).
.NET Core
image

.NET 5
image

.NET Fx 472, 461
image

Android
image

UWP
image

**Performance impact**
Runtime version is a static retrieved once.


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
```diff
@@ -152,6 +152,52 @@ public static bool IsWin10OrServerEquivalent()
             return s_win10OrServerEquivalentLazy.Value;
         }
 
+        private static string GetRuntimeVersionInternal()
```
bgavrilMS, line 155: Would it not be better for this logic to be in PlatformProxy? The desktop helper is not meant for mobile / uwp.

bgavrilMS (approved):
Looks ok, maybe move the logic from DesktopOsHelper to platform proxy

trwalke (approved):
LGTM

trwalke:
Question, did you try testing with iOS?

pmaytak:
> Question, did you try testing with iOS?

No, I haven't (no iOS device) but it should work the same as for Android.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
pmaytak: I actually initially did go that way but because RuntimeInfo is not available in some targets, still had to do # ifs But yea, if DesktopOsHelper is not the right place semantically, I can move it.

PR #2730, opened 2021-06-23
Add tenant profiles to account

neha-bhargava:
How the API looks like now: 

```csharp
                            AuthenticationResult authResult = await interactiveBuilder.ExecuteAsync().ConfigureAwait(false);
                            ClaimsPrincipal idTokenClaims = authResult.ClaimsPrincipal;

                            var accounts2 = await pca.GetAccountsAsync().ConfigureAwait(false);
                            foreach (var acc in accounts2)
                            {
                                Console.WriteLine($"Account for {acc.Username}");
                                foreach (var tp in acc.GetTenantProfiles())
                                {
                                    Console.WriteLine($"Tenant Profile in tenant {tp.TenantId} " +
                                        $"is home tenant? {tp.IsHomeTenant} " +
                                        $"claims {tp.ClaimsPrincipal.Claims.Count()}");
                                }
                            }
```


src/client/Microsoft.Identity.Client/IAccount.cs
```diff
@@ -33,5 +33,10 @@ public interface IAccount
         /// </summary>
         /// <remarks>Can be null, for example if this account was migrated to MSAL.NET from ADAL.NET v3's token cache</remarks>
         AccountId HomeAccountId { get; }
+
+        /// <summary>
+        /// Account can have multiple tenants. TenantProfiles is a map of tenant id corresponding tenant profile.
```
jmprieur, line 38:
```suggestion
        /// The same account can exist in its home tenant, but also be guest in other tenant. 
        /// TenantProfiles is a map of information of the account in each tenant. They key is the tenant ID.
```

src/client/Microsoft.Identity.Client/ITenantProfile.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+using System.Security.Claims;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// The ITenantProfile interface represents information a tenant in an account.
```
jmprieur, line 10:
```suggestion
    /// The ITenantProfile interface represents information about the account in a tenant specific tenant.
```

src/client/Microsoft.Identity.Client/ITenantProfile.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// There can be multiple tenant profiles in an account.
```
jmprieur, line 11:
```suggestion
    /// There can be multiple tenant profiles for an account (home tenant and guest accounts).
```

src/client/Microsoft.Identity.Client/ITenantProfile.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        string Oid { get; }
+
+        /// <summary>
+        /// Gets a string containing the tenant id.
+        /// </summary>
+        string TenantId { get; }
+
+        /// <summary>
+        /// Gets claims in an id token.
```
jmprieur, line 26:
```suggestion
        /// Gets claims presents in the ID Token.
```

src/client/Microsoft.Identity.Client/ITenantProfile.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        ClaimsPrincipal Claims { get; }
+
+        /// <summary>
+        /// Is the tenant the home tenant of the user?
```
jmprieur, line 31:
```suggestion
        /// Is the tenant the home tenant of the account?
```

src/client/Microsoft.Identity.Client/IAccount.cs
```diff
@@ -33,5 +33,10 @@ public interface IAccount
...
+        /// </summary>
+        IDictionary<string, ITenantProfile> TenantProfiles { get; }
```
jmprieur, line 40: Add a `<remark>`, that only tenants for which a token was acquired will be available in the TenantProfiles property 
In other words, an account can be in n tenants, but `TenantProfiles` can contain m items (m < n), because tokens were only requested for these m tenants (using `.WithTenant()`)

jmprieur (commented):
Thanks @neha-bhargava 
I've added a few suggestions for the XML comments.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -117,9 +120,11 @@ public partial class AuthenticationResult
             IAuthenticationScheme authenticationScheme,
             Guid correlationID,
             TokenSource tokenSource, 
-            ApiEvent apiEvent)
+            ApiEvent apiEvent,
+            ICacheSessionManager cacheManager)
         {
```
jennyf19, line 125: Looking at the spec, it looks like AuthenticationResult should include `.GetIdTokenClaims()` which would return just the IdToken Claims. I think id_token claims and tenant profiles are not related features, so a customer might want the claims but not necessary the tenant profile.

src/client/Microsoft.Identity.Client/TenantProfile.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public ClaimsPrincipal Claims { get; }
+
+        public bool IsHomeTenant { get; }
+
+        public override string ToString()
+        {
+            return string.Format(
+            CultureInfo.CurrentCulture,
+            "Tenant Id {1}", TenantId);
```
jennyf19, line 42: >enant Id {1} [](http://example.com/codeflow?start=14&length=12)

should be `{0}` and is it used anywhere? if for logging, tenant Id is PII.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -128,18 +133,31 @@ public partial class AuthenticationResult
...
                     username = msalIdTokenCacheItem.IsAdfs ?
                         msalIdTokenCacheItem?.IdToken.Upn :
                         msalIdTokenCacheItem?.IdToken?.PreferredUsername;
+
+                    if (tenantProfiles != null)
+                    {
+                        tenantProfiles[msalIdTokenCacheItem.TenantId] = new TenantProfile(
+                            msalIdTokenCacheItem.IdToken.ObjectId,
+                            msalIdTokenCacheItem.TenantId,
+                            msalIdTokenCacheItem.Secret,
```
jennyf19, line 151: >.Secret [](http://example.com/codeflow?start=48&length=7)

why use Secret and not `msalIdTokenCacheItem.IdToken`?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -128,18 +133,31 @@ public partial class AuthenticationResult
...
                 if (msalIdTokenCacheItem != null)
                 {
                     username = msalIdTokenCacheItem.IsAdfs ?
                         msalIdTokenCacheItem?.IdToken.Upn :
                         msalIdTokenCacheItem?.IdToken?.PreferredUsername;
+
+                    if (tenantProfiles != null)
+                    {
+                        tenantProfiles[msalIdTokenCacheItem.TenantId] = new TenantProfile(
+                            msalIdTokenCacheItem.IdToken.ObjectId,
```
jennyf19, line 149: >msalIdTokenCacheItem.IdToken [](http://example.com/codeflow?start=28&length=28)

null check.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -865,6 +876,70 @@ async Task<IEnumerable<IAccount>> ITokenCacheInternal.GetAccountsAsync(Authentic
             return accounts;
         }
 
+        async Task<IDictionary<string, ITenantProfile>> ITokenCacheInternal.GetTenantProfilesAsync(AuthenticationRequestParameters requestParameters, string homeAccountId)
+        {
```
jennyf19, line 880: this is very similar to GetAccountsAsync(). do we need the ADAL part and to do instance discovery? can the shared code be moved to another method? i would think we would have already done instance dis. do you know the perf impact?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -865,6 +876,70 @@ async Task<IEnumerable<IAccount>> ITokenCacheInternal.GetAccountsAsync(Authentic
...
+            if (string.IsNullOrEmpty(homeAccountId) || string.IsNullOrEmpty(AccountId.ParseFromString(homeAccountId).TenantId))
+            {
+                return null;
+            }
+
+            IDictionary<string, ITenantProfile> tenantProfiles = new Dictionary<string, ITenantProfile>();
+            bool filterByClientId = !_featureFlags.IsFociEnabled;
+
+            // Get id tokens in the cache and filter them with environment and home account id to get the id tokens for the account.
+            IEnumerable<MsalIdTokenCacheItem> idTokenCacheItems = GetAllIdTokensWithNoLocks(filterByClientId);
```
jennyf19, line 923: i think we are already doing this logic elsewhere in the this class, so maybe just add the part about returning the tenant profile? cache look ups are expensive.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -865,6 +876,70 @@ async Task<IEnumerable<IAccount>> ITokenCacheInternal.GetAccountsAsync(Authentic
...
+
+            return tenantProfiles;
+        }
+
+        private ITenantProfile GetTenantProfile(MsalIdTokenCacheItem msalIdTokenCacheItem)
+        {
+            AccountId homeAccountId = AccountId.ParseFromString(msalIdTokenCacheItem.HomeAccountId);
+            bool isHomeTenant = homeAccountId.TenantId.Equals(msalIdTokenCacheItem.TenantId);
+            
+            return new TenantProfile(msalIdTokenCacheItem.IdToken.ObjectId, msalIdTokenCacheItem.TenantId, msalIdTokenCacheItem.Secret, isHomeTenant);
```
jennyf19, line 940: null check here, and can you use IdToken instead of Secret?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
jennyf19: if you can pick up the instance dis info elsewhere, this won't need to be async.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -166,6 +166,7 @@ public static string iOSBrokerKeySaveFailed(string keyChainResult)
         public const string NonHttpsRedirectNotSupported = "Non-HTTPS URL redirect is not supported in webview. ";
 
         public const string IDTokenMustHaveTwoParts = "ID Token must contain at least 2 parts. ";
+        public const string IDTokenMustHaveThreeParts = "ID Token must contain at least 3 parts. ";
```
jennyf19, line 169: maybe something like "IdToken must be of form JWT" which implies 3 parts, but is more specific. OIDC specifies the format as a JWT.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -56,7 +56,8 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                         AuthenticationRequestParameters.AuthenticationScheme,
                         AuthenticationRequestParameters.RequestContext.CorrelationId,
                         TokenSource.Cache,
-                        AuthenticationRequestParameters.RequestContext.ApiEvent);
+                        AuthenticationRequestParameters.RequestContext.ApiEvent,
+                        CacheManager);
```
jennyf19, line 60: >CacheManager [](http://example.com/codeflow?start=24&length=12)

i don't think we should return a CacheManager. plus the AT is already returned above in cachedAccessTokenItem, line 46. is there another way to access the tenant profiles?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -61,15 +62,15 @@ public void TestInitialize()
         public async Task ROPC_AAD_Async()
         {
             var labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
-            await RunHappyPathTestAsync(labResponse).ConfigureAwait(false);
+            await RunHappyPathTestAsync(labResponse, authorityType: AuthorityType.Aad).ConfigureAwait(false);
```
jennyf19, line 65: why does the authority have to specified?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -285,29 +291,41 @@ AuthenticationResult authResult
...
+            AssertTelemetryHeaders(factory, false, labResponse);
+            AssertCCSRoutingInformationIsSent(factory, labResponse);
+
+            return authResult;
+        }
+
+        
+        private void AssertTenantProfiles(IDictionary<string, ITenantProfile> tenantProfiles, string tenantId)
+        {
+            Assert.IsNotNull(tenantProfiles);
```
jennyf19, line 322: i would separate the idtoken claims, as a separate API. it's not easily found, and the reason for having them is separate from having the tenant profiles.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/AadAuthorityTests.cs
```diff
@@ -270,6 +270,7 @@ public void AuthorityCustomPortTest()
 
                 harness.HttpManager.AddSuccessTokenResponseMockHandlerForPost(customPortAuthority);
                 harness.HttpManager.AddInstanceDiscoveryMockHandler(customPortAuthority);
+                harness.HttpManager.AddInstanceDiscoveryMockHandler(customPortAuthority);
```
jennyf19, line 274: this is going to be a problem. we shouldn't need to do instance dis twice.

jennyf19 (changes requested):
:clock1:

jmprieur (commented):
I proposed a few changes

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/AadAuthorityTests.cs
trwalke: The second instance discovery call should pull from the cache but I wonder how much that will affect perf. I agree it should be avoided if possible

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -268,13 +269,18 @@ private async Task RunHappyPathTestAsync(LabResponse labResponse, string federat
...
                     labResponse,
                     factory,
                     msalPublicClient,
                     federationMetadata,
                     CorrelationId).ConfigureAwait(false);
+            if (authorityType == AuthorityType.Aad)
+            {
+                AssertTenantProfiles(authResult.Account.TenantProfiles, authResult.TenantId);
+
+            }
```
trwalke, line 283: nit: extra line above

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -268,13 +269,18 @@ private async Task RunHappyPathTestAsync(LabResponse labResponse, string federat
...
 
-            AuthenticationResult authResult
+            AuthenticationResult authResult
                 = await GetAuthenticationResultWithAssertAsync(
                     labResponse,
                     factory,
                     msalPublicClient,
                     federationMetadata,
                     CorrelationId).ConfigureAwait(false);
+            if (authorityType == AuthorityType.Aad)
```
trwalke, line 279: This should also work for government clouds too right?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -128,18 +133,31 @@ public partial class AuthenticationResult
 
             if (homeAccountId != null)
             {
+                // This method call is to an async method from this constructor but the network call in the method will ideally never happen.
+                var tenantProfiles = Task.Run(async () => await _cacheManager.GetTenantProfilesAsync(homeAccountId).ConfigureAwait(false)).GetAwaiter().GetResult();
```
jmprieur, line 137: would we want to compute this lazily?
for instance have delegate in account which is called the first time a customer requests tenant profiles?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
neha-bhargava: This is not a part of this issue. As discussed earlier, once this is done we will decide how to expose the claims. @jmprieur to comment

jmprieur: I created a small extension method that computes `GetIdTokenClaims()`:

```CSharp
 public static class AuthenticationResultExtensions
    {
        public static IEnumerable<Claim> GetIdTokenClaims(this AuthenticationResult authenticationResult)
        {
            return authenticationResult.Account.TenantProfiles[authenticationResult.TenantId].Claims.Claims;
        }
    }
```

It works most of the times, but it does not work if we use `AcquireTokenSilent` with a `.WithAuthority()` calls which is called with a tenant expressed as a **domain** instead of the GUID (which is perfectly legit). Indeed, the tenantID in the tenant profile will be the domain name, not the tenant ID, whereas the authenticationResult.TenantId is the GUID

```CSharp
  result = await app.AcquireTokenSilent(new[] { "user.read" }, result.Account)
                .WithAuthority(app.Authority.Replace("common", "msidentitysamplestesting.onmicrosoft.com"))
                .ExecuteAsync();
```

We might have a little bug here:
cc: @jennyf19 @neha-bhargava

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
neha-bhargava: Secret is the id token before parsing. We need this to get the claims. IdToken in the cache is the IdToken object [see](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/698f0fd1b0daba783c9028e099595f10a0d52da2/src/client/Microsoft.Identity.Client/Cache/Items/MsalIdTokenCacheItem.cs#L65)

src/client/Microsoft.Identity.Client/ITenantProfile.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Gets a string containing the tenant id.
+        /// </summary>
+        string TenantId { get; }
+
+        /// <summary>
+        /// Gets claims present in the ID Token.
+        /// </summary>
+        ClaimsPrincipal Claims { get; }
```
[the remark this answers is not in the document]
neha-bhargava: Will rename it to ClaimsPrincipal

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
neha-bhargava: Another way would be to have a delegate in account instead as mentioned by @jmprieur in one of the comments above. But I have yet to evaluate how that can be done. We need the CacheManager to get the tenant profiles for IdTokens in the cache. I can't think of any other way.

src/client/Microsoft.Identity.Client/TenantProfile.cs
neha-bhargava: It is not used anywhere. I think I can remove this. Thanks

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: Instance discovery part cannot be moved elsewhere but by the time we reach here instance discovery is already done and cached. Since we are querying the cached instance discovery there will be no perf impact.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: No secret is the idtoken in jwt format which contains all the claims. IdToken is the class which parses the secret and stores some of the claims.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
neha-bhargava: Yes, any AAD.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
neha-bhargava: Added this since tenant profiles will only be populated for AAD accounts. Wanted to test that

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
neha-bhargava: @jennyf19 what do you mean by having a separate API for token claims? Not have them in the tenant profiles? 

I think having them in the tenant profiles as a dictionary is fine since the claims are for an id token. Also, point of having claims as a dictionary here is to get all the claims that exists in an id token and those can vary.

Aside, we already have [IdToken.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/698f0fd1b0daba783c9028e099595f10a0d52da2/src/client/Microsoft.Identity.Client/Internal/IdToken.cs#L9) which we use as an internal class for IdTokenCacheItems.

cc: @jmprieur

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/AadAuthorityTests.cs
jennyf19: i don't remember how this is set-up...why do we need a mock if it's coming from the cache?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
jennyf19: see earlier conversation where @jmprieur provided code to get the IdTokenClaims separately.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
jennyf19: right, but what if a customer doesn't set it. what will happen? do we have a test for that? I would add a test for that and leave these tests as-is. if they are AAD, it should work.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
jennyf19: ok. thanks.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/AadAuthorityTests.cs
pmaytak: Maybe it's because if authority has port, we don't cache instance discovery properly? Might need investigation on this?

neha-bhargava: Yes, that is right. This is the only case where instance discovery is not used from cache. Needs more investigation.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
bgavrilMS: We can have the Request objects fetch the data needed for creating the tenant profiles, let me have a look...

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
bgavrilMS: Was thinking that ClaimsPrincipal has everything, is there a need for IEnumerablem<Claim> ?

src/client/Microsoft.Identity.Client/Cache/Items/MsalIdTokenCacheItem.cs
```diff
@@ -53,13 +53,9 @@ internal MsalIdTokenCacheItem()
         }
 
     
-        internal bool IsAdfs { get; set; }
+        internal bool IsAdfs { get; set; } // TODO: get rid of this crap
```
henrik-me, line 56: >// TODO: get rid of this crap [](http://example.com/codeflow?start=43&length=29)

? what does this comment suggest? pls. be more specific.

src/client/Microsoft.Identity.Client/Internal/IdToken.cs
```diff
@@ -21,42 +26,39 @@ internal static class IdTokenClaim
...
+        public string GetUniqueId()
+        {
+            return ObjectId ?? Subject;
+        }
 
-        [JsonProperty(PropertyName = IdTokenClaim.FamilyName)]
-        public string FamilyName { get; set; }
+        public ClaimsPrincipal ClaimsPrincipal { get; private set; }
 
         public static IdToken Parse(string idToken)
```
henrik-me, line 63: >Parse [](http://example.com/codeflow?start=30&length=5)

Very expensive operations. it should be ensured this is only done if really needed. Overall also should be considered to use SPAN's and reduce the amount of strings created.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.cs
```diff
@@ -75,8 +75,8 @@ public async Task WithSecret_TestAsync(Cloud cloud, RunOn runOn)
         }
 
         [TestMethod]
-        [DataRow(Cloud.Public, RunOn.NetFx | RunOn.NetCore)]
-        [DataRow(Cloud.Adfs, RunOn.NetFx)]
+        //[DataRow(Cloud.Public, RunOn.NetFx | RunOn.NetCore)]
```
henrik-me, line 78: >//[DataRow(Cloud.Public, RunOn.NetFx | RunOn.NetCore)] [](http://example.com/codeflow?start=8&length=54)

why not on public?

bgavrilMS: Accidental commit of this file, I will revert it.

src/client/Microsoft.Identity.Client/Internal/IdToken.cs
bgavrilMS: Would be great, but we need to move to .net core or net5 only :) ... 
We aren't doing a LOT of extra work, because we were already parsing the idtoken and extracting `tid`, `sub`, `name`, etc. Now we are parsing ALL the claims. 

The problem is that each tenant profile will have 1 id token so 10 profiles => 10 token parsings. I'll look at how to eliminate that. 

This is still work in progress, so I'll mark the PR as "draft".

src/client/Microsoft.Identity.Client/Cache/Items/MsalIdTokenCacheItem.cs
bgavrilMS: Will do :). This was not added by our team, btw. It's wrong because when you load a token from the cache, you don't know the authority so `IsAdfs` is always false. It changes how we report UPN and a few other things.

src/client/Microsoft.Identity.Client/Internal/IdToken.cs
bgavrilMS: Implemented the lazy approach, id tokens will only be parsed when accessing things like `TenantProfile.ObjectId` or `TenantProfile.ClaimsPrincipal` etc.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: I modified this. Neha has a good point that instance discovery needs to be performed, otherwise we risk not being able to load some id tokens. 

However, there is no need for ADAL legacy at all. Only RTs from ADAL are usable. Tenant Profiles are a nice view over MSAL ID tokens.

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenParsingTests.cs
```diff
@@ -0,0 +1,120 @@
...
+            Assert.AreEqual("1538538422", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "nbf").Value);
+            Assert.AreEqual("1538542322", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "exp").Value);
+            Assert.AreEqual("Cloud IDLAB Basic User", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "name").Value);
+            Assert.AreEqual("9f4880d8-80ba-4c40-97bc-f7a23c703084", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "oid").Value);
+            Assert.AreEqual("idlab@msidlab4.onmicrosoft.com", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "preferred_username").Value);
+            Assert.AreEqual("Y6YkBdHNNLHNmTKel9KhRz8wrasxdLRFiP14BRPWrn4", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "sub").Value);
+            Assert.AreEqual("f645ad92-e38d-4d1a-b510-d1b09a74a8ca", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "tid").Value);
+            Assert.AreEqual("6nciX02SMki9k73-F1sZAA", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "uti").Value);
+            Assert.AreEqual("2.0", parsedToken.ClaimsPrincipal.Claims.Single(c => c.Type == "ver").Value);
```
bgavrilMS, line 57: tedium...

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/AadAuthorityTests.cs
bgavrilMS: I still need to investigate this. It is not ok to go to the network twice.

bgavrilMS: Good catch @jennyf19 , appreciate that you review in this level of detail!

So ... the test in its original form is wrong and the change Neha did is wrong :). A network discovery call should happen ONCE and only ONCE. There is a bug however around non-standard clouds that @trwalke is just fixing ([PR](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2760/files))

I will fix the test, but it will continue to "fail" until @trwalke merges his PR.

src/client/Microsoft.Identity.Client/Account.cs
```diff
@@ -20,25 +20,30 @@ internal sealed class Account : IAccount
...
         public string Environment { get; }
 
         public AccountId HomeAccountId { get; }
-       
 
-        internal IDictionary<string, string> WamAccountIds
-        {
-            get;
-        }
+        public IDictionary<string, TenantProfile> TenantProfiles { get; }
```
pmaytak, line 44: Should this also be `IEnumerable<TenantProfile>` for consistency?

bgavrilMS: Internally, we construct it as a dictionary, to avoid duplicates. So it is consistent with MSAL internals.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
bgavrilMS: Have changed the API, it now looks like this: 

```csharp
var authResult = await interactiveBuilder.ExecuteAsync().ConfigureAwait(false);
ClaimsPrincipal idTokenClaims = authResult.ClaimsPrincipal; 

var accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
foreach (var acc in accounts)
{
    Console.WriteLine($"Account for {acc.Username}");
    foreach (var tp in acc.GetTenantProfiles())
    {
        Console.WriteLine($"Tenant Profile in tenant {tp.TenantId} - claims {tp.ClaimsPrincipal.Claims.Count()} - is home tenant? {tp.IsHomeTenant}")
    }
}
```

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
bgavrilMS: We should not have ID token parsing logic in 2 different places and I have fixed that.

bgavrilMS: I'll update the error message as well.

bgavrilMS: Btw, IdToken actually is in format `<header>.<payload>.` i.e. missing sig.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -165,7 +165,7 @@ public static string iOSBrokerKeySaveFailed(string keyChainResult)
 
         public const string NonHttpsRedirectNotSupported = "Non-HTTPS URL redirect is not supported in webview. ";
 
-        public const string IDTokenMustHaveTwoParts = "ID Token must contain at least 2 parts. ";
+        public const string IDTokenMustHaveTwoParts = "ID Token must have a valid JWT format which contains atleast 3 parts. ";
```
jennyf19, line 168: >which contains atleast 3 parts [](http://example.com/codeflow?start=93&length=30)

nit: i would remove the end and leave valid JWT format, which implies 3 parts.

src/client/Microsoft.Identity.Client/TenantProfile.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// The user will have a different object ID in each tenant - they're considered different accounts, even though the user logs into each account with the same credentials. 
+        /// </summary>
+        public string Oid => _msalIdTokenCacheItem.ParseIdToken().ObjectId;
+
+        /// <summary>
+        /// Represents the tenant that the user is signing in to. 
+        /// For work and school accounts, the GUID is the immutable tenant ID of the organization that the user is signing in to.
+        /// For sign-ins to the personal Microsoft account tenant (services like Xbox, Teams for Life, or Outlook), the value is 9188040d-6c67-4c5b-b112-36a304b66dad. 
+        /// </summary>
+        public string TenantId => _msalIdTokenCacheItem.ParseIdToken().TenantId;
```
jennyf19, line 39: >ParseIdToken() [](http://example.com/codeflow?start=56&length=14)

any null checks needed on these?

jennyf19 (approved):
:shipit:

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenParsingTests.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            Assert.IsTrue(parsedToken.ClaimsPrincipal.Claims.Where(c => (new[] { "nbf", "iat", "exp" }).Contains(c.Type) == true).All(c => c.ValueType == ClaimValueTypes.Integer.ToString()));
+            Assert.IsTrue(parsedToken.ClaimsPrincipal.Claims.Where(c => (new[] { "nbf", "iat", "exp" }).Contains(c.Type) == false).All(c => c.ValueType == ClaimValueTypes.String.ToString()));
+
+            Assert.IsNull(parsedToken.Upn);
+            Assert.IsNull(parsedToken.FamilyName);
+            Assert.IsNull(parsedToken.GivenName);
```
pmaytak, line 64: Shouldn't we have these values in a token to know they are parsed correctly?

pmaytak (approved):

src/client/Microsoft.Identity.Client/AccountExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Extension methods for <see cref="IAccount"/>
+    /// </summary>
+    public static class AccountExtensions
+    {
+        /// <summary>        
+        /// The same account can exist in its home tenant and also as a guest in multiple other tenants. 
+        /// <see cref="TenantProfile"/> is derived from the ID token for that tenant.
+        /// </summary>
+        /// <remarks>Only tenants for which a token was acquired will be available in <c>TenantProfiles</c> property</remarks>
+        public static IEnumerable<TenantProfile> GetTenantProfiles(this IAccount account)
```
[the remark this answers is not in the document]
bgavrilMS: The Account ctor has a way to achieve this. I'll change it as you mentioned before - the param will be IEnumerable instead of IDictionary

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -47,7 +51,8 @@ public partial class AuthenticationResult
             IEnumerable<string> scopes,
             Guid correlationId,
             string tokenType = "Bearer",
-            AuthenticationResultMetadata authenticationResultMetadata = null)
+            AuthenticationResultMetadata authenticationResultMetadata = null, 
+            ClaimsPrincipal claimsPrincipal = null)
```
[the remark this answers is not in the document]
bgavrilMS: Good catch.

src/client/Microsoft.Identity.Client/Internal/IdToken.cs
```diff
@@ -21,42 +26,39 @@ internal static class IdTokenClaim
...
-        [JsonProperty(PropertyName = IdTokenClaim.Subject)]
-        public string Subject { get; set; }
+        public string Subject { get; private set; }
 
-        [JsonProperty(PropertyName = IdTokenClaim.TenantId)]
-        public string TenantId { get; set; }
+        public string TenantId { get; private set; }
 
-        [JsonProperty(PropertyName = IdTokenClaim.Version)]
-        public string Version { get; set; }
```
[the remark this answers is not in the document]
bgavrilMS: Not used by MSAL.

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenParsingTests.cs
bgavrilMS: Nope, these are not mandatory. The token at line 23 is an actual ID Token obtained via AAD. Upn is not a mandatory claim and neitehr are FamilyName and GivenName. 

`preffered_username` is expected to be there for AAD.

PR #2750, opened 2021-07-07
fix for #2747

### LINKED ISSUE #2747: [Bug] Remove the null check for .WithAzureRegion()
When a null value is passed in the .WithAzureRegion() api, we point skip region discovery and use the global endpoint and null is a valid value for this api.

Remove the null check made here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ee3cd9c2fe59c0c34ff3a65142bc16f017fd7b27/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L206

jmprieur:
Fixes #2747

Taking this one as I want to quickly update the 1P doc, and I want the product to work the way it's declared to work.
(was a partner request)

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -185,7 +185,8 @@ namespace Microsoft.Identity.Client
         /// Instructs MSAL.NET to use an Azure regional token service.
         /// </summary>
         /// <param name="azureRegion">Either the string with the region (preferred) or        
-        /// use <see cref="ConfidentialClientApplication.AttemptRegionDiscovery"/> and MSAL.NET will attempt to auto-detect the region.                
+        /// use <see cref="ConfidentialClientApplication.AttemptRegionDiscovery"/> and MSAL.NET will attempt to auto-detect the region,
+        /// or <c>null</c> to use the global endpoint.               
         /// </param>
         /// <remarks>
         /// Region names as per https://azure.microsoft.com/en-ca/global-infrastructure/geographies/.
@@ -203,11 +204,6 @@ namespace Microsoft.Identity.Client
         /// <returns>The builder to chain the .With methods</returns>
         public ConfidentialClientApplicationBuilder WithAzureRegion(string azureRegion = ConfidentialClientApplication.AttemptRegionDiscovery)
         {
-            if (string.IsNullOrEmpty(azureRegion))
-            {
-                throw new ArgumentNullException(nameof(azureRegion));
-            }
-
             Config.AzureRegion = azureRegion;
 
             return this;


src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -203,11 +204,6 @@ public ConfidentialClientApplicationBuilder WithClientAssertion(Func<string> cli
         /// <returns>The builder to chain the .With methods</returns>
         public ConfidentialClientApplicationBuilder WithAzureRegion(string azureRegion = ConfidentialClientApplication.AttemptRegionDiscovery)
         {
```
jennyf19, line 206: there should be a test that failed? right?

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
jennyf19: did you check that we have a test which passes in null?

jmprieur: yes, I was surprized that no test failed, right.
But I'll add one ...

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -203,11 +204,6 @@ public ConfidentialClientApplicationBuilder WithClientAssertion(Func<string> cli
         /// <returns>The builder to chain the .With methods</returns>
         public ConfidentialClientApplicationBuilder WithAzureRegion(string azureRegion = ConfidentialClientApplication.AttemptRegionDiscovery)
```
henrik-me, line 205: >azureRegion [](http://example.com/codeflow?start=75&length=11)

I would have assumed we should fail if null is used. This way people can believe they are using regional but they are not.

bgavrilMS: I'm pretty sure this fails with an `ArgumentNullException`

henrik-me (changes requested):
:clock1:

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
jmprieur: So, @henrik-me, shall I close this PR and the associated bug ?

jennyf19: but i don't see a test checking for this? unless i missed it.

jmprieur:
Proposing to close this PR.

I've proposed the following update to the 1P doc:

`.WithRegion(region)`, doesn't accept `null`. If you want to conditionally enable region detection, you'll want to use the following pattern:
  
  ```CSharp
  var confidentialClientApplicationBuilder = ConfidentialClientApplicationBuilder(client_id)
                  .WithCertificate(x509cert);
  if (region != null)
  {
    confidentialClientApplicationBuilder.WithAzureRegion(region);
  }
  s_cca = confidentialClientApplicationBuilder.Build();
  ```

@henrik-me @neha-bhargava : do we agree?

Issue #2736, opened 2021-06-29
MSAL 4.33 throws on mismatch authority


rabenshm:
Hello

After upgrading to MSAL 4.33.0 we are getting exception  Microsoft.Identity.Client.MsalClientException:
 The application is configured for cloud login.microsoftonline.com and the request for a different cloud - login.windows.net. This is not supported - the app and the request must target the same cloud.
See https://aka.ms/msal-net-authority-override for details.

We are using KeyVaultClient which is using callback which is probably sending the login.windows.net. We do not have control over that.

We initialize the application with login.microsoftonline.com

Thanks

bgavrilMS:
This is blocking Kusto from upgrading. The environments "login.microsoftonline.com" and "login.windows.net" are the same, MSAL should do instance discovery before throwing.

PR #2753, opened 2021-07-09
Instance discovery during authority check

trwalke:
Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2736

Calling instance discovery during authority check for aliases to be used when validating the authority environments. If a change in authority is detected but the change uses one of the aliased authorites, MSAL should not throw the authority mismatch error

**Testing**
Unit test

**Performance impact**
Evaluating


src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -134,6 +134,11 @@ public string ClientVersion
         public ITokenCacheInternal AppTokenCacheInternalForTest { get; set; }
 
         public IDeviceAuthManager DeviceAuthManagerForTest { get; set; }
+
+        public void SetAuthorityInfoForTest(AuthorityInfo authorityInfo)
+        {
+            AuthorityInfo = authorityInfo;
```
bgavrilMS, line 140: Why is this needed? AuthorityInfo already has a setter at line 68.

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -81,8 +81,8 @@ internal ClientApplicationBase(ApplicationConfiguration config)
             RequestContext requestContext,
             ITokenCacheInternal cache)
         {
-            var authority = Microsoft.Identity.Client.Instance.Authority.CreateAuthorityForRequest(
-               requestContext.ServiceBundle.Config.AuthorityInfo,
+            var authority = Instance.Authority.CreateAuthorityForRequest(
+               requestContext,
```
bgavrilMS, line 85: It's better to send exactly the objects needed (in this case the InstanceDiscoveryManager) rather than sending the entire  context object.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -208,6 +218,22 @@ private static void ValidateSameHost(AuthorityInfo requestAuthorityInfo, Authori
             }
         }
 
+        private static bool IsAuthorityAliased(RequestContext requestContext, AuthorityInfo requestAuthorityInfo)
+        {
+            var instanceDiscoveryManager = requestContext.ServiceBundle.InstanceDiscoveryManager;
+            var result = instanceDiscoveryManager.GetMetadataEntryAsync(requestContext.ServiceBundle.Config.AuthorityInfo, requestContext).Result;
```
bgavrilMS, line 224: We should not do this. This will block the thread. Never use `.Result` in production code. This is an async call, so you must make all the methods in the call chain async. There aren't that many.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -208,6 +218,22 @@ private static void ValidateSameHost(AuthorityInfo requestAuthorityInfo, Authori
             }
         }
 
+        private static bool IsAuthorityAliased(RequestContext requestContext, AuthorityInfo requestAuthorityInfo)
```
bgavrilMS, line 221: This verification is expensive, as it involves making a network call. We should only do it if the Config authority and the Request authority are different.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -1,290 +1,290 @@
...
+using Microsoft.Identity.Test.LabInfrastructure;
+using Microsoft.Identity.Test.Unit;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Integration.HeadlessTests
+{
+    // Note: these tests require permission to a KeyVault Microsoft account;
+    // Please ignore them if you are not a Microsoft FTE, they will run as part of the CI build
+    [TestClass]
+    public class UsernamePasswordIntegrationTests
```
bgavrilMS, line 29: Please check your space settings. They should be the default.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -1,290 +1,290 @@
...
+using Microsoft.Identity.Client.TelemetryCore;
+using Microsoft.Identity.Client.Utils;
+using Microsoft.Identity.Test.Common;
+using Microsoft.Identity.Test.Common.Core.Helpers;
+using Microsoft.Identity.Test.Integration.net45.Infrastructure;
+using Microsoft.Identity.Test.LabInfrastructure;
+using Microsoft.Identity.Test.Unit;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Integration.HeadlessTests
```
bgavrilMS, line 24: We should not add an integration test for this. Unit tests are sufficient.

tests/Microsoft.Identity.Test.Unit/ApiConfigTests/AuthorityTests.cs
```diff
@@ -1,15 +1,19 @@
...
+using Microsoft.Identity.Test.Common;
 using Microsoft.Identity.Test.Common.Core.Helpers;
+using Microsoft.Identity.Test.Common.Core.Mocks;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Microsoft.Identity.Test.Unit.ApiConfigTests
 {
     [TestClass]
-    public class AuthorityTests
+    public class AuthorityTests : TestBase
```
bgavrilMS, line 16: We should have a unit test that describes this issue.

bgavrilMS (changes requested):
We can't block the thread while making network calls.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
trwalke: I think I added it because intellisense was telling me that it was not settable. I think it was probably just a bug. Should have verified

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
trwalke: The instance discovery methods need the request context and since the instance discovery manager and the authority info is in the request context. I think it is better to just send the one object alone. This also reduces the number of parameters that need to be passed around.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4a0d46b37fb067163df88262ab3db473a81609c7/src/client/Microsoft.Identity.Client/Instance/Authority.cs#L224

src/client/Microsoft.Identity.Client/Instance/Authority.cs
trwalke: makes sense

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -208,6 +218,22 @@ private static void ValidateSameHost(AuthorityInfo requestAuthorityInfo, Authori
...
+
+            foreach (string alias in result.Aliases)
+            {
+                if (string.Equals(alias, requestAuthorityInfo.Host))
+                {
+                    return true;
+                }
+            }
+
+            return false;
```
pmaytak, line 234:
```suggestion
            return result.Aliases.Any(alias => alias.Equals(requestAuthorityInfo.Host));
```

pmaytak (approved):
LGTM so far, pending making that method call async.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
trwalke: I did not add a test here. this file was changed by VS for line endings

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
trwalke: Thsi came from a reccomended change to line endings by VS. not sure why but it complained about it

src/client/Microsoft.Identity.Client/Instance/Authority.cs
trwalke: Gotta remember to use Any().
Thanks

tests/Microsoft.Identity.Test.Unit/ApiConfigTests/AuthorityTests.cs
```diff
@@ -33,75 +54,107 @@ public void VerifyAuthorityTest()
...
-
-            var ex3 = Assert.ThrowsException<MsalClientException>(
-             () => Authority.CreateAuthorityForRequest(
-                 AuthorityInfo.FromAdfsAuthority(TestConstants.ADFSAuthority, true),
+            _harness.HttpManager.AddInstanceDiscoveryMockHandler();
+            _testRequestContext.ServiceBundle.Config.HttpManager = _harness.HttpManager;
+            _testRequestContext.ServiceBundle.Config.AuthorityInfo = s_commonAuthority;
+            var ex = Assert.ThrowsException<AggregateException>(
+                () =>
+                Authority.CreateAuthorityForRequestAsync(_testRequestContext, s_ppeAuthority, null).Result);
```
trwalke, line 117: Now that this is async the test ended up failing. This was the easiest way I found to assert the original conditions when using async in a lambda method. When I tried using the await Authority.CreateAuthorityForRequestAsync() the exception was not thrown and when using .Result the exception is aggregate so I just decided to go with that.

Is there a better way?

bgavrilMS (approved):

PR #2758, opened 2021-07-12
Add a test for WithAzureRegion(null)

bgavrilMS:
Just a unit test


SameerK-MSFT (approved):

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
```diff
@@ -236,6 +236,25 @@ public async Task RegionFallbackToGlobalAsync()
             }
         }
 
+        [TestMethod]
```
jennyf19, line 239: nit: you could do ` [TestMethod] [DataRow()]` to have less code. and then it's clear what's being covered.

jennyf19 (approved):

trwalke (approved):
LGTM

Issue #2701, opened 2021-06-14
[Bug] Instance discovery performed many times on non-public non-sovereign clouds


bgavrilMS:
MSAL 4.32.2

AcquireTokenForClient with a different cloud, such as PPE and use Fiddler

Actual: 
- about 6 calls to the discovery endpoint

PR #2760, opened 2021-07-13
Trwalke/instance discovery fix

bgavrilMS:
Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2701

Adding an instance discovery entry to cache when the returned metadata does not include the configured authority's environment.

**Testing**
Unit test, Manual testing

**Performance impact**
Negligible

### DIFF
--- a/tests/Microsoft.Identity.Test.Common/TestConstants.cs
+++ b/tests/Microsoft.Identity.Test.Common/TestConstants.cs
@@ -73,7 +73,8 @@ namespace Microsoft.Identity.Test.Unit
         public const string AuthorityRegional = "https://" + ProductionPrefRegionalEnvironment + "/" + TenantId + "/";
         public const string AuthorityRegionalInvalidRegion = "https://" + ProductionPrefInvalidRegionEnvironment + "/" + TenantId + "/";
         public const string AuthorityTenant = "https://" + ProductionPrefNetworkEnvironment + "/" + TenantId + "/";
-        public const string AuthorityCommonTenantNotPrefAlias = "https://" + ProductionNotPrefEnvironmentAlias + "/common/";        
+        public const string AuthorityCommonTenantNotPrefAlias = "https://" + ProductionNotPrefEnvironmentAlias + "/common/";
+        public const string AuthorityCommonPpeAuthority = "https://" + PPEEnvironment + "/common/";
 
         public const string PrefCacheAuthorityCommonTenant = "https://" + ProductionPrefCacheEnvironment + "/common/";
         public const string AuthorityOrganizationsTenant = "https://" + ProductionPrefNetworkEnvironment + "/organizations/";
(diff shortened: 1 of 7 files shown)


tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -73,7 +73,8 @@ public static HashSet<string> s_scope
         public const string AuthorityRegional = "https://" + ProductionPrefRegionalEnvironment + "/" + TenantId + "/";
         public const string AuthorityRegionalInvalidRegion = "https://" + ProductionPrefInvalidRegionEnvironment + "/" + TenantId + "/";
         public const string AuthorityTenant = "https://" + ProductionPrefNetworkEnvironment + "/" + TenantId + "/";
-        public const string AuthorityCommonTenantNotPrefAlias = "https://" + ProductionNotPrefEnvironmentAlias + "/common/";        
+        public const string AuthorityCommonTenantNotPrefAlias = "https://" + ProductionNotPrefEnvironmentAlias + "/common/";
+        public const string AuthorityCommonPpeAuthority = "https://" + PPEEnvironment + "/common/";
```
jennyf19, line 77: >PE [](http://example.com/codeflow?start=72&length=2)

nit: PpeEnvironment..

tests/Microsoft.Identity.Test.Unit/AuthorityAliasesTests.cs
```diff
@@ -182,5 +182,35 @@ private async Task ValidateCacheEntitiesEnvironmentAsync(ITokenCacheInternal cac
...
+                harness.HttpManager.AddSuccessTokenResponseMockHandlerForPost(TestConstants.AuthorityCommonPpeAuthority);
+
+                AuthenticationResult result = app
+                    .AcquireTokenInteractive(TestConstants.s_scope)
+                    .ExecuteAsync(CancellationToken.None)
+                    .Result;
+
+                Assert.IsNotNull(result);
+                Assert.IsTrue(harness.HttpManager.QueueSize == 0);
+            }
```
jennyf19, line 213: nit: `Assert.AreEqual(0, httpManager.QueueSize);` looks like most test are doing this.

jennyf19 (approved):
:shipit:

tests/Microsoft.Identity.Test.Unit/AuthorityAliasesTests.cs
bgavrilMS: It's not needed at all. The `using` statement at line 192 ensures that once the block of code finishes (at line 213), Dispose() gets called on the harness, which calls Dispose on the HttpManager, which Asserts that the queue is empty.

See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0d53f3e80e29f9c6d172e069c10e02d71e24f65e/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs#L48

bgavrilMS:
Please fix this test as well: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b51abc5960a2d4e60719cf810598f3fc0bd2bde2/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.cs#L152

bgavrilMS:
Since I opened the PR on your belhalf, I can't formally approve. So LGTM, fix pending comments and commit.

tests/Microsoft.Identity.Test.Unit/AuthorityAliasesTests.cs
jennyf19: that's what i thought, that we processed it automatically somewhere, but i couldn't find it. those other tests should probably be updated at some point.

pmaytak (approved):

PR #2774, opened 2021-07-16
Fix #2770 - do not cache authority endpoints

### LINKED ISSUE #2770: [Bug] AuthorityEndpoint caching causes memory leaks
msal 4.33

image

Repro steps: 

- MSAL + MIW  using OBO with 100,000k of tenants
- MSAL caches authority endpoints
- All those strings add up and memory slowly rises on the machine

Potential Solution: 

Review MSAL's static caches. It is acceptable for MSAL to cache enviroment (i.e. cloud) related data, but not tenant related data, since there are too many tenants!

bgavrilMS:
Fixes #2770 

Stop caching AuthorizationEndpoint strings.
Refactoring: get rid of some useless objects associated with this.

**Testing**
Existing tests cover this.

**Performance impact**
Memory is more constant now. 
@pmaytak - when looking at perf tests, this is a prime example of a problem that we should be able to catch. 

- run S2S flow for ~10 tenants (to warm start the test)
- take memory snapshot 1
- run S2S flow with 1 mil tenants
- take memory snapshot 2
- compare and observe leak


src/client/Microsoft.Identity.Client/Instance/AuthorityManager.cs
```diff
@@ -46,25 +54,29 @@ public async Task RunInstanceDiscoveryAndValidationAsync()
...
         {
-            if (!_instanceDiscoveryAndValidationExecuted)
+            s_validatedEnvironments.Clear();
+        }
+      
+        private async Task ValidateAuthorityAsync(Authority authority)
+        {
+            // race conditions could occur here, where multiple requests validate the authority at the same time
+            // but this is acceptable and once the cache is filled, no more HTTP requests will be made
+            if (!s_validatedEnvironments.Contains(authority.AuthorityInfo.Host))
```
jennyf19, line 72: why would that be acceptable in this case?

jennyf19 (approved):
:shipit:

jennyf19:
nice refactoring.

bgavrilMS:
@jennyf19 - yes, it's satisfying to delete code

src/client/Microsoft.Identity.Client/Instance/AuthorityManager.cs
bgavrilMS: We have 2 options: 

1. (status quo) - no syncronization. Multiple validation requests will be made (when system is starting), until the cache fills up (~1 second worth of requests, so we're talking few requests). 
2. use syncronization, i.e. make all requests wait for the one doing the instance dicsovery. As we found out with MyApps, if that request times out or takes ~30s to complete, then all requests will be blocked for that interval and the system is paralyzed.

To me, option 1 seems less bad.

pmaytak (approved):

PR #2776, opened 2021-07-16
Verify auto discovered region does not contain space

Avivskie:
Verify the region from the env variable does not contain space - space in the region will most likely indicates full region name is used, which is not valid as only short region names are accepted.

I did not write down the Fixed # because I am not sure this would be the entire fix at your end

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
@@ -162,7 +162,7 @@ namespace Microsoft.Identity.Client.Region
         {
             string region = Environment.GetEnvironmentVariable("REGION_NAME");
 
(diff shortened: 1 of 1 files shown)


jmprieur (commented):
Do we want to log something?
I thought we wanted to know when the environment variable is wrong using telemetry?

did I understand correctly @henrik-me ?

bgavrilMS:
Yes, we could add a logging statement and a unit test. I can take care of that. 

In this case, telemetry will tell us that regional auto-detection failed. We do send telemetry about "why" it failed.

bgavrilMS:
@Avivskie - I expanded on your fix here https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2784 - hope you don't mind I didn't include your commit, it was faster to just copy your changes.

PR #2785, opened 2021-07-20
Add public ctor for TokenCacheNotificationArgs

bgavrilMS:
Fixes # .
testability of MSAL

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
@@ -14,6 +14,31 @@ namespace Microsoft.Identity.Client
     /// </summary>
     public sealed partial class TokenCacheNotificationArgs
     {
+        /// <summary>
+        /// This constructor is for test purposes only. It allows apps to unit test their MSAL token cache implementation code.
+        /// </summary>
+        public TokenCacheNotificationArgs(
+            ITokenCacheSerializer tokenCache, 
+            string clientId, 
+            IAccount account, 
+            bool hasStateChanged, 
+            bool isApplicationCache, 
+            string suggestedCacheKey, 
+            bool hasTokens,
+            DateTimeOffset? suggestedCacheExpiry,
+            CancellationToken cancellationToken)
+        {
+            TokenCache = tokenCache;
+            ClientId = clientId;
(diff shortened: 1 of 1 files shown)


jennyf19 (approved):

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -14,6 +14,31 @@ namespace Microsoft.Identity.Client
...
+        public TokenCacheNotificationArgs(
+            ITokenCacheSerializer tokenCache, 
+            string clientId, 
+            IAccount account, 
+            bool hasStateChanged, 
+            bool isApplicationCache, 
+            string suggestedCacheKey, 
+            bool hasTokens,
+            DateTimeOffset? suggestedCacheExpiry,
+            CancellationToken cancellationToken)
```
jmprieur, line 29: We could use the defaults? 
suggestedCacheExpiry = null
cancellationToken = CancellationToken.None

jmprieur (approved):
LGTM
Thanks @bgavrilMS for the quick turn-around

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -14,6 +14,31 @@ namespace Microsoft.Identity.Client
     /// </summary>
     public sealed partial class TokenCacheNotificationArgs
     {
+        /// <summary>
+        /// This constructor is for test purposes only. It allows apps to unit test their MSAL token cache implementation code.
+        /// </summary>
+        public TokenCacheNotificationArgs(
```
pmaytak, line 20: You mean, for testing purposes for end users also, right? Or just for us internally (then it should be internal?).

pmaytak (approved):

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
bgavrilMS: For end users, yes. All our objects should either be behind an interface (for mocking) or have a public ctor. Since this is a data object, a ctor makes sense.

This is similar to your comment a while ago on `TenantProfile`, which was not testable.

Issue #2778, opened 2021-07-16
[Bug] Regional - on sovereign clouds, use preferred_network environment to construct regional env


bgavrilMS:
msal 4.34

1. Configure WithAzureRegion("westus1");
2. Configure a sovereign authority like `.WithAuthority("login.usgovcloudapi.net/common")`

Actual: 
- The actual authority becomes `westus1.login.usgovcloudapi.net`

Expected: 

- MSAL should respect the preferred_network semantics  and keep traffic on `westus1.login.microsoftonline.us`

```json
       {
            "preferred_network": "login.microsoftonline.us",
            "preferred_cache": "login.microsoftonline.us",
            "aliases": [
                "login.microsoftonline.us",
                "login.usgovcloudapi.net"
            ]
        }
```

PR #2789, opened 2021-07-21
Use preferred_network and skip discovery for known clouds

### LINKED ISSUE #2777: [Bug] Regional - if regional is configured, and original authority is known to MSAL, do not perform instance discovery
msal 4.34

1. Configure .WithAzureRegion("westus1")
2. Configure a well known authority like "login.microsotonline.com"

Expected: 

- no instance discovery should be performed
- we're ok to live with potential SSO islands. It's better than global traffic.

Actual: 

- instance discovery on the non-regional environment

bgavrilMS:
Fixes #2777 #2778 

1. Ensure instance discovery does not occur on known clouds (this was happening already, but added some tests)
2. Use "preferred_network" value for known clouds

**Testing**
unit

**Performance impact**
reduce number of http calls -> better perf
also removed a string comparison

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
@@ -117,6 +117,18 @@ namespace Microsoft.Identity.Client.Instance.Discovery
             return s_knownEnvironments.Contains(environment);
         }
 
+        public static bool TryGetKnownEnviromentPrefferedNetwork(string environment, out string preferredNetworkEnvironment)
+        {
+            if (s_knownEntries.TryGetValue(environment, out var entry))
+            {
+                preferredNetworkEnvironment = entry.PreferredNetwork;
+                return true;
+            }
+
+            preferredNetworkEnvironment = null;
+            return false;
+        }
+
         public static IDictionary<string, InstanceDiscoveryMetadataEntry> GetAllEntriesForTest()
         {
             return s_knownEntries;
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Net;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Utils;
 
@@ -117,7 +116,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
             return s_knownEnvironments.Contains(environment);
         }
 
-        public static bool TryGetKnownEnviromentPrefferedNetwork(string environment, out string preferredNetworkEnvironment)
+        public static bool TryGetKnownEnviromentPreferredNetwork(string environment, out string preferredNetworkEnvironment)
         {
             if (s_knownEntries.TryGetValue(environment, out var entry))
             {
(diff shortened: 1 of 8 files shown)


src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
```diff
@@ -117,6 +117,18 @@ public static bool IsKnownEnvironment(string environment)
             return s_knownEnvironments.Contains(environment);
         }
 
+        public static bool TryGetKnownEnviromentPrefferedNetwork(string environment, out string preferredNetworkEnvironment)
```
jennyf19, line 120: >TryGetKnownEnviromentPrefferedNetwork [](http://example.com/codeflow?start=27&length=37)

`TryGetKnownEnvironmentPreferredNetwork`

src/client/Microsoft.Identity.Client/Instance/Discovery/RegionDiscoveryProvider.cs
```diff
@@ -49,15 +45,16 @@ private static InstanceDiscoveryMetadataEntry CreateEntry(string originalEnv, st
 
         private string GetRegionalizedEnviroment(Uri authority, string region)
         {
-            var builder = new UriBuilder(authority);
+            string host = authority.Host;
 
-            // special rule for Global cloud
-            if (KnownMetadataProvider.IsPublicEnvironment(authority.Host))
+            // Regional business rule - use the PrefferedNetwork value for public and soverign clouds
```
jennyf19, line 50: >PrefferedNetwork [](http://example.com/codeflow?start=48&length=16)

sp.

jennyf19 (approved):
:shipit:

SameerK-MSFT (approved):
Minor spellings as Jenny pointed.

src/client/Microsoft.Identity.Client/Instance/Discovery/RegionDiscoveryProvider.cs
```diff
@@ -27,11 +28,7 @@ public async Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority
                 return null;
             }
 
-            // already regionalized
-            if (authority.Host.StartsWith(region))
-            {
-                return CreateEntry(requestContext.ServiceBundle.Config.AuthorityInfo.Host, authority.Host);
-            }
+
```
pmaytak, line 31: We need this. Otherwise `GetRegionalizedEnviroment` will end up regionalizing regionalized URL like so region.region.domain. I think the `SuccessfulResponseFromRegionalizedAuthorityAsync` unit test covers this scenario.

pmaytak (changes requested):

src/client/Microsoft.Identity.Client/Instance/Discovery/RegionDiscoveryProvider.cs
pmaytak: #Resolved

src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
pmaytak: #Resolved

src/client/Microsoft.Identity.Client/Instance/Discovery/RegionDiscoveryProvider.cs
pmaytak: #Resolved

pmaytak (approved):

bgavrilMS:
LGTM @pmaytak

trwalke:
PR build seems to be canceled. restarting

PR #2800, opened 2021-07-29
Disable assertion caching as it has race conditions

### LINKED ISSUE #2798: [Bug] Race condition around ConfidentialClient cached assertion
MSAL 4.35

Repro steps: 

1. Create 1 CCA app with a cert
2. Each request uses a different tenant ID
3. Fire lots of requests in parallel

**Actual**: some requests will fail because the internal JWT assertion (the client assertion) has invalid audience

Exception seen: 

Invalid audience...

The root cause for this is the cached assertion feature MSAL uses to try to save some CPU time by avoiding the expensive sign op.

Proposed fix: remove cached assertion feature. should not be needed if token cache is used. HTTP call is >10x slower than signing with a cert. Making fewer HTTP calls by good use of cache will improve app perf.

Root cause (brief description)
-----------------------------

Create confidential client app with cert. Internally this creates a ClientCredential, associated with the app. 
Each request uses this ClientCredential - it **reads and writes** to this object. This leads to race conditions.

Root cause (longer description)
-----------------------------

Request logic: 

- check if cached assertion matches audience
- if yes, use the cached assertion.
- if not, create new assertion and cache it

You should be able to see that this will not work in parallel. Or keep on reading...

Root cause (detailed description)
-----------------------------

Assume we have fired a request and have cached the assertion with audience set to aud1.

== Serial requests 

1. Create req1 with audience aud1. On execution, check if a cached assertion exists. It does. Check 
2. Create req2 for tenant t2 -> aud2. On execution, check if a cached assertion exists. It does. Check if aud is aud2. It's not. Create and sign again.

== In parallel request

Assume a cached assertion aud1
req1 is for aud1, req2 is for aud2. They fire in parallel.

req1 checks if it can use the cached assertion. It can. Hold that thought!
req1 is paused

req2 checks if it can use the cached assertion. It can't. It creates a new assertion and caches it. req2 fires and comples ok.

req1 is woken up
req1 continues - it think it can use the cached assertion because it just checked it (not knowing that req2 changed it)
it tries to use it -> failure, beucase it's for aud2

bgavrilMS:
Fixes #2798

remove assertion caching
I though about fixing this but I don't know how to do it without using locks, either implicitly or explicitly. 
ConccurentDictionary and WilsonLRU still use locking and caching logic is not trivial. 

In any case, in Confidential Client we ask ppl to use 1 app per request, hence signing already happens at scale and no issues come out of it. 

**Performance impact**
On a web api, assuming 200ms network penalty, the p95 difference between caching and no-caching is about 4ms.
Individually, the logic for creating and signing the assertion takes ~0.8 ms.

Please review the 2 commits separately - one is a just a refactoring.


src/client/Microsoft.Identity.Client/Internal/ClientCredentialWrapper.cs
```diff
@@ -150,43 +72,132 @@ private void ValidateCredentialParameters(ApplicationConfiguration config)
...
+            bool sendX5C)
+        {
+            using (logger.LogMethodDuration())
+            {
+                switch (AuthenticationType)
+                {
+                    case ConfidentialClientAuthenticationType.ClientCertificate:
+                        string tokenEndpoint = authority.GetTokenEndpoint();
+
+                        var jwtToken2 = new JsonWebToken(
```
jennyf19, line 151: can this return a null value or throw?

jennyf19 (approved):
:shipit:

src/client/Microsoft.Identity.Client/Internal/ClientCredentialWrapper.cs
bgavrilMS: I don't think so, it's just a data object, it copies those values in the ctor and keeps a reference to the crypto manager so that later on it can sign the jwt.

src/client/Microsoft.Identity.Client/Internal/ClientCredentialWrapper.cs
```diff
@@ -150,43 +72,132 @@ private void ValidateCredentialParameters(ApplicationConfiguration config)
...
+        }      
+
+        #endregion TestBuilders       
+   
         public static int MinKeySizeInBits { get; } = 2048;
-        internal string Thumbprint { get { return Base64UrlHelpers.Encode(Certificate.GetCertHash()); } }
-        internal X509Certificate2 Certificate { get; private set; }
+        internal string Thumbprint { get; }
+        internal X509Certificate2 Certificate { get; }
         // The cached assertion created from the JWT signing operation
```
trwalke, line 127: no longer needed

src/client/Microsoft.Identity.Client/Internal/ClientCredentialWrapper.cs
```diff
@@ -19,104 +23,22 @@ namespace Microsoft.Identity.Client.Internal
...
-
-        #endregion TestBuilders
-        private void CheckCertificateKeySize(X509Certificate2 cert)
-        {
-            //Currently, the min key size is enforced on desktop (net 45) as 2048 as it is the current industry standard.
-            //This is not enforced on netCore unfortunately and adding this enforcement may break customers on netCore.
-            //NetCore can only enforce a min key size of 512 since it is enforced by the portal already.
-#if DESKTOP
-            if (cert.PublicKey.Key.KeySize < MinKeySizeInBits)
+            if (config.ConfidentialClientCredentialCount == 0)
```
trwalke, line 35: is there a reason to remove this from the ValidateCredentialParameters() method?

src/client/Microsoft.Identity.Client/Internal/ClientCredentialWrapper.cs
```diff
@@ -150,43 +72,132 @@ private void ValidateCredentialParameters(ApplicationConfiguration config)
                 AuthenticationType = ConfidentialClientAuthenticationType.SignedClientAssertionDelegate;
             }
 
-            if (AuthenticationType == ConfidentialClientAuthenticationType.None)
+            switch (AuthenticationType)
```
trwalke, line 75: this switch should probably also throw NotImplementedException() as a default case?

src/client/Microsoft.Identity.Client/Internal/ClientCredentialWrapper.cs
bgavrilMS: it wasn't used anywhere and no test broke. I think this is enforced lower down the stack.

trwalke (approved):
Couple of questions but looks good overall.

pmaytak (approved):

Issue #2804, opened 2021-08-02
[Feature Request] Allow developers to set SendX5C once, rather than on every single request, as SNI creds don't work without it. 


hpsin:
**Is your feature request related to a problem? Please describe.**
When setting up SNI with Azure AD, every single STS request must come with an X5C claim, otherwise the credential cannot be used.  Currently, developers are forced to mark up every single token request with "SendX5C", and if they miss one, the request will fail. It would be easier to indicate, while setting the client credential, that it is an SNI scenario and therefore X5C must always be included. 

**Describe the solution you'd like**
Per-application setting of X5C, rather than per-request. A per-request override is provided, although marked obsolete. 

**Describe alternatives you've considered**
We are improving the SNI error messages server side as well to cut down on the number of incidents and questions we get about failure to setup SNI in MSAL. Making it easy to find the SNI option in MSAL, however, will also help.  Making this dynamic 
 (so no setup is required) is not possible at this time. 

**Additional context**
There was a comment in the codebase about how things like certs should be captured per-request, since there are some requests that "don't need them".  I am unaware of such patterns (particularly, I don't know that such patterns don't exist in ADFS), but if one existed it would make a per-application pattern less feasible.

bgavrilMS:
The only reason why a request would not need a cert if is the request can serve a token from its cache.

bgavrilMS:
I think this makes sense, we can simply add:

```csharp
            IConfidentialClientApplication cca = ConfidentialClientApplicationBuilder
                .Create(s_clientIdForConfidentialApp)
                .WithCertificate(cert, sendX5C = true)                
                .Build();
```

However, the problem with configuration options at both APP level and at REQUEST level is that they can conflict. In this case, if you'd set sendX5C to false in a request, but the app has it set to true, we'd just throw.


@hpsin - what if MSAL were to send the X5C claim all the time? Is this where you are saying that it cannot be done?

hpsin:
> I think this makes sense, we can simply add:
> 
> ```cs
>             IConfidentialClientApplication cca = ConfidentialClientApplicationBuilder
>                 .Create(s_clientIdForConfidentialApp)
>                 .WithCertificate(cert, sendX5C = true)                
>                 .Build();
> ```
> 
> However, the problem with configuration options at both APP level and at REQUEST level is that they can conflict. In this case, if you'd set sendX5C to false in a request, but the app has it set to true, we'd just throw.

[...]

PR #2805, opened 2021-08-02
ConfidentalClientApplicationBuilder WithSendX5C

hpsin:
Proposes a solution for #2804

Add WithSendX5C to ConfidentialClientApplicationBuilder. 
Makes the existing SendX5C channeled throughout confidential clients nullable, to be overwritten at request time by the default-null SendX5C within the ConfidentialClient. 
Marks the per-request SendX5C Builder methods as Obsolete, so that devs know to switch to per-app X5C. 

**Testing**
Inline updates to the existing unit tests, to switch away from the per-request model and into the per-app model.  This should validate that the claim is appearing as expected as a result of this change. 

**Performance impact**
No perf impact expected.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
@@ -85,6 +85,7 @@ namespace Microsoft.Identity.Client
         /// <param name="withSendX5C"><c>true</c> if the x5c should be sent. Otherwise <c>false</c>.
         /// The default is <c>false</c></param>
         /// <returns>The builder to chain the .With methods</returns>
+        [System.Obsolete("Set SendX5C once for this application using WithSendX5C on the ConfidentialClientApplicationBuilder object.", false)]
         public AcquireTokenByAuthorizationCodeParameterBuilder WithSendX5C(bool withSendX5C)
         {
             CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithSendX5C);
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
@@ -85,11 +85,11 @@ namespace Microsoft.Identity.Client
         /// <param name="withSendX5C"><c>true</c> if the x5c should be sent. Otherwise <c>false</c>.
         /// The default is <c>false</c></param>
         /// <returns>The builder to chain the .With methods</returns>
-        [System.Obsolete("Set SendX5C once for this application using WithSendX5C on the ConfidentialClientApplicationBuilder object.", false)]
         public AcquireTokenByAuthorizationCodeParameterBuilder WithSendX5C(bool withSendX5C)
         {
             CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithSendX5C);
-            Parameters.SendX5C = withSendX5C;
+            Parameters.SendX5C = withSendX5C; 
+            Parameters.SetPerRequestX5C = true;
             return this;
         }
 
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByRefreshTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByRefreshTokenParameterBuilder.cs
@@ -68,11 +68,11 @@ namespace Microsoft.Identity.Client
 #if !SUPPORTS_CONFIDENTIAL_CLIENT
         [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]  // hide confidential client on mobile
 #endif
-        [System.Obsolete("Set SendX5C once for this application using WithSendX5C on the ConfidentialClientApplicationBuilder object.", false)]
         public AcquireTokenByRefreshTokenParameterBuilder WithSendX5C(bool withSendX5C)
         {
             CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithSendX5C);
-            Parameters.SendX5C = withSendX5C;
+            Parameters.SendX5C = withSendX5C; 
+            Parameters.SetPerRequestX5C = true;
             return this;
         }
     }
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByRefreshTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByRefreshTokenParameterBuilder.cs
@@ -68,6 +68,7 @@ namespace Microsoft.Identity.Client
 #if !SUPPORTS_CONFIDENTIAL_CLIENT
         [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]  // hide confidential client on mobile
 #endif
+        [System.Obsolete("Set SendX5C once for this application using WithSendX5C on the ConfidentialClientApplicationBuilder object.", false)]
         public AcquireTokenByRefreshTokenParameterBuilder WithSendX5C(bool withSendX5C)
         {
             CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithSendX5C);
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
@@ -11,7 +11,9 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
 
         public string PkceCodeVerifier { get; set; }
 
-        public bool? SendX5C { get; set; }
+        public bool SendX5C { get; set; } 
+
+        internal bool SetPerRequestX5C = false;
 
         public void LogParameters(ICoreLogger logger)
         {
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
 
         public string PkceCodeVerifier { get; set; }
 
-        public bool SendX5C { get; set; }
+        public bool? SendX5C { get; set; }
 
         public void LogParameters(ICoreLogger logger)
         {
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
 
         /// <summary>
         /// </summary>
-        public bool SendX5C { get; set; }
+        public bool? SendX5C { get; set; }
 
         /// <inheritdoc />
         public void LogParameters(ICoreLogger logger)
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
@@ -14,7 +14,9 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
 
         /// <summary>
         /// </summary>
-        public bool? SendX5C { get; set; }
+        public bool SendX5C { get; set; } 
+        
+        internal bool SetPerRequestX5C = false;
 
         /// <inheritdoc />
         public void LogParameters(ICoreLogger logger)
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -97,6 +97,55 @@ namespace Microsoft.Identity.Client
             return this;
         }
 
+
+        /// <summary>
+        /// Sets the certificate associated with the application
+        /// </summary>
+        /// <param name="certificate">The X509 certificate used as credentials to prove the identity of the application to Azure AD.</param>
+        /// <param name="withSendX5C"><c>true</c> if the x5c should be sent. Otherwise <c>false</c>.
+        /// The default is <c>false</c></param>
+        /// <remarks>You should use certificates with a private key size of at least 2048 bytes. Future versions of this library might reject certificates with smaller keys. </remarks>
+        public ConfidentialClientApplicationBuilder WithCertificate(X509Certificate2 certificate, bool withSendX5C)
+        {
+            if (certificate == null)
+            {
+                throw new ArgumentNullException(nameof(certificate));
+            }
+
+            if (!certificate.HasPrivateKey)
+            {
+                throw new MsalClientException(MsalError.CertWithoutPrivateKey, MsalErrorMessage.CertMustHavePrivateKey(nameof(certificate)));
+            }
+
+            Config.ClientCredentialCertificate = certificate;
+            Config.ConfidentialClientCredentialCount++;
+            return WithSendX5C(withSendX5C);
+        }
+
+        /// <summary>
+        /// Specifies if the x5c claim (public key of the certificate) should be sent to the STS.
+        /// Sending the x5c enables application developers to achieve easy certificate roll-over in Azure AD:
+        /// this method will send the public certificate to Azure AD along with all STS requests,
+        /// so that Azure AD can use it to validate the subject name based on a trusted issuer policy.
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or PowerShell/CLI operation). For details see https://aka.ms/msal-net-sni
+        /// </summary>
+        /// <param name="withSendX5C"><c>true</c> if the x5c should be sent. Otherwise <c>false</c>.
+        /// The default is <c>false</c></param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public ConfidentialClientApplicationBuilder WithSendX5C(bool withSendX5C)
+        {
+            // Default is not to send X5C, so if called with false, this is a no-op. 
+            if (!withSendX5C)
+            {
+                return this;
+            }
+
+            Config.SendX5C= withSendX5C;
+            Config.ConfidentialClientCredentialCount++;
+            return this;
+        }
+
         /// <summary>
         /// Sets the certificate associated with the application along with the specific claims to sign.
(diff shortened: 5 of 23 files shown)


bgavrilMS:
Very cool @hpsin

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByRefreshTokenParameterBuilder.cs
```diff
@@ -68,6 +68,7 @@ internal override ApiEvent.ApiIds CalculateApiEventId()
 #if !SUPPORTS_CONFIDENTIAL_CLIENT
         [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]  // hide confidential client on mobile
 #endif
+        [System.Obsolete("Set SendX5C once for this application using WithSendX5C on the ConfidentialClientApplicationBuilder object.", false)]
         public AcquireTokenByRefreshTokenParameterBuilder WithSendX5C(bool withSendX5C)
         {
             CommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature.WithSendX5C);
```
bgavrilMS, line 74: If the user sets the x5c flag here to false, and is already set in the config as true, I think we should throw an exception. The config object can be access by: 

`this.ServiceBundle.Config`

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -97,6 +97,55 @@ public ConfidentialClientApplicationBuilder WithCertificate(X509Certificate2 cer
...
+        public ConfidentialClientApplicationBuilder WithSendX5C(bool withSendX5C)
+        {
+            // Default is not to send X5C, so if called with false, this is a no-op. 
+            if (!withSendX5C)
+            {
+                return this;
+            }
+
+            Config.SendX5C= withSendX5C;
+            Config.ConfidentialClientCredentialCount++;
```
bgavrilMS, line 145: Does this work? WithSendX5C should not increment this counter, which simply counts how many times APIs such as `WithSecret` `WithCertificate` etc. were used. It's used simply to throw an exception when 0 or >1 secrets are configured,.

bgavrilMS (changes requested):
LGTM, there is one small comment that needs to be resolved.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -97,6 +97,55 @@ public ConfidentialClientApplicationBuilder WithCertificate(X509Certificate2 cer
...
 
+
+        /// <summary>
+        /// Sets the certificate associated with the application
+        /// </summary>
+        /// <param name="certificate">The X509 certificate used as credentials to prove the identity of the application to Azure AD.</param>
+        /// <param name="withSendX5C"><c>true</c> if the x5c should be sent. Otherwise <c>false</c>.
+        /// The default is <c>false</c></param>
+        /// <remarks>You should use certificates with a private key size of at least 2048 bytes. Future versions of this library might reject certificates with smaller keys. </remarks>
+        public ConfidentialClientApplicationBuilder WithCertificate(X509Certificate2 certificate, bool withSendX5C)
```
bgavrilMS, line 108: `WithClientClaims` also needs to be modified to take in this flag.

bgavrilMS:
@jmprieur @jennyf19 - please have a look at this PR / public API surface.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
hpsin: So, I wasn't sure if WithClientClaims could itself support injecting the x5c claim. I was worried that we might collide. If it only injects claims into the body, which makes sense, then yeah, for sure.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
hpsin: My thinking was that this would cause Build to throw if you had set WithSendX5C and not set a certificate. That would be expected, because you can only use x5c with a cert (and not with a secret or provided assertion).

bgavrilMS: Yeah, I agree that if you set sendX5C but don't use WithCertificate, we should throw. I don't think we do this now, and there is no need to do it as part of this PR.

But when we throw exceptions, we need to be very explicit in describing the error message. In this case what would happen is that if the developer adds 

`app.WithCertificate(cert, true)` then that counter would be 2 and MSAL would throw an exception thinking "oh, you've used WithCertificate and WithSecret at the same time, this is not supported".

hpsin: Oh, I totally misread this to be a tracker on cert use, not credentials attached. Yep, will remove. Thanks!

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
```diff
@@ -85,6 +85,7 @@ internal override Task<AuthenticationResult> ExecuteInternalAsync(CancellationTo
         /// <param name="withSendX5C"><c>true</c> if the x5c should be sent. Otherwise <c>false</c>.
         /// The default is <c>false</c></param>
         /// <returns>The builder to chain the .With methods</returns>
+        [System.Obsolete("Set SendX5C once for this application using WithSendX5C on the ConfidentialClientApplicationBuilder object.", false)]
```
jennyf19, line 88: I don't think we should obsolete the one on the AT calls. Both should work. We can change documentation to reflect the one on the cca object, but let's not break people.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
```diff
@@ -11,7 +11,7 @@ internal class AcquireTokenByAuthorizationCodeParameters : IAcquireTokenParamete
 
         public string PkceCodeVerifier { get; set; }
 
-        public bool SendX5C { get; set; }
+        public bool? SendX5C { get; set; }
```
jennyf19, line 14: >? [](http://example.com/codeflow?start=19&length=1)

i don't think this will work in MSAL, they don't support the nullable reference types introduced in c# 8.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -97,6 +97,55 @@ public ConfidentialClientApplicationBuilder WithCertificate(X509Certificate2 cer
...
+        /// this method will send the public certificate to Azure AD along with all STS requests,
+        /// so that Azure AD can use it to validate the subject name based on a trusted issuer policy.
+        /// This saves the application admin from the need to explicitly manage the certificate rollover
+        /// (either via portal or PowerShell/CLI operation). For details see https://aka.ms/msal-net-sni
+        /// </summary>
+        /// <param name="withSendX5C"><c>true</c> if the x5c should be sent. Otherwise <c>false</c>.
+        /// The default is <c>false</c></param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public ConfidentialClientApplicationBuilder WithSendX5C(bool withSendX5C)
+        {
```
jennyf19, line 137: i don't think we need this one. It has to go with the cert or signed assertion, so makes sense that it's an additional parameter there, so no need to have it separate. I think it would lead to confusion and people setting it twice.

src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
```diff
@@ -122,6 +122,13 @@ public interface IAppConfig
 #endif
         X509Certificate2 ClientCredentialCertificate { get; }
 
+        /// <summary>
+        /// </summary>
+#if !SUPPORTS_CONFIDENTIAL_CLIENT
+        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]  // hide confidential client on mobile
+#endif
+        bool SendX5C { get; }
+
```
jennyf19, line 131: if we go w/the parameter option, this is not needed.

src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
```diff
@@ -197,7 +197,9 @@ public JWTHeaderWithCertificate(ClientCredentialWrapper credential, bool sendCer
 
                 X509CertificatePublicCertValue = null;
 
-                if (!sendCertificate)
+                bool perRequestSendX5C = sendCertificate ?? credential.SendX5C;
```
jennyf19, line 200: >credential [](http://example.com/codeflow?start=60&length=10)

null check on `credential`?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
SameerK-MSFT: I agree with Jenny that this will be an additional method and need not obsolete

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
SameerK-MSFT: Help me understand why nullable bool is required and not just bool?

jennyf19 (changes requested):
:clock1:

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
bgavrilMS: @jennyf19 - nullable types were introduced in very early C# version, we do support them in MSAL. However, nullable *reference* types are not supported because of C# 8 - https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
bgavrilMS: This comment reminds me that it would be nice to support x5c as an option.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
jennyf19: That's what I thought I wrote. `nullable reference types`

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
hpsin: Generally agreed. Thankfully setting it twice won't do anything bad. The only reason I went this way was to snap with how the other SDKs do it (e.g. [Java](https://www.javadoc.io/doc/com.microsoft.azure/msal4j/latest/com/microsoft/aad/msal4j/ConfidentialClientApplication.Builder.html)) - happy to remove.  Other consideration was that since SNI is still largely 1p only, I wasn't sure about polluting the public namespace so visibly with something they shouldn't use. WDYT - remove it?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
hpsin: Right, good for overrides if such a thing is needed.  I was expecting that this would be a warning. not an error, so not break people.  If the library setup to treat this as an error despite passing false into the Obsolete?

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
hpsin: Ah, darn.  This makes things much trickier - I'm worried that without being nullable it will default to false on a per-request basis, and overwrite the cca field that should be the primary driver of the value. 

Goal here: If the request params explicitly has SendX5C set, use that, otherwise fall back to the default-false cca object. Currently I'm using the nullable as a way to signal it's been explicitly set.  Should I add a new field instead, that's set during the per-request call, something like `bool overrideCcaSendX5C` in `IAcquireTokenParameters`?

src/client/Microsoft.Identity.Client/AppConfig/IAppConfig.cs
hpsin: Ah, what's the parameter option? Embedding it inside the credentialWrapper instead of inside the IAppConfig?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
bgavrilMS: This will emit a warning, not an error. We do this from time to time. It will not break people, but it will annoy them. 

This will warn ALL developers using `sendx5c` (1st parties only I guess) to upgrade. Some projects (e.g. Kusto) are very strict and have a 0 warnings policy. Others won't care. 

If we think it's wrong to set sendX5C at request level, let's deprecate it. But we must make sure to update our 1st party docs and that the obsolete message is very clear. 

@henrik-me - thoughts?

Aside: When we release MSAL 5 (we keep saying this for the past 3 years), we will delete all the deprecated APIs.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
bgavrilMS: Sorry, I mean that nullable **value** types was introduced since c# 2.0. Nullable reference types was introduced in C# 8.0

image

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
jennyf19: I would do one or the other, but not both. I guess Jean-Marc would need to decide. What happens if a 3rd party tries to use it? also it's false by default, so most won't even notice it.

hpsin: 3ps that try to use this will see a perf impact, but won't break (it doesn't help or harm them beyond sending extraneous bits over the wire).  Classic AAD behaviour would be to ignore the x5c claim if the cert thumbprint exists already in the app/SP registration, because while the developer is doing something "wrong" they're not doing anything illegal, so we'll let them do it.

jennyf19: It's kind of a non-optional parameter for first party, so i think it goes outside our normal "let's not pollute". 
@jmprieur to decide.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
trwalke: @hpsin I think it would be better to modify the logic internally to choose the proper value based on the parameters set rather than adding additional api here.

trwalke (changes requested):
Remove nullable reference type and update internal logic to handle sendX5c when set from various locations in MSAL

trwalke:
Also, Since this is a fork, the automation did not run. We should probably recreate this PR with a branch on our main repo and allow the automation to run.

jennyf19:
we can make a branch off of master and merge into that branch once @hpsin is done and then we can check the tests and make any final changes before merging into master.

---
In reply to: [891407708](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2805#issuecomment-891407708) [](http://example.com/codeflow?ancestors=891407708,891407708)

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenByAuthorizationCodeParameters.cs
hpsin: Can you clarify exactly what you mean by modify the logic internally? I don't thing I have the same name references for things that your team does :(  The specific line you're commenting on makes the notion of setting X5C per-request nullable so that we can perform internal logic to figure out what to do, rather than adding fields to the request parameters object. We can't do this without an API in general, as the developer needs to have some way of specifying whether we do this.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
```diff
@@ -14,7 +14,9 @@ internal class AcquireTokenForClientParameters : IAcquireTokenParameters
 
         /// <summary>
         /// </summary>
-        public bool SendX5C { get; set; }
+        public bool SendX5C { get; set; } 
+        
+        internal bool SetPerRequestX5C = false;
```
hpsin, line 19: The amount of repetition here makes me wonder if there should be an IConfidentialClientRequestParameters.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
henrik-me: +1 to Jenny: Let's not remove per request override and thus I see no reason to introduce a warning here. If anything an analyzer needs to be written to ensure recommended practices are followed.

(btw Java sends the X5C by default but allows to disable sending it)

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
henrik-me: Could be an option to change the default of sending the X5C to true

jennyf19 (approved):
:shipit:

trwalke (approved):

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForClientParameters.cs
bgavrilMS: There is, it's called `AbstractClientAppBaseAcquireTokenParameterBuilder`

trwalke (approved):
Approving so this can be recreated on our repo and finalized

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
SameerK-MSFT: It is server to server communication (i.e. high bandwidth). So it impact may not be as much. It is a trade of between reliability and perf. The perf impact will be smaller compared to the gain. - My 2 cents

Issue #2803, opened 2021-07-30
[Bug]  support for REGION_NAME env variable should be more resilient to user friendly region names


bgavrilMS:
msal 4.35.1

REGION_NAME env variable does not seem reliable. Azure Functions sometimes populate it with the long region name (which is unusable).

The solution is: 
`string region = Environment.GetEnvironmentVariable("REGION_NAME")?.Replace(" ", string.Empty).?ToLower();`

jmprieur:
@bgavrilMS @SameerK-MSFT @pmaytak 
After discussing with the Azure function PM, I learnt that this is a standard conversion, and most places in azure tools to compute the short name region by stripping out whitespace and lowercasing the string. For example, CLI, ARM template processing, etc do that

So, the data we need is there in the REGION_NAME variable. And Matthew would assert that MSAL can and should be resilient to users (reasonably) passing in the versions with whitespace.

Therefore proposing to:
- close this issue.
- Or repurpose it to make the detection by REGION_NAME more resilient: in RegionManager.cs l 163, something like the following?

  ```CSharp
  string region = Environment.GetEnvironmentVariable("REGION_NAME")?.Replace(" ", string.Empty).?ToLower();
  ```

cc: @trwalke, @jennyf19, @henrik-me

jmprieur:
Unless anybody objects proposing to move it back to the triage column
Changing the title

dannybtsai:
@jmprieur, do you have an example of the REGION_NAME?

jmprieur:
@dannybtsai 

"RegionName": "East US 2",
"RegionShortName": "eastus2"

PR #2811, opened 2021-08-05
Removing regional env variable (REGION_NAME)

trwalke:
Fixes # .
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2803

Removing regional env variable (REGION_NAME)

**Testing**
Automated testing

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Region/RegionAutodetectionSource.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Region/RegionAutodetectionSource.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Identity.Client.Region
         /// <summary>
         /// Auto-detected from IMDS
         /// </summary>
-        Imds = 3,
+        Imds = 4,
 
     }
 }
--- a/src/client/Microsoft.Identity.Client/Instance/Region/RegionAutodetectionSource.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Region/RegionAutodetectionSource.cs
@@ -26,14 +26,9 @@ namespace Microsoft.Identity.Client.Region
         Cache = 2,
 
         /// <summary>
-        /// Auto-detected from Env Variable
-        /// </summary>
-        EnvVariable = 3,
-
-        /// <summary>
         /// Auto-detected from IMDS
         /// </summary>
-        Imds = 4,
+        Imds = 3,
 
     }
 }
--- a/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
@@ -52,12 +52,12 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
             // Arrange
             var factory = new HttpSnifferClientFactory();
             var settings = ConfidentialAppSettings.GetSettings(Cloud.Public);
-            _confidentialClientApplication = BuildCCA(settings, factory);
+            _confidentialClientApplication = BuildCCA(settings, factory, false, TestConstants.Region);
 
             AuthenticationResult result = await GetAuthenticationResultAsync(settings.AppScopes).ConfigureAwait(false); // regional endpoint
             AssertTokenSourceIsIdp(result);
             AssertValidHost(true, factory);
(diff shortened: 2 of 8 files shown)


tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
```diff
@@ -46,25 +46,18 @@ public void TestInitialize()
...
-        }
-
         [TestMethod]
         public async Task AcquireTokenToRegionalEndpointAsync()
         {
             // Arrange
             var factory = new HttpSnifferClientFactory();
             var settings = ConfidentialAppSettings.GetSettings(Cloud.Public);
-            _confidentialClientApplication = BuildCCA(settings, factory);
+            _confidentialClientApplication = BuildCCA(settings, factory, false, TestConstants.Region);
```
trwalke, line 55: I am not 100% sure if this test is right with the removal of the env variable. I assumed that if I set the same region with the WithAzureRegion() api that the test would attempt discovery via IMDS but the call fails with {"A socket operation was attempted to an unreachable network."}. Is IMDS supposed to be reachable in this integration test with "centralis"?

@bgavrilMS

bgavrilMS: IMDS is an HTTP call to smth like `http://192.168.0.1/imds/metadata`. This url is not available on your local machine, and so it fails. 

However, it might work on the Azure DevOps VM, since they use actual VMs.

src/client/Microsoft.Identity.Client/Instance/Region/RegionAutodetectionSource.cs
```diff
@@ -25,15 +25,10 @@ internal enum RegionAutodetectionSource
...
-        /// <summary>
-        /// Auto-detected from Env Variable
-        /// </summary>
-        EnvVariable = 3,
-
         /// <summary>
         /// Auto-detected from IMDS
         /// </summary>
-        Imds = 4,
+        Imds = 3,
```
pmaytak, line 31: We send this in telemetry. I would keep it `Imds = 4`.

pmaytak (changes requested):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
trwalke: Seems like it is working on the cloud build but not locally. The is resulting in  test issue however since the behavior of the test is different when run locally, resulting in assertions to fail. Will need to resolve

bgavrilMS: How about 

```csharp
if (Environment.GetVariable("TF_BUILD") == null)
{
   Assert.Inconclusive("This test needs to run on Azure DevOps hosted agent, to be located on an Azure VM");
}
```

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
```diff
@@ -94,6 +86,7 @@ public async Task InvalidRegion_GoesToInvalidAuthority_Async()
 
         private void AssertTelemetry(HttpSnifferClientFactory factory, string currentTelemetryHeader, int placement = 0)
         {
+            Console.WriteLine($"Total requests for regional: {factory.RequestsAndResponses.Count}");
```
SameerK-MSFT, line 89: This looks like a leftover from debugging. Or do you need it?

trwalke: yes, removing it

pmaytak (approved):

jmprieur:
@bgavrilMS @SameerK-MSFT @pmaytak 
After discussing with the Azure function PM, I learnt that this is a standard conversion, and most places in azure tools to compute the short name region by stripping out whitespace and lowercasing the string. For example, CLI, ARM template processing, etc do that

So, the data we need is there in the REGION_NAME variable. And Matthew would assert that MSAL can and should be resilient to users (reasonably) passing in the versions with whitespace.

cc: @trwalke, @jennyf19, @henrik-me

PR #2814, opened 2021-08-10
Error message filtering as per error

### LINKED ISSUE #2808: [Bug] Fix error message and document throttling (AADSTS50196)
MSAL.NET 4.32

**Platform**
.NET Framework 4.7.2

* Desktop / Mobile
    * [x] Interactive
    * [ ] Username Password
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 

Upgraded from 4.28 (?) to 4.32 when AADSTS50196 started to appear.

**Repro**
- Create one cache per "client" (app concept) per feature test
  - Note: because all tests do that, this effectively makes the cache **miss** all the time since it's recreated for all tests and all clients trying to authn
- Run test with 2+ clients, all using the same user account
  - This will result in 2+ authn queries being raised almost at the same time, for the same AAD account
- For each client, call `AcquireTokenSilent()` (with the same account), which always fails with `MsalUiRequiredException` since the cache is brand new and not shared among test clients
- Call `AcquireTokenInteractive()` on exception caught

**Expected behavior**
All clients (which use the same AAD account) login successfully.

**Actual behavior**
AADSTS50196: The server terminated an operation because it encountered a loop while processing a request
Problems are:
- Error message is wrong; the code has no loop
- This error is an MsalUiRequiredException exception, which is not documented with `AcquireTokenInteractive()`
- #1788 hinted us that this was maybe throttling; this behavior is not documented AFAIK, and the exception type name doesn't match. Probably there should be a different exception, because unlike MsalUiRequiredException the end-user here cannot do anything about this issue, this is an error of the app code itself
- We didn't find any mention of throttling in release notes of 4.28+ => 4.32, nor what the throttling duration was

**Possible solution**
We're investigating sharing the token cache more broadly in tests to avoid multiple calls to the authn server in a short period (that is, to effectively leverage the cache and therefore have at worst one authn per test, and at best one per test run). TBD for results on whether that works.

**Additional context / logs / screenshots**
To be clear, we understand that _if_ the bug is indeed throttling _then_ this is an error on our side, misusing the API. But we found no way to confirm that, which is the actual bug being logged here.

SameerK-MSFT:
Fixes #2808.

Filtered error for 50196 and updated the doc

**Testing**
NA. Could not repro.

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Identity.Client
     {
         /// <summary>
         /// Standard OAuth2 protocol error code. It indicates that the application needs to expose the UI to the user
-        /// so that the user does an interactive action in order to get a new token.
+        /// so that the user does an interactive action in order to get a new token. Or you may be using multiple toke cache for the same user.
         /// <para>Mitigation:</para> If your application is a <see cref="T:IPublicClientApplication"/> call <c>AcquireTokenInteractive</c>
         /// perform an interactive authentication. If your application is a <see cref="T:ConfidentialClientApplication"/> chances are that the Claims member
         /// of the exception is not empty. See <see cref="P:MsalServiceException.Claims"/> for the right mitigation
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Identity.Client
     {
         /// <summary>
         /// Standard OAuth2 protocol error code. It indicates that the application needs to expose the UI to the user
-        /// so that the user does an interactive action in order to get a new token. Or you may be using multiple toke cache for the same user.
+        /// so that the user does an interactive action in order to get a new token.
         /// <para>Mitigation:</para> If your application is a <see cref="T:IPublicClientApplication"/> call <c>AcquireTokenInteractive</c>
         /// perform an interactive authentication. If your application is a <see cref="T:ConfidentialClientApplication"/> chances are that the Claims member
         /// of the exception is not empty. See <see cref="P:MsalServiceException.Claims"/> for the right mitigation
--- a/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
+++ b/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
@@ -224,7 +224,7 @@ namespace Microsoft.Identity.Client
         public const string TenantIdAndAadAuthorityInstanceAreMutuallyExclusive = "TenantId and AadAuthorityAudience are both set, but they're mutually exclusive. ";
         public const string InstanceAndAzureCloudInstanceAreMutuallyExclusive = "Instance and AzureCloudInstance are both set but they're mutually exclusive. ";
         public const string NoRefreshTokenProvided = "A refresh token must be provided. ";
-        public const string AadThrottledError = "Your app has been throttled by AAD due to too many requests. To avoid this, cache your tokens see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Client-Throttling.";
+        public const string AadThrottledError = "Your app has been throttled by AAD due to too many requests. To avoid this, cache your tokens see https://aka.ms/msal-net-throttling.";
 
         public const string NoTokensFoundError = "No Refresh Token found in the cache. ";
         public const string NoRefreshTokenInResponse = "Acquire by refresh token request completed, but no refresh token was found. ";
--- a/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
+++ b/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
@@ -224,7 +224,7 @@ namespace Microsoft.Identity.Client
         public const string TenantIdAndAadAuthorityInstanceAreMutuallyExclusive = "TenantId and AadAuthorityAudience are both set, but they're mutually exclusive. ";
         public const string InstanceAndAzureCloudInstanceAreMutuallyExclusive = "Instance and AzureCloudInstance are both set but they're mutually exclusive. ";
         public const string NoRefreshTokenProvided = "A refresh token must be provided. ";
-        public const string ThrottledTooManyCalls = "You are calling AAD too often and it has throttled you. To avoid this, cache your tokens see https://aka.ms/msal-net-token-caching.";
+        public const string AadThrottledError = "Your app has been throttled by AAD due to too many requests. To avoid this, cache your tokens see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Client-Throttling.";
 
         public const string NoTokensFoundError = "No Refresh Token found in the cache. ";
         public const string NoRefreshTokenInResponse = "Acquire by refresh token request completed, but no refresh token was found. ";
--- a/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
+++ b/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
@@ -224,6 +224,7 @@ namespace Microsoft.Identity.Client
         public const string TenantIdAndAadAuthorityInstanceAreMutuallyExclusive = "TenantId and AadAuthorityAudience are both set, but they're mutually exclusive. ";
         public const string InstanceAndAzureCloudInstanceAreMutuallyExclusive = "Instance and AzureCloudInstance are both set but they're mutually exclusive. ";
         public const string NoRefreshTokenProvided = "A refresh token must be provided. ";
+        public const string ThrottledTooManyCalls = "You are calling AAD too often and it has throttled you. To avoid this, cache your tokens see https://aka.ms/msal-net-token-caching.";
 
         public const string NoTokensFoundError = "No Refresh Token found in the cache. ";
         public const string NoRefreshTokenInResponse = "Acquire by refresh token request completed, but no refresh token was found. ";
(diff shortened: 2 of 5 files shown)


src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -12,7 +12,7 @@ public static class MsalError
     {
         /// <summary>
         /// Standard OAuth2 protocol error code. It indicates that the application needs to expose the UI to the user
-        /// so that the user does an interactive action in order to get a new token.
+        /// so that the user does an interactive action in order to get a new token. Or you may be using multiple toke cache for the same user.
```
jennyf19, line 15: nit: not sure about the additional message here. i don't think it's necessary and not a common issue. there shouldn't be multiple caches for the same user.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -121,7 +121,8 @@ private static bool IsInvalidGrant(string errorCode, string subErrorCode)
 
         private static bool IsInvalidGrantSubError(string subError)
         {
-            if (string.IsNullOrEmpty(subError))
+            if (string.IsNullOrEmpty(subError)
+                || subError.Contains("50196"))
```
jennyf19, line 125: not sure exactly how this resolves the issue. also, use a constant for the value here. Do we have any tests to update/add?

src/client/Microsoft.Identity.Client/MsalError.cs
bgavrilMS, line 15: You can use multiple caches for the same user. Not sure what this means.

bgavrilMS (changes requested):
I don't think this is the right fix. The bug states to "fix the error message", i.e. the error message that comes from ESTS is too technical and users do not understand the root cause. 

The fix should be: 

- when the exception error code is "invalid_grant" and the suberror or the message starts with "AADSTS50196", throw an `MsalUiRequiredException` with the error message "You are calling AAD too often and it has throttled you. To avoid this, cache your tokens - see https://aka.ms/msal-net-token-caching. Original exception: <original exception>"

- point https://aka.ms/msal-net-token-caching to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/token-cache-serialization

- you can easily unit test this class, there are already unit tests for it.

src/client/Microsoft.Identity.Client/MsalError.cs
SameerK-MSFT: Removed.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
SameerK-MSFT: Used const and added a test.

src/client/Microsoft.Identity.Client/MsalError.cs
SameerK-MSFT: Removed.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -224,6 +224,7 @@ public static string iOSBrokerKeySaveFailed(string keyChainResult)
         public const string TenantIdAndAadAuthorityInstanceAreMutuallyExclusive = "TenantId and AadAuthorityAudience are both set, but they're mutually exclusive. ";
         public const string InstanceAndAzureCloudInstanceAreMutuallyExclusive = "Instance and AzureCloudInstance are both set but they're mutually exclusive. ";
         public const string NoRefreshTokenProvided = "A refresh token must be provided. ";
+        public const string ThrottledTooManyCalls = "You are calling AAD too often and it has throttled you. To avoid this, cache your tokens see https://aka.ms/msal-net-token-caching.";
```
jennyf19, line 228: nit: `AadThrottledError`
and maybe:
`Your app has throttled AAD due to too many requests. See [https://aka.ms/msal-net-throttling](https://aka.ms/msal-net-throttling) for how to avoid this in the future.`

I think you should make a new aka.ms link for throttling because in addition to caching, they should also make sure they are using the right call pattern:
try{
AT Silent}
catch()
{
AT Interactive
}

also, they should not use `WithForceRefresh(true)` 

so all this can be added to one new page and link.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Identity.Client
 {
     internal class MsalServiceExceptionFactory
     {
+        private const string STSTooManyCallsError = "AADSTS50196";
+
```
jennyf19, line 15: nit: `AadThrottledErrorCode`

can you move this to the constants class (i think there is one) so you can reuse in the test?

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
pmaytak: We do have this page: [https://aka.ms/msal-net-throttling](https://aka.ms/msal-net-throttling). Maybe the above info an be added there.

This leads to a question - when does AAD return 429 Too Many Calls vs AADSTS50196; or are they the same?

trwalke: Should this be "Your has been throttled by AAD due to too many requests." ? Since AAD is doing the throttling?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
SameerK-MSFT: updated

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
SameerK-MSFT: Good suggestion. Done

jennyf19 (approved):
Thanks :shipit:

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -1,398 +1,398 @@
...
+
+        public const string LoggingCallbackAlreadySet = "LoggingCallback has already been set. ";
+        public const string TelemetryCallbackAlreadySet = "TelemetryCallback has already been set. ";
+        public const string NoClientIdWasSpecified = "No ClientId was specified. ";
+        public const string AdfsNotCurrentlySupportedAuthorityType = "ADFS is not currently a supported authority type. ";
+        public const string TenantIdAndAadAuthorityInstanceAreMutuallyExclusive = "TenantId and AadAuthorityAudience are both set, but they're mutually exclusive. ";
+        public const string InstanceAndAzureCloudInstanceAreMutuallyExclusive = "Instance and AzureCloudInstance are both set but they're mutually exclusive. ";
+        public const string NoRefreshTokenProvided = "A refresh token must be provided. ";
+        public const string AadThrottledError = "Your app has been throttled by AAD due to too many requests. To avoid this, cache your tokens see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Client-Throttling.";
```
pmaytak, line 227: Link should be aka.ms one - https://aka.ms/msal-net-throttling - (points to the same page).

pmaytak (approved):
LGTM. Just update the link in the error message to aka.ms one.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
SameerK-MSFT: Good catch. Updated

SameerK-MSFT:
Shows successful build. Will merge
image

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -26,7 +27,14 @@ internal class MsalServiceExceptionFactory
 
             if (IsInvalidGrant(oAuth2Response?.Error, oAuth2Response?.SubError) || IsInteractionRequired(oAuth2Response?.Error))
             {
-                ex = new MsalUiRequiredException(errorCode, errorMessage, innerException);
+                if (IsThrottled(oAuth2Response))
+                {
+                    ex = new MsalUiRequiredException(errorCode, MsalErrorMessage.AadThrottledError, innerException);
```
djee-ms, line 32: "The request was throttled" is not an "UI required" kind of exception/error. `MsalUiRequiredException` is an exception that tells the app developer it needs to call a different API to show some interactive UI. This is an exception that the app commonly wants to catch under normal usage pattern. The throttled error on the other hand is an internal app error where the app sends requests too fast, generally due to wrong app design. The bug was asking to throw a different exception to discriminate between those two unrelated cases, as most apps will call the non-silent token retrieval when catching `MsalUiRequiredException`, which in the case of throttling not only doesn't work but make things worse.

SameerK-MSFT: When a loop occurs (either due to app developer error or otherwise), the only way to break the loop is to ask user to login. So this exception is deliberately thrown. Also, in this case, the error message would be different "AadThrottledError" to distinguish it from the other exceptions.

djee-ms: @SameerK-MSFT  although the case you describe _can_ happen, this is not the only case where throttling occurs.
- There's no loop in the throttling case I logged the original bug for; we simply login from several tests at once in parallel, and those tests are short-lived (sometimes a few seconds), so the shear amount of requests in a short period of time causes the throttling.
- Asking the user to login is unrelated, because throttling equally occurs with silent or UI-based login. The `MsalUiRequiredException` tells the app that some stored creds are not available for silent login, and therefore an interactive login needs to happen. This is very different from what throttling is about, which is the app making requests too fast. There should be a different exception to allow 1) during development the devs to catch this misuse (our case) and fix the app logic, and 2) during prod if the app cannot be fixed for any reason, implement an app-level throttling strategy to reduce the number of requests and avoid service-side throttling. With the current fix, the app needs to catch `MsalUiRequiredException`, then also branch on the content of the exception message (substring search), which I believe is bad practice, and a separate exception type should be used instead.

PR #2815, opened 2021-08-10
Adding "always: true" to ensure scheduled builds run continuously.

trwalke:
Fixes # .
Fixing scheduled builds not running nightly

Adding "always: true" to ensure scheduled builds run always.
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml#running-even-when-there-are-no-code-changes

### DIFF
--- a/build/pipeline-ci.yaml
+++ b/build/pipeline-ci.yaml
@@ -11,6 +11,7 @@ schedules:
   branches:
     include:
     - master
+  always: true
 
 pr: none
 
--- a/build/pipeline-ci.yaml
+++ b/build/pipeline-ci.yaml
@@ -4,15 +4,6 @@ trigger:
     include:
     - 'master'
 
-# Create a daily midnight build for CI builds on master to ensure our build functions
-schedules:
-- cron: "0 0 * * *"
-  displayName: Daily midnight build
-  branches:
-    include:
-    - master
-  always: true
-
 pr: none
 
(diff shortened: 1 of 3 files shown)


bgavrilMS (approved):
Good find, thank you Travis

jennyf19 (approved):

build/pipeline-ci.yaml
```diff
@@ -11,6 +11,7 @@ schedules:
   branches:
```
pmaytak, line 11: Is there an additional benefit to running the CI build nightly? Should it be enough to run just the release build nightly, since it has everything?

pmaytak:
On slightly unrelated note: [template-ci-and-pr.yaml](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/build/template-ci-and-pr.yaml) doesn't seem to be used anywhere. Can we clean up, delete it?

SameerK-MSFT (approved):

build/pipeline-ci.yaml
trwalke: We used to have a reason before i made the updates to the pipeline as they were slightly different but now that they are so similar it should be fine to only run the release nightly and run the CI on merges to master

pmaytak (approved):

PR #2818, opened 2021-08-11
WAM + MSA-passthrough revamp

### LINKED ISSUE #2822: VS + WAM fix for MSA-PT
WAM does not work with MSA-PT config, even if the app is configured for MSA tokens.

bgavrilMS:
Fixes #2822

- simplify the MSA-PT protocol by getting the transfer token directly from the MSA plugin
- fix silent acquisition of tokens by forcing /organization 

**Testing**

ut

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
@@ -40,6 +40,12 @@ namespace Microsoft.Identity.Client.Instance
                   !TenantId.Equals(Constants.MsaTenantId);
         }
 
+        internal bool IsConsumers()
+        {
+            return TenantId.Equals(Constants.ConsumerTenant) ||
+                  TenantId.Equals(Constants.MsaTenantId);
+        }
+
         internal bool IsCommonOrganizationsOrConsumersTenant()
         {
             return IsCommonOrganizationsOrConsumersTenant(TenantId);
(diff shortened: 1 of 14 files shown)


src/client/Microsoft.Identity.Client/Instance/AadAuthority.cs
```diff
@@ -51,10 +57,10 @@ internal static bool IsCommonOrganizationsOrConsumersTenant(string tenantId)
                 s_tenantlessTenantNames.Contains(tenantId);
         }
 
-        internal override string GetTenantedAuthority(string tenantId)
+        internal override string GetTenantedAuthority(string tenantId, bool forceTenantless = false)
```
trwalke, line 60: What does forceTenantless represent here? How is it different from /common or /organizations which are also tenantless?

bgavrilMS: "l.m.o/12345-1234-1234".GetTenantAuthority("common", false)  returns "l.m.o/12345-1234-1234"\
"l.m.o/12345-1234-1234".GetTenantAuthority("common", true)  returns "l.m.o/common"

pmaytak: I would rename it to something like `useSpecifiedTenant` or `forceSpecifiedTenant`. Since I can call this method as `GetTenantedAuthority("12345-1234-1234", forceTenantless: true)` and it will return me a tenanted authority.

Also I would add some authority unit tests that test that new parameter.

pmaytak (changes requested):

tests/devapps/WAM/NetFrameworkWam/Form1.cs
```diff
@@ -246,7 +270,9 @@ private async Task<AuthenticationResult> RunAtiAsync(IPublicClientApplication pc
             var scopes = GetScopes();
 
             var builder = pca.AcquireTokenInteractive(scopes)
-                .WithUseEmbeddedWebView(true)
+                .WithUseEmbeddedWebView(false)
+                //.WithExtraQueryParameters("domain_hint=live.com")
```
trwalke, line 274: is this needed?

pmaytak (approved):

Issue #2781, opened 2021-07-19
[Bug] InvalidRegion_GoesToInvalidAuthority_Async test is failing


trwalke:
**Logs and network traces**
InvalidRegion_GoesToInvalidAuthority_Async test is failing.

"Test method Microsoft.Identity.Test.Integration.HeadlessTests.RegionalAuthIntegrationTests.InvalidRegion_GoesToInvalidAuthority_Async threw exception:
System.Net.Http.HttpRequestException: An error occurred while sending the request. ---> System.Net.WebException: The remote name could not be resolved: 'invalid.login.microsoft.com'"

**Which version of MSAL.NET are you using?**
current

PR #2819, opened 2021-08-11
Fixing regional test issue

trwalke:
Fixes # .
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2781
[conversation here](https://teams.microsoft.com/l/message/19:52a29b76367e45dda20a02de7184b4f4@thread.skype/1628720389242?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=e8095862-5f1f-4095-b996-6b78e479e4b4&parentMessageId=1628709938977&teamName=Azure%20AD%20Client%20SDK%20Team&channelName=Dotnet%20SDK%20discussions&createdTime=1628720389242)

Fixing test to point valid region. 

**Testing**
Integration test

**Performance impact**
None

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
         }
 
         [TestMethod]
-        public async Task RequestGoesToUserSpecifiedRegion_Async()
+        public async Task InvalidRegion_GoesToInvalidAuthority_Async()
         {
             // Arrange
             var factory = new HttpSnifferClientFactory();
@@ -82,13 +82,10 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
               "https://westus.r.login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/v2.0/token?allowestsrnonmsi=true",
               factory.RequestsAndResponses.Single().Item1.RequestUri.ToString());
 
-            AssertTelemetry(factory, $"{TelemetryConstants.HttpTelemetrySchemaVersion}|1004,{CacheInfoTelemetry.NoCachedAT:D},westus,3,3|0,1");
-
             _confidentialClientApplication = BuildCCA(settings, factory, true, TestConstants.Region);
             result = await GetAuthenticationResultAsync(settings.AppScopes, withForceRefresh: true).ConfigureAwait(false); // regional endpoint
             AssertTokenSourceIsIdp(result);
             AssertValidHost(true, factory, 1);
-            AssertTelemetry(factory, $"{TelemetryConstants.HttpTelemetrySchemaVersion}|1004,{CacheInfoTelemetry.ForceRefresh:D},centralus,2,1|0,1", 1);
         }
 
         private void AssertTelemetry(HttpSnifferClientFactory factory, string currentTelemetryHeader, int placement = 0)
--- a/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
@@ -68,22 +68,21 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
         }
 
         [TestMethod]
-        [Ignore] //See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2781
-        public async Task InvalidRegion_GoesToInvalidAuthority_Async()
+        public async Task RequestGoesToUserSpecifiedRegion_Async()
         {
             // Arrange
             var factory = new HttpSnifferClientFactory();
             var settings = ConfidentialAppSettings.GetSettings(Cloud.Public);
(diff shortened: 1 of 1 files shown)


trwalke:
See [conversation here](https://teams.microsoft.com/l/message/19:52a29b76367e45dda20a02de7184b4f4@thread.skype/1628720389242?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=e8095862-5f1f-4095-b996-6b78e479e4b4&parentMessageId=1628709938977&teamName=Azure%20AD%20Client%20SDK%20Team&channelName=Dotnet%20SDK%20discussions&createdTime=1628720389242) for details

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
```diff
@@ -68,22 +68,21 @@ public async Task AcquireTokenToRegionalEndpointAsync()
...
         [TestMethod]
-        [Ignore] //See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2781
-        public async Task InvalidRegion_GoesToInvalidAuthority_Async()
+        public async Task RequestGoesToUserSpecifiedRegion_Async()
         {
             // Arrange
             var factory = new HttpSnifferClientFactory();
             var settings = ConfidentialAppSettings.GetSettings(Cloud.Public);
-            _confidentialClientApplication = BuildCCA(settings, factory, true, "invalid");
+            _confidentialClientApplication = BuildCCA(settings, factory, true, "westus");
```
bgavrilMS, line 76: The point of this test is to make sure that MSAL & ESTS-R behave consistently when a bad region is used. MSAL doesn't know what a "bad" region is, if the developer tells us to use "XZY", we'll use "XZY". I expect ESTS-R to be able to serve this request.

As such: 

- the original name of the test still makes sense to me
- `westus` is not a valid region name, so you are testing the right thing :)
- If you use `invalid` what happens? `invalid` is more descriptive.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
```diff
@@ -68,22 +68,21 @@ public async Task AcquireTokenToRegionalEndpointAsync()
...
             Environment.SetEnvironmentVariable(TestConstants.RegionName, TestConstants.Region);
             AuthenticationResult result = await GetAuthenticationResultAsync(settings.AppScopes).ConfigureAwait(false); // regional endpoint
             AssertTokenSourceIsIdp(result);
             Assert.AreEqual(
-              "https://invalid.login.microsoft.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/v2.0/token?allowestsrnonmsi=true",
+              "https://westus.r.login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/v2.0/token?allowestsrnonmsi=true",
               factory.RequestsAndResponses.Single().Item1.RequestUri.ToString());
 
-            AssertTelemetry(factory, $"{TelemetryConstants.HttpTelemetrySchemaVersion}|1004,{CacheInfoTelemetry.NoCachedAT:D},invalid,3,3|0,1");
+            AssertTelemetry(factory, $"{TelemetryConstants.HttpTelemetrySchemaVersion}|1004,{CacheInfoTelemetry.NoCachedAT:D},westus,3,3|0,1");
```
bgavrilMS, line 85: Consider removing this.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
trwalke: When using invalid the following error is produced: The remote name could not be resolved: 'invalid.login.microsoft.com

Perhaps ESTS-R changed the behavior of this specific address?

pmaytak: @bgavrilMS 
- Yes, initially `InvalidRegion_GoesToInvalidAuthority_Async` tested the behavior you describe and we did expect it to work with ESTS-R. But as Travis tested, invalid regions don't work.
- `westus` is a valid region, that's why the test passes with it.
- Since invalid regions don't work, I suggested to rename the test to `RequestGoesToUserSpecifiedRegion_Async` and test that user-specified region takes precedence over the auto detected one.

jennyf19: so did the behavior from ests-r change or did we not test correctly? If the behavior changed, we should tell them, as they just broke our customers. Based on what i'm understanding, the customer can give any region and that host should be used, and if it's not available, ests-r should serve the request from the global endpoint. But that's no happening, we are getting an error. is that correct? and before it was working with an invalid region?

jennyf19 (changes requested):
:clock1:

jennyf19:
doesn't seem like the test is working yet, or that we are clear on the expected e2e.

---
In reply to: [897390527](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2819#issuecomment-897390527) [](http://example.com/codeflow?ancestors=897390527,897390527)

pmaytak (changes requested):
Based on the reply from the regional team, we can probably update this test to send `invalid` region and catch a network exception.

PR #2820, opened 2021-08-12
Add public API for long-running OBO methods with custom cache key and do not use refresh token for normal OBO

### LINKED ISSUE #2733: [Feature Request] Allow customers to provide a key to access the OBO token for long running process
1. Long running processes in web APIs are now possible since OBO tokens are refreshed automatically; however this requires the web API to keep the user assertion, to call AcquireTokenOnBehalfOf, even if it has expired. The following article [How to get an OBO token in an event handler or a long running process](https://github.com/AzureAD/microsoft-identity-web/wiki/get-token-in-event-handler) shows how to do today. But this we'd want to provide a better mechanism, for instance by using an optional claim like the sid (session id), but this could be another key. This is for a web API called for new users (contrary to the next case)
1. Doing the ADAL to MSAL migration, some partners also had a refresh token with ADAL.NET, which they used to get a new OBO token. They don't have the user assertion any longer. For them, the migration would be to call once `AcquireTokenByRefreshToken`, and then `AcquireTokenSilent` (but this would not be an OBO token)

Have 2 new public APIs in IConfidentialClientApplication:

```CSharp
interface IConfidentialClientApplication
{
        /// <summary>
        /// Acquires an access token for this web API from the authority configured in the application,
        /// in order to access another downstream protected web API on behalf of a user using the OAuth 2.0 On-Behalf-Of flow.
        /// See https://aka.ms/msal-net-on-behalf-of.
        /// This confidential client application was itself called with a token which will be provided in the
        /// <paramref name="userToken">userToken</paramref> parameter.
        /// </summary>
        /// <param name="scopes">Scopes requested to access a protected API</param>
        /// <param name="userToken">A JSON Web Token which was used to call the web API and contains the credential information
        /// about the user on behalf of whom to get a token.</param>
        /// <param name="longRunningProcessSessionKey">Key by which to look up the token in the cache.
        /// If null, it will be set to the assertion hash by default.</param>
        /// <returns>A builder enabling you to add optional parameters before executing the token request</returns>
        AcquireTokenOnBehalfOfParameterBuilder InitiateLongRunningProcessInWebApi(IEnumerable<string> scopes, string userToken, ref string longRunningProcessSessionKey)


        /// <summary>
        /// Retrieves an access token from the cache using the provided cache key that can be used to
        /// access another downstream protected web API on behalf of a user using the OAuth 2.0 On-Behalf-Of flow.
        /// See https://aka.ms/msal-net-on-behalf-of.
        /// </summary>
        /// <remarks>
        /// This method is intended to be used in the long running processes inside of web APIs.
        /// </remarks>
        /// <param name="scopes">Scopes requested to access a protected API</param>
        /// <param name="longRunningProcessSessionKey">Key by which to look up the token in the cache</param>
        /// <returns>A builder enabling you to add optional parameters before executing the token request</returns>
        AcquireTokenOnBehalfOfParameterBuilder AcquireTokenInLongRunningProcess(IEnumerable<string> scopes, string longRunningProcessSessionKey);
}
```

**work to plan**:

- [  ] New methods `InitiateLongRunningProcessInWebApi` and `AcquireTokenInLongRunningProcess` on `IConfidentialClientApplication` (Or alternatively extension methods on the IConfidentialClientApplication, or ILongRunningWebApi inheriting from IConfidentialClientApplication
- [  ] Store the key in the OBO record in the token cache
- [  ] Only get a refresh token for OBO tokens when calling `AcquireTokenInLongRunningProcess` (and `AcquireTokenInLongRunningProcess` when the token is refreshed), but no longer in `AcquireTokenOnBehaldOf`

pmaytak:
Fixes #2733.

- New public API for long-running OBO
  - Adds `InitiateLongRunningProcessInWebApi` that allows to specify a cache key which will be used for searching for an OBO access and refresh tokens instead of a user assertion hash as usual.
  - Adds `AcquireTokenInLongRunningProcess` to retrieve the token from the cache with that key without providing a user assertion.
  - Created new interface `ILongRunningWebApi`.
 - For normal OBO flow, removes refresh token from the token response.
   - Also optimization - if access token is not found, skips checking for refresh tokens.

Variations of assertions and keys for different OBO methods:
| OBO method | User assertion | User provided cache key |
|-------------------------------------|----------------|-------------------------|
| InitiateLongRunningProcessInWebApi  | Not null       | Not null                |
| AcquireTokenInLongRunningProcess    | Null           | Not null                |
| AcquireTokenOnBehalfOf              | Not null       | Null                    |

Logic flow:
**InitiateLongRunningProcessInWebApi (scopes, assertion, key)**
- AT is not found in cache by cache key - return null, look for RT
  - RT is found by cache key - Fetch new AT using existing RT and save with cache key
  - RT is not found by cache key - Fetch new AT with assertion and save with cache key
- AT is found in cache by cache key - return it

**AcquireTokenInLongRunningProcess (scopes, key)**
- AT is not found in cache by cache key - throw exception (key expected to exist)
- AT is found in cache by cache key 
  - If expired - look for RT
    - RT is found by cache key - Fetch new AT using existing RT and save with cache key
    - RT is not found by cache key - throw exception (RT is expected to exist)
  - If AT is valid - return it

**Testing**
Manual tests. Added tests for new public API flows. Tests for when calling both, normal and long-running OBO methods. Tests that suggested cache expiry should work for normal OBO. Tests that normal OBO should not have a refresh token and should not use that flow.

**Performance impact**
Less memory used for normal OBO since we don't cache RT anymore.


src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -182,15 +184,16 @@ internal override JObject ToJObject()
             SetItemIfValueNotNull(json, StorageJsonKeys.Realm, TenantId);
             SetItemIfValueNotNull(json, StorageJsonKeys.Target, _scopes);
             SetItemIfValueNotNull(json, StorageJsonKeys.UserAssertionHash, UserAssertionHash);
+            SetItemIfValueNotNull(json, StorageJsonKeys.OboCacheKey, OboCacheKey);
```
bgavrilMS, line 187: Do we need a new field in the token cache schema? I think we can reuse the UserAssertionHash. 

If the user gives us a key, we will always use it. Otherwise, we will use hash(incoming_assertion)

src/client/Microsoft.Identity.Client/Cache/Items/MsalRefreshTokenCacheItem.cs
```diff
@@ -53,6 +53,7 @@ internal MsalRefreshTokenCacheItem()
         public string FamilyId { get; set; }
 
         internal string UserAssertionHash { get; set; }
+        internal string OboCacheKey { get; set; }
```
bgavrilMS, line 56: As above, try to reuse UserAssertionHash. Maybe rename "UserAssertionHash" variable name to "OBOSessionKey" (but do not change the underlying JSON).

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -75,18 +75,18 @@ internal class AuthenticationRequestParameters
...
         #region Authority
 
-        public AuthorityManager AuthorityManager { get; set; } 
+        public AuthorityManager AuthorityManager { get; set; }
 
         /// <summary>
         /// Authority is the URI used by MSAL for communication and storage
         /// During a request it can be updated: 
-        /// - with the preffered enviroment
+        /// - with the proffered environment
```
bgavrilMS, line 83: proffered?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -71,6 +71,7 @@ public sealed partial class TokenCache : ITokenCacheInternal
                         requestParams.AuthenticationScheme.KeyId)
                     {
                         UserAssertionHash = requestParams.UserAssertion?.AssertionHash,
+                        OboCacheKey = requestParams.OboCacheKey,
```
bgavrilMS, line 74: Would it be possible to leave this class alone, i.e. just rename UserAssertionHash to OboSessionKey
Then the callers of this class would make sure to set `requestParams.OboSessionKey` to either user key or assertion hash?

jmprieur:
@pmaytak, as you said, there is a risk that the API is miss used

I think we might want to keep AcquireTokenOnBehalfOf as is today, and add 2 new methods:
- `InitiateLongRunningProcessInWebAPI(scopes, jwt, ref longRunningProcessSessionKey=null)`. This method enables the developer to initiate a long running process in a web API, by specifying the jwt used to call the web API, the scopes, and optionally a long running process session key. This is really a string that either the developer provides, or is otherwise returned by the method (and could be the jwt hash)
- `AcquireTokenInLongRunningProcess(scopes, longRunningProcessSessionKey)` which acquires the token.

Both methods could return `AuthenticationResult`.
We need to decide what happens if the first method is called with an existing key (probably throw), and if the second is called with an un-existant key (throw for sure)

Example of usage:

```CSharp
/// <summary>
        /// This methods the processing of user data where the web API periodically checks the user
        /// date (think of OneDrive producing albums)
        /// </summary>
        private void RegisterPeriodicCallbackForLongProcessing()
        {
            // Get the token incoming to the web API - we could do better here.
            string key = null;            
            cca.InitiateLongRunningProcessInWebAPI(scopes, jwt, ref key);

            // Build the URL to the callback controller, based on the request.
            var request = HttpContext.Request;
            string url = request.Scheme + "://" + request.Host + request.Path.Value.Replace("todolist", "callback") + $"?key={key}";

            // Setup a timer so that the API calls back the callback every 10 mins.
            Timer timer = new Timer(async (state) =>
            {
                HttpClient httpClient = new HttpClient();
                
                var message = await httpClient.GetAsync(url);
            }, null, 1000, 1000 * 60 * 1);  // Callback every minute
        }
``` 

```CSharp
using Microsoft.AspNetCore.Authorization;
using Microsoft.AspNetCore.Http;
using Microsoft.AspNetCore.Mvc;
using Microsoft.Extensions.Logging;
using Microsoft.Identity.Web;
using Microsoft.Identity.Web.Resource;
using System;
using System.IdentityModel.Tokens.Jwt;
using System.Linq;
using System.Net.Http;
using System.Security.Claims;
using System.Threading;
using System.Threading.Tasks;
using TodoListService.Models;

namespace TodoListService.Controllers
{
    [Authorize]
    [Route("api/[controller]")]
    //[RequiredScope("access_as_user")] 
    public class CallbackController : Controller
    {
        private readonly IConfidentialClientApplication cca;    

/* constructor omitted */

        // GET: api/values
        // [RequiredScope("access_as_user")]
        [HttpGet]
        [AllowAnonymous]
        public async Task GetAsync(string key)
        {
                var result = await cca.AcquireTokenInLongRunningProcess(new string[] { "user.read" }, key).ConfigureAwait(false); 
            }
        }
    }
}
````

**Comment by pmaytak** [comment]
@jmprieur
I implemented the changes in commits https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2820/commits/dfb40f77652c62a71a548730c753549469c3617a and https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2820/commits/e667465ba2f7ab8baa18a7bbcdafaffcd31c08ae.

Tests are in [OboRequestTests](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/pmaytak/obo-key/tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs). ([Test](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e667465ba2f7ab8baa18a7bbcdafaffcd31c08ae/tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs#L275)) for `AcquireTokenInLongRunningProcess` when key does not exist. [Tests](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e667465ba2f7ab8baa18a7bbcdafaffcd31c08ae/tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs#L200) for `InitiateLongRunningProcessInWebAPI` when cache key exists.

Should we specifically mention OBO in the method names to make it more clear? Something like `AcquireTokenForOboInLongRunningProcess`  and `InitiateLongRunningProcessForOboInWebApi`although it does make names longer.

File: src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs [hunk 5]
```diff
@@ -76,6 +75,34 @@ public partial interface IConfidentialClientApplication : IClientApplicationBase
...
+        /// This confidential client application was itself called with a token which will be provided in the
+        /// <paramref name="userToken">userToken</paramref> parameter.
+        /// </summary>
+        /// <param name="scopes">Scopes requested to access a protected API</param>
+        /// <param name="userToken">A JSON Web Token which was used to call the web API and contains the credential information
+        /// about the user on behalf of whom to get a token.</param>
+        /// <param name="longRunningProcessSessionKey">Key by which to look up the token in the cache.
+        /// If null, it will be set to the assertion hash by default.</param>
+        /// <returns>A builder enabling you to add optional parameters before executing the token request</returns>
+        AcquireTokenOnBehalfOfParameterBuilder InitiateLongRunningProcessInWebApi(IEnumerable<string> scopes, string userToken, ref string longRunningProcessSessionKey);
```
bgavrilMS, line 91: Sadly you cannot modify an existing interface. It's a breaking change :( The solution is to add a new interface to the existing ones: 

```csharp
 public sealed partial class ConfidentialClientApplication
        : ClientApplicationBase,
            IConfidentialClientApplication,
            IConfidentialClientApplicationWithCertificate,
            IByRefreshToken
```

bgavrilMS: `ILongRunningProcessWebApi` ? @jmprieur for naming ideas of the new interface

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -367,6 +372,23 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
...
             CacheInfoTelemetry cacheInfoTelemetry = CacheInfoTelemetry.None;
 
+            // Verify appropriate scenario for OBO tokens in long running process
+            if (requestParams.ApiId == ApiEvent.ApiIds.AcquireTokenOnBehalfOf)
+            {
+                // Validates correct AcquireTokenInLongRunningProcess call (user assertion is null in this case)
+                if (requestParams.UserAssertion == null && !string.IsNullOrEmpty(requestParams.OboCacheKey) && tokenCacheItems.Count == 0)
+                {
+                    logger.Error("[FindAccessTokenAsync] AcquireTokenInLongRunningProcess was called and OBO token was not found in the cache.");
+                    throw new MsalClientException(MsalError.OboCacheKeyNotInCacheError, MsalErrorMessage.OboCacheKeyNotInCache);
```
bgavrilMS, line 382: Shouldn't this throw an `MsalUiRequiredException` instead? CC @jmprieur for exception behavior.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -367,6 +372,23 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
...
+                }
+                // Validates correct InitiateLongRunningProcessInWebAPI call (user assertion is specified in this case)
+                if (requestParams.UserAssertion != null && !string.IsNullOrEmpty(requestParams.OboCacheKey) && tokenCacheItems.Count > 0)
```
bgavrilMS, line 385: I think it will be simpler to override the existing token instead of throwing? If the key is the user id and the user re-logs in. CC @jmprieur for exception behavior.

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -1004,5 +1005,23 @@ public static class MsalError
...
         public const string InitializeProcessSecurityError = "initialize_process_security_error";
+
+        /// <summary>
+        /// <para>What happens?</para>The token cache already contains a token with an OBO cache key that
+        /// matches the <c>longRunningProcessSessionKey</c> passed into <see cref="IConfidentialClientApplication.AcquireTokenInLongRunningProcess"/>.
+        /// <para>Mitigation</para>Call <see cref="IConfidentialClientApplication.InitiateLongRunningProcessInWebApi"/> with a new <c>longRunningProcessSessionKey</c>
+        /// that does not exist in the token cache or call <see cref="IConfidentialClientApplication.AcquireTokenInLongRunningProcess"/> with an already used
+        /// <c>longRunningProcessSessionKey</c>.
+        /// </summary>
+        public const string OboCacheKeyAlreadyInCacheError = "obo_cache_key_in_cache_error";
```
trwalke, line 1016: I think it would be more clear to add the word "already" to the error code. Looks very similar to the other.
maybe "obo_cache_key_in_cache_already_error"

tests/Microsoft.Identity.Test.Unit/CacheTests/CacheSerializationTests.cs
```diff
@@ -834,7 +834,7 @@ private void AssertAccessTokenCacheItemsAreEqual(MsalAccessTokenCacheItem expect
             Assert.AreEqual(expected.GetKey().ToString(), actual.GetKey().ToString());
             CollectionAssert.AreEqual(expected.ScopeSet.ToList(), actual.ScopeSet.ToList(), nameof(actual.ScopeSet));
             Assert.AreEqual(expected.TenantId, actual.TenantId, nameof(actual.TenantId));
-            Assert.AreEqual(expected.UserAssertionHash, actual.UserAssertionHash, nameof(actual.UserAssertionHash));
```
trwalke, line 837: So the cache key is replacing the userAssertionHash for these tests?

trwalke (changes requested):
Minor comment

tests/Microsoft.Identity.Test.Unit/CacheTests/CacheSerializationTests.cs
pmaytak: This is just renaming of a property to better reflect what it can represent. Previously it was only a hash, now it can be a hash or a user provided cache key - either way both act as cache keys.

bgavrilMS:
LGTM

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -127,30 +127,40 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
         private async Task<AuthenticationResult> RefreshRtOrFetchNewAccessTokenAsync(CancellationToken cancellationToken)
         {
-            // Look for a refresh token
-            MsalRefreshTokenCacheItem cachedRefreshToken = await CacheManager.FindRefreshTokenAsync().ConfigureAwait(false);
-
-            // If a refresh token is not found, fetch a new access token
-            if (cachedRefreshToken != null)
+            if (IsLongRunningObo())
```
bgavrilMS, line 130: Please add some logging statement(s) around this new branch of the code.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -186,5 +196,10 @@ protected override void EnrichTelemetryApiEvent(ApiEvent apiEvent)
 
             return new KeyValuePair<string, string>(Constants.CcsRoutingHintHeader, _ccsRoutingHint);
         }
+
+        private bool IsLongRunningObo()
+        {
+            return !string.IsNullOrEmpty(AuthenticationRequestParameters.OboCacheKey);
```
bgavrilMS, line 202: Hmm, maybe we need to rename OboCacheKey to LongRunningOboCacheKey or smth?

bgavrilMS (commented):
LGTM. Anything pending?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1173,5 +1207,21 @@ bool ITokenCacheInternal.HasTokensNoLocks()
                 .ToList()
                 .ForEach(accItem => _accessor.DeleteAccount(accItem));
         }
+
+        // Returns whether InitiateLongRunningProcessInWebAPI was called (user assertion is specified in this case)
+        private bool InitiateLongRunningOboWasCalled(AuthenticationRequestParameters requestParameters)
+        {
+            return requestParameters.ApiId == ApiEvent.ApiIds.AcquireTokenOnBehalfOf &&
```
bgavrilMS, line 1214: Is there a way to find out via telemetry if folks call AcquireTokenOBO or this new API? Maybe we should have a different ApiId here?

pmaytak:
> LGTM. Anything pending?

Hunting down a possible bug. Adding tests for more scenarios on how OBO methods are called.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -127,30 +127,46 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
         }
 
         private async Task<AuthenticationResult> FetchNewAccessTokenAsync(CancellationToken cancellationToken)
         {
             var msalTokenResponse = await SendTokenRequestAsync(GetBodyParameters(), cancellationToken).ConfigureAwait(false);
+            
+            // We always retrieve a refresh token in OBO but we don't want to cache it for normal OBO flow, only for long-running OBO
+            if (!IsLongRunningObo())
+            {
+                msalTokenResponse.RefreshToken = null;
```
pmaytak, line 167: For OBO, AAD returns a refresh token even if not sending `offline_access` scope. So solution is to set it to null after receiving the response.

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -124,6 +125,47 @@ public sealed partial class ConfidentialClientApplication
                 userAssertion);
         }
 
+        /// <inheritdoc />
```
bgavrilMS, line 128: nit: I never understood `inheritdoc`. Do they actually work in .NET?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests2.cs
```diff
@@ -43,7 +43,7 @@ public class OboTests2
         private static readonly string[] s_publicCloudOBOServiceScope = { "api://23c64cd8-21e4-41dd-9756-ab9e2c23f58c/access_as_user" };
         private static readonly string[] s_arlingtonOBOServiceScope = { "https://arlmsidlab1.us/IDLABS_APP_Confidential_Client/user_impersonation" };
 
-        //TODO: acquire scenario specific client ids from the lab resonse
+        //TODO: acquire scenario specific client ids from the lab response
```
bgavrilMS, line 46: Can we have an integration test (or modify some existing ones) around this new flow? I know it's well covered with unit tests, but it's helpful to have integration tests for high level API to be able to get Fiddler traces easily etc. 

I'm thinking: 

- one test where the AT is still valid + normal key
- one test where the AT is no longer valid (expire it or wipe it) + custom key

bgavrilMS (changes requested):
LGTM, main comment is around integration test

src/client/Microsoft.Identity.Client/ILongRunningWebApi.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Methods for long running processes in web APIs
+    /// </summary>
+    public interface ILongRunningWebApi
```
trwalke, line 11: Should probably mentiontion the exceptions that will be thrown by these apis. Similar to the comment on `AcquireTokenSilent`

src/client/Microsoft.Identity.Client/ILongRunningWebApi.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// about the user on behalf of whom to get a token.</param>
+        /// <param name="longRunningProcessSessionKey">Key by which to look up the token in the cache.
+        /// If null, it will be set to the assertion hash by default.</param>
+        /// <returns>A builder enabling you to add optional parameters before executing the token request</returns>
+        AcquireTokenOnBehalfOfParameterBuilder InitiateLongRunningProcessInWebApi(IEnumerable<string> scopes, string userToken, ref string longRunningProcessSessionKey);
+
+        /// <summary>
+        /// Retrieves an access token from the cache using the provided cache key that can be used to
+        /// access another downstream protected web API on behalf of a user using the OAuth 2.0 On-Behalf-Of flow.
+        /// See https://aka.ms/msal-net-on-behalf-of.
```
trwalke, line 31: >behalf [](http://example.com/codeflow?start=43&length=6)

im assuming we will have documentation for the long running process in the wiki?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: Im not sure it needs to, it may be that the developer simply hasnt supplied the already acquired userAssertion using `InitiateLongRunningProcess` right?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -420,6 +428,18 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
 
             CacheRefreshReason cacheInfoTelemetry = CacheRefreshReason.NotApplicable;
 
+            if (AcquireTokenInLongRunningOboWasCalled(requestParams) && tokenCacheItems.Count == 0)
+            {
+                logger.Error("[FindAccessTokenAsync] AcquireTokenInLongRunningProcess was called and OBO token was not found in the cache.");
```
trwalke, line 433: Im wondering if it makes sense to indicate that the developer needs to call InitiateLongRunningProcess first?

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -405,5 +405,9 @@ public static string CertMustHavePrivateKey(string certificateName)
         public const string TenantOverrideNonAad = "WithTenantId can only be used when an AAD authority is specified at the application level.";
 
         public const string RegionalAndAuthorityOverride = "You configured WithAuthority at the request level, and also WithAzureRegion. This is not supported when the environment changes from application to request. Use WithTenantId at the request level instead.";
+
+        public const string OboCacheKeyAlreadyInCache = "The token cache already contains a token with an OBO cache key that matches the longRunningProcessSessionKey passed into IConfidentialClientApplication.AcquireTokenInLongRunningProcess method. Call IConfidentialClientApplication.InitiateLongRunningProcessInWebApi method with a new longRunningProcessSessionKey that does not exist in the token cache or call IConfidentialClientApplication.AcquireTokenInLongRunningProcess method with an already used longRunningProcessSessionKey.";
```
trwalke, line 409: >OboCacheKeyAlreadyInCache [](http://example.com/codeflow?start=28&length=25)

Might be a good idea for supportability to add the aka.ms link here too. Sometimes people run into issues and dont know where to look in the code to figure out what to do. We also add these links for other messages

trwalke (approved):
Minor comments, Logic looks good overall

trwalke:
one more thing, is there a test that validates that the normal OBO flow removes refresh token from the token response?

pmaytak:
> one more thing, is there a test that validates that the normal OBO flow removes refresh token from the token response?

This one tests that if AT is expired, OBO call does OBO flow and not RT flow.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b4e8fbd9858d2da89e89dd573beb14269ad73243/tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs#L74

This one tests that when AT is returned from cache, there are not RT tokens in cache.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b4e8fbd9858d2da89e89dd573beb14269ad73243/tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs#L156

In contrast, this tests that RT exists for long-running OBO.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b4e8fbd9858d2da89e89dd573beb14269ad73243/tests/Microsoft.Identity.Test.Unit/RequestsTests/OboRequestTests.cs#L187

@trwalke Should I add some other tests?

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
pmaytak: Yes. Pulls documentation from base class or interface and merges it with any XML tags that are defined on the current member. We should use it.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#inheritdoc

bgavrilMS:
> Logic flow: **InitiateLongRunningProcessInWebApi (scopes, assertion, key)**
> 
> * AT is not found in cache by cache key - return null, look for RT
>   
>   * RT is found by cache key - Fetch new AT using existing RT and save with cache key
>   * RT is not found by cache key - Fetch new AT with assertion and save with cache key
> * AT is found in cache by cache key - throw exception (key is not expected to exist)
> 


Should we really throw exceptions if `Initiate` finds an AT with the same key? Why not just override the existing token? Imagine this scenario: 

- app developer decides the key is `oid` (this is the case of some services today!)
- service uses L2 caching
- machine 1 calls Initiate with key `oid1`
- machine 2 also calls initiate with key `oid1` EXCEPTION

CC @jmprieur for opinion here.

src/client/Microsoft.Identity.Client/ILongRunningWebApi.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// Acquires an access token for this web API from the authority configured in the application,
+        /// in order to access another downstream protected web API on behalf of a user using the OAuth 2.0 On-Behalf-Of flow.
+        /// See https://aka.ms/msal-net-on-behalf-of.
+        /// This confidential client application was itself called with a token which will be provided in the
+        /// <paramref name="userToken">userToken</paramref> parameter.
+        /// </summary>
+        /// <param name="scopes">Scopes requested to access a protected API</param>
+        /// <param name="userToken">A JSON Web Token which was used to call the web API and contains the credential information
+        /// about the user on behalf of whom to get a token.</param>
+        /// <param name="longRunningProcessSessionKey">Key by which to look up the token in the cache.
```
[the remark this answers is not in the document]
bgavrilMS: @pmaytak - let's add `sid` claim as the suggested key to use, and add a remark explaining that MSAL cannot read tokens on its own.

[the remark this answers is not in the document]
hpsin: @bgavrilMS is it correct to read this as "MSAL performs the OBO without validating who the token is for"? I'd suggest we vet this path for confused deputy attack surface.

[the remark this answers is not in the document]
bgavrilMS: @hpsin - MSAL doesn't look at incoming token at all. It's up to the web api to perform any validations. I believe Ms.Id.Web does some of that.

I may not understand the question, but allow me to elaborate. 

Today, we advise people wanting to implement the "long running process OBO" scenario to keep around the `incoming_assertion`. This is to avoid the "confused deputy" attack, i.e. the AT / RT from OBO can only be retrieved if the original assertion is found. However, this is pretty complicated, and we end up asking folks to keep around an expired JWT. They mostly keep it around in their own cache as "username" -> "original_jwt".  

`AcquireTokenInLongRunningProcess(scopes, key)` aims to fix that. It is a slightly more flexible version of `AcquireTokenOnBehalfOf(scopes, incomming_assertion)`, to allow the web api to let go of the expired incomming_assertion and to remember only the `sid`. Internally, MSAL will locate the AT / RT in its caches by this key. 

Shall we meet to discuss this in more detail?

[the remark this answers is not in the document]
hpsin: Thanks, I think that answers my question in general. Do we have any kind of ratchet in place that tries to ensure that the web api has performed validation?

pmaytak:
> > Logic flow: **InitiateLongRunningProcessInWebApi (scopes, assertion, key)**
> > 
> > * AT is not found in cache by cache key - return null, look for RT
> >   
> >   * RT is found by cache key - Fetch new AT using existing RT and save with cache key
> >   * RT is not found by cache key - Fetch new AT with assertion and save with cache key
> > * AT is found in cache by cache key - throw exception (key is not expected to exist)
> 
> Should we really throw exceptions if `Initiate` finds an AT with the same key? Why not just override the existing token? Imagine this scenario:
> 
> * app developer decides the key is `oid` (this is the case of some services today!)
> * service uses L2 caching
> * machine 1 calls Initiate with key `oid1`
> * machine 2 also calls initiate with key `oid1` EXCEPTION
> 
> CC @jmprieur for opinion here.

@bgavrilMS Removed that exception in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2820/commits/627b4d9325ea74186dd1e0afa59abe18d9b2f22d

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests2.cs
pmaytak: Added these extra tests in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2820/commits/f89db0b8abd37eeaf37088250881d09be26bdf8b

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
```diff
@@ -51,7 +52,279 @@ public void TestInitialize()
...
+            var cca = BuildCCA(userAuthResult.TenantId);
+
+            string oboCacheKey = null;
+            UserAssertion userAssertion = new UserAssertion(userAuthResult.AccessToken);
+
+            // InitiateLR - Empty cache - AT via OBO flow (new AT, RT cached)
+            var result = await cca.InitiateLongRunningProcessInWebApi(s_scopes, userAuthResult.AccessToken, ref oboCacheKey)
+                                    .ExecuteAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(result);
```
bgavrilMS, line 139: nit: add an assertion that oboCacheKey is now `hash(userAssertion)`

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
```diff
@@ -51,7 +52,279 @@ public void TestInitialize()
...
+            // AcquireLR - AT from cache
+            result = await cca.AcquireTokenInLongRunningProcess(s_scopes, oboCacheKey).ExecuteAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(result);
+            Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+
+            // AcquireNormal - AT from cache
+            result = await cca.AcquireTokenOnBehalfOf(s_scopes, userAssertion).ExecuteAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(result);
```
bgavrilMS, line 153: nit: this is never null.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ILongRunningWebApi.cs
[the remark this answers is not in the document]
jmprieur: > @pmaytak - let's add `sid` claim as the suggested key to use, and add a remark explaining that MSAL cannot read tokens on its own.

We cannot crack-open the token, so the claim would have to be provided by the application to MSAL.NET. Also, `sid` is an optional claim, This requires an aka.ms link.

src/client/Microsoft.Identity.Client/ILongRunningWebApi.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// Methods for long running processes in web APIs
+    /// </summary>
+    public interface ILongRunningWebApi
+    {
+        /// <summary>
+        /// Acquires an access token for this web API from the authority configured in the application,
+        /// in order to access another downstream protected web API on behalf of a user using the OAuth 2.0 On-Behalf-Of flow.
+        /// See https://aka.ms/msal-net-on-behalf-of.
```
jmprieur, line 16: Shouldn't it be https://aka.ms/msal-net-long-running-obo ?

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -159,8 +159,17 @@ public bool IsConfidentialClient
 #endif
             }
         }
+
+        /// <remarks>
+        /// Is null when <see cref="ILongRunningWebApi.AcquireTokenInLongRunningProcess"/> is called.
```
jmprieur, line 164: is it in the public API?

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -1030,5 +1031,14 @@ public static class MsalError
         /// <para>Mitigation</para> Use WithTenantId at the request level instead.
         /// </summary>
         public const string RegionalAndAuthorityOverride = "authority_override_regional";
+
+        /// <summary>
+        /// <para>What happens?</para>The token cache does not contain a token with an OBO cache key that
+        /// matches the <c>longRunningProcessSessionKey</c> passed into <see cref="ILongRunningWebApi.AcquireTokenInLongRunningProcess"/>.
+        /// <para>Mitigation</para>Call <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi"/> with a new <c>longRunningProcessSessionKey</c>
```
jmprieur, line 1038: actually if it does not exist the cache, how can we call with a key that does not exist? Is it the right sentence?

jmprieur (commented):
I think we need to improve a few XLM comments, otherwise LGTM

tested with Identity.Web and an id.web dev sample:  https://github.com/AzureAD/microsoft-identity-web/pull/1523

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
pmaytak: No, internal. Just added a comment for our clarification.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
pmaytak: Added.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
pmaytak: Updated.

src/client/Microsoft.Identity.Client/ILongRunningWebApi.cs
pmaytak: Yes, updated.

src/client/Microsoft.Identity.Client/MsalError.cs
pmaytak: Rephrased a little.

>         /// <para>What happens?</para>The token cache does not contain a token with an OBO cache key that
>         /// matches the <c>longRunningProcessSessionKey</c> passed into <see cref="ILongRunningWebApi.AcquireTokenInLongRunningProcess"/>.
>         /// <para>Mitigation</para> Call <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi"/> with this <c>longRunningProcessSessionKey</c> 
>         /// first or call <see cref="ILongRunningWebApi.AcquireTokenInLongRunningProcess"/> with an already used <c>longRunningProcessSessionKey</c>.

Issue #2285, opened 2020-12-08
[Feature Request] add an override of the constructor of PopAuthenticationConfiguration with an HttpVerb parameter


trwalke:
**Describe the solution you'd like**
Add an override of the constructor of PopAuthenticationConfiguration with an HttpVerb parameter so that this is more discoverable. There should also be a property to allow changes after initialization of the configuration object.

**Additional context**
This is a minor modification to this feature here: [POP](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1946)

PR #2840, opened 2021-08-24
Add custom nonce support to POP

### LINKED ISSUE #2809: [Feature Request] Add experimental POP support to public client with special TS nonce
API governance doc: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/2614?_a=files&path=%2F[Xplat]%20AT-PoP%20Client%20Side%2Fserver_nonce%2Fclient_handling.md

Proposed changes in MSAL

- [ ] Add back experimental support for POP to public clients. It should be noted that this does not provide security benefits because the RT remains a bearer RT

Logic for extracting the nonce:
- [ ] Add a new (mandatory?) parameter to `PoPAuthenticationConfiguration`:
HttpResponseHeaders HttpResponseHeaders; This has to contain either a WWWAuthenticate or an AutheticateInfo header.
- [ ] MSAL will parse either of these headers and find a special nonce parameter. 

OR

- [ ] Add a new (mandatory?) string parameter called `nonce` to the `PoPAuthenticationConfiguration` object. 
- [ ] Extend the existing `WwwAuthenticateParameters`  to extract this nonce
- [ ] Add a similar `AuthorityInfo` object that parses the Authority-Info headers and exposes the nonce

Then use the nonce
- [ ] Use the nonce as the `ts` claim when forming the POP token

bgavrilMS:
Fixes #2809 

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
@@ -3,39 +3,86 @@
 
 using System;
 using System.Net.Http;
-using System.Net.Http.Headers;
 using Microsoft.Identity.Client.AuthScheme.PoP;
 
 namespace Microsoft.Identity.Client.AppConfig
 {
 
     /// <summary>
-    /// Configuration properties used to construct a proof of possession request.
+    /// Details about the HTTP request and configuration properties used to construct a proof of possession request.
     /// </summary>
+    /// <remarks> 
+    /// POP tokens are signed by the process making the request. By default, MSAL will generate a key in memory.
+    /// To use a hardware key or an external key, implement <see cref="PopCryptoProvider"/>.
+    /// </remarks>
 #if !SUPPORTS_CONFIDENTIAL_CLIENT
     [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
 #endif
     public class PoPAuthenticationConfiguration
     {
         /// <summary>
-        /// Constructs the configuration properties used to construct a proof of possession request
+        /// Creates a configuration using the default key management. Use the properties <see cref="HttpMethod"/>, <see cref="HttpHost"/> etc. 
+        /// to control which elements of the request should be included in the POP token.
         /// </summary>
-        /// <param name="requestUri">An Uri to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request uri</param>
-        public PoPAuthenticationConfiguration(Uri requestUri)
+        /// <remarks>
+        /// See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#page-3 for details about signed HTTP requests.
+        /// </remarks>
+        public PoPAuthenticationConfiguration()
         {
             ConfidentialClientApplication.GuardMobileFrameworks();
-            RequestUri = requestUri ?? throw new ArgumentNullException(nameof(requestUri));
         }
 
         /// <summary>
-        /// An Uri to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request uri (to the host and query parts of it)
+        /// Creates a configuration using the default key management, and which binds all the details of the HttpRequestMessage.
         /// </summary>
-        public Uri RequestUri { get; }
+        /// <remarks>
+        /// Currently only the HtppMethod (m), UrlHost (u) and UrlPath (p) are used to create the signed HTTP request - see https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#page-3
+        /// </remarks>
+        public PoPAuthenticationConfiguration(HttpRequestMessage httpRequestMessage)
+        {
+            if (httpRequestMessage == null)
+                throw new ArgumentNullException(nameof(httpRequestMessage));
+
+            HttpMethod = httpRequestMessage.Method;
+            HttpHost = httpRequestMessage.RequestUri.Authority; // this includes the optional port
+            HttpPath = httpRequestMessage.RequestUri.AbsolutePath; 
+        }
+
+      
+
+        /// <summary>
+        /// Creates a configuration using the default key management, and which binds only the Uri part of the HTTP request.
+        /// </summary>
+        /// <remarks>
+        /// The UrlHost (u) and UrlPath (p) are used to create the signed HTTP request - see https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#page-3
+        /// </remarks>
+        public PoPAuthenticationConfiguration(Uri requestUri)
+        {
+            if (requestUri == null)
+                throw new ArgumentNullException(nameof(requestUri));
+
+            HttpHost = requestUri.Authority;
+            HttpPath = requestUri.AbsolutePath;
+        }
 
         /// <summary>
         /// The HTTP method ("GET", "POST" etc.) method that will be bound to the token. Leave null and the POP token will not be bound to the method.
+        /// Corresponds to the "m" part of the a signed HTTP request. Optional.
         /// </summary>
+        /// <remarks>
+        /// See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#section-3
+        /// </remarks>
         public HttpMethod HttpMethod { get; set; }
+
+        /// <summary>
+        /// The URL host of the protected API. The "u" part of a signed HTTP request.  This MAY include the port separated from the host by a colon in host:port format. Optional.
+        /// </summary>
+        public string HttpHost { get; set; }
+
+        /// <summary>
+        /// The "p" part of the signed HTTP request. 
+        /// </summary>
+        public string HttpPath { get; set; }
 
         /// <summary>
         /// An extensibility point that allows developers to define their own key management. 
@@ -44,5 +91,11 @@ namespace Microsoft.Identity.Client.AppConfig
         /// as MSAL.NET will keep a thumbprint of keys in memory.
         /// </summary>
         public IPoPCryptoProvider PopCryptoProvider { get; set; }
+
+        /// <summary>
+        /// If the protected resource (RP) requires use of a special nonce, they will publish it as part of the WWWAuthenticate header associated with a 401 HTTP response
+        /// or as part of the AuthorityInfo header associated with 200 response. Set it here to make it part of the Signed HTTP Request part of the POP token.
+        /// </summary>
+        public string Nonce { get; set; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
@@ -48,8 +48,6 @@ namespace Microsoft.Identity.Client.AppConfig
             HttpPath = httpRequestMessage.RequestUri.AbsolutePath; 
         }
 
-      
-
         /// <summary>
         /// Creates a configuration using the default key management, and which binds only the Uri part of the HTTP request.
         /// </summary>
--- a/src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
@@ -65,31 +65,51 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
 
         public string FormatAccessToken(MsalAccessTokenCacheItem msalAccessTokenCacheItem)
         {
-            var header = new JObject
+            JObject header = new JObject
             {
                 { JsonWebTokenConstants.ReservedHeaderParameters.Algorithm, _popAuthenticationConfiguration.PopCryptoProvider.CryptographicAlgorithm },
                 { JsonWebTokenConstants.ReservedHeaderParameters.KeyId, KeyId },
                 { JsonWebTokenConstants.ReservedHeaderParameters.Type, PoPRequestParameters.PoPTokenType}
             };
 
-            JToken popAssertion = JToken.Parse(_popAuthenticationConfiguration.PopCryptoProvider.CannonicalPublicKeyJwk);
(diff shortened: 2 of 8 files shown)


src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -3,46 +3,99 @@
...
+
+      
+
+        /// <summary>
+        /// Creates a configuration using the default key management, and which binds only the Uri part of the HTTP request.
+        /// </summary>
+        /// <remarks>
+        /// The UrlHost (u) and UrlPath (p) are used to create the signed HTTP request - see https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#page-3
+        /// </remarks>
+        public PoPAuthenticationConfiguration(Uri requestUri)
```
bgavrilMS, line 59: I think we should remove this API. The spec states that a POP request is valid even with NO Http Signed request claims. We just don't have E2E infrastructure to test it. I'll leave it for now and confirm with middleware team.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -3,46 +3,99 @@
...
-        /// Constructs the configuration properties used to construct a proof of possession request
+        /// Creates a configuration using the default key management. Use the properties <see cref="HttpMethod"/>, <see cref="HttpHost"/> etc. 
+        /// to control which elements of the request should be included in the POP token.
         /// </summary>
-        /// <param name="requestUri">An Uri to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request uri</param>
-        public PoPAuthenticationConfiguration(Uri requestUri)
+        /// <remarks>
+        /// See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#page-3 for details about signed HTTP requests.
+        /// </remarks>
+        public PoPAuthenticationConfiguration()
```
bgavrilMS, line 30: new ctor ... for when you don't want to bind to the HTTP request

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -3,46 +3,99 @@
...
 
         /// <summary>
-        /// An Uri to the protected resource which requires a PoP token. The PoP token will be cryptographically bound to the request uri (to the host and query parts of it)
+        /// Creates a configuration using the default key management, and which binds all the details of the HttpRequestMessage.
         /// </summary>
-        public Uri RequestUri { get; }
+        /// <remarks>
+        /// Currently only the HtppMethod (m), UrlHost (u) and UrlPath (p) are used to create the signed HTTP request - see https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#page-3
+        /// </remarks>
+        public PoPAuthenticationConfiguration(HttpRequestMessage httpRequestMessage)
```
bgavrilMS, line 41: new ctor ... for full binding.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -3,46 +3,99 @@
...
         /// </summary>
+        /// <remarks>
+        /// See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#section-3
+        /// </remarks>
         public HttpMethod HttpMethod { get; set; }
 
+        /// <summary>
+        /// The URL host of the protected API. The "u" part of a signed HTTP request.  This MAY include the port separated from the host by a colon in host:port format. Optional.
+        /// </summary>
+        public string HttpHost { get; set; }
```
bgavrilMS, line 80: Public properties to allow full customization.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -3,46 +3,99 @@
...
+    /// Details about the HTTP request and configuration properties used to construct a proof of possession request.
     /// </summary>
+    /// <remarks> 
+    /// POP tokens are signed by the process making the request. By default, MSAL will generate a key in memory.
+    /// To use a hardware key or an external key, implement <see cref="PopCryptoProvider"/>.
+    /// </remarks>
 #if !SUPPORTS_CONFIDENTIAL_CLIENT
     [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
 #endif
     public class PoPAuthenticationConfiguration
```
bgavrilMS, line 21: @jmprieur to review public API changes.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -3,46 +3,99 @@
...
+        }
 
         /// <summary>
         /// The HTTP method ("GET", "POST" etc.) method that will be bound to the token. Leave null and the POP token will not be bound to the method.
+        /// Corresponds to the "m" part of the a signed HTTP request. Optional.
         /// </summary>
+        /// <remarks>
+        /// See https://datatracker.ietf.org/doc/html/draft-ietf-oauth-signed-http-request-03#section-3
+        /// </remarks>
         public HttpMethod HttpMethod { get; set; }
```
jmprieur, line 75: Would we want to have Get as a default method?

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
jmprieur: We might also want to fix https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2285 at the same time?

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
jmprieur: deprecate it?

jmprieur (approved):
LGTM for the public API.
Thanks @bgavrilMS

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
SameerK-MSFT: Thought: Should we make set protected? To prevent any accidental update and security holes. With protected, developer will have to explicitly inherit from the the class and set it.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
bgavrilMS: API is "experimental", so we can delete it.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
bgavrilMS: This is more of a data-object type. We haven't used inheritance as extensibility mechanism in MSAL so far. 

From a security perspective, protected is exactly the same as public. See https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/protected-members

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
bgavrilMS: How about deprecate for 1 release, then delete? API is experimental, we can delete, but I'm mindful of not breaking MIW.

jmprieur: M.I.W can react quickly. right this is experimental..
I agree with the approach

bgavrilMS: Will leave for now, let's see how this evolves.

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
bgavrilMS: This class will evolve to have all HTTP request components: q (query), h (headers), b (body). All of these claims are optional, so it seems a bit unnatural to have m (method) take a default value.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
```diff
@@ -88,18 +79,17 @@ public async Task PoP_BearerAndPoP_CanCoexist_Async()
...
 
         private async Task BearerAndPoP_CanCoexist_Async()
         {
             // Arrange
-            var popConfig = new PoPAuthenticationConfiguration(new Uri("https://www.contoso.com/path1/path2?queryParam1=a&queryParam2=b"));
+            var popConfig = new PoPAuthenticationConfiguration(new Uri(ProtectedUrl));
             popConfig.HttpMethod = HttpMethod.Get;
 
-            //SecureString securePassword = new NetworkCredential("", user.GetOrFetchPassword()).SecurePassword;
             string clientId = PublicCloudConfidentialClientID;
```
trwalke, line 93: Variable can be deleted

src/client/Microsoft.Identity.Client/AppConfig/PopAuthenticationConfiguration.cs
```diff
@@ -3,46 +3,97 @@
...
         /// Important note: if you want to change the key (e.g. rotate the key), you should create a new instance of this object,
         /// as MSAL.NET will keep a thumbprint of keys in memory.
         /// </summary>
         public IPoPCryptoProvider PopCryptoProvider { get; set; }
+
+        /// <summary>
+        /// If the protected resource (RP) requires use of a special nonce, they will publish it as part of the WWWAuthenticate header associated with a 401 HTTP response
+        /// or as part of the AuthorityInfo header associated with 200 response. Set it here to make it part of the Signed HTTP Request part of the POP token.
+        /// </summary>
+        public string Nonce { get; set; }
```
trwalke, line 97: One of the additional goals was to enable MSAL to parse the nonce value from the HttpResponse for the customer. Have you tried this out with this? Perhaps we can mentio this class in the documentation/comments? [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ee3cd9c2fe59c0c34ff3a65142bc16f017fd7b27/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs#L105](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ee3cd9c2fe59c0c34ff3a65142bc16f017fd7b27/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs#L105)

bgavrilMS: It does not, the format of the headers is not well understood. Tracking issue: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2844

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -65,31 +65,51 @@ public PoPAuthenticationScheme(PoPAuthenticationConfiguration popAuthenticationC
...
-                    new JProperty(PoPClaimTypes.JWK, popAssertion))));
+            string popToken = CreateJWS(body.ToString(Formatting.None), header.ToString(Formatting.None));
+            return popToken;
+        }
 
-            string popToken =  CreateJWS(payload.ToString(Json.Formatting.None), header.ToString(Json.Formatting.None));
+        private JObject CreateBody(MsalAccessTokenCacheItem msalAccessTokenCacheItem)
+        {
+            JToken publicKeyJWK = JToken.Parse(_popAuthenticationConfiguration.PopCryptoProvider.CannonicalPublicKeyJwk);
+            List<JProperty> properties = new List<JProperty>(8);
```
trwalke, line 84: nit: Is there a reason this is set to a capacity of 8? Seems there aren't 8 properties atm and some may be null anyway. Not really a big deal though.

tests/Microsoft.Identity.Test.Unit/pop/PoPTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+
+            request = new HttpRequestMessage(HttpMethod.Post, new Uri(ProtectedUrlWithPort));
+            popConfig = new PoPAuthenticationConfiguration(request);
+
+            Assert.AreEqual(HttpMethod.Post, popConfig.HttpMethod);
+            Assert.AreEqual("www.contoso.com:5555", popConfig.HttpHost);
+            Assert.AreEqual("/path1/path2", popConfig.HttpPath);
+        }
+        private static void AssertSingedHttpRequestClaims(PoPAuthenticationConfiguration popConfig, System.Security.Claims.ClaimsPrincipal claims)
```
trwalke, line 152: nit: add new line

trwalke (approved):
Minor comments but LGTM overall

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
bgavrilMS: It's a good question!

These higher level collections like Lists are usually backed up by arrays (List impl [here](https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,cf7f4095e4de7646)). Array is of a fixed size, so it needs to be reallocated most likely using `Array.Copy`.

By default, a collections starts at capacity 0 (well, it dependes on .NET runtime, might be 4, but let's say 0). After you add 1 elements, it gets resized to 2. And then to 4, 8, 16 etc. Since we know the collection will have ~8 elements, I just set the capacity to 8 to avoid resizing. If we end up adding more than 8, we can set a more relevant capacity.

Issue #2796, opened 2021-07-27
[Feature Request] Add ability to add probabilistic jitter to expiry and refresh times in cache


rymeskar:
**Is your feature request related to a problem? Please describe.**
Right now, when concurrent requests arrive to fetch an expired token, they might both end up making an HTTP call to eSTS. Such calls are redundant and cause extra CPU and network cycles on both eSTS and the service.

**Describe the solution you'd like**
A probabilistic way to decrease the amount of such redundant calls to eSTS is to introduce randomization (jittering) to calculation of the times when a token needs to be refreshed. This randomization might effectively decrease either the 'refreshOn' or 'expiry' by a few minutes. 

**Describe alternatives you've considered**

Employing more locks inside the codebase.

**Additional context**

This feature is even more important when dealing with distributed caches as the overlapping concurrency window is bigger and the amount of service instances running the race is also bigger compared to amount of threads.

bgavrilMS:
refresh_in is associated with longer lived tokens, at least >2h but in practice 12-24h. With that in mind, would this work ?

- read refresh_in value from ESTS
- randomly generate a number between -N and N 
- update refresh_in = (refresh_in + N)

Questions: 

- value of N ? 5 min?
- should we do this for expires_in as well? 

I'm thinking that we do not expose a public API for refresh_in jittering, we just do it internally.

rymeskar:
I think that for the refresh_in, the internal proposal with +-5 mins makes complete sense. We could theoretically do even some statistical modeling if we had the eSTS latency numbers. 

For the expires_in, do you have some more context on the throw-away 5 mins before the actual expiry? For 90+% of cases, this effectively leads to 10 mins before the token stops being accepted at the server side (the Wilson validation lib has by default 5 mins skew).

pmaytak:
Included in [MSAL.NET 4.36.0 release](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.36.0).
cc: @rymeskar

rymeskar:
Thanks a lot for closing the issue so fast.

I noticed that in the end the jitter has been added to only refresh_in and not expiry. I would then like to follow up on the previous conversation:
"For the expires_in, do you have some more context on the throw-away 5 mins before the actual expiry? For majority of C# cases, this effectively leads to 10 mins before the token stops being accepted at the server side (the Wilson validation lib has by default 5 mins skew)."

cc: @bgavrilMS

[...]

PR #2842, opened 2021-08-24
Adding jitter to accessToken expirations

trwalke:
Fixes # .
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2796

Adding jitter with max value of 5 mins to AT expiration/refresh_in

**Testing**
Unit test

**Performance impact**
Evaluating

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
@@ -74,6 +74,8 @@ namespace Microsoft.Identity.Client.Cache.Items
             }
 
             HomeAccountId = homeAccountId;
+
+            AddJitterToTokenExpiration();
         }        
 
         private string _tenantId;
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -74,6 +74,8 @@ internal class MsalAccessTokenCacheItem : MsalCredentialCacheItemBase
             }
 
             HomeAccountId = homeAccountId;
+
+            AddJitterToTokenExpiration();
```
trwalke, line 78: @bgavrilMS The expiration is used to calculate the suggestedCacheExpiry which happens when the token is saved to the cache. If the jitter is added when the token is retrieved from the cache the suggestedCacheExpiry will no longer be accurate, although the difference will be within 5 mins. Thus, I added the jitter modification in this constructor.

Thoughts?

bgavrilMS:
@trwalke  - the work item tells you that you should change the RefreshIn value by a few minutes, not Expires In.

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
bgavrilMS: 1. Let's only modify RefreshIn
2. MsalAccessTokenCacheItem object is a data object. It should not know about jitter and other concepts like this. By adding jitter here, it's easy to make a mistake like - add jitter on token response, then later load the cache -> add jitter again.

Probably the simplest approach here is to change the refresh_in value from MSAL token response.

trwalke:
> @trwalke - the work item tells you that you should change the RefreshIn value by a few minutes, not Expires In.

The work item isnt 100% clear but I will remove the expires in

bgavrilMS:
@trwalke - let's have a look at this toghter

PR #2843, opened 2021-08-25
Discovery fails throws ex including HttpResponse

### LINKED ISSUE #2835: [Bug] user_realm_discovery_failed exception does not provide reason for failure
4.35.1

* Desktop / Mobile
    * [ ] Interactive
    * [x] Integrated Windows Authentication
    * [x] Username Password
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 

Other?

**Is this a new or existing app?** Existing

**Expected behavior**
HTTP errors while discovering the user realm are not exposed to the user. MSAL should use the standard error handling mechanism here (i.e. MsalServiceExceptionFactory.FromHttpResponse)

**Actual behavior**
Error root cause is hidden and a generic "user_realm_discovery_failed" exception is presented.

**Possible solution**
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b406142904cb0ccb8bb3aeb236db42754ba29c60/src/client/Microsoft.Identity.Client/WsTrust/CommonNonInteractiveHandler.cs#L52

SameerK-MSFT:
Fixes #2835  .

When discover fails, it throws MsalServiceException with Http response message and status code.

**Testing**
Added unit test

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/WsTrust/WsTrustWebRequestManager.cs
+++ b/src/client/Microsoft.Identity.Client/WsTrust/WsTrustWebRequestManager.cs
@@ -142,9 +142,19 @@ namespace Microsoft.Identity.Client.WsTrust
                 msalIdParams,
                 requestContext.Logger).ConfigureAwait(false);
 
-            return httpResponse.StatusCode == System.Net.HttpStatusCode.OK
-                ? JsonHelper.DeserializeFromJson<UserRealmDiscoveryResponse>(httpResponse.Body)
-                : null;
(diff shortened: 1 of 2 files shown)


jmprieur:
Is it the thing to do?
Don't we fallback on hard-coded Metadata?

bgavrilMS:
LGTM, I would also change `QueryUserRealmDataAsync` method to not expect a null response.

bgavrilMS (approved):

bgavrilMS:
@jmprieur - this is about User Realm Discovery in IWA. Too many discoveries ...

PR #2847, opened 2021-08-25
InjectAppToken prototype

### LINKED ISSUE #2806: PROTOTYPE Enable Managed Service Identity tokens obtained by Azure.Identity via IMDS to store be stored in the MSAL Cache
Issue/Feature:
Azure.Identity obtains MSI tokens via the IMDS endpoint today. In order to leverage the same caching across all libraries it may be benefitial for MSAL to allow Azure.Identity to cache MSI tokens using the MSAL cache. An alternative is to move the logic of obtaining the MSI token to MSAL.

Suggestion:
Provide an API for ConfidentialClientApplications which can take an AT and persist it in the cache. The caller will need to acquireToken silently before trying to get a token them selves and they will also need to call the persist AT API when they get a new token.

Alternative:
Move MSI token acquisition via the IMDS endpoint to MSAL and let MSAL handle this scenario.

bgavrilMS:
Fixes #2806 

This a prototype to see MSAL could cache IMDS tokens.

We could tackle this in 2 ways: 

a) `cca.InjectAppTokenAsync(string[] scopes, long expiresInSeconds, string rawAccessToken)` (this is implemented in the prototype)

b) `cca.AcquireTokenFromAzureImds(string[] scopes)` - i.e. we move the IMDS code from Azure SDK to MSAL. We still need to figure out how to cache the token though.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/SuggestedWebCacheKeyFactory.cs
@@ -54,7 +54,8 @@ namespace Microsoft.Identity.Client.Cache
                 return true;
             }
 
-            if (requestParameters.ApiId == TelemetryCore.Internal.Events.ApiEvent.ApiIds.AcquireTokenForClient)
+            if (requestParameters.ApiId == TelemetryCore.Internal.Events.ApiEvent.ApiIds.AcquireTokenForClient ||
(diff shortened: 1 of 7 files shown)


bgavrilMS:
No updates, closing this.

jmprieur:
The conversation is still going on, @bgavrilMS 
@schaabs is it something we can make work? (even if MSAL can't do the refresh)

cc: @henrik-me

Issue #2852, opened 2021-08-27
pass-in an App ID URI as a client ID, as this is supported by AAD


deep260996:
**Is your feature request related to a problem? Please describe.**
For ADAL to MSAL migration I want to use AppID URI as client ID to generate tokens for my app. The problem is discussed in detail in this Teams channel post. https://teams.microsoft.com/l/message/19:3f112ecb27024f2682379a7af3667869@thread.tacv2/1628538375509?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=7e394638-cacd-40c7-9e80-c4434c4c90df&parentMessageId=1628538375509&teamName=.NET%20Migration%20(ADAL.NET%20to%20MSAL.NET%20and%20SAL%2FMISE)&channelName=ADAL.NET%20to%20MSAL.NET%20Migration&createdTime=1628538375509 

**Describe the solution you'd like**
Allow App ID URI to be passed-in as client ID

**Additional context**
I am from the LSM team and this is blocking us on the ADAL to MSAL migration.

PR #2862, opened 2021-09-02
Enabling App Id Uri to be used with web apps

trwalke:
Fixes # .
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2852

Enabling App Id Uri to be used with web apps

**Testing**
Unit/Integration

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -353,6 +353,12 @@ namespace Microsoft.Identity.Client
         {
             base.Validate();
 
+            //ADFS does not require client id to be in the form of a GUID.
+            if (Config.AuthorityInfo?.AuthorityType != AuthorityType.Adfs && !Guid.TryParse(Config.ClientId, out _))
+            {
+                throw new MsalClientException(MsalError.ClientIdMustBeAGuid, MsalErrorMessage.ClientIdMustBeAGuid);
+            }
+
             if (string.IsNullOrWhiteSpace(Config.RedirectUri))
             {
                 Config.RedirectUri = PlatformProxyFactory.CreatePlatformProxy(null)
--- a/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
@@ -470,12 +470,6 @@ namespace Microsoft.Identity.Client
                 throw new MsalClientException(MsalError.NoClientId, MsalErrorMessage.NoClientIdWasSpecified);
             }         
 
-            //ADFS does not require client id to be in the form of a GUID.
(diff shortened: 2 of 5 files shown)


jmprieur (approved):
LGTM
Thanks @trwalke

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -353,6 +353,12 @@ internal override void Validate()
         {
             base.Validate();
 
+            //ADFS does not require client id to be in the form of a GUID.
+            if (Config.AuthorityInfo?.AuthorityType != AuthorityType.Adfs && !Guid.TryParse(Config.ClientId, out _))
+            {
+                throw new MsalClientException(MsalError.ClientIdMustBeAGuid, MsalErrorMessage.ClientIdMustBeAGuid);
+            }
```
rayluo, line 360: Overall this PR looks good. :+1: 

This implementation removes the GUID check from `ConfidentialClientApplication`, but keeps it for `PublicClientApplication`. Is there a specific reason to keep it, or is it just us wanting to make minimal change in this PR (which is understandable, too)? Basically, curious to know the answers to [@jmprieur 's earlier questions](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2852#issuecomment-907421378).

pmaytak (approved):

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
bgavrilMS: I think there's no way to configure an App Id Uri for public client. That said, today you can use the same client id for both PCA and CCA. 

@trwalke - maybe we should remove this validation altoghter?

bgavrilMS (approved):
LGTM, but I'd remove the validation altogheter.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
jennyf19: only web APIs can have an App Id URI. Public clients only have the client_id. This check is still valid for public clients.

trwalke: Yes but you can still use the app Id Uri with for public clients and the portal does not do a good job of explaining why you should not do this. Using an app Id Uri will result in a malformed request exception from AAD so we might as well fail faster with this validate method for public client apps.

pmaytak: Yea, it's not super clear because they do show the App ID URI in the App Registration Overview blade. But when you click it, it takes you to the Expose an API blade, and there they specify, in a tooltip, that it's only for a web API.
image
image

trwalke: @rayluo There are no issues with that I have come across using a non guid for the client ID. All of our tests are passing and the caching operations are working well. Also, I got confirmation from each of the SDKs in which we cache compatibility to confirm that the app id uri works in their SDK today.

PR #2874, opened 2021-09-07
Add static cache options

### LINKED ISSUE #2849: [Feature Request] In-memory shared cache cache in MSAL (like ADAL), with good performance
Many customers have problems migrating from ADAL to MSAL because of the token caching. The cache is no longer static / shared. 

Our current solutions to this problem are not great because:

- Microsoft.Identity.Web has many dependencies, making it difficult to adopt due to diamond dependency hell. 
- implemeting cache serialization using the callbacks is easy to get wrong and isn't even the fully documented anymore (Microsoft.Identity.Web being the recommended solution).
- in-memory caching today relies on JSON serialization, which is a bad solution from a perf perspective. As @pmaytak documented [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2834#issue-716518042) - the json serialization operation is about 1000x slower than using in-memory representations only. This is not noticeable for a small number of tokens per tenant, but we cannot guarantee this setup.

Proposal: 

Add a feature to allow app developers some control over how tokens are stored in memory, directly from MSAL. This can be added to both PCA and CCA. Should not be exposed on mobile platforms and UWP.

## Solution 1

`appBuilder.WithStaticMemoryCache()` // this will make the internal storage static / shared and will achieve ADAL parity

This simply involves adding static version of our Accessor objects, which are already thread safe.

## Solution 2 (preferred)

A more general approach, allowing us to move to use Wilson's Event based LRU cache which can add eviction policies. 

`appBuilder.WithMemoryCache(MemoryCacheOptions)`

```
MemoryCacheOptions // not all options need to be implemented at once
+ MaxNumberOfItems: int // P2
+ Expiration: DateTime // P2
+ Static : bool // P1
```
## Solution 3

Full extensibility. We would still need to provide a few implementations.

`appBuilder.WithMemoryCache(ITokenCacheAccessor)` where ITokenCacheAccessor is our (currently internal) interface https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/cfa8c9193e0797eeca1927729b980acd1dd18b06/src/client/Microsoft.Identity.Client/Cache/ITokenCacheAccessor.cs

bgavrilMS:
Fixes #2849

**Testing**
ut + perf

**Performance impact** No difference

### With static cache

|                                    Method | Tenants | TokensPerTenant |     Mean |     Error |    StdDev |    Median |       Min |      Max | Allocated |
|------------------------------------------ |-------- |---------------- |---------:|----------:|----------:|----------:|----------:|---------:|----------:|
| AcquireTokenForClient_Singleton_TestAsync |    1000 |            1000 | 1.029 ms | 0.0878 ms | 0.2546 ms | 0.9091 ms | 0.7325 ms | 1.746 ms |    185 KB |

### Baseline

|                          Method | Tenants | TokensPerTenant |     Mean |    Error |   StdDev |   Median |      Min |        Max | Allocated |
|-------------------------------- |-------- |---------------- |---------:|---------:|---------:|---------:|---------:|-----------:|----------:|
| AcquireTokenForClient_TestAsync |    1000 |            1000 | 956.9 us | 76.79 us | 215.3 us | 865.7 us | 661.6 us | 1,626.7 us |    185 KB |

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
-using System.Globalization;
 using System.Linq;
 using Microsoft.Identity.Client.Http;
 using Microsoft.Identity.Client.Instance;
@@ -71,7 +70,7 @@ namespace Microsoft.Identity.Client
             {
                 throw new ArgumentNullException(instanceDiscoveryJson);
             }
-            
+
             try
             {
                 InstanceDiscoveryResponse instanceDiscovery = JsonHelper.DeserializeFromJson<InstanceDiscoveryResponse>(instanceDiscoveryJson);
@@ -176,6 +175,33 @@ namespace Microsoft.Identity.Client
         {
             Config.HttpManager = httpManager;
             return (T)this;
+        }
+
+        /// <summary>
+        /// MSAL maintains a token cache internally in memory. By default, this cache object is part of each instance of <see cref="PublicClientApplication"/> or <see cref="ConfidentialClientApplication"/>.
+        /// This method allows customization of the in-memory token cache of MSAL. 
+        /// 
+        /// MSAL's memory cache is different than token cache serialization. Cache serialization pulls the tokens from a cache (e.g. Redis, Cosmos, or a file on disk), 
+        /// where they are stored in JSON format, into MSAL's internal memory cache. Memory cache operations do not involve JSON operations. 
+        /// 
+        /// External cache serialization remains the recommended way to handle desktop apps, web site and web APIs, as it provides persistence.
+        /// 
+        /// Detailed guidance for each application type and platform:
+        /// https://aka.ms/msal-net-token-cache-serialization
+        /// </summary>
+        /// <param name="options">Options for the internal MSAL token caches. </param>
+#if !SUPPORTS_CUSTOM_CACHE || WINDOWS_APP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+        public T WithInternalMemoryTokenCacheOptions(InternalMemoryTokenCacheOptions options)
+        {
+#if !SUPPORTS_CUSTOM_CACHE || WINDOWS_APP
+            throw new PlatformNotSupportedException("WithInteralTokenCacheOptions is supported only on platforms where MSAL stores tokens in memory and not on mobile platforms or UWP.");
+#else
+
+            Config.AccessorOptions = options;
+            return (T)this;
+#endif
         }
 
         internal T WithPlatformProxy(IPlatformProxy platformProxy)
@@ -467,12 +493,12 @@ namespace Microsoft.Identity.Client
             if (string.IsNullOrWhiteSpace(Config.ClientId))
             {
                 throw new MsalClientException(MsalError.NoClientId, MsalErrorMessage.NoClientIdWasSpecified);
-            }         
+            }
 
             if (Config.CustomInstanceDiscoveryMetadata != null && Config.CustomInstanceDiscoveryMetadataUri != null)
             {
                 throw new MsalClientException(
-                    MsalError.CustomMetadataInstanceOrUri, 
+                    MsalError.CustomMetadataInstanceOrUri,
                     MsalErrorMessage.CustomMetadataInstanceOrUri);
             }
 
@@ -505,11 +531,11 @@ namespace Microsoft.Identity.Client
                     Config.AuthorityInfo.AuthorityType == AuthorityType.Aad)
                 {
                     AadAuthority aadAuthority = Authority.CreateAuthority(Config.AuthorityInfo) as AadAuthority;
-                    if (!aadAuthority.IsCommonOrganizationsOrConsumersTenant() && 
+                    if (!aadAuthority.IsCommonOrganizationsOrConsumersTenant() &&
                         !string.Equals(aadAuthority.TenantId, Config.TenantId))
                     {
                         throw new MsalClientException(
-                            MsalError.AuthorityTenantSpecifiedTwice, 
+                            MsalError.AuthorityTenantSpecifiedTwice,
                             "You specified a different tenant - once in WithAuthority() and once using WithTenant().");
                     }
 
@@ -762,7 +788,7 @@ namespace Microsoft.Identity.Client
             Config.AadAuthorityAudience = authorityAudience;
             Config.ValidateAuthority = validateAuthority;
             return (T)this;
-        }       
+        }
 
         /// <summary>
         /// Adds a known Authority corresponding to an ADFS server. See https://aka.ms/msal-net-adfs
--- a/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
@@ -68,6 +68,8 @@ namespace Microsoft.Identity.Client
 
         public IPlatformProxy PlatformProxy { get; internal set; }
 
+        public InternalMemoryTokenCacheOptions AccessorOptions { get; set; }
+
         public AuthorityInfo AuthorityInfo { get; internal set; }
         public string ClientId { get; internal set; }
         public string TenantId { get; internal set; }
--- a/src/client/Microsoft.Identity.Client/AppConfig/InternalMemoryTokenCacheOptions.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/InternalMemoryTokenCacheOptions.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Options for the internal in-memory MSAL token caches.
+    /// </summary>
+    /// <remarks>
+    /// Recommended only for client credentials flow (service to service communication).
+    /// Web apps and Web APIs should use external token caching (Redis, Cosmos etc.) for scaling purposes.
+    /// Desktop apps should encrypt and persist their token cache to disk, to avoid losing tokens when app restarts. 
+    /// For detailed recommendations see: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore
+    /// </remarks>
+    public class InternalMemoryTokenCacheOptions
+    {
+        /// <summary>
+        /// Constructor for the options with default values.
+        /// </summary>
+        public InternalMemoryTokenCacheOptions()
+        {
+        }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="useSharedCache">Set to true to share the cache between all ClientApplication objects. The cache becomes static. </param>
+        public InternalMemoryTokenCacheOptions(bool useSharedCache)
+        {
+            UseSharedCache = useSharedCache;
+        }
+
+        /// <summary>
+        /// Share the cache between all ClientApplication objects. The cache becomes static. Defaults to false.
+        /// </summary>
+        /// <remarks>ADAL used a static cache by default</remarks>
+        public bool UseSharedCache { get; set; }
+
+    }
+}
--- a/src/client/Microsoft.Identity.Client/MsalError.cs
+++ b/src/client/Microsoft.Identity.Client/MsalError.cs
@@ -550,7 +550,7 @@ namespace Microsoft.Identity.Client
         /// <para>What happens?</para>You did not define an authority at the application level, so it defaults to the public cloud (login.microsoft.com), but the authority at the request level is for a different cloud. Only the tenant can be changed at the request level.
         /// <para>Mitigation</para>Add .WithAuthority("https://login.windows-ppe.net/common) at the application level and specify the tenant at the request level: .WithAuthority("https://login.windows-ppe.net/1234-567-890-12345678). 
         /// </summary>
-        public const string AuthorityHostMismatch = "authority_host_mismatch";       
+        public const string AuthorityHostMismatch = "authority_host_mismatch";
 
         /// <summary>
         /// Duplicate query parameter was found in extraQueryParameters.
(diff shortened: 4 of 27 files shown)


src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -178,6 +180,37 @@ internal T WithHttpManager(IHttpManager httpManager)
...
+        /// 
+        /// Detailed guidance for each application type and platform:
+        /// https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore
+        /// </summary>
+        /// <param name="useStaticCache">Share the static cache between all ClientApplication objects. Defaults to false.</param>
+        /// <remarks>The legacy ADAL library used a static cache by default.</remarks>
+#if !SUPPORTS_CUSTOM_CACHE || WINDOWS_APP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+        public T WithMemoryTokenCacheOptions(bool useStaticCache)
```
bgavrilMS, line 201: New API. 

CC @jmprieur

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -178,6 +180,37 @@ internal T WithHttpManager(IHttpManager httpManager)
...
+        {
+            // TODO: expose on options
```
bgavrilMS, line 203: Is this important enough to add to options? I think it is, can action this tomorrow.

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -1004,5 +1004,12 @@ public static class MsalError
         /// <para>Mitigation</para>For troubleshooting details, see https://aka.ms/msal-net-wam .
         /// </summary>
         public const string InitializeProcessSecurityError = "initialize_process_security_error";
+
+        /// <summary>
+        /// <para>What happens?</para>You configured MSAL cache serialization at the same time with <see cref="AbstractApplicationBuilder{T}.WithMemoryTokenCacheOptions(bool)"/>
+        /// These are mutually exclusive.
+        /// <para>Mitigation</para> Use only one option. Web site and web api scenarios should rely on external cache serialization, as internal cache serialization cannot scale. See https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore        
+        /// </summary>
+        public const string ExternalInternalCacheSerialization = "externa_internal_cache_serialization";
```
jennyf19, line 1013: >externa [](http://example.com/codeflow?start=66&length=7)

sp: `external`...the name is a little confusing. Also maybe use an aka.ms link for the above.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
jennyf19, line 201: >MemoryTokenCacheOptions [](http://example.com/codeflow?start=21&length=23)

will there be more options in the future? Something like[`MemoryCacheEntryOptions`](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-5.0#memorycacheentryoptions)

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -393,6 +393,11 @@ public static string CertMustHavePrivateKey(string certificateName)
                 "Failure setting process security to enable WAM Account Picker in an elevated process ({0}). For troubleshooting details, see https://aka.ms/msal-net-wam .",
                 errorCode);
 
-        public static string CcsRoutingHintMissing = "Either the userObjectIdentifier or tenantIdenifier are missing. Both are needed to create the ccs routing hint. See https://aka.ms/msal-net/ccsRouting. ";
+        public const string CcsRoutingHintMissing = "Either the userObjectIdentifier or tenantIdenifier are missing. Both are needed to create the ccs routing hint. See https://aka.ms/msal-net/ccsRouting. ";
+
+        public const string ExternalInternalCacheSerialization = 
+            "You configured MSAL cache serialization at the same time with internal caching options. These are mutually exclusive. " +
```
jmprieur, line 399:
```suggestion
            "You configured MSAL token cache serialization at the same time with internal caching options. These are mutually exclusive. " +

```

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -393,6 +393,11 @@ public static string CertMustHavePrivateKey(string certificateName)
                 "Failure setting process security to enable WAM Account Picker in an elevated process ({0}). For troubleshooting details, see https://aka.ms/msal-net-wam .",
                 errorCode);
 
-        public static string CcsRoutingHintMissing = "Either the userObjectIdentifier or tenantIdenifier are missing. Both are needed to create the ccs routing hint. See https://aka.ms/msal-net/ccsRouting. ";
+        public const string CcsRoutingHintMissing = "Either the userObjectIdentifier or tenantIdenifier are missing. Both are needed to create the ccs routing hint. See https://aka.ms/msal-net/ccsRouting. ";
+
+        public const string ExternalInternalCacheSerialization = 
+            "You configured MSAL cache serialization at the same time with internal caching options. These are mutually exclusive. " +
+            "Use only one option. Web site and web api scenarios should rely on external cache serialization, as internal cache serialization cannot scale. " +
+            "See https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore  ";
```
jmprieur, line 401:
```suggestion
            "See https://aka.ms/msal-net-cca-token-cache-serialization";

```

jmprieur (commented):
I've provided suggestions for XML comments and naming.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: That is a LOOONG name :). Suggestions for something smaller?

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -178,6 +180,34 @@ internal T WithHttpManager(IHttpManager httpManager)
...
+        /// External cache serialization remains the recommended way to handle desktop apps, web site and web apis, as it provides persistence.
+        /// 
+        /// Detailed guidance for each application type and platform:
+        /// https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore
+        /// </summary>
+        /// <param name="options">Options for the internal MSAL token caches. </param>
+#if !SUPPORTS_CUSTOM_CACHE || WINDOWS_APP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+        public T WithMemoryTokenCacheOptions(MemoryTokenCacheOptions options)
```
jmprieur, line 200: I still think that we should make it clear that this is internal and non-serialized memory.
There is an existent, and we want the message to be super clear.

I'm still proposing:

```CSharp
WithInternalMemoryTokenCacheOptions(InternalMemoryTokenCacheOptions options)
```

The length does not matter, we have very long names like WithProofOfPossession etc ..

tests/Microsoft.Identity.Test.Unit/CacheTests/InternalCacheOptionsTests.cs
```diff
@@ -0,0 +1,148 @@
...
+                                                              .WithInternalMemoryTokenCacheOptions(new InternalMemoryTokenCacheOptions() { UseSharedCache = true }).BuildConcrete();
+
+                ConfidentialClientApplication app_withoutStaticCache =
+                  ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
+                                                            .WithClientSecret(TestConstants.ClientSecret)
+                                                            .WithHttpManager(httpManager)
+                                                            .BuildConcrete();
+
+                httpManager.AddMockHandlerSuccessfulClientCredentialTokenResponseMessage();
+                await ClientCredsAssertTokenSourceAsync(app1, "S1", TokenSource.IdentityProvider).ConfigureAwait(false);
```
trwalke, line 70: This name sort of implies that you asserting the source of a token already acquired rather than doing the acquiring in the method itself. I think a more descriptive name would be
ClientCredsAcquireAndAssertTokenSourceAsync

trwalke:
> <Project Sdk="MSBuild.Sdk.Extras">

is there something different in this file? i dont see any changes

---
Refers to: src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj:1 in bfc25ae. [](commit_id = bfc25aee18151a3b8a62f43c83fa76718c68797b, deletion_comment = False)

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -178,6 +180,34 @@ internal T WithHttpManager(IHttpManager httpManager)
             return (T)this;
         }
 
+
+        /// <summary>
```
trwalke, line 184: nit: extra line above

trwalke (approved):

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
```diff
@@ -190,5 +199,11 @@ public virtual IMsalHttpClientFactory CreateDefaultHttpClientFactory()
         {
             return new SimpleHttpClientFactory();
         }
+
+        /// <summary>
+        /// On Android, iOS and UWP, MSAL will save the legacy ADAL cache in a known location.
```
SameerK-MSFT, line 204: Can this be done with #if "<Platform>" and define the value(true/false) based on the platform instead of making it virtual and have each platform override it?

src/client/Microsoft.Identity.Client/TokenCache.Notifications.cs
```diff
@@ -83,8 +89,8 @@ async Task ITokenCacheInternal.OnBeforeWriteAsync(TokenCacheNotificationArgs arg
 #endif
         public void SetBeforeAccess(TokenCacheCallback beforeAccess)
         {
-            GuardOnMobilePlatforms();
-
+            Validate();
+
```
SameerK-MSFT, line 93: nit: extra space

tests/devapps/XForms/XForms.UWP/BundleArtifacts/x64.txt
```diff
@@ -0,0 +1,2 @@
+MainPackage=c:\g\msal\tests\devapps\XForms\XForms.UWP\bin\x64\Debug\XForms.UWP_1.0.0.0_x64_Debug.msix
```
SameerK-MSFT, line 1: Please exclude the file.

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
pmaytak: Yea, but generally the pattern in MSAL is for platform specific behavior go into each proxy.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -68,6 +68,8 @@ public string ClientVersion
 
         public IPlatformProxy PlatformProxy { get; internal set; }
 
+        public InternalMemoryTokenCacheOptions TokenCacheAccessorOptions { get; set; }
```
pmaytak, line 71: @bgavrilMS Rename this to `TokenCacheOptions`?

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -1004,5 +1004,12 @@ public static class MsalError
         /// <para>Mitigation</para>For troubleshooting details, see https://aka.ms/msal-net-wam .
         /// </summary>
         public const string InitializeProcessSecurityError = "initialize_process_security_error";
+
+        /// <summary>
+        /// <para>What happens?</para>You configured MSAL cache serialization at the same time with a static internal cache via <see cref="AbstractApplicationBuilder{T}.WithInternalMemoryTokenCacheOptions(InternalMemoryTokenCacheOptions)"/>
+        /// These are mutually exclusive.
+        /// <para>Mitigation</para> Use only one option. Web site and web api scenarios should rely on external cache serialization, as internal cache serialization cannot scale. See https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore        
+        /// </summary>
+        public const string ExternalInternalCacheSerialization = "external_internal_cache_serialization";
```
pmaytak, line 1013: @bgavrilMS `StaticCacheWithExternalSerialization`?

pmaytak (approved):

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
bgavrilMS: TokenCache is such an overloaded word... How about AccessorOptions?

src/client/Microsoft.Identity.Client/AppConfig/InternalMemoryTokenCacheOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// Options for the internal MSAL token caches
+    /// </summary>
+    public class InternalMemoryTokenCacheOptions
+    {
+        /// <summary>
+        /// Creates the default options
+        /// </summary>
+        public InternalMemoryTokenCacheOptions() 
+        {
+            UseSharedCache = false;
```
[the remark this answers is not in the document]
bgavrilMS: Yes, but it makes the code more intentional. It also has not perf implications. 

See `DO explicitly declare the public parameterless constructor in classes, if such a constructor is required.` from https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/constructor

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
pmaytak: Sorry, should've added my reasoning. 
I thought `TokenCacheAccessorOptions` is too specific (options affect the cache, not just accessing it); `AccessorOptions` is too ambiguous (accessor for what?). Maybe `InternalTokenCacheOptions`? Otherwise, I would leave it at `TokenCacheAccessorOptions`. Either way, it's an internal class.

PR #2875, opened 2021-09-07
Add async delegate for WithClientAssertion

### LINKED ISSUE #2863: [Feature Request] ConfidentialClientApplicationBuilder.WithClientAssertion assertion should provide an async overload
Currently `ConfidentialClientApplicationBuilder` has two overloads of `WithClientAssertion` one which simply takes a string and once which takes a synchronous callback `Func<string>`.  There are many scenarios where the assertion, or the key to sign the assertion is not actually in memory, and fetching or creating the assertion requires some I/O. It would be good if `WithClientAssertion` had an overload which accepted an async callback, so sync over async calls could be avoided in the scenarios. Perhaps something like the following.

```c#
public ConfidentialClientApplicationBuilder WithClientAssertion(Func<Task<string>> clientAssertionDelegate)
```

bgavrilMS:
Fixes #2863 

Add async delegate. 

Note to self: always go for async + cancellation token.

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -6,6 +6,8 @@ using System.Linq;
 using System.Collections.Generic;
 using System.Security.Cryptography.X509Certificates;
 using Microsoft.Identity.Client.Internal;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Microsoft.Identity.Client
 {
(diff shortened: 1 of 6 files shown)


src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -170,14 +172,41 @@ public ConfidentialClientApplicationBuilder WithClientAssertion(string signedCli
         /// This is a delegate that computes a Base-64 encoded JWT for each authentication call.</param>
         /// <returns>The ConfidentialClientApplicationBuilder to chain more .With methods</returns>
         /// <remarks> Callers can use this mechanism to cache their assertions </remarks>
+        [Obsolete("Use WithClientAssertion(Func<CancellationToken, Task<string>> clientAssertionAsyncDelegate)")]
```
jmprieur, line 175: Do we need to obsolete this one? Why not keep both overrides?

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -170,14 +172,41 @@ public ConfidentialClientApplicationBuilder WithClientAssertion(string signedCli
...
+            };
+
+
+            Config.SignedClientAssertionDelegate = clientAssertionAsyncDelegate;
+            Config.ConfidentialClientCredentialCount++;
+            return this;
+        }
+
+        /// <summary>
+        /// Configures an async delegate that creates a client assertion. See https://aka.ms/msal-net-client-assertion
```
jmprieur, line 195: We'll have to add a paragraph in https://docs.microsoft.com/azure/active-directory/develop/msal-net-client-assertions

jmprieur (approved):
LGTM
I would propose we don't obsolete the existing delegate.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
bgavrilMS: Ok

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
bgavrilMS: PR added

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -177,7 +179,33 @@ public ConfidentialClientApplicationBuilder WithClientAssertion(Func<string> cli
                 throw new ArgumentNullException(nameof(clientAssertionDelegate));
             }
 
-            Config.SignedClientAssertionDelegate = clientAssertionDelegate;
+            Func<CancellationToken, Task<string>> clientAssertionAsyncDelegate = (token) =>
```
trwalke, line 182: >token [](http://example.com/codeflow?start=82&length=5)

Is this parameter needed?

trwalke (approved):
Minor comment but LGTM otherwise

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
bgavrilMS: It's not, I'll replace it with _

PR #2878, opened 2021-09-08
Correct XML summary element casing

### LINKED ISSUE #2869: [Feature Request] Help service owners by producing metrics to watch and warn them if they don’t respect the guidance
Partners and customers who don't follow the guidance on how to use the token cache or MSAL confidential client application lifecycle might run into problem such as:
- they will be throttled by AzureAd if they keep asking for the same token (if the cache is not used)
- they will run in perf issues if only the user token default cache is used (for the moment).

Provide metrics and alerts, and guidance on what to watch, so that services know when they don't respect the guidance.
See details in [Alerting service owners when they implement bad practices.docx](https://microsoft.sharepoint.com/teams/aad/devex/Shared%20Documents/DotNetDocs/Alerting%20service%20owners%20when%20they%20implement%20bad%20practices%20draft.docx)

Metrics to observe can be placed in AuthenticationResult.AuthenticationResultMetadata so that the developer is able to monitor these during testing/production. Rule violations specifically can go into AuthenticationResult.AuthenticationResultMetadata.RuleViolations.
We will also log these metrics.

Rule Violations to monitor:
 - [ ] Token Cache is not being used and the same request is replayed continuously which will eventually lead to throttling
 - [ ] /common or /organizations is used for client credential flows
 - [ ] Token cache serialization is not used in web app / web api scenarios involving users

Other metrics to monitor:
 - [ ] An indication of when refresh in has been triggered
 - [ ] An indication of whether or not refresh in has failed as it runs on a background thread and only logs errors

### LINKED ISSUE #2872: Pointless page
# Issue Transfer

This issue has been transferred from the Azure SDK for .NET, [#23805](https://github.com/Azure/azure-sdk-for-net/issues/23805).

## Details

[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ba0a1d89-3377-deef-faf6-8d73cba667e7
* Version Independent ID: 0d322b99-0aeb-8dcd-41be-0cadbd07d96b
* Content: [ClientApplicationBase.UserTokenCache Property (Microsoft.Identity.Client) - Azure for .NET Developers](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.clientapplicationbase.usertokencache?view=azure-dotnet#Microsoft_Identity_Client_ClientApplicationBase_UserTokenCache)
* Content Source: [xml/Microsoft.Identity.Client/ClientApplicationBase.xml](https://github.com/Azure/azure-docs-sdk-dotnet/blob/master/xml/Microsoft.Identity.Client/ClientApplicationBase.xml)
* Service: **active-directory**
* GitHub Login: @rloutlaw
* Microsoft Alias: **routlaw**

bgavrilMS:
Fixes #2872

Summary vs summary. Looks like docs.microsoft.com is case sensitive

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
@@ -123,13 +123,13 @@ namespace Microsoft.Identity.Client
         /// </summary>
         public CancellationToken CancellationToken { get; }
 
-        /// <Summary>
+        /// <summary>
         /// Suggested value of the expiry, to help determining the cache eviction time. 
         /// This value is <b>only</b> set on the <code>OnAfterAccess</code> delegate, on a cache write
         /// operation (that is when <code>args.HasStateChanged</code> is <code>true</code>) and when the cache write 
         /// is triggered from the <code>AcquireTokenForClient</code> method. In all other cases it's <code>null</code>, as there is a refresh token, and therefore the
         /// access tokens are refreshable.
-        /// </Summary> 
+        /// </summary> 
         public DateTimeOffset? SuggestedCacheExpiry { get; private set; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -19,16 +19,16 @@ using System.Threading;
 
 namespace Microsoft.Identity.Client
 {
-    /// <Summary>
+    /// <summary>
     /// Abstract class containing common API methods and properties. Both <see cref="Microsoft.Identity.Client.PublicClientApplication"/> and 
     /// ConfidentialClientApplication
     /// extend this class. For details see https://aka.ms/msal-net-client-applications
-    /// </Summary>
+    /// </summary>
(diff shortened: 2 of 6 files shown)


trwalke (approved):
So VS adds lowercase by default. Is there a way to make it use Summary instead?

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -123,13 +123,13 @@ public sealed partial class TokenCacheNotificationArgs
         /// </summary>
```
trwalke, line 123: Do you not need to change this one too?

pmaytak (approved):

jmprieur (approved):
LGTM
Thanks @bgavrilMS

jmprieur:
> So VS adds lowercase by default. Is there a way to make it use Summary instead?

We need to use lower case @trwalke, not the capitalized.

trwalke:
> > So VS adds lowercase by default. Is there a way to make it use Summary instead?
> 
> We need to use lower case @trwalke, not the capitalized.

Yes I see, thanks @jmprieur

PR #2882, opened 2021-09-09
Pii to log exception avoid reflection more timeout

### LINKED ISSUE #2795: [Bug] Preemptively refresh tokens in the background
Latest

**Platform**
All

All with the support of refresh_in coming from eSTS.

**Repro**

Right now, the request that performs the preemptive cache refresh before expiry of a token gets a live latency hit. As discussed with @bgavrilMS, this is not intended behavior.

**Expected behavior**

The expectation is that the preemptive refresh happens in the background, with the initial cached token being immediately returned. Thus, the request is not burdened with the latency of communication with eSTS and seting the token to cache.

SameerK-MSFT:
Fixes #2795  .

Pii for exceptions

**Testing**
NA. Updated the tests for more timeout

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs
@@ -53,19 +53,20 @@ namespace Microsoft.Identity.Client.Internal
                 }
                 catch (MsalServiceException ex)
                 {
-                    string logMsg = $"Background fetch failed. Is AAD down? { ex.IsAadUnavailable()}. MsalServiceException {ex.ToString()}";
+                    string logMsg = $"Background fetch failed MsalServiceException. Is AAD down? { ex.IsAadUnavailable()}";
                     if (ex.StatusCode == 400)
                     {
-                        logger.Error(logMsg);
+                        logger.ErrorPiiWithPrefix(ex, logMsg);
                     }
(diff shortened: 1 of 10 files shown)


SameerK-MSFT:
The PR Pipeline was run twice. No errors.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs
```diff
@@ -53,19 +53,20 @@ internal static void ProcessFetchInBackgroundAsync(Func<Task<AuthenticationResul
...
                     else
                     {
-                        logger.Warning(logMsg);
+                        logger.WarningPiiWithPrefix(ex, logMsg);
                     }
                 }
                 catch (Exception ex)
                 {
-                    logger.Warning($"Background fetch failed Exception {ex.ToString()}");
+                    string logMsg = $"Background fetch failed Exception.";
```
rymeskar, line 68: the 'Exception' at the end of the sentence does not feel right. Either drop it (as it would appear through the logging method) or `failed with exception`.

rymeskar (approved):
The rest LGTM

bgavrilMS (approved):

SameerK-MSFT:
It has been approved by 2 reviewers. Even though one does not have write access, it should be OK.

Issue #2887, opened 2021-09-14
[Feature Request] Log an error when confidential client application use /common or /organizations as an authority


trwalke:
**Is your feature request related to a problem? Please describe.**
It goes against our recommendations to use /common or /organizations as an authority when using confidential client applications. We need some way to notify the developer

**Describe the solution you'd like**
We should log an error indicating to the developer that the current authority is not recommended and direct them to documentation which helps explain.

PR #2888, opened 2021-09-14
Log error when common or organizations were used for client credential flow

trwalke:
Fixes # .
fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2887

Log error when common or organizations were used for client credential flow

**Testing**
unit test

**Performance impact**
negligable

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -44,8 +44,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
             AuthenticationResult authResult = null;
 
             if (AuthenticationRequestParameters.Authority is AadAuthority aadAuthority &&
-                aadAuthority.TenantId != Constants.ConsumerTenant &&
-                aadAuthority.IsCommonOrganizationsOrConsumersTenant())
+                aadAuthority.IsCommonOrOrganizationsTenant())
             {
                 logger.Error(MsalErrorMessage.ClientCredentialWrongAuthority);
             }
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -42,6 +42,12 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
             AuthenticationResult authResult = null;
 
+            if(AuthenticationRequestParameters.Authority.TenantId.Contains("common") ||
+               AuthenticationRequestParameters.Authority.TenantId.Contains("organizations"))
+            {
+                logger.Error(MsalErrorMessage.ClientCredentialWrongAuthority);
+            }
+
             if (!_clientParameters.ForceRefresh && 
                 string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
             {
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -11,6 +11,7 @@ using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 using Microsoft.Identity.Client.Utils;
 using System;
+using Microsoft.Identity.Client.Instance;
 
 namespace Microsoft.Identity.Client.Internal.Requests
 {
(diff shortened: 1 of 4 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -42,6 +42,12 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
             AuthenticationResult authResult = null;
 
+            if(AuthenticationRequestParameters.Authority.TenantId.Contains("common") ||
```
bgavrilMS, line 45: you'll get a null ref for ADFS where tenantId is null. Also, the comparison should be done in a case insensitive manner. Have a look at `AadAuthority.IsCommonOrganizationsOrConsumersTenant`

trwalke: I was looking for something like AadAuthority.IsCommonOrganizationsOrConsumersTenant
I will update

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -394,5 +394,6 @@ public static string CertMustHavePrivateKey(string certificateName)
                 errorCode);
 
         public static string CcsRoutingHintMissing = "Either the userObjectIdentifier or tenantIdenifier are missing. Both are needed to create the ccs routing hint. See https://aka.ms/msal-net/ccsRouting. ";
+        public static string ClientCredentialWrongAuthority = "The current authority is targeting the /common or /organizations endpoint which is not reccomended. See https://aka.ms/msal-net-client-credentials for more details.";
```
jmprieur, line 397: The method is named `IsCommonOrganizationsOrConsumersTenant`. Does it also verify /consumers?
If it not recommended? or not authorized?

jmprieur, line 397:
```suggestion
        public static string ClientCredentialWrongAuthority = "The current authority is targeting the /common or /organizations endpoint which is not recommended. See https://aka.ms/msal-net-client-credentials for more details.";

```

jmprieur (commented):
I proposed a typo fix and asked a question.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
trwalke: Yes, Ill add consumers

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -394,5 +394,6 @@ public static string CertMustHavePrivateKey(string certificateName)
                 errorCode);
 
         public static string CcsRoutingHintMissing = "Either the userObjectIdentifier or tenantIdenifier are missing. Both are needed to create the ccs routing hint. See https://aka.ms/msal-net/ccsRouting. ";
+        public static string ClientCredentialWrongAuthority = "The current authority is targeting the /common, /consumers or /organizations endpoint which is not reccomended. See https://aka.ms/msal-net-client-credentials for more details.";
```
pmaytak, line 397:
```suggestion
        public static string ClientCredentialWrongAuthority = "The current authority is targeting the /common, /consumers or /organizations endpoint which is not recommended. See https://aka.ms/msal-net-client-credentials for more details.";

```

pmaytak (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -42,6 +43,14 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
             AuthenticationResult authResult = null;
 
+            if (AuthenticationRequestParameters.Authority is AadAuthority aadAuthority)
```
bgavrilMS, line 46: nit: combine the 2 conditions in a single if

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -42,6 +43,14 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            {
+                if (aadAuthority.IsCommonOrganizationsOrConsumersTenant())
```
bgavrilMS, line 48: We ban this flow on /common and /organizations because these tenants really mean "I don't know the tenant, let AAD figure it out for me", which only works for user flows. 

I do not think we should bad /consumers, as this is an actual tenant id (98...)

bgavrilMS (changes requested):
don't block /consumers

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
jmprieur: indeed, we shouldn't and sorry I wasn't clear. I meant the error message isn't consistent with the test method.

trwalke: oh ok. will update

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -42,6 +43,13 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
             AuthenticationResult authResult = null;
 
+            if (AuthenticationRequestParameters.Authority is AadAuthority aadAuthority &&
+                aadAuthority.TenantId != Constants.ConsumerTenant &&
```
bgavrilMS, line 47: Please add a helper `aadAuthority.IsCommonOrOrganizations`

bgavrilMS (approved):
One small comment, otherwise LGTM

Issue #2855, opened 2021-08-30
[Bug] client_credentials throttling error is not easy to understand


bgavrilMS:
msal 4.35.1

If app developers mis-configure their app with regards to token caching (i.e. no token cache), AAD replies with HTTP status 429 and a Retry-After header. 

**Expected**:  the exception should make the potential root cause clear - throttling because lack of app cache - and provide an https://aka.ms/ link 

**Actual:** The application developer sees the following exception: 

Microsoft.Identity.Client.MsalThrottledServiceException: An error response was returned by the OAuth2 server, but it could not be parsed. Please inspect the exception properties for details. 
<long exception trace>

StatusCode: 429 
ResponseBody: Too Many Requests 
Headers: Retry-After: 60

PR #2892, opened 2021-09-16
Adding better message for retry after error for client credential request

trwalke:
Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2855

Adding better message for retry after error for client credential request

**Testing**
Unit test

**Performance impact**
none

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
+++ b/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
@@ -92,10 +92,7 @@ namespace Microsoft.Identity.Client
                 ex = new MsalServiceException(errorCode, errorMessage);
             }
 
-            if (brokerHttpResponse != null)
-            {
-                SetHttpExceptionData(ex, brokerHttpResponse);
-            }
+            SetHttpExceptionData(ex, brokerHttpResponse);
 
             ex.CorrelationId = correlationId;
             ex.SubError = subErrorCode;
@@ -116,7 +113,7 @@ namespace Microsoft.Identity.Client
             return ex;
         }
 
-        internal static MsalThrottledServiceException FromThrottledCLientCredentialResponse(HttpResponse httpResponse)
+        internal static MsalThrottledServiceException FromThrottledAuthenticationResponse(HttpResponse httpResponse)
         {
             MsalServiceException ex = new MsalServiceException(MsalError.RequestThrottled, MsalErrorMessage.AadThrottledError);
             SetHttpExceptionData(ex, httpResponse);
--- a/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
+++ b/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using Microsoft.Identity.Client.Http;
 using Microsoft.Identity.Client.Internal;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.OAuth2.Throttling;
 using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client
@@ -45,11 +46,18 @@ namespace Microsoft.Identity.Client
                     innerException);
             }
 
+            if (httpResponse.HeadersAsDictionary != null && httpResponse.HeadersAsDictionary.ContainsKey(ThrottleCommon.ThrottleRetryAfterHeaderResponseValue))
+            {
+                ex = new MsalServiceException(
+                    errorCode,
+                    MsalErrorMessage.AadThrottledError,
+                    innerException);
+            }
+
             if (ex == null)
             {
                 ex = new MsalServiceException(errorCode, errorMessage, innerException);
             }
-
 
             ex.ResponseBody = httpResponse?.Body;
             ex.StatusCode = httpResponse != null ? (int)httpResponse.StatusCode : 0;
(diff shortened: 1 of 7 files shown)


trwalke:
@bgavrilMS I am trying to simulate a throttling  from ESTS but I was not able to make it work. Just want to be sure I have the exact details of the 429 error. I tried to repro by running hundreds of CC requests with force refresh but it didnt seem to be triggering the throttled response from AAD. Any idea how to repro the throttling error?

bgavrilMS:
@trwalke - yeah, it's difficult to convince AAD to throttle you. I think you'll have to rely on injecting HTTP response of 429, which you do.

tests/Microsoft.Identity.Test.Unit/Throttling/ThrottlingAcceptanceTests.cs
```diff
@@ -203,7 +203,29 @@ public async Task RetryAfter_ConfidentialClient_Async()
...
+                var httpManager = httpManagerAndBundle.HttpManager;
+                var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
+                                                              .WithClientSecret(TestConstants.ClientSecret)
+                                                              .WithHttpManager(httpManager)
+                                                              .BuildConcrete();
 
+                httpManager.AddInstanceDiscoveryMockHandler();
+                var tokenResponse = httpManager.AddMockHandlerForThrottledResponseMessage();
+
+                var ex = await AssertException.TaskThrowsAsync<MsalServiceException>(
```
bgavrilMS, line 220: What happens if you run this test without having added the custom error handling code ? Does it reproduce the exception observed by the customer?

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -45,12 +46,19 @@ internal class MsalServiceExceptionFactory
                     innerException);
             }
 
+            if (httpResponse.HeadersAsDictionary != null && httpResponse.HeadersAsDictionary.ContainsKey(ThrottleCommon.ThrottleRetryAfterHeaderResponseValue))
```
bgavrilMS, line 49: I don't think you care about the existence of a header. You should look for not having a status code is 429 and no json response.

nit: HeadersAsDictionary  is bad property. It actually does a lot of work on the getter.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -275,6 +276,14 @@ private static MsalServiceException ExtractErrorsFromTheResponse(HttpResponse re
                 return null;
             }
 
+            if (response.HeadersAsDictionary != null && response.HeadersAsDictionary.ContainsKey(ThrottleCommon.ThrottleRetryAfterHeaderResponseValue))
```
bgavrilMS, line 279: why does the code need to be duplicated?

trwalke: it does not, this is a draft and I am still in the process of refactoring. will be removed

tests/Microsoft.Identity.Test.Unit/Throttling/ThrottlingAcceptanceTests.cs
trwalke: Yes. It does when adding AddMockHandlerForThrottledResponseMessage()

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
trwalke: ok, I was looking for the header because the 429 HttpResponse status code enum was not available in .net 4.5. and 4.6.

bgavrilMS: HttpResponse is an enum, you can cast to int and manually check for 429. You can rely on CreateTooManyRequestsNonJsonResponse() for an example HTTP response.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -96,9 +94,7 @@ private static bool IsThrottled(OAuth2ResponseBase oAuth2Response)
 
             if (brokerHttpResponse != null)
```
pmaytak, line 95: Nit: this null check is not needed, since `SetHttpExceptionData` has one.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -115,13 +111,25 @@ private static bool IsThrottled(OAuth2ResponseBase oAuth2Response)
...
 
-            ex.ResponseBody = httpResponse?.Body;
-            ex.StatusCode = httpResponse != null ? (int)httpResponse.StatusCode : 0;
-            ex.Headers = httpResponse?.Headers;
+            SetHttpExceptionData(ex, httpResponse);
 
             return ex;
         }
 
+        internal static MsalThrottledServiceException FromThrottledCLientCredentialResponse(HttpResponse httpResponse)
```
pmaytak, line 119: typo: `FromThrottledClientCredentialResponse`

pmaytak (approved):

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
bgavrilMS, line 119: Throttling can occur on any flow, not just client credentials. The method name should reflect that.

bgavrilMS (approved):
LGTM, small renaming still needed.

PR #2898, opened 2021-09-22
[DO NOT Review] Reverted the check for Invokable

### LINKED ISSUE #2897: [Bug]  Broker Authentication in 4.36.1 is broken?
**Logs and network traces**
"MSAL.Xamarin.iOS.4.36.1.0.MsalClientException: ErrorCode: uiviewcontroller_required_for_ios_broker" Microsoft.Identity.Client.MsalClientException: UIViewController is null, so MSAL.NET cannot invoke the iOS broker."

MSAL.NET 4.36.1

**Platform**
Xamarin iOS

Mobile, Interactive and with BrokerApp (MS Authenticator)

The app is in production, and I have upgraded to a new version of MSAL.

**Repro**
The Line of Code which raises the exeption is: 
IEnumerable<IAccount> accounts = await App.PCA.GetAccountsAsync();

**Expected behavior**
fill the IEnumerable<IAccount> accounts with list of Accounts

**Actual behavior**
"MSAL.Xamarin.iOS.4.36.1.0.MsalClientException"

**Additional context / logs / screenshots**
After updating from 4.35 to 4.36.1 we got an "MSAL.Xamarin.iOS.4.36.1.0.MsalClientException: ErrorCode: uiviewcontroller_required_for_ios_broker" Microsoft.Identity.Client.MsalClientException: UIViewController is null, so MSAL.NET cannot invoke the iOS broker."

The Line of Code which raises the exeption is: IEnumerable<IAccount> accounts = await App.PCA.GetAccountsAsync();
If downgrade to 4.35 everything works again... in both cases the uiviewcontroller is present....

Even the Demo App "2-With-broker" from https://github.com/Azure-Samples/active-directory-xamarin-native-v2 shows a similar issue after updating to Microsoft.Identity.Client version 4.36.1

Does 4.36.1 support Broker Auth? Please advise or fix this.

SameerK-MSFT:
Fixes #2897.

Removed the check for IsBrokerInvokeable

**Testing**
Manual

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
@@ -51,15 +51,10 @@ namespace Microsoft.Identity.Client.Platforms.iOS
         {
             using (_logger.LogMethodDuration())
             {
-                if (_uIParent?.CallerViewController == null)
-                {
-                    _logger.Error(iOSBrokerConstants.CallerViewControllerIsNullCannotInvokeBroker);
-                    throw new MsalClientException(MsalError.UIViewControllerRequiredForiOSBroker, MsalErrorMessage.UIViewControllerIsRequiredToInvokeiOSBroker);
-                }
-
+                NSObject obj = new NSObject();
                 bool canStartBroker = false;
 
-                _uIParent.CallerViewController.InvokeOnMainThread(() =>
+                obj.InvokeOnMainThread(() =>
                 {
                     if (IsBrokerInstalled(BrokerParameter.UriSchemeBrokerV3))
                     {
@@ -71,7 +66,7 @@ namespace Microsoft.Identity.Client.Platforms.iOS
 
                 if (!canStartBroker)
                 {
-                    _uIParent.CallerViewController.InvokeOnMainThread(() =>
+                    obj.InvokeOnMainThread(() =>
                     {
                         if (IsBrokerInstalled(BrokerParameter.UriSchemeBrokerV2))
                         {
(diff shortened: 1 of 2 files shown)


pmaytak (changes requested):
If I remember correctly, these changes were just a refactoring/simplification.
I think this needs a bit work.
- Previously `CanBrokerSupportSilentAuth` was called in the `if` statement above this code, and it was hardcoded to `false` for iOS. So this code block would never execute in iOS. So now it might break or have some different behavior.
 - `IsBrokerInstalledAndInvokable` is also used in a few other places, my guess, it will also throw the same exception there too in iOS.
 - `IsBrokerInstalledAndInvokable` method was added a while back and is not part of the PR that broke this. But the behavior is actually different for platforms. In Android and WAM it returns a boolean, while in iOS it throws. So maybe iOS broker needs to be somehow refactored to also just return boolean for consistency.
 - There are also tests related to these methods and proxies that should be looked at, maybe updated.

src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
```diff
@@ -51,15 +51,10 @@ public bool IsBrokerInstalledAndInvokable()
         {
             using (_logger.LogMethodDuration())
             {
-                if (_uIParent?.CallerViewController == null)
-                {
-                    _logger.Error(iOSBrokerConstants.CallerViewControllerIsNullCannotInvokeBroker);
-                    throw new MsalClientException(MsalError.UIViewControllerRequiredForiOSBroker, MsalErrorMessage.UIViewControllerIsRequiredToInvokeiOSBroker);
-                }
-
+                NSObject obj = new NSObject();
```
jennyf19, line 54: can you just invoke this if the CallerViewController is null? 

also, what is the e2e with this and the silent flow?

bgavrilMS:
Closing as I think this has been fixed by a different PR>

PR #2902, opened 2021-09-23
Added back CanBrokerSupportSilentAuth

SameerK-MSFT:
Fixes # 2897.

Fixes Silent flow for the broker

**Testing**
manual

**Performance impact**
NA

### DIFF
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -196,7 +196,7 @@ namespace Microsoft.Identity.Client
                 await UserTokenCacheInternal.RemoveAccountAsync(account, requestContext).ConfigureAwait(false);
             }
 
-            if (AppConfig.IsBrokerEnabled)
+            if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 cancellationToken.ThrowIfCancellationRequested();
 
@@ -249,7 +249,7 @@ namespace Microsoft.Identity.Client
             ICacheSessionManager cacheSessionManager, 
             CancellationToken cancellationToken)
         {
-            if (AppConfig.IsBrokerEnabled)
+            if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
                 if (broker.IsBrokerInstalledAndInvokable())
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
@@ -45,7 +45,9 @@ namespace Microsoft.Identity.Client.Internal.Requests.Silent
         protected override async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
         {
             await UpdateRequestWithAccountAsync().ConfigureAwait(false);
-           
+
+            bool isBrokerConfigured = AuthenticationRequestParameters.AppConfig.IsBrokerEnabled &&
+                                      ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
 
             try
             {
@@ -67,7 +69,7 @@ namespace Microsoft.Identity.Client.Internal.Requests.Silent
             {
                 _logger.Verbose($"Token cache could not satisfy silent request.");
 
-                if (AuthenticationRequestParameters.AppConfig.IsBrokerEnabled && ShouldTryWithBrokerError(ex.ErrorCode))
+                if (isBrokerConfigured && ShouldTryWithBrokerError(ex.ErrorCode))
                 {
                     _logger.Info("Attempting to use broker instead. ");                    
                     var brokerAuthResult = await _brokerStrategyLazy.Value.ExecuteAsync(cancellationToken).ConfigureAwait(false);
--- a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
+++ b/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
@@ -172,7 +172,12 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
             return appConfig.BrokerCreatorFunc != null ?
                 appConfig.BrokerCreatorFunc(uiParent, appConfig, Logger) :
                 new NullBroker(Logger);
-        }      
+        }
+
+        public virtual bool CanBrokerSupportSilentAuth()
+        {
+            return true;
+        }
 
         public virtual bool BrokerSupportsWamAccounts => false;
 
--- a/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
@@ -301,7 +301,7 @@ namespace Microsoft.Identity.Test.Unit.BrokerTests
             }
             public override bool IsBrokerInstalledAndInvokable()
             {
-                return true;
+                return false;
             }
         }
 
--- a/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
@@ -280,6 +280,7 @@ namespace Microsoft.Identity.Test.Unit.BrokerTests
             broker.IsBrokerInstalledAndInvokable().Returns(true);
 
             var platformProxy = Substitute.For<IPlatformProxy>();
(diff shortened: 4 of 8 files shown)


src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -249,7 +249,7 @@ private async Task<IEnumerable<IAccount>> GetAccountsInternalAsync(ApiIds apiId,
             ICacheSessionManager cacheSessionManager, 
             CancellationToken cancellationToken)
         {
-            if (AppConfig.IsBrokerEnabled)
+            if (AppConfig.IsBrokerEnabled && ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth())
             {
                 var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
                 if (broker.IsBrokerInstalledAndInvokable())
```
trwalke, line 255: Android proxy already calls this method so now it will be called twice. Should probably remove it from the proxy and add it to the places where broker is used silently, or vice versa

SameerK-MSFT: It is good to have optimization. But IMO, we should not do it in the hotfix. Hotfix should be simple reversal/just necessary changes.

trwalke: Understood, but the IsBrokerInstalledAndInvokable was also moved in the previous PR and it is a relatively simple thing to resolve. Otherwise we will need to make a new issue for this minor optimization.

SameerK-MSFT: The method id removed from Android Proxy. Kept here as this base class is called by several platforms.

pmaytak: Yep. Line 255 is needed for WAM broker too.
With current changes it makes sense. This code block will only be executed for non-iOS and if broker is enabled. And then we check if broker is actually installed.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -67,7 +69,7 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             {
```
pmaytak, line 69: Would line 65 above throw? It would be executed by iOS and call BrokerSilentStrategy. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5da17a099ef18727f821aa13e3f8a426f0578a49/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/BrokerSilentStrategy.cs#L42

Not sure but seems like before the refactoring, iOS would still execute the `BrokerSilentStrategy.ExecuteAsync` but maybe it would throw in a different place, like when it called `SendTokenRequestToBrokerAsync`.

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
```diff
@@ -280,6 +280,7 @@ public void GetAccountWithDuplicateBrokerAccountsTest()
             broker.IsBrokerInstalledAndInvokable().Returns(true);
```
pmaytak, line 280: nit: Previously (https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/4359097e755ddfd9bb1181f6f9655e57d3cca823) `PlatformProxyFactoryTests.cs ` had an extra assert.

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
```diff
@@ -280,6 +280,7 @@ public void GetAccountWithDuplicateBrokerAccountsTest()
...
```
pmaytak, line 281: `IosBrokerMock` needs to return `false` for `CanBrokerSupportSilentAuth`.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5da17a099ef18727f821aa13e3f8a426f0578a49/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs#L295

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
trwalke: I dont think this called `BrokerSilentStrategy.ExecuteAsync` for ios before the refactoring, see code below.

image

from https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2825/files#

Line 65 will throw but the catch block on line 72 will ignore it and simply return the error back to the dev. This is how it used to work

SameerK-MSFT: _clientStrategy is "CacheSilentStrategy". So it will not go to the BrokerSilentStrategy

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
```diff
@@ -547,6 +550,7 @@ public void SpecialAccount_CallsBrokerSilentAuth()
                 mockBroker.IsBrokerInstalledAndInvokable().Returns(false);
 
                 var platformProxy = Substitute.For<IPlatformProxy>();
+                platformProxy.CanBrokerSupportSilentAuth().Returns(true);
```
henrik-me, line 553: >platformProxy.CanBrokerSupportSilentAuth().Returns(true); [](http://example.com/codeflow?start=16&length=57)

seems like we have no utest validating the true state.... shouldn't that be possible to test in the e2e tests running on iOS?

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
pmaytak: Sorry, I actually misspoke there. 
`IosBrockerMock.IsBrokerInstalledAndInvokable` should return `true`. 

And in this test, method, we should add `platformProxy.CanBrokerSupportSilentAuth().Returns(false);`to represent the iOS scenario correctly.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5da17a099ef18727f821aa13e3f8a426f0578a49/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs#L415

pmaytak (approved):
Just one comment about a test. Also tested with WAM and it worked.

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
SameerK-MSFT: >> IosBrockerMock.IsBrokerInstalledAndInvokable should return true.
Done
>> And in this test, method, we should add platformProxy.CanBrokerSupportSilentAuth().Returns(false);
Done
>> Additionally we need a test for a scenario when iOSBroker.IsBrokerInstalledAndInvokable actually throws that exception
iOSBroker inherits from NSObject hence has a dependency on iOS. So, it cannot be tested here.

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
SameerK-MSFT: That will require iOS. So I would do it manually.

trwalke (approved):

bgavrilMS:
Is a hotfix necessary? Can affected folks stay on 4.35?

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -67,7 +69,7 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
                 _logger.Verbose($"Token cache could not satisfy silent request.");
 
-                if (AuthenticationRequestParameters.AppConfig.IsBrokerEnabled && ShouldTryWithBrokerError(ex.ErrorCode))
+                if (isBrokerConfigured && ShouldTryWithBrokerError(ex.ErrorCode))
```
bgavrilMS, line 72: I would have preferred a solution where we don't have to expose this flag. For example, the iosBroker object can be made to respond to the AcquireTokenSilent by throwing an exception.

But anyway, this seems to have been tested, so it will have to do.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
```diff
@@ -172,7 +172,12 @@ public virtual IBroker CreateBroker(ApplicationConfiguration appConfig, CoreUIPa
             return appConfig.BrokerCreatorFunc != null ?
                 appConfig.BrokerCreatorFunc(uiParent, appConfig, Logger) :
                 new NullBroker(Logger);
-        }      
+        }
+
+        public virtual bool CanBrokerSupportSilentAuth()
```
bgavrilMS, line 177: Should be a property, not a method. There is now work done, it's just a flag.

PR #2907, opened 2021-09-28
Refactored WwwAuthenticateParameters

abatishchev:

- Added optional `cancellationToken` to `CreateFromResourceResponseAsync()`
- Used `is null`, `is object` for null checks
- Used `string.Equals()` with `StringComparison.OrdinalIgnoreCase` for string comparison
- Removed no-op `try/catch` that does nothing but erases the exception stack trace

**Testing**

- Refactored unit tests
- Added integration test to call [ARM](https://docs.microsoft.com/en-us/rest/api/resources/subscriptions/get)

**Performance impact**

Will slightly improve the performance of the helper methods.

### DIFF
--- a/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
+++ b/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
@@ -10,6 +10,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.PlatformsCommon.Factories;
+using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client
 {
@@ -34,18 +35,17 @@ namespace Microsoft.Identity.Client
         {
             get
             {
-                if (_scopes is object)
+                if (_scopes != null)
                 {
                     return _scopes;
                 }
-                else if (!string.IsNullOrEmpty(Resource))
+
+                if (!string.IsNullOrEmpty(Resource))
                 {
                     return new[] { $"{Resource}/.default" };
                 }
-                else
-                {
-                    return null;
-                }
+
+                return null;
             }
             set
             {
@@ -98,13 +98,14 @@ namespace Microsoft.Identity.Client
         /// <param name="httpResponseHeaders">HttpResponseHeaders.</param>
         /// <param name="scheme">Authentication scheme. Default is "Bearer".</param>
         /// <returns>The parameters requested by the web API.</returns>
+        /// <remarks>Currently it only supports the Bearer scheme</remarks>
         public static WwwAuthenticateParameters CreateFromResponseHeaders(
             HttpResponseHeaders httpResponseHeaders,
             string scheme = "Bearer")
         {
             if (httpResponseHeaders.WwwAuthenticate.Any())
             {
-                // TODO: could it be Pop too?
+                // TODO: add POP support
                 AuthenticationHeaderValue bearer = httpResponseHeaders.WwwAuthenticate.First(v => string.Equals(v.Scheme, scheme, StringComparison.OrdinalIgnoreCase));
                 string wwwAuthenticateValue = bearer.Parameter;
                 return CreateFromWwwAuthenticateHeaderValue(wwwAuthenticateValue);
@@ -122,10 +123,10 @@ namespace Microsoft.Identity.Client
         {
             if (string.IsNullOrWhiteSpace(wwwAuthenticateValue))
             {
-                throw new ArgumentException($"'{nameof(wwwAuthenticateValue)}' cannot be null or whitespace.", nameof(wwwAuthenticateValue));
+                throw new ArgumentNullException(nameof(wwwAuthenticateValue));
             }
 
-            IDictionary<string, string> parameters = SplitWithQuotes(wwwAuthenticateValue, ',')
+            IDictionary<string, string> parameters = CoreHelpers.SplitWithQuotes(wwwAuthenticateValue, ',')
                 .Select(v => ExtractKeyValuePair(v.Trim()))
                 .ToDictionary(pair => pair.Key, pair => pair.Value, StringComparer.OrdinalIgnoreCase);
 
@@ -155,7 +156,7 @@ namespace Microsoft.Identity.Client
         {
             if (httpClientFactory is null)
             {
-                throw new ArgumentException($"'{nameof(httpClientFactory)}' cannot be null.", nameof(httpClientFactory));
+                throw new ArgumentNullException(nameof(httpClientFactory));
             }
 
             var httpClient = httpClientFactory.GetHttpClient();
@@ -173,11 +174,11 @@ namespace Microsoft.Identity.Client
         {
             if (httpClient is null)
             {
-                throw new ArgumentException($"'{nameof(httpClient)}' cannot be null.", nameof(httpClient));
+                throw new ArgumentNullException(nameof(httpClient));
             }
             if (string.IsNullOrWhiteSpace(resourceUri))
             {
-                throw new ArgumentException($"'{nameof(resourceUri)}' cannot be null or whitespace.", nameof(resourceUri));
+                throw new ArgumentNullException(nameof(resourceUri));
             }
 
             // call this endpoint and see what the header says and return that
@@ -202,8 +203,8 @@ namespace Microsoft.Identity.Client
                 scheme);
 
             // read the header and checks if it contains an error with insufficient_claims value.
-            if (string.Equals(parameters.Error, "insufficient_claims", StringComparison.OrdinalIgnoreCase) &&
-                parameters.Claims is object)
+            if (parameters.Claims != null &&
+                string.Equals(parameters.Error, "insufficient_claims", StringComparison.OrdinalIgnoreCase))
             {
                 return parameters.Claims;
             }
@@ -274,54 +275,6 @@ namespace Microsoft.Identity.Client
             return wwwAuthenticateParameters;
         }
 
-        internal static List<string> SplitWithQuotes(string input, char delimiter)
-        {
-            List<string> items = new List<string>();
-
-            if (string.IsNullOrWhiteSpace(input))
-            {
-                return items;
-            }
-
-            int startIndex = 0;
-            bool insideString = false;
-            int nestingLevel = 0;
-            string item;
-            for (int i = 0; i < input.Length; i++)
-            {
-                if (input[i] == delimiter && !insideString && nestingLevel == 0)
-                {
-                    item = input.Substring(startIndex, i - startIndex);
-                    if (!string.IsNullOrWhiteSpace(item.Trim()))
-                    {
-                        items.Add(item);
-                    }
-
-                    startIndex = i + 1;
-                }
-                else if (input[i] == '"')
-                {
-                    insideString = !insideString;
-                }
-                else if (input[i] == '{')
-                {
-                    nestingLevel++;
-                }
-                else if (input[i] == '}')
-                {
-                    nestingLevel--;
-                }
-            }
-
-            item = input.Substring(startIndex);
-            if (!string.IsNullOrWhiteSpace(item.Trim()))
-            {
-                items.Add(item);
-            }
-
-            return items;
-        }
-
         /// <summary>
         /// Extracts a key value pair from an expression of the form a=b
         /// </summary>
@@ -329,9 +282,10 @@ namespace Microsoft.Identity.Client
         /// <returns>Key Value pair</returns>
         private static KeyValuePair<string, string> ExtractKeyValuePair(string assignment)
         {
-            string[] segments = SplitWithQuotes(assignment, '=')
+            string[] segments = CoreHelpers.SplitWithQuotes(assignment, '=')
                 .Select(s => s.Trim().Trim('"'))
                 .ToArray();
+
             if (segments.Length != 2)
             {
                 throw new ArgumentException(nameof(assignment), $"{assignment} isn't of the form a=b");
--- a/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
+++ b/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 using System;
@@ -32,7 +32,7 @@ namespace Microsoft.Identity.Client
         {
             get
             {
-                if (_scopes != null)
+                if (_scopes is object)
                 {
                     return _scopes;
                 }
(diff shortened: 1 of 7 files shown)


jmprieur:
@abatishchev will you update unit tests?

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -32,7 +33,7 @@ public IEnumerable<string> Scopes
         {
             get
             {
-                if (_scopes != null)
+                if (_scopes is object)
```
bgavrilMS, line 36: I have not seen this type of check before. I am curious - how does it differ?

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -134,13 +135,14 @@ public static WwwAuthenticateParameters CreateFromWwwAuthenticateHeaderValue(str
         /// Create the authenticate parameters by attempting to call the resource unauthenticated, and analyzing the response.
         /// </summary>
         /// <param name="resourceUri">URI of the resource.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
         /// <returns>WWW-Authenticate Parameters extracted from response to the un-authenticated call.</returns>
-        public static async Task<WwwAuthenticateParameters> CreateFromResourceResponseAsync(string resourceUri)
+        public static async Task<WwwAuthenticateParameters> CreateFromResourceResponseAsync(string resourceUri, CancellationToken cancellationToken = default)
```
bgavrilMS, line 140: +1

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
abatishchev: This is a new feature in [C# 7.0](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching), part of added pattern matching. It bypasses any potential operator overloads (for either `==` and/or `!=`) and directly emits the intended IL for null/not null comparison.

abatishchev:
@jmprieur sorry, rushed to open a pull request. Working on adding/updating tests.

abatishchev:
@bgavrilMS, @jmprieur all tests have passed. I think it's ready for a re-review. thanks!

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
```diff
@@ -37,5 +37,22 @@ public async Task CreateWwwAuthenticateResponseFromGraphUrlAsync()
             Assert.IsNull(authParams.Claims);
             Assert.IsNull(authParams.Error);
         }
+
+        [DataRow("management.azure.com", "a645d3be-5a4a-40ef-82a1-d5f5358a424c", "login.windows.net", "72f988bf-86f1-41af-91ab-2d7cd011db47")]
+        [DataRow("api-dogfood.resources.windows-int.net", "5d04a672-05ce-492b-958f-d225b6a67926", "login.windows-ppe.net", "f686d426-8d16-42db-81b7-ab578e110ccd")]
```
bgavrilMS, line 42: Is this stable?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
```diff
@@ -37,5 +37,22 @@ public async Task CreateWwwAuthenticateResponseFromGraphUrlAsync()
...
+        [DataTestMethod]
+        public async Task CreateWwwAuthenticateResponseFromAzureResourceManagerUrlAsync(string hostName, string subscriptionId, string authority, string tenantId)
```
bgavrilMS, line 44: Could you add a comment on each line of what exactly those resources are and how they respond to the request today (i.e. actual 401 header string). This is to help us in case the resources actually do change their response.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
```diff
@@ -37,5 +37,22 @@ public async Task CreateWwwAuthenticateResponseFromGraphUrlAsync()
...
+        {
+            const string apiVersion = "2020-08-01";
+            var url = $"https://{hostName}/subscriptions/{subscriptionId}?api-version={apiVersion}";
+            var authParams = await WwwAuthenticateParameters.CreateFromResourceResponseAsync(url).ConfigureAwait(false);
+
+            Assert.IsNull(authParams.Resource);
+            Assert.AreEqual($"https://{authority}/{tenantId}", authParams.Authority);
+            Assert.IsNull(authParams.Scopes);
```
bgavrilMS, line 52: Is it worth asserting the scopes?

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
abatishchev: Discussed offline. Changed subscriptions to something more stable.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
abatishchev: I copied the assertions from the other 2 tests from the above, for consistency.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
abatishchev: Annotated the test. Please see if looks more readable now.

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
abatishchev: @bgavrilMS few questions/ideas:
1. Right now it constructs `HttpRequestMessage` and passes it to `HttpClient.SendAsync()`. Should it just call `GetAsync()` instead?
2. Both `HttpClient` and `HttpRequestMessage` implement `IDisposable`. Should both be disposed? Might speed up the freeing of the socket, avoid [socket exhaustion](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#issues-with-the-original-httpclient-class-available-in-net).

bgavrilMS: 1. I don't see why not. 
2. This is a great point. The guidance for SDKs regards to HttpClient is that SDKs should allow calling apps to inject the HttpClient themselves instead of creating it a new one (socket exhaustion) or having a static one (DNS expiration problems). MSAL allows to do this https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-provide-httpclient

So for this method I'd recommend to actually add another optional parameter `IMsalHttpClientFactory`. if people don't add use it, then we can default to MSAL's .. see `CreateDefaultHttpClientFactory`.

bgavrilMS: Obviously the latter is a bit more complex. An simple improvement would be to just use a static HttpClient.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
abatishchev: I've added 2 overload, once accepting HTTP client and another HTTP client factory. And kept the existing one which doesn't accept any. Removing it would be a braking change.

Let me in this PR just start using a static instance of HTTP client.

I couldn't figure out how to use `CreateDefaultHttpClientFactory()` since it requires an instance of proxy which in turn is an internal interface and I'm not sure how to properly instantiate it.

Another problem is that this helper is `static class` so not really DI container friendly. A solution for that would be converting it to a non-static, keep the existing static API, maybe mark it as obsolete. Looks like that's what has happened with the similar helper in ADAL earlier.

abatishchev: Ah, `PlatformProxyFactory` looks like is the way to instantiate `proxy` which will give me the default factory. Hang on, will do.

abatishchev: @bgavrilMS please re-review.

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -137,12 +138,36 @@ public static WwwAuthenticateParameters CreateFromWwwAuthenticateHeaderValue(str
...
+        /// <summary>
+        /// Create the authenticate parameters by attempting to call the resource unauthenticated, and analyzing the response.
+        /// </summary>
+        /// <param name="httpClientFactory">Factory to produce an instance of <see cref="HttpClient"/> to make the request with.</param>
+        /// <param name="resourceUri">URI of the resource.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
+        /// <returns>WWW-Authenticate Parameters extracted from response to the un-authenticated call.</returns>
+        public static Task<WwwAuthenticateParameters> CreateFromResourceResponseAsync(IMsalHttpClientFactory httpClientFactory, string resourceUri, CancellationToken cancellationToken = default)
+        {
+            var httpClient = httpClientFactory.GetHttpClient();
```
bgavrilMS, line 156: Pls add some null checks for these params as they are all mandatory. Otherwise LGTM.

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -134,13 +136,38 @@ public static WwwAuthenticateParameters CreateFromWwwAuthenticateHeaderValue(str
         /// Create the authenticate parameters by attempting to call the resource unauthenticated, and analyzing the response.
         /// </summary>
         /// <param name="resourceUri">URI of the resource.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
         /// <returns>WWW-Authenticate Parameters extracted from response to the un-authenticated call.</returns>
-        public static async Task<WwwAuthenticateParameters> CreateFromResourceResponseAsync(string resourceUri)
+        public static Task<WwwAuthenticateParameters> CreateFromResourceResponseAsync(string resourceUri, CancellationToken cancellationToken = default)
+        {
+            var httpClientFactory = PlatformProxyFactory.CreatePlatformProxy(null).CreateDefaultHttpClientFactory();
```
bgavrilMS, line 143: Yes, perfect.

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
trwalke: nit: While this is definitely interesting, these checks obfuscate the fact that we are checking for null here. The word "is" usually implies that you are checking if an object is of a certain type, not whether or not it is null so this may be confusing to someone who is not aware of this feature and they may end up adding the null check in addition to this.

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -134,13 +136,52 @@ public static WwwAuthenticateParameters CreateFromWwwAuthenticateHeaderValue(str
...
+        /// </summary>
+        /// <param name="httpClientFactory">Factory to produce an instance of <see cref="HttpClient"/> to make the request with.</param>
+        /// <param name="resourceUri">URI of the resource.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
+        /// <returns>WWW-Authenticate Parameters extracted from response to the un-authenticated call.</returns>
+        public static Task<WwwAuthenticateParameters> CreateFromResourceResponseAsync(IMsalHttpClientFactory httpClientFactory, string resourceUri, CancellationToken cancellationToken = default)
+        {
+            if (httpClientFactory is null)
+            {
+                throw new ArgumentException($"'{nameof(httpClientFactory)}' cannot be null.", nameof(httpClientFactory));
```
trwalke, line 158: nit: We typically use ArgumentNullException for these types of checks on the api surface.

trwalke (approved):
Minor nit pick and a comment about the is object check

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2907#discussion_r720451523

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
abatishchev: I agree that it's a (relatively) new feature and might be confusing. As any other new feature :)

There are some potential benefits in using it, plus many most popular OSS projects have switched to using it too.

So here's our options:
1. Continue to use `!= null` and `== null`
2. Use `is object` and `is null` (my preference)
3. Use `is not null` and `is null` (I personally find the former the least readable)

trwalke: I like option 1 or 3.
Thoughts @bgavrilMS?

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
abatishchev: Yeah, I noticed that too. Let me explain.

The only already existing check is for resourceId for not null or empty or white space. And it throws just `ArgumentException`. Some people may argue that if the input is an empty string the code can't throw `ArgumentNullException` because it's not null. But I personally think that's nitpicking and matters less than readable and consistent code.

In this PR I'm adding two more checks, both are for reference types so I'm just checking for not null.
And the caveat is in the order of parameters: `ArgumentException(message, paramName)` vs `ArgumentNullException(paramName, message)`. Why the inconsistency is a question to BCL. So in our case the code looks little bit more messy and less readable.

So what I would do is to throw `ArgumentNullException` in all checks. What you think?

bgavrilMS: Yes, we do that as well in most places, i.e. throw `ArgumentNullException` for empty string. 

Strings should really not be nullable :), but unfortunately we are not a library that uses the nullable references feature, so we have to live with these small inconsistencies.

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
bgavrilMS: Very interesting problem. It looks like "obj is null" is something we can do today. But "object is not null" is a C# 9.0 feature, which we don't currently use (but we probably should). 

I agree that "a is object" is a bit counter-intuitive.

According to https://stackoverflow.com/questions/40676426/what-is-the-difference-between-x-is-null-and-x-null it looks like the compiler (Roslyn) was actually modified so that it emits the same kind of logic as long as `==` and `!=` are not overloaded (I believe we haven't overloaded them for any object in MSAL). '

So maybe let's keep on using option 1 for consistency with MSAL, and separately we can look at moving to C# 9 and adding some analyzers for this kind of problem. Any changes should be consistent across the lib.

bgavrilMS: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2921

bgavrilMS:
@abatishchev - I made 2 small changes as per @trwalke 's comments and I'm merging this, so as to be able to review your other PR with the tenant id.

PR #2914, opened 2021-09-29
Modifying region acquisition from env variable

trwalke:
Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2803#

Modifying region acquisition from env variable

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
@@ -160,7 +160,7 @@ namespace Microsoft.Identity.Client.Region
 
         private async Task<RegionInfo> DiscoverAsync(ICoreLogger logger, CancellationToken requestCancellationToken)
         {
-            string region = Environment.GetEnvironmentVariable("REGION_NAME");
+            string region = Environment.GetEnvironmentVariable("REGION_NAME")?.Replace(" ", string.Empty)?.ToLower();
 
             if (ValidateRegion(region, "REGION_NAME env variable", logger)) // this is just to validate the region string
             {
--- a/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
@@ -160,7 +160,7 @@ namespace Microsoft.Identity.Client.Region
 
         private async Task<RegionInfo> DiscoverAsync(ICoreLogger logger, CancellationToken requestCancellationToken)
         {
-            string region = Environment.GetEnvironmentVariable("REGION_NAME")?.Replace(" ", string.Empty)?.ToLower();
+            string region = Environment.GetEnvironmentVariable("REGION_NAME")?.Replace(" ", string.Empty).ToLowerInvariant();
 
             if (ValidateRegion(region, "REGION_NAME env variable", logger)) // this is just to validate the region string
             {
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -179,7 +179,6 @@ public async Task SuccessfulResponseFromRegionalizedAuthorityAsync()
         }
 
         [DataTestMethod]
-        [DataRow("Region with spaces")]
```
trwalke, line 182: Regions with spaces are no longer invalid as we account for them in RegionManager.cs

bgavrilMS: nit: not much point for a having a DataTestMethod if there is only 1 data row ...

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -160,7 +160,7 @@ private async Task<RegionInfo> DiscoverAndCacheAsync(string azureRegionConfig, I
 
         private async Task<RegionInfo> DiscoverAsync(ICoreLogger logger, CancellationToken requestCancellationToken)
         {
-            string region = Environment.GetEnvironmentVariable("REGION_NAME");
+            string region = Environment.GetEnvironmentVariable("REGION_NAME")?.Replace(" ", string.Empty)?.ToLower();
```
bgavrilMS, line 163: Please add test that shows that this transformation is correct.

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
jmprieur: @SameerK-MSFT, what happens if the environment variable does not exist?

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -160,7 +160,7 @@ private async Task<RegionInfo> DiscoverAndCacheAsync(string azureRegionConfig, I
 
         private async Task<RegionInfo> DiscoverAsync(ICoreLogger logger, CancellationToken requestCancellationToken)
         {
-            string region = Environment.GetEnvironmentVariable("REGION_NAME");
```
bgavrilMS, line 163: `toLower` is culture sensitive, please use `toLowerInvariant` https://stackoverflow.com/questions/6225808/string-tolower-and-string-tolowerinvariant

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
bgavrilMS: I just tested and it looks like @SameerK-MSFT  is right. According to the docs https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and- the null-conditional operator is short-circuiting, so the `ToLower()` never executes. 

Great question, I didn't know about this.

jmprieur: Thanks @SameerK-MSFT and @bgavrilMS : I did not know that was the way it worked!

SameerK-MSFT (approved):

pmaytak (approved):

bgavrilMS (approved):

PR #2922, opened 2021-09-30
Added Tenant ID to WwwAuthenticateParameters

abatishchev:

- Added property `TenantID` to `WwwAuthenticateParameters`
- Which underneath calls internal class `Authority` to parse `authority uri`

**Testing**

- Updated unit and integration tests

**Performance impact**

- There should be no impact as a new property has been added

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Authority.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Authority.cs
@@ -5,7 +5,6 @@ using System;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
-using Microsoft.Identity.Client.Instance.Discovery;
 using Microsoft.Identity.Client.Internal;
 
 namespace Microsoft.Identity.Client.Instance
@@ -18,7 +17,7 @@ namespace Microsoft.Identity.Client.Instance
     {
         protected Authority(AuthorityInfo authorityInfo)
         {
-            if (authorityInfo == null)
+            if (authorityInfo is null)
             {
                 throw new ArgumentNullException(nameof(authorityInfo));
             }
@@ -43,9 +42,9 @@ namespace Microsoft.Identity.Client.Instance
         /// 2. If there is a request authority, try to use it.
         ///     2.1. If the request authority is not "common", then use it
         ///     2.2  If the request authority is "common", ignore it, and use 1.1
-        ///     
-        /// Special cases: 
-        /// 
+        ///
+        /// Special cases:
+        ///
         /// - if the authority is not defined at the application level and the request level is not AAD, use the request authority
         /// - if the authority is defined at app level, and the request level authority of is of different type, throw an exception
         ///
@@ -56,7 +55,7 @@ namespace Microsoft.Identity.Client.Instance
             string requestHomeAccountTenantId = null)
         {
             var configAuthorityInfo = requestContext.ServiceBundle.Config.AuthorityInfo;
-            if (configAuthorityInfo == null)
+            if (configAuthorityInfo is null)
             {
                 throw new ArgumentNullException(nameof(configAuthorityInfo));
             }
@@ -68,21 +67,22 @@ namespace Microsoft.Identity.Client.Instance
             {
                 // ADFS and B2C are tenant-less, no need to consider tenant
                 case AuthorityType.Adfs:
-                    return requestAuthorityInfo == null ?
-                        new AdfsAuthority(configAuthorityInfo) :
-                        new AdfsAuthority(requestAuthorityInfo);
-
+                {
+                    return requestAuthorityInfo is null ?
+                               new AdfsAuthority(configAuthorityInfo) :
+                               new AdfsAuthority(requestAuthorityInfo);
+                }
                 case AuthorityType.B2C:
-
-                    if (requestAuthorityInfo != null)
+                {
+                    if (requestAuthorityInfo is object)
                     {
                         return new B2CAuthority(requestAuthorityInfo);
                     }
                     return new B2CAuthority(configAuthorityInfo);
-
+                }
                 case AuthorityType.Aad:
-
-                    if (requestAuthorityInfo == null)
+                {
+                    if (requestAuthorityInfo is null)
                     {
                         return CreateAuthorityWithTenant(configAuthorityInfo, requestHomeAccountTenantId);
                     }
@@ -101,18 +101,20 @@ namespace Microsoft.Identity.Client.Instance
                     }
 
                     return CreateAuthorityWithTenant(configAuthorityInfo, requestHomeAccountTenantId);
-
+                }
                 default:
+                {
                     throw new MsalClientException(
                         MsalError.InvalidAuthorityType,
                         "Unsupported authority type");
+                }
             }
         }
 
         private static void ValidateTypeMismatch(AuthorityInfo configAuthorityInfo, AuthorityInfo requestAuthorityInfo)
         {
             if (!configAuthorityInfo.IsDefaultAuthority &&
-                            requestAuthorityInfo != null &&
+                            requestAuthorityInfo is object &&
                             configAuthorityInfo.AuthorityType != requestAuthorityInfo.AuthorityType)
             {
                 throw new MsalClientException(
@@ -122,6 +124,7 @@ namespace Microsoft.Identity.Client.Instance
                         requestAuthorityInfo.AuthorityType));
             }
         }
+
         public static Authority CreateAuthority(string authority, bool validateAuthority = false)
         {
             return CreateAuthority(AuthorityInfo.FromAuthorityUri(authority, validateAuthority));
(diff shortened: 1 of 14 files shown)


src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -132,18 +135,23 @@ public static Authority CreateAuthority(AuthorityInfo authorityInfo)
             switch (authorityInfo.AuthorityType)
             {
                 case AuthorityType.Adfs:
+                {
                     return new AdfsAuthority(authorityInfo);
-
+                }
```
bgavrilMS, line 140: From a style perspective, we copied the .NET Core repo styles. We might have relaxed the rules on this particular issue (1 line statements not needing braces). It keeps the code concise.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
```diff
@@ -1,67 +1,70 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
bgavrilMS, line 1: nit: spacing problem?

tests/Microsoft.Identity.Test.Unit/WwwAuthenticateParametersTests.cs
```diff
@@ -178,6 +178,28 @@ public async Task CreateFromResourceResponseAsync_HttpClientFactory_Async()
             httpClientFactory.Received().GetHttpClient();
         }
 
+        [TestMethod]
```
bgavrilMS, line 181: Is there a unit test that checks that if the WWWAuthenticate header has an ADFS or B2C authority, the tenant id is null?

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
abatishchev: It was a BOM, removed.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
abatishchev: Rolled back

tests/Microsoft.Identity.Test.Unit/WwwAuthenticateParametersTests.cs
abatishchev: Will add

abatishchev: Please assist me in authoring the test when it uses `TestConstants.B2CAuthority` as the tenant id doesn't come `null`.

bgavrilMS: Ah, sorry about @abatishchev , I made a mistake. B2C is tenanted, https://login.microsoftonline.in/tfp/tenant/{B2CSignUpSignIn}/ has tenantId `tenant`.

abatishchev: @bgavrilMS cool, thanks. updated the tests. please take a look again.

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -60,6 +60,13 @@ public IEnumerable<string> Scopes
         /// </summary>
         public string Authority { get; set; }
 
+        /// <summary>
+        /// AAD tenant ID to which the Azure subscription belongs to.
```
jmprieur, line 64: This can work without a subscription ...

```suggestion
        /// Azure AD tenant ID.
```

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -60,6 +60,13 @@ public IEnumerable<string> Scopes
...
+        /// </summary>
+        public string TenantId => Instance.Authority
```
jmprieur, line 66: @bgavrilMS : I'm not a big fan of this solution to extract the tenant

WWWAuthenticateParameters was a small class that we could have split into a small assembly that SAL and MISE could use (discussed with @brentschmaltz).
Does `Authority` brings a more dependencies?

jmprieur (commented):
LGTM
One feedback/comment for @bgavrilMS

tests/Microsoft.Identity.Test.Unit/WwwAuthenticateParametersTests.cs
```diff
@@ -193,48 +262,57 @@ public async Task CreateFromResourceResponseAsync_HttpClient_Null_Async(HttpClie
...
         [TestMethod]
-        public async Task CreateFromResourceResponseAsync_ResourceUri_Async(string resourceUri)
+        public async Task CreateFromResourceResponseAsync_Incorrect_ResourceUri_Async(string resourceUri)
         {
             Func<Task> action = () => WwwAuthenticateParameters.CreateFromResourceResponseAsync(resourceUri);
 
             await Assert.ThrowsExceptionAsync<ArgumentNullException>(action).ConfigureAwait(false);
         }
 
+        public async Task CreateFromResourceResponseAsync_ResourceUri_Async(string resourceUri)
```
pmaytak, line 309: Is this a test? Missing TestMethod attribute?

tests/Microsoft.Identity.Test.Unit/WwwAuthenticateParametersTests.cs
```diff
@@ -178,6 +179,74 @@ public async Task CreateFromResourceResponseAsync_HttpClientFactory_Async()
             httpClientFactory.Received().GetHttpClient();
         }
 
+        [TestMethod]
+        public async Task CreateFromResourceResponseAsync_HttpClientFactory_Arm_TenantId_Async()
```
pmaytak, line 183: Maybe can just add the `TenantId` assert to `CreateFromResourceResponseAsync_HttpClientFactory_Async`?

pmaytak (approved):

trwalke (approved):

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
bgavrilMS: Ok, I was not aware of this ask. 

We could split out the logic of parsing the authority URL and exposing the tenant ID. But I must warn that parsing works today because MSAL deals only with AAD, ADFS and B2C and can tell the authorities apart. If SAL and MISE want to support any IdP, extracting the `tenantid` will not be as simple. 

If we were to extract a package with WWWAuthenticateHeaders, I would move the TenantId to be an extension method in MSAL. Sadly C# does not support extension properties, so I propose that we change this in "GetTenantId" (a method is also more consistent with the fact that there is non-trivial work done there). Would this work @jmprieur  @abatishchev ?

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -60,6 +60,13 @@ public IEnumerable<string> Scopes
...
+                                          .CreateAuthority(Authority, validateAuthority: true)
+                                          .TenantId;
```
bgavrilMS, line 68: Also to @jmprieur 's point then, let's remove this helper method, as it is tied to an MSAL implementation:

```csharp
 public static Task<WwwAuthenticateParameters> CreateFromResourceResponseAsync(IMsalHttpClientFactory httpClientFactory, string resourceUri, CancellationToken cancellationToken=default(CancellationToken));
```

bgavrilMS (changes requested):
See the 2 comments related to Jean-Marc's ask

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
abatishchev: Let me make it a method, so we could easier convert it to an extension method in the future, without introducing breaking changes or anything.

tests/Microsoft.Identity.Test.Unit/WwwAuthenticateParametersTests.cs
abatishchev: This test just makes sure `httpClientFactory` is being called (and passed over). With the latest changes around `Func<HttpClient>` it became more clear.

tests/Microsoft.Identity.Test.Unit/WwwAuthenticateParametersTests.cs
abatishchev: Oh, good catch. Thanks! Fixed.

Issue #2832, opened 2021-08-19
[Feature Request] Expose indication of Refresh In


bgavrilMS:
It's difficult for app developers to know if refresh_in was configured or if it works.

We could expose some values in the AuthenticationResultMetadata: 

- [p1] time remaining before the AT will be refreshed
- [p2] the reason why a call to the identity provider was made (we already collect this for telemetry)

        None = 0, // When the cache is not hit to make the request (interactive call, username password call, device code flow, etc.)
        ForceRefresh = 1, // When the token request goes to ESTS because force_refresh was set to true
        NoCachedAT = 2, // When the token request goes to ESTS because no cached access token exists
        Expired = 3, // When the token request goes to ESTS because cached access token expired
        RefreshIn = 4, // When the token request goes to ESTS because refresh_in was used and the existing token needs to be refreshed

bgavrilMS:
We should be able to test this now, the lab has accounts that can be used for this, see

https://msidlab.com/api/config?optionalclaim=refresh_in&usertype=federated&signInAudience=azureadmyorg
https://msidlab.com/api/config?optionalclaim=refresh_in&usertype=cloud&signInAudience=azureadmyorg

PR #2924, opened 2021-10-01
Adding refresh indication and time remaining before refresh in to met…

trwalke:
Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2832

Adding refresh indication and time remaining before refresh in to metrics

**Testing**
Unit test

**Performance impact**


src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.;
+
+namespace Microsoft.Identity.Client.Cache
+{
+    /// <summary>
+    /// Enum describing the reason for cache refresh and fetching access token from ESTS.
+    /// </summary>
+    public enum CacheInfo
```
trwalke, line 9: Since we already have the cache info with the refresh in value in it from telemetry. Does it make sense to expose this object. Perhaps this other information can be useful at some point

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -259,6 +259,33 @@ public async Task AcquireToken_ManagedUsernameIncorrectPassword_AcquireTokenSucc
             Assert.Fail("Bad exception or no exception thrown");
         }
 
+        [TestMethod]
+        [TestCategory(TestCategories.ADFS)]
+        public async Task AcquireTokenRefreshInTestAsync()
+        {
+            LabResponse labResponse = await LabUserHelper.GetSpecificUserAsync("fIDLABOC_Refresh_IN@MSIDLAB6.com").ConfigureAwait(false);
```
trwalke, line 266: @bgavrilMS I tried to use this for testing the refresh in value but it is not returning the value in the auth response. I there something special that I have to do?

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
trwalke: Also, the name CacheInfoTelemetry didnt seem right if this is to be exposed. What are your thoughts on this change?

Should I only expose the refresh in value?

@bgavrilMS

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -161,6 +161,11 @@ public partial class AuthenticationResult
             CorrelationId = correlationID;
             ApiEvent = apiEvent;
             AuthenticationResultMetadata = new AuthenticationResultMetadata(tokenSource);
+
+            if (msalAccessTokenCacheItem.RefreshOn.HasValue)
+            {
+                AuthenticationResultMetadata.RemainingTimeBeforeRefresh = msalAccessTokenCacheItem.RefreshOn.Value.Millisecond - DateTimeOffset.UtcNow.Millisecond;
```
bgavrilMS, line 167: Please expose `DateTimeOffset? RefreshOn` to be consistent with AuthenticationResult.ExpiresOn

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
bgavrilMS: We have a specific request now from @rymeskar to expose this, so definitely let's do it.

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+        NoCachedAT = 2,
+        /// <summary>
+        /// When the token request goes to ESTS because cached access token expired
+        /// </summary>
+        Expired = 3,
+        /// <summary>
+        /// When the token request goes to ESTS because refresh_in was used and the existing token needs to be refreshed
+        /// </summary>
+        RefreshIn = 4,
+        //NonMsal = 5, // Reserved for non-MSAL customers (included here for consistency with the spec)
```
bgavrilMS, line 31: pls delete this line.

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.;
+
+namespace Microsoft.Identity.Client.Cache
```
bgavrilMS, line 4: nit: Please move this file to `Microsoft.Identity.Client` namepsace

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
bgavrilMS: From a naming perspective, I think this is not intuitive. This enum tells the user the reason why MSAL had to contact AAD to get a token. So maybe the name should be more like `CacheMissReason` ?

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// Enum describing the reason for cache refresh and fetching access token from ESTS.
+    /// </summary>
+    public enum CacheInfo
+    {
+        /// <summary>
+        /// When the cache is not hit to make the request (interactive call, username password call, device code flow, etc.)
+        /// </summary>
+        None = 0,
+        /// <summary>
+        /// When the token request goes to ESTS because force_refresh was set to true
```
bgavrilMS, line 16: Let's not mention ESTS here, as it's an internal acronym for AAD. I think "Identity Provider" is the better term.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
bgavrilMS: I'm pretty sure ADFS doesn't implement refresh_in. 

for testing I would recommend focusing more on unit tests. 

1. add assertions to the tests in RefreshInTests
2. find some refresh RT tests that refresh the RT because the AT expired or because withforcerefresh is used.
3. etc.

trwalke: I tried this with the cloud account as well and it still didn't work. perhaps the policies are not setup correctly on the lab account

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -132,6 +132,7 @@
     <Compile Include="$(PathToMsalSources)\**\*.cs" Exclude="$(PathToMsalSources)\obj\**\*.*" />
     <Compile Remove="$(PathToMsalSources)\Platforms\**\*.*;$(PathToMsalSources)\Resources\*.cs" />
     <Compile Remove="$(PathToMsalSources)\PlatformsCommon\PlatformNotSupported\ApiConfig\SystemWebViewOptions.cs" />
+    <Compile Include="Cache\CacheMissReason.cs" />
```
pmaytak, line 135: I think this can be removed?

src/client/Microsoft.Identity.Client/Cache/CacheMissReason.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.;
```
pmaytak, line 2: nit: remove `;`

src/client/Microsoft.Identity.Client/Cache/CacheMissReason.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// </summary>
+        None = 0,
+        /// <summary>
+        /// When the token request goes to the identity provider because force_refresh was set to true
+        /// </summary>
+        ForceRefresh = 1,
+        /// <summary>
+        /// When the token request goes to the identity provider because no cached access token exists
+        /// </summary>
+        NoCachedAT = 2,
```
pmaytak, line 22: Since it's public now, maybe spell it out - `NoCachedToken`, `NoCachedAccessToken`?

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -39,5 +42,16 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
...
         public long DurationInHttpInMs { get; set; }
+
+        /// <summary>
+        /// Time, in milliseconds, remaining before the token will be proactively refreshed.
+        /// This value may be null.
+        /// </summary>
+        public DateTimeOffset? RemainingTimeBeforeRefresh { get; set; } = null;
+
+        /// <summary>
+        /// Enum describing the reason for cache refresh and fetching access token from ESTS.
```
pmaytak, line 53:
```suggestion
        /// Specifies the reason for fetching the access token from the identity provider.
```

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -39,5 +42,16 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
...
+        /// </summary>
+        public CacheMissReason CacheInfo { get; set; }
```
pmaytak, line 55: Rename the property name `CacheInfo` too?

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
pmaytak: Not sure if `CacheMiss` is accurate for all values. I think `CacheMiss` applies to `NoCachedAT` and `Expired`. `None` and `ForceRefresh` don't even check the cache and `ForceRefresh` returns a valid token.

> This enum tells the user the reason why MSAL had to contact AAD to get a token.

This describes it well but I'm not sure of a more accurate class name.

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -39,5 +42,16 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
         /// Time, in milliseconds, MSAL spent for HTTP communication during this request.
         /// </summary>
         public long DurationInHttpInMs { get; set; }
+
+        /// <summary>
+        /// Time, in milliseconds, remaining before the token will be proactively refreshed.
+        /// This value may be null.
+        /// </summary>
+        public DateTimeOffset? RemainingTimeBeforeRefresh { get; set; } = null;
```
bgavrilMS, line 50: It's no longer a duration, it's a DateTime. So maybe rename to `RefreshOn` ?

bgavrilMS (changes requested):
Minor naming changes + revert csproj

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -145,6 +145,13 @@ public partial class AuthenticationResult
                     tenantProfiles: tenantProfiles);
             }
 
+            UniqueId = msalIdTokenCacheItem?.IdToken?.GetUniqueId();
```
SameerK-MSFT, line 148: Minor: msalIdTokenCacheItem null is checked in 3 places and idToken in 2 places. It will not make much perf difference, but if you want it can be done this way.
if(msalIdTokenCacheItem != null)
{
IdToken = msalIdTokenCacheItem.Secret;
// you can use if(msalIdTokenCacheItem.IdToken)
UniqueId = msalIdTokenCacheItem.IdToken?.GetUniqueId();
TenantId = msalIdTokenCacheItem.IdToken?.TenantId;
}

src/client/Microsoft.Identity.Client/Cache/CacheMissReason.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Specifies the reason for fetching the access token from the identity provider.
+    /// </summary>
+    public enum CacheMissReason
+    {
+        /// <summary>
+        /// When the cache is not supposed to be hit to make the request (interactive call, username password call, device code flow, etc.)
+        /// </summary>
+        None = 0,
```
jmprieur, line 14: or NotApplicable?

src/client/Microsoft.Identity.Client/Cache/CacheMissReason.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// </summary>
+        NoCachedAT = 2,
+        /// <summary>
+        /// When the token request goes to the identity provider because cached access token expired
+        /// </summary>
+        Expired = 3,
+        /// <summary>
+        /// When the token request goes to the identity provider because refresh_in was used and the existing token needs to be refreshed
+        /// </summary>
+        RefreshIn = 4
```
jmprieur, line 30: ProactivelyRefreshed?

jmprieur (approved):
LGTM
Thanks @trwalke 
A few naming suggestions

src/client/Microsoft.Identity.Client/Cache/CacheInfo.cs
trwalke: Maybe `CacheUpdateReason`, `CacheUpdateHint` or `CacheUpdateIndication`
@pmaytak @bgavrilMS?

pmaytak: `TokenSourceReason`? We have `TokenSource` already (cache, IdP, Broker), and this enum will explain why it went to that source?

trwalke: The only problem with `TokenSourceReason` is it wont make sense when the `TokenSource` value is `Cache`. For example, if you have a `TokenSource` of `Cache` and the `TokenSourceReason` is `ProactivelyRefreshed`, it doesnt explain why the token came from the cache. However, it does explain why the token cache was updated.

This is why I suggested `CacheUpdateReason` as it is more general and makes sense with all of the enums

trwalke: The only problem with `TokenSourceReason` is it wont make sense when the `TokenSource` value is `Cache`. For example, if you have a `TokenSource` of `Cache` and the `TokenSourceReason` is `ProactivelyRefreshed`, it doesnt explain why the token came from the cache. However, it does explain why the token cache was refrshed.

I suggest we do `CacheRefreshReason` as it is more general and makes sense with all of the enums

pmaytak: > The only problem with `TokenSourceReason` is it wont make sense when the `TokenSource` value is `Cache`. For example, if you have a `TokenSource` of `Cache` and the `TokenSourceReason` is `ProactivelyRefreshed`, it doesnt explain why the token came from the cache. However, it does explain why the token cache was refrshed.
> 
> I suggest we do `CacheRefreshReason` as it is more general and makes sense with all of the enums
Yea, makes sense.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: Seems not to work without this. There is a configuration somewhere I am not aware of

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -39,5 +41,16 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
         /// Time, in milliseconds, MSAL spent for HTTP communication during this request.
         /// </summary>
         public long DurationInHttpInMs { get; set; }
+
+        /// <summary>
+        /// Time, in milliseconds, remaining before the token will be proactively refreshed.
+        /// This value may be null.
```
pmaytak, line 47:
```suggestion
        /// Specifies the time when the cached token should be proactively refreshed.
        /// This value may be null if proactive refresh is not enabled.
```
Not a 100% on the exact wording though.

pmaytak (approved):

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
trwalke: I think this correct

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Specifies the reason for fetching the access token from the identity provider.
+    /// </summary>
+    public enum CacheRefreshReason
+    {
+        /// <summary>
+        /// When the cache is not supposed to be hit to make the request (interactive call, username password call, device code flow, etc.)
```
bgavrilMS, line 12: nit: can we list all the flows / methods that hit the cache (so AcquireTokenSilent, AcquireTokenForClient and AcquireTokenOnBehalfOf)

tests/Microsoft.Identity.Test.Unit/MetricsTests.cs
```diff
@@ -102,7 +104,7 @@ public async Task MetricsUpdatedSucessfully_AcquireTokenInteractiveAndSilent_Asy
 
                 PublicClientApplication pca = CreatePca(harness.HttpManager);
                 await TestAcquireTokenInteractive_Async(pca, expectedTokensFromIdp: 1, expectedTokensFromCache: 0).ConfigureAwait(false);
-                await TestAcquireTokenSilent_Async(pca, expectedTokensFromIdp: 1, expectedTokensFromCache: 1).ConfigureAwait(false);
+                await TestAcquireTokenSilent_Async(pca, expectedTokensFromIdp: 1, expectedTokensFromCache: 1, 0, refreshIn: true).ConfigureAwait(false);
```
bgavrilMS, line 107: Can you please add this as separate test instead?

tests/Microsoft.Identity.Test.Unit/MetricsTests.cs
```diff
@@ -159,6 +167,23 @@ private async Task TestAcquireTokenSilent_Async(PublicClientApplication pca, int
             Assert.AreEqual(expectedTokensFromCache, Metrics.TotalAccessTokensFromCache);
             Assert.AreEqual(expectedTokensFromBroker, Metrics.TotalAccessTokensFromBroker);
             Assert.IsTrue(Metrics.TotalDurationInMs > 0);
+
+            if (refreshIn)
+            {
+                //Force Refresh In
+                var at = pca.UserTokenCacheInternal.Accessor.GetAllAccessTokens().Single();
```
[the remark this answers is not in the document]
bgavrilMS: Ok, so I had a look at the test and there seems to be a bug when we compute this field. Namely, if the token comes from the cache, the "reason" should be `None`, but instead it is currently `RefreshIn`. This is not a problem for telemetry, because we won't actually send it (no call to the /token endpoint).

Let's make sure (via tests) all 3 of our flows that hit the cache emit the correct flag.

bgavrilMS (changes requested):
There is a bug in the existing code that you surface - if the cache is hit, the "refresh reason" should be set to none. Instead, it is set to Refresh

PR #2928, opened 2021-10-02
Revert public API changes (IAppConfig change and Validate method exposure)

### LINKED ISSUE #2927: [Bug] Fix potentially breaking changes in 4.37
Most of the changes for 4.37 are now in, so I have produced a preview package. While inspecting for breaking changes via ApiReview tool, it looks like we have 1 breaking change and 1 API exposed which should not be exposed: 
 
 
Breaking change: we cannot ADD anything to interfaces as it will break existing implementation 
 Unintended API exposed: As per "framework design guidelines, <protected> is really the same as <public> in terms of visibility"
 

image

pmaytak:
Fixes #2927

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
@@ -372,7 +372,7 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// Validates the parameters of the AcquireToken operation.
         /// </summary>
-        protected virtual void Validate()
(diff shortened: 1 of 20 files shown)


bgavrilMS (approved):

bgavrilMS:
@pmaytak - I pushed a fix for the `protected Validate` stuff. Just made it `protected internal`. Should be enough.

SameerK-MSFT:
There is a change from "protected" to "protected internal". The later is actually protected or internal. (ref: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected-internal) So it will not reduce the exposure but actually increase it. Do we want that?

PR #2930, opened 2021-10-04
Cache options renaming + API comments

bgavrilMS:
Fixes # .

**Testing**

**Performance impact**


src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Options for MSAL token caches. 
+    /// </summary>
+    /// <remarks>
+    /// These option do not allow configuration of external cache serialization, for which you should use <see cref="TokenCache.SetAfterAccessAsync(System.Func{TokenCacheNotificationArgs, System.Threading.Tasks.Task})"/> and other callbacks.
```
jennyf19, line 10: I don't think the do it yourself path should be mentioned. We have Microsoft.Identity.Web.TokenCache for a reason. A link to the documentation should be enough.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// For detailed recommendations see: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore
+    /// </remarks>
+    public class CacheOptions
+    {
+        /// <summary>
+        /// Recommended options for using a static cache. 
+        /// </summary>
+        /// <remarks>
+        /// May include some eviction policies in the future to keep memory in check.
```
jennyf19, line 19: i would remove this remark about eviction policies.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Options for MSAL token caches. 
+    /// </summary>
+    /// <remarks>
+    /// These option do not allow configuration of external cache serialization, for which you should use <see cref="TokenCache.SetAfterAccessAsync(System.Func{TokenCacheNotificationArgs, System.Threading.Tasks.Task})"/> and other callbacks.
+    /// For detailed recommendations see: https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnetcore
+    /// </remarks>
+    public class CacheOptions
```
SameerK-MSFT, line 13: Suggestion to simplify the API. Use factory pattern. So users know what they create i.e. a static method to create options and all constructors as private. 
`public class CacheOptions
{
     public static CacheOptions CreateCacheOption(bool useSharedCache) {...}
     private CacheOptions(bool useSharedCache) {...}
     bool UseSharedCache { get; }
  
}`

SameerK-MSFT: Another way can be declare CacheOptions as enum. So it can be expanded after preview. and it will make API simpler.

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
bgavrilMS: We do have other options objects (e.g. SystemWebviewOptions) and they are all simple CRUD objects. So for consistency I'd keep it the same. Also, there is currently 1 option and we don't know how many more to come, but I estimate no more than 3.

pmaytak (approved):

SameerK-MSFT (approved):

PR #2953, opened 2021-10-15
Deprecate WithAuthority at app level and add WithTenantId

### LINKED ISSUE #2837: [Bug] NullReferenceException from MSAL when using region with CCA
**Logs and network traces**
CorrelationId == "a95592bb-f6c4-4f96-8e09-1ed652ec76fd"

4.35.1.0

**Platform**

* Desktop / Mobile
    * [ ] Interactive
    * [ ] Username Password
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 
    * [x] Service to Service calls

Other?

Upgrade to MSAL

**Repro**
Create CCA
```csharp
//Create CCA
var confidentialClientApplication = ConfidentialClientApplicationBuilder.Create(ClientId)
    .WithCertificate(Cert)
    .WithAzureRegion(msalRegion)
    .WithLegacyCacheCompatibility(false)
    .WithLogging(this.LogMSALMessages, Microsoft.Identity.Client.LogLevel.Warning)
    .Build();

//Acquire Token
var authenticationResult = await this.GetConfidentialClientApplication()
    .AcquireTokenForClient(scopes)
    .WithAuthority($"https://login.microsoft.com/{tenantId}")) // this is the root cause of the failure
    .WithSendX5C(sendX5c)
    .ExecuteAsync()
    .ConfigureAwait(false);
```

**Expected behavior**
MSAL should acquire a token

**Actual behavior**
Null Ref

System.NullReferenceException : Object reference not set to an instance of an object.
   at Microsoft.Identity.Client.Region.RegionManager.IsTelemetryRecorded(ApiEvent apiEvent)
   at Microsoft.Identity.Client.Region.RegionManager.RecordTelemetry(ApiEvent apiEvent, String azureRegionConfig, RegionInfo discoveredRegion)
 ...

Root cause
------------

MSAL tries to validate that the authority at the request level is the same as the authority at the app level. Since the hosts differ, it tries to do instance discovery to find out aliases. There is no api event, so this fails.

bgavrilMS:
Fixes #2837 

Important: this PR has a large number of changes, because of the deprecation of this API. Please follow my comments which will point you to the important changes.

- Deprecate request WithAuthority 
- Add WithTenantId

**Testing**
unit + integration

**Performance impact**
same number of string operations.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Identity.Client
         internal AcquireTokenCommonParameters CommonParameters { get; } = new AcquireTokenCommonParameters();
 
         /// <summary>
-        /// Executes the Token request asynchronously, with a possibility of cancelling the
+        /// Executes the Token request asynchronously, with a possibility of canceling the
         /// asynchronous method.
         /// </summary>
         /// <param name="cancellationToken">Cancellation token. See <see cref="CancellationToken"/> </param>
@@ -323,7 +323,7 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
-        /// Overrides the tenant Id specified at the application level. This operation preserves the authority host (environment).
+        /// Overrides the tenant ID specified  in the authority the application level. This operation preserves the authority host (environment).
         /// 
         /// If an authority was not specified at the application level, the default used is https://login.microsoftonline.com/common.
         /// </summary>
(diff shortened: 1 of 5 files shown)


bgavrilMS:
@jmprieur - sure, please take the time to review. Please let me know if you'd like me to walk you through the code. 

Aside from the deprecation of WithAuthority and adding of WithTenantId, this PR also: 

- leaves behind `WithB2CAuthority` and `WithAdfsAuthority` (so no change there). We can deprecate these 2 as well if needed, but there is no technical reason for them to disapear. 
- throw an exception if the conditions of the associated P1 bug are met: 

1. App points to "login.microsoftonline.com/common"
2. Request points to "login.windows.net/727123..." (same cloud, different env)
3. Regional is used. 

This would result in the P1 bug associated with the PR. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2837

PR #2964, opened 2021-10-19
Add WithTenantId option #2280

### LINKED ISSUE #2280: [Feature Request] Include `.WithTenantId()` in `AbstractAcquireTokenParameterBuilder`
In MS Identity Web, we include the tenant if it's provided by the customer when doing the acquire token calls, but it would be better to use `.WithTenantId()`, like we can on the cca builder. This would enable us to remove the string parsing & concatenation we currently have to do to include the tenant w/the authority. 

cc: @jmprieur 

---

In fact, `WithTenantId()` should be the only option allowed, as we do not allow changes of environment. So we should obsolete (via warning) `WithAuthority` at request level and ask ppl to use `WithTenantId`

bgavrilMS:
Fixes #2280 

- Add WithTenantId
- Add some new unit tests
- Repurpose some old unit tests to use WithTenantid
- Small refactoring to use Authority instead of AuthorityInfo at app level. This should result in a fewer string operations.

**Testing**
unit and integration

**Performance impact**
not tested, but the number of string operations should be the same.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
@@ -312,15 +312,17 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
-        /// Overrides the tenant ID specified  in the authority the application level. This operation preserves the authority host (environment).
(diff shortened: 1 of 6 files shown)


jennyf19:
I didn't have time to look through everything, but what i saw looks great. will be a much appreciated feature add.

PR #2970, opened 2021-10-20
Reduce allocations for access token cache retrieval for misses

johnthcall:
For cache miss scenarios there is no reason to allocate a new list only for followup calls to check whether there is any item in the list. For .Net Core/Standard and Framework versions after 4.6 we can use Array.Empty, however this does reduce readability. I completely understand if we want to keep the code as is for simplicity sake as this is not a major bottleneck.

**Testing**
Verified that the `Array.Empty<MsalAccessTokenCacheItem>() as IReadOnlyList<MsalAccessTokenCacheItem>` does not result in null.

**Performance impact**
Verified how much time is being spent on new List vs Array.Empty

|     Method |              Runtime |       Mean |     Error |    StdDev | Ratio | Allocated |
|----------- |--------------------- |-----------:|----------:|----------:|------:|----------:|
| ArrayEmpty |             .NET 5.0 |  0.5157 ns | 0.0944 ns | 0.1628 ns |  0.09 |         - |
| ArrayEmpty |             .NET 6.0 |  0.3559 ns | 0.0896 ns | 0.1546 ns |  0.07 |         - |
| ArrayEmpty | .NET Framework 4.7.2 |  5.4722 ns | 0.1890 ns | 0.3259 ns |  1.00 |         - |
|            |                      |            |           |           |       |           |
|    NewList |             .NET 5.0 |  4.5300 ns | 0.1699 ns | 0.4622 ns |  0.44 |      32 B |
|    NewList |             .NET 6.0 |  5.7998 ns | 0.1849 ns | 0.2987 ns |  0.55 |      32 B |
|    NewList | .NET Framework 4.7.2 | 10.6371 ns | 0.2900 ns | 0.5002 ns |  1.00 |      40 B |

Source
``` csharp
[ParamsSource(nameof(ValuesForInput))]
public IEnumerable<string> Input { get; set; }
public IEnumerable<IEnumerable<string>> ValuesForInput => new IEnumerable<string>[] { null };

[Benchmark]
public IReadOnlyList<string> ArrayEmpty()
{
    return Input?.ToList() ?? Array.Empty<string>() as IReadOnlyList<string>;
}
[Benchmark]
public IReadOnlyList<string> NewList()
{
    return Input?.ToList() ?? new List<string>();
}
```

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenInteractiveParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenInteractiveParameters.cs
@@ -7,6 +7,7 @@ using System.Text;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Extensibility;
 using Microsoft.Identity.Client.UI;
+using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.ApiConfig.Parameters
 {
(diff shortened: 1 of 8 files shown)


johnthcall:
> Thanks for looking into this @johnthcall . Would be good to apply this in more places. Happy to merge your PR into a branch and apply this fix in other areas. Is that ok?
> 
> PS: I gave you write access to the repo, you don't have to use a fork anymore if you don't want to.

I've added a Util and used it everywhere where an IEnumerable or IReadOnlyList was being returned. I just did the "Fast Build" and verified everything worked, let me know if you want any of the usages undone.

PR #2985, opened 2021-11-01
Added task to publish snupkg

SameerK-MSFT:
Fixes # .
Uploads snupackages to the internal symbol server

- Added another task to publish snupkg

**Testing**
Manual

**Performance impact**
None

### DIFF
--- a/build/template-publish-packages-and-symbols.yaml
+++ b/build/template-publish-packages-and-symbols.yaml
@@ -37,7 +37,8 @@ steps:
     SymbolServerType: TeamServices
 
 - task: PublishSymbols@2
-  displayName: 'Publish snupkg'
+  displayName: 'Publish PDBs'
   inputs:
-    SearchPattern: '**/bin/**/*.snupkg'
+    SearchPattern: '**/bin/**/*.pdb'
     SymbolServerType: TeamServices
+    TreatNotIndexedAsWarning: true
--- a/build/template-publish-packages-and-symbols.yaml
+++ b/build/template-publish-packages-and-symbols.yaml
@@ -36,8 +36,8 @@ steps:
     SearchPattern: '${{ parameters.SymbolPublishWildcard }}'
     SymbolServerType: TeamServices
 
-  - task: PublishSymbols@2
-    displayName: 'Publish snupkg'
-    inputs:
-      SearchPattern: '**/bin/**/*.snupkg'
-      SymbolServerType: TeamServices
+- task: PublishSymbols@2
+  displayName: 'Publish snupkg'
+  inputs:
+    SearchPattern: '**/bin/**/*.snupkg'
+    SymbolServerType: TeamServices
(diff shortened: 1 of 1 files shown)


pmaytak (commented):
Do we still need the PublishSymbols task above?

build/template-publish-packages-and-symbols.yaml
```diff
@@ -35,3 +35,9 @@ steps:
   inputs:
     SearchPattern: '${{ parameters.SymbolPublishWildcard }}'
     SymbolServerType: TeamServices
+
+- task: PublishSymbols@2
```
pmaytak, line 39: Do we still need the PublishSymbols task above?

trwalke:
Should probably verify with the customer that this is working

build/template-publish-packages-and-symbols.yaml
SameerK-MSFT: Yes. The two tasks support two different types of files

bgavrilMS (approved):

build/template-publish-packages-and-symbols.yaml
```diff
@@ -35,3 +35,10 @@ steps:
...
+  displayName: 'Publish PDBs'
+  inputs:
+    SearchPattern: '**/bin/**/*.pdb'
```
gladjohn, line 42: can we add this to the parameters?

gladjohn (approved):

build/template-publish-packages-and-symbols.yaml
bgavrilMS: Yes, Gladwin is right. The search pattern here is too broad, and it will include tons of PDB files such as the ones that come from testing DLLs. The search pattern should be similar to  '**/bin/**/microsoft.identity.client.*'

bgavrilMS (changes requested):
Search pattern too broad

build/template-publish-packages-and-symbols.yaml
trwalke: we have a similar search pattern on line 7
`SymbolPublishWildcard: '**/bin/**/microsoft.identity.client.*'`

SameerK-MSFT: Travis is right about having the pattern. But it does not publish pdbs. I do not know the internals; but it looks like they are skipped by default. If they would publish everything qualified by "microsoft.identity.client.*", there will be too many files getting published. But again that is my hypothesis and I have not looked at the code.

pmaytak: Currently (for the current pattern) in the publish symbols task logs it does say it's uploading and succeeding processing PDBs and DLLs for both MSAL and MSAL.Desktop. So maybe then it doesn't upload into the correct location or something?

trwalke (changes requested):
Reconfigure search pattern to target source directories only

bgavrilMS:
I guess this didn't work. Should we close it?

Issue #2848, opened 2021-08-26
[Feature Request] Disable cache syncronization for confidential client apps


bgavrilMS:
WithEnableCacheSynchronization is set to true by default, but can be set to false for a perf improvement on S2S / AppTokenCache.

bgavrilMS:
I don't think so. The locking would only affect the AppTokenCache.

jmprieur:
@bgavrilMS I realize i wasn't clear. I meant.
- app is deployed on several deployment slots.
- each deployed code has the same client ID and goes to the same distributed redis app token cache

bgavrilMS:
This is a great question. If machine A and machine B compete for tokens, there is no way for MSAL to reconcile these anyway. The locking mechanism is in process and per application object only. The logic of AcquireToken* calls is: 

```
Lock
Read cache from Redis in memory
Unlock

If no token is found in the cache, get one from AAD

Lock
Read cache from Redis in memory
Merge new token
Write cache from memory to Redis
Unlock
```

By removing the locks, we can no longer guarantee the order of operations. I think this is fine for AcquireTokenForClient.

For the other scenarios, the problem I see is: 

- a Remove(Account) could potentially be missed
- a more powerful RT (with MFA) can be overwritten by a less powerful RT

Then again, these sort of problems are best left to eventual consistency. Any ideas on how we could experiment disabling syncronization for user token cache?

maliksahil:
You can do distributed locks with Redis. Basically it is built around establishing a lock with an auto-release (see redlock algorithm) https://redis.io/topics/distlock

PR #2988, opened 2021-11-02
Cache synchronization set to false for CCA by default

trwalke:
Fixes # .
fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2848

Cache synchronization set to false for CCA by default

**Testing**
Unit test

**Performance impact**
N/A

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Identity.Client
         internal int ConfidentialClientCredentialCount;
 
         public bool LegacyCacheCompatibilityEnabled { get; internal set; } = true;
-        public bool? CacheSynchronizationEnabled { get; internal set; } = null;
+        public bool CacheSynchronizationEnabled { get; internal set; } = true;
 
 
         #region Region
--- a/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Identity.Client
         internal int ConfidentialClientCredentialCount;
 
         public bool LegacyCacheCompatibilityEnabled { get; internal set; } = true;
-        public bool CacheSynchronizationEnabled { get; internal set; } = true;
+        public bool? CacheSynchronizationEnabled { get; internal set; } = null;
 
 
         #region Region
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -287,7 +287,7 @@ namespace Microsoft.Identity.Client
         /// when ConfidentialClientApplication objects are reused.
         /// </summary>
         /// <remarks>
-        /// True by default, but subject to change.
+        /// False by default.
         /// Not recommended for apps that call RemoveAsync
         /// </remarks>
         public ConfidentialClientApplicationBuilder WithCacheSynchronization(bool enableCacheSynchronization)
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -74,7 +74,9 @@ namespace Microsoft.Identity.Client
             ConfidentialClientApplication.GuardMobileFrameworks();
 
             var config = new ApplicationConfiguration();
-            return new ConfidentialClientApplicationBuilder(config).WithClientId(clientId);
+            return new ConfidentialClientApplicationBuilder(config)
+                .WithClientId(clientId)
+                .WithCacheSynchronization(false);
         }
 
         /// <summary>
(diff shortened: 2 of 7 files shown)


src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -103,7 +103,7 @@ public string ClientVersion
         internal int ConfidentialClientCredentialCount;
 
         public bool LegacyCacheCompatibilityEnabled { get; internal set; } = true;
-        public bool CacheSynchronizationEnabled { get; internal set; } = true;
+        public bool? CacheSynchronizationEnabled { get; internal set; } = null;
```
pmaytak, line 106: Hmm, wonder if we can simplify the logic? 
- Keep this line the same - normal `bool` and default to `true`. Since PCA defaults to `true` and there's no way to change the setting neither via PCA options nor via `With` method.
- CCAOptions defaults `EnableCacheSynchronization` to `false` and CCA constructor sets `CacheSynchronizationEnabled` to `false`.
- User can call `WithCacheSynchronization(true)` to change this back, if they want.
- And then we can keep the `TokenCache` as is.

pmaytak (changes requested):

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
trwalke: I was under the impression that the constructor for CCA would be called after the user calls `WithCacheSynchronization(true)` which would overwrite it, setting it back to false. ill double check

trwalke: Yeah, The CCA constructor is called from BuildConcrete() which happens after the options are configured. i could add it to the `ConfidentialClientApplicationBuilder Create()` method however

pmaytak: Hmm, yea, would it make sense to chain`.WithCacheSynchronization(false);` in CCA.Create?

trwalke: Yeah, seems like a good way to do it

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -1,340 +1,341 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
-using System;
```
bgavrilMS, line 4: What are your spacing settings in Visual Studio? You should be using the defaults - 4 spaces. 

I think when we make changes via GitHub (via those suggestions), it messes up spacing and makes this impossible to review.

bgavrilMS: Anyway, if you see this happening, I think the best is to create a commit that changes the spacing and then another commit with actual changes, to help reviewers.

tests/Microsoft.Identity.Test.Unit/AppConfigTests/ConfidentialClientApplicationBuilderTests.cs
```diff
@@ -103,7 +103,7 @@ public void CacheSynchronizationWithDefault()
                 ClientId = TestConstants.ClientId,
             };
             var app = ConfidentialClientApplicationBuilder.CreateWithApplicationOptions(options).Build();
```
bgavrilMS, line 105: Can you add a PubliClient version here as well? i.e. create a PCA using options and make sure `CacheSynchronizationEnabled` is set to true.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -1,340 +1,341 @@
...
+        }
+
+
+        /// <summary>
+        /// When set to <c>true</c>, MSAL will lock cache access at the <see cref="ConfidentialClientApplication"/> level, i.e.
+        /// the block of code between BeforeAccessAsync and AfterAccessAsync callbacks will be synchronized. 
+        /// Apps can set this flag to <c>false</c> to enable an optimistic cache locking strategy, which may result in better performance, especially 
+        /// when ConfidentialClientApplication objects are reused.
+        /// </summary>
+        /// <remarks>
```
bgavrilMS, line 289: Please update the comments.

bgavrilMS:
@trwalke - this was reviewed a long time ago. I see that you made some changes but it's very hard to review because the spacing changed. I left some comments, but would like @pmaytak to sign off.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
pmaytak: Both files seem to have 4-space tabs. This might be the Mac/Linux vs Windows new line encodings?

pmaytak:
I'll re-review once the spacing/new lines is updated.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationOptions.cs
```diff
@@ -43,6 +43,6 @@ public class ConfidentialClientApplicationOptions : ApplicationOptions
         /// True by default, but subject to change.
```
bgavrilMS, line 43: Comment update

bgavrilMS:
LGTM, 1 comment needs to be updated.

bgavrilMS (approved):
1 comment needs update

pmaytak (approved):

Issue #2891, opened 2021-09-15
[Bug][Internal] Bad test ValidateKeyExpirationAsync


bgavrilMS:
ValidateKeyExpirationAsync does not actually test that the POP keys are rotated every 8 hours. 

The test has 2 problems: 

1. It mocks the AAD response badly - client creds should not get back RT, IDT etc. This confuses MSAL and causes cache misses.
2. The test should make sure the same token is retrieved from the cache after 7 hours. Instead, it makes a call to AAD.

The first problem has been fixed in pmaytak/partitioned-usercache

bgavrilMS:
Was fixed.

PR #2994, opened 2021-11-05
Updating POP Token Test

trwalke:
Fixes # .
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2891

Adding additional check to pop token

**Testing**
Unit test

**Performance impact**
N/A

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Net.Http;
-using System.Net.Http.Headers;
 using System.Security.Cryptography;
 using System.Text;
 using Microsoft.Identity.Client.AppConfig;
@@ -22,6 +20,7 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
     {
         private static readonly DateTime s_jwtBaselineTime = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
         private readonly PoPAuthenticationConfiguration _popAuthenticationConfiguration;
+        private IPoPCryptoProvider _popCryptoProvider;
 
         /// <summary>
         /// Creates POP tokens, i.e. tokens that are bound to an HTTP request and are digitally signed.
@@ -39,9 +38,9 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
 
             _popAuthenticationConfiguration = popAuthenticationConfiguration ?? throw new ArgumentNullException(nameof(popAuthenticationConfiguration));
 
-            _popAuthenticationConfiguration.PopCryptoProvider = _popAuthenticationConfiguration.PopCryptoProvider ?? serviceBundle.PlatformProxy.GetDefaultPoPCryptoProvider();
+            _popCryptoProvider = _popAuthenticationConfiguration.PopCryptoProvider ?? serviceBundle.PlatformProxy.GetDefaultPoPCryptoProvider();
 
-            var keyThumbprint = ComputeThumbprint(_popAuthenticationConfiguration.PopCryptoProvider.CannonicalPublicKeyJwk);
+            var keyThumbprint = ComputeThumbprint(_popCryptoProvider.CannonicalPublicKeyJwk);
             KeyId = Base64UrlHelpers.Encode(keyThumbprint);
         }
 
@@ -67,7 +66,7 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
         {
             JObject header = new JObject
             {
-                { JsonWebTokenConstants.ReservedHeaderParameters.Algorithm, _popAuthenticationConfiguration.PopCryptoProvider.CryptographicAlgorithm },
+                { JsonWebTokenConstants.ReservedHeaderParameters.Algorithm, _popCryptoProvider.CryptographicAlgorithm },
                 { JsonWebTokenConstants.ReservedHeaderParameters.KeyId, KeyId },
                 { JsonWebTokenConstants.ReservedHeaderParameters.Type, PoPRequestParameters.PoPTokenType}
             };
@@ -80,7 +79,7 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
 
         private JObject CreateBody(MsalAccessTokenCacheItem msalAccessTokenCacheItem)
         {
-            JToken publicKeyJWK = JToken.Parse(_popAuthenticationConfiguration.PopCryptoProvider.CannonicalPublicKeyJwk);
+            JToken publicKeyJWK = JToken.Parse(_popCryptoProvider.CannonicalPublicKeyJwk);
             List<JProperty> properties = new List<JProperty>(8);
             
             // Mandatory parameters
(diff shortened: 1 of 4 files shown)


tests/Microsoft.Identity.Test.Unit/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -129,35 +128,57 @@ public async Task ValidateKeyExpirationAsync()
...
 
                 provider = PoPProviderFactory.GetOrCreateProvider(testClock);
-                await app.AcquireTokenForClient(TestConstants.s_scope)
+                result = await app.AcquireTokenForClient(TestConstants.s_scope)
                     .WithProofOfPossession(popConfig)
                     .ExecuteAsync(CancellationToken.None)
                     .ConfigureAwait(false);
 
                 Assert.IsTrue(JWK != provider.CannonicalPublicKeyJwk);
+                Assert.IsTrue(result.AuthenticationResultMetadata.TokenSource == TokenSource.Cache);
```
trwalke, line 168: I was thinking about shat should happen here. After the key rotates and another call is made to the token cache, the same token is returned but the key is updated. i would imagine the key rotating should trigger a new call to the token endpoint right?

@bgavrilMS

pmaytak: Is this the filter (key Id?)?
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f9083e227400c75054ba594256ce1246092d8f4c/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L652

Makes sense - token with expired key ID should be treated like a normal expired token.

bgavrilMS: Yes, correct, if the key rotates, the at_cnf in the cache no longer matches it. So the `FindAccessToken` cache operation correctly results in a cache miss and MSAL makes another request to the /token endpoint.

bgavrilMS: So the `TokenSource` should be TokenSource.IdentityProvider here. 
In fact MSAL does the right thing, but this test fails to correctly simulate the fact that time has passed.

tests/Microsoft.Identity.Test.Unit/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -129,35 +128,51 @@ public async Task ValidateKeyExpirationAsync()
...
             }
         }
 
+        private string GetAccessTokenFromPopToken(string popToken)
```
bgavrilMS, line 166: The name of this method does not describe what it does. It extracts the modulus, or `n` parameter from the POP token. The modulus is part of the `public key` in an RSA public / private key pair, alongside the exponent `e`. In practical terms, `e` is always the same and the `n` is different each time.

bgavrilMS (changes requested):
The tokenSource at line 162 should be IdentityProvider, not cache.

tests/Microsoft.Identity.Test.Unit/pop/PopAuthenticationSchemeTests.cs
trwalke: Yes i see, I am grabbing the wrong value here, it should be getting the at

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -39,9 +38,9 @@ public PoPAuthenticationScheme(PoPAuthenticationConfiguration popAuthenticationC
 
             _popAuthenticationConfiguration = popAuthenticationConfiguration ?? throw new ArgumentNullException(nameof(popAuthenticationConfiguration));
 
-            _popAuthenticationConfiguration.PopCryptoProvider = _popAuthenticationConfiguration.PopCryptoProvider ?? serviceBundle.PlatformProxy.GetDefaultPoPCryptoProvider();
+            _popCryptoProvider = _popAuthenticationConfiguration.PopCryptoProvider ?? serviceBundle.PlatformProxy.GetDefaultPoPCryptoProvider();
```
trwalke, line 41: Turns out there was actually a bug here for the default provider. Once the '_popAuthenticationConfiguration.PopCryptoProvider' was set with the default provider, the 'GetDefaultPoPCryptoProvider()' would never be called again for the application instance and the `keyID` would not update.

@bgavrilMS

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPProviderFactory.cs
```diff
@@ -16,22 +16,24 @@ internal class PoPProviderFactory
         private static readonly TimeSpan s_expirationTimespan = TimeSpan.FromHours(8);
         private static object s_lock = new object();
 
-        public static InMemoryCryptoProvider GetOrCreateProvider(/* for testing */ ITimeService timeService = null)
-        {
-            if (timeService == null)
-            {
-                timeService = new TimeService();
-            }
+        internal static ITimeService TimeService { get; set; }
```
bgavrilMS, line 19: Small perf improvement - initialize this property with `new TimeService` instead. Tests will simply override it. This will avoid you having to check if it's null at line 25 below.

tests/Microsoft.Identity.Test.Unit/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -104,60 +104,93 @@ public void ValidatePopRequestAndToken()
...
                     .ConfigureAwait(false);
 
+                Assert.AreEqual(GetAccessTokenFromPopToken(result.AccessToken), GetAccessTokenFromPopToken(initialToken));
+                Assert.AreEqual(GetModulusFromPopToken(result.AccessToken), GetModulusFromPopToken(initialToken));
                 Assert.IsTrue(JWK == provider.CannonicalPublicKeyJwk);
+                Assert.IsTrue(result.AuthenticationResultMetadata.TokenSource == TokenSource.Cache);
+
                 //Advance time 2 hours. Should be a different key
-                testClock.TestTime = testClock.TestTime + TimeSpan.FromSeconds(60 * 60 * 2);
+                testClock.TestTime = testClock.TestTime + TimeSpan.FromHours(2);
```
bgavrilMS, line 161: nit: +=

tests/Microsoft.Identity.Test.Unit/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -104,60 +104,93 @@ public void ValidatePopRequestAndToken()
...
-                //Advance time 7 hours. Should still be the same key
-                testClock.TestTime = testClock.TestTime + TimeSpan.FromSeconds(60 * 60 * 7);
 
-                provider = PoPProviderFactory.GetOrCreateProvider(testClock);
-                await app.AcquireTokenForClient(TestConstants.s_scope)
+                //Advance time 7 hours. Should still be the same key and token
+                testClock.TestTime = testClock.TestTime + TimeSpan.FromHours(7);
+
+                PoPProviderFactory.TimeService = testClock;
+                provider = PoPProviderFactory.GetOrCreateProvider();
```
bgavrilMS, line 149: Can you remove this from the test please? To make sure there aren't any side effects. It's ok not to assert the JWK.

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
bgavrilMS: Yeah, good unit testing has the tendency to spot issues.

bgavrilMS (approved):
LGTM, some small comments only

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
pmaytak: @trwalke @bgavrilMS This is a bug in non-test related code right? Should we create a separate issue for this to reference in release notes?

trwalke: @pmaytak makes sense

pmaytak (approved):

PR #2997, opened 2021-11-09
Throws MSALClientException when back is pressed

SameerK-MSFT:
Fixes 2991 .

URL parameters are analyzed to check if it is user cancelled and corresponding result is returned.

**Testing**
Unit Test

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/UI/AuthorizationResult.cs
+++ b/src/client/Microsoft.Identity.Client/UI/AuthorizationResult.cs
@@ -73,6 +73,11 @@ namespace Microsoft.Identity.Client.UI
         {
             if (parameters.ContainsKey(TokenResponseClaim.Error))
             {
+                if (parameters.ContainsKey(TokenResponseClaim.ErrorSubcode) && parameters[TokenResponseClaim.ErrorSubcode].Equals(TokenResponseClaim.ErrorSubcodeCancel, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/UI/AuthorizationResult.cs
```diff
@@ -73,6 +73,11 @@ private static AuthorizationResult FromParsedValues(Dictionary<string, string> p
         {
             if (parameters.ContainsKey(TokenResponseClaim.Error))
             {
+                if (parameters.ContainsKey(TokenResponseClaim.ErrorSubcode) && parameters[TokenResponseClaim.ErrorSubcode].Equals(TokenResponseClaim.ErrorSubcodeCancel, StringComparison.OrdinalIgnoreCase))
```
bgavrilMS, line 76: `if (parameters.TryGetValue(key, out val)`

bgavrilMS (approved):
Small suggestion

src/client/Microsoft.Identity.Client/UI/AuthorizationResult.cs
SameerK-MSFT: Good catch. Fixed.

bgavrilMS (approved):

jmprieur (approved):
LGTM
Thanks @SameerK-MSFT

Issue #2996, opened 2021-11-09
[Feature Request] Provide a way to debug dump the sanitized cache


bgavrilMS:
The MSAL team sometimes need to look in the token cache to understand bugs reported by app developers. This is not always possible because of compliance issues. To workaround it, MSAL should have a mechanism to sanitize the cache and dump it. 

To achieve this: 

- [ ] On SaveTokenResponseAsync, if verbose logging is enabled and if token cache serialization is enabled, dump the first 10 **keys** of MsalAccessTokenCacheItem. Also dump the total number of AT, RT, IdTs and Accounts.
- [ ] If PII is enabled, do not log the home_account_id part of the key

jmprieur:
I also wonder if having just the number of elements in each collection could help in some cases?

bgavrilMS:
That we could log as info every time. And if there are many items, we can log as warning.

bgavrilMS:
This was done. We are displaying the first 10 access tokens and refresh tokens that are found in the cache when logging is set to verbose mode.

PR #2999, opened 2021-11-10
Cache Dump

trwalke:
Fixes # 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2996

Dumping cache info

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs
@@ -78,6 +78,18 @@ namespace Microsoft.Identity.Client.Cache.Keys
                 _extraKeyParts);
         }
 
+        public string ToScrubbedString()
+        {
+            return MsalCacheKeys.GetCredentialKey(
+                "****",
+                _environment,
+                _credentialDescriptor,
+                _clientId,
+                _tenantId,
+                _normalizedScopes,
+                _extraKeyParts);
+        }
+
         #region iOS
 
         public string iOSAccount => MsalCacheKeys.GetiOSAccountKey(_homeAccountId, _environment);
--- a/src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Identity.Client.Cache.Keys
         public string ToLogString(bool piiEnabled = false)
         {
             return MsalCacheKeys.GetCredentialKey(
-                piiEnabled? _homeAccountId : _homeAccountId.GetHashCode().ToString(),
+                piiEnabled? _homeAccountId : _homeAccountId?.GetHashCode().ToString(),
                 _environment,
                 _credentialDescriptor,
                 _clientId,
--- a/src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs
@@ -78,10 +78,10 @@ namespace Microsoft.Identity.Client.Cache.Keys
                 _extraKeyParts);
         }
 
-        public string ToScrubbedString()
+        public string ToLogString(bool piiEnabled = false)
         {
             return MsalCacheKeys.GetCredentialKey(
-                "****",
+                piiEnabled? _homeAccountId : _homeAccountId.GetHashCode().ToString(),
                 _environment,
                 _credentialDescriptor,
                 _clientId,
--- a/src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Identity.Client.Internal.Logger
     {
         private readonly IPlatformLogger _platformLogger;
         private readonly LogCallback _loggingCallback;
-        private readonly LogLevel _minLogLevel;
         private readonly bool _isDefaultPlatformLoggingEnabled;
         private static readonly Lazy<ICoreLogger> s_nullLogger = new Lazy<ICoreLogger>(() => new NullLogger());
 
@@ -34,7 +33,7 @@ namespace Microsoft.Identity.Client.Internal.Logger
                     : " - " + correlationId;
             PiiLoggingEnabled = enablePiiLogging;
             _loggingCallback = loggingCallback;
-            _minLogLevel = logLevel;
+            LogLevel = logLevel;
             _isDefaultPlatformLoggingEnabled = isDefaultPlatformLoggingEnabled;
 
             _platformLogger = PlatformProxyFactory.CreatePlatformProxy(null).PlatformLogger;
@@ -81,6 +80,8 @@ namespace Microsoft.Identity.Client.Internal.Logger
         public string ClientVersion { get; }
 
         internal string ClientInformation { get; }
+
+        internal LogLevel LogLevel { get; }
 
         public void Info(string messageScrubbed)
         {
--- a/src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Identity.Client.Internal.Logger
     {
         private readonly IPlatformLogger _platformLogger;
         private readonly LogCallback _loggingCallback;
+        private readonly LogLevel _minLogLevel;
         private readonly bool _isDefaultPlatformLoggingEnabled;
         private static readonly Lazy<ICoreLogger> s_nullLogger = new Lazy<ICoreLogger>(() => new NullLogger());
 
@@ -33,7 +34,7 @@ namespace Microsoft.Identity.Client.Internal.Logger
                     : " - " + correlationId;
             PiiLoggingEnabled = enablePiiLogging;
             _loggingCallback = loggingCallback;
-            LogLevel = logLevel;
+            _minLogLevel = logLevel;
             _isDefaultPlatformLoggingEnabled = isDefaultPlatformLoggingEnabled;
 
             _platformLogger = PlatformProxyFactory.CreatePlatformProxy(null).PlatformLogger;
@@ -80,8 +81,6 @@ namespace Microsoft.Identity.Client.Internal.Logger
         public string ClientVersion { get; }
 
         internal string ClientInformation { get; }
-
-        internal LogLevel LogLevel { get; }
 
         public void Info(string messageScrubbed)
         {
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -237,6 +237,36 @@ namespace Microsoft.Identity.Client
                         Stopwatch sw = Stopwatch.StartNew();
                         await tokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);
                         requestParams.RequestContext.ApiEvent.DurationInCacheInMs += sw.ElapsedMilliseconds;
+
+                        if(requestParams.RequestContext.Logger.LogLevel == LogLevel.Verbose)
+                        {
+                            IEnumerable<MsalAccessTokenCacheItem> accessTokenCacheItems = GetAllAccessTokensWithNoLocks(filterByClientId: false);
+                            IEnumerable<MsalAccessTokenCacheKey> accessTokenCacheKeys = accessTokenCacheItems.Select(item => item.GetKey());
+                            IReadOnlyList<MsalRefreshTokenCacheItem> refreshTokenCacheItems = GetAllRefreshTokensWithNoLocks(filterByClientId: false);
+                            IReadOnlyList<MsalIdTokenCacheItem> idTokenCacheItems = GetAllIdTokensWithNoLocks(filterByClientId: false, partitionKey: null);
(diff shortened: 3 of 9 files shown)


src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
```diff
@@ -34,7 +33,7 @@ internal class MsalLogger : ICoreLogger
                     : " - " + correlationId;
             PiiLoggingEnabled = enablePiiLogging;
             _loggingCallback = loggingCallback;
-            _minLogLevel = logLevel;
+            LogLevel = logLevel;
```
bgavrilMS, line 36: I don't think you need to make any changes to this class. Try to use the existing method `IsLoggingEnabled`

trwalke: ok good, updated

bgavrilMS (changes requested):
There are many comments that are unresolved from the time this was a draft PR

tests/Microsoft.Identity.Test.Unit/CacheTests/TokenCacheTests.cs
```diff
@@ -1272,6 +1272,56 @@ public void TestIsFociMember_EnvAlias()
...
+                    serviceBundle.ApplicationLogger,
+                    serviceBundle.PlatformProxy.CryptographyManager,
+                    Guid.NewGuid().AsMatsCorrelationId());
+
+                var response = TokenCacheHelper.CreateMsalTokenResponse();
+
+                await cache.SaveTokenResponseAsync(requestParams, response).ConfigureAwait(false);
+
+                Assert.IsTrue(dump != string.Empty);
+                Assert.IsTrue(dump.Contains("Total number of access tokens in cache: 10"));
```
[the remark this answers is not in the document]
trwalke: Yes

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -252,6 +256,32 @@ public sealed partial class TokenCache : ITokenCacheInternal
             }
         }
 
+        //This method pulls all of the acceess and refresh tokens from the cache and can therefore be very impactful on performance.
+        private void DumpCacheToLogs(AuthenticationRequestParameters requestParameters)
+        {
+            if (requestParameters.RequestContext.Logger.IsLoggingEnabled(LogLevel.Verbose))
+            {
+                IEnumerable<MsalAccessTokenCacheItem> accessTokenCacheItems = GetAllAccessTokensWithNoLocks(filterByClientId: false);
```
bgavrilMS, line 264: Use `IReadOnlyList` here as well. And `Count` property instead of `Count()`.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -252,6 +256,32 @@ public sealed partial class TokenCache : ITokenCacheInternal
...
+                IReadOnlyList<MsalRefreshTokenCacheItem> refreshTokenCacheItems = GetAllRefreshTokensWithNoLocks(filterByClientId: false);
+                IEnumerable<MsalAccessTokenCacheKey> accessTokenCacheKeys = accessTokenCacheItems.Select(item => item.GetKey());
```
bgavrilMS, line 266: Add a `.ToList()` and use the List instead of IEnumearble. To avoid enumerating multiple times.

bgavrilMS (approved):
Look good, some comments around using List instead of IEnumerable

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -237,6 +238,9 @@ public sealed partial class TokenCache : ITokenCacheInternal
                         Stopwatch sw = Stopwatch.StartNew();
                         await tokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);
                         requestParams.RequestContext.ApiEvent.DurationInCacheInMs += sw.ElapsedMilliseconds;
+
+                        DumpCacheToLogs(requestParams);
```
SameerK-MSFT, line 242: Please do it on a background thread

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -252,6 +256,32 @@ public sealed partial class TokenCache : ITokenCacheInternal
             }
         }
 
+        //This method pulls all of the acceess and refresh tokens from the cache and can therefore be very impactful on performance.
+        private void DumpCacheToLogs(AuthenticationRequestParameters requestParameters)
+        {
+            if (requestParameters.RequestContext.Logger.IsLoggingEnabled(LogLevel.Verbose))
+            {
+                IReadOnlyList<MsalAccessTokenCacheItem> accessTokenCacheItems = GetAllAccessTokensWithNoLocks(filterByClientId: false);
+                IReadOnlyList<MsalRefreshTokenCacheItem> refreshTokenCacheItems = GetAllRefreshTokensWithNoLocks(filterByClientId: false);
```
SameerK-MSFT, line 265: If each API perf issue, then it is better to fetch them in parallel and wait till both the tasks are done. It can spread them on two cores. (note: useful only if both are taking long time. Else synchronization may erase the performance.)

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -252,6 +256,32 @@ public sealed partial class TokenCache : ITokenCacheInternal
...
+                IList<MsalAccessTokenCacheKey> accessTokenCacheKeys = accessTokenCacheItems.Select(item => item.GetKey()).ToList();
+
+                requestParameters.RequestContext.Logger.Verbose($"Total number of access tokens in cache: {accessTokenCacheItems.Count()}");
```
SameerK-MSFT, line 268: You can make it part of StringBuilder to log in one call. IMO, it will save time.

SameerK-MSFT (changes requested):

tests/Microsoft.Identity.Test.Common/Core/Mocks/TokenCacheHelper.cs
```diff
@@ -199,9 +216,37 @@ internal static class TokenCacheHelper
...
-        internal static (MsalAccessTokenCacheItem, MsalRefreshTokenCacheItem, MsalIdTokenCacheItem, MsalAccountCacheItem) PopulateCacheWithOneAccessToken(ITokenCacheAccessor accessor)
+        internal static IEnumerable<Tuple<MsalAccessTokenCacheItem, 
+                                          MsalRefreshTokenCacheItem, 
+                                          MsalIdTokenCacheItem, 
+                                          MsalAccountCacheItem>> PopulateCacheWithAccessTokens(ITokenCacheAccessor accessor, int tokensQuantity = 1)
+        {
+            IList<Tuple<MsalAccessTokenCacheItem, MsalRefreshTokenCacheItem, MsalIdTokenCacheItem, MsalAccountCacheItem>> tokens 
+                                        = new List<Tuple<MsalAccessTokenCacheItem, MsalRefreshTokenCacheItem, MsalIdTokenCacheItem, MsalAccountCacheItem>>();
+
+            bool randomizeClientInfo = tokensQuantity > 1 ? true : false;
```
pmaytak, line 227: nit: 
```suggestion
            bool randomizeClientInfo = tokensQuantity > 1;
```

tests/Microsoft.Identity.Test.Common/Core/Mocks/TokenCacheHelper.cs
```diff
@@ -199,9 +216,37 @@ internal static class TokenCacheHelper
...
+                              MsalIdTokenCacheItem, 
+                              MsalAccountCacheItem>(result.Item1, result.Item2, result.Item3, result.Item4);
+
+                tokens.Add(token);
+            }
+
+            return tokens;
+        }
+
+        internal static (MsalAccessTokenCacheItem, MsalRefreshTokenCacheItem, MsalIdTokenCacheItem, MsalAccountCacheItem) PopulateCacheWithOneAccessToken(ITokenCacheAccessor accessor, bool randomizeClientInfo = false)
```
pmaytak, line 244: nit: params can be named:
`internal static (MsalAccessTokenCacheItem AT, MsalRefreshTokenCacheItem RT, MsalIdTokenCacheItem IDT, MsalAccountCacheItem Acc) PopulateCacheWithOneAccessToken(ITokenCacheAccessor accessor, bool randomizeClientInfo = false)`

`MsalAccountCacheItem>(result.AT, result.RT, result.IDT, result.Acc);`

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -252,6 +256,32 @@ public sealed partial class TokenCache : ITokenCacheInternal
...
         }
 
+        //This method pulls all of the acceess and refresh tokens from the cache and can therefore be very impactful on performance.
+        private void DumpCacheToLogs(AuthenticationRequestParameters requestParameters)
+        {
+            if (requestParameters.RequestContext.Logger.IsLoggingEnabled(LogLevel.Verbose))
+            {
+                IReadOnlyList<MsalAccessTokenCacheItem> accessTokenCacheItems = GetAllAccessTokensWithNoLocks(filterByClientId: false);
+                IReadOnlyList<MsalRefreshTokenCacheItem> refreshTokenCacheItems = GetAllRefreshTokensWithNoLocks(filterByClientId: false);
+                IList<MsalAccessTokenCacheKey> accessTokenCacheKeys = accessTokenCacheItems.Select(item => item.GetKey()).ToList();
```
pmaytak, line 266: Add `Take(10)`? I think if the source is less than 10 items, it'll still work. Then just foreach through `accessTokenCacheKeys`.
```suggestion
                IList<MsalAccessTokenCacheKey> accessTokenCacheKeys = accessTokenCacheItems.Take(10).Select(item => item.GetKey()).ToList();
```

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -252,6 +256,32 @@ public sealed partial class TokenCache : ITokenCacheInternal
...
+
+                requestParameters.RequestContext.Logger.Verbose($"Total number of access tokens in cache: {accessTokenCacheItems.Count()}");
+                requestParameters.RequestContext.Logger.Verbose($"Total number of refresh tokens in cache: {refreshTokenCacheItems.Count()}");
```
pmaytak, line 269: Use property `Count` here and above.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: @bgavrilMS Do you know, is ReadOnlyList faster than IList?

pmaytak (approved):
A few perf refactoring suggestions.

Biggest concern top of mind is performance. We do already recommend to use Verbose logs only for short time diagnostics in prod, so that's good, but still.
Going through the scenarios: Since this code is only applicable with serialization enabled, for user and OBO flows, the cache should have only a few tokens. If for OBO the key is user account ID, then cache will be bigger, but not too much. The only issue might be with client credentials with lots of tokens/resources per tenant. But we do get all tokens already in the findAT method, and it's not as bad as filtering itself.

I did add some of these perf related thoughts in logging discussion https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/discussions/3010

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: I was looking for something like this. It works when there are less elements that what is specified.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: Seems like it is not a bad idea to run this on a background thread to reduce performance impact on the active token acquisition call.
Thoughts @bgavrilMS @pmaytak ?

pmaytak: Hmm, generally, yes, put long-running process on a different thread. In this case, I'm not sure. It was brought up before there could be a possibility of the thread pool being exhausted in a high-throughput app? I wrote down my thoughts on the possible scenarios below - I don't think the cache will be large enough here for this method to take up enough time. So I think maybe for now the risk of adding a different thread here is not worth it - keep as is on the same thread. We should explore multi-threading logging as part of the threading discussion here in #3010

trwalke: makes sense

bgavrilMS: Yes, let's discuss about background thread operation separately. The fact that this is "verbose" logs only should be sufficient perf protection.

My 2c: in a "serverless" environment like Azure Functions, there's a good change that the background thread will be cut off. Same for public client CLI apps like "git" or "powershell" - the process is killed quickly after getting the token.

src/client/Microsoft.Identity.Client/Cache/Keys/MsalAccessTokenCacheKey.cs
```diff
@@ -78,6 +78,18 @@ public override string ToString()
                 _extraKeyParts);
         }
 
+        public string ToLogString(bool piiEnabled = false)
+        {
+            return MsalCacheKeys.GetCredentialKey(
+                piiEnabled? _homeAccountId : _homeAccountId.GetHashCode().ToString(),
```
bgavrilMS, line 84: We should get a stable hash instead. See https://andrewlock.net/why-is-string-gethashcode-different-each-time-i-run-my-program-in-net-core/

bgavrilMS:
@pmaytak - I'm merging this to get a release build out, after addressing all comments. Pls let me know if you have any other comments.

PR #3012, opened 2021-11-12
Add correlation id to notification args

### LINKED ISSUE #3008: [Feature Request] Provide a correlation id as part of cache callback
The cache callback parameters today do not include correlation guids. It is worthwhile to also pass the correlation id for tracing purpose as part of the [TokenCacheNotificationArgs ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.tokencachenotificationargs?view=azure-dotnet).

bgavrilMS:
Fixes #3008 

Add correlation id to the cache notification args

**Testing**
unit 

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -121,18 +121,21 @@ namespace Microsoft.Identity.Client.Cache
                             using (_requestParams.RequestContext.CreateTelemetryHelper(cacheEvent))
                             {
                                 string key = CacheKeyFactory.GetKeyFromRequest(_requestParams);
+                                
 
                                 try
                                 {
                                     var args = new TokenCacheNotificationArgs(
-                                       TokenCacheInternal,
-                                       _requestParams.AppConfig.ClientId,
-                                       _requestParams.Account,
-                                       hasStateChanged: false,
-                                       TokenCacheInternal.IsApplicationCache,
-                                       hasTokens: TokenCacheInternal.HasTokensNoLocks(),
-                                       _requestParams.RequestContext.UserCancellationToken,
-                                       suggestedCacheKey: key);
+                                      TokenCacheInternal,
+                                      _requestParams.AppConfig.ClientId,
+                                      _requestParams.Account,
+                                      hasStateChanged: false,
+                                      isApplicationCache: TokenCacheInternal.IsApplicationCache,
+                                      suggestedCacheKey: key,
+                                      hasTokens: TokenCacheInternal.HasTokensNoLocks(),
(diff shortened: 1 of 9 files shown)


trwalke (approved):
Looks solid

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
```diff
@@ -121,35 +121,26 @@ private async Task RefreshCacheForReadOperationsAsync(CacheEvent.TokenTypes cach
...
                             {
                                 string key = CacheKeyFactory.GetKeyFromRequest(_requestParams);
+                                var args = new TokenCacheNotificationArgs(
+                                      TokenCacheInternal,
+                                      _requestParams.AppConfig.ClientId,
+                                      _requestParams.Account,
+                                      hasStateChanged: false,
+                                      isApplicationCache: TokenCacheInternal.IsApplicationCache,
+                                      suggestedCacheKey: key,
+                                      hasTokens: TokenCacheInternal.HasTokensNoLocks(),
```
pmaytak, line 131: Does this change the caching operation behavior since the `HasTokens` is called once now? Would `OnBeforeAccessAsync` ever change this value?

tests/Microsoft.Identity.Test.Unit/CacheTests/CacheSerializationTests.cs
```diff
@@ -516,7 +516,7 @@ public async Task UnknownNodesTestAsync()
                 cache = notificationArgs.TokenCache.SerializeMsalV3();
             });
 
-            var notification = new TokenCacheNotificationArgs(tokenCache, null, null, false, false, true, CancellationToken.None);
+            var notification = new TokenCacheNotificationArgs(tokenCache, null, null, false, false, null, false, null, default, default);
```
pmaytak, line 519: nit: should use named params for readability.

pmaytak (commented):
Just a comment about moving notification args creation in CacheSessionManager related to HasTokens call. If we're sure it doesn't change cache operations flow, then LGTM; otherwise, more testing needed?

src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
trwalke: Im not sure it would make any difference here. The `HasTokensNoLocks` method ultimately checks to see if there is any tokens by pulling from the cache and unless there is a change to the cache (`hasStateChanged`) the value should remain the same for the duration of this method. But @bgavrilMS can confirm.

bgavrilMS: Better to be on the safe side, I'll revert this small change.

PR #3015, opened 2021-11-15
Fix sign out guest account in WAM

bgavrilMS:
Fixes # 

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
@@ -702,14 +702,14 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
             string homeTenantId = account?.HomeAccountId?.TenantId;
             if (!string.IsNullOrEmpty(homeTenantId))
             {                
-                bool isMsa = await IsMsaRequestAsync(
+                bool isMsaRequest = await IsMsaRequestAsync(
                    appConfig.Authority,
                    appConfig.TenantId,
                    _wamOptions.MsaPassthrough).ConfigureAwait(false);
 
-                IWamPlugin wamPlugin = isMsa ? _msaPlugin : _aadPlugin;
+                IWamPlugin wamPlugin = isMsaRequest ? _msaPlugin : _aadPlugin;
                 WebAccountProvider provider;
-                if (isMsa)
+                if (isMsaRequest)
                 {
                     provider = await _webAccountProviderFactory.GetAccountProviderAsync("consumers").ConfigureAwait(false);
                 }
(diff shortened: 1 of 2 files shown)


SameerK-MSFT (approved):

trwalke (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
```diff
@@ -701,11 +701,15 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
         {
             string homeTenantId = account?.HomeAccountId?.TenantId;
             if (!string.IsNullOrEmpty(homeTenantId))
-            {
-                bool isMsaAccount = IsConsumerTenantId(homeTenantId);
-                IWamPlugin wamPlugin = isMsaAccount ? _msaPlugin : _aadPlugin;
+            {                
+                bool isMsa = await IsMsaRequestAsync(
```
pmaytak, line 705: Nit: Maybe rename to `isMsaRequest` to specify/clarify a bit more?

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
```diff
@@ -843,6 +843,42 @@ public async Task RemoveAADAccountAsync()
...
+
+                // WAM can give MSAL the home account ID of a Wam account, which MSAL matches to a WAM account
+                _aadPlugin.GetHomeAccountIdOrNull(webAccount).Returns(aadHomeAccId);
+
+                var atsParams = new AcquireTokenSilentParameters();
+                _webAccountProviderFactory.GetAccountProviderAsync("organizations").ReturnsForAnyArgs(Task.FromResult(wamAccountProvider));
+
+                // Act Assert
+                await AssertException.TaskThrowsAsync<FileNotFoundException>( // Since WebAccount is a real object, it throws this exception
+                    () => _wamBroker.RemoveAccountAsync(harness.ServiceBundle.Config, requestParams.Account))
```
pmaytak, line 877: So how do we know if the account was removed from WAM or not? What would be the difference in behavior here?

pmaytak (approved):

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
bgavrilMS: Good catch. I'll add a comment as:

`This assertion ensures that WebAccount.SignOutAsync() is called. Since the WebAccount is fake, it throws a specific exception. This can be improved with a extra layer of abstraction over WebAccount, but it is sufficient for testing`

Issue #3004, opened 2021-11-11
Add Health/Metrics logging to MSAL that cannot be turned off.


trwalke:
Ensure that there is a level below error in the logger that will always log various health metrics related to MSAL caching operations

bgavrilMS:
Call it `LogAlways` like Azure SDK.

PR #3025, opened 2021-11-18
Adding HealthMetric log level.

trwalke:
Logging important metrics from cache

Fixes # 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3004

Adding new log level to logger. This level will always be enabled

The intention here is to ensure important health metrics are always logged when logging is enabled so we can assist customers in debugging easier.

**Testing**
Unit tests

**Performance impact**
Negligible


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -215,6 +216,7 @@ public void SetiOSKeychainSecurityGroup(string keychainSecurityGroup)
         public virtual void Clear()
         {
             AccessTokenCacheDictionary.Clear();
+            _logger.HealthMetric("Clearing access token cache data.");
```
trwalke, line 219: indicates the shouldClearExistingCache flag was set

trwalke:
@pmaytak I looked through the code and logged some of the things we discussed. Thoughts?
@bgavrilMS any suggestions.

trwalke:
@bgavrilMS I just saw your comment to use `LogAlways`. I will refactor to use that.

src/client/Microsoft.Identity.Client/Internal/Logger/NullLogger.cs
```diff
@@ -15,6 +15,14 @@ internal class NullLogger : ICoreLogger
         public Guid CorrelationId { get; } = Guid.Empty;
         public bool PiiLoggingEnabled { get; } = false;
 
+        public void HealthMetric(string messageScrubbed)
```
SameerK-MSFT, line 18: nit: unused params can be just [discards](https://docs.microsoft.com/en-US/dotnet/csharp/fundamentals/functional/discards).

src/client/Microsoft.Identity.Client/LogLevel.cs
```diff
@@ -9,24 +9,29 @@ namespace Microsoft.Identity.Client
     /// </summary>
     public enum LogLevel
     {
+        /// <summary>
```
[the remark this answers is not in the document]
jmprieur: yes it will
maybe use -1 for heath metrics and not change the others?

jmprieur (commented):
one remark

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformLogger.cs
```diff
@@ -9,6 +9,11 @@ namespace Microsoft.Identity.Client.Platforms.Android
 {
     internal class AndroidPlatformLogger : IPlatformLogger
     {
+        public void LogAlways(string message)
+        {
+            //NoOp
```
bgavrilMS, line 14: Doesn't this defeat the purpose of "always log" ? You can use `Log.Info` here. 

aside: if you look at the methods available on Android `Log` object, one of the options is `wtf`, which is documented as "what a terrible failure" :D

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/EventSourcePlatformLogger.cs
```diff
@@ -14,6 +14,11 @@ static EventSourcePlatformLogger()
 
         internal static MsalEventSource MsalEventSource { get; }
 
+        public void LogAlways(string message)
+        {
+            MsalEventSource.Error(message);
```
bgavrilMS, line 19: No, we should not log it as an error. Log as is Info. 
The fact that Info(message) calls ` MsalEventSource.Error(message);` is a bug.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -215,6 +216,7 @@ public void SetiOSKeychainSecurityGroup(string keychainSecurityGroup)
         public virtual void Clear()
         {
             AccessTokenCacheDictionary.Clear();
+            _logger.LogAlways("Clearing access token cache data.");
```
bgavrilMS, line 219: Is this important?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -456,6 +456,8 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
                 return null;
             }
 
+            requestParams.RequestContext.Logger.LogAlways($"Discovered {tokenCacheItems.Count} access tokens in cache using partition key: {partitionKey}");
```
bgavrilMS, line 459: Move to line 444?

bgavrilMS:
Changes look good to me, but we need to add more to this log. In particular we should make sure we have: 

- the metrics that we expose on AuthenticationResult
- the metrics that we send via HTTP telemetry

bgavrilMS (changes requested):
Minor comments on existing logic. We should log always some of the metrics though.

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformLogger.cs
trwalke: I wasn't sure which one to pick as there isnt a log always equivalent in android and I want sure how useful this info would be for PCA yet.
Isnt error below info so maybe we should use that instead?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
trwalke: This was something discussed with peter. It indicates the value of shouldClearExistingCache.
@pmaytak thoughts on this?

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformLogger.cs
trwalke: Yeah, I saw the wtf

src/client/Microsoft.Identity.Client/LogLevel.cs
```diff
@@ -9,6 +9,11 @@ namespace Microsoft.Identity.Client
...
+        ///     LogAlways Log Level
+        /// </summary>
+        LogAlways = -1,
```
pmaytak, line 15: Nit. Wondering if we could make the wording more consistent here and throughout? (I understand we're matching AzureSDK here.) Ex.: Essential, vital, required, fundamental? Or maybe just remove `Log` prefix so it's `LogLevel.Always`.

I'm ok either way though because Essential describes the kind of data is logged, Always describes when it is logged.

src/client/Microsoft.Identity.Client/LogLevel.cs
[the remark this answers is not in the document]
pmaytak: Are there a negative consequence of using a negative enum value? 
I think the underlying type is a signed int, so this is ok.
`default(LogLevel)` returns 0 (`LogLevel.Error`)  still. Will this cause issues in the code?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -441,6 +441,9 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
             Debug.Assert(partitionKey != null || !requestParams.IsConfidentialClient, "On confidential client, cache must be partitioned.");
 
             IReadOnlyList<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true, partitionKey);
+
+            requestParams.RequestContext.Logger.LogAlways($"Discovered {tokenCacheItems.Count} access tokens in cache using partition key: {partitionKey}");
```
pmaytak, line 445: Prefix log messages with location, ex `[FindAccessTokenAsync]`?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -772,9 +775,11 @@ internal async Task ExpireAllAccessTokensForTestAsync()
             if (requestParams.Authority == null)
                 return null;
 
-            IReadOnlyList<MsalRefreshTokenCacheItem> allRts = Accessor.GetAllRefreshTokens(CacheKeyFactory.GetKeyFromRequest(requestParams));
+            var requestKey = CacheKeyFactory.GetKeyFromRequest(requestParams);
+            IReadOnlyList<MsalRefreshTokenCacheItem> allRts = Accessor.GetAllRefreshTokens(requestKey);
             if (allRts.Count != 0)
             {
+                requestParams.RequestContext.Logger.LogAlways($"Discovered {allRts.Count} refresh tokens in cache using key: {requestKey}");
```
pmaytak, line 782: Maybe this should go outside the `if`?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -175,6 +175,7 @@ public void DeleteAccount(MsalAccountCacheItem item)
         /// </summary>
         public virtual IReadOnlyList<MsalAccessTokenCacheItem> GetAllAccessTokens(string partitionKey = null)
         {
+            _logger.LogAlways($"Total number of cache partitions found while getting access tokens: {AccessTokenCacheDictionary.Count}");
```
pmaytak, line 178: This might not be accurate in all of scenarios, like shared cache, because when the last cache item is removed from the partition dictionary, the dictionary itself is not removed. It's accurate for serialization scenario since we clear the dictionaries. We can probably update the removal logic, to remove empty partition. @bgavrilMS

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: I think it's helpful to know the state of the cache. But maybe this is redundant, since we log the partition count in `GetAllAccessTokens` and token count in Find AT and Find RT methods.
Maybe we can log the value of `shouldClearExistingCache` in token cache serializer methods, to know what the user passed it when Serialize/Deserialize was called? (Because that's a question we sometimes ask.)

tests/Microsoft.Identity.Test.Unit/PublicApiTests/LoggerTests.cs
```diff
@@ -42,20 +42,25 @@ public void ConstructorComponentTest()
         }
 
         [TestMethod()]
-        public void CallbackTestErrorTest()
+        public void CallbackTestLogAlwaysTest()
```
pmaytak, line 45: Could some of these test methods be combined into parametrized tests?

src/client/Microsoft.Identity.Client/LogLevel.cs
trwalke: I think `LogLevel.Always` makes sense

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
trwalke: I think it makes sense to remove the empty partitions

tests/Microsoft.Identity.Test.Unit/PublicApiTests/LoggerTests.cs
trwalke: Yes, looking into it

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
trwalke: I double checked the code and this method is only called when `clearExistingCacheData` is set to true so if we see this in the logs, we know the value. I will look into getting the logger into the serializer as it is not straight forward from what I see so far.

trwalke:
Thanks for the review @bgavrilMS and @pmaytak. Do you feel there is anything else useful for us to log here that will help us with customer issues?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -111,6 +113,16 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
             }
         }
 
+        private static void LogMetricsFromAuthResult(AuthenticationResult authenticationResult, ICoreLogger logger)
+        {
+            logger.Always($"[LogMetricsFromAuthResult] Cache Refresh Reason: {authenticationResult.AuthenticationResultMetadata.CacheRefreshReason.ToString()}");
```
bgavrilMS, line 118: nit: i don't think you need ToString?

src/client/Microsoft.Identity.Client/LogLevel.cs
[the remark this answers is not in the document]
bgavrilMS: https://social.msdn.microsoft.com/Forums/vstudio/en-US/805eb2c7-ce76-403d-ad3f-614cab453b58/best-practices-for-negative-enumeration-values?forum=csharpgeneral details 

Looks like Enum.GetValues will return a "strange" ordering of elements, but otherwise seems ok?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: Fair point. I'm fine with keeping it here, as-is.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: Based on the discussion, this is fine then; in a shared cache we would remove token then re-add with the same key.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -175,6 +175,7 @@ public void DeleteAccount(MsalAccountCacheItem item)
         /// </summary>
         public virtual IReadOnlyList<MsalAccessTokenCacheItem> GetAllAccessTokens(string partitionKey = null)
         {
+            _logger.Always($"[GetAllAccessTokens] Total number of cache partitions found while getting access tokens: {AccessTokenCacheDictionary.Count}");
```
pmaytak, line 178: Should the same log be added to user cache also?

pmaytak (approved):
+ Bogdan's comment about possible NPE.
+ Maybe run a quick perf test. This adds some calls to `.Count` but I think it's always on a List, so should be no perf impact.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/LoggerTests.cs
trwalke: updated

src/client/Microsoft.Identity.Client/Platforms/Android/AndroidPlatformLogger.cs
bgavrilMS: The information we log under this new category LogAlways is not really an error or a warning. So when we adapt our logger to the Android logger, we can use AndroidLogger.Info. The point here is that MSAL will always log these strings.

SameerK-MSFT (approved):

PR #3030, opened 2021-11-18
Test calling OBO and Client in different order when regional is enabl…

bgavrilMS:
…ed and reusing CCA instance.

Fixes # 

**Testing**

**Performance impact**


pmaytak:
Converting to a draft because I have 2 more tests I'll add.

PR #3035, opened 2021-11-24
Pmaytak/telemetry multi thread

### LINKED ISSUE #3024: [Bug] Crash in MSAL telemetry API
MSAL 4.37.0

Stack trace below. Probably a bug related to the ApiEvent collection, which is not thread safe. 

Please consider refactoring telemetry, as the current infrastructure is very complex for our needs (i.e. deprecate WithTemelemetry and all references to MATS). 

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
at <redacted>
at System.Collections.Generic.Dictionary2.Enumerator.MoveNext()
at Microsoft.Identity.Client.Internal.Requests.RequestBase.InitializeApiEvent(String accountId)
at Microsoft.Identity.Client.Internal.Requests.RequestBase.RunAsync(CancellationToken cancellationToken)
at Microsoft.Identity.Client.ApiConfig.Executors.ConfidentialClientExecutor.ExecuteAsync(AcquireTokenCommonParameters commonParameters, AcquireTokenForClientParameters clientParameters, CancellationToken cancellationToken)
at <redacted>
--- End of inner exception stack trace ---
```
```
at <redacted>
at <redacted>
<redacted>
 ---> System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
at System.Collections.Generic.Dictionary2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
at Microsoft.Identity.Client.ApiConfig.Parameters.AcquireTokenCommonParameters.AddApiTelemetryFeature(ApiTelemetryFeature feature)
at Microsoft.Identity.Client.AbstractAcquireTokenParameterBuilder1.WithAuthority(String authorityUri, Boolean validateAuthority)
at <redacted>
```

To-do:
- [x] Fix the bug
- [x] Update dev app with repro
- [] Remove references to MATS and refactor/clean up HTTP telemetry.
- [] Run performance tests for before and after

pmaytak:
Fixes #3024

**Testing**

**Performance impact**


tests/devapps/NetCoreTestApp/Program.cs
```diff
@@ -25,18 +24,20 @@ public class Program
...
 
         // Confidential client app with access to https://graph.microsoft.com/.default
-        private static readonly string s_clientIdForConfidentialApp =
-            Environment.GetEnvironmentVariable("LAB_APP_CLIENT_ID");
+        private static readonly string s_clientIdForConfidentialApp = "50704bdf-2f17-4f9b-a78a-76726753833a";
 
         // App secret for app above 
-        private static readonly string s_confidentialClientSecret =
-            Environment.GetEnvironmentVariable("LAB_APP_CLIENT_SECRET");
+        private static readonly string s_confidentialClientSecret = ":-u/Q0r?dgfbbc4jWQpMxhReGbnvAj25";
```
bgavrilMS, line 30: Erm, please revert this and delete the secret

jeevanbalanmanoj:
@pmaytak  Will this fix make it possible to reuse AcquireTokenForClientParameterBuilder instance in multiple threads?

tests/devapps/NetCoreTestApp/Program.cs
```diff
@@ -241,6 +244,33 @@ 1. IWA
...
+                            int totalThreads = int.TryParse(Console.ReadLine(), out totalThreads) ? totalThreads : 10;
+                            Console.Write("Enter run duration in seconds (default 10): ");
+                            int durationInSeconds = int.TryParse(Console.ReadLine(), out durationInSeconds) ? durationInSeconds : 10;
+
+                            var acquireTokenBuilder = CreateCca().AcquireTokenForClient(GraphAppScope);
+
+                            var threads = new List<Thread>();
+                            for (int i = 0; i < totalThreads; i++)
+                            {
+                                var thread = new Thread(new ThreadStart(new ThreadWork(i, acquireTokenBuilder, durationInSeconds).Run));
```
bgavrilMS, line 258: nit: Threads are so 2015... Task.WhenAll() is the way you can do the same with Tasks. See https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/e7f579d3f9db9acc2e9225356ab7a72b956949ec/sample/ManualTestApp/Program.cs#L124

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/MatsConverter.cs
```diff
@@ -32,162 +19,27 @@ internal enum OsPlatform
 
     internal static class MatsConverter
```
bgavrilMS: Can't this go file go in its entirety?

src/client/Microsoft.Identity.Client/AppConfig/ITelemetryEventPayload.cs
```diff
@@ -1,14 +1,18 @@
...
 using System.Collections.Generic;
+using System.ComponentModel;
 
 namespace Microsoft.Identity.Client
 {
     /// <summary>
     /// Data that represents a single snapshot in the series of events that are collected
     /// </summary>
     /// <remarks>This API is experimental and it may change in future versions of the library without an major version increment</remarks>
+    [Obsolete("Telemetry is sent automatically by MSAL.NET. See https://aka.ms/msal-net-telemetry.", false)]
```
bgavrilMS, line 14: >false [](http://example.com/codeflow?start=101&length=5)

Should we obsolete this with error?

src/client/Microsoft.Identity.Client/AppConfig/TraceTelemetryConfig.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.Identity.Json;
+using Microsoft.Identity.Json.Linq;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// A simple <see cref="ITelemetryConfig"/> implementation that writes data using System.Diagnostics.Trace.
+    /// </summary>
+    /// <remarks>This API is experimental and it may change in future versions of the library without an major version increment</remarks>
+    public class TraceTelemetryConfig : ITelemetryConfig
```
bgavrilMS, line 20: Is this a new file? Or was it just moved?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -54,12 +54,15 @@ public enum ApiIds
         public ApiEvent(
             ICoreLogger logger,
             ICryptographyManager cryptographyManager,
-            string correlationId) : base(EventNamePrefix + "api_event", correlationId)
+            Guid correlationId) : base(EventNamePrefix + "api_event")
```
bgavrilMS: This is great, string operations take a lot of CPU time and this is on the hot path.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/EventBase.cs
```diff
@@ -23,14 +22,13 @@ internal abstract class EventBase : Dictionary<string, string>
         private readonly long _startTimestamp;
         public const string TenantPlaceHolder = "<tenant>"; // It is used to replace the real tenant in telemetry info
 
-        public EventBase(string eventName, string correlationId)
+        public EventBase(string eventName)
```
bgavrilMS: Is this base class necessary after all the refactoring? We only have ApiEvent? In paticular, the `ScrubTenant` method seems to do a lot of work.

src/client/Microsoft.Identity.Client/AppConfig/TraceTelemetryConfig.cs
bgavrilMS: Never mind, I think GH shows it as new because of spacing. CodeFlow shows me the changes.

bgavrilMS:
LGTM

tests/devapps/NetCoreTestApp/Program.cs
pmaytak: Tasks can run on the same thread though right? With threads, it's guaranteed separate.

henrik-me:
Has it been considered if our telemetry would need any of the information being removed for any metrics we would like to track or we would like our customers to track?

henrik-me:
> Fixes #3043.
> 
> **Changes proposed in this request**
> 
> **Testing**
> 
> **Performance impact**

nit (randomly chose this PR but could have picked any PR): Seems like the template information is not filled out? should it be? Is there a process around it?

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -313,15 +313,10 @@ public T WithLegacyCacheCompatibility(bool enableLegacyCacheCompatibility = true
         /// <returns>The builder to chain the .With methods</returns>
         /// <exception cref="InvalidOperationException"/> is thrown if the method was already
         /// called on the application builder.
-
+        [Obsolete("Telemetry is sent automatically by MSAL.NET. See https://aka.ms/msal-net-telemetry.", false)]
```
henrik-me, line 316: >"Telemetry is sent automatically by MSAL.NET. See https://aka.ms/msal-net-telemetry." [](http://example.com/codeflow?start=18&length=85)

I think we should refer to logging as well as our metrics for monitoring here

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -485,9 +480,10 @@ public T WithClientCapabilities(IEnumerable<string> clientCapabilities)
         /// </summary>
         /// <param name="telemetryConfig"></param>
         /// <returns></returns>
+        [Obsolete("Telemetry is sent automatically by MSAL.NET. See https://aka.ms/msal-net-telemetry.", false)]
```
henrik-me, line 483: >"Telemetry is sent automatically by MSAL.NE [](http://example.com/codeflow?start=18&length=43)

see other comment

src/client/Microsoft.Identity.Client/AppConfig/TelemetryConfig.cs
```diff
@@ -1,33 +0,0 @@
...
-// Licensed under the MIT License.
-
-using System;
-using System.Collections.Generic;
-
-namespace Microsoft.Identity.Client
-{
-    /// <summary>
-    /// </summary>
-    internal class TelemetryConfig : ITelemetryConfig
```
henrik-me, line 11: >TelemetryConfig [](http://example.com/codeflow?start=19&length=15)

Wondering if this should be marked as obsolete as well? perhaps with true?

src/client/Microsoft.Identity.Client/AppConfig/TraceTelemetryConfig.cs
```diff
@@ -23,7 +25,7 @@ public class TraceTelemetryConfig : ITelemetryConfig
         /// <remarks>This API is experimental and it may change in future versions of the library without an major version increment</remarks>
         public TraceTelemetryConfig()
         {
-            SessionId = Guid.NewGuid().AsMatsCorrelationId();
+            SessionId = Guid.NewGuid().ToString();
```
henrik-me, line 28: >SessionId [](http://example.com/codeflow?start=12&length=9)

if the class is never used perhaps this should be a fixed string. such as string.empty or 0 mor Guid.empty?

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
```diff
@@ -103,13 +102,13 @@ public string GetCurrentRequestHeader(ApiEvent eventInProgress)
...
-            eventInProgress.TryGetValue(MsalTelemetryBlobEventNames.RegionOutcome, out string regionOutcome);
-            eventInProgress.TryGetValue(MsalTelemetryBlobEventNames.IsTokenCacheSerializedKey, out string isTokenCacheSerialized);
-            eventInProgress.TryGetValue(MsalTelemetryBlobEventNames.IsLegacyCacheEnabledKey, out string isLegacyCacheEnabled);
+            string apiId = eventInProgress.ApiIdString;
+            string cacheInfo = eventInProgress.CacheInfoString;
+            string regionUsed = eventInProgress.RegionUsed;
+            string regionSource = eventInProgress.RegionAutodetectionSourceString;
+            string regionOutcome = eventInProgress.RegionOutcomeString;
+            string isTokenCacheSerialized = eventInProgress.IsTokenCacheSerializedString;
+            string isLegacyCacheEnabled = eventInProgress.IsLegacyCacheEnabledString;
```
henrik-me, line 111: wondering if there is a benefit to doing these string operations here? Perhaps that is not needed allocations?

henrik-me:
>         }

wondering if a stringbuilder or span would be faster here?

---
Refers to: src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs:120 in 1f5fb57. [](commit_id = 1f5fb57ec9d18e96310bcf372890f69d160c6d87, deletion_comment = False)

tests/devapps/NetCoreTestApp/Program.cs
```diff
@@ -241,6 +244,33 @@ 1. IWA
 
                             break;
 
+                        case 11: // AcquireTokenForClient with multiple threads
```
henrik-me, line 247: >AcquireTokenForClient with multiple threads [](http://example.com/codeflow?start=36&length=43)

did this test / sample fail before the fix?

henrik-me:
seems like this PR has not been synced with master for some time?

LogMetricsFromAuthResult is missing, seems like we should compress that information into one string operation instead of multiple... @trwalke ?

src/client/Microsoft.Identity.Client/AppConfig/TraceTelemetryConfig.cs
pmaytak: May be because I removed it initially, then readded back (since it's public).

pmaytak:
> Has it been considered if our telemetry would need any of the information being removed for any metrics we would like to track or we would like our customers to track?

This has crossed my mind. I will add this analysis to the PR description.

> > Fixes #3043.
> > **Changes proposed in this request**
> > **Testing**
> > **Performance impact**
> 
> nit (randomly chose this PR but could have picked any PR): Seems like the template information is not filled out? should it be? Is there a process around it?

For published PRs, it should be filled out. Not filled out here because it's a draft.

> seems like this PR has not been synced with master for some time?

Yes, I rebased.

tests/devapps/NetCoreTestApp/Program.cs
pmaytak: Yes, it fails with the code without the fix.

src/client/Microsoft.Identity.Client/AppConfig/TraceTelemetryConfig.cs
pmaytak: Would it even matter if the class is never instantiated/?

src/client/Microsoft.Identity.Client/AppConfig/ITelemetryEventPayload.cs
pmaytak: Even though no one should be using this, it's still a breaking change? I think we only obsoleted with error for experimental APIs.

src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs
pmaytak: No benefit. I intended for this PR to have only code removal and as little other changes as possible, so reviewing is easier and nothing gets lost. I'm thinking of refactoring this in a separate PR and maybe even split this PR into smaller ones (although that might be difficult = lots of class dependencies).

pmaytak:
> >         }
> 
> wondering if a stringbuilder or span would be faster here?
> 
> ---
> Refers to: src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs:120 in 1f5fb57. [](commit_id = 1f5fb57ec9d18e96310bcf372890f69d160c6d87, deletion_comment = False)

Maybe slightly. If I remember correctly compiler can use StringBuilder for string interpolation underneath, if needed.

bgavrilMS:
>         }

It depends, I don't think there is a silver bullet here. [https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?redirectedfrom=MSDN&view=net-6.0#the-string-and-stringbuilder-types](https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?redirectedfrom=MSDN&view=net-6.0#the-string-and-stringbuilder-types)

If we were in .NET6 / C# 10, there seems to be a way to do this: [https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/](https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/)

All in all, this code is not on the hot path. We create this string when we make an HTTP call to /token endpoint. Hopefully this does not happen often as ppl have good caching.

---
In reply to: [983250949](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3035#issuecomment-983250949) [](http://example.com/codeflow?ancestors=983250949,983250949)

---
Refers to: src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs:120 in 9a7039d. [](commit_id = 9a7039da3a57fa8b6f178c2f6f42f6e40c39a64c, deletion_comment = False)

trwalke:
> 



> seems like this PR has not been synced with master for some time?
> 
> LogMetricsFromAuthResult is missing, seems like we should compress that information into one string operation instead of multiple... @trwalke ?

Makes sense, We can use string builder to build the result and log all at once

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -63,164 +35,102 @@ public enum ApiIds
...
 
-#region Region
-        public string RegionUsed
+        private int? _regionAutodetectionSource;
+        public int RegionAutodetectionSource
         {
-            get => this.ContainsKey(RegionUsedKey) ? this[RegionUsedKey] : null;
-            set => this[RegionUsedKey] = value;
+            get { return _regionAutodetectionSource.HasValue ? _regionAutodetectionSource.Value : 0; }
+            set { _regionAutodetectionSource = value; }
```
pmaytak, line 85: Some of these properties have a nullable backing variable and a companion String variable because previously these values were stored in a dictionary. So if the value was never set, it would be null in the telemetry string. So using the property itself would default to 0. Also can't update the property itself to nullable because it will affect the usage. Will have to look at this to refactor better.

bgavrilMS (approved):

pmaytak:
Added performance test results to the PR description.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -63,164 +35,102 @@ public enum ApiIds
...
-        public ApiIds ApiId
-        {
-            get => TryGetValue(MsalTelemetryBlobEventNames.ApiIdConstStrKey, out string apiIdString) ? (ApiIds)Enum.Parse(typeof(ApiIds), apiIdString) : ApiIds.None;
+        public Guid CorrelationId { get; set; }
 
-            set => this[MsalTelemetryBlobEventNames.ApiIdConstStrKey] = ((int) value).ToString(CultureInfo.InvariantCulture);
-        }
+        public ApiIds ApiId { get; set; }
 
         public string ApiIdString
```
SameerK-MSFT, line 49: nit: public string ApiIdString => ((int)ApiId).ToString(CultureInfo.InvariantCulture);

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -63,164 +35,102 @@ public enum ApiIds
...
-        {
-            get; set;
-        }
+        public string TokenEndpoint { get; set; }
 
-        public Uri Authority
-        {
-            set => this[AuthorityKey] = ScrubTenant(value)?.ToLowerInvariant();
-        }
+        // Some of these properties like Authority, TenantId, LoginHint, etc.
```
SameerK-MSFT, line 56: This is an internal class. If they are not being used, it will be safe to remove.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -63,164 +35,102 @@ public enum ApiIds
...
+        public string RegionUsed { get; set; }
 
-#region Region
-        public string RegionUsed
+        private int? _regionAutodetectionSource;
+        public int RegionAutodetectionSource
         {
-            get => this.ContainsKey(RegionUsedKey) ? this[RegionUsedKey] : null;
-            set => this[RegionUsedKey] = value;
+            get { return _regionAutodetectionSource.HasValue ? _regionAutodetectionSource.Value : 0; }
```
SameerK-MSFT, line 84: nit: smaller code _regionAutodetectionSource ?? 0;

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
pmaytak: Fixed in PR #3058.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
pmaytak: Fixed in PR #3058.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
pmaytak: Fixed in PR #3048.

PR #3036, opened 2021-11-24
Gladjohn/b2c ccs header fix

### LINKED ISSUE #2748: [Bug] CCS routing hint sends "missing from token response" as UPN in case of B2C
image
repro:
Sign-in w/a b2c account. Close the app. Sign-in again using `.WithLoginHint()`, and you'll see "Missing from the token response" in the UI, then in the routing hint, it is sent. Probably should not send it in this case? Or not send for B2C.

gladjohn:
Fixes #2748 

Fixed by not adding CCS header when Authority Type is B2C

**Testing**
Added an Integration test

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -316,13 +316,13 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
         protected KeyValuePair<string, string>? GetCcsUpnHeader(string upnHeader)
         {
-            string OidCcsHeader = CoreHelpers.GetCcsUpnHint(upnHeader);
-
             if (AuthenticationRequestParameters.Authority.AuthorityInfo.AuthorityType == AuthorityType.B2C)
             {
                 return new KeyValuePair<string, string>();
             }
-            
+
+            string OidCcsHeader = CoreHelpers.GetCcsUpnHint(upnHeader);
+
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -316,7 +316,13 @@ protected Task ResolveAuthorityAsync()
 
         protected KeyValuePair<string, string>? GetCcsUpnHeader(string upnHeader)
```
bgavrilMS, line 317: Can we return null instead? It will a tiny bit faster. Same for line 314 above. Otherwise there's no point in returning a nullable struct, if you're not actually returning null.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -44,6 +44,8 @@ public class UsernamePasswordIntegrationTests
             "1003,ad8c894a-557f-48c0-b045-c129590c344e";
         private const string InvalidGrantError = "invalid_grant";
         private const string UPApiId = "1003";
+        private const string _b2CROPCAuthority = "https://msidlabb2c.b2clogin.com/tfp/msidlabb2c.onmicrosoft.com/B2C_1_ROPC_Auth";
```
bgavrilMS, line 47: See naming conventions used. Constants are UpperCase.

bgavrilMS (approved):

SameerK-MSFT (approved):

PR #3048, opened 2021-12-02
Refactor ApiEvent (follow up to removing MATS telemetry)

pmaytak:
Fixes #3043 (refactoring follow up, needs to be merged after #3035).

**Changes proposed in this request**
- Remove unused properties from `ApiEvent`.
- Use enum types instead of int in ApiEvent.
- Use StringBuilder to build telemetry string in `HttpTelemetryManager`.
- Update LogMetricsFromAuthResult to use StringBuilder
- Add perf test for AcquireTokenForClient and OBO call without cache.

**Testing**
Existing tests. Verified no breaking changes.

**Performance impact**
See #3035

PR #3058, opened 2021-12-08
Add OBO tests and minor changes to tests and dev apps

pmaytak:
Needs to be merged after #3048.

- Add more descriptive `LogLevel` comments.
- Minor perf test renamings.
- Add two OBO related integration tests.
- Update some dev apps to use cca config for lab.

**Testing**
N/A

**Performance impact**
N/A


tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
```diff
@@ -52,6 +53,180 @@ public void TestInitialize()
...
 
+        /// <summary>
+        /// Tests the behavior when calling OBO and silent in different orders with multiple users.
+        /// OBO calls should return tokens for correct users, silent calls should throw.
+        /// </summary>
+        [DataTestMethod]
+        [DataRow(false, false)]
+        [DataRow(true, false)]
+        [DataRow(true, true)]
+        public async Task OboAndSilent_Async(bool serializeCache, bool usePartitionedSerializationCache)
```
bgavrilMS, line 64: nit / no action needed: integration tests are slow so we need to pick and choose scenarios where we use integration testing vs unit testing. Our unit testing strategy, where we fake out AAD, is pretty good (not perfect, needs more flexibility, but pretty good).

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/OboTests.cs
pmaytak: Good point. When I was investigating these, I think I went with integration tests to make sure the E2E is correct and I don't mock something incorectly.

trwalke:
>     /// 1 - 100,000 - 100,000

Should probably update comment to reflect that these have changed

---
Refers to: tests/Microsoft.Identity.Test.Performance/AcquireTokenForOboCacheTests.cs:26 in 5db3247. [](commit_id = 5db3247fd6dd234a2ce55a2ffef19fbed9eb9ecf, deletion_comment = False)

trwalke:
>     /// 1 - 100,000 - 100,000

Should probably update comment to reflect that these have changed

---
Refers to: tests/Microsoft.Identity.Test.Performance/TokenCacheTests.cs:25 in 5db3247. [](commit_id = 5db3247fd6dd234a2ce55a2ffef19fbed9eb9ecf, deletion_comment = False)

trwalke (approved):

gladjohn (approved):
looks good. Filed a task on lab to see if they can expose a different way to get users without having to hardcode UPNs.

PR #3062, opened 2021-12-08
Catches policy protection required err + throws ex

SameerK-MSFT:
Fixes # 

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
@@ -40,6 +40,9 @@ namespace Microsoft.Identity.Client.Platforms.iOS
         private string _brokerRequestNonce;
         private bool _brokerV3Installed = false;
 
+        private string _accessToken;
+        private string _refreshToken;
+
         public iOSBroker(ICoreLogger logger, ICryptographyManager cryptoManager, CoreUIParent uIParent)
         {
             _logger = logger;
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Platforms/iOS/Broker/iOSBroker.cs
```diff
@@ -271,17 +274,27 @@ private MsalTokenResponse ResultFromBrokerResponse(Dictionary<string, string> re
...
                 {
-                    responseDictionary[BrokerResponseConst.BrokerErrorCode] = MsalError.AuthenticationCanceledError;
+                    if(errCode == BrokerResponseConst.iOSBrokerUserCancellationErrorCode)
+                    {
+                        responseDictionary[BrokerResponseConst.BrokerErrorCode] = MsalError.AuthenticationCanceledError;
+                    }
+                    else if (errCode == BrokerResponseConst.iOSBrokerProtectionPoliciesRequired)
+                    {
+                        responseDictionary[BrokerResponseConst.BrokerErrorCode] = MsalError.ProtectionPolicyRequired;
+                        _accessToken = responseDictionary[BrokerResponseConst.AccessToken];
```
bgavrilMS, line 290: It won't be enough to give MSAL just the AT and RT. You need to give it full response properties and ID token for caching like `CreateFromiOSBrokerResponse` does.

SameerK-MSFT:
Closing as there is a different set of changes that is more appropriate.

PR #3065, opened 2021-12-10
Add extensibility mechanism to support a legacy POP implementation.

bgavrilMS:
Fixes # 
Add extensibility mechanism to support a legacy POP implementation. 

- override the client assertion data
- allow binding to key

**Review Help**
- Review each commit separately
- Have a look at `LegacyPoPAsync` test (e.g. step through it)

**Testing**
integration + unit

**Performance impact**


src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Overrides the client credentials parameters (e.g. client_assertion) and optionally ties the 
+        /// resulting token to a key id, similar to POP tokens.
+        /// </summary>
+        /// <param name="clientAssertionOverride">A lambda which receives a string representing the value of the token endpoint and which needs to return key value pairs to be added to the POST payload</param>
+        /// <param name="keyId">A key id to which the access token is associated. The token will not be retrieved from the cache unless the same key id is presented. Can be null.</param>
+        /// <param name="builder">Builder to chain config options to</param>
+        /// <returns>The builder</returns>
+        public static AbstractConfidentialClientAcquireTokenParameterBuilder<T> WithClientAssertionOverride<T>
```
bgavrilMS, line 22: This extension method does 2 things: 

- inject custom params to the POST request instead of "client_assertion"
-  bind the token to a keyid

This is what's needed to support MSAuth10POP protocol, a legacy POP protocol. Still, from a design perspective of this API, it could be split into 2. Thoughts?

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Overrides the client credentials parameters (e.g. client_assertion) and optionally ties the 
+        /// resulting token to a key id, similar to POP tokens.
+        /// </summary>
+        /// <param name="clientAssertionOverride">A lambda which receives a string representing the value of the token endpoint and which needs to return key value pairs to be added to the POST payload</param>
+        /// <param name="builder">Builder to chain config options to</param>
+        /// <returns>The builder</returns>
+        public static AbstractConfidentialClientAcquireTokenParameterBuilder<T> WithClientAssertionOverride<T>
+            (this AbstractConfidentialClientAcquireTokenParameterBuilder<T> builder,
+            Func<string, IReadOnlyList<KeyValuePair<string, string>>> clientAssertionOverride)
```
bgavrilMS, line 23: Potential changes: 

- make async
- move to App config.

jmprieur: In mechanism2, you mean requestBuilder for the 2nd snippet, @bgavrilMS ?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -462,6 +462,74 @@ private enum CredentialType
...
+                var cert = new X509Certificate2(ResourceHelper.GetTestResourceRelativePath("valid.crtfile"));
+
+                var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
+                              .WithAuthority("https://login.microsoftonline.com/tid/")
+                              .WithExperimentalFeatures(true)
+                              .WithHttpManager(httpManager)
+                              .WithCertificate(cert)
+                              .Build();
+
+                Func<string, IReadOnlyList<KeyValuePair<string, string>>> clientAssertionOverride =
```
jmprieur, line 481: I don't understand what the scenario is?

jmprieur: The problem is it applies to the client assertions generated from the certificates. 
Can it (It should not) apply to the signed assertions passed instead of the client secret / cert?

jmprieur (commented):
I've made a few comments.
I guess I don't understand the why enough

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
bgavrilMS: No, in mechanism2 I mean a lambda. The lambda is invoked on a per request basis.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
bgavrilMS: Ack, I'll move it to the app. 
The sceanrio here is to add an extensibility mechanism for SAL to do MsAuth_10_AT_POP, which I call "legacy POP"

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
bgavrilMS: The lambda mechanism is too messy. I tried and the user code (in the integration test) is too complex.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/LegacyPopTest.cs
```diff
@@ -0,0 +1,178 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Identity.Test.Integration.NetFx.HeadlessTests
+{
+    [TestClass]
+    public class LegacyPopTests
+    {
+        [TestMethod]
+        public async Task LegacyPoPAsync()
```
bgavrilMS, line 27: @jmprieur - this integration test shows how SAL will use the 2 extensibility points: 

- WithKeyId
- WithClientAssertion

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/LegacyPopTest.cs
```diff
@@ -0,0 +1,178 @@
...
+                .Create(settings.ClientId)
+                .WithAuthority(settings.Authority, true)
+                .WithCertificate(clientCredsCert)
+                .WithExperimentalFeatures(true)
+                .WithTestLogging()
+                .Build();
+
+            var result = await cca.AcquireTokenForClient(settings.AppScopes)
+                .WithClientAssertionOverride(
+                    (tokenEndpoint) => CreateMs10ATPOPBodyParams(settings.ClientId, tokenEndpoint, clientCredsCert, popKey, true),
```
bgavrilMS, line 43: The content of `CreateMs10ATPOPBodyParams` is not relevant. But suffice to say it combines the client creds X509 certificate with the POP key and adds it as a param named `request` (and not `client_assertion`)

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
trwalke: For me, they seem to be more or less the same. You always have the option of setting the func equal to a defined method instead of lambda and then it just becomes another parameter. 

Also, we already have 
`ConfidentialClientApplicationBuilder WithClientAssertion(string signedClientAssertion)`
On the app builder so it may seem confusing if someone is already used to this being on the app builder but suddenly the new override is moved to the request builder. Also, the client assertions represent application credentials and we tend to put these in the application builder.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -131,8 +131,11 @@ protected override SortedSet<string> GetOverriddenScopes(ISet<string> inputScope
 
         private Dictionary<string, string> GetBodyParameters()
         {
+            //this.ServiceBundle.Config.ClientCredential
```
trwalke, line 134: can be removed

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
bgavrilMS: After playing around with this some more, it is clear to me that the app developer wanting to keep the CCA object as singleton, which is important for perf, needs to pass in the cert at request level. Having it at app level makes the programming much more complicated and prone to threading issues.

bgavrilMS:
Will do some prototyping with SAL and come back to this.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/LegacyPopTest.cs
```diff
@@ -0,0 +1,185 @@
...
+                    (tokenEndpoint, ct) => 
+                    Task.FromResult(CreateMs10ATPOPBodyParams(settings.ClientId, tokenEndpoint, clientCredsCert, _currentPopKey, true)))
+                .WithKeyId(_currentPopKey.KeyId)
+                .ExecuteAsync().ConfigureAwait(false);
+
+            Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+            MsalAccessTokenCacheItem at = (cca.AppTokenCache as ITokenCacheInternal).Accessor.GetAllAccessTokens().Single();
+            Assert.AreEqual(at.KeyId, _currentPopKey.KeyId);
+
+            result = await cca.AcquireTokenForClient(settings.AppScopes)
```
pmaytak, line 54: If we do a normal acquire token call here, it will overwrite the pop token in the cache, but it's still valid, so that's why you're adding a kid to cache key?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/LegacyPopTest.cs
```diff
@@ -0,0 +1,185 @@
...
+        private static IReadOnlyList<KeyValuePair<string, string>> CreateMs10ATPOPBodyParams(
+           string clientId,
+           string audience,
+           X509Certificate2 clientCredCert,
+           RsaSecurityKey popKey,
+           bool includeX5cClaim)
+        {
+            var clientCredsSigningCredentials = new SigningCredentials(new X509SecurityKey(clientCredCert), SecurityAlgorithms.RsaSha256, SecurityAlgorithms.Sha256);
+            string request = CreateMs10ATPOPAssertion(clientId, audience, clientCredsSigningCredentials, popKey, includeX5cClaim);
+            return new List<KeyValuePair<string, string>>(1) { new KeyValuePair<string, string>("request", request) };
```
pmaytak, line 113: So for Ms10ATPOP there has to be a body param request? Does this func need to return a collection of key-values or are you making this a bit more generic to apply to more different cases?

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// Binds the token to a key in the cache. L2 cache keys contain the key id.
+        /// No cryptographic operations is performed on the token.
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="keyId">A key id to which the access token is associated. The token will not be retrieved from the cache unless the same key id is presented. Can be null.</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static AcquireTokenForClientParameterBuilder WithKeyId(
```
pmaytak, line 22: Can this method be merged into that new `WithClientAssertion` as a param? Right now both methods would have to be called. This is similar situation that we had with the long-running OBO work, separate WithCacheKey method, kind of didn't work out.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -0,0 +1,39 @@
...
+            Func<string, CancellationToken, Task<IReadOnlyList<KeyValuePair<string, string>>>> clientAssertionAsyncDelegate)
+            where T : AbstractConfidentialClientAcquireTokenParameterBuilder<T>
+        {
+            if (clientAssertionAsyncDelegate == null)
+            {
+                throw new ArgumentNullException(nameof(clientAssertionAsyncDelegate));
+            }
+
+            builder.ValidateUseOfExperimentalFeature();
+            builder.CommonParameters.ClientAssertionOverrideDelegate = clientAssertionAsyncDelegate;
```
pmaytak, line 34: Wondering if there's a better name for this? The other `WithClientAssertion` method sets `Config.SignedClientAssertionDelegate` property. Maybe `ClientAssertionParametersDelegate` or something? To differentiate that the other method creates only the client_assertion param, this creates a bunch.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Extensibility
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public static class AbstractConfidentialClientAcquireTokenParameterBuilderExtension
```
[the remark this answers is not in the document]
pmaytak: Also would this apply to all confidential client flows or only client credentials?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/LegacyPopTest.cs
bgavrilMS: Yes, that's the general idea. The scenario where you request a Bearer token and a POP token for the same resource is probably quite obscure. 

The scenario where cache key includes kid is more in the lines of POP key rotation: 

## Before the changes

1. request with kid1 -> IDP
2. request with kid1 -> Cache
3. request with kid2 -> IDP
4. request with kid2 -> Cache
5. request with kid1 -> IDP

## After the cache key changes
1-4 remain the same
5. request with kid1 -> cache

When doing in-memory caching only it's not that important. But for distributed system + L2 caching it becomes a problem, if 2 machines have different POP keys.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/LegacyPopTest.cs
bgavrilMS: Making it generic. I am going to use this to send normal client credentials as well. It helps with having a singleton CCA.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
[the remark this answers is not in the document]
bgavrilMS: 1. I'm thinking of keeping it in a separate namespace as this is really quite advanced.
2. All confidential clients, since they all need an assertion (even if we build one for them via the cert)

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
bgavrilMS: I'm thinking that we should move away from delegates and use interfaces instead. That delegate looks horrible. Will take your naming suggestion into consideration.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -0,0 +1,39 @@
...
+            where T : AbstractConfidentialClientAcquireTokenParameterBuilder<T>
+        {
+            if (clientAssertionAsyncDelegate == null)
+            {
+                throw new ArgumentNullException(nameof(clientAssertionAsyncDelegate));
+            }
+
+            builder.ValidateUseOfExperimentalFeature();
+            builder.CommonParameters.ClientAssertionOverrideDelegate = clientAssertionAsyncDelegate;
+
```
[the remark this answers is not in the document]
bgavrilMS: Request objects should not make changes to the app config. 

Imagine the sceanrio:

- app has no cert set
- request1 has WithClientAssertion (ok)
- request2 does not have WithClientAssertion - should throw

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System;
+
+namespace Microsoft.Identity.Client.Extensibility
+{
+
+    /// <summary>
+    /// Extensions for <see cref="AcquireTokenForClientBuilderExtensions"/> class
+    /// </summary>
+    public static class AcquireTokenForClientBuilderExtensions
```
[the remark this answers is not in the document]
bgavrilMS: This one I feel more strongly that it should not pollute the namespace. I don't really see a scenario where someone would want to do this.

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
bgavrilMS: These are not really related. One is about specifying a custom client assertion, and one is about taking over the POP kid.
Also, POP is only relevant for AcquireTokenForClient now.

src/client/Microsoft.Identity.Client/Extensibility/ExternalBoundTokenScheme.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Microsoft.Identity.Client.AuthScheme;
+using Microsoft.Identity.Client.Cache.Items;
+
+namespace Microsoft.Identity.Client.Extensibility
+{
+    internal class ExternalBoundTokenScheme : IAuthenticationScheme
```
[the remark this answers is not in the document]
bgavrilMS: I'll make the access token type an input param. MSAuth10POP does not change the auth scheme to pop, it keeps it bearer.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
```diff
@@ -23,5 +25,13 @@ internal class AcquireTokenCommonParameters
         public IAuthenticationScheme AuthenticationScheme { get; set; } = new BearerAuthenticationScheme();
         public IDictionary<string, string> ExtraHttpHeaders { get; set; }
         public PoPAuthenticationConfiguration PopAuthenticationConfiguration { get; set; }
+
+        /// <summary>
+        /// If set, the client assertion parameters should be ignored and these parameters should be set.
+        /// The input string is the token endpoint and a cancellation token.
+        /// The output is a list of name / value strings to be sent to the HTTP request to the token endpoint
+        /// </summary>
+        public Func<string, CancellationToken, Task<IReadOnlyList<KeyValuePair<string, string>>>> ClientAssertionOverrideDelegate { get; set; }
```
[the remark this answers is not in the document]
bgavrilMS: I agree. I will make a change. I was pondering between: 

1. A delegate as you mention
2. An interface

# The interface 

```csharp
public interface IClientAssertionProvider
{
    Task<IReadOnlyList<KeyValuePair<string, string>>> GetAssertionAsync(string clientId, string tokenEndpoint, CancellationToken ct);
}
```


```csharp
public Task<IReadOnlyList<KeyValuePair<string, string>>> GetClientAssertionParametersAsync(string clientId, string tokenEndpoint, CancellationToken cancellationToken)
    {
        Assert.AreEqual("https://login.microsoftonline.com/tid/oauth2/v2.0/token", tokenEndpoint);
        IReadOnlyList<KeyValuePair<string, string>> result = new List<KeyValuePair<string, string>>()
                {
                    new KeyValuePair<string, string>("client_assertion", CreateAssertion(clientId, tokenEndpoint)),
                    new KeyValuePair<string, string>("client_assertion_type", "jwt-beaer"),
                };
        return Task.FromResult(result);
    }
```

```csharp
public string GetToken()
{

    var assertionProvider = new MyAssertionProvider();
    var result = await cca.AcquireTokenForClient(scope).WithClientAssertion(assertionProvider).ExecuteAsync();
    return result;
}
```

# Delegate

```csharp
public string GetToken()
{
    var assertionProvider = new MyAssertionProvider();
    var result = await cca.AcquireTokenForClient(scope).WithClientAssertion(GetClientAssertionParametersAsync).ExecuteAsync();
    return result;
}

private static Task<IReadOnlyList<KeyValuePair<string, string>>> GetClientAssertionParametersAsync(string clientId, string tokenEndpoint, CancellationToken cancellationToken)
    {
        Assert.AreEqual("https://login.microsoftonline.com/tid/oauth2/v2.0/token", tokenEndpoint);
        IReadOnlyList<KeyValuePair<string, string>> result = new List<KeyValuePair<string, string>>()
                {
                    new KeyValuePair<string, string>("client_assertion", CreateAssertion(clientId, tokenEndpoint)),
                    new KeyValuePair<string, string>("client_assertion_type", "jwt-beaer"),
                };
        return Task.FromResult(result);
    }

```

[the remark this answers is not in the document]
bgavrilMS: I think I like the delegate way better, because if we need to support a different delegate the code will be simpler, whearas the interface we will have to create a new one.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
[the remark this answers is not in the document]
pmaytak: As discussed, I think it's ok to have it in extensibility namespace (especially if it's only to support legacy behavior).

src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -103,6 +103,7 @@ await Task.Delay(TimeSpan.FromSeconds(deviceCodeResult.Interval), cancellationTo
         {
             var dict = new Dictionary<string, string>
             {
+                [OAuth2Parameter.ClientInfo] = "1",
```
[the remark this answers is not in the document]
trwalke: This used to be here in the token client for all of our requests.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/31471d957e278465bf31e66a0ec6e6c1690f9f4b/src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs#L133

Any idea why it was removed? Should it be put back instead of adding it to the different requests?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/LegacyPopTest.cs
jmprieur: @bgavrilMS: Wouldn't it be better to just focus on the token acquisition but leave the protocol part of the extension, injecting a protocol handler?

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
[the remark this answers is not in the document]
jmprieur: And is `IReadOnlyList<KeyValuePair<string, string>` needed because we care about the order of the parameters?
Or would a `IReadonlyDictionary<string, string>` work as well?

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// A delegate which return returns the key - value parameters which will be added to the token endpoint when used in WithClientAssertion.
+    /// Implementers may use the input parameters clientId and tokenEndpoint which MSAL computes.
+    /// </summary>
+    /// <param name="clientId">The client id configured in the application. May be used as issuer claim in a JWT assertion.</param>
+    /// <param name="tokenEndpoint">The token endpoint where the request will be made. May be used as audience claim in a JWT assertion.</param>
+    /// <param name="cancellationToken">The cancellation token used by the token request, if any.</param>
+    /// <returns></returns>
+    /// <remarks>For a certificate based JWT assertion, see For the JWT-Bearer assertion format see https://docs.microsoft.com/en-gb/azure/active-directory/develop/active-directory-certificate-credentials#assertion-format</remarks>
+    public delegate Task<IReadOnlyDictionary<string, string>> ClientAssertionProviderAsync(
```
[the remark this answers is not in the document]
jmprieur: Did we discuss another more generic extensibility mechanism, where an extension could intervene in the MSAL pipeline, for instance close to sending the request by tweaking the protocol parameters that are going to be sent to the IdP (a bit like the HttpMessageHandlers/Delegatinghandlers offer with the HttpMessages)?

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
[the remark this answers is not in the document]
bgavrilMS: We do not care about the order indeed. I have already moved to using IReadonlyDictionary.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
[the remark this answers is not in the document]
bgavrilMS: Interesting idea, let me think about this @jmprieur

src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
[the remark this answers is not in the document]
bgavrilMS: I'll try to separate out a commit for this. We should not send this to ALL our requests, in particular for the perf critical client_credential request.

bgavrilMS:
Closing this PR in favour of a new one, taking into account the feedback.

PR #3068, opened 2021-12-13
Remove MigrationAid

bgavrilMS:
This helps folks migrate from MSAL 1.x, 2.x, 3.x to 4.x. Since 4.x has been in existence for more than 2 years, there does not seem to be a need for these files and they pollute IntelliSense.

The ratio of MSAL 3 to MSAL 4 apps is 0.0035 (as computed by apps that made calls in the last 24h)


trwalke (approved):

pmaytak (commented):
So a lot of methods are obsolete with error, so the purpose of this is someone migrates from old version to new, then get a bunch of errors with suggestions on what new methods to use? In this case removing this shouldn't be a breaking change? Is it worth looking at telemetry if MSAL v1-3 are used anymore?

The members in TokenCache.MigrationAid are obsoleted with warnings, though. Removing HasStateChanged actually broke the build. I guess, same questions apply here. Just want to make sure this doesn't break anyone.

PR #3069, opened 2021-12-13
Add POP kid to SuggestedCacheKey for AcquireTokenForClient

bgavrilMS:
Fixes # TBD

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
@@ -40,31 +40,6 @@ namespace Microsoft.Identity.Client
         }
 
         internal IConfidentialClientApplicationExecutor ConfidentialClientApplicationExecutor { get; }
-
-        /// <summary>
-        ///  Modifies the token acquisition request so that the acquired token is a Proof of Possession token (PoP), rather than a Bearer token. 
-        ///  PoP tokens are similar to Bearer tokens, but are bound to the HTTP request and to a cryptographic key, which MSAL can manage on Windows.
-        ///  See https://aka.ms/msal-net-pop
-        /// </summary>
-        /// <param name="popAuthenticationConfiguration">Configuration properties used to construct a proof of possession request.</param>
-        /// <returns>The builder.</returns>
-        /// <remarks>
-        /// <list type="bullet">
-        /// <item><description>An Authentication header is automatically added to the request</description></item>
-        /// <item><description>The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters).</description></item>
-        /// <item><description>MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</description></item>
-        /// <item><description>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</description></item>
-        /// </list>
-        /// </remarks>
-        public T WithProofOfPossession(PoPAuthenticationConfiguration popAuthenticationConfiguration)
-        {
-            ValidateUseOfExperimentalFeature();
-
-            CommonParameters.PopAuthenticationConfiguration = popAuthenticationConfiguration ?? throw new ArgumentNullException(nameof(popAuthenticationConfiguration));
-
-            CommonParameters.AuthenticationScheme = new PoPAuthenticationScheme(CommonParameters.PopAuthenticationConfiguration, ServiceBundle);
-
-            return this as T;
-        }
+        
     }
 }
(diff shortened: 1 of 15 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
```diff
@@ -42,32 +42,6 @@ public override Task<AuthenticationResult> ExecuteAsync(CancellationToken cancel
...
-        /// <returns>The builder.</returns>
-        /// <remarks>
-        /// <list type="bullet">
-        /// <item><description>An Authentication header is automatically added to the request</description></item>
-        /// <item><description>The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters).</description></item>
-        /// <item><description>MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</description></item>
-        /// <item><description>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</description></item>
-        /// </list>
-        /// </remarks>
-        public T WithProofOfPossession(PoPAuthenticationConfiguration popAuthenticationConfiguration)
```
bgavrilMS, line 61: I moved `WithProofOfPossession` to AcquireTokenForClient only. Would be a breaking change if not for experimental flag.

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -76,18 +77,23 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
...
             key = null;
             return false;
         }
 
-        public static string GetClientCredentialKey(string clientId, string tenantId)
-        {                
-            return $"{clientId}_{tenantId}_AppTokenCache";
+        public static string GetClientCredentialKey(string clientId, string tenantId, string popKid)
+        {
+            return string.Format(
```
bgavrilMS, line 91: string.Format invariant might a tad faster than $"" which is culture sensitive.

tests/Microsoft.Identity.Test.Unit/pop/PopAuthenticationSchemeTests.cs
```diff
@@ -140,7 +139,7 @@ public async Task ValidateKeyExpirationAsync()
                 var initialToken = result.AccessToken;
 
                 //Advance time 7 hours. Should still be the same key and token
-                testClock.TestTime +=TimeSpan.FromHours(7);
+                testClock.UtcNow +=TimeSpan.FromHours(7);
```
trwalke, line 142: The name Utsnow on the test clock seems a bit misleading. Seems like it should always be the current time but it changes as modifications are made to it. Would be better to name it UtcCurrent or UtcTestTime.

trwalke:
>         public DateTime GetUtcNow()

Do we need the method and the parameter?

---
Refers to: tests/Microsoft.Identity.Test.Common/Core/Mocks/TestTimeService.cs:23 in 23862eb. [](commit_id = 23862eb71ae0a0e619384bdfc514535489a51d8d, deletion_comment = False)

tests/Microsoft.Identity.Test.Unit/pop/CryptoProviderTests.cs
```diff
@@ -5,7 +5,7 @@
 using Microsoft.Identity.Json.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace Microsoft.Identity.Test.Unit.PoP
+namespace Microsoft.Identity.Test.Unit.pop
```
trwalke, line 8: Should this be Pop instead?

trwalke:
> using Microsoft.Identity.Client.AuthScheme.PoP;

Should this also be Pop for the source code?

---
Refers to: tests/Microsoft.Identity.Test.Unit/pop/CryptoProviderTests.cs:4 in 23862eb. [](commit_id = 23862eb71ae0a0e619384bdfc514535489a51d8d, deletion_comment = False)

trwalke (approved):
Makes sense to me.

pmaytak:
@bgavrilMS Does this need to be rebased properly from the other pop branch?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -529,6 +530,79 @@ public async Task CertificateOverrideAsync()
             }
         }
 
+        [TestMethod]
+        public async Task CertificateOverrideAsync()
```
pmaytak, line 534: This seems to be an old method, should be removed.
Also cert on line 482 above is not actually added to the app.

pmaytak (approved):

Issue #2976, opened 2021-10-25
[Feature Request] Allow disabling TokenCache for Confidential clients


johnthcall:
**Is your feature request related to a problem? Please describe.**
My team uses confidential clients across millions of tenants and is currently creating a new confidential client application each request. We have been asked to start using a static confidential client application but have some reservations about the token cache implementation. It appears that the confidential client token cache has no eviction which will cause memory pressure in our service over time. We can obviously build the cache from the output of AcquireTokenForClient or via serialization events from ITokenCache but it appears in either case that the underlying token cache would still be populated.

**Describe the solution you'd like**
Add the ability for requests to configure that the entry should not be written to the built-in token cache or to configure the same on ConfidentialClientApplicationBuilder.

**Describe alternatives you've considered**
Continue using a unique confidential client application for each request and take the hit of regional discovery in each request when using "WithAzureRegion"

bgavrilMS:
@johnthcall - if you want to maintain your own cache, for perf reasons, we understand. Ideally we would get MSAL into a place where it is more performant, but this will take time. 

If the token caches here https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-net-token-cache-serialization?tabs=aspnet are not good enough, you can always implement your own token caching mechanism outside MSAL. 

There is no need to disable MSAL's token cache. Just don't configure any L2 token caching and let the `ConfidentialClientApplicationObjects` go out of scope.

PR #3070, opened 2021-12-13
Add support for cleanup of expired access token

johnthcall:
Adds support for expiring access tokens as suggested in #2976. Initially this only is being done on AppTokenCache but if the change makes sense I can push changes for UserTokenCache.

Adds nullable property `AccessTokenExpirationScanFrequency` in `CacheOptions`
Adds logic in `InMemoryPartitionedAppTokenCacheAccessor` to fire off a cleanup task when tokens are saved or retrieved and a scan hasn't recently occurred.

**Testing**
Adds a unit test which verifies the cleanup logic works as expected.

**Performance impact**
Ran Performance benchmark before and after and didn't see any regressions

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+
 namespace Microsoft.Identity.Client
 {
     /// <summary>
@@ -34,9 +36,16 @@ namespace Microsoft.Identity.Client
         /// Constructor
         /// </summary>
         /// <param name="useSharedCache">Set to true to share the cache between all ClientApplication objects. The cache becomes static. <see cref="UseSharedCache"/> for a detailed description. </param>
-        public CacheOptions(bool useSharedCache)
+        /// <param name="accessTokenExpirationScanFrequency">Sets to non-null to enable expiration of access tokens. <see cref="AccessTokenExpirationScanFrequency"/> for a detailed description. </param>
+        public CacheOptions(bool useSharedCache, TimeSpan? accessTokenExpirationScanFrequency = null)
         {
+            if (accessTokenExpirationScanFrequency.HasValue && accessTokenExpirationScanFrequency.Value <= TimeSpan.Zero)
+            {
+                throw new ArgumentException("accessTokenExpirationScanFrequency must be a positive TimeSpan", nameof(accessTokenExpirationScanFrequency));
+            }
+
             UseSharedCache = useSharedCache;
+            AccessTokenExpirationScanFrequency = accessTokenExpirationScanFrequency;
         }
 
         /// <summary>
@@ -49,6 +58,11 @@ namespace Microsoft.Identity.Client
         /// ADAL used a static cache by default.
         /// </remarks>
         public bool UseSharedCache { get; set; }
+
+        /// <summary>
+        /// The minimum time between scans for expired access tokens. Defaults to null meaning no expiration scans are performed.
+        /// </summary>
+        public TimeSpan? AccessTokenExpirationScanFrequency { get; }
 
     }
 }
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -215,14 +224,84 @@ public void SetiOSKeychainSecurityGroup(string keychainSecurityGroup)
...
+            private static void ScanForExpiredItems(CacheWrapper cacheWrapper)
+            {
+                DateTimeOffset now = cacheWrapper._lastExpirationScan = DateTimeOffset.UtcNow;
+
+                foreach (KeyValuePair<string, ConcurrentDictionary<string, MsalAccessTokenCacheItem>> partitionItem in cacheWrapper.AccessTokenCacheDictionary)
+                {
+                    var partitionItemValue = partitionItem.Value;
+                    foreach (KeyValuePair<string, MsalAccessTokenCacheItem> item in partitionItemValue)
+                    {
+                        if (item.Value.ExpiresOn < now)
```
johnthcall, line 294: Not sure if this is the only thing we should check for before removing items.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -215,14 +224,84 @@ public void SetiOSKeychainSecurityGroup(string keychainSecurityGroup)
...
+                        {
+                            partitionItemValue.TryRemove(item.Key, out _);
```
bgavrilMS, line 296: I'll let @pmaytak comment on this, but this implementation would not trigger an L2 cache update. 

We're thinking of adding proper L1 / L2 cache support in MSAL, with eviction options.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -215,14 +224,84 @@ public void SetiOSKeychainSecurityGroup(string keychainSecurityGroup)
...
             // app metadata isn't removable
         }
 
         public virtual bool HasAccessOrRefreshTokens()
         {
-            return AccessTokenCacheDictionary.Any(partition => partition.Value.Any(token => !token.Value.IsExpiredWithBuffer()));
+            return AccessTokenCacheWrapper.AccessTokenCacheDictionary.Any(partition => partition.Value.Any(token => !token.Value.IsExpiredWithBuffer()));
+        }
+
+        internal class CacheWrapper
```
trwalke, line 237: Should probably move to separate class

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -34,9 +36,16 @@ public CacheOptions()
...
+        /// <param name="accessTokenExpirationScanFrequency">Sets to non-null to enable expiration of access tokens. <see cref="AccessTokenExpirationScanFrequency"/> for a detailed description. </param>
+        public CacheOptions(bool useSharedCache, TimeSpan? accessTokenExpirationScanFrequency = null)
         {
+            if (accessTokenExpirationScanFrequency.HasValue && accessTokenExpirationScanFrequency.Value <= TimeSpan.Zero)
+            {
+                throw new ArgumentException("accessTokenExpirationScanFrequency must be a positive TimeSpan", nameof(accessTokenExpirationScanFrequency));
+            }
+
             UseSharedCache = useSharedCache;
+            AccessTokenExpirationScanFrequency = accessTokenExpirationScanFrequency;
```
trwalke, line 48: @bgavrilMS does it make sense to set this to a default value to avoid caches expanding indefinitely? Also, is there a recommended value for this?

bgavrilMS:
For now, we are rejecting this as we are working on having an actual L1 cache with eviction backed into MSAL itself.

PR #3074, opened 2021-12-14
[Internal] Ignore mobile and legacy target on env variable

bgavrilMS:
Internal improvement

Set `MSAL_DESKTOP_ONLY_DEV` and msbuild will ignore the mobile targets and net45. No more need to comment targets out, just set an env variable (and restart VS).


jmprieur (approved):
LGTM
I love this one!
Thanks @bgavrilMS

pmaytak (approved):
👍

PR #3075, opened 2021-12-16
Add descriptive x5c comment to WithCertificate

pmaytak:
Fixes # 

Update X5C related comments in WithSendX5C and WithCertificate.

**Testing**

**Performance impact**


SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -94,9 +94,15 @@ public ConfidentialClientApplicationBuilder WithCertificate(X509Certificate2 cer
 
         /// <summary>
         /// Sets the certificate associated with the application.
+        /// This method allows to specify if the x5c claim (public key of the certificate) should be sent to Azure AD.
```
bgavrilMS, line 97: 1. It's not the public key of the cert that's being sent, it's the certificate chain. Please reference https://datatracker.ietf.org/doc/html/rfc7517#section-4.7 
2. As far as I know, this is only available for 1st parties. This should be highlighted 
3. The URL is internal and we'll get flagged for posting internal stuff. Can you add or modify a page in our wiki instead where it just states "This is for 1st party Microsoft apps only. For details please see ..."

bgavrilMS (approved):

bgavrilMS:
Approved with comments. Main thing is to highlight that X5C is for 1st party only (despite the misleading internal docs)

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
pmaytak: 3. What's going to flag us? Even if it's aka.ms link, and not a direct link? 
So update aka.ms link to point to a new SNI wiki page which has a link to internal SNI doc?

bgavrilMS: An aka.ms link is fine, but not if it points to an internal doc. All aka.ms link that go onto the public docs must point to public pages. We started getting docs bugs because of this. 

You can point to https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/60

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
pmaytak: Created a wiki: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Subject-Name-and-Issuer-(SNI)-Authentication

PR #3082, opened 2021-12-21
Output the date in Universal format in the logs

bgavrilMS:
This will make it easier to find logs in LogsMiner

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
@@ -192,7 +192,19 @@ namespace Microsoft.Identity.Client.Internal.Logger
                 bool isLoggingPii = messageWithPiiExists && PiiLoggingEnabled;
                 string messageToLog = isLoggingPii ? messageWithPii : messageScrubbed;
 
-                string log = $"{isLoggingPii} MSAL {s_msalVersionLazy.Value} {s_skuLazy.Value} {s_runtimeVersionLazy.Value} {s_osLazy.Value} [{DateTime.UtcNow:MM/dd HH:mm:ss.ff}{_correlationId}]{ClientInformation} {messageToLog}";
+                string log = string.Format(
+                    CultureInfo.InvariantCulture,
+                    "{0} MSAL {1} {2} {3} {4} [{5}{6}]{7} {8}",
+                    isLoggingPii,
+                    s_msalVersionLazy.Value,
+                    s_skuLazy.Value,
+                    s_runtimeVersionLazy.Value,
+                    s_osLazy.Value,
+                    DateTime.UtcNow.ToString("u"),
+                    _correlationId,
+                    ClientInformation,
+                    messageToLog);
+
 
                 if (_isDefaultPlatformLoggingEnabled)
                 {


src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
```diff
@@ -192,7 +192,19 @@ public void Log(LogLevel logLevel, string messageWithPii, string messageScrubbed
...
-                string log = $"{isLoggingPii} MSAL {s_msalVersionLazy.Value} {s_skuLazy.Value} {s_runtimeVersionLazy.Value} {s_osLazy.Value} [{DateTime.UtcNow:MM/dd HH:mm:ss.ff}{_correlationId}]{ClientInformation} {messageToLog}";
+                string log = string.Format(
+                    CultureInfo.InvariantCulture,
+                    "{0} MSAL {1} {2} {3} {4} [{5}{6}]{7} {8}",
+                    isLoggingPii,
+                    s_msalVersionLazy.Value,
+                    s_skuLazy.Value,
+                    s_runtimeVersionLazy.Value,
+                    s_osLazy.Value,
+                    DateTime.UtcNow.ToString("u"),
```
bgavrilMS, line 203: Date in universal format which LogsMIner uses.

SameerK-MSFT: minor: Can this have string interpolation. That will make it easier to read the code with these many parameters.

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
```diff
@@ -192,7 +192,19 @@ public void Log(LogLevel logLevel, string messageWithPii, string messageScrubbed
                 bool isLoggingPii = messageWithPiiExists && PiiLoggingEnabled;
                 string messageToLog = isLoggingPii ? messageWithPii : messageScrubbed;
 
-                string log = $"{isLoggingPii} MSAL {s_msalVersionLazy.Value} {s_skuLazy.Value} {s_runtimeVersionLazy.Value} {s_osLazy.Value} [{DateTime.UtcNow:MM/dd HH:mm:ss.ff}{_correlationId}]{ClientInformation} {messageToLog}";
+                string log = string.Format(
+                    CultureInfo.InvariantCulture,
+                    "{0} MSAL {1} {2} {3} {4} [{5}{6}]{7} {8}",
```
pmaytak, line 197: Would people actually search by the date in this log string, since this string would have to be parsed? In Kusto, this string goes into a message column, and then there's a separate date time column, which is used for filtering I think.

pmaytak (approved):
+1 for better readability.

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
pmaytak: Along with that, if the date is parsed out of this whole log string usually, then changing it would break people's current parsing logic?

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
bgavrilMS: The reason I moved away from string interpolation to string.format is the culture. By default string interpolation uses current culture. This is not what we want and is a bit slower to process than invariant culture. 

There is a specify string interpolation's culture, but not in net45

https://stackoverflow.com/questions/33203261/what-is-the-default-culture-for-c-sharp-6-string-interpolation

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
bgavrilMS: If you use LogsMiner, they expect the data to be in this format. Also AAD sends timestamp in this format as well. 
I understand that some ppl might parse the logs, but we don't gurantee backwards compat at this level.

 There is an ask to provide a structured log, which we should investigate as part of the logging improvements issue

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLogger.cs
bgavrilMS: Note: we probably don't need to do this everywhere, but this method is perf critical.

PR #3104, opened 2022-01-12
[Prototype] Add cache size limit and eviction.

### LINKED ISSUE #3020: [Feature Request] MSAL internal cache eviction
# Goal

MSAL should support cache eviction options (ideal: max memory size; less than ideal: max number of items)

## North star

1. MSAL should support some eviction options (ideal: max memory size; less than ideal: max number of items)
2. L1 must be really fast
3. MSAL should allow app developers to provide their own caching mechanism (e.g. MemoryCache)
4. L1 / L2 observability (availability & hit ratios)

## Status Quo 

Today, MSAL does not have an L1 layer. Its cache operations are: 

```
1. Find
[optional] Fetch data from L2 into memory
Apply filtering logic and find token

2. Save
Get response from ESTS
Merge response with tokens in memory
[optional] Save data to L2
```

## Ideas

- We could satisfy item 1 with a tactical fix. If L2 is not defined, MSAL will use it accessors to accumulate tokens. We can gain inspiration from MemoryCache implementation on how to compute size. When size is reached, evict the entire cache. 
- Accessors cannot be implemented with MemoryCache because MemoryCache does not support "GetAllItems" operation.

- If we completely re-implement the Find 

```
1. Find
Fetch data from L1. If not hit, fetch data from L2 into accessors.
L1 is just a generic map[string] -> object, where string = cache key and object = accessor
```

- `SetCacheOptions(L1Options)` where `L1Options` defines: `Enabled`, `Static`, `MaxItems`, `MaxSize` (?), and in the future a `IL1CacheInterface`
- L2 cache continues to be defined as today
- With a proper L1, there is no need for partitioned accessors, although these can stay.

pmaytak:
Fixes #3020

- Added size limit option to the CacheOptions public API, which represent cache size in bytes.
- Added cache capacity properties to the user and app cache structures and a property that tracks combined cache size to a parent level TokenCache class.
- These properties are updated during add and remove operations. Constant token size is used. - 
- The capacity overflow check is done when adding access tokens.
- If capacity is reached, cache is cleared, app or user, depending on which operation triggered it.

Issues:
- It is difficult to keep the current cache size in sync in multi-threaded environments. When clearing the cache, after it hits capacity, a lock is needed to make sure the current size counter is not updated by other cache methods, since it’s set to 0. Locks can cause performance issues, as in past cases. Current cache size tracker also cannot be calculated dynamically based on the count of dictionary elements, since the storage structure is not a simple dictionary, but a nested one. 
- With the current cache structure, there are app and user cache accessors, which are separate from each other. So, a cache size tracker is needed in a class above them, which increases the complexity of keeping track of the three capacity variables. 

**Testing**
To be added.

**Performance impact**
Should be no impact - in add/remove accessor methods, there're a few O(1) dictionary calls.


src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -39,6 +39,17 @@ public CacheOptions(bool useSharedCache)
             UseSharedCache = useSharedCache;
         }
 
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="useSharedCache">Set to true to share the cache between all ClientApplication objects. The cache becomes static. <see cref="UseSharedCache"/> for a detailed description. </param>
+        /// <param name="sizeLimit">Token cache size limit in bytes. <see cref="SizeLimit"/> for a detailed description.</param>
+        public CacheOptions(bool useSharedCache, long sizeLimit)
```
SameerK-MSFT, line 47: Can we have default values here (as per any recommended practices)?

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -39,6 +39,17 @@ public CacheOptions(bool useSharedCache)
...
+        {
+            UseSharedCache = useSharedCache;
+            SizeLimit = sizeLimit;
+        }
```
SameerK-MSFT, line 51: Suggestion: Declare ATSizeInBytes here as internal static and validate that the value passed is greater than min required.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -50,5 +61,12 @@ public CacheOptions(bool useSharedCache)
         /// </remarks>
         public bool UseSharedCache { get; set; }
 
+        /// <summary>
+        /// Token cache size limit in bytes.
+        /// </summary>
+        /// <remarks>
+        /// Once the limit is reached, the whole cache is cleared.
+        /// </remarks>
+        public long? SizeLimit { get; set; }
```
SameerK-MSFT, line 70: Why is it nullable?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -216,13 +234,25 @@ public void SetiOSKeychainSecurityGroup(string keychainSecurityGroup)
         public virtual void Clear()
         {
             AccessTokenCacheDictionary.Clear();
```
SameerK-MSFT, line 236: Threading issue: Between Clear and setting CacheSize to zero, something may get added to the dictionary and the two can go out of sync

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -216,13 +234,25 @@ public void SetiOSKeychainSecurityGroup(string keychainSecurityGroup)
...
         {
             return AccessTokenCacheDictionary.Any(partition => partition.Value.Any(token => !token.Value.IsExpiredWithBuffer()));
         }
+
+        private bool IsOverCapacity()
+        {
+            return _tokenCacheAccessorOptions.SizeLimit.HasValue && (Interlocked.Read(ref _cacheSize) + ATSizeInBytes) > _tokenCacheAccessorOptions.SizeLimit;
+        }
+
+        private void Compact()
```
SameerK-MSFT, line 252: Do we need Compact method? It just calls clear. (logging can be done by the caller.)

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
pmaytak: In token cache we check, if it's null, then size limit doesn't apply.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
pmaytak: I don't think we can provide a realistic default value; it would be based on the user's scenario.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: Called it `Compact` as a form of abstraction, clearing is just an implementation.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: True, but we also don't lock on cache operations in general (based on eventual consistency principle). So this behavior can happen with other add/remove methods.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -50,5 +65,35 @@ public CacheOptions(bool useSharedCache)
...
+        /// </summary>
+        /// <remarks>
+        /// Once the limit is reached, either the app or user cache will be fully cleared, depending on which was most recently used.
+        /// MSAL doesn't calculate the exact memory usage and uses approximations of the token sizes. 
+        /// For instance, app token cache entry is approximately at least 4500 bytes; user access token entry - 6500 bytes,
+        /// user refresh token entry - 3700 bytes.
+        /// IMPORTANT: Monitor app health metrics (including memory usage) and cache performance (<see href="https://aka.ms/msal-net-token-cache-serialization"/>)
+        /// and adjust size limit accordingly.
+        /// </remarks>
+        public long? SizeLimit
```
pmaytak, line 81: Size limit applies only when cache serialization is disabled.

src/client/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -47,6 +47,11 @@ public sealed partial class TokenCache : ITokenCacheInternal
         /// </summary>
         internal bool UsesDefaultSerialization { get; set; } = false;
 
+        /// <summary>
+        /// Static, used by both app and user caches to track approximate token cache size.
+        /// </summary>
+        internal static long CacheSize = 0L;
```
pmaytak, line 53: Cache size includes user+ app token cache. 2 instances of TokenCache class are created (1 with app cache accessors, 1 with user accessors). So moved this size property up a level so it can be accessed by both.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
bgavrilMS: Yeah, that's a good question we're struggling with :). What would be a good default? ...

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -39,6 +41,19 @@ public CacheOptions(bool useSharedCache)
             UseSharedCache = useSharedCache;
         }
 
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="useSharedCache">Set to true to share the cache between all ClientApplication objects. The cache becomes static. <see cref="UseSharedCache"/> for a detailed description. </param>
+        /// <param name="sizeLimit">Token cache size limit in bytes. <see cref="SizeLimit"/> for a detailed description.</param>
+        public CacheOptions(bool useSharedCache, long sizeLimit)
```
bgavrilMS, line 49: `countLimit` ?

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -50,5 +65,35 @@ public CacheOptions(bool useSharedCache)
...
+            set
+            {
+                ValidateSizeLimit(value);
+                _sizeLimit = value;
+            }
+        }
+
+        private void ValidateSizeLimit(long? sizeLimit)
+        {
+            if (!sizeLimit.HasValue || sizeLimit < 0)
```
bgavrilMS, line 93: Why not allow users to set NULL ?

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -50,5 +65,35 @@ public CacheOptions(bool useSharedCache)
         /// </remarks>
         public bool UseSharedCache { get; set; }
 
+        private long? _sizeLimit;
+
+        /// <summary>
+        /// Total token cache size limit in bytes for both app and user token caches.
```
bgavrilMS, line 71: Please add a link to the caching doc and also explain that this is not a concern if distributed caching is used.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -35,6 +36,11 @@ internal class InMemoryPartitionedAppTokenCacheAccessor : ITokenCacheAccessor
         protected readonly ICoreLogger _logger;
         private readonly CacheOptions _tokenCacheAccessorOptions;
 
+        // Approximate size of cache item objects
+        private const long AccessTokenSizeInBytes = 4500;
```
bgavrilMS, line 40: My main concern is that a constant approximation will not work well: 

1. AAD allows for configuration of optional claims
2. POP tokens are a bit bigger because they contain the POP key id
3. AAD is adding support for SAML tokens (there is already an OBO flow for this). We won't be able to update this aproximation for all token types.

I think there are a few ways out of this: 

1. Allow customers to define this constant (this is what SAL does). But this is a pretty "obscure" scenario
2. Use non-constant size approximation, i.e. we know the size of the actual token (it's a string) at which we add some metadata (fairly constant)
3. Do not expose sizeLimit, expose countLimit

bgavrilMS: @dannybtsai @henrik-me - can you please review this strategy?

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
SameerK-MSFT: If default varies widely based on the user scenarios, we cannot do much. But if it primarly leans towards a certain value, that can be used. But we do not have answer, then it is good as it is.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedUserTokenCacheAccessor.cs
```diff
@@ -70,38 +79,77 @@ public InMemoryPartitionedUserTokenCacheAccessor(ICoreLogger logger, CacheOption
         #region Add
         public void SaveAccessToken(MsalAccessTokenCacheItem item)
         {
+            if (IsCacheOverCapacity(AccessTokenSizeInBytes))
+            {
+                _logger.Always("[UserCache] Cache is over capacity.");
+                Compact();
```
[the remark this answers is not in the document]
SameerK-MSFT: Good explanation. Please add that as comment to the code. So person who reads it, will understand better.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
dannybtsai, line 93: >sizeLimit [](http://example.com/codeflow?start=39&length=9)

Should we have a min value? Does 0 or a small number (1, 2, or 3) make sense?

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -39,6 +41,19 @@ public CacheOptions(bool useSharedCache)
...
+        /// <param name="useSharedCache">Set to true to share the cache between all ClientApplication objects. The cache becomes static. <see cref="UseSharedCache"/> for a detailed description. </param>
+        /// <param name="sizeLimit">Token cache size limit in bytes. <see cref="SizeLimit"/> for a detailed description.</param>
+        public CacheOptions(bool useSharedCache, long sizeLimit)
+        {
+            ValidateSizeLimit(sizeLimit);
+
+            UseSharedCache = useSharedCache;
+            _sizeLimit = sizeLimit;
+        }
+
```
dannybtsai, line 56: This can be written as the following to reuse existing logic:

        public CacheOptions(bool useSharedCache, long sizeLimit) : this(useSharedCache)
        {
            SizeLimit = sizeLimit;
        }

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
dannybtsai: I agree with Peter using Compact() instead of Clear() directly.

dannybtsai:
>     internal class InMemoryPartitionedAppTokenCacheAccessor : ITokenCacheAccessor

Why Compact() clears the entire cache instead of reducing the cache by a certain percentage and still keeping some?

---
Refers to: src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs:23 in 64f8718. [](commit_id = 64f87188794bef3e4a95560ed530ed6758ac308c, deletion_comment = False)

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
dannybtsai: Is there is reason to use a constant value instead of the actual size (the #2 option)?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
dannybtsai: Eventual consistency is not guaranteed here. For example:
1. Thread 1 calls the AccessTokenCacheDictionary.Clear() method and is paused.
2. Thread 2 adds an item and updates the _appCacheSize.
3. Thread 1 resumes execution and sets the appCacheSize to 0 (Interlocked.Exchange(ref _appCacheSize, 0)).
4. Now there is one item in the cache but the size is 0.
5. The same issue applies to the TokenCache.CacheSize, and in this case the error can accumulate over multiple Clear() calls.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
pmaytak: Yea, I mean if they don't want to set the limit, use the other constructor. I guess null is technically valid here.

pmaytak: Related discussion here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3104#discussion_r783281241

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
pmaytak: Danny also touched on a point of should there bit a minimum. Currently if someone specifies small value (like 1), cache will be compacted on each operation.

pmaytak:
> > ```
> > internal class InMemoryPartitionedAppTokenCacheAccessor : ITokenCacheAccessor
> > ```
> 
> Why Compact() clears the entire cache instead of reducing the cache by a certain percentage and still keeping some? #Closed
> 
> Refers to: src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs:23 in [64f8718](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/64f87188794bef3e4a95560ed530ed6758ac308c). [](commit_id = [64f8718](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/64f87188794bef3e4a95560ed530ed6758ac308c), deletion_comment = False)

Basically clearing full cache is simpler.  With the user cache, there are multiple token types saves in different dictionaries, so clearing them in a sensible manner becomes complex. I documented this in a doc, I'll share with the team.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: @bgavrilMS 
1. Hmm, interesting, I'd like to see the implementation in SAL. How would customers know that constant?
2. Yea, metadata should be fairly constant. This does seem like a better approximation.
3. Specific to the user cache, what would the count represent: only access tokens, all tokens? Because in both cases, the count would not represent the whole cache size accurately.

@dannybtsai 
How would you measure the actual size of objects in memory? We could measure the strings size in cache objects like Bogdan mentioned below and in Microsoft.Identity.Web we serialize the whole cache, so take size that way. But in this case cache entry is an object.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: @dannybtsai Do you suggest locking? Only when clearing or always when updating the cache size?

dannybtsai: It is hard to achieve consistency without locking as far as I can tell. And you need to always lock, not just when clearing cache otherwise it is the same problem.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -57,12 +63,26 @@ internal class InMemoryPartitionedAppTokenCacheAccessor : ITokenCacheAccessor
...
+                Compact();
+            }
+
             string itemKey = item.GetKey().ToString();
             string partitionKey = CacheKeyFactory.GetClientCredentialKey(item.ClientId, item.TenantId, item.KeyId);
 
+            // Update cache size only if cache item is added, not updated
+            if (!AccessTokenCacheDictionary.TryGetValue(partitionKey, out var partition) || !partition.TryGetValue(itemKey, out _))
+            {
+                Interlocked.Add(ref _appCacheSize, AccessTokenSizeInBytes);
```
dannybtsai, line 78: >_appCacheSize [](http://example.com/codeflow?start=36&length=13)

Is it really necessary to use the variable to track the sizes? Isn't _appCacheSize == AccessTokenCacheDictionary.Count * AccessTokenSizeInBytes?

There seems to be some risks of not setting the sizes correctly even Interlock() is atomic. And it is more complicated to set them in the Clear() method without locking.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -57,12 +63,26 @@ internal class InMemoryPartitionedAppTokenCacheAccessor : ITokenCacheAccessor
...
+                Interlocked.Add(ref TokenCache.CacheSize, AccessTokenSizeInBytes);
```
dannybtsai, line 79: >TokenCache.CacheSize [](http://example.com/codeflow?start=36&length=20)

Can this be calculated?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: `ConcurrentDictionary<string, ConcurrentDictionary<string, MsalAccessTokenCacheItem>> AccessTokenCacheDictionary`
TokenCacheDictionaries are dictionary of dictionaries. with the outer key being a partition key and an inner key is an actual cache item key. So unfortunately count does not work.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
pmaytak: Hmm, I don't think so. There are two instances of TokenCache class created and then one instance has app token cache accessor and one has user accessor, but they don't know about each other. And I don't think CacheSize can be moved into a higher parent class.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
dannybtsai: If AccessTokenSizeInBytes is added to the _appCacheSize every time a token is saved, isn't _appCacheSize == AccessTokenCacheDictionary.Count * AccessTokenSizeInBytes?

bgavrilMS:
Can we close this one @pmaytak ?

Issue #3014, opened 2021-11-15
[Bug] WithAzureRegion and WithCustomInstanceMetadata are mutually exclusive


bgavrilMS:
msal 4.37

1. Configure both WithAzureRegion("region") and `WithCustomInstanceMetadata(metadata)`
2. set medatadata to smth like 

```json
                "{{\"api-version\":\"1.1\",\"metadata\":[{{\"preferred_network\":\"{0}\",\"preferred_cache\":\"{0}\",\"aliases\":[\"{0}\",]}}]}}";
```

**Actual**: WithCustomInstanceMetadata overrides with region 
**Expected**: An exception message preventing the use of both options

PR #3116, opened 2022-01-18
Add exception when both region and custom instance metadata is config…

neha-bhargava:
Fixes # 
#3014 

Add exception when both region and custom metadata are configured

**Testing**
Added integration test

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
+++ b/src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
@@ -44,6 +44,11 @@ namespace Microsoft.Identity.Client
         {
             GuardMobileFrameworks();
 
+            if (!(string.IsNullOrEmpty(configuration.AzureRegion) || configuration.CustomInstanceDiscoveryMetadata == null))
+            {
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -44,6 +44,11 @@ public sealed partial class ConfidentialClientApplication
         {
             GuardMobileFrameworks();
 
+            if (!(string.IsNullOrEmpty(configuration.AzureRegion) || configuration.CustomInstanceDiscoveryMetadata == null))
```
bgavrilMS, line 47: Try to use the Validate method on the CCABuilder class instead for this check.

bgavrilMS, line 47: Isn't there an overload to custom instance metadata with URI?

bgavrilMS (approved):
Small changes needed.

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
neha-bhargava: Yeah.. included it.

trwalke (approved):
Looks good, Thanks @neha-bhargava

pmaytak (approved):

PR #3118, opened 2022-01-20
GetAccountAsync fix by checking for null on accountId

gladjohn:
Fixes # 

GetAccountAsync fix by checking for null on accountId

**Testing**
unit tests

**Performance impact**
n/a

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
@@ -1561,6 +1561,52 @@ namespace Microsoft.Identity.Test.Unit.PublicApiTests
                 Assert.IsTrue(log.Contains(MsalErrorMessage.ClientCredentialWrongAuthority));
             }
         }
+
+        [TestMethod]
+        public async Task ValidateGetAccountAsyncWithNullAccountIdAsync()
+        {
+            using (var httpManager = new MockHttpManager())
+            {
+                var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1561,6 +1561,52 @@ public async Task AcquireTokenForClientAuthorityCheckTestAsync()
                 Assert.IsTrue(log.Contains(MsalErrorMessage.ClientCredentialWrongAuthority));
             }
         }
+
+        [TestMethod]
```
bgavrilMS, line 1565: nit: the 2 tests are identical except that one passes in "" and one passes in null as acc_id. To avoid code duplication, you can use a parametrized test, smth like: 

```csharp
[TestMethod]
[DataRow("")]
[DataRow(null)]
public async Task ValidateGetAccountAsyncWithNullAccountIdAsync(string accountId)
{
 // test logic
}
```

Of course, you can also just do: 

```csharp
var acc = await app.GetAccountAsync("").ConfigureAwait(false);
// assert
var acc2 = await app.GetAccountAsync(null).ConfigureAwait(false);
```

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
gladjohn: Thanks @bgavrilMS, updated tests per suggestion.

trwalke (approved):
Looks good gladwin

Issue #2524, opened 2021-04-01
[Feature Request] instance aware support


henrik-me:
[Feature 824881: [MSAL.NET] Sovereign cloud aware is supported by MSAL](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/824881)

If a client indicates that it's "instance_aware" then EVO will authenticated against the cloud associated with the account attempting to authenticate and will instruct the client library (MSAL) where to redeem the resulting authorization code for a token.

https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/1450?_a=files&path=%2F%5BXplat%5D%20Instance%20Aware%2Finstance_aware_flows.md

Auth Client spec: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2F%5BXplat%5D%20Instance%20Aware%2Finstance_aware_flows.md&_a=contents&version=GBdev

**.js implementation details:**
MSAL.js v2: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/1804/files
Docs: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/instance-aware.md

pmaytak:
- [ ] Interactive flow: In `AuthCodeRequestComponent.FetchAuthCodeAndPkceInternalAsync`, if `cloud_instance_host_name` exists in auth code response, return it and in `InteractiveRequest.GetTokenResponseAsync` update the token endpoint in the request parameters.
- [ ] WIA: In `IntegratedWindowsAuthRequest.FetchAssertionFromWsTrustAsync`, if response returns `Unknown` and `cloud_instance_host_name`, create a new request and send it to the updated host.
- [ ] Username/Password: Same as for WIA but in `UsernamePasswordRequest.FetchAssertionFromWsTrustAsync`.
- [ ] Test: Interactive request with `instance_aware` calls correct token endpoint and saves in the cache correctly, then silent request retrieves from cache correctly.
- [ ] Test: Same for Windows Integrated Auth flow
- [ ] Test: Same for Username/Password flow
- [ ] Create wiki page

jennyf19:
Makes sense for PCA, but what about confidential client? The work has been done on android/ios, but they only support PCA, as noted in the above list, all PCA related concepts. I thought our initial concern was for impact on CCA...?

bgavrilMS:
[...]

PR #3141, opened 2022-02-04
Multi cloud support for interactive requests for web

neha-bhargava:
Fixes # 
#2524 

Support instance aware for interactive and silent flows
* Add instance aware to interactive requests
* Modify GetAccounts to not filter with env when MultiCloudSupport is enabled
* Silent requests to pass authority from Account returned in auth result
* Integration tests

**Testing**
Integration tests

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AccountExtensions.cs
+++ b/src/client/Microsoft.Identity.Client/AccountExtensions.cs
@@ -23,5 +23,14 @@ namespace Microsoft.Identity.Client
         {
             return (account as Account)?.TenantProfiles;
         }
+
+        /// <summary>
+        /// Returns the authority from which the account is acquired.
+        /// </summary>
+        /// <returns>Authority for the account.</returns>
+        public static string GetAuthority(this IAccount account)
+        {
+            return $"https://{account.Environment}/{account.HomeAccountId.TenantId}";
+        }
     }
 }
--- a/src/client/Microsoft.Identity.Client/AccountExtensions.cs
+++ b/src/client/Microsoft.Identity.Client/AccountExtensions.cs
@@ -25,9 +25,9 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
-        /// Returns the authority from which the account is acquired.
+        /// Gets the authority associated with the home tenant for this user.
         /// </summary>
-        /// <returns>Authority for the account.</returns>
+        /// <returns>Authority.</returns>
         public static string GetAuthority(this IAccount account)
         {
             return $"https://{account.Environment}/{account.HomeAccountId.TenantId}";
--- a/src/client/Microsoft.Identity.Client/AccountExtensions.cs
+++ b/src/client/Microsoft.Identity.Client/AccountExtensions.cs
@@ -23,14 +23,5 @@ namespace Microsoft.Identity.Client
         {
             return (account as Account)?.TenantProfiles;
         }
-
-        /// <summary>
-        /// Gets the authority associated with the home tenant for this user.
-        /// </summary>
-        /// <returns>Authority.</returns>
-        public static string GetAuthority(this IAccount account)
-        {
-            return $"https://{account.Environment}/{account.HomeAccountId.TenantId}";
-        }
     }
 }
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -105,15 +105,15 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
-        /// Configures the public client application to use multi cloud support.
-        /// It enables applications to pass in a default public cloud authority to the library and can still get tokens for resources from national clouds.
+        /// Enables multi cloud support for this instance of public client application.
+        /// It enables applications to use in a global public cloud authority to the library and can still get tokens for resources from national clouds.
         /// </summary>
         /// <param name="enabled"></param>
         /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
         /// parameters, and to create a public client application instance</returns>
-        public PublicClientApplicationBuilder WithMultiCloudSupport(bool enabled)
+        public PublicClientApplicationBuilder WithMultiCloudSupport(bool enableMultiCloudSupport)
         {
-            Config.MultiCloudSupportEnabled = enabled;
+            Config.MultiCloudSupportEnabled = enableMultiCloudSupport;
             return this;
         }
 
@@ -385,6 +385,7 @@ namespace Microsoft.Identity.Client
 
             if (Config.IsBrokerEnabled && Config.MultiCloudSupportEnabled)
             {
+                // TODO: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3139
                 throw new NotSupportedException(MsalErrorMessage.MultiCloudSupportUnavailable);
             }
         }
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Identity.Client
         /// parameters, and to create a public client application instance</returns>
         public PublicClientApplicationBuilder WithMultiCloudSupport(bool enabled)
         {
-            Config.MultiCloudSupport = enabled;
+            Config.MultiCloudSupportEnabled = enabled;
             return this;
         }
 
@@ -383,7 +383,7 @@ namespace Microsoft.Identity.Client
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
 
-            if (Config.IsBrokerEnabled && Config.MultiCloudSupport)
+            if (Config.IsBrokerEnabled && Config.MultiCloudSupportEnabled)
             {
                 throw new NotSupportedException(MsalErrorMessage.MultiCloudSupportUnavailable);
             }
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -108,7 +108,7 @@ namespace Microsoft.Identity.Client
         /// Enables multi cloud support for this instance of public client application.
         /// It enables applications to use in a global public cloud authority to the library and can still get tokens for resources from national clouds.
         /// </summary>
-        /// <param name="enabled"></param>
+        /// <param name="enableMultiCloudSupport">Enable or disable multi cloud support.</param>
         /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
         /// parameters, and to create a public client application instance</returns>
         public PublicClientApplicationBuilder WithMultiCloudSupport(bool enableMultiCloudSupport)
@@ -372,6 +372,12 @@ namespace Microsoft.Identity.Client
                 throw new MsalClientException(MsalError.ClientIdMustBeAGuid, MsalErrorMessage.ClientIdMustBeAGuid);
             }
 
+            if (Config.IsBrokerEnabled && Config.MultiCloudSupportEnabled)
+            {
+                // TODO: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3139
+                throw new NotSupportedException(MsalErrorMessage.MultiCloudSupportUnavailable);
+            }
+
             if (string.IsNullOrWhiteSpace(Config.RedirectUri))
             {
                 Config.RedirectUri = PlatformProxyFactory.CreatePlatformProxy(null)
@@ -381,12 +387,6 @@ namespace Microsoft.Identity.Client
             if (!Uri.TryCreate(Config.RedirectUri, UriKind.Absolute, out Uri uriResult))
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
-            }
-
-            if (Config.IsBrokerEnabled && Config.MultiCloudSupportEnabled)
-            {
-                // TODO: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3139
-                throw new NotSupportedException(MsalErrorMessage.MultiCloudSupportUnavailable);
             }
         }
     }
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -385,7 +385,7 @@ namespace Microsoft.Identity.Client
 
             if (Config.IsBrokerEnabled && Config.MultiCloudSupport)
             {
-                throw new NotSupportedException("Multi cloud support is not supported for broker yet.");
+                throw new NotSupportedException(MsalErrorMessage.MultiCloudSupportUnavailable);
             }
         }
     }
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -108,12 +108,12 @@ namespace Microsoft.Identity.Client
         /// Configures the public client application to use multi cloud support.
         /// It enables applications to pass in a default public cloud authority to the library and can still get tokens for resources from national clouds.
         /// </summary>
+        /// <param name="enabled"></param>
         /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
         /// parameters, and to create a public client application instance</returns>
-        public PublicClientApplicationBuilder WithMultiCloudSupport()
+        public PublicClientApplicationBuilder WithMultiCloudSupport(bool enabled)
         {
-            Config.MultiCloudSupport = true;
-            Config.ExtraQueryParameters.Add("instance_aware", "true");
+            Config.MultiCloudSupport = enabled;
             return this;
         }
 
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -105,6 +105,19 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
+        /// Configures the public client application to use multi cloud support.
+        /// It enables applications to pass in a default public cloud authority to the library and can still get tokens for resources from national clouds.
+        /// </summary>
+        /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
+        /// parameters, and to create a public client application instance</returns>
+        public PublicClientApplicationBuilder WithMultiCloudSupport()
+        {
+            Config.MultiCloudSupport = true;
+            Config.ExtraQueryParameters.Add("instance_aware", "true");
+            return this;
+        }
+
+        /// <summary>
         /// You can specify a Keychain Access Group to use for persisting the token cache across multiple applications.
         /// This enables you to share the token cache between several applications having the same Keychain access group.
         /// Sharing the token cache allows single sign-on between all of the applications that use the same Keychain access Group.
--- a/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
@@ -382,6 +382,11 @@ namespace Microsoft.Identity.Client
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (Config.IsBrokerEnabled && Config.MultiCloudSupport)
+            {
+                throw new NotSupportedException("Multi cloud support is not supported for broker yet.");
+            }
         }
     }
 }
--- a/src/client/Microsoft.Identity.Client/IAccount.cs
+++ b/src/client/Microsoft.Identity.Client/IAccount.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 
+using System;
 using System.Collections.Generic;
 
 namespace Microsoft.Identity.Client
@@ -33,6 +34,10 @@ namespace Microsoft.Identity.Client
         /// </summary>
         /// <remarks>Can be null, for example if this account was migrated to MSAL.NET from ADAL.NET v3's token cache</remarks>
         AccountId HomeAccountId { get; }
-       
-   }
+
+        /// <summary>
+        /// Returns the authority from which the token is acquired.
+        /// </summary>
+        string GetAuthority();    
+    }
 }
--- a/src/client/Microsoft.Identity.Client/IAccount.cs
+++ b/src/client/Microsoft.Identity.Client/IAccount.cs
@@ -33,11 +33,6 @@ namespace Microsoft.Identity.Client
         /// AccountId of the home account for the user. This uniquely identifies the user across AAD tenants.
         /// </summary>
         /// <remarks>Can be null, for example if this account was migrated to MSAL.NET from ADAL.NET v3's token cache</remarks>
-        AccountId HomeAccountId { get; }
-
-        /// <summary>
-        /// Returns the authority from which the token is acquired.
-        /// </summary>
-        string GetAuthority();    
+        AccountId HomeAccountId { get; } 
     }
 }
--- a/src/client/Microsoft.Identity.Client/Instance/Authority.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Authority.cs
@@ -83,15 +83,13 @@ namespace Microsoft.Identity.Client.Instance
 
                 case AuthorityType.Aad:
 
-                    bool updateAuthority = requestContext.ServiceBundle.Config.MultiCloudSupportEnabled && account != null;
+                    bool updateEnvironment = requestContext.ServiceBundle.Config.MultiCloudSupportEnabled && account != null;
 
                     if (requestAuthorityInfo == null)
                     {
-                        return CreateAuthorityWithTenant(
-                            updateAuthority ? 
-                                new AuthorityInfo(AuthorityType.Aad, $"https://{account.Environment}/{account.HomeAccountId.TenantId}", configAuthorityInfo.ValidateAuthority) : 
-                                configAuthorityInfo, 
-                            account?.HomeAccountId?.TenantId);
+                        return updateEnvironment ?
+                            CreateAuthorityWithTenant(CreateAuthorityWithEnvironment(configAuthorityInfo, account.Environment).AuthorityInfo, account?.HomeAccountId?.TenantId) :
+                            CreateAuthorityWithTenant(configAuthorityInfo, account?.HomeAccountId?.TenantId);
                     }
 
                     // In case the authority is defined only at the request level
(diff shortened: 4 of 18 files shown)


neha-bhargava:
As discussed during the standup, PR is open to review. I will add integration tests meanwhile.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -104,6 +104,19 @@ public PublicClientApplicationBuilder WithDefaultRedirectUri()
...
+        /// <summary>
+        /// Configures the public client application to use multi cloud support.
+        /// It enables applications to pass in a default public cloud authority to the library and can still get tokens for resources from national clouds.
+        /// </summary>
+        /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
+        /// parameters, and to create a public client application instance</returns>
+        public PublicClientApplicationBuilder WithMultiCloudSupport()
+        {
+            Config.MultiCloudSupport = true;
+            Config.ExtraQueryParameters.Add("instance_aware", "true");
```
bgavrilMS, line 116: We should not add this kind of logic in the config layer. App developers may override this via `WithExtraQueryParameters(Dictionary)`

src/client/Microsoft.Identity.Client/IAccount.cs
```diff
@@ -33,6 +34,10 @@ public interface IAccount
         /// </summary>
         /// <remarks>Can be null, for example if this account was migrated to MSAL.NET from ADAL.NET v3's token cache</remarks>
         AccountId HomeAccountId { get; }
-       
-   }
+
+        /// <summary>
+        /// Returns the authority from which the token is acquired.
+        /// </summary>
+        string GetAuthority();
```
bgavrilMS, line 41: IAccount is a public interface, so you cannot change it, because it'll break ppl. Both Java and C# solve this problem with "default interface methods", but sadly we cannot use this feature in C# because it is only available on .NET Core. 

So there are 2 practical ways of solving this: 

1. (we won't do this, but just for reference) Create a new interface IAccount2 which inherits from IAccount and adds the new member. Make our code use IAccount2 everywhere. Customers relying on IAccount will not be broken because of polymorphism. 

2. Create an extension method `GetAuthority()`

```csharp
public static class AccountExtensions
{
    public static string GetAuthority(this IAccount account)
    {
          $"https://{account.Environment}/{account.HomeAccountId.TenantId}";
    }
}

```

bgavrilMS: Note that extension methods are a C# concept which only works for methods. This is why you can't have a property Authority ...

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -104,6 +104,19 @@ public PublicClientApplicationBuilder WithDefaultRedirectUri()
             return this;
         }
 
+        /// <summary>
+        /// Configures the public client application to use multi cloud support.
+        /// It enables applications to pass in a default public cloud authority to the library and can still get tokens for resources from national clouds.
+        /// </summary>
+        /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
+        /// parameters, and to create a public client application instance</returns>
+        public PublicClientApplicationBuilder WithMultiCloudSupport()
```
bgavrilMS, line 113: For consistency, please have this as 

`WithMultiCloudSupport(bool enabled)`

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -227,6 +239,16 @@ private async Task<AuthenticationResult> RunTestForUserAsync(LabResponse labResp
                     .WithTestLogging()
                     .Build();
             }
+            else if (multiCloudSupport)
+            {
+                pca = PublicClientApplicationBuilder
+                    .Create(labResponse.App.AppId)
+                    .WithRedirectUri(SeleniumWebUI.FindFreeLocalhostRedirectUri())
+                    .WithMultiCloudSupport()
```
bgavrilMS, line 247: You can add just the "WithMultiCloudSupport(multiCloudSupport)" to the existing builder.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
bgavrilMS: The request can do this for example.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -227,6 +239,16 @@ private async Task<AuthenticationResult> RunTestForUserAsync(LabResponse labResp
                     .WithTestLogging()
                     .Build();
             }
+            else if (multiCloudSupport)
```
bgavrilMS, line 242: Could you please refactor this into a standalone test (maybe not so many steps like this one, just an Interactive + Silent). This will serve as a the basis for documentation and demo for this feature. As it stands, the test code doesn't show the new pattern well enough.

bgavrilMS (changes requested):
main concern is around IAccount interface which can't be modified and adding the "instance_aware" flag somewhere else.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
neha-bhargava: Done

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
neha-bhargava: Done

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -212,31 +212,51 @@ public async Task ValidateCcsHeadersForInteractiveAuthCodeFlowAsync()
...
+            Assert.IsNotNull(result);
+            Assert.IsNotNull(result.Account);
+            Assert.AreEqual(labResponse.Lab.Authority + labResponse.Lab.TenantId, result.Account.GetAuthority());
+
+            Trace.WriteLine("Part 2 - Get Accounts");
+            var accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(accounts);
+            Assert.IsTrue(accounts.Count() > 0);
+            Assert.AreEqual(labResponse.User.Upn, accounts.FirstOrDefault().Username);
```
pmaytak, line 241: I think there might be a bug?
GetAccounts returns the account with the global environment. While ATS and ATI return account correctly with national environment.

pmaytak (changes requested):

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -369,6 +382,11 @@ internal override void Validate()
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (Config.IsBrokerEnabled && Config.MultiCloudSupport)
+            {
+                throw new NotSupportedException(MsalErrorMessage.MultiCloudSupportUnavailable);
```
bgavrilMS, line 388: nit: pls make sure we have a tracking item for the broker work.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -960,8 +963,13 @@ async Task<IEnumerable<IAccount>> ITokenCacheInternal.GetAccountsAsync(Authentic
...
                 requestParameters.RequestContext).ConfigureAwait(false);
 
-            rtCacheItems = rtCacheItems.Where(rt => instanceMetadata.Aliases.ContainsOrdinalIgnoreCase(rt.Environment)).ToList();
-            accountCacheItems = accountCacheItems.Where(acc => instanceMetadata.Aliases.ContainsOrdinalIgnoreCase(acc.Environment)).ToList();
+            // If the client application is instance aware then we skip the filter with environment
+            // since the authority in request is different from the authority used to get the token
+            if (!requestParameters.AppConfig.MultiCloudSupport)
+            {
+                rtCacheItems = rtCacheItems.Where(rt => instanceMetadata.Aliases.ContainsOrdinalIgnoreCase(rt.Environment)).ToList();
+                accountCacheItems = accountCacheItems.Where(acc => instanceMetadata.Aliases.ContainsOrdinalIgnoreCase(acc.Environment)).ToList();
```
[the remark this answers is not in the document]
bgavrilMS: Should we call `ToList()` to make the code more consistent?

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -310,6 +331,42 @@ private async Task<AuthenticationResult> RunTestForUserAsync(LabResponse labResp
             return result;
         }
 
+        private Tuple<IPublicClientApplication, HttpSnifferClientFactory> getPCA(LabResponse labResponse, bool directToAdfs, bool multiCloudSupport)
```
bgavrilMS, line 334: In test projects you can use the fancier tuples from C# (in MSAL source code you cannot because of .net 45)

```csharp
 private Tuple<IPublicClientApplication pca, HttpSnifferClientFactory httpFactory> getPCA() { ... } 

// later in the code instead of var tuple = getPCA(labResponse, false, true);
(IPublicClientApplication Pca, HttpSnifferClientFactory HttpFactory) tuple = getPCA(labResponse, false, true);
tuple.Pca.AcquireToken(...)

```

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -212,31 +212,51 @@ public async Task ValidateCcsHeadersForInteractiveAuthCodeFlowAsync()
 
             Assert.IsNotNull(authResult);
             Assert.IsNotNull(authResult.AccessToken);
+        }
+
+        [TestMethod]
+        public async Task Interactive_Arlington_MultiCloudSupport_AADAsync()
```
bgavrilMS, line 218: Integration tests are expensive, especially these ones with UI. Best to execute them only for .NET Classic or for .NET Core (by default they execute on both). See comment on line 61. I'd add this test in the `#if` block.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -212,31 +212,51 @@ public async Task ValidateCcsHeadersForInteractiveAuthCodeFlowAsync()
...
+        {
+            // Arrange
+            LabResponse labResponse = await LabUserHelper.GetArlingtonUserAsync().ConfigureAwait(false);
```
bgavrilMS, line 221: Can you please create the PCA object right here in the test code, even if it leads to a small amount of duplicate code? It will make this test case more clear.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -212,31 +212,51 @@ public async Task ValidateCcsHeadersForInteractiveAuthCodeFlowAsync()
...
+
+            Assert.IsNotNull(result);
+            Assert.IsNotNull(result.Account);
+            Assert.AreEqual(labResponse.Lab.Authority + labResponse.Lab.TenantId, result.Account.GetAuthority());
+
+            Trace.WriteLine("Part 2 - Get Accounts");
+            var accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(accounts);
+            Assert.IsTrue(accounts.Count() > 0);
```
bgavrilMS, line 240: Should be exactly 1. 

nit: you can use `.Single()` which will fail if there isn't exactly 1

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
bgavrilMS: @neha-bhargava  - please add some more assertions around the account objects. IN particular, I think the accountId and ENV should be equal between the accoutn from ATI and GetAccounts()

@pmaytak - `GetAccounts` returns all the accounts, irrespective of cloud, when the "MultiCloudSupport" flag is enabled. So I think in this case `GetAccounts` returns the national cloud account. Can you give more details about where you see the bug?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -960,8 +963,13 @@ async Task<IEnumerable<IAccount>> ITokenCacheInternal.GetAccountsAsync(Authentic
                 allEnvironmentsInCache,
                 requestParameters.RequestContext).ConfigureAwait(false);
 
-            rtCacheItems = rtCacheItems.Where(rt => instanceMetadata.Aliases.ContainsOrdinalIgnoreCase(rt.Environment)).ToList();
-            accountCacheItems = accountCacheItems.Where(acc => instanceMetadata.Aliases.ContainsOrdinalIgnoreCase(acc.Environment)).ToList();
+            // If the client application is instance aware then we skip the filter with environment
+            // since the authority in request is different from the authority used to get the token
+            if (!requestParameters.AppConfig.MultiCloudSupport)
```
bgavrilMS, line 968: These changes need more unit tests

bgavrilMS (approved):
LGTM, please resolve outstanding comments and add a unit test around GetAccounts

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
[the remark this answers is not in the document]
bgavrilMS: This is a good point / discussion we can have as a team. Integration tests are expensive (time wise), particularly the UI ones. So there's a trade off between: 

- adding coverage for different clouds ... this is more of AADs domain, from MSAL's perspective if one cloud works, all should work. That said, we have tests for AcquireTokenInteractive for ALL clouds.
- slower build

src/client/Microsoft.Identity.Client/AccountExtensions.cs
```diff
@@ -23,5 +23,14 @@ public static IEnumerable<TenantProfile> GetTenantProfiles(this IAccount account
         {
             return (account as Account)?.TenantProfiles;
         }
+
+        /// <summary>
+        /// Returns the authority from which the account is acquired.
+        /// </summary>
+        /// <returns>Authority for the account.</returns>
+        public static string GetAuthority(this IAccount account)
```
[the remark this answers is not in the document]
neha-bhargava: No, this is an extension method which will provide the authority which was used to acquire the token. We particularly need it to make the silent call in this case.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
neha-bhargava: I did create an issue in github to track this. Will add it as a comment.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
pmaytak: Repro (can use NetCoreWinFormsWMA, just uncheck `Use WAM`):
- Create PCA, add `WithMultiCloudSupport`, use global authority.
- Call AcquireTokenInteractive, login with US account.
   - AuthenticationResult.Account.Environment is national.
 - Call AcquireTokenSilent, Add `WithAuthority(result.acc.GetAuthority())`.
   - AuthResult.Account.Environment is still national.
 - Call GetAccounts
   - Returns one account, but Account.Environment is global.
 - Call GetAccount with result.Account.HomeAccountId.Identifier.
   - Also returns account with global environment.

All tokens and accounts are saved with national environment in the cache.
I think the issue might be here:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/a65f7afa4eabbfc04d6696a1f6c1ce430c3da8a9/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L989

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
pmaytak: Maybe we can ToList in the foreach loops below this code, instead of here? And the Verbose logging on 974 can be moved after foreach also, since it calls Count.

Although I'm not sure if calls to .Count and .ToList reevaluate the expression tree each time (probably since expression can have reference to some external delegate code that can change result each time).

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: Are these test cases tested/ covered:
- If AT is expired, ATS refreshes using RT.
- ATS works if proactive refresh is triggered and works.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
neha-bhargava: Ok got it. Thanks for the explanation @pmaytak.

src/client/Microsoft.Identity.Client/AccountExtensions.cs
```diff
@@ -23,5 +23,14 @@ public static IEnumerable<TenantProfile> GetTenantProfiles(this IAccount account
...
+        {
+            return $"https://{account.Environment}/{account.HomeAccountId.TenantId}";
```
jmprieur, line 33: Does it work in the case of B2C ? (I would not think so)

jmprieur (commented):
I left a comment

src/client/Microsoft.Identity.Client/AccountExtensions.cs
bgavrilMS: Yes, I had the same comment. We need to have unit tests for this for ADFS and B2C

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -369,6 +382,11 @@ internal override void Validate()
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (Config.IsBrokerEnabled && Config.MultiCloudSupport)
+            {
```
SameerK-MSFT, line 387: You may want to do this check prior to assigning COnfig.RedirectUri

neha-bhargava: I have added a check to throw NotSupportedException if MultiCloudSupport is opted with Broker. I can remove the broker check from here. Is this what you are asking for @SameerK-MSFT ?

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
trwalke: I thought the mobile platforms already had support for this? Or is there something preventing us from enabling this on the broker from our side?

src/client/Microsoft.Identity.Client/IAccount.cs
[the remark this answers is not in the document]
neha-bhargava: Removed GetAuthority

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
neha-bhargava: We split this work to first implement it for web and then work for broker to make it simpler. And there is a team waiting for this work on the web.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
neha-bhargava: Marking this as resolved since we are not using GetAuthority any more.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -56,10 +56,22 @@ public async Task Interactive_AADAsync()
         {
             // Arrange
             LabResponse labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
-            await RunTestForUserAsync(labResponse).ConfigureAwait(false);
+            var result = await RunTestForUserAsync(labResponse).ConfigureAwait(false);
         }
-#if DESKTOP // no point in running these tests on NetCore - the code path is similar
 
+        [TestMethod]
+        public async Task Interactive_Arlington_MultiCloudSupport_AADAsync()
```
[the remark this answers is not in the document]
neha-bhargava: Added

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
neha-bhargava: @pmaytak I tried to remove ToList() from here and it will give a compilation error `rtCacheItems` is an object of type `IReadOnlyList` and `rtCacheItems.Where` returns an object of type `IEnumerable` so the `ToList` is used to convert it to `List` object. It will not impact any of the flow below.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -82,9 +83,15 @@ protected Authority(AuthorityInfo authorityInfo)
...
                 case AuthorityType.Aad:
 
+                    bool updateAuthority = requestContext.ServiceBundle.Config.MultiCloudSupportEnabled && account != null;
+
                     if (requestAuthorityInfo == null)
                     {
-                        return CreateAuthorityWithTenant(configAuthorityInfo, requestHomeAccountTenantId);
+                        return CreateAuthorityWithTenant(
+                            updateAuthority ? 
+                                new AuthorityInfo(AuthorityType.Aad, $"https://{account.Environment}/{account.HomeAccountId.TenantId}", configAuthorityInfo.ValidateAuthority) :
```
bgavrilMS, line 92: Can you move this logic in Authority or AuthorityInfo? We should not do any sort of string manipulation in here. There is already something very similar (and more robust) - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/Instance/Authority.cs#L165

src/client/Microsoft.Identity.Client/Instance/Authority.cs
```diff
@@ -82,9 +83,15 @@ protected Authority(AuthorityInfo authorityInfo)
 
                 case AuthorityType.Aad:
 
+                    bool updateAuthority = requestContext.ServiceBundle.Config.MultiCloudSupportEnabled && account != null;
```
bgavrilMS, line 86: nit: updateEnvironment

bgavrilMS (approved):
LGTM, some small comments remaining

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
neha-bhargava: Added unit tests for `GetAccountsAsync` and `FindRefreshTokenAsync` when multi cloud support is enabled and when it is not enabled.

src/client/Microsoft.Identity.Client/Instance/Authority.cs
neha-bhargava: Updated

pmaytak (changes requested):
Get Account still returns account with global environment, not national.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
pmaytak: GetAccount/GetAccounts still returns account with global environment, not national.
Repro for example. Passing `interactiveResult.Account` into ATS will work, passing `account`, does not work.
```
            var interactiveResult = await pca.AcquireTokenInteractive(new[] { scopeGraph })
                                .WithUseEmbeddedWebView(true)
                                .ExecuteAsync();

            var account = await pca.GetAccountAsync(result.Account.HomeAccountId.Identifier);

            var result1 = await pca.AcquireTokenSilent(new[] { scopeGraph }, account)
                .ExecuteAsync();
```

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -790,8 +790,11 @@ internal async Task ExpireAllAccessTokensForTestAsync()
                 var aliases = metadata.Aliases;
 
                 allRts = FilterRtsByHomeAccountIdOrAssertion(requestParams, allRts, familyId);
-                allRts = allRts.Where(
-                    item => aliases.ContainsOrdinalIgnoreCase(item.Environment)).ToList();
+                if (!requestParams.AppConfig.MultiCloudSupportEnabled)
```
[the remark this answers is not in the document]
bgavrilMS: That makes sense.

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
```diff
@@ -124,8 +124,51 @@ public async Task Arlington_Interactive_AdfsV2019_FederatedAsync()
...
+            Assert.IsNotNull(result.Account);
+            Assert.IsTrue(labResponse.Lab.Authority.Contains(result.Account.Environment));
+
+            Trace.WriteLine("Part 2 - Get Accounts");
+            var accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(accounts);
+            Assert.IsNotNull(accounts.Single());
+            Assert.AreEqual(labResponse.User.Upn, accounts.FirstOrDefault().Username);
+            Assert.AreEqual("login.microsoftonline.com", accounts.Single().Environment);
```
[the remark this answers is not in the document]
bgavrilMS: Yes, the account.Environment should be Arlingthon here. 
The accont from line 148 should deep equal (i.e. same account id, same env, same upn etc.) the account at line 156

bgavrilMS (changes requested):
The account from GetAccounts should have the same env as the account from result.Account, i.e. sovereign env.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
neha-bhargava: Done

tests/Microsoft.Identity.Test.Unit/CacheTests/TokenCacheTests.cs
```diff
@@ -92,6 +93,43 @@ public override void TestInitialize()
...
+
+            var requestParams = TestCommon.CreateAuthenticationRequestParameters(serviceBundle);
+            requestParams.AuthorityManager = new AuthorityManager(
+                requestContext,
+                Authority.CreateAuthorityWithTenant(
+                    requestParams.AuthorityInfo,
+                    TestConstants.Utid));
+            requestParams.Account = new Account(TestConstants.s_userIdentifier, $"1{TestConstants.DisplayableId}", TestConstants.ProductionPrefNetworkEnvironment);
+
+            await cache.FindRefreshTokenAsync(requestParams).ConfigureAwait(true);
```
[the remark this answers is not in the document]
neha-bhargava: removed

tests/Microsoft.Identity.Test.Integration.netfx/SeleniumTests/InteractiveFlowTests.cs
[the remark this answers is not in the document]
neha-bhargava: Updated. I was under the impression that this should remain same as input env for consistency. But updated this only when MultiCloudSupport is enabled.

bgavrilMS (approved):

pmaytak (changes requested):
There's an issue with tenant profiles.
When WithMultiCloudSupport is enabled (with global org authority) and national account is used - GetAccount/GetAccounts returns an account with missing tenant profiles. 
However, tenant profiles exist in these cases: the AuthResult account from ATS and ATI has tenant profiles; global accounts with MultiCloudSupport have tenant profiles; national account with national authority.

GetAccount call
image
ATS call
image

pmaytak (approved):

PR #3149, opened 2022-02-09
Fix for #3130 - if no tokens are in the cache, suggested expiry = null

### LINKED ISSUE #3130: Sequence Contains No Elements Exception from ITokenCacheInternal.SaveTokenResponseAsync
Using MSAL 4.39.0, my service logged the following exception:
image

Quickly digging into the code, the exception looks like it's coming from here:
image

Suggest the line is updated to `DateTimeOffset cacheExpiry = tokenCacheItems.DefaultIfEmpty().Max(item => item.ExpiresOn);` or something along those lines (depending on the behavior you want for this case).

bgavrilMS:
Fixes #3130 

 if no tokens are in the cache, suggested expiry = NOW

**Testing**
unit

**Performance impact**
1 extra int comparison will not degrade perf

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -12,6 +12,7 @@ using Microsoft.Identity.Client.AuthScheme.Bearer;
 using Microsoft.Identity.Client.Cache;
 using Microsoft.Identity.Client.Cache.Items;
 using Microsoft.Identity.Client.Cache.Keys;
+using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Instance;
 using Microsoft.Identity.Client.Instance.Discovery;
 using Microsoft.Identity.Client.Internal;
@@ -34,7 +35,8 @@ namespace Microsoft.Identity.Client
             AuthenticationRequestParameters requestParams,
             MsalTokenResponse response)
         {
-            response.Log(requestParams.RequestContext.Logger, LogLevel.Verbose);
+            var logger = requestParams.RequestContext.Logger;
+            response.Log(logger, LogLevel.Verbose);
 
             MsalAccessTokenCacheItem msalAccessTokenCacheItem = null;
             MsalRefreshTokenCacheItem msalRefreshTokenCacheItem = null;
@@ -44,7 +46,7 @@ namespace Microsoft.Identity.Client
             IdToken idToken = IdToken.Parse(response.IdToken);
             if (idToken == null)
             {
-                requestParams.RequestContext.Logger.Info("ID Token not present in response. ");
+                logger.Info("ID Token not present in response. ");
             }
 
             var tenantId = GetTenantId(idToken, requestParams);
@@ -138,9 +140,9 @@ namespace Microsoft.Identity.Client
 
             #endregion
 
-            requestParams.RequestContext.Logger.Verbose($"[SaveTokenResponseAsync] Entering token cache semaphore. Count {_semaphoreSlim.GetCurrentCountLogMessage()}.");
+            logger.Verbose($"[SaveTokenResponseAsync] Entering token cache semaphore. Count {_semaphoreSlim.GetCurrentCountLogMessage()}.");
             await _semaphoreSlim.WaitAsync(requestParams.RequestContext.UserCancellationToken).ConfigureAwait(false);
-            requestParams.RequestContext.Logger.Verbose("[SaveTokenResponseAsync] Entered token cache semaphore. ");
+            logger.Verbose("[SaveTokenResponseAsync] Entered token cache semaphore. ");
             ITokenCacheInternal tokenCacheInternal = this;
 
             try
@@ -174,7 +176,7 @@ namespace Microsoft.Identity.Client
 
                     if (msalAccessTokenCacheItem != null)
                     {
-                        requestParams.RequestContext.Logger.Info("Saving AT in cache and removing overlapping ATs...");
+                        logger.Info("Saving AT in cache and removing overlapping ATs...");
 
                         DeleteAccessTokensWithIntersectingScopes(
                             requestParams,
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -346,11 +343,26 @@ private bool IsLegacyAdalCacheEnabled(AuthenticationRequestParameters requestPar
...
-            DateTimeOffset cacheExpiry = tokenCacheItems.Max(item => item.ExpiresOn);
-            return cacheExpiry;
+            // If we have refresh tokens in the cache, we cannot suggest expiration
+            // because refresh token expiration is not disclosed to SDKs and RTs are long lived anyway (3 months by default)
+            if (accessor.GetAllRefreshTokens().Count == 0)
+            {
+                IReadOnlyList<MsalAccessTokenCacheItem> tokenCacheItems = accessor.GetAllAccessTokens(optionalPartitionKey: null);
+                if (tokenCacheItems.Count == 0)
+                {
+                    logger.Warning("[CalculateSuggestedCacheExpiry] No access tokens or refresh tokens found in the accessor. Suggesting immediate expiration.");
```
bgavrilMS, line 357: @pmaytak - I don't really understand how we can get in this sitatuation, unless there is a race condition in the accessor. The customers complaining are using the AppTokenCache. 

This method is only called when we save a valid MSAL token response, which has a new access token.

bgavrilMS:
@pmaytak @trwalke - please review ASAP as this should go out as a hotfix.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -346,11 +343,35 @@ private bool IsLegacyAdalCacheEnabled(AuthenticationRequestParameters requestPar
...
         {
-            IEnumerable<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true);
-            DateTimeOffset cacheExpiry = tokenCacheItems.Max(item => item.ExpiresOn);
-            return cacheExpiry;
+            // If we have refresh tokens in the cache, we cannot suggest expiration
+            // because refresh token expiration is not disclosed to SDKs and RTs are long lived anyway (3 months by default)
+            if (accessor.GetAllRefreshTokens().Count == 0)
+            {
+                IReadOnlyList<MsalAccessTokenCacheItem> tokenCacheItems = accessor.GetAllAccessTokens(optionalPartitionKey: null);
+                if (tokenCacheItems.Count == 0)
```
pmaytak, line 358: I guess there might be multiple thread access issues here too (although probably unlikely)? Thread 1 returns 0 RTs, Thread 2 clears the cache and inserts RTs and ATs, Thread 1 operates on a different set of ATs than before step 1?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -346,11 +343,35 @@ private bool IsLegacyAdalCacheEnabled(AuthenticationRequestParameters requestPar
...
+                {
+                    logger.Warning("[CalculateSuggestedCacheExpiry] No access tokens or refresh tokens found in the accessor. Not returning any expiration.");
+                    return null;
+                }
+
+                DateTimeOffset cacheExpiry = tokenCacheItems.Max(item => item.ExpiresOn);
+                
+                // do not suggest an expiration date from the past or within 5 min, as tokens will not be usable anyway
+                // and HasTokens will be set to false, letting implementers know to delete the cache node
+                if (cacheExpiry < DateTimeOffset.UtcNow + Constants.AccessTokenExpirationBuffer)
```
jennyf19, line 368: This fix looks good for us. I know one 1P who hit this issue, and they could test the ID web version for us, if you want. Also, I think you are spending too much time in SAL, as I see you remove the `{ }` on the one liners ;)

pmaytak (approved):

jennyf19 (approved):
The fix for past expiration looks good.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: This only happens when cache serialization is enabled. What about this?
- Thread 1 saves token in the cache.
- Thread 2 clears cache.
- Thread 1 gets to calculating the expiry but the cache is empty.

I updated the NetCore dev app with a multi thread test case [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ed901db2a22bd860db1050a5bfbda667468247d4/tests/devapps/NetCoreTestApp/Program.cs#L251). So maybe that can be used to repro this, just add in-memory cache serialization.

pmaytak: Also our cache is based on eventual consistency, since we don't lock, but before deserialization we clear the cache. So in a multi threaded env, we can get into the above scenario?

neha-bhargava (approved):

trwalke (approved):
Looks clean. Should definitely to more thorough checking for null/empty collections even when there is strong evidence that it will not be the case. Strange scenarios like this can slip through the cracks.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
bgavrilMS: Yes, this could explain why this issue occurs so rarely.

tests/Microsoft.Identity.Test.Unit/CacheTests/TokenCacheTests.cs
```diff
@@ -439,6 +440,77 @@ public void GetAccessTokenExpiryInRangeTest()
...
+                userTokenCache.Accessor.SaveAccessToken(t1);
+                userTokenCache.Accessor.SaveAccessToken(t2);
+                userTokenCache.Accessor.SaveAccessToken(t3);
+
+                // Act
+                var appAccessorExpiration = TokenCache.CalculateSuggestedCacheExpiry(appTokenCache.Accessor, logger);
+                var userAccessorExpiration = TokenCache.CalculateSuggestedCacheExpiry(userTokenCache.Accessor, logger);
+
+                // Assert
+                Assert.IsNull(appAccessorExpiration);
```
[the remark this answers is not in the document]
bgavrilMS: Updated to use some constants thjat will make it clear.

PR #3160, opened 2022-02-15
WwwAuthenticateParameters should not expose Resource/Scopes

### LINKED ISSUE #3144: [Bug] WwwAuthenticateParameters should not expose Resource
4.41

**Platform**
All

Other?
WWW-Authenticate parameters

**Actual behavior**
WwwAuthenticateParameters exposes the `Resource `property. It should not as the client should know which ressource it called
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/efddbb7c6e15f270c36fe8c276c3ba86997ee500/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs#L28

**Expected behavior**
Don't expose the  `Resource `property any longer.

**Possible solution**

```CSharp
  /// <summary>
  /// Resource for which to request scopes.
  /// This is the App ID URI of the API that returned the WWW-Authenticate header.
  /// </summary>
  public string Resource 
  {
    [Obsolete("The client apps should know which App ID URI it requests scopes for") 
    get; 
    set; 
   }
```

- [ ] update documentation [docs link](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/WWW-Authenticate-parameters#calling-a-protected-api-unauthenticated)

gladjohn:
Fixes #3144 

- WwwAuthenticateParameters should not expose Resource/Scopes
- Marked both resource and scopes property in WwwAuthenticateParameters as Obsolete
- Added notes to bug [3144](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3144) to update docs 

**Testing**
- unit, integration
- created a test app based on suggested usage, [How to - use the WwwAuthenticateParameters class](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/WWW-Authenticate-parameters#how-to---use-the-wwwauthenticateparameters-class), found scopes will be null if we marked Resource as obsolete. Per discussion marking scope as obsolete. 

**Performance impact**
none. we no longer calculate scope based on the resource that is returned maybe a minor improvement.

### DIFF
--- a/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
+++ b/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
@@ -33,24 +33,7 @@ namespace Microsoft.Identity.Client
         /// If it's not provided by the web API, it's computed from the Resource.
         /// </summary>
         [Obsolete("The client apps should know which scopes to request for.", true)]
-        public IEnumerable<string> Scopes
-        {
-            get
-            {
-                if (_scopes != null)
-                {
-                    return _scopes;
-                }
-
-                return null;
-            }
-            set
-            {
-                _scopes = value;
-            }
-        }
-
-        private IEnumerable<string> _scopes;
+        public IEnumerable<string> Scopes { get; set; }
 
         /// <summary>
         /// Authority from which to request an access token.
--- a/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
+++ b/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
@@ -25,12 +25,14 @@ namespace Microsoft.Identity.Client
         /// Resource for which to request scopes.
         /// This is the App ID URI of the API that returned the WWW-Authenticate header.
         /// </summary>
+        [Obsolete("The client apps should know which App ID URI it requests scopes for.", true)]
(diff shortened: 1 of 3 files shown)


jmprieur (commented):
LGTM
Thanks @GladwinJohnson 

you might want to remove line 53 and 49?

gladjohn:
> you might want to remove line 53 and 49?
thanks @jmprieur removed

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -25,35 +25,15 @@ public class WwwAuthenticateParameters
...
-                return null;
-            }
-            set
-            {
-                _scopes = value;
-            }
-        }
-
-        private IEnumerable<string> _scopes;
+        [Obsolete("The client apps should know which scopes to request for.", true)]
```
pmaytak, line 35: Obsoleting with error - is this okay, in this case?

bgavrilMS: Yes, I think so.

bgavrilMS (approved):

SameerK-MSFT (approved):

PR #3166, opened 2022-02-16
Add the transfer token flow for MSA Passthrough + MSA silent scenarios

### LINKED ISSUE #3157: Getting OS or first account when signed in with an MSA fails
When the user is signed into windows with an MSA account, and we try and get a token for the signed in user we get two different kinds of error messages but no token is retrieved. 

1) Ask for the OS account
   var appBuilder = PublicClientApplicationBuilder.Create(NewClientId)
                .WithAuthority(AadAuthorityAudience.AzureAdMultipleOrgs)
                .WithExperimentalFeatures()
                .WithWindowsBrokerOptions(new WindowsBrokerOptions
                 {
                     ListWindowsWorkAndSchoolAccounts = true,
                     MsaPassthrough = true,
                 })
                .WithWindowsBroker();

            var pca = appBuilder.Build();

            try
            {
                var resultSignedIn = await pca.AcquireTokenSilent(new string[] { "https://graph.microsoft.com/User.Read" }, PublicClientApplication.OperatingSystemAccount).ExecuteAsync();
            }
            catch (Exception ex)
            {
                MessageBox.Show(ex.Message);
            }
   
image

2)  Instead of using the operating system account we just use the first account (which is the MSA)
       var accounts = await pca.GetAccountsAsync();
       var resultFirst = await pca.AcquireTokenSilent(new string[] { "https://graph.microsoft.com/User.Read" }, accounts.First()).ExecuteAsync();
image

bgavrilMS:
Fixes #3157

- add the transfer token flow for MSA-PT + MSA silent scenarios

**Testing**

manual
some unit testing

**Performance impact**
initial request will be much slower, because it now involves to 2 AAD calls
access tokens are cached so subsequent requests should not be as slow

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/Interfaces/IMsaPassthroughHandler.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/Interfaces/IMsaPassthroughHandler.cs
@@ -11,6 +11,9 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
     internal interface IMsaPassthroughHandler
     {
         void AddTransferTokenToRequest(WebTokenRequest webTokenRequest, string transferToken);
-        Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
+        Task<string> TryFetchTransferSilentDefaultAccountAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
+        Task<string> TryFetchTransferTokenInteractiveAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
+        Task<string> TryFetchTransferTokenSilentlyAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccount account);
     }
 }
+
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/Interfaces/IMsaPassthroughHandler.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/Interfaces/IMsaPassthroughHandler.cs
@@ -11,9 +11,9 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
     internal interface IMsaPassthroughHandler
     {
         void AddTransferTokenToRequest(WebTokenRequest webTokenRequest, string transferToken);
-        Task<string> TryFetchTransferSilentDefaultAccountAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
+        Task<string> TryFetchTransferTokenSilentDefaultAccountAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
         Task<string> TryFetchTransferTokenInteractiveAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
-        Task<string> TryFetchTransferTokenSilentlyAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccount account);
+        Task<string> TryFetchTransferTokenSilentAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccount account);
     }
 }
 
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/MsaPassthroughHandler.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/MsaPassthroughHandler.cs
@@ -36,12 +36,12 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
             _parentHandle = parentHandle;
         }
 
-        public async Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider)
+        public async Task<string> TryFetchTransferTokenInteractiveAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider)
         {
             // Apps can have MSA-PT enabled and can configured to allow MSA users
             // However, some older apps have 2 incarnations, one in AAD tenant and one in MSA tenant
             // For this second case, we can't fetch the transfer token from the client_ID in AAD and this will fail
-            _logger.Verbose("WAM MSA-PT - fetching transfer token");
+            _logger.Verbose("WAM MSA-PT - fetching transfer token for interactive flow");
 
             var webTokenRequestMsa = await _msaPlugin.CreateWebTokenRequestAsync(
                 accountProvider,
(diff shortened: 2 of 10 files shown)


crmann1 (approved):
Makes sense since this is passthrough only so the infrastructure tenant is the correct choice.  However, still should follow up with AAD to confirm this is the right approach in general.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
```diff
@@ -408,12 +408,14 @@ public async Task ATS_NoAccountMatching_ThrowsUiRequiredException_Async()
         [TestMethod]
         public async Task ATS_DefaultAccount_Async()
         {
-            string homeAccId = $"{TestConstants.Uid}.{TestConstants.Utid}";
-
             // Arrange
             using (var harness = CreateTestHarness())
             {
+
```
SameerK-MSFT, line 414: nit: extra line here + one more place below

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/Interfaces/IMsaPassthroughHandler.cs
```diff
@@ -11,6 +11,9 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
     internal interface IMsaPassthroughHandler
     {
         void AddTransferTokenToRequest(WebTokenRequest webTokenRequest, string transferToken);
-        Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
+        Task<string> TryFetchTransferSilentDefaultAccountAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
+        Task<string> TryFetchTransferTokenInteractiveAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider);
+        Task<string> TryFetchTransferTokenSilentlyAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccount account);
```
neha-bhargava, line 16: Nit: TryFetchTransferTokenSilentAsync instead?

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/MsaPassthroughHandler.cs
```diff
@@ -80,16 +80,54 @@ public async Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParame
...
+            return ExtractTransferTokenAsync(authenticationRequestParameters.AppConfig.ClientId, transferResponse);
+        }
 
-            properties.TryGetValue("code", out string code);
+        public async Task<string> TryFetchTransferSilentDefaultAccountAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider)
+        {
+            // Apps can have MSA-PT enabled and can configured to allow MSA users
+            // However, some older apps have 2 incarnations, one in AAD tenant and one in MSA tenant
+            // For this second case, we can't fetch the transfer token from the client_ID in AAD and this will fail
+            _logger.Verbose("WAM MSA-PT - fetching transfer token for interactive flow");
```
neha-bhargava, line 91:
```suggestion
            _logger.Verbose("WAM MSA-PT - fetching transfer token for silent flow");
```

neha-bhargava (approved):

gladjohn (approved):
looks good

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
```diff
@@ -556,14 +629,27 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
...
             {
+                var defaultAccountProvider = await _webAccountProviderFactory.GetDefaultProviderAsync().ConfigureAwait(false);
+                if (defaultAccountProvider == null)
+                {
+                    throw new MsalUiRequiredException(
+                        MsalError.InteractionRequired,
+                        "A default account was not found");
+
+                }
+                // special case: passthrough + default MSA account. Need to use the transfer token protocol.
```
pmaytak, line 646: Wonder if it's worth it to document all these cases, maybe in OneNote, as a matrix of which accounts (AAD/MSA, MSA-PT/ default MSA/AAD) can use which authorities/tenants via WAM/browser. Seems like quiet a few scenario variations.

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/MsaPassthroughHandler.cs
```diff
@@ -80,16 +80,54 @@ public async Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParame
...
+                catch (MsalServiceException exception)
+                {
+                    _logger.Warning(
+                        "WAM MSA-PT: could not get a transfer token, ussually this is because the " +
+                        "1st party app is configured for MSA-PT but not configured to login MSA users (signinaudience =2). " +
+                        "Error was: " + exception.ErrorCode + " " + exception.Message);
+
+                }
+
+                return null;
```
pmaytak, line 127: Is this duplicate because `ExtractTransferTokenAsync` already has this code?

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/MsaPassthroughHandler.cs
```diff
@@ -80,16 +80,54 @@ public async Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParame
...
+            if (!transferResponse.ResponseStatus.IsSuccessStatus())
+            {
+                try
+                {
+                    _ = WamAdapters.CreateMsalResponseFromWamResponse(
+                        transferResponse,
+                        _msaPlugin,
+                        authenticationRequestParameters.AppConfig.ClientId,
+                        _logger,
+                        isInteractive: true);
```
pmaytak, line 116: Does the value of this parameter matter? This `Extract` method is called from ATS methods.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamMsaPassthroughTests.cs
```diff
@@ -89,7 +89,7 @@ public async Task FetchTransferTokenAsync()
                    });
 
                 // Act
-                var transferToken = await _msaPassthroughHandler.TryFetchTransferTokenAsync(requestParams, msaProvider)
+                var transferToken = await _msaPassthroughHandler.TryFetchTransferTokenInteractiveAsync(requestParams, msaProvider)
```
pmaytak, line 92: Should there also be tests for `TryFetchTransferSilentDefaultAccountAsync` and `TryFetchTransferTokenSilentlyAsync`?

pmaytak (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/MsaPassthroughHandler.cs
```diff
@@ -80,16 +80,54 @@ public async Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParame
...
             }
 
-            _ = _msaPlugin.ParseSuccessfullWamResponse(transferResponse.ResponseData[0], out var properties);
+            return ExtractTransferTokenAsync(authenticationRequestParameters.AppConfig.ClientId, transferResponse);
+        }
 
-            properties.TryGetValue("code", out string code);
+        public async Task<string> TryFetchTransferSilentDefaultAccountAsync(AuthenticationRequestParameters authenticationRequestParameters, WebAccountProvider accountProvider)
+        {
+            // Apps can have MSA-PT enabled and can configured to allow MSA users
```
trwalke, line 88:
```suggestion
            // First Party apps can have MSA-PT enabled and can be configured to allow MSA users
```

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/MsaPassthroughHandler.cs
```diff
@@ -80,16 +80,54 @@ public async Task<string> TryFetchTransferTokenAsync(AuthenticationRequestParame
...
-            WebAccount msaPtWebAccount = transferResponse.ResponseData[0].WebAccount;
-            _logger.InfoPii($"Obtained a transfer token for {msaPtWebAccount.UserName} ?  {code != null}", $"Obtained a transfer token? {code != null}");
+            WamAdapters.AddMsalParamsToRequest(authenticationRequestParameters, webTokenRequestMsa, _logger);
 
-            return code;
+            var transferResponse = await _wamProxy.GetTokenSilentlyForDefaultAccountAsync(webTokenRequestMsa)
+                .ConfigureAwait(true);
+
+            if (!transferResponse.ResponseStatus.IsSuccessStatus())
+            {
```
trwalke, line 108: Can this block of coder be replaced with a slightly modified version of ExtractTransferTokenAsync() below?

trwalke (approved):
Minor comments about refactoring but otherwise LGTM.

Issue #2698, opened 2021-06-14
[Feature Request] Add a public method to get auto-detected region


pmaytak:
**Describe the solution you'd like**
- Create a public method that returns the cached region that was auto-detected by MSAL or does a first time detection.
- Rename auto-discovery to auto-detection for consistency (variable names, comments, etc.).

```csharp
IConfidentialClientApplication.TryDetectAzureRegion() : string
```

bgavrilMS:
#3173 covers this.

PR #3173, opened 2022-02-18
Adding additional Regional Details to AuthenticationResultMetaData

### LINKED ISSUE #2975: [Feature Request] Expose the region or error used by MSAL
Part of AuthenticationResultMetadta, expose a `string RegionOrDiscoveryError` which is populated only if `WithAzureRegion` is used.

If region auto-detection failed, this metric should describe the reason (over and over again!)

{
 AuthenticationMetadata.RegionDiscoveryOutcome -> [RegionOutcome, Exception, RegionUsed]
}

gladjohn:
Fixes #2975 

[Feature Request] Expose the region or error used by MSAL

Part of AuthenticationResultMetadta, exposing a string RegionDiscoveryOutcome which is populated only if WithAzureRegion is used. 

**Testing**
unit and integration

**Performance impact**
none


src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -82,7 +86,9 @@ public async Task<string> GetAzureRegionAsync(RequestContext requestContext)
                 }
                 else
                 {
-                    logger.Warning($"[Region discovery] Auto-discovery failed.");
+                    s_regionDiscoveryFailureReason = $"Region Auto-discovery failed.";
```
bgavrilMS, line 89: Why do you override the reason here?

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -214,15 +223,17 @@ private async Task<RegionInfo> DiscoverAsync(ICoreLogger logger, CancellationTok
...
                 if (e is MsalServiceException msalEx && MsalError.RequestTimeout.Equals(msalEx?.ErrorCode))
                 {
-                    logger.Warning($"[Region discovery] Call to local IMDS timed out after {_imdsCallTimeoutMs}.");
+                    s_regionDiscoveryFailureReason = $"Call to local IMDS timed out after {_imdsCallTimeoutMs}.";
+                    logger.Warning($"[Region discovery] {s_regionDiscoveryFailureReason}.");
                 }
                 else
                 {
-                    logger.Warning($"[Region discovery] IMDS call failed with exception {e}");
+                    s_regionDiscoveryFailureReason = $"IMDS call failed with exception {e}.";
```
bgavrilMS, line 231: Please include a timestamp in this message.

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
```diff
@@ -0,0 +1,48 @@
...
+            Exception = exception;
+        }
+
+        /// <summary>
+        /// Region Outcome based on MSAL region detection
+        /// </summary>
+        internal RegionOutcome RegionOutcome { get; }
+
+        /// <summary>
+        /// Region returned to the user
```
bgavrilMS, line 38: Region used to construct /token endpoint to contact ESTS.

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        internal string RegionUsed { get; }
```
bgavrilMS, line 40: Wondering if we should output RegionDetected as well.

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Contains the result of region when MSAL region discovery is used, 
+    /// published as part of AuthenticationResultMetadata.
+    /// <see cref="AuthenticationResultMetadata"/> for additional metadata 
+    /// information of the authentication result.
+    /// </summary>
+    public class RegionDiscoveryOutcome
```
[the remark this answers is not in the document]
bgavrilMS: maybe rename to `RegionDetails` since you should fill this property in even when auto-detection is not requested.

bgavrilMS:
Looks good. Please make sure that we don't use the message "Region discovery failed in the past, not trying again", as we want to have the root cause. "Region discovery failed in the past because X" is better.

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
pmaytak: Hmm, I think it would be helpful. I just remembered that we talked about this some time ago (#2698) and even started work on that (#2684).
Maybe it would be better to have only `RegionDetected` without `RegionUsed`?
- If a user passes in a region, since it's the one that is used, they already know it. So `RegionDetected` here is more helpful to know if region passed in is correct or not; `RegionUsed` is redundant.
- If user selected auto-detection, then `RegionDetected` is more helpful, and is the same as `RegionUsed`, which is again redundant.

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Region returned to the user
+        /// </summary>
+        internal string RegionUsed { get; }
+
+        /// <summary>
+        /// Error details when region auto detect fails
+        /// </summary>
+        internal string Exception { get; }
```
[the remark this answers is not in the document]
gladjohn: removed this and displaying the exception

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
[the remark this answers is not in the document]
gladjohn: modified

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/RegionDiscoveryOutcome.cs
gladjohn: RegionUsed already has the detected region when autodetection is requested. I added a telemetry property for this instead

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -19,14 +19,16 @@ internal sealed class RegionManager : IRegionManager
...
+            public RegionInfo(string region, RegionAutodetectionSource regionSource, string regionDetails)
             {
                 Region = region;
                 RegionSource = regionSource;
+                RegionDetails = regionDetails;
             }
 
             public string Region { get; }
             public RegionAutodetectionSource RegionSource { get; }
+            public string RegionDetails { get; }
```
SameerK-MSFT, line 31: make it readonly for better compiler optimization

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -51,6 +54,7 @@ public RegionInfo(string region, RegionAutodetectionSource regionSource)
             {
```
SameerK-MSFT, line 54: In general, it will be better to remove shouldClearStaticCache which is there for test only and add another extension method or use reflection to change values for testing.

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -73,6 +73,10 @@ public string RegionOutcomeString
         {
             get => _regionOutcome.HasValue ? _regionOutcome.Value.ToString("D") : null;
         }
+        public string AutoDetectedRegion { get; set; }
```
SameerK-MSFT, line 76: Can the sets be internal (if set only by this assembly) or private (if set only by this class)?

SameerK-MSFT (approved):
I have some suggested changes. Good overall.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
gladjohn: I will create a tracking item for this

PR #3177, opened 2022-02-22
[Prototype] L1 Wilson LRU cache

pmaytak:
Fixes # 

- Added maximum items option to the CacheOptions public API, which represents the count of items in the cache. 
- Added an instance of Wilson cache, as an L1Cache, to each TokenCache instance with key being a partition string and value being a user or app accessor. 
- Added GetAllValues method to Wilson cache.
- In the cache accessor classes, remove the partitions from the token dictionaries – the partitions are now at the TokenCache level, one level above accessors.
- Replaced the Accessor property in the TokenCache class with GetAccessor method which retrieves the correct accessor instance from the Wilson cache. 

Issues: 
- Wilson cache capacity represents a count of keys in the internal dictionary, and not size. This results in an inconsistent size measurement, since each partition can have a different number of tokens in it. 
- Presently, an instance of public/confidential client app contains a reference to two TokenCache instances (one for app cache and one for user cache) which in turn have an instance of a cache accessor. These cache and accessor instances are abstracted from each other. This creates a complexity of finding a good location for the Wilson cache. The prototype has two Wilson cache instances, for app and user caches. 
- Wilson cache doesn’t have a GetAll method. GetAll (for example, GetAllAccessTokens and similar) methods without a partition key are called primarily during serialization/deserialization in confidential client apps. Technically in this scenario, the cache would only have one partition, but still the call pattern would have to be complex, since a collection of accessors would be returned, and then iterated over to get all tokens for each. In the same vein, there’s GetAccounts method in public client apps, which would have to iterate through all partitions/accesors, adding to the complexity. 
- There are about 500 references to the Accessor property in the code, albeit only about 40 being from the main library and the rest from the test code. However, it still means that changing this code will lead to wide reaching changes throughout the code base. 

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/EventBasedLRUCache.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/EventBasedLRUCache.cs
@@ -1,0 +1,603 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Cache
+{
+    /// <summary>
+    /// This is an LRU cache implementation that relies on an event queue rather than locking to achieve thread safety.
+    /// This approach has been decided on in order to optimize the performance of the get and set operations on the cache.
+    /// This cache contains a doubly linked list in order to maintain LRU order, as well as a dictionary (map) to keep track of
+    /// keys and expiration times. The linked list (a structure which is not thread-safe) is NEVER modified directly inside
+    /// an API call (e.g. get, set, remove); it is only ever modified sequentially by a background thread. On the other hand,
+    /// the map is a <see cref="ConcurrentDictionary{TKey, TValue}"/> which may be modified directly inside an API call or
+    /// through eventual processing of the event queue. This implementation relies on the principle of 'eventual consistency':
+    /// though the map and it's corresponding linked list may be out of sync at any given point in time, they will eventually line up.
+    /// See here for more details:
+    /// https://aka.ms/identitymodel/caching
+    /// </summary>
+    /// <typeparam name="TKey">The key type to be used by the cache.</typeparam>
+    /// <typeparam name="TValue">The value type to be used by the cache</typeparam>
+    internal class EventBasedLRUCache<TKey, TValue>
+    {
+        internal delegate void ItemRemoved(TValue Value);
+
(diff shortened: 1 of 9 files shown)


src/client/Microsoft.Identity.Client/Cache/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,603 @@
...
+                    _doubleLinkedList.Remove(cacheItem);
+                    _doubleLinkedList.AddFirst(cacheItem);
+                });
+            }
+
+            value = cacheItem != null ? cacheItem.Value : default;
+            return cacheItem != null;
+        }
+
+        public IList<TValue> GetAllValues()
```
pmaytak, line 486: @dannybtsai Was there any talks about adding a GetAllValues method (kind of like this) that returns actual values and not just CacheEntries?

src/client/Microsoft.Identity.Client/Cache/EventBasedLRUCache.cs
```diff
@@ -0,0 +1,603 @@
...
+    /// See here for more details:
+    /// https://aka.ms/identitymodel/caching
+    /// </summary>
+    /// <typeparam name="TKey">The key type to be used by the cache.</typeparam>
+    /// <typeparam name="TValue">The value type to be used by the cache</typeparam>
+    internal class EventBasedLRUCache<TKey, TValue>
+    {
+        internal delegate void ItemRemoved(TValue Value);
+
+        private readonly int _capacity;
```
pmaytak, line 31: @dannybtsai Similarly, any discussions about allowing a user to specify current quantity themselves? For example, kind of like [.NET MemoryCache](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-6.0#use-setsize-size-and-sizelimit-to-limit-cache-size allows to specify the cache entry size in a Set method.

brentschmaltz: @pmaytak is there a way we can use the code directly from M.IM.Tokens?

brentschmaltz: We can add MSAL as a friend.

bgavrilMS:
Can we close this one @pmaytak ?

src/client/Microsoft.Identity.Client/Cache/EventBasedLRUCache.cs
brentschmaltz: @pmaytak that could be expensive as we have seen 100,000+ items in the cache.

Issue #2921, opened 2021-09-30
[Feature Request] Better static analyzers in MSAL .NET


bgavrilMS:
Our current static analyzers do not catch enough issues. 

- [ ] Upgrade the current async analyzer, as it is deprecated
- [ ] Review .NET static analyzers rules and see the impact of enabling the recommended rules
- [ ] Add some rules which do not imply lots of work
- [ ] Discuss with the team rules that imply lots of work

PR #3179, opened 2022-02-23
Update Async analyzer to use internal

trwalke:
Fixes # 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2921

Replacing depricated asyncUsageAnalyzers package with Microsoft.VisualStudio.Threading.Analyzers


.editorconfig
```diff
@@ -176,14 +176,11 @@ end_of_line = crlf
 # Code files
 [*.{cs,vb}]
 
-dotnet_diagnostic.UseAsyncSuffix.severity = error
+dotnet_diagnostic.VSTHRD200.severity = error
```
pmaytak, line 179: nit: would be helpful to have a description above each of these rules.

pmaytak (approved):
Is Microsoft.VisualStudio.Threading.Analyzers recommended upgrade for AsyncUsageAnalyzers? I thought the core Roslyn analyzers had this async rule built-in, although maybe I'm remembering incorrectly.

tests/Microsoft.Identity.Test.Android.UIAutomation/Microsoft.Identity.Test.Android.UIAutomation.csproj
```diff
@@ -29,6 +29,11 @@
     <Reference Include="System" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers">
```
bgavrilMS, line 32: Why don't these projects pick up Directory.Build.props ? We should not duplicate this code.

.editorconfig
```diff
@@ -176,14 +176,11 @@ end_of_line = crlf
...
 
-dotnet_diagnostic.UseConfigureAwait.severity = error
-
-dotnet_diagnostic.AvoidAsyncSuffix.severity = none
-
-dotnet_diagnostic.AvoidAsyncVoid.severity = none
+dotnet_diagnostic.VSTHRD111.severity = error
 
+dotnet_diagnostic.VSTHRD100.severity = none
```
bgavrilMS, line 183: So these diagnostics replace the old ones. Would be to see if we can enable all the other diagnostics from this analyzer as well (as a separate PR).

bgavrilMS (changes requested):
We should specify the new analyzer packages in a single Directory.Builds.props file and not in individual csproj

.editorconfig
trwalke: @bgavrilMS makes sense. Here is the list of the other analyzers in this package:
https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/index.md

Definitely a good idea to enable the critical ones at least

bgavrilMS (approved):

PR #3181, opened 2022-02-23
Add instance_aware extra query param only if one does not exist

neha-bhargava:
Fixes: Update to the PR #3141 for #2524 

When user selects .WithMultiCloudSupport(true) and calls interactive request we need to add the extraQueryParam to the request. But if the query param is already added, then ExtraQueryParam.Add() will throw an exception. So checking if the key exists before adding it.

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         private readonly AcquireTokenInteractiveParameters _interactiveParameters;
         private readonly IServiceBundle _serviceBundle;
         private readonly ICoreLogger _logger;
-        private readonly string _instanceAware = "instance_aware";
+        private readonly const string InstanceAwareParam = "instance_aware";
 
         #region For Test
         private readonly IAuthCodeRequestComponent _authCodeRequestComponentOverride;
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
```diff
@@ -112,9 +113,9 @@ private async Task<MsalTokenResponse> GetTokenResponseAsync(CancellationToken ca
                 cancellationToken.ThrowIfCancellationRequested();
             }
 
-            if (_requestParams.AppConfig.MultiCloudSupportEnabled)
+            if (_requestParams.AppConfig.MultiCloudSupportEnabled && !_requestParams.AppConfig.ExtraQueryParameters.ContainsKey(_instanceAware))
```
trwalke, line 116: So the MultiCloudSupportEnabled is set internally by MSAL and the user needs to ensure "instance_aware" is added via WithExtraQueryParameters() api?

trwalke (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
neha-bhargava: No we add it to the extra query param to the interactive request. User just sets WithMultiCloudSupport(true)

trwalke: ok cool

SameerK-MSFT (approved):

PR #3189, opened 2022-02-25
Trwalke/ruleset updates

trwalke:
Fixes #
Ruleset update

Changes proposed in this request
Enabling critical Async Threading rules
Adding ruleset from asp.net repo


.editorconfig
```diff
@@ -176,18 +176,306 @@ end_of_line = crlf
...
+dotnet_diagnostic.VSTHRD003.severity = warning
+
+# Await SwitchToMainThreadAsync
+dotnet_diagnostic.VSTHRD004.severity = error
+
+# Invoke single-threaded types on Main thread
+dotnet_diagnostic.VSTHRD010.severity = warning
+
+# Use AsyncLazy<T>
+dotnet_diagnostic.VSTHRD011.severity = none #error requires the use of Microsoft.VisualStudio.Threading to get access to the AsyncLazy<T> type
```
trwalke, line 208: inorder to enable this we need to use Microsoft.VisualStudio.Threading but the package is not compatable with our .net 45 targets. same with some other older targets.

@bgavrilMS

trwalke:
These rules produce an extra 100+ warnings and it may be easier to disable the problematic rules but not sure. thoughts @bgavrilMS 

image

[Reference Here](https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/index.md)

src/client/Microsoft.Identity.Client/ITokenCache.cs
```diff
@@ -78,8 +78,10 @@ public partial interface ITokenCache
         /// want to call <see cref="ITokenCacheSerializer.DeserializeMsalV3(byte[], bool)"/></remarks>
 #if !SUPPORTS_CUSTOM_CACHE
         [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
-# endif
+#endif
+#pragma warning disable VSTHRD200 // Use "Async" suffix for async methods
```
trwalke, line 82: This rule is a combined version of two rules we used to have. UseAsyncSuffix and AvoidAsyncSuffix

Since this name is already in the public api, changing it is a breaking change. Thus, I am adding a suppression. 

https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD200.md

src/client/Microsoft.Identity.Client.Desktop/WamInterop.cs
```diff
@@ -14,20 +14,20 @@ namespace Microsoft.Identity.Client.Platforms
 {
     internal static class WebAuthenticationCoreManagerInterop
     {
-        public static IAsyncOperation<WebTokenRequestResult> RequestTokenForWindowAsync(IntPtr hWnd, WebTokenRequest request)
```
bgavrilMS, line 17: These operations are async, but not in the traditional "way", but in the Windows way. You can create your own async API which can be await-ed by simply implementing an interface. The "async" suffix should be here. Please disable the rule for this file.

tests/devapps/XamarinDev/XamarinDev/AcquirePage.xaml.cs
```diff
@@ -190,7 +190,7 @@ private async void OnAcquireSilentlyClickedAsync(object sender, EventArgs e)
             }
         }
 
-        private async void OnAcquireClickedAsync(object sender, EventArgs e)
+        private async Task OnAcquireClickedAsync(object sender, EventArgs e)
```
bgavrilMS, line 193: Does the app continue to work? I was under the impression that async event handlers cannot use Task. This is the whole reason why we still have `async void` allowed in the language.

.editorconfig
bgavrilMS: Yeah, bummer.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -140,6 +140,7 @@
     <Compile Include="$(PathToMsalSources)\**\*.cs" Exclude="$(PathToMsalSources)\obj\**\*.*" />
     <Compile Remove="$(PathToMsalSources)\Platforms\**\*.*;$(PathToMsalSources)\Resources\*.cs" />
     <Compile Remove="$(PathToMsalSources)\PlatformsCommon\PlatformNotSupported\ApiConfig\SystemWebViewOptions.cs" />
+    <Compile Remove="GlobalSuppressions3.cs" />
```
trwalke, line 143: Probably dont need this?

trwalke:
Getting some strange errors around .net 4.8 after merging in master. Are you aware of any changes to .net 48?

bgavrilMS:
@trwalke - yeah, I'm trying to convert all our dev apps to net48, since net47 and net461 are no longer installed by default with VS.

tests/devapps/DesktopTestApp/GlobalSuppressions.cs
```diff
@@ -0,0 +1,8 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
```
pmaytak, line 1: Instead of GlobalSuppressions and pragma disable in code, can these rules be disabled in tests editorconfig for all tests?

pmaytak (approved):
Main comment is that it could be cleaner to disable some rules in the test editorconfig, instead of adding pragma disable or suppression files. Also can maybe add editorconfigs to different folders, since they are hierarchical.

This should also be merged right after the WAM bug fix PR before any other ones.

Issue #3121, opened 2022-01-24
[Bug] WithAccount(result.Account) doesn't work with WAM


bgavrilMS:
MSAL 4.40

1. AcquireTokenInteractive with WAM with no login hint -> account picker shows up
2. AcquireTokenInteractive with WAM for a different scope, using `.WithAccount(result.Account)`; 

Actual: account picker shows up displaying ALL accounts
Expected: account should be pre-selected

Root cause: tenant profiles seems to override the WAM account ID:

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/AuthenticationResult.cs#L145

Workaround: 

Use `app.GetAccountsAsync()` to get the accounts again and pass in that account.

bgavrilMS:
The account object from the cache has a property named `WamAccountId` which is needed to link an MSAL account to a WAM account. 

This is particularly important when using a 3p app and trying to use MSA accounts, because we can't link MSAL account to WAM account by usename.

PR #3199, opened 2022-03-01
Pass account to Auth result

neha-bhargava:
Fixes # 
#3121 

Pass account to auth result

**Testing**
Added unit tests

**Performance impact**


src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
```diff
@@ -153,6 +154,8 @@ private async Task<AuthenticationResult> CreateAuthenticationResultAsync(MsalAcc
         {
             var msalIdTokenItem = await CacheManager.GetIdTokenCacheItemAsync(cachedAccessTokenItem).ConfigureAwait(false);
             var tenantProfiles = await CacheManager.GetTenantProfilesAsync(cachedAccessTokenItem.HomeAccountId).ConfigureAwait(false);
+            var accounts = await CacheManager.GetAccountsAsync().ConfigureAwait(false);
+            var account = accounts.FirstOrDefault() as Account;
```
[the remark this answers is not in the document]
bgavrilMS: Have a look at `CacheManager.GetIdTokenCacheItemAsync(accessTokenItem)` - this gets an IdToken from the cache, using the data of the AT (env, home account id, tenant id etc.). We should have the same logic for account. 

You can go a step further and combine `GetTenantProfilesAsync` and this new `GetAccountAssociatedWithAccessToken` as well, because an Account object can be materialized from AccountCacheItem + tenant profiles.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
```diff
@@ -161,7 +164,8 @@ private async Task<AuthenticationResult> CreateAuthenticationResultAsync(MsalAcc
                 AuthenticationRequestParameters.AuthenticationScheme,
                 AuthenticationRequestParameters.RequestContext.CorrelationId,
                 TokenSource.Cache,
-                AuthenticationRequestParameters.RequestContext.ApiEvent);
+                AuthenticationRequestParameters.RequestContext.ApiEvent,
```
bgavrilMS, line 167: For testing I suggest you look at unit tests which start from an existing cache file, such as `MultiTenantTokenCache.json`

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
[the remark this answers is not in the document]
neha-bhargava: Changed

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1150,6 +1150,35 @@ MsalIdTokenCacheItem ITokenCacheInternal.GetIdTokenCacheItem(MsalAccessTokenCach
...
+        async Task<Account> ITokenCacheInternal.GetAccountAssociatedWithAccessTokenAsync(
+            AuthenticationRequestParameters requestParameters,
+            MsalAccessTokenCacheItem msalAccessTokenCacheItem)
+        {
+            if (requestParameters.AuthorityInfo.AuthorityType != AuthorityType.Aad)
+            {
+                return null;
+            }
+
+            Debug.Assert(msalAccessTokenCacheItem.HomeAccountId != null);
```
SameerK-MSFT, line 1162: This will work only in the debug mode. You may want to use exception here for all modes.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1150,6 +1150,35 @@ MsalIdTokenCacheItem ITokenCacheInternal.GetIdTokenCacheItem(MsalAccessTokenCach
...
+
+            var tenantProfiles = await (this as ITokenCacheInternal).GetTenantProfilesAsync(requestParameters, msalAccessTokenCacheItem.HomeAccountId).ConfigureAwait(false);
```
SameerK-MSFT, line 1164: Casting is not required.

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
```diff
@@ -152,16 +153,17 @@ private async Task<AuthenticationResult> RefreshRtOrFailAsync(CancellationToken
         private async Task<AuthenticationResult> CreateAuthenticationResultAsync(MsalAccessTokenCacheItem cachedAccessTokenItem)
         {
             var msalIdTokenItem = await CacheManager.GetIdTokenCacheItemAsync(cachedAccessTokenItem).ConfigureAwait(false);
-            var tenantProfiles = await CacheManager.GetTenantProfilesAsync(cachedAccessTokenItem.HomeAccountId).ConfigureAwait(false);
+            var account = await CacheManager.GetAccountAssociatedWithAccessTokenAsync(cachedAccessTokenItem).ConfigureAwait(false);
```
bgavrilMS, line 156: I'm wondering if we can go 2 steps further with this:

1. Eliminate `GetTenantProfilesAsync` completely. It does not seem to be needed, you only use it `AcquireTokenOnBehalfOf`, which can also get an Account object instead. `GetTenantProfilesAsync` can be made private.

2. Merge `GetIdTokenCacheItemAsync` with `GetAccountAssociatedWith...` - since we always make these 2 calls one after the other. There is no code path that gets just the IdToken but not the account. 

Thoughts?

bgavrilMS (changes requested):
1. No Account for ADFS / B2C shows that we are missing some tests, or some assertion on existing tests
2. We now access the CacheManager 3 times - once for Id token, once for tenant profiles and once for account. This is not needed - a single method can do them all.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs
neha-bhargava: I think it makes sense to make `GetTenantProfilesAsync` private and add to `GetAccountAssociatedWithCache`. But leave `GetIdTokenCacheItemAsync` separate since the `TenantProfiles` needs all the id tokens for the `HomeAccountId` and IdToken is for the particular key with partitioning. Combining Tenant profiles and Id token will not have any impact on the cache.

neha-bhargava:
> 1. No Account for ADFS / B2C shows that we are missing some tests, or some assertion on existing tests
> 2. We now access the CacheManager 3 times - once for Id token, once for tenant profiles and once for account. This is not needed - a single method can do them all.

The tests do not fail since the account object was used only to populate tenant profiles and wam id. Updating the code to use the account passed.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -141,14 +140,18 @@ public partial class AuthenticationResult
...
+            }
+            else if (homeAccountId != null)
             {
                 Account = new Account(
                     homeAccountId,
                     msalIdTokenCacheItem?.GetUsername(),
                     environment,
-                    tenantProfiles: tenantProfiles,
-                    wamAccountIds: wamAccountIds);
+                    tenantProfiles: account?.TenantProfiles,
```
bgavrilMS, line 153: well, these 2 will be `null`

bgavrilMS (approved):
Some minor comments remaining (cleanup)

src/client/Microsoft.Identity.Client/Cache/ICacheSessionManager.cs
```diff
@@ -26,7 +26,7 @@ internal interface ICacheSessionManager
         
         Task<IEnumerable<IAccount>> GetAccountsAsync();
         Task<IDictionary<string, TenantProfile>> GetTenantProfilesAsync(string homeAccountId);
-
+        Task<Account> GetAccountAssociatedWithAccessTokenAsync(MsalAccessTokenCacheItem msalAccessTokenCacheItem);
```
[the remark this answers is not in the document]
neha-bhargava: Deleting `GetTenantProfilesAsync` making it private in TokenCache. Leaving `GetIdTokenCacheItemAsync` since I did not combine it with tenant profiles as mentioned in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3199#discussion_r821187217

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
neha-bhargava: Right :)

tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -843,6 +843,64 @@ public async Task AcquireTokenSilent_EmptyLoginHint_TestAsync()
...
+                account = accounts.Single(a => a.HomeAccountId.TenantId == tenant2);
+                response = await
+                    pca.AcquireTokenSilent(new[] { "User.Read" }, account)
+                    .WithTenantId(tenant2)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+
+                // Assert
+                AssertWamIds(response.Account as Account, 1, "wamId2");
+                Assert.AreEqual(tenant2, response.TenantId);
```
pmaytak, line 899: Assert tenant profiles for second account too?

pmaytak (approved):

SameerK-MSFT (approved):
Tested with B2C and SIlentAuthentication. That works too

tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs
neha-bhargava: The second account will only have 1 tenant profile since only 1 id token is present for the home account id. That is why it is not verified for multiple tenant profiles. This is same behavior as existing tests on the cache for tenant profiles.

Issue #3137, opened 2022-02-03
[Feature Request] Enable Azure.Identity to benefit from MSAL.NET token cache when used for Managed identity


jmprieur:
**Is your feature request related to a problem? Please describe.**
Azure.Identity (Azure SDK) is using MSAL.NET for some flows, and in that case it benefits from MSAL.NET cache. But in other cases, it goes directly to the MSI endpoint, and does not cache the token.
The ask is to have a way for the Azure SDK to benefit from MSAL.NET cache

**Describe the solution you'd like**
The spec is [here](https://microsoft.sharepoint.com/:w:/t/activedirectory/acmi/EUyj9w-qLt5IqsqtXydnFdAB0nIbjbBQKbjTLl-RuWSpCw?e=QftciK)

**Describe alternatives you've considered**
See the spec

**Additional context**
This will be a consistency item,

jmprieur:
See also the previous spike: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/ffb0592e10dc5938980b02486d228406c9587370

bgavrilMS:
@trwalke - please sync with @schaabs on this first and also read the spec.

PR #3204, opened 2022-03-03
Trwalke/msi

trwalke:
Fixes # 
Early prototype for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3137

Adding new api for acquiring a token with a provided callback

**Testing**

**Performance impact**


src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class ExternalTokenResult
+    {
+        public string[] scopes { get; set; } // (or resource?)
```
bgavrilMS, line 18: pls review naming rules i.e. `Scopes`

bgavrilMS, line 18: Actually AAD does not give us back scopes for this flow. We just use the input scope to store data in the cache.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,29 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+        internal static MsalTokenResponse CreateFromAppProviderResponse(ExternalTokenResult tokenResponse)
+        {
+            var response = new MsalTokenResponse
+            {
+                AccessToken = tokenResponse.RawAccessToken,
+                RefreshToken = null,
+                IdToken = null,
+                TokenType = BrokerResponseConst.Bearer,
+                CorrelationId = tokenResponse.correlationId,
+                Scope = tokenResponse.scopes[0],
```
bgavrilMS, line 158: You need to pass in input scopes from the input obj.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,43 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
...
+        {
+            var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
+                                                          .WithAppTokenProvider(async (AppTokenProviderParameters parameters) =>
+                                                          {
+                                                              Assert.IsNotNull(parameters.Scopes);
+                                                              Assert.IsNotNull(parameters.CorrelationId);
+                                                              Assert.IsNotNull(parameters.TenantId);
+                                                              Assert.IsNotNull(parameters.CancellationToken);
+
+                                                              return await Task.Run(() => GetAppTokenProviderResult()).ConfigureAwait(false);
```
bgavrilMS, line 1687: I think `Task.Run` will dispatch the execcution to a background thread. Not what you need here. It's simpler to just adapt an existing object to a Task result, i.e. 

```csharp
var appTokenProvider = GetAppTokenProviderResult();
return Task.FromResult(appTokenProvider);
```

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,29 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+                ExpiresIn = DateTimeHelpers.GetDurationFromNowInSeconds(DateTimeHelpers.DateTimeToUnixTimestamp(tokenResponse.Expiry)),
```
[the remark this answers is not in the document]
trwalke: @bgavrilMS was trying to use the same methods we have in code that already handles this. However, no reason not to simplify.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,48 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
...
+                                                          {
+                                                              Assert.IsNotNull(parameters.Scopes);
+                                                              Assert.IsNotNull(parameters.CorrelationId);
+                                                              Assert.IsNotNull(parameters.TenantId);
+                                                              Assert.IsNotNull(parameters.CancellationToken);
+
+                                                              return Task.Run(() => GetAppTokenProviderResult()).Result;
+                                                          })
+                                                          .BuildConcrete();
+            await app.AcquireTokenForClient(TestConstants.s_scope).ExecuteAsync(new CancellationToken()).ConfigureAwait(false);
```
bgavrilMS, line 1690: nit: assert that the result comes from "identity provider"

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -294,6 +294,22 @@ public ConfidentialClientApplicationBuilder WithCacheSynchronization(bool enable
             return this;
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="AppTokenProviderDelegate"></param>
+        /// <returns></returns>
+        public ConfidentialClientApplicationBuilder WithAppTokenProvider(Func<AppTokenProviderParameters, Task<ExternalTokenResult>> AppTokenProviderDelegate)
```
bgavrilMS, line 302: naming `appTokenProviderDelegate`

Am curious - is VS not warning you about the naming violations?

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Tenant Id for client application
+        /// </summary>
+        public string TenantId { get; set; }
+
+        /// <summary>
+        /// Cancellation token for async operation
+        /// </summary>
+        public CancellationToken CancellationToken { get; set; }
```
bgavrilMS, line 31: Not needed, you don't use this.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,32 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
             return response;
         }
 
+        internal static MsalTokenResponse CreateFromAppProviderResponse(ExternalTokenResult tokenResponse)
+        {
+            var response = new MsalTokenResponse
```
bgavrilMS, line 151: We should do some validation of `tokenResponse` - it is mandatory to have most of these params, otherwise the token cache can be left in a bad state.

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    /// <summary>
+    /// Token result from external token provider
+    /// </summary>
+    public class ExternalTokenResult
+    {
+        /// <summary>
+        /// Correlationn Id to track request
+        /// </summary>
+        public string CorrelationId { get; set; }
```
bgavrilMS, line 21: I think we should remove this. I don't see the point of overrideing the correlation id. We should just use the correlation id from the input.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -117,10 +117,36 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+        private async Task<MsalTokenResponse> SendTokenRequestToProviderAsync(CancellationToken cancellationToken)
+        {
+            AppTokenProviderParameters appTokenProviderParameters = new AppTokenProviderParameters();
+            appTokenProviderParameters.Scopes = GetOverriddenScopes(AuthenticationRequestParameters.Scope);
+            appTokenProviderParameters.CorrelationId = AuthenticationRequestParameters.RequestContext.CorrelationId.ToString();
+            appTokenProviderParameters.Claims = AuthenticationRequestParameters.Claims;
+            appTokenProviderParameters.TenantId = AuthenticationRequestParameters.Authority.TenantId;
+            appTokenProviderParameters.CancellationToken = cancellationToken;
+
+            ExternalTokenResult externalToken = await Task.Run(() => ServiceBundle.Config.AppTokenProviderDelegate(appTokenProviderParameters)).ConfigureAwait(false);
```
bgavrilMS, line 143: Again, Task.Run will put this logic on a different thread and run it, then wait for it. while this works, it's much simpler to do

```csharp
await ServiceBundle.Config.AppTokenProviderDelegate(appTokenProviderParameters);
```

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,32 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+            {
+                AccessToken = tokenResponse.RawAccessToken,
+                RefreshToken = null,
+                IdToken = null,
+                TokenType = BrokerResponseConst.Bearer,
+                CorrelationId = tokenResponse.CorrelationId,
```
bgavrilMS, line 157: I don't see the point in getting a new correlation id from the "idp". I think in MSAL we just rely on the fact that the ESTS gives us back the same correlation id.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,32 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+                ExpiresIn = tokenResponse.ExpiresInSeconds - DateTimeHelpers.CurrDateTimeInUnixTimestamp(),
```
bgavrilMS, line 158: I still don't understand why you need to substract `DateTimeHelpers.CurrDateTimeInUnixTimestamp()`

ExpiresInSeconds is a number like 3600
`DateTimeHelpers.CurrDateTimeInUnixTimestamp() is the numebr of seconds between Jan 1, 1970 and now, 1646389878 at the time of writing. 
Result will be negative.

This indicates that you are missing a unit test ;)

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,48 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
...
             }
         }
+
+        [TestMethod]
+        public async Task ValidateAppTokenProviderAsync()
+        {
+            var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
+                                                          .WithAppTokenProvider(async (AppTokenProviderParameters parameters) =>
+                                                          {
+                                                              Assert.IsNotNull(parameters.Scopes);
```
bgavrilMS, line 1682: Can you assert that this function delegate is only invoked once please?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,48 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
...
+            Assert.AreEqual(TestConstants.DefaultAccessToken, result.AccessToken);
+            Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+        }
+
+        private ExternalTokenResult GetAppTokenProviderResult()
+        {
+            ExternalTokenResult token = new ExternalTokenResult();
+            token.RawAccessToken = TestConstants.DefaultAccessToken;
+            token.CorrelationId = Guid.NewGuid().ToString();
+            token.ExpiresInSeconds = 3600 + DateTimeHelpers.CurrDateTimeInUnixTimestamp();
```
bgavrilMS, line 1712: Ah, now I understand why your product code works. I'm pretty sure that ESTS returns ExpiresIn as `3600` not as `12349013409123`. 

We only use UnixTimestamps when we store the values in the token cache. This is done for you at a different layer. You just need to send `3600`

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,48 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
                 Assert.IsNull(acc);
             }
         }
+
+        [TestMethod]
+        public async Task ValidateAppTokenProviderAsync()
```
bgavrilMS, line 1677: We should have a test where the expiration is 1 second to see if expiration really works.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
trwalke: no, I dont see it

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
trwalke: Yes, I missed this one.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
bgavrilMS: Something for an OCE to fix :).

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
trwalke: Ok, I mix up the fact that ESTS returns just the length of time when the token is valid rather than the point of time in which the token will expire.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
trwalke: The method above uses the same logic as it gets the correlation id from the response on line 128.
CorrelationId = responseDictionary[BrokerResponseConst.CorrelationId],

Technically this line should also simply inject the correlation id from the input params then.

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
trwalke: Yeah, it was in the spec so I put it here, will remove it.

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
trwalke: was in the spec so wasn't sure if was going to be used for something down the road. will remove

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,57 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+            {
+                response.RefreshIn = response.ExpiresIn;
+            }
+
+            return response;
+        }
+
+        private static void ValidateExternalTokenResult(ExternalTokenResult externalTokenResult)
+        {
+            if (string.IsNullOrEmpty(externalTokenResult.RawAccessToken))
```
pmaytak, line 179: Do we need a check that if `RefreshInSeconds.HasValue`, `RefreshInSeconds.HasValue >= 0`? Unless we check that when we do refresh.

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <summary>
+    /// Token result from external token provider
+    /// </summary>
+    public class ExternalTokenResult
+    {
+        /// <summary>
+        /// Access token
+        /// Mandatory
+        /// </summary>
+        public string RawAccessToken { get; set; }
```
pmaytak, line 22: `AccessToken`? Because it's an actual access token?

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,57 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+            };
+
+            if (externalTokenResponse.RefreshInSeconds.HasValue)
+            {
+                response.RefreshIn = externalTokenResponse.RefreshInSeconds.Value;
+            }
+            else
+            {
+                response.RefreshIn = response.ExpiresIn;
+            }
```
pmaytak, line 172: nit: 
```suggestion
            response.RefreshIn = externalTokenResponse.RefreshInSeconds ?? response.ExpiresIn;
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,65 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
...
+        {
+            var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
+                                                          .WithAppTokenProvider(async (AppTokenProviderParameters parameters) =>
+                                                          {
+                                                              Assert.IsNotNull(parameters.Scopes);
+                                                              Assert.IsNotNull(parameters.CorrelationId);
+                                                              Assert.IsNotNull(parameters.TenantId);
```
pmaytak, line 1684: Add `Assert.IsNotNull(parameters.Claims);`

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,65 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
...
+                                                              Assert.IsNotNull(parameters.CancellationToken);
+
+                                                              return Task.Run(() => GetAppTokenProviderResult()).Result;
+                                                          })
+                                                          .BuildConcrete();
+
+            //AcquireToken from app provider
```
pmaytak, line 1691: Wonder if we should add a new token source maybe? ExternalTokenProvider? Or at least maybe log where the token came from. I think currently we log the token endpoint, but in this case it would be null, so from the logs if would seem like this is a cached token.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1672,65 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
...
+            AuthenticationResult result = await app.AcquireTokenForClient(TestConstants.s_scope).ExecuteAsync(new CancellationToken()).ConfigureAwait(false);
+
+            Assert.IsNotNull(result.AccessToken);
+            Assert.AreEqual(TestConstants.DefaultAccessToken, result.AccessToken);
+            Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+
+            var tokens = app.AppTokenCacheInternal.Accessor.GetAllAccessTokens();
+
+            Assert.IsTrue(tokens.Count > 0);
```
pmaytak, line 1700: nit
```suggestion
            Assert.AreEqual(1, tokens.Count);
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
pmaytak: Also another call with different scopes (assert Idp source + 2 tokens in cache).

pmaytak (approved):
Mostly just naming feedback - want to ensure the public API naming is consistent and clear. Also some additional logging is needed for the new code path.

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Token result from external token provider
+    /// </summary>
+    public class ExternalTokenResult
```
bgavrilMS, line 16: We should continue to use the term "AppToken"

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Expiration of token
+        /// Mandatory
+        /// </summary>
+        public long ExpiresInSeconds { get; set; } 
+
+        /// <summary>
+        /// When the token should be refreshed proactivly. (Optional)
+        /// If not provided computed as Expiry-DateTimeOffset.Now()
```
bgavrilMS, line 37: If not provided, the token should not be refreshed proactively.

src/client/Microsoft.Identity.Client/AppConfig/ExternalTokenResult.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Access token
+        /// Mandatory
+        /// </summary>
+        public string RawAccessToken { get; set; }
+
+        /// <summary>
+        /// Tenant Id for client application
+        /// </summary>
+        public string TenantId { get; set; }
```
bgavrilMS, line 27: Is there a case where the result.tenant_id is different than input.tenant_id? I think in client_credentials we always treat it as the same tenant, because ESTS does not tell us if the token is for a different tenant. This is a user scenario only, where we derive this info from idtoken `tid` claim.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,57 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+            var response = new MsalTokenResponse
+            {
+                AccessToken = externalTokenResponse.RawAccessToken,
+                RefreshToken = null,
+                IdToken = null,
+                TokenType = BrokerResponseConst.Bearer,
+                ExpiresIn = externalTokenResponse.ExpiresInSeconds,
+                ClientInfo = null,
+                TokenSource = TokenSource.IdentityProvider,
+                TenantId = externalTokenResponse.TenantId
```
bgavrilMS, line 162: If you leave this as `null`, I think MSAL will use the input tenant id by default, which is ok for client_credentials.

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -146,6 +146,57 @@ internal static MsalTokenResponse CreateFromiOSBrokerResponse(Dictionary<string,
...
+            };
+
+            if (externalTokenResponse.RefreshInSeconds.HasValue)
+            {
+                response.RefreshIn = externalTokenResponse.RefreshInSeconds.Value;
+            }
+            else
+            {
+                response.RefreshIn = response.ExpiresIn;
```
bgavrilMS, line 171: response.RefreshIn is of type `long?` so it can be null. This block can be ` response.RefreshIn = externalTokenResponse.RefreshInSeconds`

bgavrilMS (approved):

bgavrilMS (approved):
Approved with comments, particularly around tenantid.

src/client/Microsoft.Identity.Client/AppConfig/AppTokenProviderParameters.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class AppTokenProviderParameters
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        public IEnumerable<string> Scopes { get; internal set; }
```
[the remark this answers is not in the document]
trwalke: Tests have access to the MSAL internals so they can already set these properties

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
trwalke: It is checked when we do refresh.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
bgavrilMS: Not sure about adding a new token source, because it'll confuse 99% of users who don't use this status. `IdentityProvider` seems ok in this case. Agreed that we need some logging statements to make sure this is set.

PR #3210, opened 2022-03-04
Conditional Access for Android

SameerK-MSFT:
Fixes 3185, 3203

MSAL.NET handles protection_policy_required suberror and throws IntuneAppProtectionPolicyRequiredException

**Testing**
Unit Tests

**Performance impact**
None


src/client/Microsoft.Identity.Client/Internal/Broker/BrokerResponseConst.cs
```diff
@@ -37,7 +37,7 @@ internal static class BrokerResponseConst
         public const string TenantId = "tenant_id";
         public const string UserName = "username";
         public const string iOSBrokerNonce = "broker_nonce"; // included in request and response with iOS Broker v3
-        public const string iOSBrokerTenantId = "utid";
+        public const string iOSBrokerTenantId = @"utid";
```
bgavrilMS, line 40: nit: @ is not needed, `utid` has no special chars.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -83,6 +83,9 @@ private static bool IsThrottled(OAuth2ResponseBase oAuth2Response)
                 ex = new IntuneAppProtectionPolicyRequiredException(errorCode, subErrorCode)
                 {
                     Upn = msalTokenResponse.Upn,
+                    AuthorityUrl = msalTokenResponse.AuthorityUrl,
```
bgavrilMS, line 86: Will these be avaiable on both iOS and Android?

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -151,8 +154,13 @@ private static bool IsInvalidGrant(string errorCode, string subErrorCode)
 
         private static bool IsAppProtectionPolicyRequired(string errorCode, string subErrorCode)
         {
+#if iOS
             return string.Equals(errorCode, BrokerResponseConst.iOSBrokerProtectionPoliciesRequiredErrorCode, StringComparison.OrdinalIgnoreCase)
                              && string.Equals(subErrorCode, MsalError.ProtectionPolicyRequired, StringComparison.OrdinalIgnoreCase);
+#else
```
bgavrilMS, line 160: # elif ANDROID

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -176,7 +180,11 @@ internal static MsalTokenResponse CreateFromAndroidBrokerResponse(string jsonRes
                 ExtendedExpiresIn = DateTimeHelpers.GetDurationFromNowInSeconds(authResult[BrokerResponseConst.ExtendedExpiresOn].ToString()),
                 ClientInfo = authResult[BrokerResponseConst.ClientInfo].ToString(),
                 TokenType = authResult[BrokerResponseConst.TokenType]?.ToString() ?? "Bearer",
-                TokenSource = TokenSource.Broker
+                TokenSource = TokenSource.Broker,
+                AuthorityUrl = authResult[BrokerResponseConst.Authority].ToString(),
```
bgavrilMS, line 184: Do you need to add the null protection `.?`  here as well?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerRequest.cs
```diff
@@ -121,6 +121,12 @@ private static BrokerRequest FromAuthenticationParameters(AuthenticationRequestP
                 br.Claims = _authenticationRequestParameters.Claims;
             }
 
+            if (_authenticationRequestParameters.RequestContext.ServiceBundle.Config.ClientCapabilities?.Any() == true)
```
bgavrilMS, line 124: How about just using `ClaimsAndClientCapabilities` here? i.e. `br.Claims = _authenticationRequestParameters.ClaimsAndClientCapabilities`

src/client/Microsoft.Identity.Client/Platforms/iOS/iOSPlatformProxy.cs
```diff
@@ -22,6 +22,7 @@ namespace Microsoft.Identity.Client.Platforms.iOS
     /// </summary>
     internal class iOSPlatformProxy : AbstractPlatformProxy
     {
+        private const string EnrollmentIdKey = "intune_app_protection_enrollment_id_V1";
```
bgavrilMS, line 25: Does not seem used. Is it needed?

bgavrilMS (changes requested):
Looks ok, but see the comments about `#if Android` and `ClaimsAndClientCapabilities`

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
SameerK-MSFT: Good catch

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
SameerK-MSFT: It should be present when successful. Added for safety.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerRequest.cs
SameerK-MSFT: It would override the existing claims. This is additive i.e. just adds to any existing.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
SameerK-MSFT: Authority is null in iOS. Others are available on both. Being null should not matter in iOS as it is not used.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/BrokerRequest.cs
```diff
@@ -121,6 +121,12 @@ private static BrokerRequest FromAuthenticationParameters(AuthenticationRequestP
...
+            {
+                br.Claims = ClaimsHelper.GetMergedClaimsAndClientCapabilities(br.Claims,
```
bgavrilMS, line 126: no need to merge, `ClaimsAndClientCapabilities` is already merged. so instead of block 119-122 and 124-129, you can have a single 

`if (ClaimsAndClientCapabilities.Any())
{
 br.Claims = ClaimsAndClientCapabilities;
}

SameerK-MSFT: Fixed.

tests/devapps/Intune-xamarin-Android/MainActivity.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// <param name="e">arguments</param>
+        private async void ActButton_Click(object sender, EventArgs e)
+        {
+            AuthenticationResult result = null;
+
+            try
+            {
+                // attempt silent login.
+                // If this is very first time and the device is not enrolled, it will throw MsalUiRequiredException
+                // If the device is enrolled, this will succeed.
```
pmaytak, line 52: Nit: It will still throw, since no token in cache?

tests/devapps/Intune-xamarin-Android/MainActivity.cs
```diff
@@ -0,0 +1,150 @@
...
+        private async Task DoMAMRegister(IntuneAppProtectionPolicyRequiredException exProtection)
+        {
+            // reset the registered event
+            IntuneSampleApp.MAMRegsiteredEvent.Reset();
+            
+            // Invoke compliance API on a different thread
+            await Task.Run(() =>
+                                {
+                                    IMAMComplianceManager mgr = MAMComponents.Get<IMAMComplianceManager>();
+                                    mgr.RemediateCompliance(exProtection.Upn, exProtection.AccountUserId, exProtection.TenantId, exProtection.AuthorityUrl, false);
```
pmaytak, line 112: aside: variable names...yikes...
```csharp
    public interface IMAMComplianceManager : IJavaObject, IDisposable
    {
        void RemediateCompliance(string p0, string p1, string p2, string p3, bool p4);
    }
```

tests/devapps/Intune-xamarin-Android/EnrollmentNotificationReceiver.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// If the receiver tried to take some action in response to the notification but failed to complete that action it should return false.
+        /// </returns>
+        public bool OnReceive(IMAMNotification notification)
+        {
+            if (notification.Type == MAMNotificationType.MamEnrollmentResult)
+            {
+                IMAMEnrollmentNotification enrollmentNotification = notification.JavaCast<IMAMEnrollmentNotification>();
+                MAMEnrollmentManagerResult result = enrollmentNotification.EnrollmentResult;
+
+                if (result.Equals(MAMEnrollmentManagerResult.EnrollmentSucceeded))
```
pmaytak, line 35: What if enrollment fails? The event is not reset, will the app hang?

tests/devapps/Intune-xamarin-Android/MAMWEAuthCallback.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Intune_xamarin_Android
+{
+    /// <summary>
+    /// Required by the MAM SDK. A token may be needed very early in the app lifecycle so the ideal
+    /// place to register the callback is in the OnMAMCreate() method of the app's implementation
+    /// of IMAMApplication.
+    /// See https://docs.microsoft.com/en-us/intune/app-sdk-android#account-authentication
+    /// </summary>
+    class MAMWEAuthCallback : Java.Lang.Object, IMAMServiceAuthenticationCallback
```
pmaytak, line 25: What is this class needed for? (Skimmed the docs but isn't clear.) When MAM tries to enroll the app, it needs this token? What if the silent auth fails, no interactive call?

tests/devapps/Intune-xamarin-Android/PCAWrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (acct != null)
+            {
+                var silentParamBuilder = PCA.AcquireTokenSilent(scopes, acct);
+                var authResult = await silentParamBuilder
+                                            .ExecuteAsync().ConfigureAwait(false);
+                return authResult;
+            }
+            else
+            {
+                throw new MsalUiRequiredException("ErrCode", "ErrMessage");
```
pmaytak, line 109: Should have better message - Silent call failed in PCAWrapper. No accounts.

pmaytak (approved):

bgavrilMS (approved):

LibsAndSamples.sln
```diff
@@ -177,6 +177,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "WebApi", "WebApi", "{CC07F2
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WebApi", "tests\devapps\WebApi\WebApi.csproj", "{DD23D8FF-86BA-4E9F-8AF1-0EBE0D86986A}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Intune-xamarin-Android", "tests\devapps\Intune-xamarin-Android\Intune-xamarin-Android.csproj", "{C6CA2F00-E48A-478F-ABC3-1FE3FE20286B}"
```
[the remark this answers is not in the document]
SameerK-MSFT: If it is added to other apps, the user will have to register the device with MAM service and iOS dev will not be able to user simulator. So better to be separate.

tests/devapps/Intune-xamarin-Android/MainActivity.cs
SameerK-MSFT: Yep. That is generated android binding.

tests/devapps/Intune-xamarin-Android/EnrollmentNotificationReceiver.cs
SameerK-MSFT: Good point. I added a comment in the mainactivity.cs (where the App waits) that handling of those scenarios will be specific to the app as per the business requirements and not considered in the sample app.

tests/devapps/Intune-xamarin-Android/MAMWEAuthCallback.cs
SameerK-MSFT: This is a callback when ```IMAMEnrollmentManager``` starts enrollment. ```AcquireToken``` is passed url of the MAM service as resource and expects AT from the app for it. The silent will not fail as it is called after user does interactive login. So user has been authenticated at this time. It may be confusing that silent is called after ATI. But the call is to get AT for MAM service, (which is different from the original resource that the user requested in ATI.)

tests/devapps/Intune-xamarin-Android/PCAWrapper.cs
SameerK-MSFT: Removed the redundant check. Leftover from debugging.

tests/devapps/Intune-xamarin-Android/MainActivity.cs
SameerK-MSFT: Removed the redundant chk.

trwalke (approved):
Took a look at the broker code. Seems like everything is ok. Definitly will need to run though the various scenarios during release testing.

SameerK-MSFT (commented):
The merge will be held till SP is out.

PR #3220, opened 2022-03-10
Add a workaround for PRT update failure on /organizations

### LINKED ISSUE #3217: [Bug] Tenant errors in interactive auth if authority is env/organziations with WAM
4.33.0

authority: env/organizations

Try to login interactively with an AAD account. 

Actual: the browser shows an error such as "Specified tenant identifier 'organizations' is neither a valid DNS name, nor a valid external domain."

bgavrilMS:
Fixes #3217

Replace env/organizations with env/common when we can reliably detect an AAD interactive login. 

**Testing**
unit

manual

1. {normal app, msa-pt app} with authority {global cloud, sovereign cloud}/organizations logging in an {aad/msa} account, {silent, interactive}
(16 tests)

2. log in the default account {aad, msa}, {silent, interactive}  with {normal, msa-pt} app (8 tests) 

**Performance impact**
only string comparisons

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
@@ -313,6 +313,12 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
             }
         }
 
+        /// <summary>
+        /// Some WAM operations fail for work and school accounts when the authority is env/organizations
+        /// Chaing the authority to env/common in this case works around this problem.
+        /// 
+        /// https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3217
+        /// </summary>
         private async Task<string> WorkaroundOrganizationsBugAsync(
             AuthenticationRequestParameters authenticationRequestParameters,
             WebAccountProvider webAccountProvider)
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
@@ -213,7 +213,8 @@ namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
                isAccountInWam: false)
                 .ConfigureAwait(false);
 
-            WamAdapters.AddMsalParamsToRequest(authenticationRequestParameters, webTokenRequest, _logger);
+            string differentAuthority = await WorkaroundOrganizationsBugAsync(authenticationRequestParameters, provider).ConfigureAwait(true);
+            WamAdapters.AddMsalParamsToRequest(authenticationRequestParameters, webTokenRequest, _logger, differentAuthority);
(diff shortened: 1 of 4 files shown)


crmann1 (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
```diff
@@ -313,20 +313,34 @@ private bool CanSkipAccountPicker(Authority authority)
             }
         }
 
-        private string WorkaroundOrganizationsBug(AuthenticationRequestParameters authenticationRequestParameters, WebAccount wamAccount)
+        private async Task<string> WorkaroundOrganizationsBugAsync(
+            AuthenticationRequestParameters authenticationRequestParameters,
+            WebAccountProvider webAccountProvider)
         {
             string differentAuthority = null;
-            if (!_wamOptions.MsaPassthrough &&
```
pmaytak, line 319: Previous filter:
Not MSA-PT **AND** Organizations request **AND** (Organizations provider **OR** Any Default OS account)

Current filter:
Organizations request **AND** (Organizations provider **OR** Default OS AAD account)

Are we missing the inclusion of Default OS MSA account in the filter?

Also in which cases is organizations request is true but organizations provider is false? Is it when Default OS account is used? Provider is default then?

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
```diff
@@ -634,7 +634,7 @@ public async Task ATI_WithoutPicker_AccountMatch_Organizations_Async()
                     "https://login.microsoft.com", "organizations");
                 var webAccount = new WebAccount(wamAccountProvider, "user@contoso.com", WebAccountState.Connected);
                 var webTokenRequest = new WebTokenRequest(wamAccountProvider);
-
+                _webAccountProviderFactory.IsOrganizationsProvider(wamAccountProvider).Returns(true);
```
pmaytak, line 637: See my comment in WamBroker, do we need a test where this would return false?

src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/WamBroker.cs
bgavrilMS: MSA accounts are not affected by this bug. Previously I did not test with this setup (machine with only 1 MSA account on it).

bgavrilMS: `Also in which cases is organizations request is true but organizations provider is false? Is it when Default OS account is used? Provider is default then?`

In MSA-PT scenario, the authority is env/organizations, and account provider can be MSA.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
bgavrilMS: I think there are tests like this, but let me add another one.

Issue #3028, opened 2021-11-18
MSAL logging improvements


bgavrilMS:
### Discussed in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/discussions/3010

<div type='discussions-op-text'>

<sup>Originally posted by **bgavrilMS** November 12, 2021</sup>
Suggested improvements that can be made to MSAL logging: 

## Common log interface for MSAL/SAL/Wilson/MISE
- [x] https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3245

## Apps should not have to recompile to get verbose logs

- [x] Allow apps to change the logging verbosity based on an env variable (this will require a process restart)
- [x] Allow apps to change logging verbosity programmatically #2868 
- [ ] Allow apps to collect event logs, even if a logging callback has not been enabled (WithLogging + enable default logger already uses event source, but it's not turned on by default).

## Better logging in token caches
- [ ] Microsoft.Identity.TokenCache should reuse MSAL's logger 
- [x] The desktop cache should reuse MSAL's logger
- [x] Expose the correlation ID to token cache implementations #3008

## Logging perf
- [ ] Logging should not compute log strings unless the log level is met. 

## Add more info in the logs
- [x] #2029
- [x] Log important metrics irrespective of the log level #3004 

Additional discussion around logging: #1993 
</div>

bgavrilMS:
This quarter we want to do only: 

- token cache logic uses MSAL loggers
- #2868

jmprieur:
@bgavrilMS dynamically? you mean on AcquireTokenXXBuilder?

bgavrilMS:
no, I mean: 

```csharp
...
.WithLogging(() =>_logLevel , LogCallback);

private static LogLevel _logLevel;

// apps monitor config or env changes
private void OnConfigChange()
{
    _logLevel = // read log level from config
}

```

trwalke:
updating details with https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3245

bgavrilMS:
This has not been completed.

trwalke:
@bgavrilMS yeah, not sure how this got closed. Must a bug somewhere.

bgavrilMS:
Tracking item for logging improvements.

bgavrilMS:
T

PR #3232, opened 2022-03-17
Adding external logger for token cache

trwalke:
Fixes # 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3028

Adding external logging interface for external token cache logging

**Testing**
Unit


trwalke:
It turns out removing ICoreLogger and the Microsoft.Identity.Client.Core affects a lot of files, I will need to create a list of relevant files fo the review. I used VS to remove invalid usings for the "*.Core" namespace since it was present in over 100+ files and it ended up removing a lot of extra ones too. however, everything is working fine and the usings in the src code are now cleaned up.

@bgavrilMS

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -1672,6 +1674,55 @@ public async Task ValidateGetAccountAsyncWithNullEmptyAccountIdAsync(string acco
                 Assert.IsNull(acc);
             }
         }
+
+        [TestMethod]
+        public async Task ExternalMsalLoggerTestAsync()
```
trwalke, line 1679: Test demonstrating the use of the new API

bgavrilMS:
@trwalke - best thing to do imo is to separate out the important changes in a commit, and all the refactoring in a separate commit. Otherwise it's impossible to review.

PR #3233, opened 2022-03-17
Reuse lists in token cache filtering logic.

### LINKED ISSUE #3178: [Bug] Improve token filtering logic
4.40.0

**Actual behavior**
In [FindAccessTokenAsync](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b32458fde68d27929cc35f8af2eeda1b8257de22/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L464) and [FilterWithLogging](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b32458fde68d27929cc35f8af2eeda1b8257de22/src/client/Microsoft.Identity.Client/Utils/EnumerableExtensions.cs#L44), inefficient Linq is used when filtering the token collection resulting in unnecessary List object allocation.

**Possible solution**
Main tasks:
- [ ] Replace `IReadOnlyList` with `List` in accessor and filtering methods.
- [ ] In `FilterWithLogging`, replace `list = list.Where(predicate).ToList();` with `list.RemoveAll(e => !predicate(e));`
- [ ] Investigate if the same changes have to be made on other methods, like the refresh token filtering.
- [ ] (optional) Investigate using `SegmentedList` instead of `List`
- [ ] Run performance tests

Context from conversation:

> Notice there are 4+ branches, all from a single method. Here is the source code in MSAL library:
>  
>             IReadOnlyList<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true, partitionKey);
>             ...
>  
>             tokenCacheItems = FilterByHomeAccountTenantOrAssertion(requestParams, tokenCacheItems);
>             tokenCacheItems = FilterByTokenType(requestParams, tokenCacheItems);
>             tokenCacheItems = FilterByScopes(requestParams, tokenCacheItems);
>             tokenCacheItems = await FilterByEnvironmentAsync(requestParams, tokenCacheItems).ConfigureAwait(false);
>  
>        internal static IReadOnlyList<T> FilterWithLogging<T>(
>            this IReadOnlyList<T> list,
>            Func<T, bool> predicate,
>            ICoreLogger logger,
>            string logPrefix)
>         {
>             if (logger.IsLoggingEnabled(LogLevel.Verbose))
>             {
>                 logger.Verbose($"{logPrefix} - item count before: {list.Count} ");
>             }
>  
>             list = list.Where(predicate).ToList();
>  
>             if (logger.IsLoggingEnabled(LogLevel.Verbose))
>             {
>                 logger.Verbose($"{logPrefix} - item count after: {list.Count} ");
>             }
>  
>             return list;
>         }
>  
> What if you change the code to:
>  
>             List<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true, partitionKey);
>  
>             requestParams.RequestContext.Logger.Always($"[FindAccessTokenAsync] Discovered {tokenCacheItems.Count} access tokens in cache using partition key: {partitionKey}");
>  
>             if (tokenCacheItems.Count == 0)
>             {
>  
>                 logger.Verbose("No access tokens found in the cache. Skipping filtering. ");
>                 requestParams.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.NoCachedAccessToken;
>  
>                 if (AcquireTokenInLongRunningOboWasCalled(requestParams))
>                 {
>                     logger.Error("[FindAccessTokenAsync] AcquireTokenInLongRunningProcess was called and OBO token was not found in the cache.");
>                     throw new MsalClientException(MsalError.OboCacheKeyNotInCacheError, MsalErrorMessage.OboCacheKeyNotInCache);
>                 }
>  
>                 return null;
>             }
>  
>             FilterByHomeAccountTenantOrAssertion(requestParams, tokenCacheItems);
>             FilterByTokenType(requestParams, tokenCacheItems);
>             FilterByScopes(requestParams, tokenCacheItems);
>             await FilterByEnvironmentAsync(requestParams, tokenCacheItems).ConfigureAwait(false);
>  
> With:
>  
>         internal static void FilterWithLogging<T>(
>            this List<T> list,
>            Func<T, bool> predicate,
>            ICoreLogger logger,
>            string logPrefix)
>         {
>             if (logger.IsLoggingEnabled(LogLevel.Verbose))
>             {
>                 logger.Verbose($"{logPrefix} - item count before: {list.Count} ");
>             }
>  
>             list.RemoveAll(e => !predicate(e));
>  
>             if (logger.IsLoggingEnabled(LogLevel.Verbose))
>             {
>                 logger.Verbose($"{logPrefix} - item count after: {list.Count} ");
>             }
>         }
>  
> Now you just need a single list, with much reduced LOH allocations. The single list can either be reused or switched to use segmented list. So no LOH allocation is totally achievable.
>  
> Given that we’re doing multiple levels of filtering, it’s even possible to apply the most effective filter first, and/or combine them into a single stage filtering.
>  
> I said quite a few times that Linq.Enumerable is total disregard of proper data structure and algorithm design. The flip side of the story is you need precise selection/craftmanship in your data structure and algorithm design.
>

pmaytak:
Fixes #3178

- Replace `IReadOnlyList` with `List` in accessor and filtering methods (`RemoveAll` is in concrete `List` implementation).
- In `FilterWithLogging`, replace `list = list.Where(predicate).ToList();` with `list.RemoveAll(e => !predicate(e));`
  - Add a boolean parameter whether to filter on the current list or create a new one.
- In other methods, replace `Where` with `RemoveAll` and reverse filtering predicate (since instead of keeping the items based on a filter, we now remove items based on the opposite of that filter).

**Testing**
Existing tests.

**Performance impact**
Tested with 1 tenant and 10K tokens in cache (tenants don't matter in this scenario). About 40% reduction in memory allocated for client, OBO, and silent calls. Perf improvement is minimal, only about .5 ms for silent call.
**Before**
|                Method |  CacheSize |        Mean |     Error |      StdDev |      Median |         Min |         Max |     Gen 0 | Allocated |
|---------------------- |----------- |------------:|----------:|------------:|------------:|------------:|------------:|----------:|----------:|
| AcquireTokenForClient | (1, 10000) |  5,353.3&nbsp;μs | 411.87&nbsp;μs | 1,207.93&nbsp;μs |  5,424.8&nbsp;μs | 2,887.60&nbsp;μs |  7,847.5&nbsp;μs |         - |  1,988&nbsp;KB |
|    AcquireTokenForOBO | (1, 10000) |  6,193.0&nbsp;μs | 341.06&nbsp;μs |   978.57&nbsp;μs |  6,045.6&nbsp;μs | 4,182.15&nbsp;μs |  8,638.4&nbsp;μs |         - |  2,256&nbsp;KB |
|    AcquireTokenSilent | (1, 10000) |  5,652.5&nbsp;μs | 378.56&nbsp;μs | 1,086.15&nbsp;μs |  5,605.9&nbsp;μs | 3,390.45&nbsp;μs |  8,480.9&nbsp;μs |         - |  2,249&nbsp;KB |
|            GetAccount | (1, 10000) |    184.6&nbsp;μs |  31.33&nbsp;μs |    87.85&nbsp;μs |    152.3&nbsp;μs |    92.75&nbsp;μs |    432.4&nbsp;μs |         - |     11&nbsp;KB |
|           GetAccounts | (1, 10000) |    175.9&nbsp;μs |  29.71&nbsp;μs |    81.82&nbsp;μs |    151.8&nbsp;μs |    95.20&nbsp;μs |    431.2&nbsp;μs |         - |     11&nbsp;KB |
|         RemoveAccount | (1, 10000) | 10,414.1&nbsp;μs | 211.16&nbsp;μs |   615.97&nbsp;μs | 10,262.3&nbsp;μs | 8,266.30&nbsp;μs | 12,159.5&nbsp;μs | 1000.0000 | 11,075&nbsp;KB |

**After**
|                Method |  CacheSize |        Mean |     Error |      StdDev |      Median |         Min |         Max |     Gen 0 | Allocated |
|---------------------- |----------- |------------:|----------:|------------:|------------:|------------:|------------:|----------:|----------:|
| AcquireTokenForClient | (1, 10000) |  4,716.3&nbsp;μs  | 349.99&nbsp;μs | 1,020.94&nbsp;μs |  4,567.2&nbsp;μs | 2,504.50&nbsp;μs |  6,964.0&nbsp;μs |         - |  1,219&nbsp;KB |
|    AcquireTokenForOBO | (1, 10000) |  5,997.8&nbsp;μs | 300.82&nbsp;μs |   882.25&nbsp;μs |  5,898.9&nbsp;μs | 4,146.40&nbsp;μs |  8,369.4&nbsp;μs |         - |  1,230&nbsp;KB |
|    AcquireTokenSilent | (1, 10000) |  4,847.9&nbsp;μs | 298.60&nbsp;μs |   861.54&nbsp;μs |  4,939.9&nbsp;μs | 2,915.85&nbsp;μs |  7,107.8&nbsp;μs |         - |  1,224&nbsp;KB |
|            GetAccount | (1, 10000) |    167.0&nbsp;μs |  29.82&nbsp;μs |    83.61&nbsp;μs |    129.5&nbsp;μs |    81.10&nbsp;μs |    449.9&nbsp;μs |         - |     10&nbsp;KB |
|           GetAccounts | (1, 10000) |    180.4&nbsp;μs |  36.04&nbsp;μs |   102.24&nbsp;μs |    138.6&nbsp;μs |    82.80&nbsp;μs |    514.6&nbsp;μs |         - |     10&nbsp;KB |
|         RemoveAccount | (1, 10000) | 10,475.1&nbsp;μs | 376.17&nbsp;μs | 1,085.33&nbsp;μs | 10,465.9&nbsp;μs | 7,783.30&nbsp;μs | 13,363.2&nbsp;μs | 1000.0000 | 10,562&nbsp;KB |

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -263,8 +263,8 @@ namespace Microsoft.Identity.Client
 
             if (requestParameters.RequestContext.Logger.IsLoggingEnabled(LogLevel.Verbose))
             {
-                var accessTokenCacheItems = GetAllAccessTokensWithNoLocks(filterByClientId: false);
-                var refreshTokenCacheItems = GetAllRefreshTokensWithNoLocks(filterByClientId: false);
+                var accessTokenCacheItems = Accessor.GetAllAccessTokens();
+                var refreshTokenCacheItems = Accessor.GetAllRefreshTokens();
                 var accessTokenCacheKeys = accessTokenCacheItems.Take(10).Select(item => item.GetKey()).ToList();
 
                 StringBuilder tokenCacheKeyDump = new StringBuilder();
@@ -459,7 +459,7 @@ namespace Microsoft.Identity.Client
             string partitionKey = CacheKeyFactory.GetKeyFromRequest(requestParams);
             Debug.Assert(partitionKey != null || !requestParams.IsConfidentialClient, "On confidential client, cache must be partitioned.");
 
-            var accessTokens = GetAllAccessTokensWithNoLocks(true, partitionKey);
+            var accessTokens = Accessor.GetAllAccessTokens(partitionKey);
 
             requestParams.RequestContext.Logger.Always($"[FindAccessTokenAsync] Discovered {accessTokens.Count} access tokens in cache using partition key: {partitionKey}");
 
@@ -482,6 +482,7 @@ namespace Microsoft.Identity.Client
             FilterByTokenType(requestParams, accessTokens);
             FilterByScopes(requestParams, accessTokens);
             accessTokens = await FilterByEnvironmentAsync(requestParams, accessTokens).ConfigureAwait(false);
+            FilterByClientId(accessTokens);
 
             CacheRefreshReason cacheInfoTelemetry = CacheRefreshReason.NotApplicable;
 
@@ -749,6 +750,11 @@ namespace Microsoft.Identity.Client
         }
         #endregion
 
+        private void FilterByClientId<T>(List<T> tokenCacheItems) where T : MsalCredentialCacheItemBase
+        {
+            tokenCacheItems.RemoveAll(x => !x.ClientId.Equals(ClientId, StringComparison.OrdinalIgnoreCase));
+        }
+
         /// <summary>
         /// For testing purposes only. Expires ALL access tokens in memory and fires OnAfterAccessAsync event with no cache key
         /// </summary>
@@ -944,8 +950,13 @@ namespace Microsoft.Identity.Client
             // this will either be the home account ID or null, it can never be OBO assertion or tenant ID
             string partitionKey = CacheKeyFactory.GetKeyFromRequest(requestParameters);
 
-            var refreshTokenCacheItems = GetAllRefreshTokensWithNoLocks(filterByClientId, partitionKey);
+            var refreshTokenCacheItems = Accessor.GetAllRefreshTokens(partitionKey);
             var accountCacheItems = Accessor.GetAllAccounts(partitionKey);
+
+            if (filterByClientId)
+            {
+                FilterByClientId(refreshTokenCacheItems);
+            }
 
             if (logger.IsLoggingEnabled(LogLevel.Verbose))
                 logger.Verbose($"GetAccounts found {refreshTokenCacheItems.Count} RTs and {accountCacheItems.Count} accounts in MSAL cache. ");
@@ -1109,7 +1120,8 @@ namespace Microsoft.Identity.Client
 
             Debug.Assert(homeAccountId != null);
 
-            var idTokenCacheItems = GetAllIdTokensWithNoLocks(true, homeAccountId);
+            var idTokenCacheItems = Accessor.GetAllIdTokens(homeAccountId);
+            FilterByClientId(idTokenCacheItems);
 
             if (!requestParameters.AppConfig.MultiCloudSupportEnabled)
             {
@@ -1248,7 +1260,7 @@ namespace Microsoft.Identity.Client
 
             string partitionKey = account.HomeAccountId.Identifier;
 
-            var refreshTokens = GetAllRefreshTokensWithNoLocks(false, partitionKey);
+            var refreshTokens = Accessor.GetAllRefreshTokens(partitionKey);
             refreshTokens.RemoveAll(item => !item.HomeAccountId.Equals(account.HomeAccountId.Identifier, StringComparison.OrdinalIgnoreCase));
 
             // To maintain backward compatibility with other MSALs, filter all credentials by clientID if
@@ -1258,7 +1270,7 @@ namespace Microsoft.Identity.Client
             // Delete all credentials associated with this IAccount
             if (filterByClientId)
             {
-                refreshTokens.RemoveAll(x => !x.ClientId.Equals(ClientId, StringComparison.OrdinalIgnoreCase));
+                FilterByClientId(refreshTokens);
             }
 
             foreach (MsalRefreshTokenCacheItem refreshTokenCacheItem in refreshTokens)
@@ -1268,8 +1280,12 @@ namespace Microsoft.Identity.Client
 
             requestContext.Logger.Info($"Deleted {refreshTokens.Count} refresh tokens.");
 
-            var accessTokens = GetAllAccessTokensWithNoLocks(filterByClientId, partitionKey);
+            var accessTokens = Accessor.GetAllAccessTokens(partitionKey);
             accessTokens.RemoveAll(item => !item.HomeAccountId.Equals(account.HomeAccountId.Identifier, StringComparison.OrdinalIgnoreCase));
+            if (filterByClientId)
+            {
+                FilterByClientId(accessTokens);
+            }
 
             foreach (MsalAccessTokenCacheItem accessTokenCacheItem in accessTokens)
             {
@@ -1278,8 +1294,12 @@ namespace Microsoft.Identity.Client
 
             requestContext.Logger.Info($"Deleted {accessTokens.Count} access tokens.");
 
-            var idTokens = GetAllIdTokensWithNoLocks(filterByClientId, partitionKey);
+            var idTokens = Accessor.GetAllIdTokens(partitionKey);
             idTokens.RemoveAll(item => !item.HomeAccountId.Equals(account.HomeAccountId.Identifier, StringComparison.OrdinalIgnoreCase));
+            if (filterByClientId)
+            {
+                FilterByClientId(idTokens);
+            }
 
             foreach (MsalIdTokenCacheItem idTokenCacheItem in idTokens)
             {
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -256,28 +256,26 @@ namespace Microsoft.Identity.Client
             }
         }
 
-        //This method pulls all of the acceess and refresh tokens from the cache and can therefore be very impactful on performance.
+        //This method pulls all of the access and refresh tokens from the cache and can therefore be very impactful on performance.
         //This will run on a background thread to mitigate this.
         private void DumpCacheToLogs(AuthenticationRequestParameters requestParameters)
         {
 
             if (requestParameters.RequestContext.Logger.IsLoggingEnabled(LogLevel.Verbose))
             {
-                IReadOnlyList<MsalAccessTokenCacheItem> accessTokenCacheItems = GetAllAccessTokensWithNoLocks(filterByClientId: false);
-                IReadOnlyList<MsalRefreshTokenCacheItem> refreshTokenCacheItems = GetAllRefreshTokensWithNoLocks(filterByClientId: false);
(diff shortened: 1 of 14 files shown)


src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
```diff
@@ -25,6 +24,14 @@ public static IReadOnlyList<T> GetEmptyReadOnlyList<T>()
 #endif
         }
 
+#if HAVE_METHOD_IMPL_ATTRIBUTE
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+#endif
+        public static List<T> GetEmptyList<T>()
```
bgavrilMS, line 30: Is this worth having? There's no interesting logic around it.

pmaytak: Yea, I was thinking if there's more efficient way of creating an empty list, but since it's a concrete type, seems like no. Was leaning towards keeping this as a form of abstraction, but might remove it.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1236,50 +1225,54 @@ bool ITokenCacheInternal.HasTokensNoLocks()
 
             string partitionKey = account.HomeAccountId.Identifier;
 
-            var allRefreshTokens = GetAllRefreshTokensWithNoLocks(false, partitionKey)
-                .Where(item => item.HomeAccountId.Equals(account.HomeAccountId.Identifier, StringComparison.OrdinalIgnoreCase))
-                .ToList();
+            var refreshTokens = GetAllRefreshTokensWithNoLocks(false, partitionKey);
```
pmaytak, line 1228: Can probably combine the filtering + removal here more.

src/client/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -167,10 +167,8 @@ private void UpdateAppMetadata(string clientId, string environment, string famil
             if (!requestParams.IsClientCredentialRequest)
```
pmaytak, line 167: Can probably combine this further, if not for the log on line 165

bgavrilMS: Perhaps the log at line 160 is sufficient?

bgavrilMS:
@pmaytak  - did you invert the 2 perf tables in the description? Because the "before" looks better than the "After".

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -461,11 +459,11 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
             string partitionKey = CacheKeyFactory.GetKeyFromRequest(requestParams);
             Debug.Assert(partitionKey != null || !requestParams.IsConfidentialClient, "On confidential client, cache must be partitioned.");
 
-            IReadOnlyList<MsalAccessTokenCacheItem> tokenCacheItems = GetAllAccessTokensWithNoLocks(true, partitionKey);
+            var accessTokens = GetAllAccessTokensWithNoLocks(true, partitionKey);
```
bgavrilMS, line 462: I'm thinking that this filter here is wrong. We get all tokens and filter by client_id, but it's highly unlikely that there will be more than 1 client id in the whole cache.

So we are not filtering by the most restrictive filter first, which for S2S is the tenant id. `GetAllAccessTokensWithNoLocks` should not even have this flag to filter by client id, as it promotes misuse.

Thoughts?

src/client/Microsoft.Identity.Client/TokenCache.cs
```diff
@@ -200,33 +200,39 @@ private bool RtMatchesAccount(MsalRefreshTokenCacheItem rtItem, MsalAccountCache
             return homeAccIdMatch && clientIdMatch;
         }
 
-        private IReadOnlyList<MsalRefreshTokenCacheItem> GetAllRefreshTokensWithNoLocks(bool filterByClientId, string partitionKey = null)
+        private List<MsalRefreshTokenCacheItem> GetAllRefreshTokensWithNoLocks(bool filterByClientId, string partitionKey = null)
```
bgavrilMS, line 203: As stated above, it might be better to remove this filter from here, as it leads to misuse.

bgavrilMS (commented):
Please see the comment around filtering by tenantId first and then by clientid.

pmaytak:
> @pmaytak - did you invert the 2 perf tables in the description? Because the "before" looks better than the "After".

Yes, :) Corrected.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: So your suggestion is to still keep client ID filter, and just move it down the filtering list?

For client credential calls, the partition includes client ID, so this filter is unnecessary. For user related calls, the partition is per user, which could have tokens for different client IDs (whether the cache is static or serialized), right? But yes, the filter is the least restrictive, and probably least likely to happen.

Also `WithNoLocks` in the method name is unnecessary, since we never lock when we use accessors anymore.

pmaytak: @bgavrilMS Changes made in 292de9f2312ce83e8afe7a264054549322d8a0c6.

bgavrilMS: It's a good point about the partition key.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -480,27 +478,28 @@ private static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdTo
                 return null;
             }
 
-            tokenCacheItems = FilterByHomeAccountTenantOrAssertion(requestParams, tokenCacheItems);
-            tokenCacheItems = FilterByTokenType(requestParams, tokenCacheItems);
-            tokenCacheItems = FilterByScopes(requestParams, tokenCacheItems);
-            tokenCacheItems = await FilterByEnvironmentAsync(requestParams, tokenCacheItems).ConfigureAwait(false);
+            FilterByHomeAccountTenantOrAssertion(requestParams, accessTokens);
```
trwalke, line 481: For these filtering methods, it is better to return the object that is being operated on to reveal which parameter the method is changing, rather than returning void. And or change the name to include the work "tokens" at least.

tests/Microsoft.Identity.Test.Performance/Program.cs
```diff
@@ -13,13 +13,19 @@ class Program
...
-                    .AddDiagnoser(MemoryDiagnoser.Default)
-                    .AddJob(
-                        Job.Default
-                            .WithId("Job-PerfTests")));
+            BenchmarkSwitcher.FromTypes(new[] {
+                typeof(AcquireTokenForClientCacheTests),
+                typeof(AcquireTokenForOboCacheTests),
+                typeof(TokenCacheTests),
+            }).RunAll(DefaultConfig.Instance
+                //.WithOptions(ConfigOptions.DisableLogFile)
```
trwalke, line 21: is this needed?

trwalke (approved):

PR #3239, opened 2022-03-22
Fix for UWP packaging

### LINKED ISSUE #3184: [Bug] Error in creating UWP package for Microsoft Store Upload
MSAL.NET 4.41.0

**Platform**
UWP

* Desktop / Mobile
    *  Interactive

The app is in production and currently using legacy Live SDK, 
I am trying to upgraded MSAL, but started seeing this issue.

**Expected behavior**
It should create a UWP app package to upload to store.

**Actual behavior**
App package creation fails with following error
C:\Users\User\\.nuget\packages\microsoft.net.native.compiler\2.2.11-rel-30601-02\tools\Microsoft.NetNative.targets(805,5): warning :       MCG : warning MCG0007: Unresolved P/Invoke method 'ntdll.dll!RtlGetVersion' for method 'System.Int32 Microsoft.Identity.Client.PlatformsCommon.Shared.Win32VersionApi.RtlGetVersion(Microsoft.Identity.Client.PlatformsCommon.Shared.Win32VersionApi.OSVERSIONINFOEXW)'. Calling this method would throw exception at runtime. Please make sure the P/Invoke either points to a Windows API allowed in UWP applications, or a native DLL that is part of the package. If for some reason your P/Invoke does not satisfy those requirements, please use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP APIs.

C:\Users\User\\.nuget\packages\microsoft.net.native.compiler\2.2.11-rel-30601-02\tools\Microsoft.NetNative.targets(805,5): error : Internal compiler error: Exception of type 'System.OutOfMemoryException' was thrown.

bgavrilMS:
Fixes #3184

UWP does not provide a mechanism to tell if machine is running Win Server 2016, so we just assume WAM is present on all UWP compatible OSes. 

On UWP, it seems that we cannot have any "illegal" DllImport statetems in the code, even if these are not in the call path.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/Win32VersionApi.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/Win32VersionApi.cs
@@ -6,9 +6,11 @@ using System.Runtime.InteropServices;
 
 namespace Microsoft.Identity.Client.PlatformsCommon.Shared
 {
+
     /// <summary>
     /// Windows OS Version checks
     /// </summary>
+    /// <remarks>Do not include this code in UWP, it causes packaging errors</remarks>
     internal static class Win32VersionApi
     {
 
@@ -215,4 +217,5 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
             }
         }
     }
+
 }
--- a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
+++ b/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
@@ -89,12 +89,14 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
         {
 #if WINDOWS_APP
             return true;
-#else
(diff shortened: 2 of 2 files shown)


src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/Win32VersionApi.cs
```diff
@@ -6,9 +6,11 @@
 
 namespace Microsoft.Identity.Client.PlatformsCommon.Shared
 {
+
     /// <summary>
     /// Windows OS Version checks
     /// </summary>
+    /// <remarks>Do not include this code in UWP, it causes packaging errors</remarks>
     internal static class Win32VersionApi
```
bgavrilMS, line 14: The "DesktopOs/Win32" directory is excluded from the UWP build.

pmaytak (approved):
Anything useful that we can add to the WAM wiki, https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/WAM?

trwalke (approved):

Issue #3245, opened 2022-03-23
[Feature Request] Add new logger interface that is in alignment with SAL/Wilson/MISE


trwalke:
**Is your feature request related to a problem? Please describe.**
We need to add a new interface that customers will be able to use with MSAL.NET where the customer will be able to reuse the same logger implementation with SAL/Wilson/MISE without having to reimplement anything. 

**Describe the solution you'd like**
Add a new api where MSAL will accept an implementation of a logger interface (IIdentityLogger) that MSAL will use for all of its logging operations.

Expose this logger implementation to the external cache api (TokenCacheNotificationArgs) so that the customer can log messages during caching operations if needed.

**Additional context**
The current logging api:

    public T WithLogging(
                LogCallback loggingCallback,
                LogLevel? logLevel = null,
                bool? enablePiiLogging = null,
                bool? enableDefaultPlatformLogging = null)
        
will be made obsolete by the new logging interface api. An adapter can be added in MSAL to adapt this api to use the new logger implementation to maintain backwards compatibility so that we don't break current customer functionality.

Should Msal.net extensions should probably be updated to use the same logging strategy?

bgavrilMS:
Fixed in MSAL 4.44

PR #3247, opened 2022-03-24
IdentityLogger Integration

trwalke:
Fixes # 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3245

Adding identity logger interface to api. The Identity Logger will be implemented by the developer and provided to MSAL via the WithLogging() api. The logger callback is still functional but is considered legacy now. You will see the two implementations of these loggers in the logger adapter classes

Critical changes are in these files
image

including LoggerTests.cs

**Testing**

**Performance impact**

**Code Tour**
[identityloggerupdates.zip](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/files/8856070/identityloggerupdates.zip)

Download code tour extension in VS Code
Open code tour:
image


src/client/Microsoft.Identity.Client/Internal/Logger/MsalLoggerExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+            logEntry.EventLevel = EventLevel.Verbose;
+
+            logger.Log(logEntry);
+        }
+
+        //Since there is not a good way to determine if the pii value is needed by the logger at the moment, we need to make a check.
+        private static void LogToAvailableLogger(this IIdentityLogger logger, LogEntry logEntry, bool containsPii)
+        {
+            //Not the best way to do this. Needs refactoring.
+            if (containsPii && logger is LegacyIdentityLoggerAdapter loggerAdapter)
```
trwalke, line 85: This is clearly not the best way to do this but it does demonstrate that we need some way of sending the containsPii value to the legacy logger callback.

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLoggerExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+using Microsoft.IdentityModel.Logging.Abstractions;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Extension methods for ICoreLogger
+    /// </summary>
+    internal static class MsalLoggerExtensions
+    {
+        public static void Always(this IIdentityLogger logger, string message)
```
trwalke, line 18: Adding extension methods here to help avoid the need to change 100s of lines of logger statements in the codebase. these will simply adapt the current log statements to the new LogEntry parameter

src/client/Microsoft.Identity.Client/Internal/Logger/LegacyIdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            InvokeLogCallback(logEntry, true);
+        }
+
+        private void InvokeLogCallback(LogEntry logEntry, bool containsPii = false)
+        {
+            _logCallback.Invoke(GetLegacyLogLevel(logEntry.EventLevel), logEntry.Message, containsPii);
+        }
+
+        public LegacyIdentityLoggerAdapter(LogLevel logLevel, LogCallback logCallback)
```
bgavrilMS, line 41: Add `isPiiEnabled` here.

src/client/Microsoft.Identity.Client/Internal/Logger/LogEntry.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.Identity.Client;
+
+//Will be replaced by logger package when available
+
+namespace Microsoft.IdentityModel.Logging.Abstractions
+{
+    /// <summary>
+    /// Defines the structure of a log entry.
+    /// </summary>
+    public class LogEntry
```
bgavrilMS, line 19: Shouldn't a `LogEntry` also have a `IsPii` property?

src/client/Microsoft.Identity.Client/Internal/Logger/MsalLoggerExtensions.cs
bgavrilMS: The log adapter should encapsulate this better.

src/client/Microsoft.Identity.Client/Internal/Logger/LogEntry.cs
trwalke: The design of the new identity logger is to assume that messages will contain Pii if the IsPiiEnabled property on the IIdentityLogger is set to true and if it is set to false, messages will not contain Pii. This is different from the old way we used to do it where we would identify if each message contains Pii or not. This is why this class does not have isPii. 

However, after creating this prototype, I noticed that we do need some way to identity whether or not a message contains Pii since the legacy log adapter will need that information for the callback. However, I'm not sure adding a property to the new LogEntry just for our legacy logger callback is the right thing to do?

src/client/Microsoft.Identity.Client/Internal/Logger/LegacyIdentityLoggerAdapter.cs
trwalke: yes

src/client/Microsoft.Identity.Client/Internal/Logger/MsalCacheLoggerWrapper.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Diagnostics.Tracing;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.Internal.Logger
+{
+    internal class MsalCacheLoggerWrapper : IIdentityLogger
```
trwalke, line 15: This is the class to be exposed in the TokenCacheNotificationArgs for logging. Im assuming there is no Pii from the logs but I am thinking about adding an extension method to the interface to handle Pii

trwalke: @bgavrilMS

trwalke, line 15: This is the class to be exposed in the TokenCacheNotificationArgs for logging. Im assuming there is no Pii from the logs at the momemnt but I am thinking about adding an extension method to the interface to handle Pii messages.

@bgavrilMS

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerAdapterHelper.cs
```diff
@@ -0,0 +1,166 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.PlatformsCommon.Factories;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.Internal.Logger
+{
+    internal class LoggerAdapterHelper
```
trwalke, line 19: I ran into a potential pitfall by having this logic in the extensions class. It is solved by moving it to a helper class used internally by the logger adapter

trwalke:
Tests will be committed soon. just trying to figure out one more issue preventing them from running:

Error	CS0012	The type 'CLSCompliantAttribute' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.	Microsoft.Identity.Client (net45)	C:\git\MSAL\src\client\Microsoft.Identity.Client\CSC	1	C:\git\MSAL\src\client\Microsoft.Identity.Client	N/A

src/client/Microsoft.Identity.Client/Internal/Logger/LegacyLogger.cs
```diff
@@ -0,0 +1,285 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.PlatformsCommon.Factories;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.Internal.Logger
+{
+    //internal abstract class MsalLoggerx
```
trwalke, line 16: Legacy logger to be deleted

NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Local" value="C:\git\Nuget" />
```
bgavrilMS, line 6: Don't check this in. Better to have this in your VS settings.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -10,9 +10,9 @@
 
   <!-- Mobile and legacy targets - comment these out or set env variable MSAL_DESKTOP_ONLY_DEV to speedup the build -->
   <PropertyGroup Condition="'$(MSAL_DESKTOP_ONLY_DEV)' == ''">
-    <TargetFrameworkNetStandard>netstandard1.3</TargetFrameworkNetStandard>
+    <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
```
bgavrilMS, line 13: Would have been better to do this change as a separate PR.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -141,6 +140,8 @@
     <Compile Remove="$(PathToMsalSources)\Platforms\**\*.*;$(PathToMsalSources)\Resources\*.cs" />
     <Compile Remove="$(PathToMsalSources)\PlatformsCommon\PlatformNotSupported\ApiConfig\SystemWebViewOptions.cs" />
     <Compile Remove="GlobalSuppressions3.cs" />
```
bgavrilMS, line 142: Delete all the *cs files from here.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -293,5 +294,7 @@
     </None>
   </ItemGroup>
 
-
+  <ItemGroup>
+    <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="6.18.0" />
```
bgavrilMS, line 298: 6.18 ? What is the versioning story for `Microsoft.IdentityModel.Abstractions` ?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -94,7 +94,6 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkUap)' ">
     <!-- These are for MSAL json/* build infra -->
     <DefineConstants>$(DefineConstants);PORTABLE;DOTNET;HAVE_ASYNC;HAVE_BIG_INTEGER;HAVE_COVARIANT_GENERICS;HAVE_DATA_CONTRACTS;HAVE_DATE_TIME_OFFSET;HAVE_FSHARP_TYPES;HAVE_GUID_TRY_PARSE;HAVE_HASH_SET;HAVE_IGNORE_DATA_MEMBER_ATTRIBUTE;HAVE_INOTIFY_COLLECTION_CHANGED;HAVE_INOTIFY_PROPERTY_CHANGING;HAVE_ISET;HAVE_LINQ;HAVE_METHOD_IMPL_ATTRIBUTE;HAVE_NON_SERIALIZED_ATTRIBUTE;HAVE_READ_ONLY_COLLECTIONS;HAVE_REFLECTION_BINDER;HAVE_SERIALIZATION_BINDER_BIND_TO_NAME;HAVE_STRING_JOIN_WITH_ENUMERABLE;HAVE_TIME_SPAN_PARSE_WITH_CULTURE;HAVE_TIME_SPAN_TO_STRING_WITH_CULTURE;HAVE_TIME_ZONE_INFO;HAVE_UNICODE_SURROGATE_DETECTION;HAVE_VARIANT_TYPE_PARAMETERS;HAVE_VERSION_TRY_PARSE;HAVE_OBSOLETE_FORMATTER_ASSEMBLY_STYLE;HAVE_CONCURRENT_DICTIONARY;HAVE_ICONVERTIBLE;$(AdditionalConstants)</DefineConstants>
```
bgavrilMS, line 96: I think you'll need to change the JSON compilation flags - it might be enough to just use the ones from https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Newtonsoft.Json.csproj

bgavrilMS: We'll need to test this. The only way I can think about testing this is to manually compile out our netcore and netfx versions and run the tests. The test would pick up the netstndard part of the lib.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: Not sure, Ill ask Sruthi

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,98 @@
...
+    internal class IdentityLoggerAdapter : ILoggerAdapter
+    {
+        private readonly IIdentityLogger _identityLogger;
+        private string _clientInfo;
+        private string _correlationId;
+
+        public bool PiiLoggingEnabled { get; }
+        public bool IsDefaultPlatformLoggingEnabled { get; } = false;
+        public string ClientName { get; }
+        public string ClientVersion { get; }
```
neha-bhargava, line 23: Can we include [ApiId](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5e29d349dde99775bf60f23812e2a2437bb0cca9/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs#L11) here? This will help understand which flow the log is for.

src/client/Microsoft.Identity.Client/Internal/Logger/LegacyIdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using Microsoft.Identity.Client.Core;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.Internal.Logger
+{
+    internal class LegacyIdentityLoggerAdapter : ILoggerAdapter
+    {
+        LogLevel _minLogLevel = LogLevel.Always;
```
neha-bhargava, line 14: Should this not be LogLevel.Error? Since it is the minLogLevel

src/client/Microsoft.Identity.Client/Internal/Logger/LegacyIdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            PiiLoggingEnabled = enablePiiLogging;
+            _logCallback = loggingCallback;
+            _minLogLevel = logLevel;
+            IsDefaultPlatformLoggingEnabled = isDefaultPlatformLoggingEnabled;
+        }
+
+        public void Log(LogLevel logLevel, string messageWithPii, string messageScrubbed)
+        {
+            //LogEntry entry = _loggerAdapterHelper.Log(this, logLevel, messageWithPii, messageScrubbed);
```
neha-bhargava: Remove commented code?

src/client/Microsoft.Identity.Client/Internal/Logger/LegacyIdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            string message = LoggerAdapterHelper.FormatLogMessage(messageWithPii, messageScrubbed, PiiLoggingEnabled, _correlationId, _clientInfo);
+
+            if (IsLoggingEnabled(logLevel))
+            {
+                _logCallback.Invoke(logLevel, message, string.IsNullOrEmpty(messageWithPii) ? true : false);
+            }
+        }
+
+        //public LogEntry Log(ILoggerAdapter logger, EventLevel logLevel, string messageWithPii, string messageScrubbed)
```
neha-bhargava: Commented code

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerAdapterHelper.cs
```diff
@@ -0,0 +1,161 @@
...
+            return clientInformation;
+        }
+
+        public static ILoggerAdapter CreateLogger(
+            Guid correlationId,
+            ApplicationConfiguration config)
+        {
+            if (config.IdentityLogger == null)
+            {
+                return LegacyIdentityLoggerAdapter.Create(correlationId, config);
```
neha-bhargava, line 50: Should this be renamed to DefaultLogger instead?

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -273,6 +274,26 @@ public T WithLegacyCacheCompatibility(bool enableLegacyCacheCompatibility = true
...
+        /// Sets the Identity Logger 
+        /// </summary>
+        /// <param name="identityLogger">IdentityLogger</param>
+        /// <param name="enablePiiLogging">Boolean used to enable/disable logging of
+        /// Personally Identifiable Information (PII).
+        /// PII logs are never written to default outputs like Console, Logcat or NSLog
+        /// Default is set to <c>false</c>, which ensures that your application is compliant with GDPR.
+        /// You can set it to <c>true</c> for advanced debugging requiring PII
+        /// </param>
+        /// <returns></returns>
```
pmaytak, line 287:
```suggestion
        /// <summary>
        /// Sets an instance of <c>IIdentityLogger</c>. For details see https://aka.ms/msal-net-logging
        /// </summary>
        /// <param name="identityLogger">IIdentityLogger</param>
        /// <param name="enablePiiLogging">Boolean used to enable/disable logging of
        /// Personally Identifiable Information (PII).
        /// PII logs are never written to default outputs like Console, Logcat or NSLog
        /// Default is set to <c>false</c>, which ensures that your application is compliant with GDPR.
        /// You can set it to <c>true</c> for advanced debugging requiring PII
        /// </param>
        /// <returns></returns>
```

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            return LoggerAdapterHelper.LogBlockDuration(this, measuredBlockName, logLevel);
+        }
+
+        public DurationLogHelper LogMethodDuration(LogLevel logLevel = LogLevel.Verbose, [CallerMemberName] string methodName = null, [CallerFilePath] string filePath = null)
+        {
+            return LoggerAdapterHelper.LogMethodDuration(this, logLevel, methodName, filePath);
+        }
+
+        public EventLogLevel GetEventLogLevel(LogLevel logLevel)
```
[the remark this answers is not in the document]
trwalke: Also, It needs to be public so developers can implement the interface using the new logger

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public DurationLogHelper LogBlockDuration(string measuredBlockName, LogLevel logLevel = LogLevel.Verbose)
+        {
+            return LoggerAdapterHelper.LogBlockDuration(this, measuredBlockName, logLevel);
+        }
+
+        public DurationLogHelper LogMethodDuration(LogLevel logLevel = LogLevel.Verbose, [CallerMemberName] string methodName = null, [CallerFilePath] string filePath = null)
+        {
+            return LoggerAdapterHelper.LogMethodDuration(this, logLevel, methodName, filePath);
+        }
```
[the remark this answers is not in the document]
trwalke: the LogBlockDuration is definitely redundant and can be removed. All it is doing is returning a new DurationLogHelper. 
The LogMethodDuration however is different. 

Also, I didnt want to mix interface implementation with inheritance implementation so I only used interfaces and helpers.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/LoggerTests.cs
```diff
@@ -23,33 +29,51 @@ public void TestInit()
...
-        [DataRow(LogLevel.Warning)]
-        [DataRow(LogLevel.Info)]
-        [DataRow(LogLevel.Verbose)]
-        public void CallbackLoggerTest(LogLevel level)
+        [DataRow(LogLevel.Always, true)]
+        [DataRow(LogLevel.Error, true)]
+        [DataRow(LogLevel.Warning, true)]
+        [DataRow(LogLevel.Info, true)]
+        [DataRow(LogLevel.Verbose, true)]
+        public void CallbackLoggerTest(LogLevel level, bool UseLegaccyLogger)
```
pmaytak, line 74: Add a similar test for IdentityLogger? Although ExternalMsalLoggerTestAsync kind of covers that?

Was also thinking a test when both WithLogging methods are called - assert that IdentityLogger is used.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -10,9 +13,9 @@
 
   <!-- Mobile and legacy targets - comment these out or set env variable MSAL_DESKTOP_ONLY_DEV to speedup the build -->
   <PropertyGroup Condition="'$(MSAL_DESKTOP_ONLY_DEV)' == ''">
-    <TargetFrameworkNetStandard>netstandard1.3</TargetFrameworkNetStandard>
+    <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
     <TargetFrameworkNetDesktop45>net45</TargetFrameworkNetDesktop45>
-    <TargetFrameworkUap>uap10.0</TargetFrameworkUap>
+    <TargetFrameworkUap>uap10.0.17763</TargetFrameworkUap>
```
pmaytak, line 18: So this will be a breaking change for projects using these targets?
netstandard13
.Net Core 1.x
.Net 4.6 (except this, Msal net45 will be used)
Mono 4.6
XamariniOS 10 (and this will use explicit iOS target)
Xamarin Mac 3
Xamarin Android 7
UAP 10

This has been approved by Bogdan but just mentioning for completeness.

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
trwalke: Lets add this in a separate PR. Lots of stuff in here already. Plus I have some questions

src/client/Microsoft.Identity.Client/Internal/Logger/LegacyIdentityLoggerAdapter.cs
trwalke: Minimum is always instead of error since it is the lowest one. Also, this value will always be set by the user. This default is really just for testing.

src/client/Microsoft.Identity.Client/OAuth2/Throttling/RetryAfterProvider.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http.Headers;
+using Microsoft.Identity.Client.Core;
```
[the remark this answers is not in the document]
trwalke: Some methods moved to a new namespace. these are valid.

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerAdapterHelper.cs
trwalke: technically the nullLogger is the default logger. the LegacyIdentityLoggerAdapter is only used when the developer sets it.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -273,6 +274,26 @@ public T WithLegacyCacheCompatibility(bool enableLegacyCacheCompatibility = true
             return (T)this;
         }
 
+        /// <summary>
+        /// Sets the Identity Logger
```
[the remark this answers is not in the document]
trwalke: Yeah, that make sense to do.
@bgavrilMS FYI

src/client/Microsoft.Identity.Client/Internal/Logger/ILoggerAdapter.cs
```diff
@@ -1,41 +1,25 @@
...
 using Microsoft.Identity.Client.Internal.Logger;
+using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.Identity.Client.Core
 {
-    internal interface ICoreLogger
+    internal interface ILoggerAdapter
     {
+        bool PiiLoggingEnabled { get; }
+        bool IsDefaultPlatformLoggingEnabled { get; }
```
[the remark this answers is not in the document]
trwalke: This is part of the current implementation

src/client/Microsoft.Identity.Client/Internal/Logger/ILoggerAdapter.cs
```diff
@@ -1,41 +1,25 @@
...
-using System;
-using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using Microsoft.Identity.Client.Internal.Logger;
+using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.Identity.Client.Core
 {
-    internal interface ICoreLogger
+    internal interface ILoggerAdapter
```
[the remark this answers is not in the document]
trwalke: Well, the `LoggerAdapters` aren't actually loggers. They are only here to adapt to the new logging method. Also, the new logger that the user implements is the `IdentityLogger` so it will be confusing calling the `LoggerAdapter` `IdentityLogger` while also calling the passes in logger the `IdentityLogger`. 

I think `LoggerAdapterHelper` to `LoggerHelper` makes sense though

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public EventLogLevel GetEventLogLevel(LogLevel logLevel)
+        {
+            return (EventLogLevel)((int)logLevel + 1);
```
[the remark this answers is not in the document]
trwalke: Not sure. Might be a good thing to test with the performance tests though

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,98 @@
...
+                config?.EnablePiiLogging ?? false);
+        }
+
+        public void Log(LogLevel logLevel, string messageWithPii, string messageScrubbed)
+        {
+            LogEntry entry = Log(this, logLevel, messageWithPii, messageScrubbed);
+            _identityLogger.Log(entry);
+        }
+
+        public LogEntry Log(ILoggerAdapter logger, LogLevel logLevel, string messageWithPii, string messageScrubbed)
```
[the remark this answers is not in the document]
trwalke: Yesh, this is redundant after refactoring. removing

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,98 @@
...
+                correlationId,
+                config?.ClientName ?? string.Empty,
+                config?.ClientVersion ?? string.Empty,
+                config?.EnablePiiLogging ?? false);
+        }
+
+        public void Log(LogLevel logLevel, string messageWithPii, string messageScrubbed)
+        {
+            LogEntry entry = Log(this, logLevel, messageWithPii, messageScrubbed);
+            _identityLogger.Log(entry);
```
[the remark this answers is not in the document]
trwalke: Actually, I think we should do nothing if it is not enabled. Ill double check.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -141,6 +140,7 @@
     <Compile Remove="$(PathToMsalSources)\Platforms\**\*.*;$(PathToMsalSources)\Resources\*.cs" />
     <Compile Remove="$(PathToMsalSources)\PlatformsCommon\PlatformNotSupported\ApiConfig\SystemWebViewOptions.cs" />
     <Compile Remove="GlobalSuppressions3.cs" />
+    <Compile Include="Internal\Logger\MsalCacheLoggerWrapper.cs" />
```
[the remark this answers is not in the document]
trwalke: yes

src/client/Microsoft.Identity.Client/json/Serialization/DefaultContractResolver.cs
```diff
@@ -1379,7 +1379,7 @@ protected virtual IValueProvider CreateMemberValueProvider(MemberInfo member)
             // warning - this method use to cause errors with Intellitrace. Retest in VS Ultimate after changes
             IValueProvider valueProvider;
             // UNITY flag based on https://github.com/jilleJr/Newtonsoft.Json-for-Unity
-#if !(PORTABLE40 || PORTABLE || DOTNET || ANDROID || iOS || MAC || UNITY)
+#if !(PORTABLE40 || PORTABLE || DOTNET || ANDROID || iOS || MAC || UNITY || UAP10_0_17763)
```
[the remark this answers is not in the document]
trwalke: Yes, Updating to NetStandard 2 and Uap 10.0.17763 caused some weird errors that needed to be fixed with things like this.
Mostly because the version of the JSON serializer is old and was created during before NetStandard 2

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: Yes.
@bgavrilMS FYI

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -279,7 +281,7 @@ public T WithLegacyCacheCompatibility(bool enableLegacyCacheCompatibility = true
             return (T)this;
         }
 
-#if !XAMARINMAC2_0
+#if !XAMARINMAC2_0 !ANDROID && !iOS
```
bgavrilMS, line 284: This is going to be hell to maintain :(. Not a fan.

bgavrilMS: What is the error you are getting from the automated tests? Can we find a way to fix that instead?

pmaytak (approved):
👍 
Should probably let R9/Karel know? This was one of their asks.

build/template-build-and-prep-automation.yaml
```diff
@@ -25,16 +25,21 @@ steps:
...
     SourceFolder: '$(Agent.TempDirectory)'
-    Contents: 'test-cloud.exe'
+    Contents: 'test-cloud.exe2'
     TargetFolder: '$(build.artifactstagingdirectory)'
     flattenFolders: true
 
+- task: PowerShell@2
+  inputs:
+    targetType: 'inline'
+    script: 'Rename-Item -Path $(build.artifactstagingdirectory)\test-cloud.exe2 -NewName "test-cloud.exe"'
```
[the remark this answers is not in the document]
trwalke: This is only to test updates to the pipeline without breaking the pipeline on the main branch, Before I merge I will revert.

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -142,5 +171,16 @@ public sealed partial class TokenCacheNotificationArgs
         /// access tokens are refreshable.
         /// </summary> 
         public DateTimeOffset? SuggestedCacheExpiry { get; private set; }
+
```
trwalke, line 174: new api exposed to developers during cache serialization.

src/client/Microsoft.Identity.Client/Internal/Logger/IdentityLoggerAdapter.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using System.Diagnostics.Tracing;
+using System.Globalization;
+using System.Runtime.CompilerServices;
+using Microsoft.Identity.Client.Core;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.Internal.Logger
+{
+#if !XAMARINMAC20
```
bgavrilMS, line 13: I dislike this very much. Please create a work item to explore getting rid of XamarinMac20

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -93,8 +96,7 @@
...
...ER;HAVE_BINARY_SERIALIZATION;HAVE_BINARY_EXCEPTION_SERIALIZATION;HAVE_CHAR_TO_LOWER_WITH_CULTURE;HAVE_CHAR_TO_STRING_WITH_CULTURE;HAVE_COM_ATTRIBUTES;HAVE_COMPONENT_MODEL;HAVE_CONCURRENT_COLLECTIONS;HAVE_COVARIANT_GENERICS;HAVE_DATA_CONTRACTS;HAVE_DATE_TIME_OFFSET;HAVE_DB_NULL_TYPE_CODE;HAVE_DYNAMIC;HAVE_EMPTY_TYPES;HAVE_ENTITY_FRAMEWORK;HAVE_EXPRESSIONS;HAVE_FAST_REVERSE;HAVE_FSHARP_TYPES;HAVE_FULL_REFLECTION;HAVE_GUID_TRY_PARSE;HAVE_HASH_SET;HAVE_ICLONEABLE;HAVE_ICONVERTIBLE;HAVE_IGNORE_DATA_MEMBER_ATTRIBUTE;HAVE_INOTIFY_COLLECTION_CHANGED;HAVE_INOTIFY_PROPERTY_CHANGING;HAVE_ISET;HAVE_LINQ;HAVE_MEMORY_BARRIER;HAVE_METHOD_IMPL_ATTRIBUTE;HAVE_NON_SERIALIZED_ATTRIBUTE;HAVE_READ_ONLY_COLLECTIONS;HAVE_SECURITY_SAFE_CRITICAL_ATTRIBUTE;HAVE_SERIALIZATION_BINDER_BIND_TO_NAME;HAVE_STREAM_READER_WRITER_CLOSE;HAVE_STRING_JOIN_WITH_ENUMERABLE;HAVE_TIME_SPAN_PARSE_WITH_CULTURE;HAVE_TIME_SPAN_TO_STRING_WITH_CULTURE;HAVE_TIME_ZONE_INFO;HAVE_TRACE_WRITER;HAVE_TYPE_DESCRIPTOR;HAVE_UNICODE_SURROGATE_DETECTION;HAVE_VARIANT_TYPE_PARAMETERS;HAVE_VERSION_TRY_PARSE;HAVE_XLINQ;HAVE_XML_DOCUMENT;HAVE_XML_DOCUMENT_TYPE;HAVE_CONCURRENT_DICTIONARY;HAVE_REGEX_TIMEOUTS;$(AdditionalConstants)</DefineConstants>
```
bgavrilMS, line 99: This will require some smoke testing of UWP scenarios.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -292,4 +294,8 @@
       <DependentUpon>Splash.resx</DependentUpon>
     </None>
   </ItemGroup>
+
+  <ItemGroup>
```
bgavrilMS, line 298: Does this get referenced even on xamarinmac ?

src/client/Microsoft.Identity.Client/Platforms/Android/SystemWebview/AuthenticationActivity.cs
```diff
@@ -11,8 +11,10 @@
 using Android.OS;
 #if __ANDROID_29__
 using AndroidX.Browser.CustomTabs;
+using Microsoft.Identity.Client.Core;
```
bgavrilMS, line 14: no need for 2 of these.

bgavrilMS (approved):
Small comments.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: yes

PR #3250, opened 2022-03-26
Add perf tests for Acquire Token calls with serialization cache

pmaytak:
Fixes # 

**Testing**

**Performance impact**


bgavrilMS:
For S2S: 

I don't know if it's realistic to have 1000 tokens per partition. Maybe we should keep the number of resources at 10, and vary the number of tenants: 1k, 10k, 100k, 1M ? The (1M, 10) would be interesting to see. 

For OBO: same comment.

tests/Microsoft.Identity.Test.Performance/Helpers/InMemoryCache.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Concurrent;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Test.Performance.Helpers
+{
+    public class InMemoryCache
+    {
+        private readonly ConcurrentDictionary<string, byte[]> _memoryCache = new ConcurrentDictionary<string, byte[]>();
```
bgavrilMS, line 11: It would be better to use the MemoryCache from Id.Web here, maybe with an eviction policy.

tests/Microsoft.Identity.Test.Performance/Helpers/InMemoryCache.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Concurrent;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Test.Performance.Helpers
+{
+    public class InMemoryCache
```
bgavrilMS, line 9: It would be better to use the MemoryCache from Id.Web here, maybe with an eviction policy.

bgavrilMS (changes requested):
1. We should use the memory cache from Id.Web for comparison, as that is the one we recomend. The Id.Web layer does extra work, such as setting policies etc. 
2. Not sure if a lower number of partitions is relevant. We should use a high number of partitions and a lower number of tokens per partition. These are the real life scenarios IMO.

pmaytak:
> 1. We should use the memory cache from Id.Web for comparison, as that is the one we recomend. The Id.Web layer does extra work, such as setting policies etc.
> 2. Not sure if a lower number of partitions is relevant. We should use a high number of partitions and a lower number of tokens per partition. These are the real life scenarios IMO.

- Added results above for Identity Web cache. Compared to in-memory cache, Id.Web cache is usually a bit slower, but not significantly.
- Added results for test cases with multiple partitions and few tokens per partition.
- Added test case of 1M partitions, but there were issues running the tests, probably due to memory constraints. As expected though, the perf is similar to other tests - number of partitions don't affect perf too much.
- Yes, high number of tokens per partition is probably not common / realistic, but I think it can give good perspective on perf of internal methods, like filtering.

bgavrilMS (approved):

PR #3255, opened 2022-03-30
Bogavril/3251

### LINKED ISSUE #3251: [Bug] IsSystemWebViewAvailable doesn't work correctly in WSL2
**Logs and network traces**
Not applicable

MSAL.NET 4.39.0

**Platform**
.NET 6, WSL2 Debian

Not applicable. The auth flow works correctly but they cannot be invoked correctly.

* Desktop / Mobile
    * [ ] Interactive
    * [ ] Username Password
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 

Other?

The app is in production, I haven't upgraded MSAL, but started seeing this issue.
The flow works correctly before but somehow doesn't work anymore but I'm not sure when.

**Repro**

```csharp
            try
            {
                // The IsSystemWebViewAvailable doesn't seem to work correctly on WSL
                if (!azureApp.IsSystemWebViewAvailable()) // azureApp is a PublicClientApplicationBuilder.Build() object.
                {
                    // Device code flow to get openid only, if client has no web view.
                    result = azureApp.AcquireTokenWithDeviceCode(
                            new[] { "openid" },
                            codeResult =>
                            {
                                console.Output.WriteLine(codeResult.Message);
                                return Task.CompletedTask;
                            })
                        .ExecuteAsync(cancellationToken).GetAwaiter().GetResult();
                }
                else
                {
                    // Common authorization code flow within browser.
                    result = azureApp
                        .AcquireTokenInteractive(new[] { "openid" })
                        .WithPrompt(Prompt.SelectAccount)
                        .ExecuteAsync(cancellationToken).GetAwaiter().GetResult();
                }
            }
            catch (MsalException ex)
            {
                throw new CommandException($"Unable to authenticate to Azure AD: {ex.Message}");
            }

```

**Expected behavior**

The expected behavior is that if a user is in an environment where UI/web browser is available, it should go to interactive login page otherwise device code login should be use.

**Actual behavior**

The actual behavior is that the `IsSystemWebViewAvailable` always returns true and it tries to open a browser in WSL shell which never succeeded. I attempted `IsUserInteractive` and that doesn't work either.

One thing to notice is that it works on a real Linux server without UI (it did jump to device code flow).

**Possible solution**
No idea how this works under the hood. I can imagine platform-dependent code.

bgavrilMS:
Fixes #3251 
Add wslview which works on WSL

Manually tested.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
@@ -189,7 +189,8 @@ namespace Microsoft.Identity.Client.Platforms.netcore
                         }
                     }
                     
-                    if (opened == false)
+                    if (!opened)
+
                     {
                         throw new MsalClientException(
                             MsalError.LinuxXdgOpen,
--- a/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
@@ -170,26 +170,30 @@ namespace Microsoft.Identity.Client.Platforms.netcore
                 string sudoUser = Environment.GetEnvironmentVariable("SUDO_USER");
                 if (!string.IsNullOrWhiteSpace(sudoUser))
                 {
-                    throw new MsalClientException(MsalError.LinuxXdgOpen, MsalErrorMessage.LinuxOpenAsSudoNotSupported);
+                    throw new MsalClientException(
+                        MsalError.LinuxXdgOpen, 
+                        MsalErrorMessage.LinuxOpenAsSudoNotSupported);
                 }
 
                 try
                 {
-                    ProcessStartInfo psi = null;
+                    bool opened = false;
 
                     foreach (string openTool in GetOpenToolsLinux(isBrokerConfigured))
                     {
                         if (TryGetExecutablePath(openTool, out string openToolPath))
                         {
-                            psi = OpenLinuxBrowser(openToolPath, url);
-
+                            OpenLinuxBrowser(openToolPath, url);
+                            opened = true;
                             break;
                         }
                     }
                     
-                    if (psi == null)
+                    if (opened == false)
                     {
-                        throw new Exception("Failed to locate a utility to launch the default web browser.");
+                        throw new MsalClientException(
+                            MsalError.LinuxXdgOpen,
+                            MsalErrorMessage.LinuxOpenToolFailed);
                     }
                 }
                 catch (Exception ex)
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -1,218 +1,222 @@
...
                         if (TryGetExecutablePath(openTool, out string openToolPath))
                         {
-                            psi = OpenLinuxBrowser(openToolPath, url);
-
+                            OpenLinuxBrowser(openToolPath, url);
+                            opened = true;
                             break;
                         }
                     }
-
```
bgavrilMS, line 189: whitespace was wrong here, probably because the file was worked on from Linux some time back. Only this and another block has changes.

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -222,52 +226,51 @@ private ProcessStartInfo OpenLinuxBrowser(string openToolPath, string url)
...
-        }
-
+        }
+
         private string[] GetOpenToolsLinux(bool isBrokerConfigured)
         {
             if (isBrokerConfigured)
             {
-                return new[] { "microsoft-edge", "xdg-open", "gnome-open", "kfmclient" };
+                return new[] { "microsoft-edge", "xdg-open", "gnome-open", "kfmclient", "wslview" };
```
bgavrilMS, line 235: Added wslview tool here.

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -1,218 +1,222 @@
...
+                {
+                    bool opened = false;
+
                     foreach (string openTool in GetOpenToolsLinux(isBrokerConfigured))
                     {
                         if (TryGetExecutablePath(openTool, out string openToolPath))
                         {
-                            psi = OpenLinuxBrowser(openToolPath, url);
-
+                            OpenLinuxBrowser(openToolPath, url);
```
trwalke, line 186: So if this fails it will throw n exception of some kind?

trwalke (approved):

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -1,218 +1,222 @@
...
-            {
-                throw new PlatformNotSupportedException(RuntimeInformation.OSDescription);
-            }
-            return Task.FromResult(0);
-
-        }
-
-        private ProcessStartInfo OpenLinuxBrowser(string openToolPath, string url)
+                    
+                    if (opened == false)
```
pmaytak, line 192: nit:
```suggestion
                    if (!opened)

```

pmaytak (approved):

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
bgavrilMS: We were simply not using this at all.

PR #3258, opened 2022-03-31
Internal log improvements

trwalke:
Fixes # 
Internal loggin updates

Using Request logger in more places as opposed to the internal logger.


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -173,9 +173,10 @@ public void DeleteAccount(MsalAccountCacheItem item)
         /// WARNING: if partitonKey = null, this API is slow as it loads all tokens, not just from 1 partition. 
         /// It should only support external token caching, in the hope that the external token cache is partitioned.
         /// </summary>
-        public virtual List<MsalAccessTokenCacheItem> GetAllAccessTokens(string partitionKey = null)
+        public virtual IReadOnlyList<MsalAccessTokenCacheItem> GetAllAccessTokens(string partitionKey = null, ICoreLogger requestlogger = null)
```
bgavrilMS, line 176: Merge conflict probably resolved wrongly. I think Peter moved all these to `List`

bgavrilMS (approved):
LGTM, but you have 1 merge conflict - please resolve it before checking in.

pmaytak (approved):
What's the difference between app and request loggers? Would be helpful to just use app level one, so we don't have to pass request logger into every method where we want to log. Maybe this can be investigated as part of the bigger logger refactoring?

bgavrilMS:
> What's the difference between app and request loggers? Would be helpful to just use app level one, so we don't have to pass request logger into every method where we want to log. Maybe this can be investigated as part of the bigger logger refactoring?

Most of our logging is done via a request logger, which has the correlation ID. But some logging is done outside the scope of a request (e.g. when you construct the application object). The TokenCache is at APP level, most of the methods are used in the context of a request.

bgavrilMS:
Let's merge this?

Issue #3265, opened 2022-04-05
[Feature Request] no instance discovery metadata for PPE endpoint.


Qi-Zhang-MSFT:
**Is your feature request related to a problem? Please describe.**
There is no instance discovery metadata for PPE endpoint.

**Describe the solution you'd like**
There is no instance discovery metadata for login.windows-ppe.net, which is the PPE endpoint.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs#L40

Also, I see that when we set the authority to login.windows-ppe.net, the instance discovery called made by MSAL looks like:

https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https:%2F%2Flogin.windows-ppe.net%2F0d035206-f7e3-4ad9-b326-1cd88c6c24d6%2Foauth2%2Fv2.0%2Fauthorize

which is weird given login.microsoftonline.com is the prod endpoint, and the response does not contain login.windows-ppe.net.

**Describe alternatives you've considered**

**Additional context**

bgavrilMS:
ESTS-R team are getting a lot of support cases because of this.

SlyryD:
I can't sign in with an MSA PPE account. Should https://login.live-int.com/oauth20_authorize.srf be added as well?

bgavrilMS:
Can you login with an MSA PPE account by using AAD's PPE authority ? https://login.windows-ppe.net/common ?

SlyryD:
It might be a WAM integration issue, but it doesn't seem like it. This is an MSA PPE account I can use to sign into word.officeppe.com.
![image](https://user-images.githubusercontent.com/4186968/191084444-dc90ad71-5368-4574-a8cf-2b8cc3920cc2.png)

bgavrilMS:
@SlyryD - can you please send an email to me (bogavril) and I will forward it to some people who can help.

PR #3273, opened 2022-04-13
Adding PPE to known discovery endpoints

trwalke:
Fixes # 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3265

Adding PPE to known instance discovery metadata

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
@@ -72,11 +72,19 @@ namespace Microsoft.Identity.Client.Instance.Discovery
                 PreferredCache = "login-us.microsoftonline.com"
             };
 
+            InstanceDiscoveryMetadataEntry ppeCloudEntry = new InstanceDiscoveryMetadataEntry()
+            {
+                Aliases = new[] { "login.windows-ppe.net" },
+                PreferredNetwork = "login.windows-ppe.net",
+                PreferredCache = "login.windows-ppe.net"
+            };
+
             AddToKnownCache(publicCloudEntry);
             AddToKnownCache(cloudEntryChina);
             AddToKnownCache(cloudEntryGermanay);
             AddToKnownCache(usGovCloudEntry);
             AddToKnownCache(usCloudEntry);
+            AddToKnownCache(ppeCloudEntry);
             AddToPublicEnvironment(publicCloudEntry);
         }
 
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
 
             InstanceDiscoveryMetadataEntry ppeCloudEntry = new InstanceDiscoveryMetadataEntry()
             {
-                Aliases = new[] { "login.windows-ppe.net" },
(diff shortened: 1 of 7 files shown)


src/client/Microsoft.Identity.Client/Instance/Discovery/KnownMetadataProvider.cs
```diff
@@ -72,11 +72,19 @@ void AddToPublicEnvironment(InstanceDiscoveryMetadataEntry entry)
                 PreferredCache = "login-us.microsoftonline.com"
             };
 
+            InstanceDiscoveryMetadataEntry ppeCloudEntry = new InstanceDiscoveryMetadataEntry()
```
trwalke, line 75: The solution seems to be to simply add this entry to the known list. Seems like something we should have as it is well known for our first party customers? Thoughts @bgavrilMS ?

bgavrilMS: Yes, this is correct.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/InstanceProviderTests.cs
```diff
@@ -66,6 +66,10 @@ public void KnownMetadataProvider_RespondsIfEnvironmentsAreKnown()
                 "login.partner.microsoftonline.cn", new[] { "login.windows.net", "login.microsoft.com", "login.partner.microsoftonline.cn" }, _logger);
             Assert.IsNotNull(result);
 
+            result = knownMetadataProvider.GetMetadata(
```
bgavrilMS, line 69: Can we have a test around regional that confirms that when using PPE, an instance discovery call is not made ?

bgavrilMS: You'll also need to fix the test `OtherCloudWithAuthorityValidationAsync`, as we still want to test with a "different cloud".

trwalke: > Can we have a test around regional that confirms that when using PPE, an instance discovery call is not made ?

Does this test already validate this?

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5e29d349dde99775bf60f23812e2a2437bb0cca9/tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs#L398

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
```diff
@@ -340,7 +340,7 @@ public async Task OtherCloudWithAuthorityValidationAsync()
 
                 var tokenHttpCallHandler = new MockHttpMessageHandler()
                 {
-                    ExpectedUrl = "https://eastus.login.windows-ppe.net/17b189bc-2b81-4ec5-aa51-3e628cbc931b/oauth2/v2.0/token",
+                    ExpectedUrl = "https://eastus.login.windows-ppe.org/17b189bc-2b81-4ec5-aa51-3e628cbc931b/oauth2/v2.0/token",
```
bgavrilMS, line 343: That's quite subtle ... at least add a comment stating that `login.windows-ppe.org` is not known to MSAL.

bgavrilMS (approved):

pmaytak (approved):

gladjohn (approved):

Issue #3270, opened 2022-04-12
[Bug] WebView2 does not certify that device is managed


bgavrilMS:
AcquireTokenInteractive and login with a user from a tenant where CA policy is enabled fails if using WebView2.

Repro

1. Create console app and set tfm to net5.0-windows10.0.177763.0
2. Add the following code (this is Az SDK client id, Microsoft tenant and storage scope)
3. Login interactively with @microsoft account from a domain joined device

```csharp
 var pca = PublicClientApplicationBuilder.Create("04b07795-8ddb-461a-bbee-02f9e1bf7b46")                
                .WithTenantId("72f988bf-86f1-41af-91ab-2d7cd011db47")
                .WithDefaultRedirectUri()
                .Build();

           var result = await pca.AcquireTokenInteractive(new[] { "https://storage.azure.com/.default" })
               .WithUseEmbeddedWebView(true)
               .ExecuteAsync().ConfigureAwait(false);

```

**Actual**: 

error_description=AADSTS50155: Device authentication failed.
Trace ID: d54b18e2-4c07-41fe-9c02-3c0d0fb38400
Correlation ID: ed9eda78-a6fe-46ca-a90b-243368af7ddd
Timestamp: 2022-04-12 10:07:08Z

**Expected**: a token

**Note**: webview1 (which is the default on .NET classic) is not affected by this, nor is system browser. The recommendation is to use WAM on Windows.

bgavrilMS:
Note: upgrading to latest SDK does not fix this problem. Proposing to drop WV2.

bgavrilMS:
See PR https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3276 with a partial fix.

bgavrilMS:
Reopening this as P2.

PR #3276, opened 2022-04-14
Default to WebView1 instead of WebView2 when using AAD or ADFS author…

bgavrilMS:
…ities

Partial fix for #3270

- if authority is AAD or ADFS, use WebView1
- if authority is B2C - use previous logic (use WebView2 if available)

**Testing**
unit 
dev app which reproduces the problem

**Performance impact**
interactive flow is not perf critical


gladjohn (approved):
Looks good

pmaytak (approved):

PR #3289, opened 2022-04-21
Fix NPE in IsBrokerAvailable and update method comment.

### LINKED ISSUE #3261: [Bug] IsBrokerAvailable() has NullReferenceException
**Logs and network traces**
```
   at Microsoft.Identity.Client.PublicClientApplicationBuilder.IsBrokerAvailable()
   at Program.<Main>$(String[] args) in W:\DEV\TestConsoleAuthenticationBroker\TestConsoleAuthenticationBroker\Program.cs:line 10
```

4.42.1

**Platform**
.NET 6, executed on Windows 10 21H2

Public Client with Broker

New app

**Repro**

```csharp
using Microsoft.Identity.Client;

var options = new PublicClientApplicationOptions();
options.AadAuthorityAudience = AadAuthorityAudience.AzureAdMultipleOrgs;
options.ClientId = "F3120210-46AA-4D3B-B7EA-72E69AEC0B5C";
options.AzureCloudInstance = AzureCloudInstance.AzurePublic;
options.RedirectUri = "http://localhost";

var builder = PublicClientApplicationBuilder.CreateWithApplicationOptions(options);
if (builder.IsBrokerAvailable())
{
    builder = builder
        .WithBroker()
        .WithWindowsBrokerOptions(new WindowsBrokerOptions
        {
                ListWindowsWorkAndSchoolAccounts = true,
                MsaPassthrough = true,
                HeaderText = "TEST"
        });
}

builder.Build();

Console.ReadLine();
```

**Expected behavior**
Working code :)

**Actual behavior**
A NullReferenceException is thrown upon the IsBrokerAvailable() call

pmaytak:
Fixes #3261

**Root cause:**
NPE happens if authority is not set in the config before calling this method. Resolving authority happens later when `Build` is called.

- In the method setting authority to AAD by default, if it's null (same as in general)
- Added more details to the method comment.
- Updated TrueMAM samples with a check for broker availability. 
- Did not obsolete since the method can still be used in the TrueMAM apps.
- Created #3303 to investigate if TrueMAM scenario can be improved.

**Testing**
Manual + existing.

**Performance impact**
N/A


gladjohn (approved):
looks good, but few tests are failing after marking IsBrokerAvailable as obsolete

pmaytak:
@SameerK-MSFT related to TrueMAM:
- What is the current TrueMAM behavior if no brokers are installed? (If I remember, it falls back to browser, but browser shows an error, since the request can't be complete?)
- Is there a way where we can check if broker is installed ourselves and throw an exception with a link to install it? (I think we did this before. There's also an internal method, something like IsBrokerInvokable.)
- If the above is not possible, we could keep this method as is, but would have to update the samples/wiki to show that broker availability needs to be checked manually.

SameerK-MSFT:
> What is the current TrueMAM behavior if no brokers are installed? (If I remember, it falls back to browser, but browser shows an error, since the request can't be complete?)

You are right that it will fallback to the browser. But the request cannot be completed.

> Is there a way where we can check if broker is installed ourselves and throw an exception with a link to install it? (I think we did this before. There's also an internal method, something like IsBrokerInvokable.)
> If the above is not possible, we could keep this method as is, but would have to update the samples/wiki to show that broker availability needs to be checked manually.

PCA has a method " public bool IsBrokerAvailable()". That can be used. But please test to ensure that it serves the purpose before making it obsolete.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -334,13 +334,14 @@ public PublicClientApplicationBuilder WithKerberosTicketClaim(string servicePrin
...
-        /// 
         /// On mobile, the device must be Intune joined and Authenticator or Company Portal must be installed. See https://aka.ms/msal-brokers
         /// </remarks>
+        [Obsolete("Use WithBroker by itself, which will fall back to using a web view if broker is unavailable.", false)]
         public bool IsBrokerAvailable()
-        {            
-            return PlatformProxyFactory.CreatePlatformProxy(null)
-                .CreateBroker(base.Config, null).IsBrokerInstalledAndInvokable(base.Config.Authority.AuthorityInfo.AuthorityType);
+        {
+            return Config.Authority?.AuthorityInfo?.AuthorityType == null ? false :
```
bgavrilMS, line 342: How about we assume the authority type is AAD instead if it is not already configured?

bgavrilMS (approved):

bgavrilMS (approved):

gladjohn (approved):
looks good

PR #3295, opened 2022-04-28
Discovery of region is done only once

SameerK-MSFT:
Fixes # 3277

There is semaphore to perform the discovery operation only once.

**Testing**
Validated unit tests. The actual may be harder to test due to dependency on IMDS.

**Performance impact**
Semaphore may affect the performance one time.


src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -176,67 +171,103 @@ private async Task<RegionInfo> DiscoverAndCacheAsync(string azureRegionConfig, I
...
                 {
-                    { "Metadata", "true" }
-                };
+                    var headers = new Dictionary<string, string>
+                    {
+                        { "Metadata", "true" }
+                    };
 
-                Uri imdsUri = BuildImdsUri(DefaultApiVersion);
+                    Uri imdsUri = BuildImdsUri(DefaultApiVersion);
```
bgavrilMS, line 191: spacing seems strange, makes it difficult to review.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -38,6 +38,9 @@ public RegionInfo(string region, RegionAutodetectionSource regionSource, string
         private readonly IHttpManager _httpManager;
         private readonly int _imdsCallTimeoutMs;
 
+        // lock for discovery. So it is done only once
+        private readonly SemaphoreSlim _lockDiscover = new SemaphoreSlim(1);
```
bgavrilMS, line 42: We create 1 RegionManager per Application object, so 2 different apps would have 2 different locks. Notice how we use statics for autoDiscoveredRegion etc. Consider making this lock static as well.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
```diff
@@ -75,7 +75,7 @@ public async Task TelemetryAcceptanceTestAsync()
 
             // we can assert telemetry here, as it will be sent to AAD. However, AAD is down, so it will not record it.
             AssertCurrentTelemetry(result.HttpRequest, ApiIds.AcquireTokenForClient,
-                RegionAutodetectionSource.Cache.ToString("D"),
+                RegionAutodetectionSource.EnvVariable.ToString("D"),
```
bgavrilMS, line 78: We should not change existing behavior. If region auto-detection completed, we should reuse that result.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
bgavrilMS, line 42: We need to unit test this change, i.e. 

1. start request1 and make it pause inside the crtical section (for example by adding a pause in the http manager)
2. start request2 and ensure it is blocked waiting
3. complete request1 and assert request2 does not make HTTP request.

bgavrilMS (changes requested):
Needs a unit test

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
SameerK-MSFT: Let us discuss this tomorrow.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -151,18 +154,10 @@ private bool IsTelemetryRecorded(ApiEvent apiEvent)
 
         private async Task<RegionInfo> DiscoverAndCacheAsync(string azureRegionConfig, ICoreLogger logger, CancellationToken requestCancellationToken)
```
pmaytak, line 155: nit: remove `azureRegionConfig` parameter, it's unused.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -151,18 +154,10 @@ private bool IsTelemetryRecorded(ApiEvent apiEvent)
...
-            {
-                logger.Info($"[Region discovery] Auto-discovery already ran and found {s_autoDiscoveredRegion}.");
-                return new RegionInfo(s_autoDiscoveredRegion, RegionAutodetectionSource.Cache, null);
+            var regionInfo = GetCachedRegion(logger);
+            if (regionInfo != null)
+            {
+                return regionInfo;
             }
 
             var result = await DiscoverAsync(logger, requestCancellationToken).ConfigureAwait(false);
```
pmaytak, line 163: These need to be set inside of a lock. Otherwise
- Thread 1 discovers region, releases semaphore.
- Thread 2 calls GetCachedRegion and uses old values, sees that region was not found, starts discovery.
- Thread 1 updates the static region values to the region it found.

```csharp
            s_failedAutoDiscovery = result.RegionSource == RegionAutodetectionSource.FailedAutoDiscovery;
            s_autoDiscoveredRegion = result.Region;
            s_regionDiscoveryDetails = result.RegionDetails;
```

pmaytak (changes requested):
- Some values need to be updated inside the lock.
- A unit test that Bogdan described would be helpful.
- Suggest try to repro the original bug or that at least these changes don't have regressions.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/RegionalTelemetryTests.cs
```diff
@@ -93,6 +98,43 @@ public async Task TelemetryAcceptanceTestAsync()
...
+                                                        {
+                                                            Assert.Fail(ex.Message);
+                                                        }
+                                                    });
+
+            void DelayLogCallback(LogLevel logLevel, string message, bool hasPii)
+            {
+                if (message.Contains("[Region discovery]"))
+                {
+                    Thread.Sleep(1 * 1000);// sleep 1 sec
```
bgavrilMS, line 125: We should try to not pause the tests. Better to use signaling instead.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -267,6 +267,7 @@ private RegionInfo GetCachedRegion(ICoreLogger logger)
                 return new RegionInfo(s_autoDiscoveredRegion, RegionAutodetectionSource.Cache, null);
             }
 
+            logger.Info($"[Region discovery] Auto-discovery successful but found null or empty region.");
```
bgavrilMS, line 270: This is very subtle / unexpected. Removing a logging statemetn will break the test. 

Here's an example of how to mock the IMDS response: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs#L290

This will be a much more realistic test, i.e. when we make an HTTP call to IMDS, the call takes a clong time.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
SameerK-MSFT: Good catch

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
SameerK-MSFT: Unit test added

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
SameerK-MSFT: Added test

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -149,94 +152,130 @@ private bool IsTelemetryRecorded(ApiEvent apiEvent)
...
+            return result;
+        }
 
-            }
-            catch (Exception e)
+        // returns cached region if any.
+        // if nothing is cached, returns null.
+        private RegionInfo GetCachedRegion(ICoreLogger logger)
+        {
+            if (s_failedAutoDiscovery)
```
neha-bhargava, line 263: Does it make sense to add a test around this, when auto discovery has already failed it is not performed again?

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
SameerK-MSFT: It is a private method. So as long as corresponding public method is checked, it is fine.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -85,6 +85,43 @@ public async Task SuccessfulResponseFromLocalImdsAsync()
...
+                    ValidateInstanceMetadata(regionalMetadata);
+                    Interlocked.Decrement(ref threadCount);
+                }
+                catch (Exception ex)
+                {
+                    Assert.Fail(ex.Message);
+                }
+            });
+
+            Assert.IsTrue(result.IsCompleted);
```
pmaytak, line 117: I think this needs to be after `while`?

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -85,6 +85,43 @@ public async Task SuccessfulResponseFromLocalImdsAsync()
...
+                {
+
+                    _testRequestContext.ServiceBundle.Config.AzureRegion =
+                        ConfidentialClientApplication.AttemptRegionDiscovery;
+
+                    InstanceDiscoveryMetadataEntry regionalMetadata = await _regionDiscoveryProvider.GetMetadataAsync(
+                        new Uri("https://login.microsoftonline.com/common/"), _testRequestContext).ConfigureAwait(false);
+
+                    ValidateInstanceMetadata(regionalMetadata);
+                    Interlocked.Decrement(ref threadCount);
```
pmaytak, line 109: I think this needs to be in a `finally` statement.

I removed the semaphores from the RegionManager code and this test hung up. There were exception in all threads because regionalMetadata was null, for some reason. So threadCount was never decremented.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
```diff
@@ -85,6 +85,43 @@ public async Task SuccessfulResponseFromLocalImdsAsync()
             ValidateInstanceMetadata(regionalMetadata);
         }
 
+        [TestMethod]
+        public void MultiThreadSuccessfulResponseFromLocalImds_HasOnlyOneImdsCall()
```
pmaytak, line 89: I removed the semaphore from the RegionManager and this test still passed. So it can't reliably test the related code... Although I'm not sure of a better way to test this, since it involves a specific order and time of threads executing the same code. I did test it manually by debugging through 2 threads, and everything worked as expected though.

pmaytak (approved):
Since this test has threading stuff, it might be fail intermittently like that other pro-active refresh test. Long term we might need to find a way to more reliably test the threading stuff.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
SameerK-MSFT: Done.

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
SameerK-MSFT: Good catch

tests/Microsoft.Identity.Test.Unit/CoreTests/RegionDiscoveryProviderTests.cs
SameerK-MSFT: If we find a better way to test it, I will be more than happy to improve it. But sounds like good for now.

SameerK-MSFT:
Failure in the pipeline is because of the expired iOS cert. Will merge.

PR #3297, opened 2022-04-30
Refactor perf tests and create a YAML perf build

pmaytak:
- Fix broken benchmarks.
- Return benchmark result value (to prevent dead-code elimination).
- Use explicit type in benchmarks (for consistency between runs).
- Remove IterationSetup and Random use where possible, use constant values (for consistency between runs).
- Minor other clean up.
- Review and clean up test cases for cache size test cases, AcquireTokenNoCache, and the constant test params. Make sure test cases are covered completely without overlap.
- Convert classic pipeline to YAML.
   - Set up to run as a CI off main branch only when changes to Microsoft.Identity.Client project are made.
   - Can also be run manually.
- After this PR is merged  - update pipeline to run as CI off main branch.
- Update documentation.


tests/Microsoft.Identity.Test.Performance/TokenCacheTests.cs
```diff
@@ -99,13 +79,13 @@ public async Task<IAccount> GetAccountAsync_TestAsync()
                 .ConfigureAwait(false);
         }
 
-        [Benchmark(Description = "GetAccounts")]
-        [BenchmarkCategory("With cache")]
+        //GetAccounts is only available in PCA and is not high-perf scenario.
+        //[Benchmark(Description = "GetAccounts")]
```
neha-bhargava, line 83: Is this supposed to be commented?

tests/Microsoft.Identity.Test.Performance/TokenCacheTests.cs
```diff
@@ -45,10 +36,8 @@ public class TokenCacheTests
         // This is a workaround to specify the exact param combinations to be used.
         public IEnumerable<(int, int)> CacheSizeSource => new[] {
             (1, 10),
-            (1, 10000),
```
neha-bhargava, line 48: What is the reason to remove some combinations? How are these chosen?

neha-bhargava (approved):
Few questions otherwise looks good.

tests/Microsoft.Identity.Test.Performance/TokenCacheTests.cs
pmaytak: Yea, don't really need to run this.

tests/Microsoft.Identity.Test.Performance/TokenCacheTests.cs
pmaytak: The goal is to have some common, unique scenarios. The perf is essentially the same for cases with any number of tenants. And then also two cases that test with small and larger number of tokens per tenant.

build/pipeline-perf-tests-automation.yaml
```diff
@@ -0,0 +1,46 @@
...
+pr: none
+
+pool:
+  name: MsalNetPerfTest1ESPool
+  demands: msbuild
+
+variables:
+  BuildPlatform: 'any cpu'
+  BuildConfiguration: 'release'
+  MSAL_DESKTOP_ONLY_DEV: 'false' # skips building mobile and legacy targets
```
bgavrilMS, line 19: Good idea.

build/pipeline-perf-tests-automation.yaml
```diff
@@ -0,0 +1,46 @@
...
+      displayName: Build PerformanceTests.sln
+      inputs:
+        solution: PerformanceTests.sln
+        msbuildArchitecture: x64
+        platform: '$(BuildPlatform)'
+        configuration: '$(BuildConfiguration)'
+        msbuildArguments: '/p:RunCodeAnalysis=false /p:MSAL_DESKTOP_ONLY_DEV=$(MSAL_DESKTOP_ONLY_DEV)'
+        clean: true
+
+    - script: '$(Build.SourcesDirectory)\tests\Microsoft.Identity.Test.Performance\bin\$(BuildConfiguration)\$(BuildRuntime)\Microsoft.Identity.Test.Performance.exe -BranchName $(Build.SourceBranch) -BranchCommit $(Build.SourceVersion)'
```
bgavrilMS, line 45: I have a few questions about this, perhaps let's discuss at standup.

- how are the 2 params used?
- how are the results reported / used? Or will this be part of another PR?

bgavrilMS (approved):

build/pipeline-perf-tests-automation.yaml
pmaytak: Currently the results are printed to the console window in the build pipeline and those parameters are not used. But I added them because I was curious if it's possible to pass the branch info into the perf app, and it is. So in the future the branch info can be stored along with the perf run results.

bgavrilMS: Ack, maybe add a comment in the code or in the yaml about this.

PR #3313, opened 2022-05-09
Throw TaskCancelledException when request gets cancelled.

### LINKED ISSUE #3283: [Bug] Operation cancelled exception transformed to service exception
**Logs and network traces**
Without logs or traces, it is unlikely that the team can investigate your issue. Capturing logs and network traces is described in [Logging wiki](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging).

4.42.1

**Platform**
Windows

* Desktop / Mobile
    * [X ] Device code flow (browserless)

Visual studio and storage explorer

**Repro**

Cancel device code flow while it is polling

Issue:
  When canceling device code flow sometimes a MSAL service exception is thrown instead of an operation cancelled exception. 

This is because
Microsoft.Identity.Client\Http\HttpManager.cs
ExecuteWithRetryAsync
~176  transforms a cancel into a MsalServiceException with the text "Request to the endpoint timed out."
A check should be performed before doing that transformation and if the cancellation token passed into the method call was cancelled instead of the webapi timing outthen the operation cancelled exception should be thrown instead.

neha-bhargava:
Fixes #3283 

When auth is cancelled, throw TaskCancelledException instead of converting it to TimeoutException

**Testing**
Updated an existing unit test

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -143,13 +143,14 @@ namespace Microsoft.Identity.Client.Http
             }
             catch (TaskCanceledException exception)
             {
-                logger.Error("The HTTP request failed or it was canceled. " + exception.Message);
-                isRetryable = true;
-
                 if (cancellationToken.IsCancellationRequested)
                 {
-                    throw exception;
+                    logger.Info("The HTTP request was cancelled. ");
+                    throw;
                 }
+
+                logger.Error("The HTTP request failed. " + exception.Message);
+                isRetryable = true;
 
                 timeoutException = exception;
             }
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -148,7 +148,7 @@ namespace Microsoft.Identity.Client.Http
 
                 if (cancellationToken.IsCancellationRequested)
                 {
-                    isRetryable = false;
+                    throw exception;
                 }
 
                 timeoutException = exception;
--- a/tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpManagerTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpManagerTests.cs
@@ -135,7 +135,7 @@ namespace Microsoft.Identity.Test.Unit.CoreTests.HttpTests
                 catch (AggregateException exc)
                 {
                     Assert.IsNotNull(exc);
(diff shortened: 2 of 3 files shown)


pmaytak (approved):
Are there any other places in code where we hide the TaskCancelledException or don't use the CancellationToken that was passed in?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -148,7 +148,7 @@ private HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string
 
                 if (cancellationToken.IsCancellationRequested)
                 {
-                    isRetryable = false;
+                    throw exception;
```
bgavrilMS, line 151: Prefer to use `throw` instead. See https://stackoverflow.com/questions/730250/is-there-a-difference-between-throw-and-throw-ex

Also, consioder adding a `log.Info` for this.

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpManagerTests.cs
```diff
@@ -135,7 +135,7 @@ public void TestSendGetWithCanceledToken()
                 catch (AggregateException exc)
                 {
                     Assert.IsNotNull(exc);
-                    Assert.IsTrue(exc.InnerException is MsalServiceException);
+                    Assert.IsTrue(exc.InnerException is TaskCanceledException);
```
bgavrilMS, line 138: Please make this test `async` and use `await httpManager.SendGetAsync` to avoid this whole AggregateException stuff. And maybe add a comment on the test that this is a regression test for bug xyz.

bgavrilMS (approved):

bgavrilMS:
Approved with comments.

neha-bhargava:
> Are there any other places in code where we hide the TaskCancelledException or don't use the CancellationToken that was passed in?

Good catch there were a few places where the CancellationToken was not passed in. I have added it to the flow. As far as the handling is concerned, this method is called for all the HTTP flows so that is covered. Thanks!

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpManagerTests.cs
neha-bhargava: Done.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: Done

SameerK-MSFT (approved):

PR #3315, opened 2022-05-10
Refactor AuthorityType functionality

SameerK-MSFT:
Fixes # 

Shifting the functionality to Authority and derived classes would be the ideal way to refactor and have O in the code. But due to the tight coupling and functionality spread over various segments, that will require significant efforts. In the interest of time, it has a few changes.

Here is the summary of changes:

- Many callers are using AuthorityInfo.AuthorityType to determine the behavior. The behavior + functionality of a type of Authority has been encapsulated in the AuthorityInfo in properties.
- There is a new class AuthorityInfoHelper. This class extends the functionality of AuthorityInfo. It is intentionally a nested class inside AuhtorityInfo for the following reasons:
  - The callers have access to AuthorityInfo; but not Authority itself. So the functionality needs to be wrapped inside AuthorityInfo itself.
  - AuthoirityInfo should not be bloated. This class provides the extended functionality.
  - It has to be nested to access the private members of the AuthorityInfo
  - Ideally this should be nested class of Authority itself. But since it requires access to the private members of Authority and the callers lack access to Authority, it has to be nested class of AuthorityInfo
- To centralize the extended functionality, some helper static functionality has been shifted to AuthorityInfo. In order to minimize the impact on the callers, it Authority delegates that to AuthorityInfo
 - AuthorityValidatorFactory was a helper class with single method to create. Functionality has been shifted to AuthorityInoHelper.
- Changed the AuthorityType from public to internal. Making it private would require many more changes even in test. So this may be the intermediate solution to increase the O factor.

**Testing**
Existing tests

**Performance impact**
None


src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -87,10 +116,18 @@ internal class AuthorityInfo
 
         public string Host { get; }
         public string CanonicalAuthority { get; }
-        public AuthorityType AuthorityType { get; }
+        internal AuthorityType AuthorityType { get; }
```
pmaytak, line 119: Shouldn't matter since enclosing AuthorityInfo is internal?

pmaytak (commented):
nit: Can you set the dsts branch as base for this review, so the dsts changes are excluded?

Reviewed partially, will finish tomorrow.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/DstsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+                result = await app
+                    .AcquireTokenForClient(TestConstants.s_scope)
+                    .ExecuteAsync(CancellationToken.None)
+                    .ConfigureAwait(false);
+
+                Assert.IsNotNull(result);
+                Assert.IsNotNull(result.AccessToken);
+                Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);    
+                //request to token endpoint
```
neha-bhargava, line 93: This comment can be removed. I added this as reminder to assert it and added the checks above at line 47 but forgot to remove this.

neha-bhargava (dismissed):

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -87,10 +116,18 @@ internal class AuthorityInfo
...
         public string UserRealmUriPrefix { get; }
         public bool ValidateAuthority { get; }
 
+        internal bool IsInstanceDiscoverySupported => (AuthorityType == AuthorityType.Aad);
+
+        internal bool IsClientInfoSupported => (AuthorityType == AuthorityType.Aad || AuthorityType == AuthorityType.Dsts);
```
pmaytak, line 125: Should include B2C.

pmaytak (approved):
Didn't review DSTS code.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
SameerK-MSFT: Now includes B2C in the latest check-in

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
SameerK-MSFT: Just getting closer to O.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -112,12 +116,15 @@ private string[] GetPathSegments(string absolutePath)
 
         public string Host { get; }
         public string CanonicalAuthority { get; }
-        public AuthorityType AuthorityType { get; }
+        internal AuthorityType AuthorityType { get; }
```
bgavrilMS, line 119: Maybe add a comment saying that this property should be removed.

bgavrilMS (approved):

Issue #3284, opened 2022-04-15
[Feature Request] Add AT POP support in the runtime broker


gladjohn:
On a very high level : today we cache all access tokens, but we need to cache only Bearer tokens.

- [x] Add POP support in runtime broker
- [x] Unit testing

PR #3316, opened 2022-05-10
Acquire token via WAM Runtime with POP

trwalke:
Fixes # 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3284

Adding new API to enable POP with new MSAL runtime
New api can also be used by confidential client for AcquireTokenSilent
When pop is enabled, MSAL runtime will parse the Pop token from the Authorization header returned from the native broker

Main changes are in RuntimeBroker.cs and WamAdapters.cs

**Testing**
Unit tests
Manual tests


src/client/Microsoft.Identity.Client.Brokers/RuntimeBroker.cs
```diff
@@ -108,6 +108,16 @@ internal class RuntimeBroker : IBroker
                     _logger.Verbose("[WamBroker] Account information was not provided. Using an account picker.");
                 }
 
+                // if PopAuthenticationConfiguration is set, proof of possesion will be performed via the runtime broker
+                if (authenticationRequestParameters.PopAuthenticationConfiguration != null)
```
bgavrilMS, line 112: There are a bunch of other places like AcquireTokenSilent where POP needs to be set.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -35,6 +36,36 @@ public override Task<AuthenticationResult> ExecuteAsync(CancellationToken cancel
...
         }
 
+        /// <summary>
+        ///  Modifies the token acquisition request so that the acquired token is a Proof of Possession token (PoP), rather than a Bearer token. 
+        ///  PoP tokens are similar to Bearer tokens, but are bound to the HTTP request and to a cryptographic key, which MSAL can manage on Windows.
+        ///  See https://aka.ms/msal-net-pop
+        /// </summary>
+        /// <param name="httpMethod">The HTTP method ("GET", "POST" etc.) method that will be bound to the token. Leave null and the POP token will not be bound to the method.
+        /// Corresponds to the "m" part of the a signed HTTP request.</param>
+        /// <param name="requestUri">The URI to bind the signed HTTP request to.</param>
```
bgavrilMS, line 46: Please add a note that only the host and path parts of the URI will get bound.

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
```diff
@@ -179,5 +179,11 @@ private async Task<MsalTokenResponse> RunBrokerWithInstallUriAsync(string broker
             Metrics.IncrementTotalAccessTokensFromBroker();
             return tokenResponse;
         }
+
+        protected override bool ShouldNotCacheTokenResponse()
```
bgavrilMS, line 183: AFAIK all request support POP. We should not cache pop tokens. So I don't understand why we'd need a virtual method in RequestBase which some requests can override.

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
```diff
@@ -179,5 +179,11 @@ private async Task<MsalTokenResponse> RunBrokerWithInstallUriAsync(string broker
...
+        {
+            return AuthenticationRequestParameters.PopAuthenticationConfiguration != null && //limited to pop tokens for the moment. Will be expanded later
+                ServiceBundle.Config.BrokerCreatorFunc != null; //Check to see if MSAL runtime is used.
```
bgavrilMS, line 186: not a realiable way to check for WAM. But in any case, having a POP config should be enough.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -193,6 +198,38 @@ protected async Task<AuthenticationResult> CacheTokenResponseAndCreateAuthentica
                 msalTokenResponse.SpaAuthCode);
         }
 
+        /// <summary>
+        /// Determines if the token response should be saved in the cached.
+        /// This may be ovveritten by the child types that perform additional checks.
+        /// Token responses recieved from brokers should not be cached for example.
```
bgavrilMS, line 204: I don't think we need a virtual method here. We should simply access tokens WAM irrespective of flow. Today WAM can do InteractiveRequest and SilentRequest, and we also plan to add support for ROPC.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -175,6 +175,11 @@ protected async Task<AuthenticationResult> CacheTokenResponseAndCreateAuthentica
 
             ValidateAccountIdentifiers(fromServer);
 
+            if (ShouldNotCacheTokenResponse())
+            {
+                return GetAuthenticationResultFromTokenResponse(msalTokenResponse);
```
bgavrilMS, line 180: This is not correct. We need to cache Account object - it is very imporant. It should be ok to not cache AT and ID token.

src/client/Microsoft.Identity.Client/TokenResponseHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Instance;
+using Microsoft.Identity.Client.Internal;
+using Microsoft.Identity.Client.Internal.Requests;
+using Microsoft.Identity.Client.OAuth2;
+
+namespace Microsoft.Identity.Client
+{
+    internal class TokenResponseHelper
```
bgavrilMS, line 16: is this all refactoring or did you add any new logic?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke: Now caching id tokens  so that the account object can be created with tenant profiles in the tokenCacheInternal logic

src/client/Microsoft.Identity.Client/TokenResponseHelper.cs
trwalke: all refactoring

src/client/Microsoft.Identity.Client.Brokers/RuntimeBroker.cs
```diff
@@ -153,6 +156,8 @@ internal class RuntimeBroker : IBroker
             using (var core = new NativeInterop.Core())
             using (var authParams = WamAdapters.GetCommonAuthParameters(authenticationRequestParameters, _wamOptions.MsaPassthrough))
             {
+                AddPopParams(authenticationRequestParameters, authParams);
```
bgavrilMS, line 159: Why not add it to `WamAdapters.GetCommonAuthParameters` ?

src/client/Microsoft.Identity.Client.Brokers/RuntimeBroker.cs
```diff
@@ -13,6 +13,7 @@
 using Microsoft.Identity.Client.Instance.Discovery;
 using Microsoft.Identity.Client.Internal.Broker;
 using Microsoft.Identity.Client.Internal.Requests;
+using Microsoft.Identity.Client.NativeInterop;
```
bgavrilMS, line 16: Did you E2E test this manually yet? I see that `WamAdapter.ParseRuntimeResponse` hardcodes "Bearer" token type.

bgavrilMS: You can use `authResult.IsPopAuthorization` to determine if the auth result in Bearer or POP.

src/client/Microsoft.Identity.Client.Brokers/RuntimeBroker.cs
trwalke: will be added in next commit, still refactoring

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -35,6 +36,37 @@ public override Task<AuthenticationResult> ExecuteAsync(CancellationToken cancel
...
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item><description>An Authentication header is automatically added to the request</description></item>
+        /// <item><description>The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters).</description></item>
+        /// <item><description>MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</description></item>
+        /// <item><description>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</description></item>
+        /// </list>
+        /// </remarks>
+        public T WithProofOfPossession(HttpMethod httpMethod, Uri requestUri, string nonce)
+        {
```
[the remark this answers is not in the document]
pmaytak: Wouldn't a resource throw exception anyway since it would expect a POP token, but would get a Bearer?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -147,5 +149,51 @@ public AcquireTokenSilentParameterBuilder WithProofOfPossession(PoPAuthenticatio
...
+        {
+            ClientApplicationBase.GuardMobileFrameworks();
+            PoPAuthenticationConfiguration popConfig = new PoPAuthenticationConfiguration(requestUri ?? throw new ArgumentNullException(nameof(requestUri)));
+            popConfig.HttpMethod = httpMethod ?? throw new ArgumentNullException(nameof(httpMethod));
+            popConfig.Nonce = nonce ?? throw new ArgumentNullException(nameof(nonce));
+
+            CommonParameters.PopAuthenticationConfiguration = popConfig;
+            var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
+
+            if (CommonParameters.PopAuthenticationConfiguration != null)
```
trwalke, line 185: TODO: Throw exception if not on windows

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -35,6 +37,53 @@ public override Task<AuthenticationResult> ExecuteAsync(CancellationToken cancel
...
+            ClientApplicationBase.GuardMobileFrameworks();
+            PoPAuthenticationConfiguration popConfig = new PoPAuthenticationConfiguration(requestUri ?? throw new ArgumentNullException(nameof(requestUri)));
+            popConfig.HttpMethod = httpMethod ?? throw new ArgumentNullException(nameof(httpMethod));
+            popConfig.Nonce = nonce ?? throw new ArgumentNullException(nameof(nonce));
+
+            CommonParameters.PopAuthenticationConfiguration = popConfig;
+            var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
+
+            if (CommonParameters.PopAuthenticationConfiguration != null)
```
trwalke, line 73: TODO: Throw exception if not on windows

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -147,5 +149,53 @@ public AcquireTokenSilentParameterBuilder WithProofOfPossession(PoPAuthenticatio
...
+        /// <item><description>An Authentication header is automatically added to the request</description></item>
+        /// <item><description>The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters).</description></item>
+        /// <item><description>MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</description></item>
+        /// <item><description>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</description></item>
+        /// </list>
+        /// </remarks>
+#if iOS || ANDROID || WINDOWS_UWP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+        public AcquireTokenSilentParameterBuilder WithProofOfPossession(HttpMethod httpMethod, Uri requestUri, string nonce)
```
trwalke, line 175: Need to figure out how to prevent this from running on confidential clients.

The best idea I have now is to throw an exception if this is running on Confidential cleints. We also have our previous POP api which is available on public clients when it should not be. Should probably throw in that situation as well.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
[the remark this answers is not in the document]
bgavrilMS: No, I don't think any resource out there will support just POP. They will support Bearer and POP. However, a CA policy may prevent a Bearer token from gaining access.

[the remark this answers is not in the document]
bgavrilMS: I think this decision point warrants a bigger discussion with MSAL C++ team, the aim is to be consistent with them. 

CC @rayluo and @siddhijain and @sameerag - the context here is "POP is only supported if WAM is present on the box, i.e. Win10. If on Win8, Mac or Linux and app developer configures POP, what should MSAL do? Return a bearer token or throw?"

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -35,6 +37,55 @@ public override Task<AuthenticationResult> ExecuteAsync(CancellationToken cancel
...
```
bgavrilMS, line 73: Can this ever be null? You just set it at line 70

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
bgavrilMS, line 175: Thinking about this some more, only pca,ATI and pca.ATS will support POP. So, I guess this `WithProofOfPossesion` setting needs to be set on ATI and ATS. Next sprint we will try to get it to pca.AcquireTokenByUsernamePassword as well.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -147,5 +149,53 @@ public AcquireTokenSilentParameterBuilder WithProofOfPossession(PoPAuthenticatio
...
+        {
+            ClientApplicationBase.GuardMobileFrameworks();
+            PoPAuthenticationConfiguration popConfig = new PoPAuthenticationConfiguration(requestUri ?? throw new ArgumentNullException(nameof(requestUri)));
+            popConfig.HttpMethod = httpMethod ?? throw new ArgumentNullException(nameof(httpMethod));
+            popConfig.Nonce = nonce ?? throw new ArgumentNullException(nameof(nonce));
+
+            CommonParameters.PopAuthenticationConfiguration = popConfig;
+            var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
+
+            if (CommonParameters.PopAuthenticationConfiguration != null)
```
bgavrilMS, line 185: `CommonParameters.PopAuthenticationConfiguration` cannot be null, you set it at line 70

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -177,6 +181,33 @@ public partial class AuthenticationResult
...
         }
 
+        internal AuthenticationResult(
+            MsalTokenResponse tokenResponse,
+            AuthenticationRequestParameters requestParams,
+            IdToken idToken,
+            IAccount account)
+        {
+            AccessToken = tokenResponse.AccessToken;
+            IsExtendedLifeTimeToken = false; //TODO: How should this be determined?
```
bgavrilMS, line 191: it is a deprecated property and it's fine to always be set to false.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -64,7 +63,7 @@ public sealed partial class TokenCache : ITokenCacheInternal
                 await requestParams.AuthorityManager.GetInstanceDiscoveryEntryAsync().ConfigureAwait(false);
 
             #region Create Cache Objects
-            if (!string.IsNullOrEmpty(response.AccessToken))
+            if (!string.IsNullOrEmpty(response.AccessToken) && !requestParams.ShouldNotCacheAccessTokens)
```
bgavrilMS, line 66: instead of relying on `requestParams` to tell you this (which it shouldn't, because it doesn't deal with caching), better to add the logic here: 

`
bool shouldCache = (msalTokenResponse.TokenSource == Broker && msalTokenResponse.TokenType == POP) 
`

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -193,6 +198,32 @@ protected async Task<AuthenticationResult> CacheTokenResponseAndCreateAuthentica
...
+        {
+            if (account == null)
+            {
+                bool isAdfsAuthority = AuthenticationRequestParameters.AuthorityInfo.AuthorityType == AuthorityType.Adfs;
+                string preferredUsername = TokenResponseHelper.GetPreferredUsernameFromIdToken(isAdfsAuthority, idTokenCacheItem?.IdToken);
+                string username = isAdfsAuthority ? idTokenCacheItem?.IdToken?.Upn : preferredUsername;
+                string homeAccountId = TokenResponseHelper.GetHomeAccountId(AuthenticationRequestParameters, msalTokenResponse, idTokenCacheItem?.IdToken);
+                Dictionary<string, string> wamAccountIds = TokenResponseHelper.GetWamAccountIds(AuthenticationRequestParameters, msalTokenResponse);
+
+                account = new Account(
```
bgavrilMS, line 214: by re-creating the account, you lose out on tenant profiles.

Maybe we don't need a new ctor for `AuthenticationResult`. You already have idTokenCacheItem and account object. Maybe you can just create a simple AccessTokenCacheItem object and use the existing AuthenticationResult ctor?

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
```diff
@@ -320,11 +322,12 @@ private async Task<AuthenticationResult> RunAtiAsync(IPublicClientApplication pc
 
             AuthenticationResult result = null;
             var scopes = GetScopes();
-
+            var guid = Guid.NewGuid();
             var builder = pca.AcquireTokenInteractive(scopes)
                 .WithUseEmbeddedWebView(true)
                 //.WithExtraQueryParameters("domain_hint=live.com") -- will force AAD login with browser
                 //.WithExtraQueryParameters("msafed=0")             -- will force MSA login with browser
+                .WithProofOfPossession(System.Net.Http.HttpMethod.Get, new Uri( pca.Authority), "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6bnVsbH0.eyJ0cyI6MTY1MjI4NTAzNH0.Nh-mAJwRphv57IdpdIrYzmYp6vP_BmmYy4UrNKj5A2x4XKLbp_H3aH4J5_s9hP5MzoiHE2SgVaDG8YUbP4xOjFYmpNG884pWqI-z9RjFNKJgBTXUhwv8HsUnxUHq1KTvpLmd1K1gJZORdeUI2LDr07EEH3-aT0PkRt-wT1YNNh5gU_RHV5KvlsyDWCvCJpEbZmGUf8JX9tHO2ux7XAKD77lVb5m6lFq_8Wr5nhJDyREHrXKWQq-X4rTxnBCZ4KBAufImSVHAeVi7ihlGbcobU2CuyJscTZkyELWMG8rBD6QK57AzrM77mua9-QClKIHArL8_d2fgyksLLS89wxy25A")
```
bgavrilMS, line 330: We need fancy UI this :). I'm joking, but we do need at least a checkbox or dropdown to select between Bearer and POP.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke: I am only creating the account here when the token cache cannot create it and it is null. under normal circumstances, the account will already be created and have the tenant profiles

We have similar logic when using the other constructor https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9c881e80995f0fcd04cb684181d8e9e99482f0ae/src/client/Microsoft.Identity.Client/AuthenticationResult.cs#L147-L157

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
trwalke: Yes, next update will fix

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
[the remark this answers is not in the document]
siddhijain: I think the library should throw an exception to let the user know that their device does not support PoP. If we return a bearer, they will never know that they have not received the right kind of token.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
```diff
@@ -35,6 +37,55 @@ public override Task<AuthenticationResult> ExecuteAsync(CancellationToken cancel
...
+        /// <item><description>An Authentication header is automatically added to the request</description></item>
+        /// <item><description>The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters).</description></item>
+        /// <item><description>MSAL creates, reads and stores a key in memory that will be cycled every 8 hours.</description></item>
+        /// <item><description>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</description></item>
+        /// </list>
+        /// </remarks>
+#if iOS || ANDROID || WINDOWS_UWP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+        public T WithProofOfPossession(HttpMethod httpMethod, Uri requestUri, string nonce)
```
pmaytak, line 63: Should this be in `AcquireTokenInteractiveParameterBuilder` instead? Otherwise currently it's avaiable with other AcquireToken methods like UP, IWA, device code.

pmaytak (changes requested):

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -172,7 +175,25 @@ public bool IsConfidentialClient
 
         public KeyValuePair<string, string>? CcsRoutingHint { get; set; }
 
-        #endregion
+        /// <summary>
+        /// Determines if the token response should be saved in the cached.
+        /// This may be ovveritten by the child types that perform additional checks.
+        /// Token responses recieved from brokers should not be cached for example.
+        /// </summary>
+        public bool ShouldNotCacheAccessTokens
```
bgavrilMS, line 183: Should not be here - ideally the caching logic should know about this.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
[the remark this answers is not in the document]
bgavrilMS: @siddhijain - yes, erroring out soon is a good point, but I am not sure if it will help with the dev experience. The dev experience would be something like: 

```csharp
var pca = new PubliClientApplication().PreferBroker();

if (pca.IsBrokerAvailable)
{
     pca.AcquireTokenInteractive().WithProofOfPosession("http://www.bing.com", nonce);
}
else
{
    pca.AcquireTokenInteractive();
}

```

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
trwalke: We were advised to not cache ATs to ensure the there is no possibility of the AT in MSAL's cache to be different from the broker cache. Also, RTs are not returned from the broker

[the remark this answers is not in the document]
trwalke: In prior commits, I was under the impression that we were not caching anything (AT/RT/Id/Account) so it made more sense to not call into this method but simply create an auth result from the response directly.  But recently I learned we still need to cache the id and the account so might as well reuse what we have.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractPublicClientAcquireTokenParameterBuilder.cs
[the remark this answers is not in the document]
pmaytak: I think the concern is if someone wants to use a more secure PoP token but gets less secure Bearer without knowing. We'd have to suggest for them to monitor the token type. This can be forced by the resource to only accept PoP tokens. If the resource accepts both token types, then having this method fallback to Bearer might not be as bad. Also, we could later add an overload with a boolean parameter whether to fallback or not.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
pmaytak: Yep, I reached the same conclusion - can't think of a way without breaking changes.

src/client/Microsoft.Identity.Client.Brokers/WamAdapters.cs
```diff
@@ -158,9 +160,27 @@ private enum ResponseStatus : int
...
+        /// <summary>
+        /// Configures the Msal Runtime authenticaion request to use proof of posession.
+        /// </summary>
+        private static void AddPopParams(AuthenticationRequestParameters authenticationRequestParameters, NativeInterop.AuthParameters authParams)
+        {
+            // if PopAuthenticationConfiguration is set, proof of possesion will be performed via the runtime broker
+            if (authenticationRequestParameters.PopAuthenticationConfiguration != null)
+            {
+                //_logger.Verbose("[WamBroker] Proof of posession configuration provided. Using proof of posession with broker request.");
+                authParams.PopParams.HttpMethod = authenticationRequestParameters.PopAuthenticationConfiguration.HttpMethod.Method;
```
bgavrilMS, line 177: Can the HttpMethod be null? Possible null ref here.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
```diff
@@ -329,6 +332,53 @@ public AcquireTokenInteractiveParameterBuilder WithParentActivityOrWindow(NSWind
...
+            }
+
+            var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
+
+            if (!broker.IsPopSupported)
+            {
+                throw new MsalClientException(MsalError.BrokerDoesNotSupportPop, MsalErrorMessage.BrokerDoesNotSupportPop);
+            }
+
+            PoPAuthenticationConfiguration popConfig = new PoPAuthenticationConfiguration(requestUri ?? throw new ArgumentNullException(nameof(requestUri)));
```
bgavrilMS, line 373: I believe that all the POP params are optional, definitely the method and uri are optional. 
It's ok to consider the nonce as mandatory but please check with `string.isnullorempty`

src/client/Microsoft.Identity.Client/TokenResponseHelper.cs
bgavrilMS, line 16: static class

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
```diff
@@ -329,6 +332,53 @@ public AcquireTokenInteractiveParameterBuilder WithParentActivityOrWindow(NSWind
...
+            ClientApplicationBase.GuardMobileFrameworks();
+
+            if (!ServiceBundle.Config.IsBrokerEnabled)
+            {
+                throw new MsalClientException(MsalError.BrokerRequiredForPop, MsalErrorMessage.BrokerRequiredForPop);
+            }
+
+            var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
+
+            if (!broker.IsPopSupported)
```
bgavrilMS, line 368: Can you add some unit test coverage for this part? You should be able to inject a mock IBroker

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -147,5 +149,56 @@ public AcquireTokenSilentParameterBuilder WithProofOfPossession(PoPAuthenticatio
...
+        /// </remarks>
+#if iOS || ANDROID || WINDOWS_UWP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+        public AcquireTokenSilentParameterBuilder WithProofOfPossession(HttpMethod httpMethod, Uri requestUri, string nonce)
+        {
+            ClientApplicationBase.GuardMobileFrameworks();
+            var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
+
+            if (ServiceBundle.Config.IsBrokerEnabled)
```
bgavrilMS, line 180: Can this logic be consolidated with the one from AcquireTokenInteractive?

bgavrilMS (approved):
Looks good, added some minor comments and try to write more unit tests. Please work with Gladwin to add some manual E2E tests around this, which we will look to automate next sprint.

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
trwalke: The logic is a bit different between the two. The api for the ATI flow as it is not shared between PCA and CCA while this one is.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -88,7 +91,6 @@ internal class RuntimeBroker : IBroker
                     {
                         msalTokenResponse = WamAdapters.ParseRuntimeResponse(result, authenticationRequestParameters, _logger);
                         _logger.Verbose("[WamBroker] Successfully retrieved token.");
-
                     }
```
SameerK-MSFT, line 94: nit: extra line

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
```diff
@@ -12,6 +12,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\src\client\Microsoft.Identity.Client.Brokers\Microsoft.Identity.Client.Broker.csproj" />
+    <ProjectReference Include="..\..\src\client\Microsoft.Identity.Client.Broker\Microsoft.Identity.Client.Broker.csproj" />
```
SameerK-MSFT, line 16: Somehow the same project ref added twice.

SameerK-MSFT (approved):
Minor changes requested.

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
trwalke: not needed

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ClientApplicationBaseExecutor.cs
```diff
@@ -27,7 +27,21 @@ public ClientApplicationBaseExecutor(IServiceBundle serviceBundle, ClientApplica
             AcquireTokenSilentParameters silentParameters,
             CancellationToken cancellationToken)
         {
-            var requestContext = CreateRequestContextAndLogVersionInfo(commonParameters.CorrelationId, cancellationToken);
+            //Since the AcquireTokenSilentParameterBuilder api is shared between public and confidential clients,
```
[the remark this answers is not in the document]
trwalke: Also added new test to validate behavior

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -128,6 +128,10 @@ public static string iOSBrokerKeySaveFailed(string keyChainResult)
         public const string BrokerResponseError = "Broker response returned error: ";
         public const string CannotInvokeBroker = "MSAL cannot invoke the broker. The Authenticator App (Broker) may not be installed on the user's device or there was an error invoking the broker. " +
             "Check logs for more details and see https://aka.ms/msal-brokers. ";
+        public const string BrokerDoesNotSupportPop = "The broker does not support Proof-of-Possession on the current platform.";
+        public const string BrokerRequiredForPop = "The request has Proof-of-Possession configured but does not have broker enabled. Broker is required for Proof-of-Possession on public clients.";
+        public const string NonceRequiredForPop = "The request has Proof-of-Possession configured for public cleints but does not have a nonce provided. A nonce is required for Proof-of-Possession on public clients.";
```
gladjohn, line 133:
```suggestion
        public const string NonceRequiredForPop = "The request has Proof-of-Possession configured for public clients but does not have a nonce provided. A nonce is required for Proof-of-Possession on public clients.";

```

src/client/Microsoft.Identity.Client/TokenResponseHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+            // otherwise, this will result in cache misses
+            return Authority.CreateAuthorityWithTenant(
+                requestParams.Authority.AuthorityInfo,
+                idToken?.TenantId).TenantId;
+        }
+
+        public static string GetPreferredUsernameFromIdToken(bool isAdfsAuthority, IdToken idToken)
+        {
+            // The preferred_username value cannot be null or empty in order to comply with the ADAL/MSAL Unified cache schema.
+            // It will be set to "preferred_username not in id token"
```
gladjohn, line 32:
```suggestion
            // It will be set to "Missing from the token response"
```

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
```diff
@@ -320,7 +332,7 @@ private async Task<AuthenticationResult> RunAtiAsync(IPublicClientApplication pc
 
             AuthenticationResult result = null;
             var scopes = GetScopes();
-
+            var guid = Guid.NewGuid();
             var builder = pca.AcquireTokenInteractive(scopes)
                 .WithUseEmbeddedWebView(true)
                 //.WithExtraQueryParameters("domain_hint=live.com") -- will force AAD login with browser
```
gladjohn, line 338: can be removed?

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -158,9 +161,27 @@ private enum ResponseStatus : int
                 }
             }
 
+            AddPopParams(authenticationRequestParameters, authParams);
+
             return authParams;
         }
 
+        /// <summary>
+        /// Configures the Msal Runtime authenticaion request to use proof of posession.
```
gladjohn, line 170:
```suggestion
        /// Configures the MSAL Runtime authentication request to use proof of possession .
```

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -158,9 +161,27 @@ private enum ResponseStatus : int
...
+        /// </summary>
+        private static void AddPopParams(AuthenticationRequestParameters authenticationRequestParameters, NativeInterop.AuthParameters authParams)
+        {
+            // if PopAuthenticationConfiguration is set, proof of possesion will be performed via the runtime broker
```
gladjohn, line 174:
```suggestion
            // if PopAuthenticationConfiguration is set, proof of possession will be performed via the runtime broker
```

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -158,9 +161,27 @@ private enum ResponseStatus : int
...
+            if (authenticationRequestParameters.PopAuthenticationConfiguration != null)
+            {
+                authenticationRequestParameters.RequestContext.Logger.Info("[WamBroker] Proof-of-Posession is configured. Using Proof-of-Posession with broker request");
```
gladjohn, line 177:
```suggestion
                authenticationRequestParameters.RequestContext.Logger.Info("[WamBroker] Proof-of-Possession is configured. Using Proof-of-Posession with broker request");
```

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
```diff
@@ -329,6 +332,59 @@ public AcquireTokenInteractiveParameterBuilder WithParentActivityOrWindow(NSWind
 
         #endregion
 
+        /// <summary>
+        ///  Modifies the token acquisition request so that the acquired token is a Proof of Possession token (PoP), rather than a Bearer token. 
+        ///  PoP tokens are similar to Bearer tokens, but are bound to the HTTP request and to a cryptographic key, which MSAL can manage on Windows.
+        ///  Note that only the host and path parts of the request URI will be bound.
+        ///  See https://aka.ms/msal-net-pop
```
gladjohn, line 339: This page may need some updates as we release to include public flows as well

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -884,6 +884,23 @@ public static class MsalError
...
         public const string CryptoNet45 = "crypto_net45";
 
+        /// <summary>
+        /// <para>What happens?</para>The request has broker enabled and proof of possesion configured but the boker does not support proof of possesion
+        /// <para>Mitigation</para>only configure proof of possesion for public clients on windows.
+        /// </summary>
+        public const string BrokerDoesNotSupportPop = "broker_does_not_support_pop";
+
+        /// <summary>
+        /// <para>What happens?</para>The request has proof of possesion configured but does not have broker enabled. Broker is required for proof of possesion on public clients
```
gladjohn, line 894:
```suggestion
        /// <para>What happens?</para>The request has proof of possession configured but does not have broker enabled. Broker is required for proof of possession on public clients
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -884,6 +884,23 @@ public static class MsalError
...
+        /// <para>Mitigation</para>Enable the broker when proof of possesion is configured.
```
gladjohn, line 895:
```suggestion
        /// <para>Mitigation</para>Enable the broker when proof of possession is configured.
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -884,6 +884,23 @@ public static class MsalError
...
+        /// </summary>
+        public const string BrokerRequiredForPop = "broker_required_for_pop";
+
+        /// <summary>
+        /// <para>What happens?</para>The request has proof of possesion configured but does not have a nonce configured. A nonce is required for proof of possesion on public clients
```
gladjohn, line 900:
```suggestion
        /// <para>What happens?</para>The request has proof of possession configured but does not have a nonce configured. A nonce is required for proof of possession on public clients
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -884,6 +884,23 @@ public static class MsalError
...
+        /// <para>Mitigation</para>Provide a nonce when proof of possesion is configured for public clients.
```
gladjohn, line 901:
```suggestion
        /// <para>Mitigation</para>Provide a nonce when proof of possession is configured for public clients.
```

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -884,6 +884,23 @@ public static class MsalError
...
+        /// </summary>
+        public const string NonceRequiredForPopOnPCA = "nonce_required_for_pop_on_pca";
 #if iOS
         /// <summary>
         /// Xamarin.iOS specific. This error indicates that keychain access has not be enabled for the application.
```
gladjohn, line 906:
```suggestion
        /// Xamarin.iOS specific. This error indicates that keychain access has not been enabled for the application.
```

gladjohn (approved):
looks good, just a few comments

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -884,6 +884,23 @@ public static class MsalError
         /// </summary>
         public const string CryptoNet45 = "crypto_net45";
 
+        /// <summary>
+        /// <para>What happens?</para>The request has broker enabled and proof of possesion configured but the boker does not support proof of possesion
```
neha-bhargava, line 888:
```suggestion
        /// <para>What happens?</para>The request has broker enabled and proof of possession configured but the broker does not support proof of possession
```

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
```diff
@@ -329,6 +332,59 @@ public AcquireTokenInteractiveParameterBuilder WithParentActivityOrWindow(NSWind
...
+        /// <item><description>The PoP token is bound to the HTTP request, more specifically to the HTTP method (GET, POST, etc.) and to the Uri (path and query, but not query parameters).</description></item>
+        /// <item><description>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</description></item>
+        /// </list>
+        /// </remarks>
+#if iOS || ANDROID || WINDOWS_UWP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+#endif
+        public AcquireTokenInteractiveParameterBuilder WithProofOfPossession(string nonce, HttpMethod httpMethod, Uri requestUri)
+        {
+            ClientApplicationBase.GuardMobileFrameworks();
```
bgavrilMS, line 359: Can we also make these method experimental?

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -183,16 +204,19 @@ private enum ResponseStatus : int
                     logger.Warning("No correlation ID in response");
                     correlationId = null;
                 }
+                
+                //parsing Pop token from auth header if pop was performed. Otherwise use access token field.
+                var token = authResult.IsPopAuthorization ? authResult.AuthorizationHeader.Split(' ')[1] : authResult.AccessToken;
```
pmaytak, line 209: This behavior is a bit funky and not intuitive. I guess, why the difference? Should be documented a bit more (maybe here in code). Maybe changed on the interop side to be consistent? 
+@gladjohn

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
trwalke: Seems like it is used to enable certain scenarios by uncommenting when needed

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
```diff
@@ -329,6 +332,60 @@ public AcquireTokenInteractiveParameterBuilder WithParentActivityOrWindow(NSWind
...
+            if (!broker.IsPopSupported)
+            {
+                throw new MsalClientException(MsalError.BrokerDoesNotSupportPop, MsalErrorMessage.BrokerDoesNotSupportPop);
+            }
+
+            PoPAuthenticationConfiguration popConfig = new PoPAuthenticationConfiguration(requestUri);
+
+            if (string.IsNullOrEmpty(nonce))
+            {
+                throw new ArgumentNullException(nameof(nonce));
```
pmaytak, line 378: If nonce is required for this WithPop method, maybe we should mention (in XML comment?) that you can use the other overload without a nonce.

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ClientApplicationBaseExecutor.cs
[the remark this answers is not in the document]
pmaytak: Could you add repro steps/code snippet? Wouldn't checks in the PCA.ATS().WithPop() take care of this?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -174,7 +173,7 @@ public sealed partial class TokenCache : ITokenCacheInternal
                         requestParams.RequestContext.ApiEvent.DurationInCacheInMs += sw.ElapsedMilliseconds;
                     }
 
-                    if (msalAccessTokenCacheItem != null)
+                    if (msalAccessTokenCacheItem != null && !(response.TokenSource == TokenSource.Broker && response.TokenType == Constants.PoPAuthHeaderPrefix))
```
pmaytak, line 176: nit:
```suggestion
                    // Don't cache PoP access tokens from broker
                    if (msalAccessTokenCacheItem != null && !(response.TokenSource == TokenSource.Broker && response.TokenType == Constants.PoPAuthHeaderPrefix))
```

tests/Microsoft.Identity.Test.Unit/pop/PoPTests.cs
```diff
@@ -234,6 +271,103 @@ public async Task CacheKey_Includes_POPKid_Async()
...
+
+                // Act
+                AuthenticationResult result = await pca.AcquireTokenInteractive(TestConstants.s_graphScopes)
+                            .WithProofOfPossession(TestConstants.Nonce, HttpMethod.Get, new Uri(TestConstants.AuthorityCommonTenant))
+                            .ExecuteAsync()
+                            .ConfigureAwait(false);
+
+                Assert.IsNotNull(result.AccessToken);
+                Assert.AreEqual(TestConstants.s_graphScopes.FirstOrDefault(), result.Scopes.FirstOrDefault());
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
```
pmaytak, line 315: Add an assert for token type.

pmaytak (approved):

pmaytak:
Suggestion: suggested changes can be added to a batch and committed at once, rather than one by one. Otherwise, PR build is triggered for each commit. Although this has to be done from the Files page in a PR.

image

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenInteractiveParameterBuilder.cs
trwalke: The overloaded method is actually not available for the interactive flow so we are good here

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ClientApplicationBaseExecutor.cs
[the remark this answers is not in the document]
trwalke: No, the checks are not in the withPop api because the ATS api is shared with confidential clients that do not require a nonce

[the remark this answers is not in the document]
trwalke: Will add repro to issue here
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3334

PR #3319, opened 2022-05-12
ATS split (prototype)

bgavrilMS:
Today, AcquireTokenSilent is on both PCA and CCA. This prevents us from adding `WithPOP(string, string, string)` on PCA only and throwing meaningful expcetions if broker is not configured.

Can we re-create the object hirearchy on PCA and CCA without incurring a brekaing change?

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
@@ -15,12 +15,83 @@ using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 
 namespace Microsoft.Identity.Client
 {
+    /// <summary>
+    /// 
+    /// </summary>
+    public class PCAAcquireTokenSilentParameterBuilder :
+        AcquireTokenSilentParameterBuilder
+    {
+        internal PCAAcquireTokenSilentParameterBuilder(IClientApplicationBaseExecutor clientApplicationBaseExecutor) : base(clientApplicationBaseExecutor)
+        {
+        }
+
+        private new AcquireTokenSilentParameterBuilder WithSendX5C(bool withSendX5C)
+        {
+            return null;
+        }
+
+
+        /// <summary>
+        ///  Modifies the token acquisition request so that the acquired token is a Proof of Possession token (PoP), rather than a Bearer token. 
+        ///  PoP tokens are similar to Bearer tokens, but are bound to the HTTP request and to a cryptographic key, which MSAL can manage on Windows.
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -15,12 +15,91 @@
...
+                if (!ServiceBundle.Config.IsBrokerEnabled)
+                {
+                    throw new MsalClientException(MsalError.BrokerRequiredForPop, MsalErrorMessage.BrokerRequiredForPop);
+                }
+            }
+
+            return this;
+        }
+
+        private new PCAAcquireTokenSilentParameterBuilder WithProofOfPossession(PoPAuthenticationConfiguration config)
```
bgavrilMS, line 91: private new doesn't seem to work to hide the method from the base class.

src/client/Microsoft.Identity.Client/IPublicClientApplication.cs
```diff
@@ -8,13 +8,68 @@
...
+        ///
+        /// See also the additional parameters that you can set chain:
+        /// <see cref="AbstractAcquireTokenParameterBuilder{T}.WithTenantId(string)"/> 
+        /// to request a token for a different authority than the one set at the application construction
+        /// <see cref="AcquireTokenSilentParameterBuilder.WithForceRefresh(bool)"/> to bypass the user token cache and
+        /// force refreshing the token, as well as
+        /// <see cref="AbstractAcquireTokenParameterBuilder{T}.WithExtraQueryParameters(Dictionary{string, string})"/> to
+        /// specify extra query parameters
+        /// </remarks>
+        new PCAAcquireTokenSilentParameterBuilder AcquireTokenSilent(IEnumerable<string> scopes, IAccount account);
```
bgavrilMS, line 44: Unfortunately this is tehnically a breaking change. One thing to explore is to use extension methods instead of changing the interface.

src/client/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -234,5 +236,15 @@ public bool IsBrokerAvailable()
                 username,
                 password);
         }
+
+        PCAAcquireTokenSilentParameterBuilder IPublicClientApplication.AcquireTokenSilent(IEnumerable<string> scopes, IAccount account)
+        {
+            throw new NotImplementedException();
```
bgavrilMS, line 242: Did not implement this, but need to explore using extension methods.

bgavrilMS:
@trwalke - I won't be able to make more progress on this in the short term - please have a look to see if you can use this.

PR #3338, opened 2022-05-20
ATS extra check for auth scheme

trwalke:
Fixes # 
ATS extra check for auth scheme

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
@@ -195,14 +195,16 @@ namespace Microsoft.Identity.Client
             popConfig.HttpMethod = httpMethod ?? throw new ArgumentNullException(nameof(httpMethod));
             popConfig.Nonce = nonce;
 
-            CommonParameters.PopAuthenticationConfiguration = popConfig;
-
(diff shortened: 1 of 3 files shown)


SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -195,6 +195,12 @@ public AcquireTokenSilentParameterBuilder WithProofOfPossession(string nonce, Ht
             popConfig.HttpMethod = httpMethod ?? throw new ArgumentNullException(nameof(httpMethod));
             popConfig.Nonce = nonce;
 
+            //POP Auth scheme should not wrap and sign token when broker is enabled for public clients
+            if (ServiceBundle.Config.IsBrokerEnabled)
+            {
+                popConfig.SignHttpRequest = false;
```
bgavrilMS, line 201: What about AcquireTokenInteractive and other flows? Should we put this somewhere else?

trwalke: Only ATS is shared between CCA and PCA so it needs the auth scheme. This is the only place this is needed.

pmaytak (approved):

PR #3342, opened 2022-05-23
Ensuring tokens are not pulled from cache when pop with broker is executed

trwalke:
Fixes # 
Ensuring tokens are not pulled from cache when pop with broker is executed

Skip checking local cache when pop with broker is executed

**Testing**
Unit test


src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -48,6 +48,13 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             bool isBrokerConfigured = AuthenticationRequestParameters.AppConfig.IsBrokerEnabled &&
                                       ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
 
+            if (isBrokerConfigured && AuthenticationRequestParameters.PopAuthenticationConfiguration != null)
+            {
```
pmaytak, line 52: Would we ever get here if someone specifies ADFS authority, enables broker, enables Pop? I don't think so, right, we have checks in builders for this?

pmaytak (approved):
Nit: add a context to the PR description or the associated issue for the documentation purposes - why this change is needed.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
trwalke: It will get there, but im not sure that is a problem? Im assuming WAM can handle ADFS accounts? 
Also, not sure if ADFS supports pop at the moment.
@gladjohn @bgavrilMS?

bgavrilMS: A good test case to code up.

bgavrilMS (approved):

Issue #3323, opened 2022-05-16
[Bug] MSAL cache can get confused when WAM has two accounts that match an Identifier


JPWilli:
**Logs and network traces**
Without logs or traces, it is unlikely that the team can investigate your issue. Capturing logs and network traces is described in [Logging wiki](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging).
Due to PII concerns, I've given @bgavrilMS and @pmaytak logs directly.

**Which version of MSAL.NET are you using?**
4.42.1

**Platform**
.Net 5

**What authentication flow has the issue?**
* Desktop / Mobile
    * [x] Interactive
    * [ ] Integrated Windows Authentication
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 
    * [ ] Service to Service calls

Other?

**Is this a new or existing app?**
Visual Studio is seeing this when enabling WAM for a subset of users.

**Repro**
This one is hard to nail down, the best I can say if we ask MSAL  to GetTokenSilent for ARM, on a non-home tenant, and given [not-sure-what] circumstance, that can cause WAM to get/maintain a 2nd account for the same identifier.

Then, subsequent calls to GetTokenSilent can use the wrong WAM account to back the MSAL account, and return something with the non-home-tenant as the Home Tenant.

Again, logs provided separately through email due to PII.

**Expected behavior**
Home Tenant should be constant for a given MSAL account, regardless of what tokens are requested.

bgavrilMS:
WAM does have some logic to match an IAccount and WebAccount based on ID with a fallback to UPN.

pmaytak:
After some investigation, not sure if I was able to repro this exactly. Provided info/updates via email. At the moment pending more specific repro steps/debugging session.

Some notes on the scenarios tested (using a home and a guest account):
1. GetAccounts (ListOSAccounts=true, 0 accounts in MSAL cache, 2 accounts in WAM)
	- Returns 1 account, 1 saved into MSAL cache (whichever account was the last to be processed from WAM)
2. ATI with Guest account  (ListOSAccounts=true)
	- Returns 1 account, 1 saved into MSAL cache (whichever account was the last to be processed from WAM)
	- Then GetAccounts
		- Same as above.
3. ATI with Guest account  (ListOSAccounts=false)
	- Returns 1 guest account
[...]

PR #3356, opened 2022-05-27
Return accounts with tenant profiles from GetAccounts for token received from broker

### LINKED ISSUE #3349: [Bug] Tenant profiles are not returned in GetAccounts for broker tokens
Latest

**Repro**
1. Use [NetCoreWinFormsWam](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/tree/master/tests/devapps/WAM/NetCoreWinFormsWam) dev app.
2. Clean MSAL cache.
3. Select `Use WAM`.
4. Login interactively with an account.
5. MSAL cache should have an access token, account, ID token, but no refresh token.
6. Call GetAccounts.
7. Observe that one account is returned but tenant profiles are empty.

**Actual behavior**
In GetAccounts [this code](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/89d45db6ecfe79b4d39a5285c60fa9a44d3bfba9/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L937) that adds tenant profiles is skipped because there are no refresh tokens in broker flow. And [this code](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/89d45db6ecfe79b4d39a5285c60fa9a44d3bfba9/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L963) doesn't add them.

**Expected behavior**
Tenant profiles should be returned.

pmaytak:
Fixes #3349, related to [#3323](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3323)

Added a missing call to retrieve tenant profiles for accounts which do not have an associated refresh token (such as in WAM scenario)

**Testing**
Updated a test + manual testing.

**Performance impact**
Should be negligible since this should only be called in PCA.


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -962,17 +962,23 @@ async Task<IEnumerable<IAccount>> ITokenCacheInternal.GetAccountsAsync(Authentic
...
-                    acc.WamAccountIds.ContainsKey(requestParameters.AppConfig.ClientId)))
+                foreach (MsalAccountCacheItem cachedAccount in accountCacheItems)
                 {
-                    var wamAccount = new Account(
-                        wamAccountCache.HomeAccountId,
-                        wamAccountCache.PreferredUsername,
-                        environment,
-                        wamAccountCache.WamAccountIds);
+                    if (!clientInfoToAccountMap.ContainsKey(cachedAccount.HomeAccountId) &&
+                        cachedAccount.WamAccountIds != null &&
```
pmaytak, line 968: Thought about adding a home tenant filter here to return only home accounts with tenant profiles, as per guest account spec. However, that breaks a scenario when (1) there's only one guest account in the cache retrieved from WAM and (2) List OS Accounts filter is false. In that scenario the filter would filter that account out and return nothing. We might need to refactor some WAM code related to this because per spec when querying broker for a guest account, it will actually return a home account.

bgavrilMS: Hmm, we have an algorithm to resolve guest account from WAM, i.e. we can always find the client_info. It's pretty complex, i.e. sometimes we'll have to actually make a token request.

bgavrilMS (approved):

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
pmaytak: So for like an ATI with a guest authority, we do use client info and home account ID as a partition key but the account object represents guest account.

trwalke (approved):

PR #3360, opened 2022-05-30
Fix Expires In on Runtime Broker

### LINKED ISSUE #3354: [Bug] Auth result from runtime broker is not parsed correctly
**Repro**
Enable broker preview and acquire token interactively.

**Actual behavior**
In [`ParseRuntimeResponse`](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/89d45db6ecfe79b4d39a5285c60fa9a44d3bfba9/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs#L217), `ExpiresIn` ends up being set to 0 because a DateTime and not a duration is passed into `DateTimeHelpers.GetDurationFromWindowsTimestamp`, which fails parsing.

**Expected behavior**
`ExpiresIn` should be set to a valid duration in seconds.

**Possible solution**
Something like:
```csharp
ExpiresIn = DateTimeHelpers.GetDurationFromNowInSeconds(DateTimeHelpers.DateTimeToUnixTimestamp(DateTime.SpecifyKind(authResult.ExpiresOn, DateTimeKind.Utc))),
```

gladjohn:
Fixes #3354 

Fix Expires In on Runtime Broker WAM

**Testing**
manual testing

**Performance impact**
none

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -196,7 +196,6 @@ namespace Microsoft.Identity.Client.Broker
         {
             try
             {
-                string expiresOn = authResult.ExpiresOn.ToString();
                 string correlationId = authenticationRequestParameters.CorrelationId.ToString("D");
 
                 if (string.IsNullOrWhiteSpace(correlationId))
@@ -214,7 +213,7 @@ namespace Microsoft.Identity.Client.Broker
                     IdToken = authResult.RawIdToken,
                     CorrelationId = correlationId,
                     Scope = authResult.GrantedScopes,
-                    ExpiresIn = DateTimeHelpers.GetDurationFromWindowsTimestamp(expiresOn, logger),
+                    ExpiresIn = GetExpiresInFromAuthResult(authResult.ExpiresOn),
                     ClientInfo = authResult.Account.ClientInfo.ToString(),
                     TokenType = authResult.IsPopAuthorization ? Constants.PoPAuthHeaderPrefix : BrokerResponseConst.Bearer,
                     WamAccountId = authResult.Account.Id,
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -240,5 +239,15 @@ private static string GetExpectedRedirectUri(string clientId)
         {
             return $"ms-appx-web://microsoft.aad.brokerplugin/{clientId}";
         }
+
+        /// <summary>
+        /// Get ExpiresIn from Auth Result
+        /// </summary>
+        /// <param name="expiresOn"></param>
+        private static long GetExpiresInFromAuthResult(DateTime expiresOn)
```
bgavrilMS, line 247: Is it possible to get this covered by an integration test? Maybe it's worth waiting a bit until we have an integration test.

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -240,5 +239,15 @@ private static string GetExpectedRedirectUri(string clientId)
...
+        {
+            return DateTimeHelpers.GetDurationFromNowInSeconds
```
bgavrilMS, line 249: This seems very inefficient to transform a DateTime into Unix timestamp. The `expiresOn` can be assumed to be UTC. Can you find a better way to transform a DateTime into what MSAL needs?

bgavrilMS (changes requested):
Needs some test. Should not transform to Unix timestamp just to get the total seconds.

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
gladjohn: updated

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -214,7 +213,7 @@ private static void AddPopParams(AuthenticationRequestParameters authenticationR
                     IdToken = authResult.RawIdToken,
                     CorrelationId = correlationId,
                     Scope = authResult.GrantedScopes,
-                    ExpiresIn = DateTimeHelpers.GetDurationFromWindowsTimestamp(expiresOn, logger),
+                    ExpiresIn = (long)(authResult.ExpiresOn - DateTime.UtcNow).TotalSeconds,
```
pmaytak, line 216: `authResult.ExpiresOn.Kind` is `Unspecified`. Wonder we need to explicitly specify Utc kind, otherwise might get some funky datetime business.

bgavrilMS: I propose that we use `DateTimeOffset` instead of `DateTime`. MSAL should not be using `DateTime` in general, because we don't display zone-specific dates.

bgavrilMS (approved):

pmaytak (approved):

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
gladjohn: merging to get this in for this week's release. Will add a test when the Integration pipeline is ready

Issue #3355, opened 2022-05-27
[Feature Request] Support for KeyVault sign when using ClientCredentials


svrooij:
**Is your feature request related to a problem? Please describe.**
All samples related to Client Credentials suggest to put a certificate in the Key Vault, registering that certificate in the application and accessing the certificate using a managed identity. Which is all great, but I think no one is aware that the certificate might be exported either by a hack changing the application code or by getting in from memory.

Not even talking about what would happen if a malicious developer would extract the certificate.

All samples show something like:

```csharp
using Azure.Identity;
using Azure.Security.KeyVault.Secrets;

public async Task<string> GetToken(...)
{
  const string ceyVaultUri = "https://{kv-domain}.vault.azure.net/";
  const string certName = "{some-certificate-name}";
  var tokenCredential = new DefaultAzureCredential();
  var secretClient = new SecretClient(new Uri(keyVaultUri), tokenCredential);
  var secretResult = await secretClient.GetSecretAsync(certName).ExecuteAsync();
  // Certificate is now present on client machine
  var certificate = new X509Certificate2(Convert.FromBase64String(secretResult.Value.Value));

  const string clientId = "d294e746-425b-44fa-896c-dacf2c7938b8";
  const string tenantId = "42a26c5d-b8ed-4f1b-8760-655f98154373";

  var app = ConfidentialClientApplicationBuilder
    .Create(clientId)
    .WithAuthority(AzureCloudInstance.AzurePublic, tenantId)
    .WithCertificate(certificate)
    .Build();

  var tokenResult = await app
    .AcquireTokenForClient(new[] { "https://graph.microsoft.com/.default" })
    .ExecuteAsync();

  return tokenResult.AccessToken;
}
```

**Describe the solution you'd like**

I would like to see a warning in the documentation about the possibility that people can extract the certificate and possibly gain persistent access to customer data.
The solution is in the KeyVault, instead of downloading the certificate and signing the client assertion on the application side, you can also use the Key Vault to sign the unsigned assertion in the cloud, mitigating the possibility to extract the data.

I created a [poc](https://github.com/Smartersoft/identity-client-assertion/tree/main/src/Smartersoft.Identity.Client.Assertion) to do exactly that and I could use some assistance in getting this in MSAL.net if you're open to that.

```csharp
// Seeing support for 'WithKeyVaultKey' instead of 'WithCertificate' is the main goal here
[...]

PR #3376, opened 2022-06-08
Access Client ID and TokenEndpoint

### LINKED ISSUE #3352: [Feature Request] Add a new WithClientAssertion API that exposes the token endpoint
Applications that create their own client assertion typically do it as: 

```csharp
WithClientAssertion( "aud: ${input_authority}/v2.0/token, iss: ${client_id}")
```
However, MSAL may decide to change the input_authority, because ESTS wants the SDK to use the "preferred_network". For example, for the public cloud, if the app developer configures "sts.windows.net/12345", MSAL will use "login.microsoftonline.com/12345". And it looks like ESTS does not support sending client_assertion with `aud: sts.windows.net/12345/` over `login.microsoftonline.com/12345`. These 2 must match. 

Proposal: 

```csharp
WithClientAssertion(Func<RequestDetails, Task<string>>);

// where RequestDetails has following properties: 
// TokenEndpoint
// ClientId
// CancellationToken
```  

CC @jmprieur

svrooij:
WithClientAssertion did not have access to client id or token endpoint
Which can be very useful

Fixes #3352
Related to #3355

- `WithClientAssertion(...)` now has access to both the client id and the token endpoint, both of which are key properties of the payload in a client assertion.
- Maybe also allow public access to `JsonWebTokenConstants.JwtToAadLifetimeInSeconds` or add it as a property to `AssertionRequestOptions`?

**Testing**

- Added tests, but not able to run them locally. Or even run the build on my machine, flying blind here 😉 

**Performance impact**

None


bgavrilMS:
I'm ok to take this. I think creating the assertion without knowing the token endpoint or the client id is pretty difficult. @jmprieur  for thoughts on public API and soft-deprecating the other overloads of `WithClientAssertion(Func)`

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -226,6 +226,29 @@ public ConfidentialClientApplicationBuilder WithClientAssertion(Func<string> cli
...
         }
 
+        /// <summary>
+        /// Configures a delegate that creates a client assertion. See https://aka.ms/msal-net-client-assertion
+        /// </summary>
+        /// <param name="clientAssertionDelegate">delegate computing the client assertion used to prove the identity of the application to Azure AD.
+        /// This is a delegate that computes a Base-64 encoded JWT for each authentication call.</param>
+        /// <returns>The ConfidentialClientApplicationBuilder to chain more .With methods</returns>
+        /// <remarks> Callers can use this mechanism to cache their assertions </remarks>
+        public ConfidentialClientApplicationBuilder WithClientAssertion(Func<AssertionRequestOptions, string> clientAssertionDelegate)
```
bgavrilMS, line 236: We should allow support for async, i.e. `Func<AssertionRequestOptions, Task<string>>`. Getting an assertion signed by KV is most likely an async operation.

bgavrilMS, line 236: I'd probably refactor the existing `WithClientAssertion` overload to use this mechansm now. And probably hide the old mechanism (via EditorBrowsable=false).

svrooij: It supports both, see below. In the current solution there is both a sync and an async version. Just supported both in this proposed solution

svrooij: The idea was to not break existing functionality. Marking it as EditorBrowsable=false is a great idea. Had not thought about that.

How do the documentation pages get updated? Because if we mark this as not browseable the docs won't match the functions.

bgavrilMS: Not sure I understand this. If I want to implement a delegate like: 

```csharp
public string AsyncImplementation(AssertionRequestOptions options)
{
     string clientId = options.ClientId;
     string tokenEndpoint = options.TokenEndpoint;

     string json = CreateJson(clientId, tokenEndpoint);

     // async operation -> method must return a Task<string> not a string.
     string signedJson = await KeyVaultSignAsync(json);
     
    return signedJson;
}

```

svrooij: @bgavrilMS your call, do you want me to remove it or leave it to support the same as previously?

bgavrilMS: "Async all the way" :). We've learned that the hard way. 

I can update the code as well if you wish.

bgavrilMS:
@svrooij - with regards to the docs, I am not sure this will be possible. You need to login first, and an edit button should appear

image


On edit, it'll take you to a GitHub repo where the docs are stored. Not sure you are able to propose a PR there.

svrooij:
@bgavrilMS I think I should be able to... since I'm already mentioned there as a contributor:

image

bgavrilMS (approved):

bgavrilMS:
@trwalke  and @pmaytak - can you please help review this ?

src/client/Microsoft.Identity.Client/Internal/ClientCredential/SignedAssertionDelegateClientCredential.cs
```diff
@@ -13,15 +13,28 @@ namespace Microsoft.Identity.Client.Internal.ClientCredential
...
         internal Func<CancellationToken, Task<string>> _signedAssertionDelegate { get; }
+        internal Func<Microsoft.Identity.Client.AssertionRequestOptions, Task<string>> _signedAssertionWithInfoDelegate { get; }
 
+        [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
         public SignedAssertionDelegateClientCredential(Func<CancellationToken, Task<string>> signedAssertionDelegate)
         {
             _signedAssertionDelegate = signedAssertionDelegate;
         }
 
+        public SignedAssertionDelegateClientCredential(Func<Microsoft.Identity.Client.AssertionRequestOptions, Task<string>> signedAssertionDelegate)
```
jmprieur, line 24: Does it need to be fully qualified?

svrooij:
@jmprieur thanks for the cliffnotes on federated credentials. Can I summarize that the client would generate a token for resource in the AD (which should be created first) and then that token from Azure AD can be used as a client assertion for some application. Since the application admin needs to specify the aud, iss and sub claim. It might as well be possible to authorize specific developers to use there own credentials during development. Can I contribute a guide on how to do just that?

jmprieur:
@bgavrilMS @trwalke @pmaytak 
I believe the doc in doc.ms is already merged and published ... I had not realized the PR was not merged ....?

jmprieur (approved):
LGTM
Thanks @svrooij

svrooij:
@bgavrilMS what is the usual release schedule for MSAL.NET? And is it an idea to document this?

Something like, "if your PR is merged to main, it will be included in the next release which is every x days/weeks".

Maybe even a PR bot telling just that.

PR #3381, opened 2022-06-09
Ensure region configured by user is logged

bgavrilMS:
Fixes # 

**Testing**

**Performance impact**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
@@ -174,7 +174,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         public string LongRunningOboCacheKey { get; set; }
 
         public KeyValuePair<string, string>? CcsRoutingHint { get; set; }
-#endregion
+        #endregion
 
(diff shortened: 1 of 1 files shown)


pmaytak (approved):

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -199,6 +199,7 @@ public void LogParameters()
             builder.AppendLine("CorrelationId - " + CorrelationId);
             builder.AppendLine("UserAssertion set: " + (UserAssertion != null));
             builder.AppendLine("LongRunningOboCacheKey set: " + !string.IsNullOrWhiteSpace(LongRunningOboCacheKey));
+            builder.AppendLine("Region configured: " + AppConfig.AzureRegion);
```
neha-bhargava, line 202: I think with the configured region we also need the source of region discovery. If we are suspecting the region discovered is wrong then it could be either environment variable or imds

bgavrilMS: Agreed, but that is actually logged as part of the response.

Issue #3308, opened 2022-05-06
[Feature Request] WAM runtime support for AcquireTokenByUsernamePassword


gladjohn:
This is tracking work item for SigninSilently() with additional parameters `MSALRuntime_Username` and `MSALRuntime_Password

More info on the API Review doc linked [here](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/%5BXplat%5D%20MSALRuntime/Windows%20MSALRuntime/IntegrationWithMsalRuntime.md&version=GBdev&line=136&lineEnd=136&lineStartColumn=73&lineEndColumn=166&lineStyle=plain&_a=contents)

- [ ] Get it working for Bearer tokens
- [ ] Integration test for Bearer
- [ ] Also enable "WithProofOfPosession" API with this flow.
- [ ] Integration test for POP

pmaytak:
[Released in 4.46.0](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.46.0)

PR #3400, opened 2022-06-16
Add username password flow msal runtime

neha-bhargava:
Fixes # 
#3308 

Add username password flow for msal runtime

**Testing**
Add integration tests

**Performance impact**


src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs
```diff
@@ -38,12 +41,31 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
-                await SendTokenRequestAsync(GetAdditionalBodyParameters(userAssertion), cancellationToken).ConfigureAwait(false);
+
+            MsalTokenResponse msalTokenResponse = await GetTokenResponseAsync(cancellationToken).ConfigureAwait(false);
+            
             return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
         }
 
+        private async Task<MsalTokenResponse> GetTokenResponseAsync(CancellationToken cancellationToken)
+        {
+            if (_requestParameters.AppConfig.IsBrokerEnabled)
```
bgavrilMS, line 52: You need to also check `broker.IsBrokerInstalledAndInvokable(_authenticationRequestParameters.AuthorityInfo.AuthorityType)`

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -227,6 +228,40 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
...
+                        msalTokenResponse = WamAdapters.ParseRuntimeResponse(result, authenticationRequestParameters, _logger);
+                    }
+                    else
+                    {
+                        WamAdapters.ThrowExceptionFromWamError(result, authenticationRequestParameters, _logger);
+                    }
+                }
+            }
+
+            return msalTokenResponse;
```
bgavrilMS, line 262: Am curious about the following: 

1. is there an RT given back to the app if the flow works?
2. does it work with POP?

neha-bhargava: Implemented the POP and it works with username and password and returns the pop token

bgavrilMS:
@neha-bhargava - can we have separate PRs for the ROPC and for POP support? Feel free to separate the work item as well if it helps tracking. 

ROPC + integration test can be one task
ROPC + POP + integration test second task

I think it will make it easier for folks to review the code.

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
```diff
@@ -64,6 +64,12 @@ public Task<MsalTokenResponse> AcquireTokenSilentDefaultUserAsync(Authentication
             return Task.FromResult<MsalTokenResponse>(null);
         }
 
+        public Task<MsalTokenResponse> AcquireTokenByUsernamePasswordAsync(AuthenticationRequestParameters authenticationRequestParameters, AcquireTokenSilentParameters acquireTokenSilentParameters)
```
pmaytak, line 67: Remove, unused.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerUsernamePasswordRequestComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+
+    internal class BrokerUsernamePasswordRequestComponent : ITokenRequestComponent
+    {
+        internal IBroker Broker { get; }
+        private readonly AcquireTokenByUsernamePasswordParameters _usernamePasswordParameters;
+        private readonly string _optionalBrokerInstallUrl; // can be null
+        private readonly AuthenticationRequestParameters _authenticationRequestParameters;
+        private readonly IServiceBundle _serviceBundle;
+        private readonly ICoreLogger _logger;
```
pmaytak, line 24: Needs to be merged with main to use new logger.

src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs
pmaytak, line 52: Add some logs like in `InteractiveRequest.GetTokenResponseAsync`.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerUsernamePasswordRequestComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Internal.Broker
+{
+
+    internal class BrokerUsernamePasswordRequestComponent : ITokenRequestComponent
+    {
+        internal IBroker Broker { get; }
+        private readonly AcquireTokenByUsernamePasswordParameters _usernamePasswordParameters;
+        private readonly string _optionalBrokerInstallUrl; // can be null
```
pmaytak, line 21: Some of these variables can be removed, unused.

tests/Microsoft.Identity.Test.Integration.netcore/WAM/RuntimeBrokerTests.cs
```diff
@@ -135,5 +136,100 @@ public async Task WamInteractiveAuthNoWindowHandleAsync()
             }
 
         }
+
+        [TestMethod]
+        public async Task WamUsernamePasswordRequestAsync()
```
pmaytak, line 141: I would maybe add a test for old broker returning correct exception.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerUsernamePasswordRequestComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+                _authenticationRequestParameters, 
+                _usernamePasswordParameters)
+                .ConfigureAwait(false);
+
+            ValidateResponseFromBroker(tokenResponse);
+
+            return tokenResponse;
+        }
+
+        internal /* internal for test */ void ValidateResponseFromBroker(MsalTokenResponse msalTokenResponse)
```
bgavrilMS, line 60: I think it's the broker's responsability to throw exceptions.

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
```diff
@@ -448,5 +448,10 @@ private void CheckPackageForPowerOptimization(string package)
                        null,
                        UiRequiredExceptionClassification.AcquireTokenSilentFailed);
         }
+
+        public Task<MsalTokenResponse> AcquireTokenByUsernamePasswordAsync(AuthenticationRequestParameters authenticationRequestParameters, AcquireTokenByUsernamePasswordParameters acquireTokenByUsernamePasswordParameters)
+        {
+            throw new PlatformNotSupportedException();
```
bgavrilMS, line 454: Why do you throw on Android but not on iOS? Should be consistent. Probably return null to fallback to MSAL's default logic.

tests/Microsoft.Identity.Test.Integration.netcore/WAM/RuntimeBrokerTests.cs
```diff
@@ -135,5 +136,100 @@ public async Task WamInteractiveAuthNoWindowHandleAsync()
...
+
+            // Assert the account is removed
+            accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(accounts);
+            Assert.AreEqual(0, accounts.Count());
+        }
+
+        [TestMethod]
+        public async Task WamUsernamePasswordRequestMsaPassthroughAsync()
```
bgavrilMS, line 179: MSA-PT, as the name suggests, is to allow an MSA users to acquire a token for certain client apps which do not natively support MSA.

However: 

1. your test does fetch an MSA user from the lab, it fetches an AAD user
2. MSA does not suport ROPC. In any shape or form.

Pls delete the test

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -227,6 +228,40 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
...
+            var cancellationToken = authenticationRequestParameters.RequestContext.UserCancellationToken;
+            MsalTokenResponse msalTokenResponse = null;
+
+            _logger.Verbose("[WamBroker] Acquiring token with Username Password flow.");
+
+            using (var core = new NativeInterop.Core())
+            using (AuthParameters authParams = WamAdapters.GetCommonAuthParameters(authenticationRequestParameters, _wamOptions.MsaPassthrough))
+            {
+                authParams.Properties["MSALRuntime_Username"] = acquireTokenByUsernamePasswordParameters.Username;
+                authParams.Properties["MSALRuntime_Password"] = new System.Net.NetworkCredential(string.Empty, acquireTokenByUsernamePasswordParameters.Password).Password;
```
bgavrilMS, line 244: To transform from a SecureString to a string, please use https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerUsernamePasswordRequestComponent.cs
```diff
@@ -0,0 +1,84 @@
...
+            _authenticationRequestParameters = authenticationRequestParameters;
+            _usernamePasswordParameters = acquireTokenByUsernamePasswordParameters;
+            _serviceBundle = authenticationRequestParameters.RequestContext.ServiceBundle;
+            Broker = broker;
+            _logger = _authenticationRequestParameters.RequestContext.Logger;
+        }
+
+        public async Task<MsalTokenResponse> FetchTokensAsync(CancellationToken cancellationToken)
+        {
+            if (Broker.IsBrokerInstalledAndInvokable(_authenticationRequestParameters.AuthorityInfo.AuthorityType))
```
bgavrilMS, line 40: The fallback condition is not unit tested: if broker is not installed or available, MSAL should do ROPC on its own. This should be tested via unit tests, not integration test.

bgavrilMS (changes requested):
Fallback when broker is not installed is not correct. Mobile brokers not consistent.

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
neha-bhargava: This method is added to IBroker and needs to be implemented here.

trwalke: I think this overload is not needed. I don't see an overload that accepts a acquireTokenSilentParameters parameter. Also, not sure this flow needs these parameters anyway as they are for silent requests?? I think the logic should live in the method on line 84.

tests/Microsoft.Identity.Test.Integration.netcore/WAM/RuntimeBrokerTests.cs
```diff
@@ -135,5 +136,100 @@ public async Task WamInteractiveAuthNoWindowHandleAsync()
...
+            await pca.RemoveAsync(account).ConfigureAwait(false);
+
+            // Assert the account is removed
+            accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(accounts);
+            Assert.AreEqual(0, accounts.Count());
+        }
+
+        private void AssertAuthResult(AuthenticationResult result, TokenSource tokenSource, string tenantId)
```
trwalke, line 220: Maybe this can move to the MsalAssert class to usable everywhere. seems like a useful assert. There is already a method there named AssertAuthResult so this will be a useful overload

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
trwalke, line 67: I think this overload is not needed. I don't see an overload that accepts a acquireTokenSilentParameters parameter. Also, not sure this flow needs these parameters anyway as they are for silent requests?? I think the logic should live in the method on line 84.

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerUsernamePasswordRequestComponent.cs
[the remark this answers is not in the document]
neha-bhargava: This is for all the brokers android, ios and wam?

src/client/Microsoft.Identity.Client/Platforms/Android/Broker/AndroidAccountManagerBroker.cs
neha-bhargava: Added logic to return null to fallback to msal's default logic.

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
```diff
@@ -83,7 +77,8 @@ public Task<MsalTokenResponse> AcquireTokenByUsernamePasswordAsync(Authenticatio
 
         public Task<MsalTokenResponse> AcquireTokenByUsernamePasswordAsync(AuthenticationRequestParameters authenticationRequestParameters, AcquireTokenByUsernamePasswordParameters acquireTokenByUsernamePasswordParameters)
         {
-            throw new PlatformNotSupportedException();
+            _logger.Info("NullBroker - returning null on ROPC request.");
```
bgavrilMS, line 80: Verbose?

src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs
```diff
@@ -51,15 +54,30 @@ private async Task<MsalTokenResponse> GetTokenResponseAsync(CancellationToken ca
...
+
+                    if (brokerTokenResponse != null)
+                    {
+                        _logger.Info("Broker attempt completed successfully. ");
+                        Metrics.IncrementTotalAccessTokensFromBroker();
+                        return brokerTokenResponse;
+                    }
+                }
+
+                _logger.Info("Broker attempt did not complete.");
```
bgavrilMS, line 78: "Broker request not attempted because the broker is not available"

tests/Microsoft.Identity.Test.Integration.netcore/WAM/RuntimeBrokerTests.cs
```diff
@@ -174,47 +174,19 @@ public async Task WamUsernamePasswordRequestAsync()
             Assert.IsNotNull(accounts);
             Assert.AreEqual(0, accounts.Count());
         }
+      
 
         [TestMethod]
-        public async Task WamUsernamePasswordRequestMsaPassthroughAsync()
+        public async Task WamUsernamePasswordRequestOldBrokerAsync()
```
bgavrilMS, line 180: Can we move this to be a unit test?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
neha-bhargava: This is the pattern followed by other methods here. Do you think these should be changed to Verbose?

tests/Microsoft.Identity.Test.Integration.netcore/WAM/RuntimeBrokerTests.cs
neha-bhargava: Removing this since it is no longer valid as per https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3400#discussion_r908418469.

pmaytak (approved):

tests/Microsoft.Identity.Test.Integration.netcore/WAM/RuntimeBrokerTests.cs
```diff
@@ -135,5 +136,43 @@ public async Task WamInteractiveAuthNoWindowHandleAsync()
...
+            MsalAssert.AssertAuthResult(result, TokenSource.Cache, labResponse.Lab.TenantId);
+
+            // Remove Account
+            await pca.RemoveAsync(account).ConfigureAwait(false);
+
+            // Assert the account is removed
+            accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(accounts);
+            Assert.AreEqual(0, accounts.Count());
```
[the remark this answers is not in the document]
neha-bhargava: This is added to test MSAL Runtime flows since those can be tested via integration test only using username password flow

src/client/Microsoft.Identity.Client/Internal/Broker/BrokerUsernamePasswordRequestComponent.cs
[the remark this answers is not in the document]
bgavrilMS: Theoretically yes, we should have the same behaviour everywhere. But note that we do not expose ROPC, as far as I remmeber, on android and ios.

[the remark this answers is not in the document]
bgavrilMS: I think your unit test, where you use a NullBroker tests this, because NullBroker acts like a broker that is not available.

PR #3440, opened 2022-07-01
[Prototype] For client credentials, use cache item key as cache key, instead of partition key.

pmaytak:
Continuation off of #3439. Related to cache improvements #3020.

Rough prototype for client credentials flow to use access token cache item key as cache key, instead of partition key. This was something mentioned/discussed some time ago when we started talking about caching improvements. This is a performance optimization. Currently in FindAccessToken method, we first retrieve all access tokens from the cache partition (for client credentials it's clientID+tenantID), and then apply filters to the collection. We can try to retrieve the AT by AT key, and since app ATs have `resource/.default` scope, we can skip the scope filtering. We can also skip tenant filtering. Additionally the internal accessors can be then changed from dictionary of dictionaries to just flat dictionary.

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -176,11 +176,6 @@ namespace Microsoft.Identity.Client
                         requestParams.RequestContext.ApiEvent.DurationInCacheInMs += sw.ElapsedMilliseconds;
                     }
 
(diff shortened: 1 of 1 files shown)


jmprieur:
@pmaytak what is the underlying reason for doing this change? What's the work item? problem to solve?

pmaytak:
@jmprieur This was something mentioned/discussed some time ago when we started talking about caching improvements. This is a performance optimization. Basically keying app tokens by AT key and skipping some filtering. Updated the PR description.

PR #3443, opened 2022-07-01
Hard deprecate AcquireTokenSilent(scopes, login_hint)

### LINKED ISSUE #3403: [Bug] AcquireTokenSilent(scopes, login_hint) is not valid on CCA
msal 4.45, web site scenario

`AcquireTokenSilent(scopes, login_hint)` is a convenience API on PCA only. Internally it does a "GetAccounts" operation, which is banned on CCA, because it does not enforce caching. 

Solution: 

- hard deprecate this method on CCA
- deprecation message should ask customers to use `AcquireTokenSilent(scopes, account)`

MichelZ:
Fixes #3403

### DIFF
--- a/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
+++ b/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
@@ -76,6 +76,12 @@ namespace Microsoft.Identity.Client
         AcquireTokenOnBehalfOfParameterBuilder AcquireTokenOnBehalfOf(IEnumerable<string> scopes, UserAssertion userAssertion);
 
         /// <summary>
+        /// Please use <see cref="IClientApplicationBase.AcquireTokenSilent(IEnumerable{string}, IAccount)"/>.
+        /// </summary>
+        [Obsolete("Use AcquireTokenSilent(scopes, account)", true)]
+        AcquireTokenSilentParameterBuilder AcquireTokenSilent(IEnumerable<string> scopes, string loginHint);
+
+        /// <summary>
         /// [V3 API] Computes the URL of the authorization request letting the user sign-in and consent to the application accessing specific scopes in
         /// the user's name. The URL targets the /authorize endpoint of the authority configured in the application.
         /// This override enables you to specify a login hint and extra query parameter.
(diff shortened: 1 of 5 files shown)


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
```diff
@@ -75,6 +75,12 @@ public partial interface IConfidentialClientApplication : IClientApplicationBase
         /// </remarks>
         AcquireTokenOnBehalfOfParameterBuilder AcquireTokenOnBehalfOf(IEnumerable<string> scopes, UserAssertion userAssertion);
 
+        /// <summary>
+        /// Please use <see cref="IClientApplicationBase.AcquireTokenSilent(IEnumerable{string}, IAccount)"/>.
+        /// </summary>
+        [Obsolete("Use AcquireTokenSilent(scopes, account)", true)]
+        AcquireTokenSilentParameterBuilder AcquireTokenSilent(IEnumerable<string> scopes, string loginHint);
```
pmaytak, line 82: I think this needs `new` and implementation in CCA class is not needed, similar to 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7fff414e166d6e10eb68b44802830749a91841fb/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs#L95-L97

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
```diff
@@ -75,6 +75,12 @@ public partial interface IConfidentialClientApplication : IClientApplicationBase
         /// </remarks>
         AcquireTokenOnBehalfOfParameterBuilder AcquireTokenOnBehalfOf(IEnumerable<string> scopes, UserAssertion userAssertion);
 
+        /// <summary>
+        /// Please use <see cref="IClientApplicationBase.AcquireTokenSilent(IEnumerable{string}, IAccount)"/>.
```
pmaytak, line 79: nit:
```suggestion
        /// In confidential client apps use <see cref="IClientApplicationBase.AcquireTokenSilent(IEnumerable{string}, IAccount)"/> instead.
```

pmaytak (changes requested):
See comment in the issue.

PR #3445, opened 2022-07-01
Remove more build warnings

### LINKED ISSUE #2912: [Internal] Decrease build warnings to under 5 (currently >100)
It is difficult to release because the builds produce so many warnings. We might be missing important warnings because of this.

MichelZ:
Fixes #2912

### DIFF
--- a/LibsAndSamples.sln
+++ b/LibsAndSamples.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.32428.217
+# Visual Studio Version 17
+VisualStudioVersion = 17.2.32616.157
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tests", "tests", "{9B0B5396-4D95-4C15-82ED-DC22B5A3123F}"
 EndProject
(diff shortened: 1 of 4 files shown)


LibsAndSamples.sln
```diff
@@ -185,6 +185,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Intune-xamarin-Android", "t
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Identity.Client.Broker", "src\client\Microsoft.Identity.Client.Broker\Microsoft.Identity.Client.Broker.csproj", "{6839F934-45F0-4026-8AF3-C3AEFB7D48A9}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AriaTelemetryProvider", "tests\devapps\AriaTelemetryProvider\AriaTelemetryProvider.csproj", "{A4A15A55-1106-45CB-8F4B-A15A8E8DCA50}"
```
pmaytak, line 188: @MichelZ Why is this project added?
@bgavrilMS We can probably just delete it? It's obsolete? Hasn't been used in 3+ years?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -2,7 +2,7 @@
   
   <PropertyGroup Label="Targets required for unit tests to run">
     <TargetFrameworkNetDesktop461>net461</TargetFrameworkNetDesktop461>
-    <TargetFrameworkNetCore>netcoreapp2.1</TargetFrameworkNetCore>
+    <TargetFrameworkNetCore>netcoreapp3.1</TargetFrameworkNetCore>
```
pmaytak, line 5: Why is this made? Is it because of 2.1 obsolete warning?
It's a possible breaking change, since .NET 2.1 projects will now reference the NetStandard2.0 binaries instead. There are some API breaking changes (although they might only affect NetFx). Either way, should be reverted for now.

pmaytak (changes requested):

pmaytak:
Thanks @MichelZ. Closing in favor of #3463 so CI builds run.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
MichelZ: Yes, it was because of the obsolete warning. I did not think there would be side-effects, as the label states that it's only for unit tests?

PR #3446, opened 2022-07-02
Replace obsolete KeyVault library

MichelZ:
**Changes proposed in this request**
- Replace obsolete Microsoft.Azure.KeyVault with Azure.Security.KeyVault.* libraries
- As long as the integration tests pass, this seems like a low risk PR. Only test projects changed.

**Testing**
- We'll see when Integration tests run :)

PR #3448, opened 2022-07-06
Replace obsolete KeyVault library (#3446)

pmaytak:
Merges #3446 into main after tests are run.


bgavrilMS:
Merging this, as both Peter and Bogdan reviewed it. Thanks a lot for the help @MichelZ !

PR #3450, opened 2022-07-06
Interactive cancellation

### LINKED ISSUE #3225: [Bug] Windows Broker and Embedded Web View do check cancellation tokens
Version: MSAL.NET @ 4.42
Platform: Windows 11, `net5.0-windows10.0.17763.0`
Auth flow: Broker and Interactive Web auth (embedded web view only)

**Repro**
@bgavrilMS - we have an easy repro internally and a current bugfix in the auth CLI.

```csharp
private class MsalHttpClientFactoryAdaptor : IMsalHttpClientFactory
        {
            private HttpClient instance;

            public MsalHttpClientFactoryAdaptor(HttpClient instance)
            {
                if (instance == null)
                {
                    throw new ArgumentNullException(nameof(instance));
                }

                this.instance = instance;
            }

            public HttpClient GetHttpClient()
            {
                // MSAL calls this method each time it wants to use an HTTP client.
                // We ensure we only create a single instance to avoid socket exhaustion.
                return this.instance;
            }
        }
//---

ver pca = PublicClientApplicationBuilder
    .Create(this.clientId.ToString())
    .WithAuthority(this.Authority)
    .WithBroker();

var tokenSource = new CancellationStokenSource();
var timeout = TimeSpan.FromSeconds(15);

tokenSource.CancelAfter(timeout);

await pca.AcquireTokenInteractive(scopes)
    .ExecuteAsync(tokenSource.Token);

// Expect OperationCancelledException to be thrown - but the Web or broker UI hangs forever.
```

**Expected behavior**
An `OperationCancelledException` should be thrown by calling `token.ThrowIfCancellationRequested()`.

**Possible solution**
Internally we have a workaround to run await a timer task that checks if the task is done every 100ms and after the timeout, returns a `Task.FromResult<T>(null)`.

We have debugged the un-cancellable call to this Task that runs with a custom Task Scheduler https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/96ec2757d1a18c48eb9b7abc02028a82bc601cc9/src/client/Microsoft.Identity.Client/Platforms/Features/WebView2WebUi/WebView2WebUi.cs#L74

bgavrilMS:
Fixes #3225 

**Testing**

**Performance impact**


trwalke (approved):

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
```diff
@@ -329,7 +329,7 @@ private async void atiBtn_Click(object sender, EventArgs e)
             }
         }
 
-        private async Task<AuthenticationResult> RunAtiAsync(IPublicClientApplication pca)
+        private async Task<AuthenticationResult> RunAtiAsync(IPublicClientApplication pca, CancellationToken cancellationToken)
```
pmaytak, line 332: nit: instead of passing the cancellation token to ATS and ATI methods, just create it right before ExecuteAsync?

pmaytak (approved):

PR #3457, opened 2022-07-07
Remove reference to netstndard

bgavrilMS:
Not sure why this was added (@trwalke ) . Curious to see if it builds. Maybe it needs to be added only one some select TFMs like net461 ?


pmaytak:
Locally, seems like only xamarinmac20 target needs it.
image

PR #3464, opened 2022-07-09
Obsolete AcquireTokenSilent(scopes, login_hint) in CCA

### LINKED ISSUE #3403: [Bug] AcquireTokenSilent(scopes, login_hint) is not valid on CCA

pmaytak:
Fixes #3403
Merged from #3443
Related to #2464 when `GetAccounts` was deprecated in CCA.

Obsolete AcquireTokenSilent(scopes, login_hint) in CCA. The new method in IConfidentialClientApplication interface hides the one in IClientApplicationBase. No breaking changes.

**Testing**
Tested manually.

**Performance impact**


gladjohn (approved):

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
```diff
@@ -91,8 +91,17 @@ public partial interface IConfidentialClientApplication : IClientApplicationBase
         /// </remarks>
         GetAuthorizationRequestUrlParameterBuilder GetAuthorizationRequestUrl(IEnumerable<string> scopes);
 
-        /// <inheritdoc/>
-        [Obsolete("Use GetAccountAsync in web apps and web APIs, and use a token cache serializer for better security and performance. See https://aka.ms/msal-net-cca-token-cache-serialization.")]
+        /// <summary>
+        /// In confidential client apps use <see cref="IClientApplicationBase.AcquireTokenSilent(IEnumerable{string}, IAccount)"/> instead.
+        /// </summary>
+        [Obsolete("In confidential client apps use AcquireTokenSilent(scopes, account) instead.")]
```
bgavrilMS, line 97: Should we not set the "error" flag to true? Is there any scenario where this would work?

pmaytak: Static cache/reusing CCA instance? Any test code? Easier to use this than refactoring to GetAccount(userId) since this requires creating an identifier.

bgavrilMS (approved):

PR #3504, opened 2022-07-21
Remove Newtonsoft

MichelZ:
Replace Newtonsoft Json references with Microsoft.Identity.Client.Json implementation for consistency

Not sure if that's wanted or not, but I thought it makes sense


bgavrilMS:
For test projects we probably want to use System.Text.Json ... For MSAL itself it's more complicated, but there is a separate work item tracking it and someone is looking at it.

bgavrilMS (changes requested):
We should use System.Text.Json instead of Newtonsoft if possible. A project reference is better than direct refrence for test projects, because in case of security issues we will be updated.

bgavrilMS:
Is it ok to close this? I think we can let the existing projects as they are.

MichelZ:
Sure :)

PR #3506, opened 2022-07-22
Line feed config

### LINKED ISSUE #3425: [Bug] TestSerializeAccountWithWamId Test fails due to difference in line ending
I tried cloning the repo and running the Microsoft.Identity.Test.Unit tests, in netcoreapp3.1, and a single test failed to run:
`TestSerializeAccountWithWamId`

Investigating this further, I found out that the reason for this is that this file is the only test that uses multi-line string literals.
This means that it uses the current file's CRLF definition, which is not specified by this project, neither on .gitattributes, nor or .editorconfig file. Running:
`git ls-files --eol tests\Microsoft.Identity.Test.Unit\CacheTests\CacheSerializationTests.cs`
Returns:
`i/lf    w/lf    attr/                   tests/Microsoft.Identity.Test.Unit/CacheTests/CacheSerializationTests.cs`

Which means that the repository is storing this file as LF, and the cloned version is also LF. If I open the file in Visual Studio and change the line endings to CRLF, then the test passes.

**Repro**
Clone the repo with only LF at your global git configs, run the tests.

**Expected behavior**
C# files use consistent line endings so this issue doesn't happen.

**Actual behavior**
The build will work, but one test will fail.

**Possible solution**
Configure .gitattributes and .editoconfig files so line endings are consistent.

MichelZ:
Fixes #3425

Consistent line endings. gitattributes inspired by https://github.com/dotnet/runtime/blob/main/.gitattributes

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -176,6 +176,8 @@ end_of_line = crlf
 # Code files
 [*.{cs,vb}]
 
+end_of_line = crlf
+
 # Async Threading Rules
 
 # Use Async suffix for async methods
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,15 @@
+# Set default behavior to automatically normalize line endings.
+* text=auto
+
+# Force bash scripts to always use lf line endings so that if a repo is accessed
+# in Unix via a file share from Windows, the scripts will work.
+*.in text eol=lf
+*.sh text eol=lf
+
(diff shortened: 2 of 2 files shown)


bgavrilMS (approved):

bgavrilMS:
Very nice, appreciate the effort @MichelZ

pmaytak (approved):

PR #3508, opened 2022-07-22
Update Xamarin

MichelZ:
Working on #2912 

This gets rid of some Xamarin MinPlatformVersion warnings


pmaytak (commented):
Seems like there are a few other projects that can be upgraded. Otherwise, we get package downgrade warnings.
image

bgavrilMS:
To be honest, I would not work on these. Xamarin is on its way out. Can't wait to GC Maui, wait for 1 year and remove Xamarin support (probably wishful thinking, but one is allowed to dream).

MichelZ:
> To be honest, I would not work on these. Xamarin is on its way out. Can't wait to GC Maui, wait for 1 year and remove Xamarin support (probably wishful thinking, but one is allowed to dream).

Good one :)
Things tend to stick around unfortunately. It will depend on how well MAUI works

bgavrilMS:
> 

Well, MAUI is an evolution of Xamarin, I'm can only assume Xamarin will be put in maintenance mode (not official MSFT position, I did not chat to the Maui folks about this). Especially as .NET is moving so fast nowadays.

bgavrilMS:
@MichelZ - if you want to help with some more interesting stuff, we have a label "Good First Issue" - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

Right now there are only 2 issues there, but I can add some more.

SameerK-MSFT (changes requested):
I would not upgrade Intune project. Intune SDK has some dependencies. Also, we need to support the minimum Android version of 10.0. Not below that.

MichelZ:
> I would not upgrade Intune project. Intune SDK has some dependencies. Also, we need to support the minimum Android version of 10.0. Not below that.

Readme says:
`Mono Android 9.0`

So I should just ignore that?
(I might try again to get this working)

PR #3514, opened 2022-07-23
Refactor AuthorityInfo for Uri

### LINKED ISSUE #3487: [Bug] Perf - MSAL stores authorities as strings and doing too many string operations. Try to use Uri instead.
Refactor AuthorityInfo and related code. We should not be doing so much string operations.

MichelZ:
Fixes #3487
Tried not to change too many API's at once (so some .ToString() sprinkled in)


bgavrilMS (approved):
Very nice, thanks for the effort. Hope it was not too tedious.

pmaytak (approved):

PR #3518, opened 2022-07-24
Improve build/test speed

MichelZ:
- Move from PublishBuildArtifacts / DownloadBuildArtifacts to PublishPipelineArtifacts / DownloadPipelineArtifacts per [this](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/build-artifacts?view=azure-devops&tabs=yaml) and [this](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-build-artifacts?view=azure-devops) recommendation (They are deprecated)
- Removed the second build for CacheCompat tests and use common build Output from Build stage
- Leverage Parallel Build on VSBuild


MichelZ:
@bgavrilMS Would you mind posting me the relevant test failure logs? I'm sure it's a path to something, but hard to say without the logs

bgavrilMS:
The actual build takes longer though. For example the PR build of your "Unifi/Upgrade NetFramework for Tests" took 21 min:
image


While this one took 29 min:

image


I don't understand where the `RunDesktopTest X` come from... Is it because of the parallelism  option?

MichelZ:
@bgavrilMS I think it's misleading because I restarted failed tests for the "faster" build.
This is what I got as a benchmark: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3516/checks?check_run_id=7488220808

image

Yes, it's because of this:
```
- job: 'RunDesktopTests'
  strategy:
    parallel: 5
```
But it's probably not as effective as I hoped. Would need to mainly distribute integration tests across multiple agents, which doesn't seem to be easy to configure.

bgavrilMS:
Probably a more effective way of improving the CI speed is to stop building the dev apps. It's too time consuming to keep them all up to date anyway.

MichelZ:
> Probably a more effective way of improving the CI speed is to stop building the dev apps. It's too time consuming to keep them all up to date anyway.

Hmm... that would mean that they could break though, not? Is this desired?

MichelZ:
Some more stats now after merging main to the branches:
- Refactor AuthorityInfo for Uri 3514 
  - 47m 0s
- Consolidate Packages, Update MSTest 3515  
  - 42m 56s
- Ignore WithTenantId for B2C, Refactor AADAuthority Tenant Checks 3517 
  - 44m 3s
-  Add GetTenantId 3505 
   - 43m 49s
-  Improve build/test speed 3518 
   - 29m 46s

bgavrilMS (approved):

bgavrilMS:
This is a great result, thank you so much @MichelZ

SameerK-MSFT:
Looks good. However, I would like to run this through OneBranch pipeline as there is change in some tasks and some tasks are restricted on OneBranch. Just to be safe before merging.

MichelZ:
> Looks good. However, I would like to run this through OneBranch pipeline as there is change in some tasks and some tasks are restricted on OneBranch. Just to be safe before merging.

Sure, please do

build/template-android-appcenter-tests.yaml
```diff
@@ -3,17 +3,19 @@ parameters:
...
-- task: DownloadBuildArtifacts@0
-  displayName: 'Download Drop'
+- task: DownloadPipelineArtifact@2
+  displayName: 'Download Xamarin.Android'
   inputs:
-    artifactName: drop
-    itemPattern: '**/*'
+    artifact: Xamarin.Android
+    patterns: '**/*'
+    path: $(System.ArtifactsDirectory)/Xamarin.Android
```
pmaytak, line 11: Any reason why Android is dropped into a separate folder? IOS stuff is dropped into root artifact directory.

build/template-ios-appcenter-tests.yaml
```diff
@@ -3,17 +3,19 @@ parameters:
...
 
-- task: DownloadBuildArtifacts@0
+- task: DownloadPipelineArtifact@2
   displayName: 'Download Xamarin.iOS'
   inputs:
-    artifactName: Xamarin.iOS
-    itemPattern: '**/*'
+    artifact: Xamarin.iOS
+    patterns: '**/*'
+    path: $(Build.artifactstagingdirectory)/drop
```
pmaytak, line 11: Should this be:
```suggestion
    path: $(Build.artifactstagingdirectory)/Xamarin.iOS
```

pmaytak (changes requested):

pmaytak:
@MichelZ How easy/confusing was it to figure out which templates use what other templates? Wondering if it would help to have a doc [like this](https://github.com/AzureAD/microsoft-identity-web/blob/master/build/build.md) that kind of shows the usage hierarchy?

build/template-ios-appcenter-tests.yaml
MichelZ: It probably should be, yes. There is no iOS pipeline that I can test this with, that's probably why I did not notice. Will fix, thanks!

MichelZ:
> @MichelZ How easy/confusing was it to figure out which templates use what other templates? Wondering if it would help to have a doc [like this](https://github.com/AzureAD/microsoft-identity-web/blob/master/build/build.md) that kind of shows the usage hierarchy?

@pmaytak As I do have some experience with it, it wasn't difficult for me. And I don't think many people (especially inexperienced) will mess with the build, so I'd suggest not spending too much time on this IMHO :)

pmaytak:
> > @MichelZ How easy/confusing was it to figure out which templates use what other templates? Wondering if it would help to have a doc [like this](https://github.com/AzureAD/microsoft-identity-web/blob/master/build/build.md) that kind of shows the usage hierarchy?
> 
> @pmaytak As I do have some experience with it, it wasn't difficult for me. And I don't think many people (especially inexperienced) will mess with the build, so I'd suggest not spending too much time on this IMHO :)

Good point.

build/template-android-appcenter-tests.yaml
MichelZ: Bug, as I can't start an iOS build to validate. Fixed now, thanks!

build/template-ios-appcenter-tests.yaml
pmaytak: Sounds good. Thanks. Yea, we disabled it for a while since doesn't mesh with some net6 mobile targets. We can test once we reenable it.

pmaytak (approved):
Thanks, these speed improvements are always appreciated :). 
Looks good from my side. Although I think @SameerK-MSFT wanted to take a look too.

SameerK-MSFT (changes requested):
Looks good. Since there are a lot of pipeline changes; I would make sure that it runs fine in the OneBranch before the merge. SO we do not have to revert, if it fails.

MichelZ:
> Looks good. Since there are a lot of pipeline changes; I would make sure that it runs fine in the OneBranch before the merge. SO we do not have to revert, if it fails.

Yes, please do. :)

MichelZ:
> Looks good. Since there are a lot of pipeline changes; I would make sure that it runs fine in the OneBranch before the merge. SO we do not have to revert, if it fails.

@SameerK-MSFT Were you able to test a build?

bgavrilMS:
> 

@SameerK-MSFT - did you run it through OneBranch?

SameerK-MSFT:
The build is failing in NuGet restore in Microsoft.Identity.Test.Integration.NetStandard.csproj. This would be fixed in the main by @trwalke Please merge the latest main in your branch and I will run it again to make sure that other parts work. Thanks.

MichelZ:
> The build is failing in NuGet restore in Microsoft.Identity.Test.Integration.NetStandard.csproj. This would be fixed in the main by @trwalke Please merge the latest main in your branch and I will run it again to make sure that other parts work. Thanks.

@SameerK-MSFT done

SameerK-MSFT:
Started the build with the latest. I will let you know how it goes.

SameerK-MSFT:
Ran the build twice. The first ran for approx 57 minutes and the second was approx 53 minutes. These are higher than common times that are in the low 40s. But there have been times in the low 50s and sometimes1hr. This does not break or degrade anything for OneBranch build. Since it shaves time from the PR builds, that will benefit the team efficiency. Thanks for the contribution @MichelZ

SameerK-MSFT (approved):

MichelZ:
@SameerK-MSFT Out of curiosity, isn't OneBranch a totally separate build system?
Or does it use any of these build YAMLS from the project?

bgavrilMS:
@MichelZ - it is a build system in AzureDevOps. It is more locked down than the regular ADO build system and it was designed for services, not for SDKs. It forces the use of some compliance tasks, which is nice, but otherwise it's a bit of a pain. Currently the machines there do not have VS 2022 and they do not seem to support net6 (!!!) - but we are expecting a fix for this, otherwise we might have become creative ...

Issue #3413, opened 2022-06-22
[Bug] Upgrade Newtonsoft to the latest version


SameerK-MSFT:
**Logs and network traces**
NA

**Which version of MSAL.NET are you using?**
4.44

**Platform**
Applicable to all

**issue?**
GitHub has warning to upgrade Newtonsoft. Ref: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/security/dependabot

**Possible solution**
- [x] Upgrade Newtonsoft references (PRs [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pulls?q=is%3Apr+is%3Aclosed+label%3Adependencies+newtonsoft.json))
- [x] Upgrade Newtonsoft direct source code ref

bgavrilMS:
We also need to upgrade Newtonsoft version in MSAL.

bgavrilMS:
This was alraedy done for Wilson if you need inspiration.

bgavrilMS:
@trwalke - please coordinate with @pmaytak on this one or pick something else.

PR #3521, opened 2022-07-25
Update MSAL's Newton.Json version to 13.0.1

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3413

Updating newtonsoft.Json to 13.0.1.
Updating namespace
Applying Unity fixes
Applying Netstandard 2 fix
Applying UAP fix
Setting C# lang version to `latest`

**Testing**
Unity test app

**Performance impact**

**Documentation**


MichelZ (commented):
Do you want to include non-code files like .snk, .csproj, pics, etc?

bgavrilMS (approved):

bgavrilMS:
> Do you want to include non-code files like .snk, .csproj, pics, etc?

Not sure if there is any benefit for including those, since we're building the source code as part of MSAL.

MichelZ:
> > Do you want to include non-code files like .snk, .csproj, pics, etc?
> 
> Not sure if there is any benefit for including those, since we're building the source code as part of MSAL.

Exactly. But they ARE included in this PR (I'd remove them)

pmaytak (commented):
> Applying Unity fixes
> Applying Netstandard 2 fix
> Applying UAP fix

- Can you specify what/where are the above specific changes? (Is it just a few flags?)
- Maybe helpful to add a Readme in the top `json` folder with what the version of this Json.Net code is.
- Are there any breaking changes in 13.x that affect us?

bgavrilMS:
> > Applying Unity fixes
> > Applying Netstandard 2 fix
> > Applying UAP fix
> 
> * Can you specify what/where are the above specific changes? (Is it just a few flags?)
> * Maybe helpful to add a Readme in the top `json` folder with what the version of this Json.Net code is.
> * Are there any breaking changes in 13.x that affect us?

The Unity fix in particular should ideally be re-applied as a separate commit. It will help us track it down in the future.

src/client/Microsoft.Identity.Client/json/Bson/BsonBinaryType.cs
```diff
@@ -25,6 +25,8 @@
 
 using System;
 
+#nullable disable
```
MichelZ, line 28: Try this:
In the .editorconfig, specify generated_code = true in a section that applies to that file.

This should disable nullable stuff (and a lot of other warnings)
See: https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies

> 
> Important
> 
> The global nullable context does not apply for generated code files. Under either strategy, the nullable context is disabled for any source file marked as generated. This means any APIs in generated files are not annotated. There are four ways a file is marked as generated:
> 
>     In the .editorconfig, specify generated_code = true in a section that applies to that file.
>     Put <auto-generated> or <auto-generated/> in a comment at the top of the file. It can be on any line in that comment, but the comment block must be the first element in the file.
>     Start the file name with TemporaryGeneratedFile_
>     End the file name with .designer.cs, .generated.cs, .g.cs, or .g.i.cs.
> 
> Generators can opt-in using the [#nullable](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#nullable-context) preprocessor directive.

trwalke:
> > > Applying Unity fixes
> > > Applying Netstandard 2 fix
> > > Applying UAP fix
> > 
> > 
> > 
> > * Can you specify what/where are the above specific changes? (Is it just a few flags?)
> > * Maybe helpful to add a Readme in the top `json` folder with what the version of this Json.Net code is.
> > * Are there any breaking changes in 13.x that affect us?
> 
> The Unity fix in particular should ideally be re-applied as a separate commit. It will help us track it down in the future.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3536

@bgavrilMS

src/client/Microsoft.Identity.Client/json/Newtonsoft.Json.csproj
```diff
@@ -0,0 +1,82 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
trwalke: File is not needed. Removing

trwalke:
> > Applying Unity fixes
> > Applying Netstandard 2 fix
> > Applying UAP fix
> 
> * Can you specify what/where are the above specific changes? (Is it just a few flags?)
> * Maybe helpful to add a Readme in the top `json` folder with what the version of this Json.Net code is.
> * Are there any breaking changes in 13.x that affect us?

The net standard and UAP fixes are just flags. since the fixes are the same as we have in the main branch, they do not show in the diff and nuget wont let me comment on the lines where the are.

Added readme

no breaking changes that I have seen so far. They are not mentioned in the Release Notes

@pmaytak

MichelZ:
Files are not needed:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/00bb5f47ac04d36d78a753ec63c547775b72c042/src/client/Microsoft.Identity.Client/json/packageIcon.png

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/00bb5f47ac04d36d78a753ec63c547775b72c042/src/client/Microsoft.Identity.Client/json/Dynamic.snk

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/00bb5f47ac04d36d78a753ec63c547775b72c042/src/client/Microsoft.Identity.Client/json/Newtonsoft.Json.ruleset

tests/Microsoft.Identity.Test.Unit/pop/CryptoProviderTests.cs
```diff
@@ -19,8 +19,8 @@ public void ValidateCannonicalJwk()
             string jwk = provider.CannonicalPublicKeyJwk;
             JObject jwkObj = JObject.Parse(jwk);
 
-            Assert.IsNotNull(jwkObj["E"]);
-            Assert.IsNotNull(jwkObj["N"]);
+            Assert.IsNotNull(jwkObj["e"]);
```
trwalke, line 22: I discovered this. Seems like the JObject would take a key that was case insensitive. But now that is no longer the case. Not sure if breaking change or an actual bug fix. Didn't see it mentioned in the release notes anywhere.

@pmaytak @bgavrilMS

pmaytak: afaik Newtonsoft is case-insensitive by default. So this would be a breaking change imo. Yea, the 13.x changelog is a bit sparse, but I guess this change would make sense why they upped the major version. One way to verify could be create a test project and use 12.x and 13.x packages and see what the behavior is. If you can't find any info on this, maybe open an issue in their repo? Otherwise, we might need to set this setting globally somewhere to keep the behavior consistent with how it is now. (Don't think that case-sensitivity should matter but just in case.)

pmaytak (changes requested):

tests/Microsoft.Identity.Test.Unit/pop/CryptoProviderTests.cs
pmaytak: There's this issue https://github.com/JamesNK/Newtonsoft.Json/issues/815 that suggests that it's case-insensitive. But also maybe different serialization/deserialization methods have different case sensitivity...

pmaytak (approved):

pmaytak:
UWP can be tested by building in `Release` and checking two properties in project properties - `Optimize code` and `Compile with .NET Native tool chain`:
image

PR #3538, opened 2022-07-28
MsalRuntime - init core only once.

bgavrilMS:
Fix for #3578


gladjohn:
@bgavrilMS I was able to repro the VS installer team bug and not been able to repro with this fix. Suggest we merge this so we can test more and ask VS installer to try this as well.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -299,9 +344,9 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
 
             _logger.Info($"Removing WAM Account. Correlation ID : {correlationId} ");
 
-            using (var core = new NativeInterop.Core())
+            
             {
```
gladjohn, line 348: { can be removed

gladjohn (approved):

PR #3544, opened 2022-08-01
Add Package Icon

MichelZ:
Not a fan of packages without icons...  so, behold! A PACKAGE ICON!  :)

The logo is copied from:
http://go.microsoft.com/fwlink/?LinkID=288890

This logo is used in other Microsoft projects like:
https://github.com/microsoft/OpenAPI.NET/commit/74fd5143c553758a74ddb7600fdce1b32eaed176

(The PackageIconUrl approach from the package above is deprecated, so including the icon in the package is the recommended way currently. See https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageiconurl and  https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file)


bgavrilMS:
The approach is sound, but I'm not sure if this is the icon we're suppose to use. Let me try to find out.

pmaytak:
[Azure.Identity](https://www.nuget.org/packages/Azure.Identity) is using a more specific Azure one.

MichelZ:
They do have a mix of icons in use:
https://www.nuget.org/profiles/azure-sdk?page=2
It seems like projects that start with Azure.* are favoring the Azure icon, and projects that start with Microsoft.* are favoring the Microsoft icon. (Just my 2 cents)

bgavrilMS:
I don't see any of the Identity libraries having an icon (e.g. Graph SDK) :(. We may not be able to use an icon here.

MichelZ:
> I don't see any of the Identity libraries having an icon (e.g. Graph SDK) :(. We may not be able to use an icon here.

Or maybe someone needs to get the ball rolling ;)

bgavrilMS:
@MichelZ - I'm not getting any traction on this. I think everybody is too busy than to think about icon and the default is ok. Moving to draft.

bgavrilMS:
Still no traction. I think we will have to live with default icon, if anything to be consistent with Graph SDK etc.

Issue #3533, opened 2022-07-27
[Feature Request] Add API to get Telemetry Client in Builder


neha-bhargava:
Add api to pass in the telemetry client to MSAL

Ref: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3511

PR #3545, opened 2022-08-01
Add api for telemetry client

neha-bhargava:
Fixes #
#3533 

Add api for telemetry client as experimental

**Testing**
Unit test

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,23 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
             return (T)this;
         }
 
+        /// <summary>
+        /// Sets telemetry client for the application.
```
trwalke, line 521: Any documentation that describes how to use this? Or what it is for? Or is this for internal use only?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -308,6 +308,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="6.18.0" />
```
trwalke, line 311: Did you need to update this?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public string ClientId { get; set; }
+
+        public void Initialize()
+        {
+            throw new NotImplementedException();
+        }
+
+        public bool IsEnabled()
```
trwalke, line 45: Is there no need to test these other methods in this class? are they already tested somewhere else?

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
neha-bhargava: I will add documentation here after I create it. This api is for first party only

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
neha-bhargava: Yes pulling in the ITelemetryClient from this.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
neha-bhargava: The other method implementation will follow with the next task where I am logging the telemetry. For now they are not needed.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,23 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
...
+        /// </summary>
+        /// <param name="telemetryClients">List of telemetry clients to add telemetry logs.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public T WithTelemetryClient(params ITelemetryClient[] telemetryClients)
+        {
+            ValidateUseOfExperimentalFeature("ITelemetryClient");
+
+            if ((telemetryClients != null) && telemetryClients.Any())
+            {
```
pmaytak, line 530: Thoughts on throwing if `telemetryClients` is null?

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,23 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
...
+        /// <summary>
+        /// Sets telemetry client for the application.
+        /// </summary>
+        /// <param name="telemetryClients">List of telemetry clients to add telemetry logs.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public T WithTelemetryClient(params ITelemetryClient[] telemetryClients)
+        {
+            ValidateUseOfExperimentalFeature("ITelemetryClient");
+
+            if ((telemetryClients != null) && telemetryClients.Any())
```
pmaytak, line 529:
```suggestion
            if ((telemetryClients != null) && telemetryClients.Count > 0)
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,23 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
             return (T)this;
         }
 
+        /// <summary>
+        /// Sets telemetry client for the application.
+        /// </summary>
+        /// <param name="telemetryClients">List of telemetry clients to add telemetry logs.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public T WithTelemetryClient(params ITelemetryClient[] telemetryClients)
```
pmaytak, line 525: Do we expect multiple implementations of clients? How does SAL implement this? The spec doesn't specify either.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -44,6 +44,8 @@ public string ClientVersion
             internal set { _clientVersion = string.IsNullOrWhiteSpace(value) ? DefaultClientVersion : value; }
         }
 
+        public ITelemetryClient[] TelemetryClients { get; internal set; }
```
pmaytak, line 47: Should maybe initialize this to empty array somewhere, so we don't have to check if it's null before use each time.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -0,0 +1,65 @@
...
+                .Create(TestConstants.ClientId)
+                .WithClientSecret("secret")
+                .WithTelemetryClient(telemetryClient)
+                .Build());
+
+            Assert.AreEqual(MsalError.ExperimentalFeature, e.ErrorCode);
+        }
+    }
+
+    internal class TelemetryClient : ITelemetryClient
```
pmaytak, line 31: nit: > something like `TestTelemetryClient` or you can just pass a mocked `ITelemetryClient`.

pmaytak (approved):

TimHannMSFT:
Thanks Neha, I likely won't get the chance to take a look today but plan to block some time for a review tomorrow.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
bgavrilMS: Initialize to empty? So that `foreach` just jumps over it?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
bgavrilMS: Agreed. It doesn't seem like you are doing very complicated stuff with this implementation, so using `Substitute.For<ITelemetryClient>` is probably enough.

bgavrilMS (approved):
Approving, but would like @TimHannMSFT 's opinion on the public API. We should have the same entry point, i.e. a collection of `ITelemetryClient`. Maybe we should use the same type of collection too? Array or IEnumerable or something.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
TimHannMSFT: The SAL implementation will currently throw an ArgNull exception in this case as another data point.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
TimHannMSFT: The idea behind multiple clients as an extensibility model is that say substrate or teams can pass their own ITelemetryClient implementation that can decide which events it cares about and send them back to whatever backend they want. They might decide to take just one event type and push that to their Geneva tenant. That won't interfere with the common telemetry client we implement for 1P partners which emits to our shared store.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
TimHannMSFT: SAL also does the initialize to empty as an additional data point.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
TimHannMSFT: Does MSAL leverage Moq at all? seems like that could be useful here.

pmaytak: We use NSubstitute.

TimHannMSFT:
> Approving, but would like @TimHannMSFT 's opinion on the public API. We should have the same entry point, i.e. a collection of `ITelemetryClient`. Maybe we should use the same type of collection too? Array or IEnumerable or something.

For middleware libraries we don't have any DI concepts so the entry point is new constructors which take in a single ITelemetryClient or an IEnumerable<ITelemetryClient> (either way it's added to the internal set which is initialized to empty by default.)

TimHannMSFT (approved):
Looks like this change merely allows an entry point to add an ITelemetryClient to the MSAL instance. I'm assuming a later change will be added which will actually instrument some parts of MSAL to send events to the ITelemetryClient if enabled for the respective client?

I think that's fine to do, perhaps clarifying in the release notes somewhere that this change merely adds the foundation for future capabilities and isn't meant to add any new capabilities on its own.

pmaytak:
Yea, I think this work was just split into more manageable pieces.
My 2 cents - I would only merge this and other related PRs when the full feature is ready.

bgavrilMS:
> Looks like this change merely allows an entry point to add an ITelemetryClient to the MSAL instance. I'm assuming a later change will be added which will actually instrument some parts of MSAL to send events to the ITelemetryClient if enabled for the respective client?
> 
> I think that's fine to do, perhaps clarifying in the release notes somewhere that this change merely adds the foundation for future capabilities and isn't meant to add any new capabilities on its own.

Yes, correct, this just adds a new public API but no data collection occurs.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
neha-bhargava: I am adding implementation to this in the next PR to test what is logged. So leaving this as is.

neha-bhargava:
> rely allows an entry point to add an ITelemetryClient to the MSAL instance. I'm assuming a later change will be added which will actually instrument some parts of MSAL to send events to the ITelemetryClient if enabled for the respective client?
> 
> I think that's fine to do, perhaps clarifying in the release notes somewhere that this change merely adds the foundation for future capabilities and isn't meant to add any new capabilities on its own.

The follow up PR is ready for review. I will address these comments and merge it in the other PR. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3574

neha-bhargava:
All the comments here are addressed and I have merged this PR into https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3574. Closing this one.

Issue #3534, opened 2022-07-27
[Feature Request] Log data to telemetry pipeline


neha-bhargava:
Add logic to log data to telemetry.

Currently for acquire_token following data points are included:

| AttributeName | Log for MSAL |
| ----------------- | --------------- |
| EventInfo_Time | Logged by client |
| DeviceInfo_OsVersion | Logged by client |
| DeviceInfo_OsName | Logged by client |
| PipelineInfo_ClientCountry | Logged by client |
| Succeeded | Log in MSAL |
| PopToken | Log in MSAL |
| SentByS2STelemetryClient | Logged by client |
| Duration | Log in MSAL |
| RemainingLifetime | Log in MSAL |
| ActivityId |
| ClientId | Log in MSAL |
| CacheInfoTelemetry |
| TokenSource | Log in MSAL |
| Resource |

Additionally we can add cache and http latency.

Ref: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3511

PR #3574, opened 2022-08-08
Nebharg/log telemetry

neha-bhargava:
Fixes #
#3534 

Add logs to telemetry pipeline

**Testing**
Unit tests

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -128,8 +127,8 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
 
         private void LogSuccessfulTelemetry(AuthenticationResult authenticationResult, MsalTelemetryEventDetails telemetryEventDetails)
         {
-            telemetryEventDetails.SetProperty(TelemetryConstants.CacheInfoTelemetry, authenticationResult.AuthenticationResultMetadata.CacheRefreshReason.ToString());
-            telemetryEventDetails.SetProperty(TelemetryConstants.TokenSource, authenticationResult.AuthenticationResultMetadata.TokenSource.ToString());
+            telemetryEventDetails.SetProperty(TelemetryConstants.CacheInfoTelemetry, Convert.ToInt64(authenticationResult.AuthenticationResultMetadata.CacheRefreshReason));
```
bgavrilMS, line 130: why int64 (i.e. long) and not int32?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -13,6 +18,24 @@ namespace Microsoft.Identity.Test.Unit.PublicApiTests
...
+        public override void TestInitialize()
+        {
+            _telemetryClient = new TelemetryClient(TestConstants.ClientId);
+            base.TestInitialize();
+        }
+
+        [TestCleanup] 
+        public override void TestCleanup()
+        {
+            _harness?.Dispose();
```
bgavrilMS, line 35: The harness should not be shared between tests, becasue the each test should consume all HTTP message from it.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        /// <summary>
+        /// Gets a set of <see cref="ITelemetryClient"/>s which are enabled for a given <paramref name="eventName"/>.
+        /// </summary>
+        /// <param name="clients">Set of clients to check.</param>
+        /// <param name="eventName">Event name to evaluate.</param>
+        /// <returns>The relevant set of <see cref="ITelemetryClient"/>s.</returns>
+        internal static IEnumerable<ITelemetryClient> GetEnabledClients(this IEnumerable<ITelemetryClient> clients, string eventName)
+        {
+            foreach (var client in clients)
```
bgavrilMS, line 22: nit: the Linq way of doing it is: 

```csharp
return clients?.Where(c => c.IsEnabled(eventName))
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -104,9 +108,35 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                     AuthenticationRequestParameters.RequestContext.Logger.ErrorPii(ex);
                     throw;
                 }
+                finally
+                {
+                    TrackTelemetryIfNecessary(
```
bgavrilMS, line 113: Since we have 2 telemetry systems, I suggest using more descriptive naming (or comments) to differentiate.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -104,9 +108,35 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+                        AuthenticationRequestParameters.RequestContext.ServiceBundle.Config.TelemetryClients, 
+                        telemetryEventDetails);
+                }
             }
         }
 
+        private void TrackTelemetryIfNecessary(
+            IEnumerable<ITelemetryClient> telemetryClients, MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            var relevantClients = telemetryClients.GetEnabledClients(TelemetryConstants.AcquireTokenEventName);
```
bgavrilMS, line 123: nit: whole thing could be an extension method

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: The method to set property does not have an iverload for int32. Only for long. Looks like a limitation of Aria SDK. @TimHannMSFT can confirm this.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -44,6 +44,8 @@ public string ClientVersion
             internal set { _clientVersion = string.IsNullOrWhiteSpace(value) ? DefaultClientVersion : value; }
         }
 
+        public ITelemetryClient[] TelemetryClients { get; internal set; }
```
TimHannMSFT, line 47: >public ITelemetryClient[] TelemetryClients { get; internal set; } [](http://example.com/codeflow?start=8&length=65)

init to Empty set? having a public setter will let any caller modify the set of telemetry clients, is that intentional?

What's the library stance on property headers?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
TimHannMSFT: Yes, the AriaSDK only supports long natively which we decided to reflect in the basic contract.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava: Actually changed this on the other PR and pulled in changes from there. So this is updated and initialized.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -104,9 +113,25 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                     AuthenticationRequestParameters.RequestContext.Logger.ErrorPii(ex);
                     throw;
                 }
+                finally
+                {
+                    telemetryClients?.TrackEvent(telemetryEventDetails, TelemetryConstants.AcquireTokenEventName);
```
TimHannMSFT, line 118: >telemetryClients?.TrackEvent(telemetryEventDetails, TelemetryConstants.AcquireTokenEventName); [](http://example.com/codeflow?start=20&length=94)

I would think you'd only want to call TrackEvent IIF at least one telemetry client was enabled for the given event. Take a look how SAL does this (first checking if there are any enabled clients then, only for those clients passing the event along to it.)

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -104,9 +113,25 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+                }
             }
         }
 
+        private void LogSuccessfulTelemetry(AuthenticationResult authenticationResult, MsalTelemetryEventDetails telemetryEventDetails)
```
TimHannMSFT, line 123: >LogSuccessfulTelemetry [](http://example.com/codeflow?start=21&length=22)

take a look at how this is done in SAL, to my eye it's much more readable and clear what's going on. Perhaps just a style thing (S2STelemetryEventDetails).

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            return clients?.Where(c => c.IsEnabled(eventName));
+        }
+
+        /// <summary>
+        /// Sends the same input events to each telemetry client.
+        /// </summary>
+        /// <param name="clients">Clients to emit telemetry to.</param>
+        /// <param name="eventDetails">Telemetry details to emit.</param>
+        internal static void TrackEvent(this IEnumerable<ITelemetryClient> clients, TelemetryEventDetails eventDetails, string eventName)
```
TimHannMSFT, line 31: >string eventName [](http://example.com/codeflow?start=120&length=16)

Why does this take in eventName? TelemetryEventDetails already contains that.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            foreach (var client in clients.GetEnabledClients(eventName))
```
TimHannMSFT, line 33: >clients.GetEnabledClients(eventName) [](http://example.com/codeflow?start=35&length=36)

If you collapse the enabled check into this method, then you'll always execute the work to build out the event details no matter if any clients have that event enabled.

Better to separately check if any client is enabled for the event THEN create the details and pass it to the enabled clients.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        public bool IsEnabled()
+        {
+            return true;
+        }
+
+        public bool IsEnabled(string eventName)
+        {
+            return TelemetryConstants.AcquireTokenEventName.Equals(eventName) ? true : false;
```
TimHannMSFT, line 133: >? true : false; [](http://example.com/codeflow?start=78&length=15)

no need to add the extra ternary statement, Equals already returns what you want right?

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,23 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
             return (T)this;
         }
 
+        /// <summary>
+        /// Sets telemetry client for the application.
+        /// </summary>
+        /// <param name="telemetryClients">List of telemetry clients to add telemetry logs.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public T WithTelemetryClient(params ITelemetryClient[] telemetryClients)
```
TimHannMSFT, line 525: >WithTelemetryClient [](http://example.com/codeflow?start=17&length=19)

SAL doesn't expect the client to be initialized before being passed in. I think you'd want to call initialize here for each client. It's idempotent so if it's already initialized it's a no-op.

TimHannMSFT (approved):
:shipit:

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClientExtensions.cs
neha-bhargava: I am building the event details separately and not when logging. The idea is to have the event details populated as we go through the request and in future when we add the error details or L1/L2 cache details we might have to pass the event details object around. See [this](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/nebharg%2FLogTelemetry/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs#L123)

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -95,9 +95,9 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                     UpdateTelemetry(sw, apiEvent, authenticationResult);
                     LogMetricsFromAuthResult(authenticationResult, AuthenticationRequestParameters.RequestContext.Logger);
 
-                    if (telemetryClients != null && telemetryClients.Count() > 0)
+                    if (telemetryClients.GetEnabledClients(TelemetryConstants.AcquireTokenEventName).Count() > 0)
```
bgavrilMS, line 98: nit: put the entire if block in the private method.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -528,6 +528,11 @@ public T WithTelemetryClient(params ITelemetryClient[] telemetryClients)
 
             if ((telemetryClients != null) && telemetryClients.Any())
             {
+                foreach (var telemetryClient in telemetryClients)
```
bgavrilMS, line 531: The `params` will automatically create a non-null array for you. It will have 0 elements if the user does `WithTelemetryClient()`.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -528,6 +528,11 @@ public T WithTelemetryClient(params ITelemetryClient[] telemetryClients)
 
             if ((telemetryClients != null) && telemetryClients.Any())
```
bgavrilMS, line 529: Better to use `Length` on arrays.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        /// <summary>
+        /// Sends the same input events to each telemetry client.
+        /// </summary>
+        /// <param name="clients">Clients to emit telemetry to.</param>
+        /// <param name="eventDetails">Telemetry details to emit.</param>
+        internal static void TrackEvent(this IEnumerable<ITelemetryClient> clients, TelemetryEventDetails eventDetails)
+        {
+            foreach (var client in clients.GetEnabledClients(eventDetails.Name))
```
pmaytak, line 33: In this code, when  `GetEnabledClients` is called, since it's an IEnumerable, a new list will be created behind the scenes with only enabled clients, over which we then iterate.

Suggestion is to iterate over the original `clients` collection, then check if that client is enabled, track the event. This saves an allocation.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.TelemetryCore
+{
+    /// <summary>
+    /// Extension methods used for telemetry client.
+    /// </summary>
+    internal static class TelemetryClientExtensions
```
pmaytak, line 13: For the two methods in this class I would change them to extend an `ITelemetryClient[]`.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -0,0 +1,146 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.PublicApiTests
+{
+    [TestClass]
+    public class TelemetryClientTests : TestBase
+    {
+        private MockHttpAndServiceBundle _harness;
+        private PublicClientApplication _pca;
+        private ITelemetryClient _telemetryClient;
```
pmaytak, line 23: To eliminate the need for casting later
```suggestion
        private TelemetryClient _telemetryClient;
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -0,0 +1,146 @@
...
+using Microsoft.Identity.Test.Common.Core.Helpers;
+using Microsoft.Identity.Test.Common.Core.Mocks;
+using Microsoft.Identity.Test.Common.Mocks;
+using Microsoft.IdentityModel.Abstractions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.PublicApiTests
+{
+    [TestClass]
+    public class TelemetryClientTests : TestBase
```
pmaytak, line 19: Additional test cases for the public API could be:
`.WithTelemetryClient()` => Empty params array - should work
`.WithTelemetryClient(null)` => should throw ArgumentNullEx
`.WithTelemetryClient(telemetryClient, null)` => should throw ArgumentNullEx

pmaytak (changes requested):

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,38 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
...
+            if (telemetryClients != null && telemetryClients.Length > 0)
+            {
+                foreach (var telemetryClient in telemetryClients)
+                {
+                    if (telemetryClient == null)
+                    { 
+                        throw new ArgumentNullException(nameof(telemetryClient));
+                    }
+
+                    telemetryClient.Initialize();
```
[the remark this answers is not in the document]
neha-bhargava: The Initialize call is idempotent, so if it is already initialized its no-op. [See](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3574/files/b230cfe842ec89d7614705a331725125c68d27f3#r943982390)

src/client/Microsoft.Identity.Client/TelemetryCore/MsalTelemetryEventDetails.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.TelemetryCore
+{
+    internal class MsalTelemetryEventDetails: TelemetryEventDetails
```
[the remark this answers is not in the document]
neha-bhargava: Yeah was thinking of putting it in another folder. I can move it.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -15,5 +15,20 @@ internal static class TelemetryConstants
         public const string Zero = "0";
         public const string CommaDelimiter = ",";
         public const string PlatformFields = "platform_fields";
+
+#region Telemetry Client Constants
```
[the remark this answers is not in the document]
neha-bhargava: Or I can move the constants to a different file in the TelemetryClient folder to keep them separate.

pmaytak (approved):

PR #3593, opened 2022-08-11
Automatically handle the case when `loginHint`== OperatingSystemAccount

### LINKED ISSUE #3200: [Feature Request] AcquireTokenByIntegratedWindowsAuthentication + WAM should just log in the default user
AcquireTokenSilent(PublicClientApplication.OperatingSystemAccount)

johnterickson:
Fixes #3200

If the loginHint is the current user, then don't just check the cache - actually use the current user account.

**Testing**
I ran the broker test app.

**Performance impact**
I didn't do perf testing.

**Documentation**
I'm not sure which documentation should be updated.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
@@ -21,6 +21,10 @@ namespace Microsoft.Identity.Client.Internal.Requests.Silent
         private readonly Lazy<ISilentAuthRequestStrategy> _brokerStrategyLazy;
         private readonly ILoggerAdapter _logger;
 
+        private bool IsBrokerConfigured => 
+            AuthenticationRequestParameters.AppConfig.IsBrokerEnabled &&
+            ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
+
         public SilentRequest(
             IServiceBundle serviceBundle,
             AuthenticationRequestParameters authenticationRequestParameters,
@@ -46,10 +50,7 @@ namespace Microsoft.Identity.Client.Internal.Requests.Silent
         {
             await UpdateRequestWithAccountAsync().ConfigureAwait(false);
 
-            bool isBrokerConfigured = AuthenticationRequestParameters.AppConfig.IsBrokerEnabled &&
-                                      ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
-
-            if (isBrokerConfigured && AuthenticationRequestParameters.PopAuthenticationConfiguration != null)
+            if (IsBrokerConfigured && AuthenticationRequestParameters.PopAuthenticationConfiguration != null)
             {
                 _logger.Info("[Silent Request] Attempting to use broker instead of searching local cache for Proof-of-Possession tokens. ");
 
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -138,11 +139,32 @@ private async Task<IAccount> GetSingleAccountForLoginHintAsync(string loginHint)
         {
             if (!string.IsNullOrEmpty(loginHint))
             {
-                IReadOnlyList<IAccount> accounts = (await CacheManager.GetAccountsAsync().ConfigureAwait(false))
+                IList<IAccount> accounts = (await CacheManager.GetAccountsAsync().ConfigureAwait(false))
                     .Where(a => !string.IsNullOrWhiteSpace(a.Username) &&
                            a.Username.Equals(loginHint, StringComparison.OrdinalIgnoreCase))
                     .ToList();
 
+                if (accounts.Count == 0 && DesktopOsHelper.IsWindows())
```
bgavrilMS, line 147: Interesting idea. But what is the point here? To get AcquireTokenSilent + WAM to succeed when login_hint matches UPN of the Windows user?

I think this logic would need to move to the broker

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -138,11 +139,32 @@ private async Task<IAccount> GetSingleAccountForLoginHintAsync(string loginHint)
...
+                        currentUpn = await ServiceBundle.PlatformProxy.GetUserPrincipalNameAsync();
+                    }
+                    catch
+                    {
+                        // NetDesktopPlatformProxy throws instead of returning string.Empty.
+                        currentUpn = null;
+                    }
+
+
+                    if (currentUpn == loginHint)
```
bgavrilMS, line 162: also if `string.IsNotNullOrEmpty(currentUpn)`, otherwise this might login the current OS account on .NET Classic.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
bgavrilMS, line 147: we'd probably need to add `if brokerEnabled`

bgavrilMS: Can you please open a bug or a feature request to tie this PR to ?

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
johnterickson: This shouldn't happen as this is all under `if (!string.IsNullOrEmpty(loginHint))` but it doesn't hurt to make the code clearer :)

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
johnterickson: Should I do this *instead of* the OS check?

bgavrilMS:
I'm ok with this, but one of us will need to write some unit test

bgavrilMS:
Hi @johnterickson - we expect this scenario to work without having to make this change (new broker only https://aka.ms/msal-net-wam). If it doesn't, we should get a bug logged.

CC @gladjohn

PR #3601, opened 2022-08-13
Use Release configuration.

pmaytak:
In previous PR the build configurations were refactored, the config was changed from Debug to Release in some places. So I updated to use Release everywhere. Plus my guess is maybe it'll make the builds faster too?

### DIFF
--- a/build/pipeline-ci.yaml
+++ b/build/pipeline-ci.yaml
@@ -8,7 +8,7 @@ pr: none
 
 variables:
   BuildPlatform: 'any cpu'
-  BuildConfiguration: 'debug'
+  BuildConfiguration: 'Release'
   #intended for running all ui automation tests as one to reduce time
   ConsolidateAppCenterTests: true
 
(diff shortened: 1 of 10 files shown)


bgavrilMS (approved):

trwalke:
what is the reason for this?

pmaytak:
> what is the reason for this?

@trwalke In previous PR where Bogdan refactored the build configurations,  the config was changed from Debug to Release in some places. So I updated to use Release everywhere. Plus my guess is maybe it'll make the builds faster too?

gladjohn:
Should the android failure be addressed along with this? would be nice to pass build configuration parameter so we do not hardcode debug|Release in test path

gladjohn (approved):

neha-bhargava (approved):

trwalke (approved):

pmaytak:
> Should the android failure be addressed along with this? would be nice to pass build configuration parameter so we do not hardcode debug|Release in test path

Yeah, but I just wanted to quickly fix the build without refactoring anything.

PR #3606, opened 2022-08-16
Compile MAUI

SameerK-MSFT:
Fixes #
Builds MAUI solution

Added Libs.sln file. Tmpelates to compile lIbs.sln

**Testing**
Manual

**Performance impact**
NA

**Documentation**
NA


LIbs/LIbs.sln
```diff
@@ -0,0 +1,28 @@
+﻿
```
MichelZ, line 1: Nit: Directory/Filename casing is wrong (LIbs vs Libs)

bgavrilMS:
@SameerK-MSFT - is this still relevant?

Issue #2877, opened 2021-09-08
[Feature Request] Provide ability to turn off the default retry-once policy and recommend retry policy


rymeskar:
By default, MSAL retries http calls once on 5xx with [a delay of 1 second](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ef3b2f61d9f0d441a1d2232ca2062a8d60e13437/src/client/Microsoft.Identity.Client/Http/HttpManager.cs#L162). 

**Scenario**
In expert and high traffic scenarios, customers might want to provide own retry strategies, retry periods and resilient http client strategies (circuit breaker).
The behavior within MSAL then intercepts with the logic that the clients might want to inject through [the WithHttpClientFactory](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.abstractapplicationbuilder-1.withhttpclientfactory?view=azure-dotnet) configuration.

**Proposal**
Provide a knob to turn off the retry-once behavior of MSAL.

Note: see larger issue tracking this: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3561

**Design**

While we could add a complex extensiblity point such as the one described [here](https://gist.github.com/mjcheetham/44880df497c147b1f90a73fe86f8a3b7), we recognize that retry libraries like Poly work perfectly fine at HttpClient level and Msal already exposes this layer. So the proposal is to add an overload to `WithHttpClientFactory`:

```WithHttpClientFactory(IMsalHttpClientFactory factory, bool disableDefaultRetryPolicy)```

We should provide docs on how to add retry policy for HTTP.

bgavrilMS:
CC @trwalke

bgavrilMS:
We will remove the default retry policy completely, as there are concenrns about retry storms. Apps should deal with retries.

gladjohn:
MSAL 4.47.0 has been released, This issue is now fixed.

PR #3608, opened 2022-08-16
Adding api to disable retry on 5xx errors

trwalke:
Fixes #
fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2877

Adding api to disable retry on 5xx error.

**Testing**
unit test

**Performance impact**
N/A

**Documentation**


src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,19 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
             return (T)this;
         }
 
+        /// <summary>
+        /// Configures MSAL to retry on 5xx server errors. When enabled (on by default), MSAL will wait 1 second after recieving a 5xx error and then retry the
+        /// http request again.
+        /// </summary>
+        /// <param name="shouldRetry">Determines if MSAL should retry on 5xx errors.</param>
+        /// <returns></returns>
+        public T WithRetryOnServerErrors(bool shouldRetry)
```
bgavrilMS, line 526: @jmprieur for new public API review

tests/Microsoft.Identity.Test.Unit/AppConfigTests/PublicClientApplicationBuilderTests.cs
```diff
@@ -696,5 +696,11 @@ public void IsBrokerAvailable_NoAuthorityInBuilder()
             Assert.IsTrue(builder1.IsBrokerAvailable());
 #endif
         }
+
+        [TestMethod]
+        public void RetryOn5xxApiTest()
+        {
+            //WithRetryOnServerErrors
```
bgavrilMS, line 703: ?

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,19 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
             return (T)this;
         }
 
+        /// <summary>
+        /// Configures MSAL to retry on 5xx server errors. When enabled (on by default), MSAL will wait 1 second after recieving a 5xx error and then retry the
+        /// http request again.
+        /// </summary>
+        /// <param name="shouldRetry">Determines if MSAL should retry on 5xx errors.</param>
+        /// <returns></returns>
```
bgavrilMS, line 525: remove

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -517,6 +517,19 @@ public T WithTelemetry(ITelemetryConfig telemetryConfig)
             return (T)this;
         }
 
+        /// <summary>
+        /// Configures MSAL to retry on 5xx server errors. When enabled (on by default), MSAL will wait 1 second after recieving a 5xx error and then retry the
```
bgavrilMS, line 521: Pls mention that the default is true

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS, line 526: Was thinking more of `WithRetryOnIdentityProviderErrors(bool retry5xxErrors)`

pmaytak: I'm feelin like it would be good to spec out the whole retry feature so that we do this holistically. We discussed having some retry adapter or something that maybe user can pass.
Or I was thinking something like `.WithRetryOptions(retryOptions)`, then `RetryOptions` can have something like `RetryOptions.Retry5xx`, `RetryOptions.Retry418`, `RetryOptions.RetryTaskCancellationEx`, `RetryOptions.RetryDelay`, `RetryOptions.RetryMaxCount`.

I understand it'll take more time, but I think it'll help us deliver a  better feature holistically.

bgavrilMS: Yes, I was thinking about a more general-purpose API as well, which could include cache retries, retries on HTTP errors etc. At the same time, it's better if retries were handled externally, where specialized libraries like Poly can provide the golden standard of retries, i.e. exponential backoff. 

Alternative to not add a new API which we might never extend, is to reuse `WithHttpClientFactory(IHttpClientFactory factory, RetryOptions retryOptions)`.

Thoughts? 

CC: @jmprieur and @rymeskar

bgavrilMS: Here's an idea based on feedback from @mjcheetham 

```csharp

WithRetryStrategy(IHttpRetryStrategy httpRetryStrategy);

// Return null to not retry. Return a timestap to retry after waiting for it.
interface IHttpRetryStrategy
{
    // Invoked when the HTTP response is 4xx or 5xx. 
    TimeStamp? OnHttpResponseFailed(HttpResponse httpResponse4xx5xx);

    // Invoked on exceptions
    TimeStamp? OnHttpException(Exception ex);

```

trwalke: I think we want to avoid providing retries on all 4xx since ESTS already has a mechanism for 425 with the retry after header that the user is expected to handle. Seems like the new implementations may potentially class with it.

bgavrilMS: It's a fair point, we need to document 4xx and especially 429 TooManyRetries

Alternative proposal from @mjcheetham  https://gist.github.com/mjcheetham/44880df497c147b1f90a73fe86f8a3b7

bgavrilMS: Discussed this with @jmprieur as well, and the perspective is:

While we could add a complex extensiblity point such as the one described [here](https://gist.github.com/mjcheetham/44880df497c147b1f90a73fe86f8a3b7), we recognize that retry libraries like Poly work perfectly fine at HttpClient level and Msal already exposes this layer. So the proposal is to add an overload to `WithHttpClientFactory`:

```WithHttpClientFactory(IMsalHttpClientFactory factory, bool disableDefaultRetryPolicy)```

Comment (SameerK-MSFT) [reply src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs:526]: It sounds right to simplify our API and let retry libraries do the job. But we need to document this well as we are getting several questions regarding it.

**Comment by bgavrilMS** [comment]
.WithHttpClientFactory(IHttpClientFactory, bool retryOn5xx /* true by default /* )

File: src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs (hunk 1 above)
Comment (jmprieur) [inline src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs:526]: in the end did we decide not to add it to:

```csharp
WithHttpClientFactory(IHttpClientFactory factory, RetryOptions retryOptions).
```

@bgavrilMS: I thought that's what we agreed on? Do I miss something?

bgavrilMS: We decided to not provide a complex retry option and to focus only on: 

```csharp
WithHttpClientFactory(IHttpClientFactory factory, bool retryOnceOn5xx=true).
```

Or did you want to have a `RetryOptions` object with a single `bool RetryOnceOn5xx` property, as it would be easier to extend?

rymeskar: Can you also provide an MSAL extensibility point (separate package) where you would provide the default poly implementation/configuration of what you think are the right values?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -139,7 +143,7 @@ private HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string
                     (int)response.StatusCode, response.StatusCode));
 
                 is5xxError = (int)response.StatusCode >= 500 && (int)response.StatusCode < 600;
-                isRetryable = is5xxError && !HasRetryAfterHeader(response);
+                isRetryable = is5xxError && !HasRetryAfterHeader(response) && _retryConfig;
```
rymeskar, line 146: Nit:
might be more optimized to go from the lease expensive bool checks (_retryConfig) to the most expensive checks (HasRetryAfterHeader).

But don't have strong opinions as the effect is fairly negligible.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -37,16 +37,20 @@ internal AbstractApplicationBuilder(ApplicationConfiguration configuration)
...
         /// <remarks>MSAL does not guarantee that it will not modify the HttpClient, for example by adding new headers.
         /// Prior to the changes needed in order to make MSAL's httpClients thread safe (https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2046/files),
         /// the httpClient had the possibility of throwing an exception stating "Properties can only be modified before sending the first request".
         /// MSAL's httpClient will no longer throw this exception after 4.19.0 (https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.19.0)
         /// see (https://aka.ms/msal-httpclient-info) for more information.
+        /// If you only want to configure the retryOnceOn5xx parameter, set httpClientFactory to null and MSAL will use the default Http client.
         /// </remarks>
         /// <returns>The builder to chain the .With methods</returns>
-        public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
+        public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory, bool retryOnceOn5xx = true)
```
bgavrilMS, line 50: I believe we said in the past that we'd add new overloads instead of adding new default params.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -25,12 +25,16 @@ namespace Microsoft.Identity.Client.Http
     internal class HttpManager : IHttpManager
     {
         private readonly IMsalHttpClientFactory _httpClientFactory;
+        //Determines whether or not to retry on 5xx errors. Configurable on application creation. default is true;
+        private readonly bool _retryConfig;
         public long LastRequestDurationInMs { get; private set; }
 
-        public HttpManager(IMsalHttpClientFactory httpClientFactory)
+        public HttpManager(IMsalHttpClientFactory httpClientFactory, bool retry = true)
```
jmprieur, line 32: why have a default value, given it's an internal method?

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
```diff
@@ -24,8 +24,8 @@ internal sealed class MockHttpManager : HttpManager,
         private readonly TestContext _testContext;
         public Func<MockHttpMessageHandler> MessageHandlerFunc;
 
-        public MockHttpManager(TestContext testContext = null) :
-            base(new SimpleHttpClientFactory())
+        public MockHttpManager(TestContext testContext = null, bool retry = true) :
```
jmprieur, line 27: same thing

jmprieur (approved):

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
trwalke: Yes I think it makes sense. I was following what was in this comment so I wasn't sure.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3608#discussion_r953699033

I can update

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: You are probably in a better position to advise about good retry policies @rymeskar, since you work with high scale services. We'd take your advice as recommendation for others.

bgavrilMS (approved):

PR #3623, opened 2022-08-23
Update Microsoft.Identity.Client.NativeInterop reference to 0.12.1

### LINKED ISSUE #3612: [Bug] MSAL.NativeInterop only targets .net fwk
It looks like since we introduced the new target framework for MSAL.NativeInterop, we stopped packing the netstandard version of the managed DLL and only pack the NetFwk 4.6.1 version. You can use ILSpy to observe this:

image

This prevents PowerShell from updating past 0.9.0 

Possible cause: a bug in the packaging logic

gladjohn:
Fixes #3612 

Fixes a bug where netstandard2.0 support was dropped for Interop. [See 3067](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3067)

**Testing**
Dev Apps

**Performance impact**
None

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -245,7 +245,7 @@ namespace Microsoft.Identity.Client.Broker
                     ExpiresIn = (long)(DateTime.SpecifyKind(authResult.ExpiresOn, DateTimeKind.Utc) - DateTimeOffset.UtcNow).TotalSeconds,
                     ClientInfo = authResult.Account.ClientInfo.ToString(),
                     TokenType = authResult.IsPopAuthorization ? Constants.PoPAuthHeaderPrefix : BrokerResponseConst.Bearer,
-                    WamAccountId = authResult.Account.Id,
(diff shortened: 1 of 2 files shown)


bgavrilMS (approved):

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -245,7 +245,7 @@ private static void AddPopParams(AuthenticationRequestParameters authenticationR
                     ExpiresIn = (long)(DateTime.SpecifyKind(authResult.ExpiresOn, DateTimeKind.Utc) - DateTimeOffset.UtcNow).TotalSeconds,
                     ClientInfo = authResult.Account.ClientInfo.ToString(),
                     TokenType = authResult.IsPopAuthorization ? Constants.PoPAuthHeaderPrefix : BrokerResponseConst.Bearer,
-                    WamAccountId = authResult.Account.Id,
+                    WamAccountId = authResult.Account.AccountId,
```
bgavrilMS, line 248: nit: In the future, would be good to maintain backwards compat to avoid problems if ppl decide to upgarde only 1 package.

PR #3626, opened 2022-08-25
Use common  HTTP Client on .NET Fx.

### LINKED ISSUE #3546: [Bug] Provide better HttpClient factory on .NET fwk (use the factory from .NET Core)
Looks like the .NET fwk factory is throwing a lot of timeout errors: 

image

Suggested fix: use .NET Core's factory on .net

pmaytak:
Fixes #3546

- On .NET Fx targets, use the common HTTP Client which is used on NET Core, (which has better performance and is more resilient).

**Testing**
Existing tests. Tested manually.

**Performance impact**
See related issue. The client should throw less exceptions.

**Documentation**
N/A


trwalke (approved):
So this will use the default abstract virtual method which initializes the simple cleint factory right? LGTM

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopPlatformProxy.cs
```diff
@@ -245,11 +244,6 @@ public override IDeviceAuthManager CreateDeviceAuthManager()
             return new NetDesktopDeviceAuthManager();
         }
 
-        public override IMsalHttpClientFactory CreateDefaultHttpClientFactory()
-        {
-            return new NetDesktopHttpClientFactory();
```
bgavrilMS, line 250: Are the associated `NetDesktopHttpClientFactory` and dependencies classes used anywhere? If not, please delete the dead code.

bgavrilMS (approved):

pmaytak:
> So this will use the default abstract virtual method which initializes the simple client factory right? LGTM

Yes.

bgavrilMS:
Merging so that it catches 4.46.2 release

PR #3641, opened 2022-08-30
Fix for #3637 - use correct authority in multi-cloud + WAM scenario

### LINKED ISSUE #3637: [Bug] Incorrect authority in AuthResult using instance_aware
I used the below mentioned configuration and tried logging into a non-public cloud. I got logged in, successfully got access token as well, everything is correct Except Authority is not of respective cloud. It seems library is overriding the authority it gets in their result.

Also, when I enable instance_aware manually using query params it works fine except authority but when I use WithMultiCloudSupport(true) it doesn't work.

MSAL.NET 4.46.0

**Platform**
.NET Framework: 4.7

* Desktop
    * Integrated Windows Authentication

The app is in production, I haven't upgraded MSAL, but started seeing this issue.

**Repro**
1.Use this configuration of WAM
```csharp
return PublicClientApplicationBuilder.Create(clientId)
.WithDefaultRedirectUri()
.WithAuthority(authority, tenant)
.WithExtraQueryParameters(new Dictionary<string, string>()
{{"instance_aware", "true" }
})
.WithWindowsBroker().WithMsalLogging();
```
2.Login using a non-public cloud account which is not added to your OS.

**Expected behavior**
Environment of respective cloud in AuthResult (account used is non-public).

**Actual behavior**
Got Public Environment in AuthResult but the account was non-public cloud.

**Possible solution**
Do not override the auth result, instead return the authority as it is at least for non-public cloud OR fix WithMultiCloudSupport(true) with WAM.

bgavrilMS:
Fixes #3637 

**Testing**
unit + manual

**Performance impact**
there is new logic in hot path, but it is behind the "IsMultiCloud" flag.

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -95,7 +95,7 @@ namespace Microsoft.Identity.Client.Broker
                         $" Is Retryable: false \n" +
                         $" Possible causes: \n" +
                         $"- Invalid redirect uri - ensure you have configured the following url in the AAD portal App Registration: " +
-                        $"{WamAdapters.GetExpectedRedirectUri(authenticationRequestParameters.AppConfig.ClientId)} \n" +
+                        $"{GetExpectedRedirectUri(authenticationRequestParameters.AppConfig.ClientId)} \n" +
                         $"- No Internet connection \n" +
                         $"Please see https://aka.ms/msal-net-wam for details about Windows Broker integration";
                     logger.Error($"[WamBroker] WAM_provider_error_{errorCode} {errorMessage}");
@@ -157,7 +157,7 @@ namespace Microsoft.Identity.Client.Broker
 
             if (authenticationRequestParameters.AppConfig.MultiCloudSupportEnabled)
             {
-                authParams.Properties["discover"] = "home";
+                authParams.Properties["instance_aware"] = "true";
             }
 
             //pass extra query parameters if there are any
@@ -199,13 +199,13 @@ namespace Microsoft.Identity.Client.Broker
 
             if (authResult.IsSuccess)
             {
-                msalTokenResponse = WamAdapters.ParseRuntimeResponse(authResult, authenticationRequestParameters, logger);
+                msalTokenResponse = ParseRuntimeResponse(authResult, authenticationRequestParameters, logger);
                 logger.Verbose("[WamBroker] Successfully retrieved token.");
             }
             else
             {
                 logger.Error($"[WamBroker] {errorMessage} {authResult.Error}");
-                WamAdapters.ThrowExceptionFromWamError(authResult, authenticationRequestParameters, logger);
+                ThrowExceptionFromWamError(authResult, authenticationRequestParameters, logger);
             }
 
             return msalTokenResponse;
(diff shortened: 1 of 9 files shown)


src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -8,6 +8,7 @@
     <TargetFrameworkNet5Win>net5.0-windows10.0.17763.0</TargetFrameworkNet5Win>
     <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
     <Configurations>Debug;Release;Debug + MobileApps</Configurations>
+
```
gladjohn, line 11: nit : extra line

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -157,7 +157,7 @@ private enum ResponseStatus
 
             if (authenticationRequestParameters.AppConfig.MultiCloudSupportEnabled)
             {
-                authParams.Properties["discover"] = "home";
+                authParams.Properties["instance_aware"] = "true";
```
gladjohn, line 160: Do we still have ["discover"] = "home", if so what is that used for?

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
bgavrilMS: "discover" = "home" is the magic sauce if you talk directly to WAM. 
"instance_aware" = "true" is the magic sauce if you talk to MSAL C++ or directly to ESTS. MSAL C++ will transform "instace_aware"="true" into "discover"="home" when talking to WAM on its own.

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
```diff
@@ -856,6 +846,41 @@ public void SilentStrategy_ProtectionPolicyNotEnabled_Throws_Exception()
...
+                var tokenResponse = CreateTokenResponseForTest();
+                tokenResponse.AuthorityUrl = "https://login.microsoftonline.us/organizations";
+                broker.AcquireTokenInteractiveAsync(null, null).ReturnsForAnyArgs(Task.FromResult(tokenResponse));
+
+                // Act - interactive flow logs-in Arlighton user
+                var result = await globalPca.AcquireTokenInteractive(TestConstants.s_graphScopes).ExecuteAsync().ConfigureAwait(false);
+                Assert.AreEqual("login.microsoftonline.us", result.Account.Environment);
+                
+                var account = (await globalPca.GetAccountsAsync().ConfigureAwait(false)).Single();
+                Assert.AreEqual("login.microsoftonline.us", account.Environment);
```
pmaytak, line 882:
```suggestion
                Assert.AreEqual("login.microsoftonline.us", result.Account.Environment);
                Assert.AreEqual(TestConstants.Organizations, result.TenantId);
                
                var account = (await globalPca.GetAccountsAsync().ConfigureAwait(false)).Single();
                Assert.AreEqual("login.microsoftonline.us", account.Environment);
                Assert.AreEqual(TestConstants.Organizations, result.TenantId);
```

pmaytak (changes requested):

pmaytak (approved):

PR #3660, opened 2022-09-09
Update logic that calculates if exception is retry-able

### LINKED ISSUE #3661: [Bug] MsalServiceException.IsRetryable is not correctly set
MSAL 4.46.2

Expected: 
`MsalServiceException.IsRetriable` should be true when the HTTP Status code is 408, 429, 5xx or when AAD error code is `temporarily_unavailable`, as per https://docs.microsoft.com/en-us/azure/active-directory/develop/reference-aadsts-error-codes#handling-error-codes-in-your-application

Actual: MSAL always sets it false

bgavrilMS:
Fixes #3661

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -416,10 +416,10 @@ namespace Microsoft.Identity.Client.Internal.Requests
         internal async Task<AuthenticationResult> HandleTokenRefreshErrorAsync(MsalServiceException e, MsalAccessTokenCacheItem cachedAccessTokenItem)
         {
             var logger = AuthenticationRequestParameters.RequestContext.Logger;
-            bool isAadUnavailable = e.IsAadUnavailable();
-            logger.Warning($"Fetching a new AT failed. Is AAD down? {isAadUnavailable}. Is there an AT in the cache that is usable? {cachedAccessTokenItem != null}");
+            
+            logger.Warning($"Fetching a new AT failed. Is exception retry-able? {e.IsRetryable}. Is there an AT in the cache that is usable? {cachedAccessTokenItem != null}");
 
-            if (cachedAccessTokenItem != null && isAadUnavailable)
+            if (cachedAccessTokenItem != null && e.IsRetryable)
             {
                 logger.Info("Returning existing access token. It is not expired, but should be refreshed. ");
 
--- a/src/client/Microsoft.Identity.Client/MsalServiceException.cs
+++ b/src/client/Microsoft.Identity.Client/MsalServiceException.cs
@@ -224,8 +224,10 @@ namespace Microsoft.Identity.Client
         /// </summary>
         private void UpdateIsRetryable()
         {
-            IsRetryable = 
-                HttpManager.IsRetryableStatusCode(StatusCode) ||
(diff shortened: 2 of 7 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -416,10 +416,10 @@ private void LogReturnedToken(AuthenticationResult result)
         internal async Task<AuthenticationResult> HandleTokenRefreshErrorAsync(MsalServiceException e, MsalAccessTokenCacheItem cachedAccessTokenItem)
         {
             var logger = AuthenticationRequestParameters.RequestContext.Logger;
-            bool isAadUnavailable = e.IsAadUnavailable();
-            logger.Warning($"Fetching a new AT failed. Is AAD down? {isAadUnavailable}. Is there an AT in the cache that is usable? {cachedAccessTokenItem != null}");
+            
+            logger.Warning($"Fetching a new AT failed. Is exception retry-able? {e.IsRetryable}. Is there an AT in the cache that is usable? {cachedAccessTokenItem != null}");
 
-            if (cachedAccessTokenItem != null && isAadUnavailable)
```
bgavrilMS, line 422: Too many terms for the same concept. Better to keep only 1 concept - `is retryable`

src/client/Microsoft.Identity.Client/MsalServiceException.cs
```diff
@@ -147,10 +151,15 @@ public MsalServiceException(string errorCode, string errorMessage, int statusCod
         /// http://msdn.microsoft.com/en-us/library/bb268233(v=vs.85).aspx).
         /// You can use this code for purposes such as implementing retry logic or error investigation.
         /// </summary>
-        public int StatusCode { get; internal set; } = 0;
-
-#if !DESKTOP
-#endif
+        public int StatusCode
```
bgavrilMS, line 154: This is such a nasty bug I found here. `StatusCode` can be updated, and in fact the exception factory does not use the ctor to create this exception. But when `StatusCode` is updated (e.g. from 0 to 502), the `IsRetryable` property is not updated. 

Note to self: never trust mutable objects.

tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionTests.cs
```diff
@@ -67,6 +68,49 @@ public void MsalClientException_FromMessageAndCode()
             Assert.IsFalse(piiMessage.Contains(ExMessage));
         }
 
+        [TestMethod]        
+        public void IsRetryable()
```
bgavrilMS, line 72: Sadly we didn't write unit tests when we added this property in the first place....

src/client/Microsoft.Identity.Client/MsalServiceException.cs
```diff
@@ -81,7 +86,6 @@ public MsalServiceException(string errorCode, string errorMessage, int statusCod
             Exception innerException)
             : base(errorCode, errorMessage, innerException)
         {
-            IsRetryable = IsAadUnavailable();
```
pmaytak, line 84: Missing `UpdateIsRetryable()` call.

pmaytak (approved):

PR #3668, opened 2022-09-13
Update Broker csproj to use "Microsoft.Identity.Client.NativeInterop" Version="0.12.4"

gladjohn:
Fixes #

Use "Microsoft.Identity.Client.NativeInterop" Version="0.12.4"

Also include runtime, build and lib assets by adding `all`.

**Testing**
Dev Apps

**Performance impact**
None

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
+++ b/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
@@ -56,7 +56,7 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.12.3" />
+    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.12.4" IncludeAssets="all" />
     <ProjectReference Include="..\Microsoft.Identity.Client\Microsoft.Identity.Client.csproj" />
   </ItemGroup>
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
```diff
@@ -56,7 +56,7 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.12.3" />
+    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.12.4" IncludeAssets="all" />
```
pmaytak, line 59: > Also include runtime, build and lib assets by adding all.

Should only the specific ones be included then? `IncludeAssets=""build;runtime;compile`

pmaytak (approved):

src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
gladjohn: We only have bin, lib and runtime. Will all make sure these three are included? or should we be specific?

pmaytak: https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

I think all should work. I wasn't sure if all would include extra things we don't need, so was trying to clarify with you. :)

PR #3674, opened 2022-09-15
Run .NET 6 unit tests, separate .NET 6 and .NET Core 3.1 integration tests

pmaytak:
Run .NET 6 unit tests
Separate .NET 6 and .NET Core 3.1 integration tests into separate tasks for clarity

image

### DIFF
--- a/build/template-run-integration-tests.yaml
+++ b/build/template-run-integration-tests.yaml
@@ -19,11 +19,25 @@ steps:
     pathtoCustomTestAdapters: 'C:\temp' # Workaround for test failure, as NUnit Test Adapter that gets detected seems to screw something up
 
 - task: VSTest@2
-  displayName: 'Run integration tests (.NET Core)'
+  displayName: 'Run integration tests (.NET Core 3.1)'
   condition: and(succeeded(), eq(variables['RunTests'], 'true'))
   inputs:
     testSelector: 'testAssemblies'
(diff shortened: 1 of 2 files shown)


gladjohn (approved):

build/template-run-integration-tests.yaml
```diff
@@ -19,11 +19,25 @@ steps:
     pathtoCustomTestAdapters: 'C:\temp' # Workaround for test failure, as NUnit Test Adapter that gets detected seems to screw something up
 
 - task: VSTest@2
-  displayName: 'Run integration tests (.NET Core)'
+  displayName: 'Run integration tests (.NET Core 3.1)'
```
bgavrilMS, line 22: Do we need .NET Core 3.1 integration tests? Can't we rely just on unit tests being run there?

bgavrilMS (changes requested):
Avoid running integration tests everywhere.

build/template-run-integration-tests.yaml
pmaytak: Yea, I think Net Core can be removed. NetFx tests would cover Newtonsoft flow, Net6 tests would cover STJ flow + existing non-NetFx NetCore tests.

pmaytak:
Will include this as part of fix for #3682

PR #3677, opened 2022-09-15
Fix for empty scopes. HasScopes does return true when empty scopes are passed

### LINKED ISSUE #3654: [Bug] Crashing Microsoft.Identity.Client.Broker - WithBrokerPreview()
**Logs and network traces**
##### During logout:
```
Ausnahme ausgelöst: "Microsoft.Identity.Client.NativeInterop.MsalRuntimeException" in Microsoft.Identity.Client.NativeInterop.dll
GlobalExceptionHandler: Status: ApiContractViolation
Context: (pii)
Tag: 0x2039c1cb 
Ausnahme ausgelöst: "Microsoft.Identity.Client.NativeInterop.MsalRuntimeException" in Microsoft.Identity.Client.NativeInterop.dll
GlobalExceptionHandler: Status: ApiContractViolation
Context: (pii)
Tag: 0x2039c1cb 
```
##### During login:
no logs, app close without any error in VisualStudio. But not always (~70% crash), sometimes login is working.

Microsoft.Identity.Client 4.46.2

**Platform**
.NET 6 WPF

* Desktop / Mobile
    * [x] Interactive
    * [ ] Username Password
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 

This is a new app or experiment.

**Repro**

```csharp
readonly string tenantId = "common";
private static string Instance = "https://login.microsoftonline.com/";
readonly string clientId = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx";
private IPublicClientApplication? _clientApp;
private IPublicClientApplication? PublicClientApp { get { return _clientApp; } }
private static AuthenticationResult? authResult;

...

_clientApp = PublicClientApplicationBuilder.Create(clientId)
                    .WithAuthority($"{Instance}{tenantId}")
                    .WithRedirectUri("http://localhost")
                    .WithBrokerPreview()
                    .Build();

...
authResult = await PublicClientApp.AcquireTokenInteractive(scopes)
                        .WithParentActivityOrWindow(new WindowInteropHelper(window).Handle)
                        .WithUseEmbeddedWebView(true)
                        .ExecuteAsync();
...

//logout
IEnumerable<IAccount> accounts = await PublicClientApp.GetAccountsAsync();
                if (accounts.Any())
                {
                    try
                    {
                        await PublicClientApp.RemoveAsync(accounts.First());
                    }
                    catch (MsalException ex)
                    {
                        Trace.WriteLine($"MsalException Error signing-out user: {ex.Message}");
                    }
                    catch (Exception ex)
                    {
                        Trace.WriteLine($"Exception Error signing-out user: {ex.Message}");
                    }
                }
```

**Expected behavior**
No crashing app, no errors on logout.

**Actual behavior**
During login, the app is crashing very often, during logout I get errors as above written

**Possible solution**
not using `WithBrokerPreview()`, instead see alternative code.

**Alternative working code**
same as above, just remove
```csharp
.WithBrokerPreview()
```
and you receive the "old" Loginscreen from Microsoft.

gladjohn:
Fixes #3654 

Fix for empty/null scopes. HasScopes does return true when empty scopes are passed but works for null scopes.  

This fixes the user issue 

`GlobalExceptionHandler: 
Status: ApiContractViolation
Context: (pii)`

**Testing**
Dev Apps

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -322,7 +322,7 @@ namespace Microsoft.Identity.Client.Broker
         {
             //MSAL Runtime throws an ApiContractViolation Exception with Tag: 0x2039c1cb (InvalidArg)
             //When no scopes are passed, this will check if user is passing scopes
-            var scopes = authenticationRequestParameters.Scope.ElementAtOrDefault(0);
+            var scopes = authenticationRequestParameters.Scope?.ElementAtOrDefault(0);
 
             if (string.IsNullOrWhiteSpace(scopes))
             {
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -322,7 +322,9 @@ namespace Microsoft.Identity.Client.Broker
         {
             //MSAL Runtime throws an ApiContractViolation Exception with Tag: 0x2039c1cb (InvalidArg)
             //When no scopes are passed, this will check if user is passing scopes
-            if (!authenticationRequestParameters.HasScopes)
+            var scopes = authenticationRequestParameters.Scope.ElementAtOrDefault(0);
+
+            if (string.IsNullOrWhiteSpace(scopes))
             {
                 logger.Error($"[WamBroker] {MsalError.WamScopesRequired} " +
                     $"{MsalErrorMessage.ScopesRequired}");


src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -322,7 +322,9 @@ private static string GetExpectedRedirectUri(string clientId)
         {
             //MSAL Runtime throws an ApiContractViolation Exception with Tag: 0x2039c1cb (InvalidArg)
             //When no scopes are passed, this will check if user is passing scopes
-            if (!authenticationRequestParameters.HasScopes)
+            var scopes = authenticationRequestParameters.Scope.ElementAtOrDefault(0);
+
+            if (string.IsNullOrWhiteSpace(scopes))
```
trwalke, line 327: nit, use IsNullOrEmpty
There is a slight performance improvement and whitespaces will not return true when using this

neha-bhargava (approved):

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
gladjohn: we do want to throw for this as well, s_scopes = new[] { "   " }

trwalke: ok, makes sense

trwalke (approved):

gladjohn:
@pmaytak please take a look at this and approve, if I need to merge this.

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -322,7 +322,9 @@ private static string GetExpectedRedirectUri(string clientId)
         {
             //MSAL Runtime throws an ApiContractViolation Exception with Tag: 0x2039c1cb (InvalidArg)
             //When no scopes are passed, this will check if user is passing scopes
-            if (!authenticationRequestParameters.HasScopes)
+            var scopes = authenticationRequestParameters.Scope.ElementAtOrDefault(0);
```
pmaytak, line 325: `Scope` can be null, needs null check?

pmaytak (approved):

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
gladjohn: if null scopes are passed then this will return null and the next check if catch it

pmaytak: I mean if `Scope` itself is null, this will throw? `HasScopes` does a null check on `Scope`.

gladjohn:
```
            // Creating HashSet
            HashSet<string> myhash1 = new HashSet<string>();

            // Add the elements in HashSet
            //myhash1.Add("pass");
            myhash1.Add(null);
            //myhash1.Add(" ");
            //myhash1.Add("");

            var scopes = myhash1.ElementAtOrDefault(0);

            if (string.IsNullOrWhiteSpace(scopes))
            {
                Console.WriteLine("Scopes required!!!");
            }
            else
            {
                Console.WriteLine("Passed!!!");
            }
```

trwalke: I think we set the scopes to an empty sorted set if we detect that it is null so technically it should always be something. However, not sure if broker flow does something different. Doesn't hurt to add the check.

PR #3679, opened 2022-09-16
Remove net6 target

pmaytak:
Remove net6 target from MSAL and unit tests, instead of reverting the whole code base. Since SUPPORTS_SYSTEM_TEXT_JSON flag is only added for net6, the code paths for other targets are unchanged. Left net6 target in integration tests, but it'll run using netcoreapp2.1 MSAL binary.


src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -7,7 +7,7 @@
     <TargetFrameworkNetCore>netcoreapp2.1</TargetFrameworkNetCore>
     <TargetFrameworkNet5Win>net5.0-windows10.0.17763.0</TargetFrameworkNet5Win>
     <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
-    <TargetFrameworkNet6>net6.0</TargetFrameworkNet6>
+    <!--<TargetFrameworkNet6>net6.0</TargetFrameworkNet6>-->
```
gladjohn, line 10: we can just remove this for now?

gladjohn (approved):

SameerK-MSFT (approved):

Issue #3683, opened 2022-09-19
[Bug] MSAL 4.37.- logs critical errors when they are not critical


jmprieur:
**Logs and network traces**
Without logs or traces, it is unlikely that the team can investigate your issue. Capturing logs and network traces is described in [Logging wiki](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging).

**Which version of MSAL.NET are you using?**

**Platform**
ASP.NET Core 6.0

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Authentication
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web app 
    * [ x] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 
    * [ ] Service to Service calls

Other?

**Is this a new or existing app?**

**Repro**
- Create an ASP.NET Core web app (using Visual Studio File new project ASP.NET core web app, and checking Microsoft identity platform for the authorization. Check 'use graph" in the wizard.
- Update MSAL.NET to 4.47.0
- Run it

**Expected behavior**
The logs shouldn't call out critical errors when they are not

**Actual behavior**
![image](https://user-images.githubusercontent.com/13203188/190942380-eaacd17e-fd45-445b-8dbf-fcefeb676315.png)

Does it means that MSAL logs as "always" things that should not be logged as "always"?

See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/search?q=logger.Always&type=code.
I don't think that any of these should be logged as "always (which will translate to critical in ASP.NET Core)

bgavrilMS:
I would think https://github.com/AzureAD/microsoft-identity-web/pull/1891 is enough? 

`Always` log level concept is similar to Azure Identity log level, which should be translated to "asp.net core's info level"

PR #3684, opened 2022-09-19
Change the logger.Always messages to logger.Info as it creates too mu…

jennyf19:
…ch noise for services. Many are already taking the metrics from the authResult, so they may not want it always again in the logs.

Fixes #
#3683 

In .NET Core, the mapping Always to Critical is not going to work, because they appear as Critical (see the linked item for details). Service owners will not want this level appearing in their logs. Based on the logs, they can all be added an information and not always.

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -427,7 +427,7 @@ namespace Microsoft.Identity.Client
 
             var accessTokens = Accessor.GetAllAccessTokens(partitionKey, logger);
 
-            requestParams.RequestContext.Logger.Always($"[FindAccessTokenAsync] Discovered {accessTokens.Count} access tokens in cache using partition key: {partitionKey}");
+            requestParams.RequestContext.Logger.Info($"[FindAccessTokenAsync] Discovered {accessTokens.Count} access tokens in cache using partition key: {partitionKey}. ");
 
             if (accessTokens.Count == 0)
             {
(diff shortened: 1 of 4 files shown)


jmprieur (approved):
LGTM
Thanks @jennyf19 

@pmaytak @trwalke @neha-bhargava @gladjohn @sameerkapps : why do you want to log always this information (given it would be translated in Critical and displayed in Red in .NET Core)

bgavrilMS (approved):

bgavrilMS:
> LGTM Thanks @jennyf19
> 
> @pmaytak @trwalke @neha-bhargava @gladjohn @sameerkapps : why do you want to log always this information (given it would be translated in Critical and displayed in Red in .NET Core)

Maybe we should not translate MSAL's `Always` to ASP.NET Core's `Critical` ? Always are log messages we always want to log. They are not indicative of an error. The purpose is to avoid asking customer who already has logging enabled to change log verbosity.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -756,7 +756,7 @@ internal async Task ExpireAllAccessTokensForTestAsync()
 
             var requestKey = CacheKeyFactory.GetKeyFromRequest(requestParams);
             var refreshTokens = Accessor.GetAllRefreshTokens(requestKey);
-            requestParams.RequestContext.Logger.Always($"[FindRefreshTokenAsync] Discovered {refreshTokens.Count} refresh tokens in cache using key: {requestKey}");
+            requestParams.RequestContext.Logger.Info($"[FindRefreshTokenAsync] Discovered {refreshTokens.Count} refresh tokens in cache using key: {requestKey}. ");
```
bgavrilMS, line 759: These 2 are quite useful.

bgavrilMS (changes requested):
Isn't https://github.com/AzureAD/microsoft-identity-web/pull/1891 enough to fix the issue? The "log always" concept was copied from Azure Identity, it's just that it should not translate to ASP.NET Core's "critical"

https://learn.microsoft.com/en-us/dotnet/azure/sdk/logging#map-to-aspnet-core-logging

bgavrilMS:
Closing this one, the information there is pretty useful. Solution to the root bug was added in Id.Web

PR #3692, opened 2022-09-21
Adding a PublishNugetAndSymbols Build Variable

### LINKED ISSUE #3689: [Feature Request] Remove publishing to VSTS Feed from the build pipeline
Remove publishing to VSTS Feed from the build pipeline. 

Our daily pipeline is pushing all 3 packages (client, desktop and broker) to our internal IDDP feed. But when we release version it, the internal feed gets the release versioned NuGet, but we may not necessarily release the one we tagged as release. Just like nuget.org, internal feed can take only one copy of a given version, this blocks us from publishing the actual released version to the internal feed. 

For 4.47.0 release we had 3 RCs, and the first RC was pushed to the internal feed. But we released RC3 to NuGet.org. 

After MSAL 4.47.0 release, we had to release a 4.47.1 to fix this issue for our 1P customers (MS Internal). This bug needs to be handled on priority to avoid such issues in the future.

gladjohn:
Fixes #3689 

Adding a PublishNugetAndSymbols Build Variable check on Task `'Publish packages to VSTS feed'` and on `'Publish symbols'`

**Testing**
none

**Performance impact**
n/a

**Documentation**

### DIFF
--- a/build/template-publish-packages-and-symbols.yaml
+++ b/build/template-publish-packages-and-symbols.yaml
@@ -25,11 +25,11 @@ steps:
     publishVstsFeed: '46419298-b96c-437f-bd4c-12c8df7f868d'
     allowPackageConflicts: true
   continueOnError: true
-  condition: eq(variables['PublishNugetAndSymbols'], 'true')
+  condition: eq(variables['PublishToVSTSFeed'], 'true')
 
 - task: PublishSymbols@2
   displayName: 'Publish symbols'
   inputs:
     SearchPattern: '${{ parameters.SymbolPublishWildcard }}'
     SymbolServerType: TeamServices
-  condition: eq(variables['PublishNugetAndSymbols'], 'true')
+  condition: eq(variables['PublishSymbols'], 'true')
--- a/build/template-publish-packages-and-symbols.yaml
+++ b/build/template-publish-packages-and-symbols.yaml
@@ -25,11 +25,11 @@ steps:
     publishVstsFeed: '46419298-b96c-437f-bd4c-12c8df7f868d'
     allowPackageConflicts: true
   continueOnError: true
-  condition: eq(${{ parameters.PublishArtifacts }}, 'true')
+  condition: eq(variables['PublishNugetAndSymbols'], 'true')
 
 - task: PublishSymbols@2
   displayName: 'Publish symbols'
   inputs:
     SearchPattern: '${{ parameters.SymbolPublishWildcard }}'
     SymbolServerType: TeamServices
-  condition: eq(${{ parameters.PublishArtifacts }}, 'true')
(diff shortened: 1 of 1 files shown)


build/template-publish-packages-and-symbols.yaml
```diff
@@ -25,11 +25,11 @@ steps:
     publishVstsFeed: '46419298-b96c-437f-bd4c-12c8df7f868d'
     allowPackageConflicts: true
   continueOnError: true
-  condition: eq(${{ parameters.PublishArtifacts }}, 'true')
+  condition: eq(variables['PublishNugetAndSymbols'], 'true')
```
pmaytak, line 28: nit: Wonder if it's clearer to specify that it's a publish to Vsts feed?

build/template-publish-packages-and-symbols.yaml
```diff
@@ -25,11 +25,11 @@ steps:
...
 
 - task: PublishSymbols@2
   displayName: 'Publish symbols'
   inputs:
     SearchPattern: '${{ parameters.SymbolPublishWildcard }}'
     SymbolServerType: TeamServices
-  condition: eq(${{ parameters.PublishArtifacts }}, 'true')
+  condition: eq(variables['PublishNugetAndSymbols'], 'true')
```
pmaytak, line 35: Does this variable need to be passed in from the top level yaml to work, like here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b529f3024f4aed17aa9bdc48e23294f03a5d18e6/build/pipeline-releasebuild.yaml#L6-L10

bgavrilMS (approved):

SameerK-MSFT (approved):

build/template-publish-packages-and-symbols.yaml
gladjohn: Yes this will be passed as a build variable. I split them into `PublishToVSTSFeed` and `PublishSymbols`  variables. 

image

Ran a [test build](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=995169&view=logs&j=9906ff93-67e4-5c55-442d-8f18d4fd3348&t=daaef45a-0156-5701-28d1-cab03afd58b7), without the variable and confirmed that these are not executed 

image

and[ ran one with the variables ](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=995192&view=logs&j=9906ff93-67e4-5c55-442d-8f18d4fd3348)and these were pushed to the feed, 

image

Basically with this fix, the default is turned off so we are not publishing it nightly or for test builds. Even for release build we should not set this, instead we should rely on the publish package pipeline. This variable should be set only when we know there is an internal release that needs to be pushed.

gladjohn: @pmaytak

PR #3696, opened 2022-09-22
Improve some WAM error messages 

### LINKED ISSUE #3675: [Request] WAM support for no scopes
**Logs and network traces**
none

latest

**Platform**
any

**Repro**

```csharp
var authResult = await pca.AcquireTokenInteractive(new[] { "" })
                                          .ExecuteAsync();
```

**Expected behavior**
Friendly exception

**Actual behavior**
Flow proceeds and results in unfriednly exceptions

Fix: authenticationRequestParameters.HasScopes returns true if empty string is passed as scopes

bgavrilMS:
Fixes #3675 

- improve error messages
- fix "HasScopes" helper and use it
- unit tests

**Testing**

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -93,7 +94,7 @@ internal class AuthenticationRequestParameters
         public ICacheSessionManager CacheSessionManager { get; }
         public HashSet<string> Scope { get; }
 
-        public bool HasScopes => Scope != null && Scope.Count > 0;
+        public bool HasScopes => Scope != null && Scope.Count > 0 && Scope.Any(s => s != string.Empty);
```
SameerK-MSFT, line 97: Scopes.Count > 0 not required .Any will cover it

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -84,12 +87,57 @@ public void NoWamOnADFS()
...
+               .Create(TestConstants.ClientId)
+               .WithBrokerPreview()
+               .Build();
+
+            // no window handle - throw
+            var ex = await AssertException.TaskThrowsAsync<MsalClientException>(
+                () => pca.AcquireTokenInteractive(new[] { "" }).ExecuteAsync()).ConfigureAwait(false);
+
+            Assert.AreEqual("window_handle_required", ex.ErrorCode);
+
```
SameerK-MSFT, line 103: nit: extra line. Here + next method

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -316,15 +316,13 @@ private static string GetExpectedRedirectUri(string clientId)
         /// <param name="authenticationRequestParameters"></param>
         /// <param name="logger"></param>
         /// <exception cref="MsalClientException"></exception>
-        public static void ValidateAuthParams(
+        private static void ValidateAuthParams(
```
gladjohn, line 319: Thanks, I missed this.

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
pmaytak, line 97: - This still misses a case when scope is whitespace.
- Is this in a hot path? Just thinking so that Linq doesn't regress the perf then.
- Hmm, what about changing the logic when we add to `Scope` so we don't add empty scopes or whitespace; so we can forgo this check?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -95,7 +95,7 @@ private static void OnProcessExit(object sender, EventArgs e)
             {
                 throw new MsalClientException(
                     "window_handle_required",
-                    "Public Client applications wanting to use WAM need to provide their window handle. Console applications can use GetConsoleWindow Windows API for this.");
+                    "Desktop applications wanting to use the broker need to provide their window handle. See https://aka.ms/msal-net-wam#parent-window-handles");
```
pmaytak, line 98: +1, I was just thinking of updating this message myself...

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
bgavrilMS: Fixed. I removed the `HasScopes` property, although it wasn't used in the hot path.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -93,6 +95,20 @@ protected override void Validate()
             {
                 Parameters.SendX5C = this.ServiceBundle.Config.SendX5C;
             }
+
+            // During AT Silent with no scopes, Unlike AAD, B2C will not issue an access token if no scopes are requested
+            // And we don't want to refresh the RT on every ATS call
+            // See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/715 for details
+            if (ServiceBundle.Config.Authority.AuthorityInfo.AuthorityType == AuthorityType.B2C &&
```
bgavrilMS, line 102: I just moved this to "Validate" instead of "ExecuteAsync"

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -316,15 +316,13 @@ private static string GetExpectedRedirectUri(string clientId)
...
             AuthenticationRequestParameters authenticationRequestParameters,
             ILoggerAdapter logger)
         {
             //MSAL Runtime throws an ApiContractViolation Exception with Tag: 0x2039c1cb (InvalidArg)
             //When no scopes are passed, this will check if user is passing scopes
-            var scopes = authenticationRequestParameters.Scope?.ElementAtOrDefault(0);
-
-            if (string.IsNullOrWhiteSpace(scopes))
+            if (!ScopeHelper.HasNonMsalScopes(authenticationRequestParameters.Scope))
```
bgavrilMS, line 325: @gladjohn - since MSAL C++ has undefined behavior when using one of the 3 reserved scopes, let's throw an error message here.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
gladjohn, line 98:
```suggestion
                    "Desktop applications wanting to use the broker need to provide their own window handle. See https://aka.ms/msal-net-wam#parent-window-handles");
```

gladjohn (approved):

PR #3697, opened 2022-09-22
Soft deprecate WithAuthority on request level

bgavrilMS:
Fixes #

Hide `WithAuthority` methods.

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
@@ -122,6 +123,8 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
+        /// Important: Use WithTenantId or WithTenantIdFromAuthority instead, or WithB2CAuthority for B2C authorities.
+        /// 
         /// Specific authority for which the token is requested. Passing a different value than configured
         /// at the application constructor narrows down the selection to a specific tenant.
         /// This does not change the configured value in the application. This is specific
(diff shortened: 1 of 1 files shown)


jmprieur (approved):
LGTM
Thanks @bgavrilMS 

do we want to add an obsolete(false) attribute?

SameerK-MSFT (approved):

gladjohn (approved):

pmaytak (commented):
What about:
- https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9af5689b52dbd05c3d23e857cb9f3f2530f46955/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs#L260
- https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9af5689b52dbd05c3d23e857cb9f3f2530f46955/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs#L360

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -122,6 +123,8 @@ public T WithExtraQueryParameters(string extraQueryParameters)
         }
 
         /// <summary>
+        /// Important: Use WithTenantId or WithTenantIdFromAuthority instead, or WithB2CAuthority for B2C authorities.
```
pmaytak, line 126: > Use WithTenantId or WithTenantIdFromAuthority instead

Does this include ADFS and DSTS authortities?

bgavrilMS:
> LGTM Thanks @bgavrilMS
> 
> do we want to add an obsolete(false) attribute?

I had a long conversation with Azure SDK folks about this and they advise against it. This is because of diamond dependencies again, which can cause warnings. I think the sceanrio is like this: 

- App references MSAL 4.40, uses an API which is not obsolete
- App also references Azure.Identiy, which references MSAL 4.48, where API is obsolete
- App now gets a warning (and many apps use "report warnings as errors", which breaks the build)

So they recommend just hiding obsolete APIs.

CC: @pmaytak @neha-bhargava @SameerK-MSFT @trwalke @gladjohn

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
bgavrilMS: No, those are not supported scenarios

PR #3698, opened 2022-09-22
Setting broker and desktop AssemblyVersion to a pre-released version

gladjohn:
Fixes #

Setting broker and desktop AssemblyVersion to a pre-released version. This the version we use in the client project as well. Our build pipeline sets the `MsalClientSemVer` but if it is not set then `4.7.1` will be used. 

**Testing**
none

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
+++ b/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
@@ -12,7 +12,7 @@
 
   <PropertyGroup Label="NuGet and AssemblyInfo metadata">
     <!--This should be passed from the VSTS build-->
-    <MsalClientSemVer Condition="'$(MsalClientSemVer)' == ''">4.7.1-preview</MsalClientSemVer>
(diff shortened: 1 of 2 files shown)


pmaytak (commented):
Why is the main MSAL project set to 4.46, in the first place? The issue I see with this is just updating this version to latest all the time. Who's going to do that? And if we won't update it regularly, then unless there's a specific reason to set this version to latest, I would make these all `4.7.1-local` or something.

bgavrilMS (approved):

gladjohn:
> Why is the main MSAL project set to 4.46, in the first place? The issue I see with this is just updating this version to latest all the time. Who's going to do that? And if we won't update it regularly, then unless there's a specific reason to set this version to latest, I would make these all `4.7.1-local` or something.

4.46 is a released version it is Ok to have a released version in here, we are just setting broker and desktop AssemblyVersion to a released version (4.46) to, this way even if someone removes the version variable we will not be using 4.7. which is a future release. This is a onetime change, we should make sure no one edits this.

SameerK-MSFT (approved):

PR #3715, opened 2022-09-28
WAM - throw when msalruntime.dll is not found

### LINKED ISSUE #3699: [Bug] WAM - throw when msalruntime.dll is not found
**Logs and network traces**
none

4.47.1

**Platform**
Net.Core (consuming netstandard 2.0 interop lib)

* Desktop / Mobile
    * [x] Interactive
    * [ ] Username Password
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 

Problem?
We only [log the failure](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs#L46) when interop is not able to load msalruntime.dll, instead we should throw

gladjohn:
Fixes #3699

throwing when MSAL runtime init fails

**Testing**
Dev App

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -44,12 +44,22 @@ namespace Microsoft.Identity.Client.Broker
             catch (MsalRuntimeException ex) when (ex.Status == ResponseStatus.ApiContractViolation)
             {
                 // failed to initialize msal runtime - can happen on older versions of Windows. Means broker is not available.
+                // We will never get here with our current OS version check. Instead in this scenario we will fallback to the browser
+                // but MSALRuntime does it's internal check for OS compatibility and throws an ApiContractViolation MsalRuntimeException.
+                // For any reason, if our OS check fails then this will catch the MsalRuntimeException and 
+                // log but we will not fallback to the browser in this case. 
                 s_initException = ex;
 
                 // ignored
                 return null;
             }
-
+            catch (Exception ex)
+            {
+                // When MSAL Runtime dlls fails to load then we catch the exception and throw with a meaningful
+                // message with information on how to troubleshoot
+                throw new MsalClientException(
+                    "wam_runtime_init_failed", ex.Message + " See https://aka.ms/msal-net-wam#troubleshooting");
+            }
         });
 
         /// <summary>
@@ -438,7 +448,7 @@ namespace Microsoft.Identity.Client.Broker
 
             if (s_lazyCore.Value == null)
             {
-                _logger.Info("[WAM Broker] MsalRuntime init failed...");
+                _logger.Info("[WAM Broker] MsalRuntime initialization failed. See https://aka.ms/msal-net-wam##wam-limitations");
                 _logger.InfoPii(s_initException);
                 return false;
             }
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -41,8 +41,9 @@ namespace Microsoft.Identity.Client.Broker
 
                 return new NativeInterop.Core();
             }
-            catch (Exception ex)
+            catch (MsalRuntimeException ex) when (ex.Status == ResponseStatus.ApiContractViolation)
             {
+                // failed to initialize msal runtime - can happen on older versions of Windows. Means broker is not available.
                 s_initException = ex;
 
                 // ignored
@@ -440,7 +441,6 @@ namespace Microsoft.Identity.Client.Broker
                 _logger.Info("[WAM Broker] MsalRuntime init failed...");
                 _logger.InfoPii(s_initException);
 
-                throw new MsalServiceException("wam_runtime_init_failed", "MsalRuntime init failed.");
             }
 
             _logger.Verbose($"[WAM Broker] MsalRuntime init successful.");
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -58,7 +58,7 @@ namespace Microsoft.Identity.Client.Broker
                 // When MSAL Runtime dlls fails to load then we catch the exception and throw with a meaningful
                 // message with information on how to troubleshoot
                 throw new MsalClientException(
-                    "wam_runtime_init_failed", ex.Message + " See https://aka.ms/msal-net-wam#troubleshooting");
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -440,7 +440,7 @@ public bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
                 _logger.Info("[WAM Broker] MsalRuntime init failed...");
                 _logger.InfoPii(s_initException);
 
-                return false;
+                throw new MsalServiceException("wam_runtime_init_failed", "MsalRuntime init failed.");
```
pmaytak, line 443: This should be MsalClientException. 
Also please add a descriptive and actionable message. How can the user fix this error? If the solution is verbose, then maybe adding a aka.ms link to WAM wiki would be good (and add a Troubleshooting section there, if we already don't have it).

pmaytak (changes requested):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -440,7 +440,7 @@ public bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
                 _logger.Info("[WAM Broker] MsalRuntime init failed...");
                 _logger.InfoPii(s_initException);
 
-                return false;
```
bgavrilMS, line 443: This is not the correct fix I don't think. We want to propagate all exceptions except the ApiContractViolation which states that msal_runtime cannot be initialized (e.g. we're on Win8).

I would suggest instead to update the initialization method like this:


```csharp
      private static Lazy<NativeInterop.Core> s_lazyCore = new Lazy<NativeInterop.Core>(() =>
        {
            try
            {
                AppDomain.CurrentDomain.ProcessExit += OnProcessExit;

                return new NativeInterop.Core();
            }                  
            catch (MsalRuntimeException ex) when (ex.Status == ResponseStatus.ApiContractViolation)
            {
                // failed to initialize msal runtime - can happen on older versions of Windows. Means broker is not available.
                s_initException = ex;

                // ignored
                return null;
            }

            // any other exceptions are not expected and should bubble up to the app developer, so that they can log bugs.
        });

```

bgavrilMS: (if this is acceptable, consider it as sign off from me).

gladjohn: agreed, did not think of `MsalRuntimeException`

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
gladjohn: Added https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/wam#troubleshooting

gladjohn: And we now throw the MSALRuntimeException that provides the details of the error

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -439,7 +440,6 @@ public bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
             {
                 _logger.Info("[WAM Broker] MsalRuntime init failed...");
```
pmaytak, line 441:
```suggestion
                _logger.Info("[WAM Broker] MsalRuntime initialization failed. See https://aka.ms/msal-net-wam#troubleshooting"); 
```

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -41,8 +41,9 @@ internal class RuntimeBroker : IBroker
 
                 return new NativeInterop.Core();
             }
-            catch (Exception ex)
+            catch (MsalRuntimeException ex) when (ex.Status == ResponseStatus.ApiContractViolation)
```
pmaytak, line 44: So this would catch if MSAL fails to load runtime DLL? I guess the `ApiContractViolation` status is a bit misleading in this situation.

In what other error scenarios would this status appear?

What is end-to-end developer experience here if runtime DLL is missing? So we call IsBrokerInstalledAndInvokable, which returns falls. If OS is unsupported, I think we give descriptive error, right? But what if some package restore failed, how would the user know? Will the user be able to self help in that scenario?

gladjohn: right, so based on the testing with the changes I did earlier here are the findings. 

`With catch (MsalRuntimeException ex) when (ex.Status == ResponseStatus.ApiContractViolation)`

This will happen on older versions of Windows. Which means broker is not available. But, We will never get here with our current OS version check. Instead in this scenario we will fallback to the browser. MSALRuntime does it's internal check for OS compatibility and throws an ApiContractViolation MsalRuntimeException. if, For any reason, if our OS check fails then this will catch the MsalRuntimeException and log but we will not fallback to the browser.

**for the actual fix we need this line,** 

```
catch (Exception ex)
{
    // When MSAL Runtime dlls fails to load then we catch the exception and throw with a meaningful
    // message with information on how to troubleshoot
        "wam_runtime_init_failed", ex.Message + " See https://aka.ms/msal-net-wam#troubleshooting");
}
```
This catches the dll load exception and throws with instructions on how to fix the issue.

gladjohn: @bgavrilMS please review

pmaytak (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -41,14 +41,25 @@ internal class RuntimeBroker : IBroker
...
                 // ignored
                 return null;
             }
-
+            catch (Exception ex)
+            {
+                // When MSAL Runtime dlls fails to load then we catch the exception and throw with a meaningful
+                // message with information on how to troubleshoot
+                throw new MsalClientException(
+                    "wam_runtime_init_failed", ex.Message + " See https://aka.ms/msal-net-wam#troubleshooting");
```
pmaytak, line 61: `ex` may need to be included as an inner exception.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
gladjohn: added

gladjohn:
image

Issue #3596, opened 2022-08-12
[Feature Request] Log L1/L2 cache details to telemetry pipeline


neha-bhargava:
Log data points for 
-	L1 / L2 cache latency 
-	L1 / L2 cache hit (true / false)
-	Snapshot of L1 cache item remaining lifetime

Ref: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3511

neha-bhargava:
Created issues in Microsoft.Identity.Web to track work there:
* https://github.com/AzureAD/microsoft-identity-web/issues/1900
* https://github.com/AzureAD/microsoft-identity-web/issues/1901

neha-bhargava:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3738: capture cache latency in MSAL.

PR #3737, opened 2022-10-06
Add cache details to token cache notification args

neha-bhargava:
Fixes #
#3596 

Add cache details to token cache notification args and log data to telemetry pipeline.

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/RequestContext.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/RequestContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Internal.Logger;
@@ -23,6 +24,8 @@ namespace Microsoft.Identity.Client.Internal
         public ApiEvent ApiEvent { get; set; }
 
         public CancellationToken UserCancellationToken { get; }
+
+        public Dictionary<string, object> CacheDetails { get; } = new Dictionary<string, object>();
 
         public RequestContext(IServiceBundle serviceBundle, Guid correlationId, CancellationToken cancellationToken = default)
         {
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -96,7 +96,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
                     LogReturnedToken(authenticationResult);
                     UpdateTelemetry(sw, apiEvent, authenticationResult);
                     LogMetricsFromAuthResult(authenticationResult, AuthenticationRequestParameters.RequestContext.Logger);
-                    LogSuccessfulTelemetryToClient(authenticationResult, telemetryEventDetails, telemetryClients);
+                    LogSuccessfulTelemetryToClient(authenticationResult, AuthenticationRequestParameters.RequestContext.CacheDetails, telemetryEventDetails, telemetryClients);
 
                     return authenticationResult;
                 }
@@ -129,7 +129,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
             }
         }
 
-        private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationResult, MsalTelemetryEventDetails telemetryEventDetails, ITelemetryClient[] telemetryClients)
+        private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationResult, Dictionary<string, object> cacheDetails, MsalTelemetryEventDetails telemetryEventDetails, ITelemetryClient[] telemetryClients)
         {
             if (telemetryClients.HasEnabledClients(TelemetryConstants.AcquireTokenEventName))
             {
@@ -142,6 +142,39 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 telemetryEventDetails.SetProperty(TelemetryConstants.PopToken, authenticationResult.TokenType.Equals(Constants.PoPTokenType));
                 telemetryEventDetails.SetProperty(TelemetryConstants.RemainingLifetime, (authenticationResult.ExpiresOn - DateTime.Now).TotalMilliseconds);
                 telemetryEventDetails.SetProperty(TelemetryConstants.ActivityId, authenticationResult.CorrelationId);
+                telemetryEventDetails.SetProperty(TelemetryConstants.RefreshOn, 
+                    authenticationResult.AuthenticationResultMetadata.RefreshOn.HasValue ?
+                    DateTimeHelpers.DateTimeToUnixTimestampMilliseconds(authenticationResult.AuthenticationResultMetadata.RefreshOn.Value)
+                    : 0);
+
+                LogCacheDetailsToTelemetryClient(cacheDetails, telemetryEventDetails);
+            }
+        }
+
+        private void LogCacheDetailsToTelemetryClient(Dictionary<string, object> cacheDetails, MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            if ((cacheDetails == null) || (cacheDetails.Count() == 0)) return;
+
+            object value;
+
+            if (cacheDetails.TryGetValue(TelemetryCacheConstants.CacheUsed, out value))
+            {
+                telemetryEventDetails.SetProperty(TelemetryCacheConstants.CacheUsed, ((CacheUsed)value).ToString());
+            }
+
+            if (cacheDetails.TryGetValue(TelemetryCacheConstants.CacheUsed, out value))
+            {
+                telemetryEventDetails.SetProperty(TelemetryCacheConstants.CacheUsed, ((CacheUsed)value).ToString());
+            }
+
+            if (cacheDetails.TryGetValue(TelemetryCacheConstants.L1Latency, out value))
+            {
+                telemetryEventDetails.SetProperty(TelemetryCacheConstants.L1Latency, (long)value);
+            }
+
+            if (cacheDetails.TryGetValue(TelemetryCacheConstants.L2Latency, out value))
+            {
+                telemetryEventDetails.SetProperty(TelemetryCacheConstants.L2Latency, (long)value);
             }
         }
 
--- a/src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs
+++ b/src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.TelemetryCore.TelemetryClient
+{
+    /// <summary>
+    /// Contains the parameters of cache to log to telemetry.
+    /// </summary>
+    public class TelemetryCacheConstants
+    {
+        /// <summary>
+        /// Constant to use as key to Cache Details dictionary.
+        /// </summary>
+        public const string CacheUsed = "CacheUsed";
+
+        /// <summary>
+        /// Constant to use as key to Cache Details dictionary.
+        /// </summary>
+        public const string L1Latency = "L1Latency";
+
+        /// <summary>
+        /// Constant to use as key to Cache Details dictionary.
+        /// </summary>
+        public const string L2Latency = "L2Latency";
+    }
+
+    /// <summary>
+    /// Indicates the cache which was used to get the token.
+    /// </summary>
+    public enum CacheUsed
+    {
+        /// <summary>
+        /// Indicates that the token was not cached
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Token was obtained from L1
+        /// </summary>
+        L1 = 1,
+
+        /// <summary>
+        /// Token was obtained from L2
+        /// </summary>
+        L2 = 2
+    }
+}
--- a/src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
+++ b/src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Identity.Client.TelemetryCore
         public const string DurationInHttp = "DurationInHttp";
         public const string ActivityId = "ActivityId";
         public const string Resource = "Resource";
+        public const string RefreshOn = "RefreshOn";
 
 #endregion
     }
--- a/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
@@ -147,6 +147,44 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
+        /// This constructor is for test purposes only. It allows apps to unit test their MSAL token cache implementation code.
+        /// </summary>
+        public TokenCacheNotificationArgs(    // only use this constructor in product code
+            ITokenCacheSerializer tokenCache,
+            string clientId,
+            IAccount account,
+            bool hasStateChanged,
+            bool isApplicationCache,
+            string suggestedCacheKey,
+            bool hasTokens,
+            DateTimeOffset? suggestedCacheExpiry,
+            CancellationToken cancellationToken,
+            Guid correlationId,
+            IEnumerable<string> requestScopes,
+            string requestTenantId,
+            IIdentityLogger identityLogger,
+            bool piiLoggingEnabled,
+            Dictionary<string, object> cacheDetails)
+
+        {
+            TokenCache = tokenCache;
+            ClientId = clientId;
+            Account = account;
+            HasStateChanged = hasStateChanged;
+            IsApplicationCache = isApplicationCache;
+            SuggestedCacheKey = suggestedCacheKey;
+            HasTokens = hasTokens;
+            CancellationToken = cancellationToken;
+            CorrelationId = correlationId;
+            RequestScopes = requestScopes;
+            RequestTenantId = requestTenantId;
+            SuggestedCacheExpiry = suggestedCacheExpiry;
+            IdentityLogger = identityLogger;
+            PiiLoggingEnabled = piiLoggingEnabled;
+            CacheDetails = cacheDetails;
+        }
+
+        /// <summary>
         /// Gets the <see cref="ITokenCacheSerializer"/> involved in the transaction
         /// </summary>
         /// <remarks><see cref="TokenCache" > objects</see> implement this interface.</remarks>
@@ -248,5 +286,10 @@ namespace Microsoft.Identity.Client
         /// Boolean used to determine if Personally Identifiable Information (PII) logging is enabled.
         /// </summary>
         public bool PiiLoggingEnabled { get; }
+
+        /// <summary>
+        /// Cache Details contains the details of L1/ L2 cache for telemetry logging.
+        /// </summary>
+        public Dictionary<string, object> CacheDetails { get; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -129,7 +129,8 @@ namespace Microsoft.Identity.Client.Cache
                                   requestScopes: _requestParams.Scope, 
                                   requestTenantId: _requestParams.AuthorityManager.OriginalAuthority.TenantId,
                                   identityLogger: _requestParams.RequestContext.Logger.IdentityLogger,
-                                  piiLoggingEnabled: _requestParams.RequestContext.Logger.PiiLoggingEnabled);
+                                  piiLoggingEnabled: _requestParams.RequestContext.Logger.PiiLoggingEnabled,
+                                  cacheDetails: RequestContext.CacheDetails);
 
                                 stopwatch.Start();
                                 await TokenCacheInternal.OnBeforeAccessAsync(args).ConfigureAwait(false);
@@ -155,7 +156,8 @@ namespace Microsoft.Identity.Client.Cache
                                   requestScopes: _requestParams.Scope,
                                   requestTenantId: _requestParams.AuthorityManager.OriginalAuthority.TenantId,
                                   identityLogger: _requestParams.RequestContext.Logger.IdentityLogger,
-                                  piiLoggingEnabled: _requestParams.RequestContext.Logger.PiiLoggingEnabled);
+                                  piiLoggingEnabled: _requestParams.RequestContext.Logger.PiiLoggingEnabled,
+                                  cacheDetails: RequestContext.CacheDetails);
 
                                 await TokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);
                                 RequestContext.ApiEvent.DurationInCacheInMs += stopwatch.ElapsedMilliseconds;
--- a/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
@@ -177,7 +177,8 @@ namespace Microsoft.Identity.Client
                             requestScopes: requestParams.Scope,
                             requestTenantId: requestParams.AuthorityManager.OriginalAuthority.TenantId,
                             identityLogger: requestParams.RequestContext.Logger.IdentityLogger,
-                            piiLoggingEnabled: requestParams.RequestContext.Logger.PiiLoggingEnabled);
+                            piiLoggingEnabled: requestParams.RequestContext.Logger.PiiLoggingEnabled,
+                            cacheDetails: requestParams.RequestContext.CacheDetails);
 
                         Stopwatch sw = Stopwatch.StartNew();
 
@@ -249,7 +250,8 @@ namespace Microsoft.Identity.Client
                             requestScopes: requestParams.Scope,
                             requestTenantId: requestParams.AuthorityManager.OriginalAuthority.TenantId,
                             identityLogger: requestParams.RequestContext.Logger.IdentityLogger,
-                            piiLoggingEnabled: requestParams.RequestContext.Logger.PiiLoggingEnabled);
+                            piiLoggingEnabled: requestParams.RequestContext.Logger.PiiLoggingEnabled,
+                            cacheDetails: requestParams.RequestContext.CacheDetails);
 
                         Stopwatch sw = Stopwatch.StartNew();
                         await tokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);
@@ -741,7 +743,8 @@ namespace Microsoft.Identity.Client
                             requestScopes: null,
                             requestTenantId: null,
                             identityLogger: null,
-                            piiLoggingEnabled: false);
+                            piiLoggingEnabled: false,
+                            cacheDetails: new Dictionary<string, object>());
 
                 await tokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);
             }
@@ -1162,7 +1165,8 @@ namespace Microsoft.Identity.Client
                             requestScopes: requestParameters.Scope,
                             requestTenantId: requestParameters.AuthorityManager.OriginalAuthority.TenantId,
                             identityLogger: requestParameters.RequestContext.Logger.IdentityLogger,
-                            piiLoggingEnabled: requestParameters.RequestContext.Logger.PiiLoggingEnabled);
+                            piiLoggingEnabled: requestParameters.RequestContext.Logger.PiiLoggingEnabled,
+                            cacheDetails: requestParameters.RequestContext.CacheDetails);
 
 
                         await tokenCacheInternal.OnBeforeAccessAsync(args).ConfigureAwait(false);
@@ -1198,7 +1202,8 @@ namespace Microsoft.Identity.Client
                            requestScopes: requestParameters.Scope,
                            requestTenantId: requestParameters.AuthorityManager.OriginalAuthority.TenantId,
                            identityLogger: requestParameters.RequestContext.Logger.IdentityLogger,
-                            piiLoggingEnabled: requestParameters.RequestContext.Logger.PiiLoggingEnabled);
+                           piiLoggingEnabled: requestParameters.RequestContext.Logger.PiiLoggingEnabled,
+                           cacheDetails: requestParameters.RequestContext.CacheDetails);
 
 
                         await tokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);


src/client/Microsoft.Identity.Client/Internal/RequestContext.cs
```diff
@@ -24,6 +25,8 @@ internal class RequestContext
 
         public CancellationToken UserCancellationToken { get; }
 
+        public Dictionary<string, object> CacheDetails { get; } = new Dictionary<string, object>();
```
bgavrilMS, line 28: Can we name this `TelemetryDatapoints` or something more general purpose?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -142,6 +142,33 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                    DateTimeHelpers.DateTimeToUnixTimestampMilliseconds(authenticationResult.AuthenticationResultMetadata.RefreshOn.Value)
+                    : 0);
+
+                LogCacheDetailsToTelemetryClient(cacheDetails, telemetryEventDetails);
+            }
+        }
+
+        private void LogCacheDetailsToTelemetryClient(Dictionary<string, object> cacheDetails, MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            foreach (var cacheDetail in cacheDetails)
```
bgavrilMS, line 156: perf: enumerating in dictionary is O(n) operation. Better to use typical dictionary operations which are O(1)

See `TryGetValue` in particular.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -142,6 +142,33 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+            {
+                switch (cacheDetail.Key)
+                {
+                    case TelemetryCacheConstants.CacheUsed:
+                        telemetryEventDetails.SetProperty(TelemetryCacheConstants.CacheUsed, ((CacheUsed)cacheDetail.Value).ToString());
+                        break;
+                    case TelemetryCacheConstants.L1Latency:
+                        telemetryEventDetails.SetProperty(TelemetryCacheConstants.L1Latency, (long)cacheDetail.Value);
```
bgavrilMS, line 164: I think you need to sanitize the values, there is no guarantee that those objects are ints.

src/client/Microsoft.Identity.Client/Internal/RequestContext.cs
bgavrilMS: When you create the dictionary, specify the culture, so that key operations are case insensitive. https://stackoverflow.com/questions/13988643/case-insensitive-dictionary-with-string-key-type-in-c-sharp

bgavrilMS, line 28: Is it worth having `object` here and not `string`? I guess `object` is more generic, but will there ever be datapoints which are not strings?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /// <summary>
+        /// Indicates that the token was not cached
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Token was obtained from L1
+        /// </summary>
+        L1 = 1,
```
bgavrilMS, line 42: Should we differentiate between MSAL's internal cache and an external memory cache?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public enum CacheUsed
+    {
+        /// <summary>
+        /// Indicates that the token was not cached
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Token was obtained from L1
```
bgavrilMS, line 40: We should name this "MemoryCache" and "DistributedCache" instead of L1 / L2

bgavrilMS (changes requested):
Naming is only blocking change, code looks ok otherwise with minor changes suggested.

src/client/Microsoft.Identity.Client/Internal/RequestContext.cs
neha-bhargava: I kept it as object because the cache latency is logged as a long and cache used as an int. But I guess all of that can be converted to string. And convert them to long when logging as we are logging latencies as long.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs
neha-bhargava: Good point. We can add another enum to indicate which cache was used and record the latency in L1Latency or just DurationInCache. We can do this work as part of: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3738

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs
neha-bhargava: Sure, would make more sense.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    /// Indicates the cache which was used to get the token.
+    /// </summary>
+    public enum CacheUsed
+    {
+        /// <summary>
+        /// Indicates that the token was not cached
+        /// </summary>
+        None = 0,
```
bgavrilMS, line 37: I think that the "None" value here is misleading, because it combines several states: 

- no cache was used (e.g. cache miss)
- MSAL's internal caching was used 
- some other cache was used, but the cache did not emit any metrics

I think we'd need values for all these.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -142,6 +142,40 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+            }
+        }
+
+        private void LogDatapointsToTelemetryClient(Dictionary<string, string> telemetryDatapoints, MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            if ((telemetryDatapoints == null) || (telemetryDatapoints.Count() == 0)) return;
+
+            string value;
+
+            if (telemetryDatapoints.TryGetValue(TelemetryCacheConstants.CacheUsed, out value))
```
bgavrilMS, line 160: nit: 

```csharp
 if (telemetryDatapoints.TryGetValue(TelemetryCacheConstants.CacheUsed, out string value) && int.TryParse(value, out int cacheUsed)```

Comment (bgavrilMS) [reply src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs:160]: i.e. you can declare the out variable inline, e.g. `out string x`

Comment (neha-bhargava) [reply src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs]: Changed to class.

File: src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs (hunk 6 above)
Comment (neha-bhargava) [reply src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryCacheConstants.cs]: Added.

File: src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs [hunk 8]
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.TelemetryCore.TelemetryClient
+{
+    /// <summary>
+    /// Stores the cache details to log to <see cref="ITelemetryClient"/>.
+    /// </summary>
+    public class TelemetryDatapoints
```
bgavrilMS, line 11: How about `CacheTelemetryDatapoints` ?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -142,9 +142,24 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                    DateTimeHelpers.DateTimeToUnixTimestampMilliseconds(authenticationResult.AuthenticationResultMetadata.RefreshOn.Value)
+                    : 0);
+
+                LogDatapointsToTelemetryClient(telemetryDatapoints, telemetryEventDetails);
             }
         }
 
+        private void LogDatapointsToTelemetryClient(TelemetryDatapoints telemetryDatapoints, MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            if (telemetryDatapoints == null) return;
```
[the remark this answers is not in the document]
neha-bhargava: Ideally it will never be null as I instantiate this class in RequestContext so that we can capture latencies incrementally.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// <summary>
+    /// Indicates the cache which was used to get the token.
+    /// </summary>
+    public enum CacheUsed
+    {
+        /// <summary>
+        /// Indicates that the token was not found in the cache.
+        /// </summary>
+        CacheMiss = 0,
```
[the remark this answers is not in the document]
neha-bhargava: Yeah makes sense. I think we can keep MsalInternalCache as default here.

neha-bhargava:
Removed the CacheUsed variable as this can be inferred as:
```
if (L2Latency > 0) 
    CacheUsed = L2
else 
    CacheUsed = L1
```

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -30,7 +30,11 @@ internal static class TelemetryConstants
         public const string DurationInHttp = "DurationInHttp";
         public const string ActivityId = "ActivityId";
         public const string Resource = "Resource";
+        public const string RefreshOn = "RefreshOn";
+        public const string CacheUsed = "CacheUsed";
```
bgavrilMS, line 34: still needed?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
pmaytak: So this check is not needed then.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -142,9 +142,23 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
                 telemetryEventDetails.SetProperty(TelemetryConstants.PopToken, authenticationResult.TokenType.Equals(Constants.PoPTokenType));
                 telemetryEventDetails.SetProperty(TelemetryConstants.RemainingLifetime, (authenticationResult.ExpiresOn - DateTime.Now).TotalMilliseconds);
                 telemetryEventDetails.SetProperty(TelemetryConstants.ActivityId, authenticationResult.CorrelationId);
+                telemetryEventDetails.SetProperty(TelemetryConstants.RefreshOn,
```
pmaytak, line 145: Why are we logging this? What's reporting value of datetime when refresh has to happen? Would something like time remaining time until refresh (kind of like remaining lifetime until expiration above) or refresh on duration be more useful? I also think this value is always non-null.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
pmaytak: Do we log CacheUsed anywhere? Also in the scenario when static Msal is used (so MsalInternalCache only) we would always log 0 for L1Latency and 0 for DurationInCache. So from telemetry it would seem like cache is not used, but then DurationInHttp would also be 0 (since internal cache actually is used). I think we would have to measure and set L1Latency for internal cache if cache serialization is not enabled. (just brainstorming).

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -146,6 +147,44 @@ public sealed partial class TokenCacheNotificationArgs
             PiiLoggingEnabled = piiLoggingEnabled;
         }
 
+        /// <summary>
+        /// This constructor is for test purposes only. It allows apps to unit test their MSAL token cache implementation code.
+        /// </summary>
+        public TokenCacheNotificationArgs(    // only use this constructor in product code
```
TimHannMSFT, line 153: I'm confused on the comments (but same for other methods here so maybe a style thing? The method header says it's test purposes only and the other comment says it's only used for product code. That conflicts, no?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
TimHannMSFT: Will this ever store more stuff? e.g. aggregate errors or track latency of other sub operations? If no, agree with above comments to make type more descriptive. If yes, class header should be updated to reflect a broader scope than cache.

src/client/Microsoft.Identity.Client/Internal/RequestContext.cs
```diff
@@ -2,11 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
```
TimHannMSFT, line 5: nit: no longer used?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Total latency of L1 cache access. This data is captured in MSAL when accessing the internal cache or Microsoft.Identity.Web when accessing the memory cache.
+        /// </summary>
+        public long L1Latency { get; set; }
+
+        /// <summary>
+        /// Total latency of L2 cache access. This data is captured in Microsoft.Identity.Web when accessing the distributed cache.
+        /// </summary>
+        public long L2Latency { get; set; }
```
TimHannMSFT, line 21: I think we've talked before about whether L1 and L2 latency needs to be stored separately. In my conception it's just latency and then we'd log what the token source was to inform what kind of latency. I don't think we'd ever have the case where we found a token in L1 and still looked in L2 and if we have to look in L2 it should include the L1 look up since that's part of the latency e2e (imo).

SEAL folks seemed to be of the same opinion in "MSAL's schema logged to telemetry pipeline" thread.

bgavrilMS: Ack, @neha-bhargava  is pausing this until we have a clearer picture of the schema. Will discuss with E&D team on Tuesday. Transforming this into a draft PR.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,26 +139,23 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
                 telemetryEventDetails.SetProperty(TelemetryConstants.DurationInCache, authenticationResult.AuthenticationResultMetadata.DurationInCacheInMs);
                 telemetryEventDetails.SetProperty(TelemetryConstants.DurationInHttp, authenticationResult.AuthenticationResultMetadata.DurationInHttpInMs);
                 telemetryEventDetails.SetProperty(TelemetryConstants.Succeeded, true);
-                telemetryEventDetails.SetProperty(TelemetryConstants.PopToken, authenticationResult.TokenType.Equals(Constants.PoPTokenType));
+                telemetryEventDetails.SetProperty(TelemetryConstants.TokenType, nameof(AuthenticationRequestParameters.RequestContext.ApiEvent.TokenType));
```
bgavrilMS, line 142: What are the possible values of this?

src/client/Microsoft.Identity.Client/Internal/ClientCredential/IClientCredential.cs
```diff
@@ -11,12 +11,15 @@
...
 using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+using Microsoft.Identity.Client.TelemetryCore;
 using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Internal.ClientCredential
 {
     internal interface IClientCredential
     {
+        AssertionType AssertionType { get; }
```
bgavrilMS, line 21: This is a code smell. You already know the type by looking at the `obj.Type` or `typeof(obj)` where obj implements `IClientCredential`. We should as much as possible not complicate the object model for telemetry.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: It was requested by the folks. Maybe @rymeskar remembers what they plan to do with `RefreshOn` ?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -193,13 +204,29 @@ private ApiEvent InitializeApiEvent(string accountId)
...
+        private AssertionType GetAssertionType()
+        {
+            if (ServiceBundle.Config.UseManagedIdentity)
+            {
+                return AssertionType.MSI;
+            }
+
+            if (AuthenticationRequestParameters.OnBeforeTokenRequestHandler != null)
+            {
+                return AssertionType.TokenRequestHandler;
```
bgavrilMS, line 224: Let's not expose this one and just assume it's MSI. As it's only Azure SDK that need it.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -193,13 +204,29 @@ private ApiEvent InitializeApiEvent(string accountId)
...
+            }
+
+            return (AssertionType)ServiceBundle.Config.ClientCredential.AssertionType;
```
bgavrilMS, line 227: Possible null ref if `ClientCredential` is null (e.g. PCA?)

src/client/Microsoft.Identity.Client/TelemetryCore/AssertionType.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.TelemetryCore
+{
+    internal enum AssertionType
+    {
+        Certificate,
```
bgavrilMS, line 14: Please add `None` as default value and assign int values to each member + update the doc (add an Annex). Folks creating Kusto queries based on this data will need to know what the numbers mean.

bgavrilMS, line 14:
```suggestion
        CertificateWithoutSNI
```

src/client/Microsoft.Identity.Client/TelemetryCore/AssertionType.cs
```diff
@@ -0,0 +1,21 @@
...
+        CertificateWithSNI,
+        Secret,
+        UserProvided,
```
bgavrilMS, line 17:
```suggestion
        ClientAssertion,
```

src/client/Microsoft.Identity.Client/Internal/ClientCredential/CertificateAndClaimsClientCredential.cs
```diff
@@ -42,7 +46,9 @@ public CertificateAndClaimsClientCredential(X509Certificate2 certificate, IDicti
                 tokenEndpoint,
                 _claimsToSign,
                 _appendDefaultClaims);
-            
+
+            _sendX5c = sendX5C;
```
bgavrilMS, line 50: I don't think this will work.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
rymeskar: MSI will in the future be also needed by non-AzureSDK scenarios, or? For example, federated credentials

bgavrilMS: We are exposing MSI via higher level APIs @rymeskar e.g. 

```csharp
ManagedIdentityApplication app = ManagedIdentityApplicationBuilder.Create(user_assigned_identity_client_id).Build();
var token = await app.AcquireToken("https://azure.keyvault.com/.default");
```

Afaik federated credentials are just another type of client assertion, we did not have to expose new API in MSAL.

```csharp
var cca = ConfidentialClientApplicationBuilder....WithClientAssertion(tokenFromOtherIdp).Build();
```

rymeskar: These higher-Level APIs stay inside MSAL?

bgavrilMS: Yes, these APIs will stay in MSAL.

src/client/Microsoft.Identity.Client/Internal/ClientCredential/CertificateAndClaimsClientCredential.cs
trwalke: Oh right, I forgot that this is set at the application level so there can be interactions collisions between two requests.

src/client/Microsoft.Identity.Client/Internal/RequestContext.cs
```diff
@@ -24,6 +26,8 @@ internal class RequestContext
 
         public CancellationToken UserCancellationToken { get; }
 
+        public TelemetryDatapoints TelemetryDatapoints { get; } = new TelemetryDatapoints();
```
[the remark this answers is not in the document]
trwalke: This is the class that is exposed to IDWeb to gather additional telemetry related to the cache (L1 vs L2 for now). I think it is best to expose a type that we can add to as needed.

src/client/Microsoft.Identity.Client/Cache/CacheTypeUsed.cs
```diff
@@ -17,10 +17,10 @@ public enum CacheTypeUsed
         /// <summary>
         /// Specifies if the L1 cache is used
         /// </summary>
-        L1Cache,
+        L1Cache = 0,
```
bgavrilMS, line 20: Should there be a None value ?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Identity.Client.TelemetryCore.TelemetryClient
+{
+    /// <summary>
+    /// Stores the cache details to log to <see cref="ITelemetryClient"/>.
+    /// </summary>
+    public class TelemetryDatapoints
+    {
+        /// <summary>
+        /// Total latency of L1 cache access. This data is captured in MSAL when accessing the internal cache or Microsoft.Identity.Web when accessing the memory cache.
```
neha-bhargava, line 15:
```suggestion
        /// Type of cache used. This data is captured from MSAL or Microsoft.Identity.Web to log to telemetry.
```

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Identity.Client.Cache;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.TelemetryCore.TelemetryClient
+{
+    /// <summary>
+    /// Stores the cache details to log to <see cref="ITelemetryClient"/>.
```
neha-bhargava, line 10:
```suggestion
    /// Stores details to log to <see cref="ITelemetryClient"/>.
```

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
neha-bhargava: We can keep this as TelemetryDatapoints as we can use this class to capture more information in future.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -32,7 +32,13 @@ internal static class TelemetryConstants
...
+        public const string L1Latency = "L1Latency";
```
neha-bhargava, line 37: This can be removed as we are no longer logging L1Latency and L2Latency

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -147,6 +147,8 @@ public X509Certificate2 ClientCredentialCertificate
             }
         }
 
+
```
neha-bhargava, line 150: nit: remove extra line

src/client/Microsoft.Identity.Client/Cache/CacheTypeUsed.cs
neha-bhargava: Lets have default as None for scenario when no cache is used.

src/client/Microsoft.Identity.Client/Internal/ClientCredential/IClientCredential.cs
```diff
@@ -11,12 +11,15 @@
...
 using Microsoft.Identity.Client.OAuth2;
 using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+using Microsoft.Identity.Client.TelemetryCore;
 using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Internal.ClientCredential
 {
     internal interface IClientCredential
     {
+        AssertionType TelemetryDetails { get; }
```
neha-bhargava, line 21: `LogAssertionType` or `AssertionTypeForTelemetry` instead of `TelemetryDetails`?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,9 +139,20 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
                 telemetryEventDetails.SetProperty(TelemetryConstants.RemainingLifetime, (authenticationResult.ExpiresOn - DateTime.Now).TotalMilliseconds);
                 telemetryEventDetails.SetProperty(TelemetryConstants.ActivityId, authenticationResult.CorrelationId);
+                telemetryEventDetails.SetProperty(TelemetryConstants.RefreshOn, 
+                    authenticationResult.AuthenticationResultMetadata.RefreshOn.HasValue ?
+                    DateTimeHelpers.DateTimeToUnixTimestampMilliseconds(authenticationResult.AuthenticationResultMetadata.RefreshOn.Value)
+                    : 0);
+                telemetryEventDetails.SetProperty(TelemetryConstants.AssertionType, (int)AuthenticationRequestParameters.RequestContext.ApiEvent.AssertionType);
+                telemetryEventDetails.SetProperty(TelemetryConstants.Endpoint, AuthenticationRequestParameters.Authority.AuthorityInfo.CanonicalAuthority.ToString());
+
+                if (telemetryDatapoints.CacheTypeUsed != null)
```
neha-bhargava, line 152: Can we add a default value of the enum as None and check `telemetryDatapoints.CacheTypeUsed != CacheTypeUsed.None`? AFAIK the value of enum will never be null but the default value.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -153,56 +326,72 @@ private void AssertLoggedTelemetry(AuthenticationResult authenticationResult, Ms
...
-
-        public void Initialize()
-        {
-
-        }
-
-        public bool IsEnabled()
-        {
-            return true;
+            if (eventDetails.Properties.ContainsKey(TelemetryConstants.CacheUsed))
```
neha-bhargava, line 333: Assert that the cacheUsed is null when it is not set.

trwalke: I set it to the default value in the parameter so it will test that case.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
trwalke: Added cache used value instead based on discussions

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
bgavrilMS, line 153: We should not introduce a new ctor. Existing constructors are sufficient.

trwalke: This is a public class so I believe Neha's intention was to avoid breaking changes. I guess I can add it with a default value.

TimHannMSFT:
@GeoK and others during the hackathon saw quite a bit of success with OpenTelemetry which may have implications for continued changes here. Would ask whoever's owning this to sync with him before merging.

trwalke:
@bgavrilMS please see the last commit, I updated the PR according to your last comment [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3784#issuecomment-1476524161).

Can you verify if this is correct?

Im assuming scopes should be in a JSON format?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,9 +139,34 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                ParseScopesForTelemetry(telemetryEventDetails);
+            }
+        }
+
+        private void ParseScopesForTelemetry(MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            if (AuthenticationRequestParameters.Scope.Count > 0)
+            {
+                var firstScope = AuthenticationRequestParameters.Scope.First().ToString();
+                if (Uri.IsWellFormedUriString(firstScope, UriKind.Absolute))
```
trwalke, line 161: Im assuming here that if a resource is set, it will be the only element in the array.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,9 +139,34 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                {
+                    telemetryEventDetails.SetProperty(TelemetryConstants.Resource, firstScope);
+                }
+                else
+                {
+                    telemetryEventDetails.SetProperty(TelemetryConstants.Scopes, JsonHelper.SerializeToJson(AuthenticationRequestParameters.Scope));
```
pmaytak, line 167: Why serialize as Json and not just as string?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -193,13 +218,41 @@ private ApiEvent InitializeApiEvent(string accountId)
...
             EnrichTelemetryApiEvent(apiEvent);
 
             return apiEvent;
         }
 
+        private AssertionType GetAssertionType()
+        {
+            if (!string.IsNullOrWhiteSpace(ServiceBundle.Config.ManagedIdentityUserAssignedClientId) ||
+                !string.IsNullOrWhiteSpace(ServiceBundle.Config.ManagedIdentityUserAssignedResourceId) ||
+                ServiceBundle.Config.AppTokenProvider != null)
```
pmaytak, line 233: Is AppTokenProvider always Msi? Technically no?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -193,13 +218,41 @@ private ApiEvent InitializeApiEvent(string accountId)
...
+            if (ServiceBundle.Config.ClientCredential != null)
+            {
+                if (ServiceBundle.Config.ClientCredential.TelemetryAssertionType == AssertionType.CertificateWithoutSni)
+                {
+                    if (ServiceBundle.Config.SendX5C)
+                    {
+                        return AssertionType.CertificateWithSni;
+                    }
+
+                    return AssertionType.CertificateWithoutSni;
```
pmaytak, line 247: nit: can this logic be moved into CertificateAndClaimsClientCredential.AssertionType? It should know about sendX5c

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Stores details to log to the <see cref="ITelemetryClient"/>.
+    /// </summary>
+    public class TelemetryDatapoints
+    {
+        /// <summary>
+        /// Type of cache used. This data is captured from MSAL or Microsoft.Identity.Web to log to telemetry.
+        /// </summary>
+        public CacheTypeUsed CacheTypeUsed { get; set; } = CacheTypeUsed.None;
```
pmaytak, line 17: What does `Used` mean? Cache read? Read or write? Or if L2 is set up at all? The life time of this class is per request, but both L1 and L2 can be used.

src/client/Microsoft.Identity.Client/Cache/CacheTypeUsed.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Cache
+{
+    /// <summary>
+    /// Identifies the type of cache used when accessing the cache.
```
bgavrilMS, line 13:
```suggestion
    /// Identifies the type of cache used when accessing the cache. Cache implementations must provide this.
```

src/client/Microsoft.Identity.Client/Cache/CacheTypeUsed.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    public enum CacheTypeUsed
+    {
+        /// <summary>
+        /// Specifies that no cache type is used
+        /// </summary>
+        None = 0,
```
bgavrilMS, line 20: Unknown?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -96,7 +96,7 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                     LogReturnedToken(authenticationResult);
                     UpdateTelemetry(sw, apiEvent, authenticationResult);
                     LogMetricsFromAuthResult(authenticationResult, AuthenticationRequestParameters.RequestContext.Logger);
-                    LogSuccessfulTelemetryToClient(authenticationResult, telemetryEventDetails, telemetryClients);
+                    LogSuccessfulTelemetryToClient(authenticationResult, AuthenticationRequestParameters.RequestContext.TelemetryDatapoints, telemetryEventDetails, telemetryClients);
```
bgavrilMS, line 99: Can you consolidate all these telemetry objects? I don't think we want to have both datapoints and eventdetails.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,9 +139,34 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                telemetryEventDetails.SetProperty(TelemetryConstants.AssertionType, (int)AuthenticationRequestParameters.RequestContext.ApiEvent.AssertionType);
+                telemetryEventDetails.SetProperty(TelemetryConstants.Endpoint, AuthenticationRequestParameters.Authority.AuthorityInfo.CanonicalAuthority.ToString());
+                telemetryEventDetails.SetProperty(TelemetryConstants.CacheUsed, (int)telemetryDatapoints.CacheTypeUsed);
+                ParseScopesForTelemetry(telemetryEventDetails);
+            }
+        }
+
+        private void ParseScopesForTelemetry(MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            if (AuthenticationRequestParameters.Scope.Count > 0)
```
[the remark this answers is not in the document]
bgavrilMS: The spec mentions: 
```
resource - the uri
scopes - scopes without the resource uri
```

This is to avoid putting string that are too long into the telemetry storage.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -144,7 +334,17 @@ public async Task AcquireTokenUnSuccessfulTelemetryTestAsync()
...
+        private void AssertLoggedTelemetry(
+                        AuthenticationResult authenticationResult, 
+                        MsalTelemetryEventDetails eventDetails, 
+                        TokenSource tokenSource, 
+                        CacheRefreshReason cacheRefreshReason,
+                        AssertionType assertionType,
+                        string endpoint,
+                        TokenType? tokenType = TokenType.Bearer,
+                        CacheTypeUsed cacheTypeUsed = CacheTypeUsed.None,
+                        string scopes = null,
```
bgavrilMS, line 346: Why don't you assert the resource and scope everytime?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -144,7 +334,17 @@ public async Task AcquireTokenUnSuccessfulTelemetryTestAsync()
...
 
-        private void AssertLoggedTelemetry(AuthenticationResult authenticationResult, MsalTelemetryEventDetails eventDetails, TokenSource tokenSource, CacheRefreshReason cacheRefreshReason)
+        private void AssertLoggedTelemetry(
+                        AuthenticationResult authenticationResult, 
+                        MsalTelemetryEventDetails eventDetails, 
+                        TokenSource tokenSource, 
+                        CacheRefreshReason cacheRefreshReason,
+                        AssertionType assertionType,
+                        string endpoint,
+                        TokenType? tokenType = TokenType.Bearer,
```
bgavrilMS, line 344: why don't you assert the token type everytime?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
bgavrilMS: We should have a few validation tests for this: 

```
Input: https:\\graph.microsoft.com\.default
Resource: https:\\graph.microsoft.com
Scopes: .default

Input: https:\\graph.microsoft.com\User.Read https:\\graph.microsoft.com\Mail.Read 
Resource: https:\\graph.microsoft.com
Scopes: User.Read Mail.Read

Input: User.Read Mail.Read
Resource: graph
Scopes: User.Read Mail.Read

Input: https:\\sharepoint.com\scope
Resource: https:\\sharepoint.com\
Scopes: scope

File: src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs [hunk 43]
```diff
@@ -139,9 +139,34 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+            {
+                var firstScope = AuthenticationRequestParameters.Scope.First().ToString();
+                if (Uri.IsWellFormedUriString(firstScope, UriKind.Absolute))
+                {
+                    telemetryEventDetails.SetProperty(TelemetryConstants.Resource, firstScope);
```
bgavrilMS, line 163: I don't think this is right, see the spec above.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,9 +139,34 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                telemetryEventDetails.SetProperty(TelemetryConstants.CacheUsed, (int)telemetryDatapoints.CacheTypeUsed);
+                ParseScopesForTelemetry(telemetryEventDetails);
+            }
+        }
+
+        private void ParseScopesForTelemetry(MsalTelemetryEventDetails telemetryEventDetails)
+        {
+            if (AuthenticationRequestParameters.Scope.Count > 0)
+            {
+                var firstScope = AuthenticationRequestParameters.Scope.First().ToString();
```
bgavrilMS, line 160: Possible null ref (see tests above)

bgavrilMS (changes requested):
Needs more in depth look / tests around resource + scope

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -98,15 +109,24 @@ public async Task AcquireTokenSuccessfulTelemetryTestAsync()
                 CreateApplication();
                 _harness.HttpManager.AddMockHandlerSuccessfulClientCredentialTokenResponseMessage();
 
-                // Acquire token interactively
+                // Acquire token interactively with scope
```
bgavrilMS, line 112: nit: this is client_credes, not acquire token interactively

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -98,15 +109,24 @@ public async Task AcquireTokenSuccessfulTelemetryTestAsync()
...
                 var result = await _cca.AcquireTokenForClient(TestConstants.s_scope)
                     .WithAuthority(TestConstants.AuthorityUtidTenant)
                     .ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
 
                 Assert.IsNotNull(result);
 
                 MsalTelemetryEventDetails eventDetails = _telemetryClient.TestTelemetryEventDetails;
-                AssertLoggedTelemetry(result, eventDetails, TokenSource.IdentityProvider, CacheRefreshReason.NoCachedAccessToken);
+                AssertLoggedTelemetry(
```
bgavrilMS, line 120: For the fuuture - it helps reviewers when the tests are easier to read. Keep prod code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), test code can be "more" WET to increase readability. 

For example, if the assertion was more explicit, smth like:

```
Assert.AreEqual(CacheLevel.L2.ToString(), telemetry["cache_level"]);
Assert.AreEqual("https:\\graph.microsoft.com", telemetry["resource"]);
```

bgavrilMS:
Is this missing the error datapoints? If it does, I think you should have a separate PR as this one is getting a bit big.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
pmaytak: Thanks, yes, examples help. My first thought top of mind is the concern for all the string parsing. The other scope parsing logic that we have already also shows up on flamegraphs as high CPU usage. Now since this will be done only when telemetry is enabled, maybe it's not too bad; but again if we want telemetry to be used widely, it'll be an issue. This is also done for every request.

This case means I'll have to parse every scope?
```text
Input: https:\\graph.microsoft.com\User.Read https:\\graph.microsoft.com\Mail.Read
Resource: https:\\graph.microsoft.com
Scopes: User.Read Mail.Read
```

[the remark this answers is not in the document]
trwalke: > We should have a few validation tests for this. Let me know if any of these don't make sense.
> 
> ```
> Input: https:\\graph.microsoft.com\.default
> Resource: https:\\graph.microsoft.com
> Scopes: .default
> 
> Input: https:\\graph.microsoft.com\User.Read https:\\graph.microsoft.com\Mail.Read 
> Resource: https:\\graph.microsoft.com
> Scopes: User.Read Mail.Read
> 
> Input: User.Read Mail.Read
> Resource: 
> Scopes: User.Read Mail.Read
> 
> Input: https:\\sharepoint.com\scope
> Resource: https:\\sharepoint.com\
> Scopes: scope
> 
> Input: null
> Resource: null
> Scopes: null
> 
> Input: any combo of the 3 magic scopes (offline_access, openid, profile)
> Resource: null
> Scopes: null
> ```
> 
> I'm hoping that we don't have to do extra parsing for the last 2 use cases because MSAL alraedy has scope merging logic.
> 
> The use cases above should be covered by tests.

Ah, ok ,this makes sense. ill go ahead and update the logic to reflect this

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
trwalke: The spec mentions that we want a distinction between scopes and resources

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke: Technically no but we discussed this and the decision was made to make it Msi. It should be Msi almost all of the time

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
trwalke: Yea, that makes sense, I can seperate the values so both levels can be set.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,9 +139,34 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
                 telemetryEventDetails.SetProperty(TelemetryConstants.DurationInHttp, authenticationResult.AuthenticationResultMetadata.DurationInHttpInMs);
                 telemetryEventDetails.SetProperty(TelemetryConstants.Succeeded, true);
-                telemetryEventDetails.SetProperty(TelemetryConstants.PopToken, authenticationResult.TokenType.Equals(Constants.PoPTokenType));
+                telemetryEventDetails.SetProperty(TelemetryConstants.TokenType, (int)AuthenticationRequestParameters.RequestContext.ApiEvent.TokenType);
                 telemetryEventDetails.SetProperty(TelemetryConstants.RemainingLifetime, (authenticationResult.ExpiresOn - DateTime.Now).TotalMilliseconds);
                 telemetryEventDetails.SetProperty(TelemetryConstants.ActivityId, authenticationResult.CorrelationId);
+                telemetryEventDetails.SetProperty(TelemetryConstants.RefreshOn, 
+                    authenticationResult.AuthenticationResultMetadata.RefreshOn.HasValue ?
+                    DateTimeHelpers.DateTimeToUnixTimestampMilliseconds(authenticationResult.AuthenticationResultMetadata.RefreshOn.Value)
+                    : 0);
```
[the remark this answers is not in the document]
trwalke: indeed. good catch

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -22,17 +22,23 @@ internal static class TelemetryConstants
...
         public const string Duration = "Duration";
         public const string Succeeded = "Succeeded";
         public const string DurationInCache = "DurationInCache";
         public const string DurationInHttp = "DurationInHttp";
         public const string ActivityId = "ActivityId";
         public const string Resource = "Resource";
+        public const string RefreshOn = "RefreshOn";
+        public const string CacheUsed = "CacheUsed";
+        public const string AssertionType = "AssertionType";
+        public const string Endpoint = "Endpoint";
```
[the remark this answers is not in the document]
trwalke: This is the name agreed upon in the spec. Am I at liberty to change it? @bgavrilMS

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke: I was trying to find a reasonable format for the list of scopes in case it would be parsed later. I guess I can use Scopes.AsSingleString()

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke: Yeah, I was missing some of the specifics on how to parse the scopes. Updating logic.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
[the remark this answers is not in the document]
bgavrilMS: Let's keep it as is for now.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
trwalke: > 



> Thanks, yes, examples help. My first thought top of mind is the concern for all the string parsing. The other scope parsing logic that we have already also shows up on flamegraphs as high CPU usage. Now since this will be done only when telemetry is enabled, maybe it's not too bad; but again if we want telemetry to be used widely, it'll be an issue. This is also done for every request.
> 
> This case means I'll have to parse every scope?
> 
> ```
> Input: https:\\graph.microsoft.com\User.Read https:\\graph.microsoft.com\Mail.Read
> Resource: https:\\graph.microsoft.com
> Scopes: User.Read Mail.Read
> ```

I guess it would depend on whether or not the host portion of the scopes are the same for all scopes. If so, only one scope needs to be parsed and the extracted host can simply be removed from the other scopes with something like `string.Split(host)`. 

However, if we have something like the following then we may have to parse all of them.

 ```
 Input: https:\\hostA\User.Read https:\\HostB\Mail.Read
 Resource: https:\\hostA https:\\hostB
 Scopes: User.Read Mail.Read
 ```
 
 @bgavrilMS Will the host portion of the scopes all be the same?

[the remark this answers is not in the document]
trwalke: @pmaytak discussed this with Bogdan, We only need to parse one of the Uri and then make the assumption that other scopes will contain the same host

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
trwalke: Actually, i added a new entry to the enum to indicate when both caches are used.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke: The data points is what is exposed to the cache for IDWeb/other SDKs. I figured it would be better to keep the object in case we wanted to expand the data we request from the cache.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -22,17 +22,23 @@ internal static class TelemetryConstants
...
         public const string ErrorCode = "ErrorCode";
+        public const string ErrorMessage = "ErrorMessage";
         public const string Duration = "Duration";
         public const string Succeeded = "Succeeded";
         public const string DurationInCache = "DurationInCache";
         public const string DurationInHttp = "DurationInHttp";
         public const string ActivityId = "ActivityId";
         public const string Resource = "Resource";
+        public const string RefreshOn = "RefreshOn";
+        public const string CacheUsed = "CacheUsed";
```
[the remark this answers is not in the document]
trwalke: `Cachelevel` is better

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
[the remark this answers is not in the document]
trwalke: @bgavrilMS cant do this test since null cant be used as a scope with confidential client
```
Input: null
Resource: null
Scopes: null
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
trwalke: The CertificateAndClaimsClientCredential lasts for the duration of the app but Send x5c can change which is why it is not stored within the CertificateAndClaimsClientCredential object.

src/client/Microsoft.Identity.Client/Cache/CacheTypeUsed.cs
```diff
@@ -10,21 +10,25 @@
...
         L1Cache = 1,
         /// <summary>
         /// Specifies if the L2 cache is used
         /// </summary>
-        L2Cache = 2
+        L2Cache = 2,
+        /// <summary>
+        /// Specifies if both the L1 and L2 caches are used
+        /// </summary>
+        L1AndL2Cache = 3
```
bgavrilMS, line 32: As discussed offline, this is not possible. We are only interested in READs, and a READ operation can happen in L1 or in L2, but not in both.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: I suggest consolidating them. This detail should not bleed throughout all our codebase.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -140,11 +425,41 @@ public async Task AcquireTokenUnSuccessfulTelemetryTestAsync()
...
+
+                Assert.IsNotNull(exClient);
+                Assert.IsNotNull(exClient.ErrorCode);
+
+                eventDetails = _telemetryClient.TestTelemetryEventDetails;
+                Assert.AreEqual(exClient.ErrorCode, eventDetails.Properties[TelemetryConstants.ErrorCode]);
+                Assert.AreEqual(exClient.Message, eventDetails.Properties[TelemetryConstants.ErrorMessage]);
+                Assert.IsFalse((bool?)eventDetails.Properties[TelemetryConstants.Succeeded]);
+
+                //Test for generic Exception
```
trwalke, line 448: Trying to figure out the best way to simulate a generic exception for this test.

trwalke:
cc @neha-bhargava for one final look

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -120,16 +121,38 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
             }
         }
 
-        private void LogErrorTelemetryToClient(string errorCode, MsalTelemetryEventDetails telemetryEventDetails, ITelemetryClient[] telemetryClients)
+        private void LogMsalErrorTelemetryToClient(MsalException ex, MsalTelemetryEventDetails telemetryEventDetails, ITelemetryClient[] telemetryClients)
+        {
+            if (telemetryClients.HasEnabledClients(TelemetryConstants.AcquireTokenEventName))
+            {
+                if (ex is MsalClientException clientException)
```
bgavrilMS, line 128: For all other exceptions, we should still log the exception type and message

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
bgavrilMS: How about - inject a custom HttpClient (via `WithHttpManager` or `WithHttpClientFactory`). See example here https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/PublicApiTests/AccountTests.cs#L59 (here it asserts, you can just throw a SocketNotFoundException or smth)_

bgavrilMS (approved):
Approved with comments. 

@neha-bhargava  - can you please also show Travis how you tested this E2E so that we can see some values in Kusto ? Some manual validation would be great.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -341,8 +341,6 @@ public ConfidentialClientApplicationBuilder WithGenericAuthority(string authorit
         /// <returns>The builder to chain the .With methods</returns>
         public ConfidentialClientApplicationBuilder WithTelemetryClient(params ITelemetryClient[] telemetryClients)
         {
-            ValidateUseOfExperimentalFeature("ITelemetryClient");
```
pmaytak, line 344: Created a separate issue for tracking for GAing #4119

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -104,13 +104,14 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                 {
                     apiEvent.ApiErrorCode = ex.ErrorCode;
                     AuthenticationRequestParameters.RequestContext.Logger.ErrorPii(ex);
-                    LogErrorTelemetryToClient(ex.ErrorCode, telemetryEventDetails, telemetryClients);
+                    LogMsalErrorTelemetryToClient(ex, telemetryEventDetails, telemetryClients);
```
neha-bhargava, line 107: Can we combine logging all the errors in a single method? `LogErrorToTelemetryClient`?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: They are logged in a separate method. My suggestion is to combine in one method.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -120,12 +121,34 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
             }
         }
 
-        private void LogErrorTelemetryToClient(string errorCode, MsalTelemetryEventDetails telemetryEventDetails, ITelemetryClient[] telemetryClients)
+        private void LogMsalErrorTelemetryToClient(MsalException ex, MsalTelemetryEventDetails telemetryEventDetails, ITelemetryClient[] telemetryClients)
```
neha-bhargava, line 124: Can also be written as 
```
private void LogMsalErrorTelemetryToClient(MsalException msalException, MsalTelemetryEventDetails telemetryEventDetails, ITelemetryClient[] telemetryClients)
        {
            if (telemetryClients.HasEnabledClients(TelemetryConstants.AcquireTokenEventName))
            {
                telemetryEventDetails.SetProperty(TelemetryConstants.Succeeded, false);
                telemetryEventDetails.SetProperty(TelemetryConstants.ErrorCode, msalException.ErrorCode);
                telemetryEventDetails.SetProperty(TelemetryConstants.ErrorMessage, msalException.Message);

                if (msalException is MsalServiceException serviceException)
                {
                    telemetryEventDetails.SetProperty(TelemetryConstants.StsErrorCode, serviceException.ErrorCodes.AsSingleString());
                }
            }
        }
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,12 +162,64 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                else
+                {
+                    telemetryEventDetails.SetProperty(TelemetryConstants.Scopes, AuthenticationRequestParameters.Scope.AsSingleString());
+                }
+            }
+        }
+
+        private CacheLevel GetCacheLevel(AuthenticationResult authenticationResult)
+        {
+            if (authenticationResult.AuthenticationResultMetadata.TokenSource == TokenSource.Cache) //Check if token scource is cache
```
neha-bhargava, line 209: Nit
```suggestion
            if (authenticationResult.AuthenticationResultMetadata.TokenSource == TokenSource.Cache) //Check if token source is cache
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -193,13 +268,42 @@ private ApiEvent InitializeApiEvent(string accountId)
...
 
             // Give derived classes the ability to add or modify fields in the telemetry as needed.
             EnrichTelemetryApiEvent(apiEvent);
 
             return apiEvent;
         }
 
+        private AssertionType GetAssertionType()
+        {
+            if (!string.IsNullOrWhiteSpace(ServiceBundle.Config.ManagedIdentityUserAssignedClientId) ||
```
neha-bhargava, line 281: You can change this comdition to 
`if (ServiceBundle.Config.IsManagedIdentity)`

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/TelemetryDatapoints.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using Microsoft.Identity.Client.Cache;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Client.TelemetryCore.TelemetryClient
+{
+    /// <summary>
+    /// Stores details to log to the <see cref="ITelemetryClient"/>.
+    /// </summary>
+    public class TelemetryData
```
neha-bhargava, line 12: The name of the file is still TelemetryDatapoints.cs.

bgavrilMS:
@trwalke - please create a work item on ID.Web to track the work in the cache project.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -22,17 +22,24 @@ internal static class TelemetryConstants
...
+        public const string StsErrorCode = "StsErrorCode";
+        public const string ErrorMessage = "ErrorMessage";
         public const string Duration = "Duration";
         public const string Succeeded = "Succeeded";
         public const string DurationInCache = "DurationInCache";
         public const string DurationInHttp = "DurationInHttp";
         public const string ActivityId = "ActivityId";
         public const string Resource = "Resource";
+        public const string RefreshOn = "RefreshOn";
+        public const string CacheLevel = "CacheLevel";
```
pmaytak, line 38:
```suggestion
        public const string CacheSourceLevel = "CacheSourceLevel";
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -139,12 +162,64 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
+                if (Uri.IsWellFormedUriString(firstScope, UriKind.Absolute))
+                {
+                    Uri firstScopeAsUri = new Uri(firstScope);
+                    telemetryEventDetails.SetProperty(TelemetryConstants.Resource, $"{firstScopeAsUri.Scheme}://{firstScopeAsUri.Host}");
+
+                    StringBuilder stringBuilder = new StringBuilder();
+
+                    foreach (string scope in AuthenticationRequestParameters.Scope)
+                    {
+                         stringBuilder.Append(scope.Split(new[] { firstScopeAsUri.Host }, StringSplitOptions.None)[1].Replace("/", string.Empty) + " ");
```
pmaytak, line 195:
```suggestion
                         stringBuilder.Append(scope.Split(new[] { firstScopeAsUri.Host }, StringSplitOptions.None)[1].TrimStart('/') + " ");
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -116,7 +125,282 @@ public async Task AcquireTokenSuccessfulTelemetryTestAsync()
...
+        }
+
+        [TestMethod]
+        [DataRow(new[] { "https://graph.microsoft.com/.default" }, "https://graph.microsoft.com", ".default")]
+        [DataRow(new[] { "https://graph.microsoft.com/User.Read", "https://graph.microsoft.com/Mail.Read" }, "https://graph.microsoft.com", "User.Read Mail.Read")]
+        [DataRow(new[] { "api://23c64cd8-21e4-41dd-9756-ab9e2c23f58c/access_as_user" }, "api://23c64cd8-21e4-41dd-9756-ab9e2c23f58c", "access_as_user")]
+        [DataRow(new[] { "User.Read", "Mail.Read" }, null, "User.Read Mail.Read")]
+        [DataRow(new[] { "https://sharepoint.com/scope" }, "https://sharepoint.com", "scope")]
+        [DataRow(new[] { "offline_access", "openid", "profile" }, null, "offline_access openid profile")]
+        public async Task AcquireTokenSuccessfulTelemetryTestForScopesAsync(IEnumerable<string> intput, string expectedResource, string expectedScope)
```
pmaytak, line 171:
```suggestion
        public async Task AcquireTokenSuccessfulTelemetryTestForScopesAsync(IEnumerable<string> input, string expectedResource, string expectedScope)
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -116,7 +125,282 @@ public async Task AcquireTokenSuccessfulTelemetryTestAsync()
...
+                    TestConstants.AuthorityUtidTenant,
+                    TokenType.Bearer,
+                    CacheLevel.None,
+                    null,
+                    "https://graph.windows.net");
+            }
+        }
+
+        [TestMethod]
+        [DataRow(new[] { "https://graph.microsoft.com/.default" }, "https://graph.microsoft.com", ".default")]
```
pmaytak, line 165: Add 
`[DataRow(new[] { "https://graph.microsoft.com/.default", "User.Read" }, "https://graph.microsoft.com", ".default User.Read")]`

src/client/Microsoft.Identity.Client/Cache/CacheLevel.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Cache
+{
+    /// <summary>
+    /// Identifies the type of cache used when accessing the cache. Cache implementations must provide this.
+    /// </summary>
+    public enum CacheLevel
```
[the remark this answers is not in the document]
trwalke: Hmm, not sure about the cache "source" level. What prompted you to add source?

pmaytak (approved):

src/client/Microsoft.Identity.Client/Cache/CacheLevel.cs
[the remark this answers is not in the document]
pmaytak: My initial thoughts were that CacheLevel is too general, so was thinking about making it more specific. But this will work.

Issue #3745, opened 2022-10-12
[Feature Request] Add MSAL version to TelemetryClient logging


neha-bhargava:
Add MSAL version to used to the telemetry pipeline logging

PR #3747, opened 2022-10-12
Log MsalVersion to ITelemetryClient

neha-bhargava:
Fixes #
#3745 

Log MSAL version

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -352,11 +352,11 @@ namespace Microsoft.Identity.Client
 
         private void TelemetryClientLogMsalVersion()
         {
-            MsalTelemetryEventDetails telemetryEventDetails = new MsalTelemetryEventDetails(TelemetryConstants.ConfigurationUpdateEventName);
-            telemetryEventDetails.SetProperty(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion());
-
             if (Config.TelemetryClients.HasEnabledClients(TelemetryConstants.ConfigurationUpdateEventName))
             {
+                MsalTelemetryEventDetails telemetryEventDetails = new MsalTelemetryEventDetails(TelemetryConstants.ConfigurationUpdateEventName);
+                telemetryEventDetails.SetProperty(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion());
+
                 Config.TelemetryClients.TrackEvent(telemetryEventDetails);
             }
         }
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -355,12 +355,9 @@ namespace Microsoft.Identity.Client
             MsalTelemetryEventDetails telemetryEventDetails = new MsalTelemetryEventDetails(TelemetryConstants.ConfigurationUpdateEventName);
             telemetryEventDetails.SetProperty(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion());
 
-            foreach (var telemetryClient in Config.TelemetryClients)
+            if (Config.TelemetryClients.HasEnabledClients(TelemetryConstants.ConfigurationUpdateEventName))
             {
-                if (telemetryClient.IsEnabled(TelemetryConstants.ConfigurationUpdateEventName))
-                {
-                    telemetryClient.TrackEvent(telemetryEventDetails);
-                }
+                Config.TelemetryClients.TrackEvent(telemetryEventDetails);
             }
         }
 
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -10,6 +10,8 @@ using System.Threading.Tasks;
 using Microsoft.Identity.Client.Extensibility;
 using Microsoft.Identity.Client.Internal;
 using Microsoft.Identity.Client.Internal.ClientCredential;
+using Microsoft.Identity.Client.TelemetryCore;
+using Microsoft.Identity.Client.TelemetryCore.TelemetryClient;
 using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.Identity.Client
@@ -343,7 +345,23 @@ namespace Microsoft.Identity.Client
                 Config.TelemetryClients = telemetryClients;
             }
 
+            TelemetryClientLogMsalVersion();
+
             return this;
+        }
+
+        private void TelemetryClientLogMsalVersion()
+        {
+            MsalTelemetryEventDetails telemetryEventDetails = new MsalTelemetryEventDetails(TelemetryConstants.ConfigurationUpdateEventName);
+            telemetryEventDetails.SetProperty(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion());
+
+            foreach (var telemetryClient in Config.TelemetryClients)
+            {
+                if (telemetryClient.IsEnabled(TelemetryConstants.ConfigurationUpdateEventName))
+                {
+                    telemetryClient.TrackEvent(telemetryEventDetails);
+                }
+            }
         }
 
         internal ConfidentialClientApplicationBuilder WithAppTokenCacheInternalForTest(ITokenCacheInternal tokenCacheInternal)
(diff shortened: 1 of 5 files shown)


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -343,9 +345,25 @@ public ConfidentialClientApplicationBuilder WithTelemetryClient(params ITelemetr
...
             }
 
+            TelemetryClientLogMsalVersion();
+
             return this;
         }
 
+        private void TelemetryClientLogMsalVersion()
+        {
+            MsalTelemetryEventDetails telemetryEventDetails = new MsalTelemetryEventDetails(TelemetryConstants.ConfigurationUpdateEventName);
```
pmaytak, line 355: Do we need this event, if MSAL version is sent with every acquire token event?
Additionally, if we upload this event, when querying data how can we correlate MSAL version to the app, since we don't send like client ID?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
```diff
@@ -153,6 +153,7 @@ private void AssertLoggedTelemetry(AuthenticationResult authenticationResult, Ms
             Assert.AreEqual(authenticationResult.AuthenticationResultMetadata.DurationInHttpInMs, eventDetails.Properties[TelemetryConstants.DurationInHttp]);
             Assert.AreEqual(authenticationResult.AuthenticationResultMetadata.DurationInCacheInMs, eventDetails.Properties[TelemetryConstants.DurationInCache]);
             Assert.AreEqual(authenticationResult.AuthenticationResultMetadata.DurationTotalInMs, eventDetails.Properties[TelemetryConstants.Duration]);
+            Assert.AreEqual(typeof(ConfidentialClientApplication).Assembly.GetName().Version.ToString(), eventDetails.Properties[TelemetryConstants.MsalVersion]);
```
pmaytak, line 156: Use MsalIdHelper.GetMsalVersion() instead? Plus wouldn't the assembly version be different than Msal Version (ex 4.46.2.0 vs 4.46.2), so this should fail?

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryClient/MsalTelemetryEventDetails.cs
```diff
@@ -7,9 +7,9 @@ namespace Microsoft.Identity.Client.TelemetryCore.TelemetryClient
 {
     internal class MsalTelemetryEventDetails : TelemetryEventDetails
     {
-        public MsalTelemetryEventDetails()
+        public MsalTelemetryEventDetails(string eventName)
```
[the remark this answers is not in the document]
bgavrilMS: We do not enforce this in the codebase though. internal class + public method is the same as internal class + internal method.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
bgavrilMS: The consumer is very COGS sensitive, so they don't want extra data like client_id. We do have correlation id at the moment, but not sure if it will stay. This is a good question to follow up with the partners wanting to consume this.

neha-bhargava: We don't log the client id but it is logged by default in the telemetry client as when a telemetry client object is created with a client id.

neha-bhargava: Yes, correct. Client id is logged in all the tables by the telemetry client.

pmaytak (approved):

pmaytak (approved):
Main feedback is then to not send version with every request and only once as part of config_update event.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs
neha-bhargava: Removed this test as no longer relevant

neha-bhargava:
> Main feedback is then to not send version with every request and only once as part of config_update event.

Missed to remove that earlier. Removed now. Thanks

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -343,9 +345,25 @@ public ConfidentialClientApplicationBuilder WithTelemetryClient(params ITelemetr
...
+            telemetryEventDetails.SetProperty(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion());
+
+            foreach (var telemetryClient in Config.TelemetryClients)
```
jmprieur, line 358: Shouldn't the code in the foreach be factorized somewhere?
This seems like a lot of responsibility for the application.

cc: @jennyf19 as we were discussing the R9 telemetry.

SameerK-MSFT:
Looking at the API "WithTelemetryClient", https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/86ffe422e668fc8fb88c4672d764a70cd8de8181/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L322 two thoughts.

1. Since this can accept one or more ITelemetryClient objects, can this be renamed to plural version WithTelemetryClient**s**.
2. Since most developers will use a single Telemetry client, an override can be added that accept single client. This will keep API simple for the developers. This will be a breaking change. But if minimal customers are using it, it may be worth. Thoughts?

pmaytak:
> Looking at the API "WithTelemetryClient",
> 
> https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/86ffe422e668fc8fb88c4672d764a70cd8de8181/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L322
> 
> two thoughts.
> 1. Since this can accept one or more ITelemetryClient objects, can this be renamed to plural version WithTelemetryClient**s**.
> 2. Since most developers will use a single Telemetry client, an override can be added that accept single client. This will keep API simple for the developers. This will be a breaking change. But if minimal customers are using it, it may be worth. Thoughts?

The current API uses `params` so it can already accept a single telemetry client.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
neha-bhargava: Thanks for the feedback @jmprieur, this is consistent with how we log in SAL. As discussed with @bgavrilMS, we can probably resolve this by adding a delegate to perform this task. @TimHannMSFT, thoughts on adding a delegate to iterate over ITelemetryClient list?

TimHannMSFT: I don't understand the delegate comment. My read of Jean-Marc's comment was to not have the concept of iterating over clients here but refactored into a method that does this in a standard way for all times you're emitting an event. I agree. See https://identitydivision.visualstudio.com/DevEx/_git/Microsoft-IdentityModel-S2S?path=/src/Microsoft.IdentityModel.S2S.Tokens/Telemetry/TelemetryExtension.cs for what we do in SAL.

TimHannMSFT:
>             if (telemetryClients.Length > 0)

nit: this check isn't needed right? If it's length zero the foreach won't be entered.

---
Refers to: src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs:333 in 839cb78. [](commit_id = 839cb783e8dfcd7a45b788f72b5c062473762fca, deletion_comment = False)

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -343,9 +345,25 @@ public ConfidentialClientApplicationBuilder WithTelemetryClient(params ITelemetr
...
 
+            TelemetryClientLogMsalVersion();
+
             return this;
         }
 
+        private void TelemetryClientLogMsalVersion()
+        {
+            MsalTelemetryEventDetails telemetryEventDetails = new MsalTelemetryEventDetails(TelemetryConstants.ConfigurationUpdateEventName);
+            telemetryEventDetails.SetProperty(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion());
```
TimHannMSFT, line 356: The way SAL does this, which I prefer, is to not construct this until you know there's a client who would receive it. i.e. to put it inside the foreach/if blocks and only create the event details there.

Here it's pretty trivial extra work (but it is extra work) but it's a good pattern to set when emitting telemetry for the library.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
neha-bhargava: Thanks @TimHannMSFT, I have updated the code.

Issue #3743, opened 2022-10-12
[Bug] NullReferenceException using preview broker with MSA account and MSA-PT enabled for ATS


mjcheetham:
**Logs and network traces**

<details>
  <summary>MSAL Logs</summary>

```log
[Info] False MSAL 4.47.2.0 MSAL.Desktop 4.8 or later Windows 10 Enterprise [2022-10-11 21:36:04Z - 20e6084f-19cb-40c1-a2ef-3a30b2110dd2] MSAL MSAL.Desktop with assembly version '4.47.2.0'. CorrelationId(20e6084f-19cb-40c1-a2ef-3a30b2110dd2)
[Info] False MSAL 4.47.2.0 MSAL.Desktop 4.8 or later Windows 10 Enterprise [2022-10-11 21:36:04Z - 20e6084f-19cb-40c1-a2ef-3a30b2110dd2] === AcquireTokenSilent Parameters ===
[Info] False MSAL 4.47.2.0 MSAL.Desktop 4.8 or later Windows 10 Enterprise [2022-10-11 21:36:04Z - 20e6084f-19cb-40c1-a2ef-3a30b2110dd2] LoginHint provided: True
[Info] False MSAL 4.47.2.0 MSAL.Desktop 4.8 or later Windows 10 Enterprise [2022-10-11 21:36:04Z - 20e6084f-19cb-40c1-a2ef-3a30b2110dd2] Account provided: False
[Info] False MSAL 4.47.2.0 MSAL.Desktop 4.8 or later Windows 10 Enterprise [2022-10-11 21:36:04Z - 20e6084f-19cb-40c1-a2ef-3a30b2110dd2] ForceRefresh: False
[Info] False MSAL 4.47.2.0 MSAL.Desktop 4.8 or later Windows 10 Enterprise [2022-10-11 21:36:04Z - 20e6084f-19cb-40c1-a2ef-3a30b2110dd2] 
=== Request Data ===
Authority Provided? - True
Scopes - 499b84ac-1321-427f-aa17-267ca6975798/.default
Extra Query Params Keys (space separated) - 
ApiId - AcquireTokenSilent
IsConfidentialClient - False
SendX5C - False
LoginHint ? False
IsBrokerConfigured - True
HomeAccountId - False
CorrelationId - 20e6084f-19cb-40c1-a2ef-3a30b2110dd2
UserAssertion set: False
LongRunningOboCacheKey set: False
Region configured: 

[...]

PR #3763, opened 2022-10-27
Broker Silent NRE

gladjohn:
Fixes : #3743 

With the new broker, ATS with loginHint fails with a Null Reference Exception. When we do a silent call with the new broker we need to pass an account. This works when we call ATS with the IAccount overload. With login_hint the acquireTokenSilentParameters.Account comes back as null. Using authenticationRequestParameters.Account instead.

```cs
// Fails with NRE
var result2 = await app.AcquireTokenSilent(scopes, loginHint).ExecuteAsync();
Console.WriteLine(result2.AccessToken);
```

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -257,7 +257,7 @@ namespace Microsoft.Identity.Client.Broker
                 _logger))
             {
                 using (var readAccountResult = await s_lazyCore.Value.ReadAccountByIdAsync(
-                    acquireTokenSilentParameters.Account.HomeAccountId.ObjectId,
+                    authenticationRequestParameters.Account.HomeAccountId.ObjectId,
                     authenticationRequestParameters.CorrelationId.ToString("D"),
                     cancellationToken).ConfigureAwait(false))
                 {
--- a/tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
+++ b/tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
@@ -208,13 +208,6 @@ namespace NetDesktopWinForms
 
             if (!string.IsNullOrEmpty(loginHint))
             {
-                //if (IsMsaPassthroughConfigured())
-                //{
-                //    // TODO: bogavril - move this exception in WAM
-                //    throw new InvalidOperationException(
-                //        "[TEST APP FAILURE] Do not use login hint on AcquireTokenSilent for MSA-Passthrough. Use the IAccount overload.");
-                //}
-
                 Log($"ATS with login hint: " + loginHint);
                 var builder = pca.AcquireTokenSilent(GetScopes(), loginHint);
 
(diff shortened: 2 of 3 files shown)


src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -257,7 +257,7 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
                 _logger))
             {
                 using (var readAccountResult = await s_lazyCore.Value.ReadAccountByIdAsync(
-                    acquireTokenSilentParameters.Account.HomeAccountId.ObjectId,
+                    authenticationRequestParameters.Account.HomeAccountId.ObjectId,
```
bgavrilMS, line 260: Yes, that should work @gladjohn

pmaytak, line 260: So `acquireTokenSilentParameters.Account` ends up being the argument that was passed into ATS (null in case of login hint used); while `authenticationRequestParameters.Account` ends up being the account that MSAL finds in cache when login hint is used.

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
```diff
@@ -208,12 +208,12 @@ private async Task<AuthenticationResult> RunAtsAsync(IPublicClientApplication pc
 
             if (!string.IsNullOrEmpty(loginHint))
             {
-                if (IsMsaPassthroughConfigured())
-                {
-                    // TODO: bogavril - move this exception in WAM
-                    throw new InvalidOperationException(
-                        "[TEST APP FAILURE] Do not use login hint on AcquireTokenSilent for MSA-Passthrough. Use the IAccount overload.");
-                }
+                //if (IsMsaPassthroughConfigured())
```
pmaytak, line 211: This check/error can be removed then right?

pmaytak (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
trwalke: If what peter said is true, should we be checking local cache for an account instead of going directly to WAM? We have decided to avoid checking local cache for tokens so wondering if it is the same for accounts with brokers?

bgavrilMS: The logic of `AcquireTokenSilent(login_hint)` with WAM is: 

1. go to our cache and try to find `Account` object that matches `login_hint`. Fail with UI Required ex if not found or if 2+ are found.
2. call `AcquireTokenSilent(Account)`
3. look in our cache for AT 
4. look in our cache for RT (this is a bug, it is tracked). If found, refresh the RT.
5. go to WAM

bgavrilMS (approved):

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
gladjohn: removed

SameerK-MSFT (approved):

PR #3768, opened 2022-10-28
Refactor IsConfidentialClient and remove experimental flag from PCA POP

### LINKED ISSUE #3767: [Feature Request] Remove experimental API flag on WithProofOfPossession (PCA)
Visual Studio Installer teams is planning on using this feature and releasing with it. 

Remove experimental API flag on WithProofOfPossession (PCA)

bgavrilMS:
Fixes #3767

- We need a better way to figure out if an app config is for CCA or for PCA. Since app config is mutable object, best is to rely on ctor.

- 

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
@@ -192,7 +192,15 @@ namespace Microsoft.Identity.Client
 #endif
         public AcquireTokenSilentParameterBuilder WithProofOfPossession(string nonce, HttpMethod httpMethod, Uri requestUri)
         {
-            ValidateUseOfExperimentalFeature();
+            if (ServiceBundle.Config.IsConfidentialClient)
+                ValidateUseOfExperimentalFeature();
+
+            // On public client, we only support POP via the broker
+            if (!ServiceBundle.Config.IsConfidentialClient &&
+                !ServiceBundle.Config.IsBrokerEnabled)
+            {
+                throw new MsalClientException(MsalError.BrokerRequiredForPop, MsalErrorMessage.BrokerRequiredForPop);
+            }
 
             ClientApplicationBase.GuardMobileFrameworks();
             var broker = ServiceBundle.PlatformProxy.CreateBroker(ServiceBundle.Config, null);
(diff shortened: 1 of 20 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -192,7 +192,15 @@ public AcquireTokenSilentParameterBuilder WithProofOfPossession(PoPAuthenticatio
 #endif
         public AcquireTokenSilentParameterBuilder WithProofOfPossession(string nonce, HttpMethod httpMethod, Uri requestUri)
         {
-            ValidateUseOfExperimentalFeature();
+            if (ServiceBundle.Config.IsConfidentialClient)
+                ValidateUseOfExperimentalFeature();
```
bgavrilMS, line 196: Still experimental on CCA.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -25,6 +25,11 @@ namespace Microsoft.Identity.Client
 {
     internal sealed class ApplicationConfiguration : IAppConfig
     {
+        public ApplicationConfiguration(bool isConfidentialClient)
```
bgavrilMS, line 28: Could't find a better way to do this since ApplicationConfiguration is mutable. We can add logic to property setters but it gets ugly.

gladjohn (approved):

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
SameerK-MSFT: Alternative can be to have set as internal.

pmaytak (approved):

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
pmaytak: I think having this be set in constructor is good, since it forces you to set this property when config is created, so it's not missed by accident somewhere and set to false by default.

Issue #3718, opened 2022-09-29
[Feature Request] Add HTTP telemetry to understand POP usage


bgavrilMS:
I propose we log a single data point as a number with the following values: 

- 1 = bearer
- 2 = pop
- 3 = ssh-cert
- 4 = extensibilty api WithPOPKey is used (this covers MSAuth10AT_POP which SAL uses)

PR #3771, opened 2022-10-28
Log token type to indicate pop usage

neha-bhargava:
Fixes #
#3718 

Add token type to telemetry to indicate the token type

**Testing**
Updated unit and integration tests

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
@@ -10,6 +10,8 @@ namespace Microsoft.Identity.Client.AuthScheme.Bearer
     {
         internal const string BearerTokenType = "bearer";
 
+        public int Id => 1;
+
         public string AuthorizationHeaderPrefix => "Bearer";
 
         public string AccessTokenType => BearerTokenType;
--- a/src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Identity.Client.AuthScheme.Bearer
     {
         internal const string BearerTokenType = "bearer";
 
-        public int Id => 1;
+        public int TelemetryId => 1;
 
         public string AuthorizationHeaderPrefix => "Bearer";
 
--- a/src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Identity.Client.AuthScheme.Bearer
     {
         internal const string BearerTokenType = "bearer";
 
-        public int TelemetryId => 1;
+        public TokenType TelemetryTokenType => TokenType.Bearer;
 
         public string AuthorizationHeaderPrefix => "Bearer";
 
--- a/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Identity.Client.AuthScheme
         /// <summary>
         /// Value to log to telemetry to indicate pop usage.
         /// </summary>
-        int Id { get; }
+        int TelemetryId { get; }
 
         /// <summary>
         /// Prefix for the HTTP header that has the token. E.g. "Bearer" or "POP"
--- a/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs
@@ -12,6 +12,11 @@ namespace Microsoft.Identity.Client.AuthScheme
     internal interface IAuthenticationScheme
     {
         /// <summary>
+        /// Value to log to telemetry to indicate pop usage.
+        /// </summary>
+        int Id { get; }
+
+        /// <summary>
         /// Prefix for the HTTP header that has the token. E.g. "Bearer" or "POP"
         /// </summary>
         string AuthorizationHeaderPrefix { get; }
--- a/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Identity.Client.AuthScheme
         /// <summary>
         /// Value to log to telemetry to indicate pop usage.
         /// </summary>
-        int TelemetryId { get; }
+        TokenType TelemetryTokenType { get; }
 
         /// <summary>
         /// Prefix for the HTTP header that has the token. E.g. "Bearer" or "POP"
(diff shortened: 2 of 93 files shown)


src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs
```diff
@@ -11,6 +11,11 @@ namespace Microsoft.Identity.Client.AuthScheme
     /// </summary>
     internal interface IAuthenticationScheme
     {
+        /// <summary>
+        /// Value to log to telemetry to indicate pop usage.
+        /// </summary>
+        int Id { get; }
```
trwalke, line 17: TelemetryId?

src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
```diff
@@ -10,6 +10,8 @@ internal class BearerAuthenticationScheme : IAuthenticationScheme
     {
         internal const string BearerTokenType = "bearer";
 
+        public int Id => 1;
```
trwalke, line 13: TelemetryId

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -49,6 +49,8 @@ public PopAuthenticationScheme(PoPAuthenticationConfiguration popAuthenticationC
             KeyId = Base64UrlHelpers.Encode(keyThumbprint);
         }
 
+        public int Id => 2;
```
trwalke, line 52: Should this AuthScheme and the BrokerAuthScheme have the same value? This one is used for CCA and the other one is used for broker.

trwalke (approved):
Couple of clarifying questions but generally looks good

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
neha-bhargava: Since this is logging to telemetry to see which auth scheme is used I think this should be fine. @bgavrilMS should this be another value for broker?

src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
```diff
@@ -10,6 +10,8 @@ internal class BearerAuthenticationScheme : IAuthenticationScheme
     {
         internal const string BearerTokenType = "bearer";
 
+        public int TelemetryId => 1;
```
pmaytak, line 13: I would create an enum like we use [CacheRefreshReason](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs). I think this makes the code more readable.

Something like:
```csharp
internal enum TelemetryTokenType // or TokenType
{
Bearer = 1,
Pop = 2,
SshCert = 3,
PopExtensibilityApi = 4,
}
```

pmaytak (commented):
Suggestion to use an enum.
Also the telemetry API review doc needs to be updated, I think.

src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
neha-bhargava: I was thinking of using an enum earlier but here the scenario is different. We don't have to evaluate the value it simply depends on which type of auth scheme object is created and hence used this,

neha-bhargava:
> Suggestion to use an enum. Also the telemetry API review doc needs to be updated, I think.

Since we are adding to the platform field, we don't update the telemetry doc. Api review doc is updated when there is a change in the telemetry schema for all the sdk's.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/HttpTelemetryTests.cs
```diff
@@ -567,7 +567,9 @@ private async Task<(MockHttpMessageHandler MockHttpHandler, Guid Correlationid)>
 
             Assert.AreEqual(isCacheSerialized ? "1" : "0", telemetryCategories[2].Split(',')[0]); // is_cache_serialized
 
-            Assert.AreEqual(isLegacyCacheEnabled ? "1" : "0", telemetryCategories[2].Split(',')[1]); // is_legacy_cache_enabled
+            Assert.AreEqual(isLegacyCacheEnabled ? "1" : "0", telemetryCategories[2].Split(',')[1]); // is_legacy_cache_enabled
+
+            Assert.AreEqual("1", telemetryCategories[2].Split(',')[2]);
```
pmaytak, line 572: Should also add tests/asserts for other telemetry token types.

src/client/Microsoft.Identity.Client/AuthScheme/Bearer/BearerAuthenticationScheme.cs
pmaytak: imo it helps eliminate scenarios like this where we have to do extra explanation of what the numbers are - with enums, the code is self documenting.
```csharp
// Token type is to indicate 1 - bearer, 2 - pop, 3 - ssh-cert, 4 - external.
sb.Append(eventInProgress.TokenType);
```
Similar thing like in tests: 
instead of
```csharp
private readonly string XClientCurrentTelemetryROPCFailure = $"{TelemetryConstants.HttpTelemetrySchemaVersion}|1003,{CacheRefreshReason.NotApplicable:D},,,|0,1,1";
```
can be this
```csharp
private readonly string XClientCurrentTelemetryROPCFailure = $"{TelemetryConstants.HttpTelemetrySchemaVersion}|1003,{CacheRefreshReason.NotApplicable:D},,,|0,1,{TokenType.Bearer}";
```

neha-bhargava: Makes sense. I will update it.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/HttpTelemetryTests.cs
neha-bhargava: Updated a few tests to assert the token type set for auth scheme. This should cover the scenario that the correct value is set.

src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
neha-bhargava: Discussed offline with Bogdan. We can have the same value. If there a need to differentiate we can use the app id to know if it is public client or confidential client.

src/client/Microsoft.Identity.Client/AuthScheme/TokenType.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client.AuthScheme
+{
+    /// <summary>
+    /// Specifies the token type to log to telemetry.
+    /// </summary>
+    public enum TokenType
```
pmaytak, line 9: Needs to be internal.

neha-bhargava: public for testing.

pmaytak (approved):
👍 
Just make enum internal.

neha-bhargava:
> 👍 Just make enum internal.

It is left as public so that we can use it in tests to compare the results.

pmaytak:
> > 👍 Just make enum internal.
> 
> It is left as public so that we can use it in tests to compare the results.

Which tests? If it's for our unit/integration tests, then internal MSaL classes are still visible there. We only make things public for testing for other users (so it's visible in the public API); but in this case it's fully internal feature.

PR #3777, opened 2022-10-31
MSAL Runtime logging integration

SameerK-MSFT:
Fixes #
Adds unit tests

Added Unit tests

**Testing**
Unit Tests

**Performance impact**
None

**Documentation**
No doc needed

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
+++ b/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
@@ -4,6 +4,7 @@
     <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
     <TargetFrameworkNetDesktop Condition="$([MSBuild]::IsOsPlatform('Windows')) and '$(MSAL_DESKTOP_ONLY_DEV)' == ''">net461</TargetFrameworkNetDesktop>
 
+    <LangVersion>8.0</LangVersion>
     <PlatformTarget>AnyCPU</PlatformTarget>
     <TargetFramework Condition="'$(TargetFrameworkNetDesktop)' == ''">$(TargetFrameworkNetStandard)</TargetFramework>
     <TargetFrameworks Condition="'$(TargetFrameworkNetDesktop)' != ''">$(TargetFrameworkNetStandard);$(TargetFrameworkNetDesktop)</TargetFrameworks>
@@ -56,7 +57,7 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.13.0" IncludeAssets="all" />
+    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.13.2" IncludeAssets="all" />
     <ProjectReference Include="..\Microsoft.Identity.Client\Microsoft.Identity.Client.csproj" />
   </ItemGroup>
   <ItemGroup Label="Build Tools" Condition="$([MSBuild]::IsOsPlatform('Windows'))">
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -106,7 +106,6 @@ namespace Microsoft.Identity.Client.Broker
 
         private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
         {
-            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
             LogLevel msalLogLevel = LogLevelMap[args.LogLevel];
             if (_logger.IsLoggingEnabled(msalLogLevel))
             {
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -27,6 +27,16 @@ namespace Microsoft.Identity.Client.Broker
         private readonly WindowsBrokerOptions _wamOptions;
         private static Exception s_initException;
 
+        private readonly Dictionary<NativeInterop.LogLevel, LogLevel> _logLevelMap = new Dictionary<NativeInterop.LogLevel, LogLevel>()
+        {
+            { NativeInterop.LogLevel.Trace, LogLevel.Verbose },
+            { NativeInterop.LogLevel.Debug, LogLevel.Verbose },
+            { NativeInterop.LogLevel.Info, LogLevel.Info },
+            { NativeInterop.LogLevel.Warning, LogLevel.Warning },
+            { NativeInterop.LogLevel.Error, LogLevel.Error },
+            { NativeInterop.LogLevel.Fatal, LogLevel.Error },
+        };
+
         public bool IsPopSupported => true;
 
         /// <summary>
@@ -83,6 +93,11 @@ namespace Microsoft.Identity.Client.Broker
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
+            if (_logger.IsDefaultPlatformLoggingEnabled)
+            {
+                s_lazyCore.Value.LogEvent += LogEventRaised;
+            }
+
             _parentHandle = GetParentWindow(uiParent);
 
             _wamOptions = appConfig.WindowsBrokerOptions ??
@@ -91,6 +106,60 @@ namespace Microsoft.Identity.Client.Broker
             if (_wamOptions.ListWindowsWorkAndSchoolAccounts)
             {
                 throw new NotImplementedException("The new broker implementation does not yet support Windows account discovery (ListWindowsWorkAndSchoolAccounts option)");
+            }
+        }
+
+        private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
+        {
+            LogLevel msalLogLevel = _logLevelMap[args.LogLevel];
+            if (_logger.IsLoggingEnabled(msalLogLevel))
+            {
+                if (_logger.PiiLoggingEnabled)
+                {
+                    switch (msalLogLevel)
+                    {
+                        case LogLevel.Always:
+                            _logger.AlwaysPii(args.Message, string.Empty);
+                            break;
+                        case LogLevel.Error:
+                            _logger.ErrorPii(args.Message, string.Empty); // TODO confirm with Sam
+                            break;
+                        case LogLevel.Warning:
+                            _logger.WarningPii(args.Message, string.Empty);
+                            break;
+                        case LogLevel.Info:
+                            _logger.InfoPii(args.Message, string.Empty);
+                            break;
+                        case LogLevel.Verbose:
+                            _logger.VerbosePii(args.Message, string.Empty);
+                            break;
+                        default:
+                            break;
+                    }
+                }
+                else
+                {
+                    switch (msalLogLevel)
+                    {
+                        case LogLevel.Always:
+                            _logger.Always(args.Message);
+                            break;
+                        case LogLevel.Error:
+                            _logger.Error(args.Message);
+                            break;
+                        case LogLevel.Warning:
+                            _logger.Warning(args.Message);
+                            break;
+                        case LogLevel.Info:
+                            _logger.Info(args.Message);
+                            break;
+                        case LogLevel.Verbose:
+                            _logger.Verbose(args.Message);
+                            break;
+                        default:
+                            break;
+                    }
+                }
             }
         }
 
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -91,7 +91,7 @@ namespace Microsoft.Identity.Client.Broker
             ApplicationConfiguration appConfig,
             ILoggerAdapter logger)
         {
-            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _logger = logger;
 
             _parentHandle = GetParentWindow(uiParent);
 
@@ -106,55 +106,25 @@ namespace Microsoft.Identity.Client.Broker
 
         private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
         {
-            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
-            LogLevel msalLogLevel = LogLevelMap[args.LogLevel];
-            if (_logger.IsLoggingEnabled(msalLogLevel))
+            if (_logger != null)
             {
-                if (_logger.PiiLoggingEnabled)
+                using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
+                LogLevel msalLogLevel = LogLevelMap[args.LogLevel];
+                if (_logger.IsLoggingEnabled(msalLogLevel))
                 {
-                    switch (msalLogLevel)
+                    string msgWithPii = string.Empty;
+                    string msgNoPii = string.Empty;
+
+                    if (_logger.PiiLoggingEnabled)
                     {
-                        case LogLevel.Always:
-                            _logger.AlwaysPii(args.Message, string.Empty);
-                            break;
-                        case LogLevel.Error:
-                            _logger.ErrorPii(args.Message, string.Empty);
-                            break;
-                        case LogLevel.Warning:
-                            _logger.WarningPii(args.Message, string.Empty);
-                            break;
-                        case LogLevel.Info:
-                            _logger.InfoPii(args.Message, string.Empty);
-                            break;
-                        case LogLevel.Verbose:
-                            _logger.VerbosePii(args.Message, string.Empty);
-                            break;
-                        default:
-                            break;
+                        msgWithPii = args.Message;
                     }
-                }
-                else
-                {
-                    switch (msalLogLevel)
+                    else
                     {
-                        case LogLevel.Always:
-                            _logger.Always(args.Message);
-                            break;
-                        case LogLevel.Error:
-                            _logger.Error(args.Message);
-                            break;
-                        case LogLevel.Warning:
-                            _logger.Warning(args.Message);
-                            break;
-                        case LogLevel.Info:
-                            _logger.Info(args.Message);
-                            break;
-                        case LogLevel.Verbose:
-                            _logger.Verbose(args.Message);
-                            break;
-                        default:
-                            break;
+                        msgNoPii = args.Message;
                     }
+
+                    _logger.Log(msalLogLevel, msgWithPii, msgNoPii);
                 }
             }
         }
@@ -183,7 +153,7 @@ namespace Microsoft.Identity.Client.Broker
 
             var cancellationToken = authenticationRequestParameters.RequestContext.UserCancellationToken;
 
-            _logger.Verbose("[WamBroker] Using Windows account picker.");
+            _logger?.Verbose("[WamBroker] Using Windows account picker.");
 
             if (authenticationRequestParameters?.Account?.HomeAccountId?.ObjectId != null)
             {
@@ -212,9 +182,9 @@ namespace Microsoft.Identity.Client.Broker
                         }
                         else
                         {
-                            _logger.Warning(
+                            _logger?.Warning(
                                 $"[WamBroker] Could not find a WAM account for the selected user {authenticationRequestParameters.Account.Username}");
-                            _logger.Info(
+                            _logger?.Info(
                                 $"[WamBroker] Calling SignInInteractivelyAsync this will show the account picker.");
 
                             msalTokenResponse = await SignInInteractivelyAsync(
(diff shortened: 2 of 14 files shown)


src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -83,6 +93,11 @@ private static void OnProcessExit(object sender, EventArgs e)
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
+            if (_logger.IsDefaultPlatformLoggingEnabled)
```
bgavrilMS, line 96: Don't use `IsDefaultPlatformLoggingEnabled`. We should probably get rid of it, it's poorly documented. 

Use `_logger.IsLoggingEnabled()`

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -27,6 +27,16 @@ internal class RuntimeBroker : IBroker
         private readonly WindowsBrokerOptions _wamOptions;
         private static Exception s_initException;
 
+        private readonly Dictionary<NativeInterop.LogLevel, LogLevel> _logLevelMap = new Dictionary<NativeInterop.LogLevel, LogLevel>()
```
bgavrilMS, line 30: make it static

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
bgavrilMS: CC @trwalke for help here.

bgavrilMS: Quite possibly we should always assume a logger exists...

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -94,6 +109,60 @@ private static void OnProcessExit(object sender, EventArgs e)
...
+        private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
+        {
+            LogLevel msalLogLevel = _logLevelMap[args.LogLevel];
+            if (_logger.IsLoggingEnabled(msalLogLevel))
+            {
+                if (_logger.PiiLoggingEnabled)
+                {
+                    switch (msalLogLevel)
+                    {
+                        case LogLevel.Always:
```
bgavrilMS, line 121: nit: you can get rid of the switches and just use 

`logger.Log(LogLevel, messageWithPii, messageWithoutPii)`

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -153,7 +188,115 @@ public void HandleInstallUrl_Throws()
...
+        // Done - Logs in one
+        // NA - logs in multiple
+        // Done - maps correctly
+        // Done - Is Pii enabled
+        // Done - does not emit when loglevel is not appropriate
+        // Is API Platform logging enabled?
+        // no mem leaks
+
+        [TestMethod]
+        public async Task ATS_CallsLog_When_CalledAsync()
```
bgavrilMS, line 201: We have ROPC integration tests. Can you add real logging there and maybe assert that something comes from MSAL C++?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -83,6 +93,11 @@ private static void OnProcessExit(object sender, EventArgs e)
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
+            if (_logger.IsDefaultPlatformLoggingEnabled)
+            {
+                s_lazyCore.Value.LogEvent += LogEventRaised;
```
bgavrilMS, line 98: This is an event handle leak. You keep adding handles, but you never remove them. 

I think you need to register the callback on every IBroker method and remove it in a "finally" block.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Done.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Done. In a different way.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -94,10 +104,66 @@ private static void OnProcessExit(object sender, EventArgs e)
             }
         }
 
+        private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
+        {
+            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
+            LogLevel msalLogLevel = LogLevelMap[args.LogLevel];
+            if (_logger.IsLoggingEnabled(msalLogLevel))
```
trwalke, line 111: In regards to [this comment](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3777#discussion_r1010386713)

This is the best way to do it. If the logger is not enabled, there should be a null logger which will always return null.

trwalke: I meant to say false above. made a correction

SameerK-MSFT: Done.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -456,5 +528,38 @@ public bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
             _logger.Verbose($"[WAM Broker] MsalRuntime initialization successful.");
             return true;
         }
+
+        internal class LogEventWrapper : IDisposable
```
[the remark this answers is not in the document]
SameerK-MSFT: This is used only by this class. So better to have it as part of  RuntimeBroker

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
SameerK-MSFT: Added

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Spoke with @trwalke .Please check the changes made. We can discuss if you disagree.

src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
     <TargetFrameworkNetDesktop Condition="$([MSBuild]::IsOsPlatform('Windows')) and '$(MSAL_DESKTOP_ONLY_DEV)' == ''">net461</TargetFrameworkNetDesktop>
 
+    <LangVersion>8.0</LangVersion>
```
bgavrilMS, line 7: Inconsistent with MSAL? Why is this here?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -94,10 +104,36 @@ private static void OnProcessExit(object sender, EventArgs e)
...
+            if (_logger != null)
+            {
+                using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
+                LogLevel msalLogLevel = LogLevelMap[args.LogLevel];
+                if (_logger.IsLoggingEnabled(msalLogLevel))
+                {
+                    string msgWithPii = string.Empty;
+                    string msgNoPii = string.Empty;
+
+                    if (_logger.PiiLoggingEnabled)
```
bgavrilMS, line 118: This is not how msal_runtime handles PII. They handle it by passing flag (in auth params?), in which case both exception emssages and log messages have PII.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -435,26 +477,59 @@ public bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
...
+                _logger?.InfoPii(s_initException);
                 return false;
             }
 
-            _logger.Verbose($"[WAM Broker] MsalRuntime initialization successful.");
+            _logger?.Verbose($"[WAM Broker] MsalRuntime initialization successful.");
             return true;
         }
+
+        internal class LogEventWrapper : IDisposable
```
bgavrilMS, line 502: Please move this to its own class.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -94,10 +104,36 @@ private static void OnProcessExit(object sender, EventArgs e)
...
+                    string msgNoPii = string.Empty;
+
+                    if (_logger.PiiLoggingEnabled)
+                    {
+                        msgWithPii = args.Message;
+                    }
+                    else
+                    {
+                        msgNoPii = args.Message;
+                    }
```
bgavrilMS, line 125: args.Message is the same thing. See my comment about PII.

bgavrilMS (changes requested):
PII is not handled correctly.

src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
SameerK-MSFT: 8.0 is required to declare using of a local variable. e.g.
``` CSharp
using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
```

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/MsalLoggerValidator.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using Microsoft.IdentityModel.Abstractions;
+
+/// <summary>
+/// Performs validation on logs
+/// </summary>
+public class MsalLoggerValidator : IIdentityLogger
+{
+    // string that i tpart of MSAL logs
```
trwalke, line 12:
```suggestion
    // string that is part of MSAL logs
```

trwalke (approved):
Besides PII, LGTM

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -153,7 +187,116 @@ public void HandleInstallUrl_Throws()
...
+            }
+            catch (MsalUiRequiredException)
+            {
+                // Assert
+                _logger.Received().Log(Arg.Any<Client.LogLevel>(), Arg.Any<string>(), Arg.Any<string>());
+            }
+            catch (Exception ex)
+            {
+                Assert.IsTrue(false, ex.Message);
+            }
```
pmaytak, line 236: I think it's ok to remove this and for test to just fail by throwing.

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -153,7 +187,116 @@ public void HandleInstallUrl_Throws()
...
+        [DataRow(Client.NativeInterop.LogLevel.Error, Client.LogLevel.Error)]
+        [DataRow(Client.NativeInterop.LogLevel.Fatal, Client.LogLevel.Error)]
+        public void LogEventRaised_MapsEvents_Correctly_When_No_Pii(Client.NativeInterop.LogLevel nativeLogLevel, Client.LogLevel msalLogLevel)
+        {
+            const string logMessage = "This is test";
+            _logger.IsLoggingEnabled(msalLogLevel).Returns(true);
+
+            Type wamBrokerType = _wamBroker.GetType();
+            MethodInfo fireLogMethod = wamBrokerType.GetMethod("LogEventRaised", BindingFlags.NonPublic | BindingFlags.Instance);
+            fireLogMethod.Invoke(_wamBroker, new object[] { null, new LogEventArgs(nativeLogLevel, logMessage) });
```
pmaytak, line 253: Any way to test this without reflection? Make LogEventRaised internal?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -435,26 +477,59 @@ public bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
...
+
+        internal class LogEventWrapper : IDisposable
+        {
+            private bool _disposedValue;
+            RuntimeBroker _broker;
+
+            public LogEventWrapper(RuntimeBroker broker)
+            {
+                _broker = broker;
+                s_lazyCore.Value.LogEvent += _broker.LogEventRaised;
```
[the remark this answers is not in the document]
pmaytak: I think it will be difficult to repro this manually. Maybe using blocking tasks, like when we test WithCacheSynchronization.

But it could happen if request 1 is disposing and unsubscribed the event handler, while request 2 is in the middle of doing work.

I forget, is he event collection a set or a list? Would calling `+= _broker.LogEventRaised;` multiple times, add multiple delegates or just 1?

pmaytak (changes requested):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -81,7 +91,7 @@ private static void OnProcessExit(object sender, EventArgs e)
             ApplicationConfiguration appConfig,
             ILoggerAdapter logger)
         {
-            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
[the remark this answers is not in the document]
SameerK-MSFT: @trwalke we discussed this please clarify.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Pii will not be supported in this release. Created a feature request. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3822

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Updated as there will be no Pii in the first release

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/MsalLoggerValidator.cs
```diff
@@ -0,0 +1,36 @@
...
+    private const string MSALCPPIdentifier = "[MSAL:000";
+
+    // count of MsalCPP logs
+    private int MsalCPPLogCount { get; set; }
```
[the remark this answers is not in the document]
SameerK-MSFT: Good catch. I was using it earlier and was a leftover.

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -1,40 +1,71 @@
...
-using Microsoft.Identity.Test.Common;
-using Microsoft.Identity.Client.Internal.Logger;
-using Microsoft.Identity.Client.Platforms.Features.DesktopOs.Kerberos;
-using System.Runtime.InteropServices;
-using System.Linq;
+using Account = Microsoft.Identity.Client.Account;
 
 namespace Microsoft.Identity.Test.Unit.BrokerTests
 {
+    internal class TestLogger : ILoggerAdapter
```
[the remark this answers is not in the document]
SameerK-MSFT: Removed.

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -153,7 +187,116 @@ public void HandleInstallUrl_Throws()
...
+        {
+            // Arrange
+            var appTokenCache = new TokenCache(_serviceBundle, isApplicationTokenCache: true);
+            var requestContext = new RequestContext(_serviceBundle, Guid.NewGuid());
+            var tenantAuthority = AuthorityInfo.FromAadAuthority(AzureCloudInstance.AzurePublic, tenant: TestConstants.AadTenantId, validateAuthority: false);
+            var acquireTokenCommonParameters = new AcquireTokenCommonParameters
+            {
+                ApiId = ApiEvent.ApiIds.AcquireTokenForClient,
+                AuthorityOverride = tenantAuthority
+            };
```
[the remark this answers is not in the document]
SameerK-MSFT: This is done with OS current user. I am missing something. Let's discuss when you come back.

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
SameerK-MSFT: I do not want to change the access modifier just of testing. If we can do testing with reflection, I avoid it.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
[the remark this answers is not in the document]
SameerK-MSFT: > forget, is he event collection a set or a list? W
This is a collection and ad adn remove operations are thread safe Ref: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes

src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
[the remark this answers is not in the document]
SameerK-MSFT: I would like the latest. But we may face issues later as some of our users may not be able to use it. Still thinking 🤔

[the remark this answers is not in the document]
SameerK-MSFT: Let's play safe and not use the latest to avoid any backward compatibility issues down the line.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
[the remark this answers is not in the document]
SameerK-MSFT: Added parallel ATS for 25 ATS calls in the sample App (tests/devapps/WAM/NetFrameworkWam/Form1.cs). Still works.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -94,10 +104,26 @@ private static void OnProcessExit(object sender, EventArgs e)
...
+        private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
+        {
+            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
+            LogLevel msalLogLevel = LogLevelMap[args.LogLevel];
+            if (_logger.IsLoggingEnabled(msalLogLevel))
+            {
+                string msgWithPii = string.Empty;
+                string msgNoPii = string.Empty;
+
+                // as of now, there will be no Pii.
```
bgavrilMS, line 116: Please add a link to the GH issue

bgavrilMS (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Done.

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
```diff
@@ -11,6 +11,8 @@
     <LangVersion>latest</LangVersion>
     
     <Configurations>Debug;Release;Debug + MobileApps</Configurations>
+
+    <Platforms>AnyCPU;x64</Platforms>
```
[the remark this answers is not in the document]
SameerK-MSFT: it was required for different platforms

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -94,10 +104,27 @@ private static void OnProcessExit(object sender, EventArgs e)
             }
         }
 
+        private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
+        {
+            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
```
pmaytak, line 109: This should be removed since LogEventWrapper constructor adds LogEventRaised to the static event . So each time LogEventRaised is raised, we're adding one more handler.

pmaytak (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Good catch. That was copy/paste for each method added here too.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
pmaytak: @SameerK-MSFT Was just thinking about this again. I may have misheard today in standup. So is this correct - errors from runtime time **do not** have PII; they are scrubbed? (I think I just heard you say the opposite).

Issue #3775, opened 2022-10-31
[Feature Request] Authority Validation enhacements


bgavrilMS:
This is a consistency item. 

Spec is here: https://microsoft.sharepoint.com/:w:/t/ADAL/ESzbQvzlMuBDjAsqfaHdjNUBsTmwIUR5g17IAY4rvOY_Hw?e=7dv5hZ 

Tasks 

- [x] Implement phase 2 `WithInstanceDiscoevryEndpoint("DISABLED")` which turns off all instance discovery 
- [ ] Ensure ESTS-R scenarios work in conjunction with this setting
- [ ] Testing

@rayluo to confirm that only phase 2 is required and if phase1 ("WithKnownAuthorityHosts") is optional / recommended etc.

rayluo:
> @rayluo to confirm that only phase 2 is required and if phase1 ("WithKnownAuthorityHosts") is optional / recommended etc.

Yes. Also, [that internal specs](https://microsoft.sharepoint.com/:w:/t/ADAL/ESzbQvzlMuBDjAsqfaHdjNUBsTmwIUR5g17IAY4rvOY_Hw?e=7dv5hZ) grew along with the lengthy conversation. If you just want to focus on what ended up happening, [this implementation PR in MSAL Python](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/496) shipped only a boolean instance_discovery=True or False. You can for example take a look into its unit test cases.

PR #3818, opened 2022-11-22
Implementation of WithInstanceDiscoevryEndpoint api to control Author…

trwalke:
…ity validation/instance discovery

Fixes #
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3775

Adding api to disable all instance discovery functinality according to Ray's authority validation [spec](https://microsoft.sharepoint.com/:w:/t/ADAL/ESzbQvzlMuBDjAsqfaHdjNUBsTmwIUR5g17IAY4rvOY_Hw?e=7dv5hZ). 
User provided metadata is still allowed when instance discovery endpoint is disabled.
Regional behavior is unchanged

**Testing**
Additional unit tests

**Performance impact**
N/A

**Documentation**


src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -109,14 +108,26 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
...
 
             if (authorityInfo.IsInstanceDiscoverySupported)
             {
                 var entry = _userMetadataProvider?.GetMetadataOrThrow(environment, requestContext.Logger);
 
+                //Check if instance discovery endpoint is disabled
+                if (entry == null && requestContext.ServiceBundle.Config.InstanceDiscoveryEndpoint == Constants.InstanceDiscoveryEndpointDisabled)
+                {
+                    //Check if regional discovery provider returns an entry. Regional should not be affected by the disabling of instance discovery endpoint.
+                    entry = await _regionDiscoveryProvider.GetMetadataAsync(authorityUri, requestContext).ConfigureAwait(false);
```
trwalke, line 122: Not sure if regional metadata acquisition should be disabled as well.
@bgavrilMS

trwalke: Discussed offline, will keep regional metadata functional.

cc @rayluo

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -528,6 +529,20 @@ public T WithClientCapabilities(IEnumerable<string> clientCapabilities)
...
         }
 
+        /// <summary>
+        /// Determines whether or not instance discovery is performed when attempting to authenticate. Setting this to false will completely disable
+        /// instance discovery and authority validation. This will not affect the behavior of application configured with regional endpoints however.
+        /// Also, you can still provide your own instance metadata.
+        /// </summary>
+        /// <param name="enableInstanceDiscovery">Determines if instance discovery/Authority validation is performed</param>
+        /// <returns></returns>
+        public T WithInstanceDiscoevryEndpoint(bool enableInstanceDiscovery)
```
gladjohn, line 539:
```suggestion
        public T WithInstanceDiscoveryEndpoint(bool enableInstanceDiscovery)
```

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -109,14 +108,26 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
...
+
+                    if (entry == null)
+                    {
+                        requestContext.Logger.Info("[Instance Discovery] Skipping Instance discovery because it is disabled. ");
```
gladjohn, line 126: can we also include logging for "Instance discovery is enabled. Instance discovery will be performed"?

gladjohn (approved):

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -109,14 +108,26 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
...
             {
                 var entry = _userMetadataProvider?.GetMetadataOrThrow(environment, requestContext.Logger);
 
+                //Check if instance discovery endpoint is disabled
+                if (entry == null && requestContext.ServiceBundle.Config.InstanceDiscoveryEndpoint == Constants.InstanceDiscoveryEndpointDisabled)
+                {
+                    //Check if regional discovery provider returns an entry. Regional should not be affected by the disabling of instance discovery endpoint.
+                    entry = await _regionDiscoveryProvider.GetMetadataAsync(authorityUri, requestContext).ConfigureAwait(false);
+
+                    if (entry == null)
```
neha-bhargava, line 124: This check is not needed.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
neha-bhargava, line 539: Can this be renamed to `WithInstanceDiscoverDisabled(bool disableInstanceDiscovery = true)` or `WithDisabledInstanceDiscovery(bool disableInstanceDiscovery = true)`

bgavrilMS: I think most of our APIs that take in a "bool" have it as `WithXYZ(enabled)`. 

But I believe that @rayluo 's design mentions `WithInstanceDiscovery` and not `WithInstanceDiscoveryEndpoint`. Naming in hard.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -528,6 +529,20 @@ public T WithClientCapabilities(IEnumerable<string> clientCapabilities)
...
+        {
+            Config.InstanceDiscoveryEndpoint = Constants.InstanceDiscoveryEndpointDisabled;
```
bgavrilMS, line 541: Why is `InstanceDiscoveryEndpoint` a string and not a bool? It should be called `IsInstanceDiscoveryEnabled` as per suggestion of naming bool fields by @SameerK-MSFT

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -109,14 +108,26 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
...
+                    {
+                        requestContext.Logger.Info("[Instance Discovery] Skipping Instance discovery because it is disabled. ");
+                        return CreateEntryForSingleAuthority(authorityUri);
+                    }
```
bgavrilMS, line 128: I'd return the non-null entry here to be safe. The logic is hard to follow otherwise.

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -109,14 +108,26 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
...
-
             Uri authorityUri = authorityInfo.CanonicalAuthority;
             string environment = authorityInfo.Host;
 
             if (authorityInfo.IsInstanceDiscoverySupported)
             {
                 var entry = _userMetadataProvider?.GetMetadataOrThrow(environment, requestContext.Logger);
 
+                //Check if instance discovery endpoint is disabled
+                if (entry == null && requestContext.ServiceBundle.Config.InstanceDiscoveryEndpoint == Constants.InstanceDiscoveryEndpointDisabled)
```
bgavrilMS, line 119: You also need to update `GetMetadataEntryTryAvoidNetworkAsync` and have a test for this.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/AuthorityTests.cs
```diff
@@ -113,6 +115,61 @@ public async Task AuthorityValidationTestWithFalseValidateAuthorityAsync()
...
+                .ConfigureAwait(false);
+
+            Assert.AreEqual("No such host is known.", exception.Message);
+            Assert.AreEqual(11001,((SocketException)exception.InnerException).ErrorCode);
+        }
+
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public async Task CommonAuthorityValidationTestAsync(bool validateAuthority)
```
bgavrilMS, line 150: I don't see the point in making actual AAD tests here. This should be a unit test

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/AuthorityTests.cs
```diff
@@ -113,6 +115,61 @@ public async Task AuthorityValidationTestWithFalseValidateAuthorityAsync()
...
+            HttpRequestException exception = await AssertException.TaskThrowsAsync<HttpRequestException>(() =>
+                 pca.AcquireTokenByUsernamePassword(
+                    s_scopes,
+                     user.Upn,
+                     user.GetOrFetchPassword())
+                     .ExecuteAsync())
+                .ConfigureAwait(false);
+
+            Assert.AreEqual("No such host is known.", exception.Message);
+            Assert.AreEqual(11001,((SocketException)exception.InnerException).ErrorCode);
```
bgavrilMS, line 144: I think what this test is trying to say is that authority validation is not performed and MSAL will fail when reaching out to the /token endpoint. This is fine, but I don't see why we'd have an expensive integration test for this, when a unit test can work just as well. We are not interested in testing the integration point between MSAL and an unknwon authority, we're just interested in making sure instance discovery is not done (which by the way you don't assert here).

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -109,14 +108,26 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
             RequestContext requestContext, 
             bool forceValidation = false)
         {
-
             Uri authorityUri = authorityInfo.CanonicalAuthority;
             string environment = authorityInfo.Host;
 
             if (authorityInfo.IsInstanceDiscoverySupported)
             {
                 var entry = _userMetadataProvider?.GetMetadataOrThrow(environment, requestContext.Logger);
```
bgavrilMS, line 116: I would throw if app dev sets their own metadata and this flag at the same time.

bgavrilMS (changes requested):
1. GetMetadataEntryTryAvoidNetworkAsync needs to have the same logic 
2. unit tests should cover this, not integration tests

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
pmaytak: Why? `GetMetadataEntryTryAvoidNetworkAsync` doesn't go over the network, it ends up calling this method.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/AuthorityTests.cs
```diff
@@ -113,6 +115,61 @@ public async Task AuthorityValidationTestWithFalseValidateAuthorityAsync()
                 .ConfigureAwait(false);
         }
 
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public async Task UncommonAuthorityValidationTestAsync(bool validateAuthority)
```
pmaytak, line 121: - Are there existing authority validation and/or instance discovery unit tests that we can just add `WithInstanceDiscovery(true/false)` to, plus extra asserts.
- I think we also need to update tests with disabled instance discovery that cover other scenarios, like user provided metadata, or known-metadata used; to protect from regressions.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
pmaytak, line 539: The parameter is not used anywhere. Should be used to set the config value.

pmaytak (changes requested):

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
trwalke: I think we want to keep it inside the block. we wont know if discovery failed or not until we check to see if entry is null

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
trwalke: The design mentions `InstanceDiscoveryEndpoint` but I think `WithInstanceDiscovery` is ok

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
trwalke: not sure what you mean. can you clarify? Null is not being returned here.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
rayluo: > The design mentions `InstanceDiscoveryEndpoint` but I think `WithInstanceDiscovery` is ok

I can explain. The design doc written as a WORD doc with table, was the first several iterations. At that time, the `InstanceDiscoveryEndpoint` was indeed proposed, because we wanted the polymorphism to accept not only a boolean but also a new string endpoint, and the latter was considered a major reason to "justify" this new parameter. However, that "customizing a new endpoint" idea was later dropped due to lack of real-world requirements. And we eventually ended up with a [simplified design](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/AuthorityValidation/authority_validation_concepts_and_api.md&_a=preview&anchor=skip-authority-validation-and-alias-metadata-behaviors-without-a-set-of-predefined-authorities), which chose the name `WithInstanceDiscovery`.

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
bgavrilMS: I think it does https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs

Maybe the name or the design is wrong. i.e. "TryAvoid" is not the same as "Avoid". It avoids going to the network if all the hostnames in the cache == authority.host

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
trwalke: I think I was incorrect in assuming that user metadata was a separate concept from network metadata. So, in actuality, usermatada should also be ignored when IsInstanceDiscoverySupported is set to false. Which is why @bgavrilMS is suggesting to throw if both are set.

pmaytak: Are we talking about IsInstanceDiscoverySupported  or InstanceDiscoveryEndpoint?
I think it's valid if user provides metadata and disables network discovery. In this case we just check user + regional + cached + known metadta.

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
pmaytak: Yea, but `GetMetadataEntryTryAvoidNetworkAsync ` calls `GetMetadataEntryAsync` todo the network call, which already has the check.

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
trwalke: Im talking about IsInstanceDiscoverySupported. Also, not sure about "if user provides metadata and disables network discovery. In this case we just check user + regional + cached + known metadta."

bgavrilMS: Yeah. I think the current implemenation around user metadata is also fine taking precedence over this flag is also fine.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ClientCredentialsTests.WithRegion.cs
```diff
@@ -137,7 +139,8 @@ private void AssertTelemetry(HttpSnifferClientFactory factory, string currentTel
                 builder.WithCertificate(settings.GetCertificate());
             }
 
-            builder.WithAuthority($@"https://{settings.Environment}/{settings.TenantId}")
+            builder.WithAuthority($@"https://{settings.Environment}/{settings.TenantId}")
+                .WithInstanceDiscovery(settings.InstanceDiscoveryEndpoint)
```
bgavrilMS, line 143: I would undo this test. It doesn't assert anything.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -42,14 +42,19 @@ public override void TestInitialize()
             _networkCacheMetadataProvider = Substitute.For<INetworkCacheMetadataProvider>();
             _userMetadataProvider = Substitute.For<IUserMetadataProvider>();
 
+            InitializeTestObjects();
+        }
+
+        private void InitializeTestObjects(bool isInstanceDiscoveryEnabled = true)
```
bgavrilMS, line 48: Can we have a simple unit test around the end to end, similar to `ConfidentialClientUsingSecretNoCacheProvidedTestAsync`: 

- set this flag to true
- remove the mock `httpManager.AddInstanceDiscoveryMockHandler();`

bgavrilMS (approved):

pmaytak (approved):

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -96,16 +95,28 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
...
+                //Check if instance discovery endpoint is disabled
+                if (entry == null && requestContext.ServiceBundle.Config.IsInstanceDiscoveryEnabled)
+                {
+                    entry = _networkCacheMetadataProvider.GetMetadata(environment, requestContext.Logger) ??
                     _knownMetadataProvider.GetMetadata(environment, existingEnvironmentsInCache, requestContext.Logger) ??
                     await GetMetadataEntryAsync(authorityInfo, requestContext).ConfigureAwait(false);
+                }
+
+                if (entry == null)
+                {
```
bgavrilMS, line 111: nit: Pls add a log statement here.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/InstanceDiscoveryManagerTests.cs
```diff
@@ -122,6 +127,59 @@ public async Task NetworkCacheProvider_IsUsedFirst_Async()
...
+        [TestMethod]
+        public async Task InstanceDiscoveryDisabled_Async()
+        {
+            // Arrange
+            InitializeTestObjects(false);
+            INetworkMetadataProvider networkMetadataProvider = new NetworkMetadataProvider(
+                Substitute.For<IHttpManager>(),
+                _networkCacheMetadataProvider);
+
+            _networkCacheMetadataProvider.GetMetadata("some_env.com", Arg.Any<ILoggerAdapter>()).Returns(_expectedResult);
```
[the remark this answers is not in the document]
trwalke: Yes, it should not be called, but I put this here because I want to be sure that the cached entry is not returned when the flag is set to false. There is an assert that validates that this is true.

PR #3834, opened 2022-11-30
Code to enable Pii

### LINKED ISSUE #3822: [Feature Request] Support Pii logging in MsalRuntime Interop layer
Currently msalruntime interop layer does not support Pii

Provide API to enable/disable Pii

**Describe alternatives you've considered**
No alternative

**Additional context**
None

SameerK-MSFT:
Fixes #3822

Added API to dis/enable Pii

**Testing**
Intetgration tests to verify
Unit test
Updated app

**Performance impact**
None

**Documentation**


bgavrilMS (approved):

trwalke (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -95,6 +95,11 @@ private static void OnProcessExit(object sender, EventArgs e)
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
 
+            if(_logger.PiiLoggingEnabled)
```
pmaytak, line 98: I think `if` needs to be removed and we need to call `EnablePii` every time RuntimeBroker is created.

The scenario I'm thinking:
- PCA instance is created with PII logging.
- ATI call is made
- Interop instance is created statically.
- RuntimeBroker instance is created, PII enabled in Interop static instance.
- For some other ATI call, a new PCA instance is created. PII logging is disabled (dynamically somehow, through config or whatever).
- Interop instance is already created.
- RuntimeBroker instance is created, **PII needs to be disabled here** for the static interop instance.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -119,6 +119,63 @@ public async Task WamUsernamePasswordRequestAsync()
...
+
+            Func<IntPtr> windowHandleProvider = () => intPtr;
+
+            WamLoggerValidator testLogger = new WamLoggerValidator();
+
+            IPublicClientApplication pca = PublicClientApplicationBuilder
+               .Create(labResponse.App.AppId)
+               .WithParentActivityOrWindow(windowHandleProvider)
+               .WithAuthority(labResponse.Lab.Authority, "organizations")
+               .WithLogging(testLogger, enablePiiLogging: true)
```
pmaytak, line 162: Related to my previous comment - we should also test that if new PCA instance disables PII logging, the runtime broker also does.

pmaytak (changes requested):
Does `LogEventRaised` in `RuntimeBroker` needs to be updated also to log both messages?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Good point. The irony here is that there is only one static instance of the interop and there can be multiple instances of runtime broker. Let's take the above condition a step further. 
User starts utilizing the first instance again. In that case, Pii will not be enabled as the constructor will not be called.

In most cases Pii is enabled/disabled for the lifespan of the application. So, it is unlikely that the case you mentioned will ever happen.
The reason if was put was there to avoid unnecessary interop call that may impact perf.

pmaytak: > User starts utilizing the first instance again. In that case, Pii will not be enabled as the constructor will not be called.

Yes, that's right. I think that scenario of having multiple PCA instances that are reused is uncommon. The more common scenarios are either using a static PCA instance or recreate a new instance for each request.

> In most cases Pii is enabled/disabled for the lifespan of the application. So, it is unlikely that the case you mentioned will ever happen.

Yea, I think for PCA that's common, since PCA apps are closed/opened all the time, as opposed to CCA running on a server. The scenario I was thinking of is if I'm using an app and something breaks, so I enable logs in the settings, then try to repro again - this would mean I'd have to restart the app (so that interop instance gets recreated with new settings).

Probably unlikely I guess, but wanted to bring it up so it doesn't create weird issues down the road.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -50,13 +50,19 @@ public async Task WamSilentAuthUserInteractionRequiredAsync()
             try
             {
                 var result = await pca.AcquireTokenSilent(scopes, PublicClientApplication.OperatingSystemAccount).ExecuteAsync().ConfigureAwait(false);
-
+                Assert.IsTrue(false, "AcquireTokenSilent successfull!");
```
gladjohn, line 53:
```suggestion
                Assert.IsTrue(false, "AcquireTokenSilent successful!");
```

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -50,13 +50,16 @@ public async Task WamSilentAuthUserInteractionRequiredAsync()
             try
             {
                 var result = await pca.AcquireTokenSilent(scopes, PublicClientApplication.OperatingSystemAccount).ExecuteAsync().ConfigureAwait(false);
-
+                Assert.IsTrue(false, "AcquireTokenSilent successful!");
```
pmaytak, line 53:
```suggestion
                Assert.Fail(" AcquireTokenSilent was successful. MsalUiRequiredException should have been thrown.");
```

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -50,13 +50,16 @@ public async Task WamSilentAuthUserInteractionRequiredAsync()
...
             }
             catch (MsalUiRequiredException ex)
             {
                 Assert.IsTrue(!string.IsNullOrEmpty(ex.ErrorCode));
             }
-
+            catch (Exception ex)
```
pmaytak, line 59: nit: don't think this is needed. We can let the test just fail through.

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -53,6 +53,7 @@ public void Init()
             _coreUIParent = new CoreUIParent() { SynchronizationContext = _synchronizationContext };
             ApplicationConfiguration applicationConfiguration = new ApplicationConfiguration(isConfidentialClient: false);
             _logger = Substitute.For<ILoggerAdapter>();
+            _logger.PiiLoggingEnabled.Returns(true);
```
pmaytak, line 56: This doesn't do anything, since there are not pii related tests in this class? Are they missing?

pmaytak (commented):
I think [LogEventRaised ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/01c52461b643352b0b375cdc4ed2cd891dbaa463/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs#L117-L120)needs to be updated also to log both messages.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
SameerK-MSFT: This is to catch any other exception

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
SameerK-MSFT: Added WamUsernamePasswordRequestAsync_WithPiiAsync

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -106,20 +111,18 @@ private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
...
-                string msgWithPii = string.Empty;
-                string msgNoPii = string.Empty;
+                string msgWithPii = _logger.PiiLoggingEnabled ? args.Message : string.Empty;
+                string msgNoPii = !_logger.PiiLoggingEnabled ? args.Message : string.Empty;
 
-                // as of now, there will be no Pii.
-                // TODO change when Pii is added
-                // https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3822
-                _logger.Log(msalLogLevel, string.Empty, args.Message);
+                _logger.Log(msalLogLevel, msgWithPii, msgNoPii);
```
pmaytak, line 117: nit: maybe something like this instead? This way we don't allocate a duplicate string each time; which may be an issue if LogEventRaised gets called a lot.

```csharp
                if (_logger.PiiLoggingEnabled)
                {
                    _logger.Log(msalLogLevel, args.Message, string.Empty);
                }
                else
                {
                    _logger.Log(msalLogLevel, string.Empty, args.Message);
                }
```

pmaytak (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Done.

Issue #3829, opened 2022-11-29
[Feature Request] Add IMDS managed identity source to MSI integration


neha-bhargava:
Add IMDS managed identity source.

- [ ] Add code for IMDS source
- [ ] Test the flow manually
- [ ] Add unit tests 
- [ ] Add error handling 

For more details see: #3754

PR #3849, opened 2022-12-05
Add IMDS source to managed identity

neha-bhargava:
Fixes #
#3829 

- [x] Add code for IMDS source
- [x] Test the flow manually
- [x] Add unit tests 
- [x] Add error handling 

**Testing**
Unit tests

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -0,0 +1,167 @@
...
+            catch (MsalServiceException e) when (e.ErrorCode == MsalError.ManagedIdentityRequestFailed)
+            {
+                _requestContext.Logger.Error(NoResponseError + e.Message);
+                throw;
+            }
+            catch (TaskCanceledException)
+            {
+                _requestContext.Logger.Error(TimeoutError);
+                throw;
+            }
```
pmaytak, line 103: Can this be moved to wrap the HTTP calls in `ManagedIdentitySource.AuthenticateAsync` instead?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -0,0 +1,167 @@
...
+                _requestContext.Logger.Verbose("[Managed Identity] Adding user assigned client id to the request.");
+                request.QueryParameters[Constants.ManagedIdentityClientId] = _clientId;
+            }
+            if (!string.IsNullOrEmpty(_resourceId))
+            {
+                _requestContext.Logger.Verbose("[Managed Identity] Adding user assigned resource id to the request.");
+                request.QueryParameters[Constants.ManagedIdentityResourceId] = _resourceId;
+            }
+
+            request.ComputeUri();
```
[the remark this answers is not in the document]
bgavrilMS: I agree. Ideally every parameter in this `ManagedIdentityRequest` class would be read-only / immutable (and they are, except for Endpoint), and you'd only call `ComputeUri` when you need the URI.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsTests.cs
```diff
@@ -0,0 +1,238 @@
...
+        [DataRow(null, DefaultResource, DefaultResource, null)]
+        [DataRow("http://169.254.169.254", "https://management.azure.com/.default", DefaultResource, null)]
+        [DataRow("http://169.254.169.254", "https://management.azure.com/.default", DefaultResource, TestConstants.ClientId, UserAssignedIdentityId.ClientId)]
+        [DataRow("http://169.254.169.254", "https://management.azure.com/.default", DefaultResource, "resource_id", UserAssignedIdentityId.ResourceId)]
+        [DataRow("http://169.254.169.254", "https://management.azure.com/.default", DefaultResource, "", UserAssignedIdentityId.None)]
+        [DataRow("http://169.254.169.254", "https://management.azure.com/.default", DefaultResource, "  ", UserAssignedIdentityId.None)]
+        public async Task ImdsHappyPathAsync(
+            string endpoint, 
+            string scope, 
+            string resource,
```
bgavrilMS, line 46: This param is always set to `DefaultResource`. Do you actually need it?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -0,0 +1,167 @@
...
+                _requestContext.Logger.Error(TimeoutError);
+                throw;
+            }
+        }
+
+        protected override ManagedIdentityResponse HandleResponse(
+            AppTokenProviderParameters parameters, 
+            HttpResponse response)
+        {
+            // handle error status codes indicating managed identity is not available
```
bgavrilMS, line 110: Please log the actual status code

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -0,0 +1,167 @@
...
+            var baseMessage = response.StatusCode switch
+            {
+                HttpStatusCode.BadRequest => IdentityUnavailableError,
+                HttpStatusCode.BadGateway => GatewayError,
+                HttpStatusCode.GatewayTimeout => GatewayError,
+                _ => default(string)
+            };
```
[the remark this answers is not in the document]
bgavrilMS: Agreed, it's pretty cool, although a bit difficult to read. But prompted me to find the article on C# pattern matching https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#compare-discrete-values

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+            _clientId = requestContext.ServiceBundle.Config.ManagedIdentityUserAssignedClientId;
+            _resourceId = requestContext.ServiceBundle.Config.ManagedIdentityUserAssignedResourceId;
+        }
+
+        protected override ManagedIdentityRequest CreateRequest(string[] scopes)
+        {
+            // covert the scopes to a resource string
+            string resource = ScopeHelper.ScopesToResource(scopes);
```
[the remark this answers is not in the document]
neha-bhargava: Yeah good point. I think we can make this change.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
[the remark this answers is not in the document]
neha-bhargava: I also liked this format of switch statement. Less lines of code :)

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -74,8 +74,6 @@ public AcquireTokenForClientParameterBuilder WithSendX5C(bool withSendX5C)
         /// Specifies to use the Managed Identity to fetch the token from the Managed Identity Endpoint.
         /// For more details see https://aka.ms/msal-net-managed-identity.
         /// </summary>
-        /// <param name="userAssignedClientIdOrResourceId">This is an optional parameter to pass client id or resource id for a user assigned managed identity. 
-        /// For system assigned managed identity, no need to pass a value to this parameter. </param>
```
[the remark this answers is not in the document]
neha-bhargava: Not sure how it got removed. Adding back.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/AppServiceManagedIdentitySource.cs
```diff
@@ -83,14 +80,12 @@ protected override ManagedIdentityRequest CreateRequest(string[] scopes)
                 _requestContext.Logger.Info("[Managed Identity] Adding user assigned client id to the request.");
                 request.QueryParameters[Constants.ManagedIdentityClientId] = _clientId;
             }
-
-            if (!string.IsNullOrEmpty(_resourceId))
+            else if (!string.IsNullOrEmpty(_resourceId))
```
trwalke, line 83: Why is the else needed? Are you not able to add both client id and resource id to the QueryParameters?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -29,14 +30,31 @@ protected ManagedIdentitySource(RequestContext requestContext)
 
         public virtual async Task<ManagedIdentityResponse> AuthenticateAsync(AppTokenProviderParameters parameters, CancellationToken cancellationToken)
         {
-            ManagedIdentityRequest request = CreateRequest(parameters.Scopes.ToArray());
+            // Convert the scopes to a resource string.
+            string resource = ScopeHelper.ScopesToResource(parameters.Scopes.ToArray());
 
-            HttpResponse response =
+            if (cancellationToken.IsCancellationRequested)
```
trwalke, line 36: Shouldn't this also be higher than the ScopesToResource(). Should probably be the first thing checked for.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -29,14 +30,31 @@ protected ManagedIdentitySource(RequestContext requestContext)
...
+
+            ManagedIdentityRequest request = CreateRequest(resource);
+
+            try
+            {
+                HttpResponse response =
             request.Method == HttpMethod.Get ?
-            await _requestContext.ServiceBundle.HttpManager.SendGetForceResponseAsync(request.Endpoint, request.Headers, _requestContext.Logger, cancellationToken: cancellationToken).ConfigureAwait(false) :
-            await _requestContext.ServiceBundle.HttpManager.SendPostForceResponseAsync(request.Endpoint, request.Headers, request.BodyParameters, _requestContext.Logger, cancellationToken: cancellationToken).ConfigureAwait(false);
+            await _requestContext.ServiceBundle.HttpManager.SendGetForceResponseAsync(request.ComputeUri(), request.Headers, _requestContext.Logger, cancellationToken: cancellationToken).ConfigureAwait(false) :
```
trwalke, line 48: Is it possible to compute the uri only once and store it as a property rather than computing it multiple times.

trwalke (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/AppServiceManagedIdentitySource.cs
```diff
@@ -67,12 +67,9 @@ private static bool TryValidateEnvVars(string msiEndpoint, string secret, ILogge
...
         }
 
-        protected override ManagedIdentityRequest CreateRequest(string[] scopes)
+        protected override ManagedIdentityRequest CreateRequest(string resource)
         {
-            // convert the scopes to a resource string
-            string resource = ScopeHelper.ScopesToResource(scopes);
-
-            ManagedIdentityRequest request = new ManagedIdentityRequest(System.Net.Http.HttpMethod.Get, _endpoint);
+            ManagedIdentityRequest request = new(System.Net.Http.HttpMethod.Get, _endpoint);
```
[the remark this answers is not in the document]
neha-bhargava: Its a constructor. VS suggestion :)

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityRequest.cs
```diff
@@ -13,9 +13,9 @@ namespace Microsoft.Identity.Client.ManagedIdentity
 {
     internal class ManagedIdentityRequest
     {
-        public Uri Endpoint { get; private set; }
+        public Lazy<Uri> Endpoint => new Lazy<Uri>(() => ComputeUri(this)) ;
```
pmaytak, line 16: `Lazy` executes only on the first call. So if `Endpoint` is called, then some other query params are added, then `Endpoint is called again - it will return the original. This works correctly with the current code, but can introduce bugs down the line.

pmaytak (approved):
Approving but there maybe be a potential bug in ManagedIdentityRequest.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityRequest.cs
bgavrilMS: The `ComputeUri` does not seem to be that expensive / used too much. If there are perf concerns, we should measure first and then optimize. 
Please remove the Lazy here, beucase the class is not immutable (all those collections are mutable because you can add / remove elements from them). Hence `ComputeUri` needs to always re-compute URI.

neha-bhargava: Updated.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -82,13 +100,21 @@ protected ManagedIdentityResponse GetSuccessfulResponse(HttpResponse response)
             return managedIdentityResponse;
         }
 
-        internal static string GetMessageFromResponse(HttpResponse response)
+        internal string GetMessageFromErrorResponse(HttpResponse response)
         {
             var managedIdentityResponse = JsonHelper.TryToDeserializeFromJson<ManagedIdentityErrorResponse>(response?.Body);
```
gladjohn, line 105:
```suggestion
            var managedIdentityErrorResponse = JsonHelper.TryToDeserializeFromJson<ManagedIdentityErrorResponse>(response?.Body);
```

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
```diff
@@ -95,11 +95,31 @@ public static string GetMsiSuccessfulResponse()
           "\"Bearer\",\"client_id\":\"client_id\"}";
         }
 
+        public static string GetMsiImdsSuccessfulResponse()
+        {
+            string expiresOn = DateTimeHelpers.DateTimeToUnixTimestamp(DateTime.UtcNow.AddHours(1));
+            return
+          "{\"access_token\":\"" + TestConstants.ATSecret + "\",\"client_id\":\"client-id\"," +
+          "\"expires_in\":\"12345\",\"expires_on\":\"" + expiresOn + "\",\"resource\":\"https://management.azure.com/\"," +
```
gladjohn, line 103: not_before is also returned for IMDS

gladjohn (approved):

gladjohn:
everything looks good for IMDS signed off, but while testing I am not able to get the clientID in the response for a user managed identity. MSI endpoint exposes this, do we need to expose it as well?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -54,20 +72,20 @@ public virtual async Task<ManagedIdentityResponse> AuthenticateAsync(AppTokenPro
                     return GetSuccessfulResponse(response);
```
gladjohn, line 72: seems like we are ignoring clientID in this response and not passing it on to MSALTokenResponse. MSI Endpoint gives clientID as well. Do we know if this needs to be passed on?

neha-bhargava: We can only send `AccessToken`, `ExpiresIn` and `RefreshIn` to MSAL through [AppTokenProviderResult.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Extensibility/AppTokenProviderResult.cs#L13)

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
neha-bhargava: Not used here as we don't extract it.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -0,0 +1,150 @@
...
+                requestContext.Logger.Verbose("[Managed Identity] Environment variable for IMDS returned endpoint: " + EnvironmentVariables.PodIdentityEndpoint);
+                var builder = new UriBuilder(EnvironmentVariables.PodIdentityEndpoint)
+                {
+                    Path = ImdsTokenPath
+                };
+                _imdsEndpoint = builder.Uri;
+			}
+			else
+			{
+                requestContext.Logger.Verbose("[Managed Identity] Unable to find an endpoint in environment variable for IMDS, using the default endpoint.");
```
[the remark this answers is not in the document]
neha-bhargava: I don't think we should expose this environment variable in the logs. We don't want anyone to set the environment variable looking at this.

src/client/Microsoft.Identity.Client/ManagedIdentity/EnvironmentVariables.cs
```diff
@@ -12,5 +12,6 @@ internal class EnvironmentVariables
     {
         public static string IdentityEndpoint => Environment.GetEnvironmentVariable("IDENTITY_ENDPOINT");
         public static string IdentityHeader => Environment.GetEnvironmentVariable("IDENTITY_HEADER");
+        public static string PodIdentityEndpoint => Environment.GetEnvironmentVariable("AZURE_POD_IDENTITY_AUTHORITY_HOST");
```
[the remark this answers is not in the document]
neha-bhargava: This the what it is in the azure sdk code. I did not find this on the VM as well but we do have the endpoint hardcoded.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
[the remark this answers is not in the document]
bgavrilMS: The SDK is open source, so it's better to prioritize supportability over security by obscurity.

PR #3855, opened 2022-12-09
Proxy HTTP instead of Managed Identity

bgavrilMS:
Shows an example of prox-ing HTTP call at a more general level

Sadly I could not get it to work end to end, but it should be close.

### DIFF
--- a/tests/Microsoft.Identity.Test.Common/Http/ProxyHttpClientFactory.cs
+++ b/tests/Microsoft.Identity.Test.Common/Http/ProxyHttpClientFactory.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using System.Text;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Test.Common.Http
+{
+    public class RemoteHttpFactory : IMsalHttpClientFactory
+    {
+        private readonly HttpClient _httpClient;
+
+        public RemoteHttpFactory(Uri proxyUri)
+        {
+
+            _httpClient = new HttpClient(new RemoteServiceHostHandler(proxyUri));
+        }
+
+        public HttpClient GetHttpClient()
+        {
+            return _httpClient;
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


tests/Microsoft.Identity.Test.Common/Http/ProxyHttpClientFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using System.Text;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Test.Common.Http
+{
+    public class RemoteHttpFactory : IMsalHttpClientFactory
```
bgavrilMS, line 11: `IMsalHttpClientFactory` can easily be injected in MSAL.

tests/devapps/MSI/HelperService/MSIHelperService/Controllers/GetMSITokenController.cs
```diff
@@ -2,11 +2,36 @@
...
+        private readonly ILogger _logger;
+        private readonly IHttpClientFactory _httpClientFactory;
+
+        public HttpController(ILogger<GetMSITokenController> logger, IHttpClientFactory httpClientFactory)
+        {
+            _logger = logger;
+            _httpClientFactory = httpClientFactory;
+        }
+
+        [HttpPost]
```
bgavrilMS, line 25: Not sure if POST is needed, but probably safer that GET since url length can be an issue.

tests/devapps/NetCoreTestApp/Program.cs
```diff
@@ -68,9 +69,12 @@ private static string GetAuthority()
 
         private static IPublicClientApplication CreatePca()
         {
+            var proxyHttpFactory = new RemoteHttpFactory(new Uri("https://localhost:7059/Http"));
+            
             var pcaBuilder = PublicClientApplicationBuilder
                             .Create(s_clientIdForPublicApp)
                             .WithAuthority(GetAuthority())
+                            .WithHttpClientFactory(proxyHttpFactory)
```
bgavrilMS, line 77: Example of how to use this with MSAL

Issue #3843, opened 2022-12-01
[Feature Request] MSAL MSI Integration test for App Service


gladjohn:
Add Integration test for App Service managed identity source.

- [x] Call into the helper Service from MSAL Integration test project
- [x] Add App Service test specific integration tests
- [x] Fine tune MSI Helper service as needed to support the integration tests

PR #3857, opened 2022-12-10
[MSI] Integration tests for MSI source (Web App / Function App / VM)

gladjohn:
Fixes [#3843](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3843)

- Calls into the msi helper service to get environment variables based on resource type
- sets environment variable
- calls the azure resource to get the msi token
- test coverage for Function App, Web App and VM

**Testing**
n/a

**Performance impact**
none

**Documentation**


tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+                AuthenticationResult authenticationResult = await cca.AcquireTokenForClient(s_msi_scopes)
+                    .WithManagedIdentity()
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+            }
+            catch (MsalClientException ex)
+            {
+                result = ex.Message;
+            }
+            catch (Exception ex)
```
bgavrilMS, line 66: We should not catch general exceptions here only MsalClientException

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+        private static readonly string[] s_msi_scopes = { "https://management.azure.com" };
+        private static readonly string s_baseURL = "https://service.msidlab.com/";
+        private static readonly string s_clientId = "client_id";
+        private static Dictionary<string, string> s_envVariables = new Dictionary<string, string>();
+        private const string UserAssignedClientID = "3b57c42c-3201-4295-ae27-d6baec5b7027";
+        private const string Mi_res_id = "/subscriptions/c1686c51-b717-4fe0-9af3-24a20a41fb0c/resourcegroups/MSAL_MSI/providers/Microsoft.ManagedIdentity/userAssignedIdentities/MSAL_MSI_USERID";
+
+        public enum AzureResource
+        {
+            webapp,
```
bgavrilMS, line 35: Naming - use UpperCase for enum members. If you need to transfor to string, use ToString().ToLowerInvariant()

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/HttpProxyManager.cs
```diff
@@ -0,0 +1,162 @@
...
+            CancellationToken cancellationToken = default)
+        {
+            //Get token for the MSIHelperService
+            var labApi = new LabServiceApi();
+            var token = await labApi.GetMSIHelperServiceTokenAsync()
+                .ConfigureAwait(false);
+
+            //Add the Authorization header
+            s_httpClient.DefaultRequestHeaders.Authorization = 
+                new AuthenticationHeaderValue("Bearer", token);
```
bgavrilMS, line 131: Cool

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+        [DataRow(AzureResource.function, UserAssignedClientID, DisplayName = "User Identity Function App")]
+        [DataRow(AzureResource.webapp, Mi_res_id, DisplayName = "ResourceID Web App")]
+        [DataRow(AzureResource.function, Mi_res_id, DisplayName = "ResourceID Function App")]
+        public async Task AcquireMSITokenAsync(AzureResource azureResource, string userIdentity)
+        {
+            //Arrange
+            AuthenticationResult result = null;
+
+            //Set the Environment Variables
+            bool isEnvironmentVariableSet = await SetEnvironmentVariablesAsync(azureResource)
```
bgavrilMS, line 89: I think what you are doing here is "downloading" all the env variables from the service onto the box. The name should suggest that, e.g. `SetRemoveEnvVariablesAsync` or smth like that.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+                .ExecuteAsync().ConfigureAwait(false);
+
+            Assert.IsNotNull(result.Scopes);
+            Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+
+            ClearEnvironmentVariables();
+
+        }
+
+        private async Task<bool> SetEnvironmentVariablesAsync(AzureResource resource)
```
bgavrilMS, line 138: Not a fan of returning true / false. C# pattern is to use exceptions, because they convey more information.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+                .ConfigureAwait(false);
+
+            //process the response
+            if (!string.IsNullOrEmpty(environmentVariableResponse))
+            {
+                s_envVariables = JsonConvert.DeserializeObject<Dictionary<string, string>>(environmentVariableResponse);
+            }
+            else
+            {
+                return false;
```
bgavrilMS, line 154: If this fails, throw an exception. The exception will propagate and fail the test.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+            IConfidentialClientApplication cca = CreateCCAWithProxy(uri);
+
+            //Act
+            try
+            {
+                result = await cca.AcquireTokenForClient(s_msi_scopes)
+                    .WithManagedIdentity(userIdentity)
+                    .ExecuteAsync().ConfigureAwait(false);
+            }
+            catch (Exception ex)
```
bgavrilMS, line 110: Why don't you just let the exception propagate and fail the test?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                Trace.WriteLine(ex);
+                Assert.IsNull(ex.Message);
+            }
+
+            //Assert
+            Assert.IsNotNull(result.AccessToken);
```
bgavrilMS, line 117: It's never null, you can remove this.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+            Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
```
bgavrilMS, line 118: Can you add an assert on the expiration. Something like expiration is between 0 and 48h or smth? I've seen bugs where the ExpireOn field was set years in advance and we didn't catch it.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+                Environment.SetEnvironmentVariable(kvp.Key, kvp.Value);
+            }
+
+            return true;
+        }
+
+        private void ClearEnvironmentVariables()
+        {
+            //Clear the environment variables
+            foreach (KeyValuePair<string, string> kvp in s_envVariables)
```
bgavrilMS, line 169: If env variable A is already present on the box, e.g. A = value1 and you download it from the remote, e.g. A = value2, when you call Clear() you'll reset it. But really what you want to do is to set it back to value1.

Have a look at https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3849/files#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154R35 - it captures env variables at point X and on Dispose it resets the state to point X.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        [DataTestMethod]
+        [DataRow(AzureResource.webapp, "", DisplayName = "System Identity Web App")]
+        [DataRow(AzureResource.function, "", DisplayName = "System Identity Function App")]
+        [DataRow(AzureResource.webapp, UserAssignedClientID, DisplayName = "User Identity Web App")]
+        [DataRow(AzureResource.function, UserAssignedClientID, DisplayName = "User Identity Function App")]
+        [DataRow(AzureResource.webapp, Mi_res_id, DisplayName = "ResourceID Web App")]
+        [DataRow(AzureResource.function, Mi_res_id, DisplayName = "ResourceID Function App")]
+        public async Task AcquireMSITokenAsync(AzureResource azureResource, string userIdentity)
```
bgavrilMS, line 83: This is very cool!

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: added please take a look

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: Fixed all comments except this one, please review the rest. I need to merge main into my branch and work on this, I will hopefully get to this tomorrow. Will update once I am done with it. I also see @neha-bhargava  has merged IMDS, I will also update the tests for IMDS once I get the main

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,325 @@
...
+            Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+
+            ClearEnvironmentVariables(environmentVariables);
+
+        }
+
+        [DataTestMethod]
+        [DataRow(MsiAzureResource.WebApp, "", SendMSIHeader.None, DisplayName = "Web App No Header")]
+        [DataRow(MsiAzureResource.WebApp, "", SendMSIHeader.WithWrongValue, DisplayName = "Web App Wrong Header")]
+        public async Task TryAcquireMSITokenHeaderTestAsync(
```
gladjohn, line 132: @bgavrilMS @neha-bhargava please take a look at these tests, most of these may not be applicable since I am tweaking the request before it is sent out. But find a few cases where the error message is missing a co-correlation ID or cases where we do not throw the MSI endpoint error message.

gladjohn: I disabled these as they may not be useful for integration, but @neha-bhargava just the https://portal.azure.com token did not work, maybe you can see why?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
bgavrilMS: How did you resolve this?

tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -460,4 +460,34 @@ internal static class Adfs2019LabConstants
...
+    {
+        WebApp,
+        Function,
+        Vm,
+        AzureArc,
+        CloudShell,
+        ServiceFabric
+    }
+
+    public enum SendMSIHeader
```
bgavrilMS, line 474: It's not clear what these enums do, please add comments.

tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -460,4 +460,34 @@ internal static class Adfs2019LabConstants
...
+    {
+        Original,
+        WithWrongValue,
```
bgavrilMS, line 477: We have not really discussed negative testing - but in negative scenarios we are: 

1. testing to ensure correct, easy to understand exceptions are thrown back to user (this is unit test territory, try not to use integration tests for these since they are expensive)
2. ensure that the integration point (e.g. MSI service) behaves in a consistent way when hit with "bad" values. 

I'd like to better understand what the negative test strategy is first before reviewing code. Also, to keep PRs reasonable, aim to deliver the smallest piece of code that provides value - the positive tests are enough. Negative tests should be added as separate PR.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
bgavrilMS: When would we send MSI header with none or wrong value?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: resolved

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: sorry missed this, added an alternate test to check for length, so we make sure that the token is always trimmed

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: we get the value and then the value changes in the web app. Web restarts or MSI endpoint updates can change the header value. We can discuss more when we write the negative cases

tests/Microsoft.Identity.Test.Common/TestConstants.cs
gladjohn: removed negative cases

tests/Microsoft.Identity.Test.Common/TestConstants.cs
gladjohn: added

gladjohn:
> Please create smaller PR with infra + positive tests and resolve comments. If you do not agree with a comment, please reply or reach out directly to reviewers.

removed negative cases and updated the PR, also there is a separate PR for infra

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        private const string Mi_res_id = "/subscriptions/c1686c51-b717-4fe0-9af3-24a20a41fb0c/" +
+            "resourcegroups/MSAL_MSI/providers/Microsoft.ManagedIdentity/userAssignedIdentities/" +
+            "MSAL_MSI_USERID";
+
+        [TestMethod]
+        public async Task ManagedIdentitySourceCheckAsync()
+        {
+            //Arrange
+            string result = string.Empty;
+            string expectedClientException = "[Managed Identity] Authentication unavailable. " +
```
neha-bhargava, line 47: I am moving the error messages to MsalErrorMessage class so that these can be referred in the tests.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
neha-bhargava: Should we tweak the request sent out? Will this ever happen?

neha-bhargava: Yes, lets discuss these.

neha-bhargava (approved):
Looks good. Thanks.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+            }
+            catch (MsalServiceException ex)
+            {
+                result = ex.Message;
+            }
+
+            //Assert
+            Assert.AreSame(result, expectedClientException);
```
pmaytak, line 67: Use [Assert.ThrowsException](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/8e21adc1c5f3c4440044f07b0edcdcc25fe0ba2a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/AppServiceTests.cs#L95)

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/HttpProxyManager.cs
```diff
@@ -0,0 +1,162 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
pmaytak, line 1: Rename the file to match the class name.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                {
+                    result = await cca
+                        .AcquireTokenForClient(s_msi_scopes)
+                        .WithManagedIdentity(userIdentity)
+                        .ExecuteAsync().ConfigureAwait(false);
+                }
+                catch (MsalClientException ex)
+                {
+                    throw new Exception(ex.Message);
+                }
```
pmaytak, line 109: Not needed? Just let it bubble up and break the test?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            
+                //Assert
+                //1. MSI Helper service trims the access token, so that the access token from the MSI
+                //   is not fully returned to the calling service or test for security reasons
+                Assert.IsTrue(result.AccessToken.Length == 28);
```
pmaytak, line 114: The length requirement seems a bit flaky. Maybe AT returned by the service should just be some constant - "MSI_token" or something?

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/HttpProxyManager.cs
```diff
@@ -0,0 +1,162 @@
...
+{
+    /// <remarks>
+    /// We invoke this class from different threads and they all use the same HttpClient.
+    /// To prevent race conditions, make sure you do not get / set anything on HttpClient itself,
+    /// instead rely on HttpRequest objects which are thread specific.
+    ///
+    /// In particular, do not change any properties on HttpClient such as BaseAddress, buffer sizes and Timeout. You should
+    /// also not access DefaultRequestHeaders because the getters are not thread safe (use HttpRequestMessage.Headers instead).
+    /// </remarks>
+    internal class ProxyHttpManager : IHttpManager
```
pmaytak, line 29: Maybe this could extend HttpManager, to reduce code duplication, and the ExecuteAsync that has logic can be something like protected override.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: modified the test

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/HttpProxyManager.cs
gladjohn: Yes we can, but the idea is to extend the proxyHttpManager for other http methods as well. that's the reason of implementing iHttpManager.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        [DataRow(MsiAzureResource.WebApp, UserAssignedClientID, DisplayName = "User Identity Web App")]
+        [DataRow(MsiAzureResource.Function, UserAssignedClientID, DisplayName = "User Identity Function App")]
+        [DataRow(MsiAzureResource.Vm, UserAssignedClientID, DisplayName = "User Identity Virtual Machine")]
+        [DataRow(MsiAzureResource.WebApp, Mi_res_id, DisplayName = "ResourceID Web App")]
+        [DataRow(MsiAzureResource.Function, Mi_res_id, DisplayName = "ResourceID Function App")]
+        [DataRow(MsiAzureResource.Vm, Mi_res_id, DisplayName = "ResourceID Virtual Machine")]
+        public async Task AcquireMSITokenAsync(MsiAzureResource azureResource, string userIdentity)
+        {
+            //Arrange
+
```
SameerK-MSFT, line 74: nit: extra line

SameerK-MSFT (approved):
minor extra lines (more than 1 place)

tests/Microsoft.Identity.Test.LabInfrastructure/LabServiceApi.cs
```diff
@@ -161,5 +161,17 @@ public async Task<string> GetUserSecretAsync(string lab)
             string result = await SendLabRequestAsync(LabApiConstants.LabUserCredentialEndpoint, queryDict).ConfigureAwait(false);
             return JsonConvert.DeserializeObject<LabCredentialResponse>(result).Secret;
         }
+
+        public async Task<string> GetMSIHelperServiceTokenAsync()
```
trwalke, line 165: Isnt this the same token we are already using for other things? maybe it is better to make the token a property or something that can be accessed when needed rather than making a specific method for MSI?

gladjohn: it's the same app but since the service is not hosted under the same lab api, added a different secret so the resources do not share the same secret. This way external teams can use the service with a different secret

tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -462,4 +462,15 @@ internal static class Adfs2019LabConstants
         public const string ADFS2019ClientSecretURL = "https://buildautomation.vault.azure.net/secrets/ADFS2019ClientCredSecret/";
         public const string ADFS2019ClientSecretName = "ADFS2019ClientCredSecret";
     }
+
+    //MSI Azure resources
+    public enum MsiAzureResource
```
neha-bhargava, line 467: In the MSI PR, I have created a class to hold MSI related enums, lets place this enum there as well: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/nebharg%2FMsiApiPrototype/tests/Microsoft.Identity.Test.Common/Core/Helpers/ManagedIdentityTestUtil.cs#L10

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,266 @@
...
+                MsalServiceException ex = await AssertException.TaskThrowsAsync<MsalServiceException>(async () =>
+                {
+                    await cca
+                    .AcquireTokenForClient(s_msi_scopes)
+                    .WithManagedIdentity(userIdentity)
+                    .ExecuteAsync().ConfigureAwait(false);
+                }).ConfigureAwait(false);
+
+                //Assert
+                Assert.IsTrue(ex.Message.Contains(UserAssignedIdDoesNotExist));
```
neha-bhargava, line 166: Assert error code instead of error message? Since the error message is based on the server response.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,266 @@
...
+                MsalServiceException ex = await AssertException.TaskThrowsAsync<MsalServiceException>(async () =>
+                {
+                    await cca
+                    .AcquireTokenForClient(s_wrong_msi_scopes)
+                    .WithManagedIdentity(userIdentity)
+                    .ExecuteAsync().ConfigureAwait(false);
+                }).ConfigureAwait(false);
+
+                //Assert
+                Assert.IsTrue(ex.Message.Contains(WrongScopesErrorMessage));
```
neha-bhargava, line 203: Assert error code.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,266 @@
...
+
+        //Non Existent User Assigned Client ID 
+        private const string NonExistentUserAssignedClientID = "3b57c42c-9999-9999-zz99-d6baec5b7027";
+
+        //Error Messages
+        private const string UserAssignedIdDoesNotExist = "[Managed Identity] Error Message: No User Assigned or Delegated Managed Identity found for specified ClientId/ResourceId/PrincipalId.";
+        private const string WrongScopesErrorMessage = "An unexpected error occured while fetching the AAD Token";
+
+        //Resource ID of the User Assigned Identity 
+        private const string Mi_res_id = "/subscriptions/c1686c51-b717-4fe0-9af3-24a20a41fb0c/" +
```
neha-bhargava, line 46: nit:
```suggestion
        private const string UamiResourceId = "/subscriptions/c1686c51-b717-4fe0-9af3-24a20a41fb0c/" +
```

neha-bhargava (approved):
Looks good. Mainly, I would suggest to assert the error code instead of error message since the error message is not a constant but is constructed from response.

tests/Microsoft.Identity.Test.Common/TestConstants.cs
gladjohn: I will wait for you to merge and then take this change

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: this seems to be a very generic error code but the message seems to be specific in this scenario. thoughts? 

image

gladjohn:
> Please create smaller PR with infra + positive tests and resolve comments. If you do not agree with a comment, please reply or reach out directly to reviewers.

@bgavrilMS can you please review the msi helper service?

pmaytak (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,301 @@
...
+
+        [TestMethod]
+        public async Task ManagedIdentity_WithoutEnvironmentVariables_ThrowsAsync()
+        {
+            //Arrange
+            string result = string.Empty;
+
+            IConfidentialClientApplication cca = CreateCCAWithProxy(s_baseURL);
+
+            MsalServiceException ex = await AssertException.TaskThrowsAsync<MsalServiceException>(async () =>
```
neha-bhargava, line 60: I don't think this will be MsalServiceException. When there is no environment variable, IMDS source is created and tries to acquire token from IMDS endpoint which results in HttpRequestException ` System.Net.Http.HttpRequestException: A socket operation was attempted to an unreachable network.`

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
neha-bhargava: I am thinking how will asserting this message help us? Are we going to take some action if the error message changes?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,301 @@
...
+
+                //Assert
+                Assert.IsTrue(ex.Message.Contains(UserAssignedIdDoesNotExist));
+            }
+        }
+
+        [DataTestMethod]
+        [DataRow(MsiAzureResource.Function, NonExistentUserAssignedClientID, DisplayName = "User Identity Function App")]
+        [DataRow(MsiAzureResource.Function, Non_Existent_UamiResourceId, DisplayName = "ResourceID Function App")]
+        public async Task FunctionAppErrorNotInExpectedFormatAsync(MsiAzureResource azureResource, string userIdentity)
```
neha-bhargava, line 172: nit: This can be combined with the test for web app?

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/ProxyHttpManager.cs
```diff
@@ -0,0 +1,162 @@
...
+            ILoggerAdapter logger,
+            bool retry = true,
+            CancellationToken cancellationToken = default)
+        {
+            return await ExecuteAsync(endpoint, headers, null, HttpMethod.Get, logger, cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Performs the GET request just like <see cref="SendGetAsync(Uri, IDictionary{string, string}, ILoggerAdapter, bool, CancellationToken)"/>
+        /// but does not throw a ServiceUnavailable service exception. Instead, it returns the <see cref="HttpResponse"/> associated
```
neha-bhargava, line 73: nit
```suggestion
        /// but does not throw a ServiceUnavailable exception. Instead, it returns the <see cref="HttpResponse"/> associated
```

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/ProxyHttpManager.cs
```diff
@@ -0,0 +1,162 @@
...
+            ILoggerAdapter logger,
+            bool retry = true,
+            CancellationToken cancellationToken = default)
+        {
+            return await ExecuteAsync(endpoint, headers, null, HttpMethod.Get, logger, cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Performs the POST request just like <see cref="SendPostAsync(Uri, IDictionary{string, string}, IDictionary{String, String}, ILoggerAdapter, CancellationToken)"/>
+        /// but does not throw a ServiceUnavailable service exception. Instead, it returns the <see cref="HttpResponse"/> associated
```
neha-bhargava, line 88:
```suggestion
        /// but does not throw a ServiceUnavailable exception. Instead, it returns the <see cref="HttpResponse"/> associated
```

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/ProxyHttpManager.cs
```diff
@@ -0,0 +1,162 @@
...
+            IDictionary<string, string> bodyParameters,
+            ILoggerAdapter logger,
+            CancellationToken cancellationToken = default)
+        {
+            throw new NotSupportedException();
+        }
+
+        /// <summary>
+        /// Performs the POST request just like <see cref="SendPostAsync(Uri, IDictionary{string, string}, HttpContent, ILoggerAdapter, CancellationToken)"/>
+        /// but does not throw a ServiceUnavailable service exception. Instead, it returns the <see cref="HttpResponse"/> associated
```
neha-bhargava, line 103:
```suggestion
        /// but does not throw a ServiceUnavailable exception. Instead, it returns the <see cref="HttpResponse"/> associated
```

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,301 @@
...
+        private static readonly string s_clientId = "client_id";
+
+        //http proxy base URL 
+        private static readonly string s_baseURL = "https://service.msidlab.com/";
+
+        //Shared User Assigned Client ID
+        private const string UserAssignedClientID = "3b57c42c-3201-4295-ae27-d6baec5b7027";
+
+        //Non Existent User Assigned Client ID 
+        private const string NonExistentUserAssignedClientID = "3b57c42c-9999-9999-zz99-d6baec5b7027";
```
bgavrilMS, line 39: Nit: don't use `zz` in there as it is an invalid GUID digit and a potential Guid.Parse in MSAL's logic will break your test. Consider using the Microsoft tenant id - that's a valid invalid clientid :)

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,301 @@
...
+        [DataRow(MsiAzureResource.VM, UserAssignedClientID, DisplayName = "User Identity Virtual Machine")]
+        [DataRow(MsiAzureResource.WebApp, UamiResourceId, DisplayName = "ResourceID Web App")]
+        [DataRow(MsiAzureResource.Function, UamiResourceId, DisplayName = "ResourceID Function App")]
+        [DataRow(MsiAzureResource.VM, UamiResourceId, DisplayName = "ResourceID Virtual Machine")]
+        public async Task AcquireMSITokenAsync(MsiAzureResource azureResource, string userIdentity)
+        {
+            //Arrange
+            using (new EnvVariableContext())
+            {
+                //Get the Environment Variables
```
bgavrilMS, line 86:
```suggestion
                // Fetch the env variables from the resource and set them locally
```

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,301 @@
...
+                        .ExecuteAsync().ConfigureAwait(false);
+            
+                //Assert
+                //1. Access token
+                Assert.IsNotNull(result.AccessToken);
+
+                //2. First token response is from the MSI Endpoint
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+
+                //3. Validate the ExpiresOn falls within a 24 hour range from now
```
bgavrilMS, line 113: Is this true? Don't they cache some tokens as well?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,301 @@
...
+                CoreAssert.IsWithinRange(
+                                DateTimeOffset.UtcNow + TimeSpan.FromHours(0),
+                                result.ExpiresOn,
+                                TimeSpan.FromHours(24));
+
+                result = await cca
+                    .AcquireTokenForClient(s_msi_scopes)
```
bgavrilMS, line 120: You'll need to rebase on top of Neha's changes (they are in main now). No more ConfidentialClient and AcquireTokenForClient

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
bgavrilMS: `Assert.InformButDontFail` :)

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: this message is from the Web App Azure Resource when you pass a wrong client id, this is not from the helper service. Helper service does not manipulate any error codes or error messages it simply passes it down

image

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: will need to test this a bit more but I do get a MSALServiceException

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: just keeping it as a separate error case

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
neha-bhargava: No there is no caching on their side.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                //Get the Environment Variables
+                Dictionary<string, string> envVariables = 
+                    await GetEnvironmentVariablesAsync(azureResource).ConfigureAwait(false);
+
+                //Set the Environment Variables
+                SetEnvironmentVariables(envVariables);
+
+                //form the http proxy URI 
+                string uri = s_baseURL + $"MSIToken?" +
```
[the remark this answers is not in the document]
gladjohn: I did not take this change as there are two endpoints. but I can revisit

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ManagedIdentityTests.cs
gladjohn: modified to HttpRequestException, the problem was going through the proxy. thanks @neha-bhargava

gladjohn:
build machine seems to have an IMDS endpoint 

image

Issue #3613, opened 2022-08-18
[Bug] If broker is enabled, MSAL still looks to refresh a cached RT, which lowers the security stance


bgavrilMS:
**Logs and network traces**
Without logs or traces, it is unlikely that the team can investigate your issue. Capturing logs and network traces is described in [Logging wiki](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/logging).

**Which version of MSAL.NET are you using?**
4.46

**Repro**

1. Create an app (Android or Desktop) and use a webview to login 
2. Upgrade the app to use the broker now, keeping the same cache location
3. Acquire a token silently

**Actual**: MSAL will look for AT in cache. If AT is valid return it. Otherwise, MSAL will look for RT in cache and use refrsh_token flow
**Expected**: MSAL should not use the local RT when broker is enabled, 

**Impact**: the CA policy that states that RT must be bounded will always block the user.

**Suggested fix**: do not attempt to look for either AT / RT when broker is enabled, the perf regression on Android is not significant.

Issue #3375, opened 2022-06-08
[Feature Request] GA the WAM Integration with MSAL C++ Interop / Runtime 


gladjohn:
## Remaining work for GA

- [x]  [Bug - on some configurations msal_runtime.dll is not found](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3964) (IN [PR](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3397))
- [x] #3447 (design almost in final stages, this should be completed by end of feb 2023) (IN [PR](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3976))

- [x] [Product Backlog Item 1897893](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1897893) and #3772 
- [x] [Task 1903740](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1903740): Automated testing for MSAL .Net
- [x] [Product Backlog Item 1913500](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1913500): MSAL .NET Partner Integration testing (Several teams have this in production already, waiting for sign-off from VS)
- [x] #3613 (in code review)
- [x] #3788
- [x] #3685
- [x] #3740 
- [x] #3822
- [x] #3458
- [x] #3927
- [x] #3952
- [x] #3964
## Non-GA issues

- [x] #3772 
- [x] #3301 
- [ ] Add automated test to validate Wam Id in account in authResult [See](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/40157d950b7314e22a842303b8a58bdd9aaba231/tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs#L857-L901)
- [x] #3961 (Will be fixed in the next MSAL release. [This is fixed in C++](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3394))
- [ ] #3965
- [x] #3916 (Will be fixed in the next MSAL release. [This is fixed in C++](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3394))

## Post-GA work

- [ ] #4006 (blocked feature needs to be added in MSAL C++ layer)

## External issues

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues?q=is%3Aissue+is%3Aopen+label%3Aexternal+label%3AWAM

## Cut

[...]

PR #3859, opened 2022-12-12
Removing local cache lookup from broker

trwalke:
Fixes #
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3613

Msal will no longer look in local cache for tokens when broker is enabled

**Testing**
Unit test.
Manual test

**Performance impact**
Negligable

**Documentation**


src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -187,8 +187,8 @@ public sealed partial class TokenCache : ITokenCacheInternal
                         requestParams.RequestContext.ApiEvent.DurationInCacheInMs += sw.ElapsedMilliseconds;
                     }
 
-                    // Don't cache PoP access tokens from broker
-                    if (msalAccessTokenCacheItem != null && !(response.TokenSource == TokenSource.Broker && response.TokenType == Constants.PoPAuthHeaderPrefix))
+                    // Don't cache access tokens from broker
+                    if (msalAccessTokenCacheItem != null && !(response.TokenSource == TokenSource.Broker))
```
bgavrilMS, line 191: To avoid re-testing Android / iOS, it would be safer to do this only on non-mobile

trwalke:
@bgavrilMS it looks like this test requires locally cached tokens to function. Is this test no longer possible since we can't return local tokens in acquireTokenSilent? Would this need to be an integration test with the WAM broker?
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/31a3ae3343a40b6312da7cf4a950161fd6dbaa41/tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs#L859

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -187,9 +187,13 @@ public sealed partial class TokenCache : ITokenCacheInternal
                         requestParams.RequestContext.ApiEvent.DurationInCacheInMs += sw.ElapsedMilliseconds;
                     }
 
-                    // Don't cache PoP access tokens from broker
-                    if (msalAccessTokenCacheItem != null && !(response.TokenSource == TokenSource.Broker && response.TokenType == Constants.PoPAuthHeaderPrefix))
-                    {
+                    // Don't cache access tokens from broker
+#if iOS
```
bgavrilMS, line 191: nit: better to place this complicated `if` condition in a private function, `ShouldPersistAccessToken()`

tests/Microsoft.Identity.Test.Unit/RequestsTests/SilentRequestTests.cs
```diff
@@ -95,35 +96,11 @@ public void ExpiredTokenRefreshFlowTest()
...
-
-        [TestMethod]
-        public async Task BrokerSilentRequestBrokerNotConfiguredAndNotInstalledTestAsync()
-        {
-            //Broker is not configured by user and is not installed
-            await BrokerSilentRequestTestExecutorAsync(false, false).ConfigureAwait(false);
-        }
-
-        public async Task BrokerSilentRequestTestExecutorAsync(bool brokerConfiguredByUser, bool brokerIsInstalledAndInvokable)
+        [DataRow(true, true)] //Broker is configured by user and is installed.
```
bgavrilMS, line 99: Can we have a unit test where we mock the broker so that it matches how iOS broker behaves? Have a look at [BrokerTests.cs ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs#L444)

tests/Microsoft.Identity.Test.Unit/RequestsTests/SilentRequestTests.cs
```diff
@@ -156,13 +133,27 @@ public async Task BrokerSilentRequestTestExecutorAsync(bool brokerConfiguredByUs
 
                 var request = new SilentRequest(harness.ServiceBundle, parameters, silentParameters);
 
-                var result = await request.RunAsync(default).ConfigureAwait(false);
+                if (brokerConfiguredByUser) //When broker is enabled, local cache should not be used
```
bgavrilMS, line 136: Is it worth splitting this 4 data row test into 2 tests? I think they would be more readable.

bgavrilMS:
> @bgavrilMS it looks like this test requires locally cached tokens to function. Is this test no longer possible since we can't return local tokens in acquireTokenSilent? Would this need to be an integration test with the WAM broker?
> 
> https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/31a3ae3343a40b6312da7cf4a950161fd6dbaa41/tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs#L859

Yes, this is a good point. But I don't think we have infrastructure for such a test, i.e. do we have guest users in the lab? I recommend doing a manual test for now and adding a task to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3375

bgavrilMS:
> > @bgavrilMS it looks like this test requires locally cached tokens to function. Is this test no longer possible since we can't return local tokens in acquireTokenSilent? Would this need to be an integration test with the WAM broker?
> > https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/31a3ae3343a40b6312da7cf4a950161fd6dbaa41/tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs#L859
> 
> Yes, this is a good point. But I don't think we have infrastructure for such a test, i.e. do we have guest users in the lab? I recommend doing a manual test for now and adding a task to #3375

I did some manual testing myself and it seems ok. I pushed a small update to the dev app.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -48,59 +48,48 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             bool isBrokerConfigured = AuthenticationRequestParameters.AppConfig.IsBrokerEnabled &&
                                       ServiceBundle.PlatformProxy.CanBrokerSupportSilentAuth();
 
-            if (isBrokerConfigured && AuthenticationRequestParameters.PopAuthenticationConfiguration != null)
```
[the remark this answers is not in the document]
trwalke: I think so, Im trying to get my iOS setup so I can test/fine tune the code.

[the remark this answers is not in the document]
trwalke: Actually, the code is fine as is, isBrokereConfigured will be false for iOS because CanBrokerSupportSilentAuth() will return false.

tests/Microsoft.Identity.Test.Unit/TestBase.cs
```diff
@@ -92,5 +94,38 @@ internal MsalTokenResponse CreateMsalRunTimeBrokerTokenResponse(string accessTok
...
+
+        internal class IosBrokerMock : NullBroker
+        {
+            public IosBrokerMock(ILoggerAdapter logger) : base(logger)
+            {
+
+            }
+            public override bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
+            {
+                // WAM does not work on pure ADFS environments
```
bgavrilMS, line 106: nit: this is ios broker mock, not wam

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
bgavrilMS: iOS broker gives MSAL back the RT, because iOS does not allow two apps to talk to each other in the background. So iOS broker is the only broker for which we have the flag `CanBrokerSupportSilentAuth` set to false. 

IMO, I'd prefer to have this logic at runtime rather than at build time, i.e. if (BrokerSupportsSilentAuth) { ... } but I understand that we may not want to create new broker just for this. Perhaps this flag is better to be set in PlatformProxy and not in Broker ? Thoughts @trwalke ?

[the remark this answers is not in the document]
SameerK-MSFT: I support runtime check i.e. CanBrokerSupportSilentAuth rather than compile time check, if things change in the future regarding which OS supports the silent broker.

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
```diff
@@ -441,39 +441,6 @@ public void CreateFromiOSBroker_HandlesNoUpn()
             Assert.AreEqual(null, token.Upn);
         }
 
-        internal class IosBrokerMock : NullBroker
```
SameerK-MSFT, line 444: This may be needed, if we do runtime check instead of compile time check.

SameerK-MSFT (changes requested):

tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
trwalke: It is still there. Just in a new location (TestBase.cs)

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -48,59 +48,47 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
-                _logger.Verbose("Attempting to acquire token using using local cache...");
-                return await _clientStrategy.ExecuteAsync(cancellationToken).ConfigureAwait(false);
-
-            }
-            catch (MsalException ex)
-            {
-                _logger.Verbose($"Token cache could not satisfy silent request.");
-
-                if (isBrokerConfigured && ShouldTryWithBrokerError(ex.ErrorCode))
+                if (isBrokerConfigured)
```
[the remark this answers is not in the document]
trwalke: The `isBrokerConfigured` value does not know if the broker is installed but the, the `BrokerSilentStrategy` will check and return null if it is not available. The code will handle this scenario by throwing an exception.

bgavrilMS:
@SameerK-MSFT @trwalke - please aim to close this
@pmaytak - can you please review it as well?

@trwalke - from my point of view this looks good. I am just not sure if we don't break the iOS scenario. If you are able to manually test iOS, I'm signing off.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -48,59 +48,47 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
-            if (isBrokerConfigured && AuthenticationRequestParameters.PopAuthenticationConfiguration != null)
-            {
-                _logger.Info("[Silent Request] Attempting to use broker instead of searching local cache for Proof-of-Possession tokens. ");
-
-                return await _brokerStrategyLazy.Value.ExecuteAsync(cancellationToken).ConfigureAwait(false);
-            }
-
             try
             {
                 if (AuthenticationRequestParameters.Account == null)
```
neha-bhargava, line 53: Do we need this check? I guess it is for silent call and we removed that call.

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -48,59 +48,47 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+                        throw new MsalUiRequiredException(
+                           MsalError.FailedToAcquireTokenSilentlyFromBroker,
+                           "Broker could not satisfy the silent request.",
+                           null,
+                           UiRequiredExceptionClassification.AcquireTokenSilentFailed);
+                    }
                 }
 
-                throw;
+                _logger.Verbose("Attempting to acquire token using using local cache...");
```
pmaytak, line 84:
```suggestion
                _logger.Verbose("Attempting to acquire token using local cache.");

```

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -273,6 +273,20 @@ public sealed partial class TokenCache : ITokenCacheInternal
...
+            if (msalAccessTokenCacheItem != null)
+#else
+            if (msalAccessTokenCacheItem != null && tokenSource != TokenSource.Broker)
+#endif
+            {
+                return true;
+            }
+
+            return false;
```
pmaytak, line 287: nit:
```suggestion
#if iOS
            return msalAccessTokenCacheItem != null;
#else
            return msalAccessTokenCacheItem != null && tokenSource != TokenSource.Broker;
#endif
```

pmaytak (approved):

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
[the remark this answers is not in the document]
trwalke: this flag is already set in PlatformProxy so we are good

trwalke:
> @SameerK-MSFT @trwalke - please aim to close this @pmaytak - can you please review it as well?
> 
> @trwalke - from my point of view this looks good. I am just not sure if we don't break the iOS scenario. If you are able to manually test iOS, I'm signing off.

PR has been updated. just waiting for PR feedback/approvals

bgavrilMS (approved):
Approved assuming manual testing of iOS scenario is ok.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs
```diff
@@ -856,6 +856,9 @@ public async Task AcquireTokenSilent_EmptyLoginHint_TestAsync()
         /// <returns></returns>
         [TestMethod]
         [DeploymentItem(@"Resources\TokenCacheWithWamId.json")]
+        //https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3859
+        //https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3375
+        [Ignore("This test no longer possible since MSAL doesnt return local tokens in acquireTokenSilent anymore.")]
```
SameerK-MSFT, line 861: If you are ignoring the test, why do you want to keep the code?

SameerK-MSFT (approved):

trwalke:
I am trying to merge this PR in but I am getting an error that does not make sense to me. The error is: 

Test method Microsoft.Identity.Test.Integration.Broker.RuntimeBrokerTests.WamUsernamePasswordRequestAsync_WithPiiAsync threw exception:
Microsoft.Identity.Client.MsalUiRequiredException: WAM Error

Error Code: 0
Error Message: Account has previously been signed out of this application.
Internal Error Code: 540940121

I cant seem to find this error in our SDK so it looks like it is from WAM. However, it does not make sense. My changes are only affecting local MSAL cache so not sure how calling ATS on WAM can generate this error. There is also an identical test that does not use PII that works fine too.

Test: WamUsernamePasswordRequestAsync_WithPiiAsync - fails
Test: WamUsernamePasswordRequestAsync - passes

Any thoughts @bgavrilMS @SameerK-MSFT @gladjohn?

image

gladjohn:
@SameerK-MSFT does this test `WamUsernamePasswordRequestAsync_WithPiiAsync` not run locally?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/PublicClientApplicationTests.cs
bgavrilMS: Agreed, delete it.

PR #3865, opened 2022-12-14
Fix PII messages

### LINKED ISSUE #3788: [Bug] MSAL exceptions should not have PII
Assembly Microsoft.Identity.Client, Version=4.47.1.0,

**Platform**
WPF

* Desktop
    * [ ] Interactive
    * [ ] Username Password

Not sure. The app tries to acquire the token silently and AcquireTokenSilent throws an MsalUiRequiredException.

Other?
The issue is that the exception message accessed through System.Exception.Message.ToStirng() contains PII: the email ID of the user.
e.g. (Message: (Could not find a WAM account for the selected user **[user@hotmail.com]**. Status: AccountNotFound Context: Account with id '(pii)' not found Tag: some-hex-number)).

Some information is scrubbed '(pii)' but not the email id of the user.

b. The app is in production, I haven't upgraded MSAL, but started seeing this issue.

**Repro**

```csharp

var builder = PublicClientApplicationBuilder.Create(s_clientId)
                    .WithBrokerPreview(true)
                    .WithAuthority(authority)
                    .WithDefaultRedirectUri()
                    .WithLogging(OnMsalLoggingInvoked, LogLevel.Warning, enablePiiLogging: false, enableDefaultPlatformLogging: false);

public class ExceptionLogging
{
        public static string GetExceptionMessage(Exception _ex)
        {
            return $"(Message: ({_ex.Message.ToString()}))";
        }
}


try
{
    AuthenticationResult result = await s_clientApp.AcquireTokenSilent(scope, account)
                        .WithCorrelationId(correlationId)
                        .ExecuteAsync();
    ....
}
catch (MsalUiRequiredException ex)
{
    Log(ExceptionLogging.GetExceptionMessage(ex));
}

```

**Expected behavior**
With enablePiiLogging set to false, email ID should be scrubbed in the exception message.

**Actual behavior**
Email ID is not scrubbed in the exception message.

bgavrilMS:
Fixes #3788

Log msg only

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -169,8 +169,10 @@ namespace Microsoft.Identity.Client.Broker
                         }
                         else
                         {
-                            _logger?.Warning(
-                                $"[WamBroker] Could not find a WAM account for the selected user {authenticationRequestParameters.Account.Username}");
+                            _logger?.WarningPii(
+                                $"[WamBroker] Could not find a WAM account for the selected user {authenticationRequestParameters.Account.Username}, error: {readAccountResult.Error}",
+                                $"[WamBroker] Could not find a WAM account for the selected user. Error: {readAccountResult.Error}");
+                            
                             _logger?.Info(
                                 $"[WamBroker] Calling SignInInteractivelyAsync this will show the account picker.");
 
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -290,12 +292,12 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
                     if (!readAccountResult.IsSuccess)
                     {
                         _logger?.WarningPii(
-                            $"[WamBroker] Could not find a WAM account for the selected user {acquireTokenSilentParameters.Account.Username}",
-                            $"[WamBroker] Could not find a WAM account for the selected user {readAccountResult.Error}");
+                            $"[WamBroker] Could not find a WAM account for the selected user {acquireTokenSilentParameters.Account.Username} error: {readAccountResult.Error}",
```
SameerK-MSFT, line 295: nit: E capital in "error:" to be consistent with other messages.

SameerK-MSFT (approved):

pmaytak (approved):

neha-bhargava (approved):

trwalke (approved):

PR #3875, opened 2022-12-19
Disable runtime logging from new preview broker

gladjohn:
Fixes #

- disabling runtime logger by not passing the log callback delegate from broker
- Verified that logger is not initialized after removing the code using the Dev App (NetDesktopWinFormsWAM)
- Minimal amount of code removal so it will be easier to plugin the feature for the next release
- updated the failing tests (WamUsernamePasswordRequestWithPOPAsync/WamUsernamePasswordRequestAsync) by commenting it. 

**Testing**
NetDesktopWinFormsWAM Dev App 

**Performance impact**
should improve perf for new broker as we are not logging broker messages anymore

**Documentation**


neha-bhargava (approved):

pmaytak (approved):
Tests will have to be updated too.

PR #3881, opened 2022-12-23
Create a separate build pipeline for netonly targets.

### LINKED ISSUE #3863: [Feature Request] Create a separate MSAL package without net6-windows target
A workaround for #3236. It will allow the Azure Function app to pull net6 binaries instead of net6-windows which don't work because of missing Desktop runtime dependencies.

- [x] Investigate which targets should be included in the package
- [x] Test with Az Functions repro
- [ ] Update the build pipeline to produce two packages
- [ ] Update release pipeline to push this package to internal feed only

pmaytak:
Fixes #3863 

- Added a `NET_ONLY_INTERNAL_PACKAGE` flag to Msal.csproj to only build net6, net461, netstandard2, netcore2.1 targets.
- Updated to add `-netonly` suffix to the package.
- Created a new pipeline to build this package. Created a separate pipeline so that the final binaries and packages for main release and this one are not in the same location, and we don't accidentally publish them to Nuget.
- Only building the main Msal project.
- Not building/running tests, since those are run on source code in the main pipeline.
- The final binaries and packages are signed.
- Added some duplicate compliance tools, especially binskim that verifies binaries.


build/pipeline-releasebuild-netonly.yaml
```diff
@@ -0,0 +1,122 @@
+
+# Release build for -netonly package should be triggered after the normal release build
```
gladjohn, line 2: Instead of creating another new build pipeline, can `NET_ONLY_INTERNAL_PACKAGE` be passed as a build variable, and we use the existing logic? the csproj changes can remain as is. 

When `NET_ONLY_INTERNAL_PACKAGE` is passed as a build variable then we should not build the client package, we should only build the -netonly_internal package

pmaytak: I tried something like this and our tasks usually build the whole solution. So if we exclude most of the targets, the test projects and devapps won't build. We then would have to add more logic to probably exclude certain tasks, and I thought that would complicate the current pipeline. And even if we build only the main proj; we'd still have to add logic to do that. If there's a way to just add minimal flags to the current pipeline, then maybe that'll work, but I'm not sure.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -34,19 +34,25 @@
...
     <TargetFrameworks Condition="$([MSBuild]::IsOsPlatform('OSX'))">$(TargetFrameworkNetStandard);$(TargetFrameworkNetCore);$(TargetFrameworkIos);$(TargetFrameworkAndroid9);$(TargetFrameworkAndroid10);$(TargetFrameworkNet6Ios);$(TargetFrameworkNet6Android);$(TargetFrameworkNet6Mac);$(TargetFrameworkNet6)</TargetFrameworks>
     <TargetFrameworks Condition="$([MSBuild]::IsOsPlatform('Linux')) Or '$(NetCoreOnly)' !='' ">$(TargetFrameworkNetStandard);$(TargetFrameworkNetCore);$(TargetFrameworkNet6)</TargetFrameworks>
-    <PlatformTarget>AnyCPU</PlatformTarget>
-    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+  
+  <!-- Package for Azure Functions targeting .NET Windows TFMs -->
+  <PropertyGroup Condition="'$(NET_ONLY_INTERNAL_PACKAGE)' != ''">
+    <TargetFrameworks Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(TargetFrameworkNetDesktop461);$(TargetFrameworkNetStandard);$(TargetFrameworkNetCore);$(TargetFrameworkNet6);</TargetFrameworks>
```
bgavrilMS, line 45: Is it worth building just net6?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -34,19 +34,25 @@
     <SupportedOSPlatformVersion Condition="'$(TargetFramework)' == '$(TargetFrameworkNet6Android)'">21.0</SupportedOSPlatformVersion>
   </PropertyGroup>
 
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(NET_ONLY_INTERNAL_PACKAGE)' == ''">
```
bgavrilMS, line 37: Please add a comment here indicating he bug for future reference

bgavrilMS (approved):

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -34,19 +34,25 @@
...
+    <TargetFrameworks Condition="$([MSBuild]::IsOsPlatform('Windows'))">$(TargetFrameworkNetDesktop461);$(TargetFrameworkNetStandard);$(TargetFrameworkNetCore);$(TargetFrameworkNet6);</TargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Label="NuGet and AssemblyInfo metadata">
     <!--This should be passed from the VSTS build-->
     <MsalClientSemVer Condition="'$(MsalClientSemVer)' == ''">4.46.0</MsalClientSemVer>
     <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
-    <Version>$(MsalClientSemVer)</Version>
+    <Version Condition="'$(NET_ONLY_INTERNAL_PACKAGE)' == ''">$(MsalClientSemVer)</Version>
+    <Version Condition="'$(NET_ONLY_INTERNAL_PACKAGE)' != ''">$(MsalClientSemVer)-netonly</Version>
```
pmaytak, line 53: Did it this way, since `-netonly` suffix apparently has to be in the Version tag, otherwise it doesn't get added to some metadata in the binaries.

build/pipeline-releasebuild-netonly.yaml
```diff
@@ -0,0 +1,122 @@
...
+        Pattern: '**\bin\**\Microsoft.Identity.Client.dll'
+
+    - task: VSBuild@1
+      displayName: 'Pack ${{ variables.ProjectFilePath }}'
+      inputs:
+        solution: ${{ variables.ProjectFilePath }}
+        msbuildArgs: '/t:pack /p:nobuild=true /p:IncludeSymbols=true /p:NET_ONLY_INTERNAL_PACKAGE=true'
+        configuration: ${{ variables.BuildConfiguration }}
+
+    - task: CopyFiles@2
```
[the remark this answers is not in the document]
pmaytak: We brainstormed suffixes a bit, I think -internal is a bit too generic. If we release some other internal package, it may be confusing; so -netonly is a bit more specific.

neha-bhargava (approved):

gladjohn (approved):

trwalke (approved):
I understand why it is safer to create a separate build pipeline to produce the artifacts but it is possible to ensure the existing pipeline separates this new package from the existing ones so they are not pushed in accidently. Maybe something we can look at later if we find that we have too many builds which tends to happen over time. Ill add this to the pipeline dashboard.

PR #3888, opened 2023-01-06
Fix WAM exception thrown during sign out for the new preview broker 

### LINKED ISSUE #3685: [Bug] ApiContractViolation exception with WithBrokerPreview() during logout
**Logs and network traces**
```
GlobalExceptionHandler: Status: ApiContractViolation
Context: (pii)
Tag: 0x2039c1cb 
GlobalExceptionHandler: Status: ApiContractViolation
Context: (pii)
Tag: 0x2039c1cb
```

Microsoft.Identity.Client 4.47.0
Microsoft.Identity.Client.Broker 4.47.0-preview

**Platform**
.NET 6 WPF

* Desktop / Mobile
    * [x] Interactive
    * [ ] Username Password
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 

This is a new app or experiment.

**Repro**
```csharp
using Microsoft.Graph;
using Microsoft.Identity.Client;
using Microsoft.Identity.Client.Broker;
using System;
using System.Diagnostics;
using System.Linq;
using System.Net.Http.Headers;
using System.Runtime.InteropServices;
using System.Threading.Tasks;
using System.Windows;
using System.Windows.Interop;
using WinCopies.Linq;

namespace mcpat
{
    public class GraphApi
    {
        private GraphServiceClient? graphClient;
        readonly string tenantId = "common";
        readonly string clientId = "xxxxxxxxxxxxxxxxxx";
        readonly string[] scopes = new[] { "User.Read" };
        private readonly IPublicClientApplication? _clientApp;
        private IPublicClientApplication? PublicClientApp { get { return _clientApp; } }
        public GraphApi()
        {
            try
            {
                _clientApp = PublicClientApplicationBuilder.Create(clientId)
                    .WithAuthority(AzureCloudInstance.AzurePublic, tenantId)
                    .WithDefaultRedirectUri()
                    .WithBrokerPreview()
                    .Build();

                TokenCacheHelper.EnableSerialization(_clientApp.UserTokenCache);
            }
            catch (Exception ex)
            {
                Trace.WriteLine("Error GraphApi: " + ex.Message);
            }
        }
        public async Task SignOut()
        {
            try
            {
                if (PublicClientApp == null)
                    return;
                IEnumerable<IAccount> accounts = await PublicClientApp.GetAccountsAsync();

                if (accounts.Any())
                {
                    try
                    {
                        await PublicClientApp.RemoveAsync(accounts.FirstOrDefault());
                        Trace.WriteLine("User has signed-out");
                    }
                    catch (MsalException ex)
                    {
                        Trace.WriteLine($"MsalException Error signing-out user: {ex.Message}");
                    }
                    catch (Exception ex)
                    {
                        Trace.WriteLine($"Exception Error signing-out user: {ex.Message}");
                    }
                }
            } catch (Exception ex)
            {
                Trace.WriteLine($"Error signing-out user: {ex.Message}");
            }  
        }
        public async Task<AuthenticationResult?> GetTokenForUserAsync(Window window)
        {
            try
            {
                IAccount? firstAccount;
                var app = PublicClientApp;
                if (app == null)
                    return null;
                var accounts = await app.GetAccountsAsync();
                firstAccount = accounts.FirstOrDefault();
                AuthenticationResult? authResult = null;

                if (firstAccount != null)
                    authResult = await app.AcquireTokenSilent(scopes, firstAccount).ExecuteAsync();
                else
                    authResult = await app.AcquireTokenInteractive(scopes)
                        .WithParentActivityOrWindow(new WindowInteropHelper(window).Handle)
                        .WithUseEmbeddedWebView(true)
                        .ExecuteAsync();

                return authResult;
            }
            catch (MsalClientException ex)
            {
                int errorCode = Marshal.GetHRForException(ex) & ((1 << 16) - 1);
                Trace.WriteLine("MsalClientException (ErrCode " + errorCode + "): " + ex.ErrorCode);
            }
            catch (Exception ex)
            {
                int errorCode = Marshal.GetHRForException(ex) & ((1 << 16) - 1);
                Trace.WriteLine("Error Acquiring Token (ErrCode " + errorCode + "): " + ex);
            }

            return null;
        }
        public async Task LoginAsync(Window window)
        {
            try
            {
                if (_clientApp == null)
                    return;
                AuthenticationResult? token = await GetTokenForUserAsync(window);
                if (token == null)
                    return;
                
                DelegateAuthenticationProvider authenticationProvider = new((request) =>
                {
                    request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token.AccessToken);
                    return Task.FromResult(0);
                });

                graphClient = new GraphServiceClient(authenticationProvider);

            } catch(Exception ex)
            {
                Trace.WriteLine("Error LoginAsync: " + ex.Message);
            }
        }
    }
}
```

**Expected behavior**
During logout no exception

**Actual behavior**
During logout I receive the exception as written above in the log output

**Possible solution**
?

I receive the exception with this piece of code:
```csharp
//catch exceptions
            AppDomain.CurrentDomain.FirstChanceException += (sender, eventArgs) =>
            {
                Trace.WriteLine("GlobalExceptionHandler: " + eventArgs.Exception.Message + " " + eventArgs.Exception.InnerException);
            };
```

gladjohn:
Fixes #3685 

- Was wrongly passing SignOutResult object to ThrowExceptionFromWamError(), the exception handler only accepts authResult object
- Fixes the APIContractViolation error during signout 

**Testing**
Dev App

**Performance impact**
None

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -412,7 +412,7 @@ namespace Microsoft.Identity.Client.Broker
                     }
                     else
                     {
-                        _logger?.ErrorPii(
+                        _logger?.WarningPii(
                             $"[WamBroker] Could not find a WAM account for the selected user {account.Username} - error: {readAccountResult.Error}",
                             $"[WamBroker] Could not find a WAM account for the selected user, error: {readAccountResult.Error}");
                     }
@@ -428,7 +428,7 @@ namespace Microsoft.Identity.Client.Broker
                         }
                         else
                         {
-                            _logger?.ErrorPii(
+                            _logger?.WarningPii(
                             $"[WamBroker] Could not sign out user {account.Username} - error: {result.Error}",
                             $"[WamBroker] Could not sign out user, error: {result.Error}");
                         }
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -412,12 +412,9 @@ namespace Microsoft.Identity.Client.Broker
                     }
                     else
                     {
-                        _logger?.WarningPii(
+                        _logger?.ErrorPii(
                             $"[WamBroker] Could not find a WAM account for the selected user {account.Username} - error: {readAccountResult.Error}",
                             $"[WamBroker] Could not find a WAM account for the selected user, error: {readAccountResult.Error}");
-
-                        string errorMessage = $"{readAccountResult.Error} (error code : {readAccountResult.Error.ErrorCode})";
-                        throw new MsalServiceException("wam_no_account_found", errorMessage);
                     }
 
                     using (NativeInterop.SignOutResult result = await s_lazyCore.Value.SignOutSilentlyAsync(
@@ -431,12 +428,9 @@ namespace Microsoft.Identity.Client.Broker
                         }
                         else
                         {
-                            _logger?.WarningPii(
+                            _logger?.ErrorPii(
                             $"[WamBroker] Could not sign out user {account.Username} - error: {result.Error}",
                             $"[WamBroker] Could not sign out user, error: {result.Error}");
-
-                            string errorMessage = $"{result.Error} (error code : {result.Error.ErrorCode})";
-                            throw new MsalServiceException("wam_sign_out_failed", errorMessage);
                         }
                     }
                 }
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -431,7 +431,12 @@ namespace Microsoft.Identity.Client.Broker
                         }
                         else
                         {
-                            WamAdapters.ThrowExceptionFromWamError(result, _logger);
+                            _logger?.WarningPii(
+                            $"[WamBroker] Could not sign out user {account.Username} - error: {result.Error}",
+                            $"[WamBroker] Could not sign out user, error: {result.Error}");
+
+                            string errorMessage = $"{result.Error} (error code : {result.Error.ErrorCode})";
+                            throw new MsalServiceException("wam_sign_out_failed", errorMessage);
                         }
                     }
                 }


src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -431,7 +431,12 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
                         }
                         else
                         {
-                            WamAdapters.ThrowExceptionFromWamError(result, _logger);
+                            _logger?.WarningPii(
+                            $"[WamBroker] Could not sign out user {account.Username} - error: {result.Error}",
+                            $"[WamBroker] Could not sign out user, error: {result.Error}");
+
+                            string errorMessage = $"{result.Error} (error code : {result.Error.ErrorCode})";
+                            throw new MsalServiceException("wam_sign_out_failed", errorMessage);
```
bgavrilMS, line 439: IMO, the SingOut operation is best effort. We (MSAL.NET) will delete all artefacts from our side.

And then we call broker to sign out the user as well. There seem to be cases where this won't work, because we don't have the correct account id (we use home_oid, MSAL C++ uses local_oid). 

It might be better to do `try { singout from broker} catch { log exception as error, but continue }`

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -431,7 +431,12 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
...
                         }
```
bgavrilMS, line 440: Also at line 420, I don't think that exception is useful. If an account is not found, let's just move on.

gladjohn:
The actual bug will get fixed with MSAL C++ Runtime fix ([PR](https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3353))

This fix will help with better logging and fixing the bug where wrong params were being passed to `ThrowExceptionFromWamError`

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -412,12 +412,9 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
                     }
                     else
                     {
-                        _logger?.WarningPii(
+                        _logger?.ErrorPii(
```
SameerK-MSFT, line 415: Since we do not care, if the account is not found, it should be just a warning and not error.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -431,7 +428,9 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
                         }
                         else
                         {
-                            WamAdapters.ThrowExceptionFromWamError(result, _logger);
+                            _logger?.ErrorPii(
```
SameerK-MSFT, line 431: Like above, this should be warning and not error.

SameerK-MSFT (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
SameerK-MSFT: Also, my thoughts are on the border of warning/error. Leaning towards warning.

bgavrilMS: Agreed. We do sometimes have customers opening bugs if they discover error messages in their logs. So unless it is truly an exceptional case, we should log this as a warning.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
gladjohn: fixed

Issue #3892, opened 2023-01-09
[Bug] The JSON value could not be converted to System.Int64. ROPC flow


dmitriyi-affinity:
**Logs and network traces**
```
System.Text.Json.JsonException
The JSON value could not be converted to System.Int64. Path: $.expires_in | LineNumber: 0 | BytePositionInLine: 1332.
   at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& state, Utf8JsonReader& reader, Exception ex)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, JsonSerializerOptions options, ReadStack& state)
   at System.Text.Json.Serialization.JsonConverter`1.ReadCoreAsObject(Utf8JsonReader& reader, JsonSerializerOptions options, ReadStack& state)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 utf8Json, JsonTypeInfo jsonTypeInfo, Nullable`1 actualByteCount)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 json, JsonTypeInfo jsonTypeInfo)
   at System.Text.Json.JsonSerializer.Deserialize(String json, Type returnType, JsonSerializerContext context)
   at Microsoft.Identity.Client.Utils.JsonHelper.DeserializeFromJson[T](String json)
   at Microsoft.Identity.Client.OAuth2.OAuth2Client.CreateResponse[T](HttpResponse response, RequestContext requestContext)
   at Microsoft.Identity.Client.OAuth2.OAuth2Client.ExecuteRequestAsync[T](Uri endPoint, HttpMethod method, RequestContext requestContext, Boolean expectErrorsOn200OK, Boolean addCommonHeaders, Func`2 onBeforePostRequestData)
   at Microsoft.Identity.Client.OAuth2.OAuth2Client.GetTokenAsync(Uri endPoint, RequestContext requestContext, Boolean addCommonHeaders, Func`2 onBeforePostRequestHandler)
   at Microsoft.Identity.Client.OAuth2.TokenClient.SendHttpAndClearTelemetryAsync(String tokenEndpoint, ILoggerAdapter logger)
[...]

PR #3902, opened 2023-01-13
Fix for System.Text.Json on .NET 7 when deserializing ints represented as strings

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3892

adding `[JsonNumberHandling(JsonNumberHandling.AllowReadingFromString)]` attribute to values that cannot convert string to long during serialization. 

**Testing**
Manual testing (No .NET 7 target yet)

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
```diff
@@ -31,14 +32,19 @@ internal static string SerializeToJson<T>(T toEncode)
 #endif
         }
 
+#if SUPPORTS_SYSTEM_TEXT_JSON
+        [RequiresUnreferencedCode("Calls System.Text.Json.JsonSerializer.Deserialize(String, Type, JsonSerializerOptions)")]
```
trwalke, line 36: This is needed when using JsonSerializer.Deserialize(json, typeof(T), MsalJsonSerializerOptions.Options to avoid broken functionality when using trimming.

Message:
Warning	IL2026	Using member 'System.Text.Json.JsonSerializer.Deserialize(String, Type, JsonSerializerOptions)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. JSON serialization and deserialization might require types that cannot be statically analyzed. Use the overload that takes a JsonTypeInfo or JsonSerializerContext, or make sure all of the required types are preserved.

trwalke: Im going to keep trying different things to see if I can find a better solution but at least this one works.

trwalke:
There is no .NET 7 target for MSAL yet. @SameerK-MSFT do you plan to add this target? If so, I can use that to create a unit test for this fix.

trwalke:
Found a potential solution.
So there are two ways to set serialization/deserialization behavior in `System.Test.Json` (there may be others)
- `JsonSerializerContext` 
- `JsonSerializerOptions`. 

It seems like tergetting .NET 7 causes the 
`object? Deserialize(string json, Type returnType, JsonSerializerContext context)` api to ignore the configuration in the `JsonSerializerContext`. 

However, using `object? Deserialize(string json, Type returnType, JsonSerializerOptions? options = null)` and setting `JsonSerializerOptions` seems to work.

Since this isn't a documented change, perhaps I should open an issue on the .NET 7 repo? I doubt they intended to ignore `JsonSerializerContext` when deserializing.

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
```diff
@@ -31,14 +32,19 @@ internal static string SerializeToJson<T>(T toEncode)
...
         {
             if (string.IsNullOrEmpty(json))
             {
                 return default;
             }
 #if SUPPORTS_SYSTEM_TEXT_JSON
-            return (T)JsonSerializer.Deserialize(json, typeof(T), MsalJsonSerializerContext.Custom);
+            //Workaround for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3892
+            //This also requires the RequiresUnreferencedCode attribute to maintain functionality during trimming.
+            return (T)JsonSerializer.Deserialize(json, typeof(T), MsalJsonSerializerOptions.Options);
```
trwalke, line 47: Found a potential solution.
So there are two ways to set serialization/deserialization behavior in `System.Test.Json` (there may be others)
- `JsonSerializerContext` 
- `JsonSerializerOptions`. 

It seems like tergetting .NET 7 causes the 
`object? Deserialize(string json, Type returnType, JsonSerializerContext context)` api to ignore the configuration in the `JsonSerializerContext`. 

However, using `object? Deserialize(string json, Type returnType, JsonSerializerOptions? options = null)` and setting `JsonSerializerOptions` seems to work.

Since this isn't a documented change, perhaps I should open an issue on the .NET 7 repo? I doubt they intended to ignore `JsonSerializerContext` when deserializing. @pmaytak, @bgavrilMS, @SameerK-MSFT?

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
pmaytak: Seems like Context class is better to use... They do use both Options and Context in the samples. There are different modes of source generation, so maybe we need to set some specific properties for Options to work the same as Context.
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
pmaytak: Yea, I think opening an issue there would be good. You can also search through 7.0.x [milestones](https://github.com/dotnet/runtime/milestones?state=closed) or issues, maybe this was already reported or fixed. Otherwise, getting some clarify would be good.

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
bgavrilMS: Trimming is important, we have customers using it in public client world. 

If we can't find a good solution at JSON level, consider having separate logic for AAD response / B2C response.

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
SameerK-MSFT: Not honoring JsonSerializerContext would be a serious bug. It would be good to check if it already exists.

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
SameerK-MSFT: I agree with Bogdan that if the response format is different for AAD and B2C, it would be good to have separate deserialization logic for each.

SameerK-MSFT (changes requested):

src/client/Microsoft.Identity.Client/Platforms/net6/MsalJsonSerializerOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.Json.Serialization;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Platforms.net6
+{
+    internal class MsalJsonSerializerOptions
```
bgavrilMS, line 14: 1. static class
2. Maybe call it `SystemTextJsonOptions` ?

src/client/Microsoft.Identity.Client/Platforms/net6/MsalJsonSerializerOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        private static JsonSerializerOptions s_serializerOptions;
+
+        public static JsonSerializerOptions Options
+        {
+            get
+            {
+                return s_serializerOptions ??=
```
bgavrilMS, line 22: nice use of ??=

bgavrilMS (changes requested):
We can't break IL trimming

trwalke:
I have reworked this to avoid breaking IL Trimming and a dependency on .Net7. Working on unit testing with .Net 7

pmaytak (commented):
Looks good. Tentatively approving. I think it's still a bug on .NET that they should fix, but we can release with this faster to unblock users.

Also let's make sure [other classes](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/39ff025c091d5860893f32ecd2cec3ef5da2042d/src/client/Microsoft.Identity.Client/Platforms/net6/MsalJsonSerializerContext.cs#L24-L40) with long properties are updated.

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
trwalke: Removed Il trimming attribute and using a better solution.

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
trwalke: Changing solution to use attribute on serializable values

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
```diff
@@ -7,6 +7,7 @@
 using System.Text;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Internal;
+using System.Diagnostics.CodeAnalysis;
```
bgavrilMS, line 10: pls remove

bgavrilMS (changes requested):
Looks like a great solution, but missing a unit test. Would be great to get a E2E test added.

trwalke:
@bgavrilMS added a new net7 target for unit tests. I can you take one more look and confirm if we want to remove the .net 6 target and use net 7 going forward?

bgavrilMS:
> @bgavrilMS added a new net7 target for unit tests. I can you take one more look and confirm if we want to remove the .net 6 target and use net 7 going forward?

Sure, if it's not too much build work, let's upgrade net6 to net7

bgavrilMS (approved):

trwalke:
Seems like the .net 7 target is not available on hosted agents at this time

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
```diff
@@ -15,6 +16,10 @@
     <Platforms>AnyCPU;x64</Platforms>
   </PropertyGroup>
 
+  <!--<ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet7)'">
```
pmaytak, line 19: nit: remove?

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
```diff
@@ -4,8 +4,9 @@
     <TargetFrameworkNetDesktop461>net48</TargetFrameworkNetDesktop461>
     <TargetFrameworkNetCore>netcoreapp3.1</TargetFrameworkNetCore>
     <TargetFrameworkNet6Win>net6.0-windows10.0.17763.0</TargetFrameworkNet6Win>
+    <TargetFrameworkNet7Win>net7.0-windows10.0.18362.0</TargetFrameworkNet7Win>
     
-    <TargetFrameworks>$(TargetFrameworkNetDesktop461);$(TargetFrameworkNetCore);$(TargetFrameworkNet6Win);</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworkNetDesktop461);$(TargetFrameworkNetCore);$(TargetFrameworkNet6Win);$(TargetFrameworkNet7Win)</TargetFrameworks>
```
pmaytak, line 9: Also need to update: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/build/template-run-unit-tests.yaml to run net7 tests.

pmaytak (approved):

src/client/Microsoft.Identity.Client/Internal/DeviceCodeResponse.cs
```diff
@@ -32,9 +33,15 @@ internal class DeviceCodeResponse : OAuth2ResponseBase
         public string VerificationUri { get; set; }
 
         [JsonProperty("expires_in")]
+#if SUPPORTS_SYSTEM_TEXT_JSON
```
bgavrilMS, line 36: Can you do a similar trick to `JsonProperty` so that we can avoid adding these pragmas everywhere?

src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
```diff
@@ -181,9 +181,15 @@ internal class JWTPayload
...
 
             [JsonProperty(JsonWebTokenConstants.ReservedClaims.NotBefore)]
+#if SUPPORTS_SYSTEM_TEXT_JSON
+            [JsonNumberHandling(JsonNumberHandling.AllowReadingFromString)]
+#endif
             public long ValidFrom { get; set; }
 
             [JsonProperty(JsonWebTokenConstants.ReservedClaims.ExpiresOn)]
+#if SUPPORTS_SYSTEM_TEXT_JSON
+            [JsonNumberHandling(JsonNumberHandling.AllowReadingFromString)]
```
bgavrilMS, line 191: Is this needed? This class is used to go from OBJ to JSON, not the other way around?

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
bgavrilMS, line 9: How much time does this add to the build? Is it worth removing: 

- the net6 target?
- the netcore 3.1 target ? 

to mitigate ?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
pmaytak: Counter-point: I would it add it here for the sake of consistency - add the attribute for any integer properties.

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
pmaytak: Unit tests run is about 1.5 minutes. I'm fine with adding additional frameworks, as long as it adds coverage. Net7 broke stuff, but it's also a new platform. net3.1 and net6 should really only get patch changes, so maybe worth replacing at least net6. Netcore3.1 I think we still need, since it tests the Newtonsoft json related code.

More OCE work: maybe we should add test coverage tools or something for each PR run.

src/client/Microsoft.Identity.Client/Internal/DeviceCodeResponse.cs
pmaytak: It would be similar to [JsonObject](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Platforms/net6/JsonObjectAttribute.cs) which is an empty class on net6 platform, but a proper Json.Net class on non-net6 platforms.
So `JsonNumberHandling` would be opposite - it would be a dummy class on non-net6 platforms. I think it's possible but maybe a bit tricky, since you'd have to add that class on every non-net6 platform.

[JsonIgnore](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e774d1d9791ce070355dd3aa17709683636de111/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs#L250) is similar and I just used `#if` flags.

Up to you whichever way looks cleaner.

build/platform_and_feature_flags.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetCore)' or '$(TargetFramework)' == '$(TargetFrameworkNet6)'">
     <DefineConstants>$(DefineConstants);NET_CORE;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_CUSTOM_CACHE;SUPPORTS_BROKER;SUPPORTS_WIN32;</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet6)' or '$(TargetFramework)' == '$(TargetFrameworkNet6Win)'">
+  <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet6)' or '$(TargetFramework)' == '$(TargetFrameworkNet6Win)' or '$(TargetFramework)' == '$(TargetFrameworkNet7)' or '$(TargetFramework)' == '$(TargetFrameworkNet7Win)'">
```
bgavrilMS, line 5: I don't think this is needed. MSAL doesn't explicitly target net7 and our tests don't need to know what JSON parsing lib MSAL is using?

trwalke: This is primarily for the unit tests to use SUPPORTS_SYSTEM_TEXT_JSON as a directive for .net 7. Technically it should be in the unit test project, Ill move it there

PR #3912, opened 2023-01-23
Bypass the Accounts Control when there is an account hint

### LINKED ISSUE #3301: [Bug] WAM broker does not honor LoginHint
MSAL 4.46

1. Use WAM preview
2. Declare that the app is MSA-PT, i.e. only AAD plugin should handle auth
3. AcquireTokenInteractive using a login_hint

**Actual**: the full account picker pops up every time

image

**Expected** - a better interactive login experience, at the very least the Account Picker should not be invoked and instead the AAD plugin's browser should accelerate the authentication, e.g. Note that if account is part of the OS, the AAD plugin seems to be able to give back token silently.

image

Tracking ADO issue: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2040431

Note: when implementing a fix, please see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3520 as well for internal API details

gladjohn:
Fixes #3301

In majority of cases, MSALRuntime will know which identity provider to invoke based on the authority. However, in the consumer pass through scenario where the authority is /organizations or /<tanent_id> but the application accepts MSA, passing the account_hint itself will not able to disambiguate the identity provider. 
Hence, if there is an account_hint in the SignInInteractively, MSAL need to pass this additional key value pair if the identity provider is MSA:
````
{"msal_identity_provider":"msa"}
````
If this key value is missing or the value is not msa, the account_hint will be treated as AAD by default.  
If the account was signed in before and has an account object in MSAL C++ (By calling ReadAccountById), the token acquisition path should be to re-hydrated the account and then call acquireToken* with the account object instead of using SignInInteractively with the account_hint.

[Adding link to the design](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3301#issuecomment-1399941331) 

**Testing**
Dev App with MSA-PT

**Performance impact**
None

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -15,6 +15,7 @@ using Microsoft.Identity.Client.Internal;
 using Microsoft.Identity.Client.Internal.Broker;
 using Microsoft.Identity.Client.Internal.Requests;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 using Microsoft.Identity.Client.Utils;
 using static Microsoft.Identity.Client.Broker.RuntimeBroker;
 
@@ -240,6 +241,10 @@ namespace Microsoft.Identity.Client.Broker
             NativeInterop.AuthParameters authParams,
             ILoggerAdapter logger)
         {
+            //Set MSAL Identity Provider only for AcquireTokenInteractive API
+            if (authenticationRequestParameters.ApiId != ApiEvent.ApiIds.AcquireTokenInteractive)
+                return;
+
             //Set MsalIdentityProvider Based on Tenant ID
             if (authenticationRequestParameters?.Account?.HomeAccountId != null)
             {
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -27,6 +27,8 @@ namespace Microsoft.Identity.Client.Broker
         //MSA-PT Auth Params
         private const string NativeInteropMsalRequestType = "msal_request_type";
         private const string ConsumersPassthroughRequest = "consumer_passthrough";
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by gladjohn** [comment]
still in testing so marking it as draft for now. Will need to check with Sam on a few scerarios

**Comment by bgavrilMS** [comment]
I provided a spec on the GH item https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3301

File: src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs [hunk 1]
```diff
@@ -178,6 +180,7 @@ private enum ResponseStatus
             if (brokerOptions.MsaPassthrough)
```
bgavrilMS, line 180: If it's MSA-PT, then msal_runtime doesn't display the account picker anyway. Please see the original work item, I added a spec there.

gladjohn:
> I provided a spec on the GH item #3301

Thanks @bgavrilMS, that worked. Will do some more testing and update the PR

gladjohn:
tested with dev app and it works as expected in the scenario mentioned in this bug for MSA-PT scenarios. 

pending to be done : test by passing QP from the dev app.

bgavrilMS (approved):

SameerK-MSFT:
Unit test?

SameerK-MSFT (approved):
Please add unit test(s).

neha-bhargava (approved):

bgavrilMS:
> Please add unit test(s).

@neha-bhargava @gladjohn @SameerK-MSFT - when we started the msal_runtime effort we specifically discussed about unit testing and we decided to not add them, due to low ROI. We rely exclusively on integration testing and manual testing. Happy to discuss about this.

gladjohn:
> > Please add unit test(s).
> 
> @neha-bhargava @gladjohn @SameerK-MSFT - when we started the msal_runtime effort we specifically discussed about unit testing and we decided to not add them, due to low ROI. We rely exclusively on integration testing and manual testing. Happy to discuss about this.

added a test case to the release sheet for manual verification

PR #3918, opened 2023-01-26
[WAM] Let MSALRuntime handle the no scope validation

### LINKED ISSUE #3675: [Request] WAM support for no scopes

gladjohn:
Fixes #3675 

- Remove no scope check from MSAL .NET code 
- MSAL Runtime handles the no scope validation, catch the exception and throw a MSALServiceException 

**Testing**
Dev App

**Performance impact**
None

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -15,6 +15,7 @@ using Microsoft.Identity.Client.Internal;
 using Microsoft.Identity.Client.Internal.Broker;
 using Microsoft.Identity.Client.Internal.Requests;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 using Microsoft.Identity.Client.Utils;
 using static Microsoft.Identity.Client.Broker.RuntimeBroker;
 
@@ -27,6 +28,9 @@ namespace Microsoft.Identity.Client.Broker
         //MSA-PT Auth Params
         private const string NativeInteropMsalRequestType = "msal_request_type";
         private const string ConsumersPassthroughRequest = "consumer_passthrough";
+        private const string MsalIdentityProvider = "msal_identity_provider";
+        private const string IdentityProviderTypeMSA = "msa";
+        private const string IdentityProviderTypeAAD = "aad";
         private const string WamHeaderTitle = "msal_accounts_control_title";
 
         //MSAL Runtime Error Response 
@@ -175,6 +179,9 @@ namespace Microsoft.Identity.Client.Broker
             //this is used internally by the interop to fallback to the browser 
             authParams.RedirectUri = authenticationRequestParameters.RedirectUri.ToString();
 
+            //MSAL Identity Provider
+            SetMSALIdentityProvider(authenticationRequestParameters, authParams, logger);
+
             //MSA-PT
             if (brokerOptions.MsaPassthrough)
             {
@@ -227,6 +234,40 @@ namespace Microsoft.Identity.Client.Broker
                 authParams.PopParams.UriHost = authenticationRequestParameters.PopAuthenticationConfiguration.HttpHost;
                 authParams.PopParams.UriPath = authenticationRequestParameters.PopAuthenticationConfiguration.HttpPath;
                 authParams.PopParams.Nonce = authenticationRequestParameters.PopAuthenticationConfiguration.Nonce;
+            }
+        }
+
+        private static void SetMSALIdentityProvider(
+            AuthenticationRequestParameters authenticationRequestParameters, 
+            NativeInterop.AuthParameters authParams,
+            ILoggerAdapter logger)
+        {
+            //Set MSAL Identity Provider only for AcquireTokenInteractive API
+            if (authenticationRequestParameters.ApiId != ApiEvent.ApiIds.AcquireTokenInteractive)
+                return;
+
+            //Set MsalIdentityProvider Based on Tenant ID
+            if (authenticationRequestParameters?.Account?.HomeAccountId != null)
+            {
+                if (!string.IsNullOrEmpty(authenticationRequestParameters.Account.HomeAccountId.TenantId))
+                {
+                    var tenantObjectId = authenticationRequestParameters.Account.HomeAccountId.TenantId;
+
+                    if (tenantObjectId.Equals(Constants.MsaTenantId, StringComparison.OrdinalIgnoreCase))
+                    {
+                        logger.Verbose($"[WamBroker] MSALRuntime Identity provider set to " +
+                            $"{ IdentityProviderTypeMSA }.");
+
+                        authParams.Properties[MsalIdentityProvider] = IdentityProviderTypeMSA;
+                    }
+                    else
+                    {
+                        logger.Verbose($"[WamBroker] MSALRuntime Identity provider set to " +
+                            $"{ IdentityProviderTypeAAD }.");
+
+                        authParams.Properties[MsalIdentityProvider] = IdentityProviderTypeAAD;
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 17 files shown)


pmaytak:
What is the reasoning for moving this check to MsalRuntime layer?

I think having it in MSAL is better - the error is more specific. Otherwise, MsalRuntime will throw a generic ApiContractViolation exception and the user will have to figure out what's wrong. Also I think it should stay a client exception, not service exception.

Can you remind me, what was the end result of the conversation about why MsalRuntime won't support empty (or oidc only) scopes?

gladjohn:
> What is the reasoning for moving this check to MsalRuntime layer?
> 
> I think having it in MSAL is better - the error is more specific. Otherwise, MsalRuntime will throw a generic ApiContractViolation exception and the user will have to figure out what's wrong. Also I think it should stay a client exception, not service exception.
> 
> Can you remind me, what was the end result of the conversation about why MsalRuntime won't support empty (or oidc only) scopes?

Scopes are a required parameter. We need to pass scopes for MSAL CPP. This check was recently added on the C++ layer. Also there was a bug when we used magic scopes, it will let you sing in but error out during sign out. Now that has been fixed. If we decide to add the no scope check in MSAL .Net, we just need to check for empty / null scopes. The magic scope check we used to do needs to be removed as that scenario works now.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -239,6 +241,36 @@ public async Task WamListWindowsWorkAndSchoolAccountsAsync()
             var account = accounts.FirstOrDefault();
             Assert.AreEqual(labResponse.User.Upn, account.Username);
         }
+
+        [DataTestMethod]
+        [DataRow("")]
+        [DataRow(" ")]
+        [DataRow(null)]
```
bgavrilMS, line 248: Too expensive for integration test, pick one :)

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -239,6 +241,36 @@ public async Task WamListWindowsWorkAndSchoolAccountsAsync()
...
+            Func<IntPtr> windowHandleProvider = () => intPtr;
+
+            IPublicClientApplication pca = PublicClientApplicationBuilder
+               .Create("04f0c124-f2bc-4f59-8241-bf6df9866bbd")
+               .WithAuthority("https://login.microsoftonline.com/organizations")
+               .WithParentActivityOrWindow(windowHandleProvider)
+               .WithBrokerPreview()
+               .Build();
+
+            var ex = await AssertException.TaskThrowsAsync<MsalServiceException>(
```
bgavrilMS, line 266: I was under the impression that MSAL C++ now allows no scopes?

gladjohn: Per discussion I had with @MSamWils and @logdog16, scopes is a required parameter for MSAL C++. But the magic scopes can be passed and token retrieved. There was a bug with magic scopes and signout but that is now fixed. @MSamWils can you please confirm the behavior

MSamWils: That's correct, passing the magic scope should works fine including sign out and sign back in.

pmaytak: I thought if the user passes empty scopes, we add [default/reserved](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/80c7824871bd6369e740479ba0497850bfbe51de/src/client/Microsoft.Identity.Client/OAuth2/OAuthConstants.cs#L101) scopes, like [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/80c7824871bd6369e740479ba0497850bfbe51de/src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs#L62). Seems like for old WAM, we check for null only, not empty, see [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/80c7824871bd6369e740479ba0497850bfbe51de/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AadPlugin.cs#L266).

So for MSAL runtime, we should pass default scopes, if user scopes are null/empty. Yes?

pmaytak: I thought if the user passes empty scopes, we add [default/reserved](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/80c7824871bd6369e740479ba0497850bfbe51de/src/client/Microsoft.Identity.Client/OAuth2/OAuthConstants.cs#L101) scopes, like [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/80c7824871bd6369e740479ba0497850bfbe51de/src/client/Microsoft.Identity.Client/OAuth2/TokenClient.cs#L62). Seems like for old WAM, we check for null only, not empty, see [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/80c7824871bd6369e740479ba0497850bfbe51de/src/client/Microsoft.Identity.Client/Platforms/Features/WamBroker/AadPlugin.cs#L266).

So for MSAL runtime, we should pass default scopes, if user scopes are null/empty. Yes?

pmaytak (changes requested):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
gladjohn: sounds good, let me fix it

gladjohn: @pmaytak please review

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -165,14 +165,23 @@ private enum ResponseStatus
...
             var authParams = new NativeInterop.AuthParameters
                 (authenticationRequestParameters.AppConfig.ClientId,
                 authenticationRequestParameters.Authority.AuthorityInfo.CanonicalAuthority.ToString());
 
-            //scopes
-            authParams.RequestedScopes = string.Join(" ", authenticationRequestParameters.Scope);
+            //When no scopes are passed, add the default scopes
+            if (!ScopeHelper.HasNonMsalScopes(authenticationRequestParameters.Scope))
+            {
+                string defaultScopes = "email offline_access openid profile";
```
pmaytak, line 176: We don't send email scope. Maybe create a `const string = "offline_access openid profile"` in OAuthConstants.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/a927f8c716677893c116333d6d8d2d12618f102b/src/client/Microsoft.Identity.Client/OAuth2/OAuthConstants.cs#L97-L99

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -239,6 +241,43 @@ public async Task WamListWindowsWorkAndSchoolAccountsAsync()
...
+
+            }
+            catch (MsalUiRequiredException ex)
+            {
+                Assert.IsTrue(!string.IsNullOrEmpty(ex.ErrorCode));
+            }
+            catch (MsalServiceException)
+            {
+                Assert.Fail("MsalServiceException should not have been thrown.");
+            }
```
pmaytak, line 279:
```suggestion
            var ex = await AssertException.TaskThrowsAsync<MsalServiceException>(
                 () => pca.AcquireTokenSilent(new string[] { scopes }, PublicClientApplication.OperatingSystemAccount)
                        .ExecuteAsync())
                        .ConfigureAwait(false);

            Assert.IsFalse(string.IsNullOrEmpty(ex.ErrorCode));
```

pmaytak (approved):

pmaytak (approved):

PR #3919, opened 2023-01-26
Audit all info and verbose logs and use lambdas to generate strings

### LINKED ISSUE #3901: [Bug] Logs are created even if logging is disabled
**Issue**
See [AuthenticationRequestParameters](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs#L166). The log string is created even if the specific log level is disabled. This method is also called on every request. Other parameter classes have the same behavior; and there may be other cases. Usually users use log level Warning, so logs are generated which are not used.

**Solution**
Go over any logger calls where we call Log (especially the perf sensitive areas) and wrap log creation into a conditional to check if the log level is enabled (`if (logger.IsLoggingEnabled(LogLevel.X))`). Verify the results with perf tests.

bgavrilMS:
Fixes #3901 

- add logging methods that use lambdas, which do not execute unless the log level is met (this is inspired by high-performance logging from https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-7.0)
- for verbose logging, i removed the overloads with strings. Use only Func<string>
- for info it's a mix bag - strings and Func<string>, since info is ON by default and the Func<string> introduces an extra check.

**Testing** 
unit

**Performance impact**
|                                                Method |        Mean |     Error |    StdDev |      Median |         Min |         Max |   Gen0 | Allocated |
|------------------------------------------------------ |------------:|----------:|----------:|------------:|------------:|------------:|-------:|----------:|
|  'Compute and log expensive string via direct string' | 1,292.38 ns | 35.423 ns | 33.135 ns | 1,287.09 ns | 1,253.46 ns | 1,371.78 ns | 0.1488 |     944 B |
|           'Compute and log expensive string via Func' | 1,369.27 ns | 41.625 ns | 34.759 ns | 1,362.51 ns | 1,296.23 ns | 1,436.84 ns | 0.1678 |    1056 B |
|           'Compute and skip logging expensive string' |   628.37 ns | 10.844 ns | 10.143 ns |   626.69 ns |   611.36 ns |   650.44 ns | 0.0591 |     376 B |
| 'No compute (Func) and skip logging expensive string' |    14.14 ns |  0.286 ns |  0.239 ns |    14.14 ns |    13.72 ns |    14.66 ns | 0.0153 |      96 B |

** AcquireTokenForClient perf tests **

These were modified to run with `WithLogging(Console.WriteLine(msg), LogLevel.Info)`

|                Method |   CacheSize | EnableCacheSerialization |      Mean |       Min |       Max |      Gen0 |      Gen1 |     Gen2 |   Allocated |
|---------------------- |------------ |------------------------- |----------:|----------:|----------:|----------:|----------:|---------:|------------:|
| AcquireTokenForClient |     (1, 10) |                    False |  3.114 ms |  3.057 ms |  3.204 ms |    7.8125 |         - |        - |    51.26 KB |
| AcquireTokenForClient |     (1, 10) |                     True |  3.204 ms |  3.163 ms |  3.280 ms |   46.8750 |   15.6250 |        - |   287.65 KB |
| AcquireTokenForClient |   (1, 1000) |                    False |  3.121 ms |  3.059 ms |  3.160 ms |   27.3438 |         - |        - |   190.96 KB |
| AcquireTokenForClient |   (1, 1000) |                     True | 30.336 ms | 29.677 ms | 31.857 ms | 2687.5000 | 1312.5000 | 625.0000 | 17010.89 KB |
| AcquireTokenForClient | (10000, 10) |                    False |  3.116 ms |  3.030 ms |  3.184 ms |    7.8125 |         - |        - |    51.28 KB |
| AcquireTokenForClient | (10000, 10) |                     True |  4.117 ms |  3.383 ms |  4.852 ms |   46.8750 |   15.6250 |        - |   287.65 KB 

|                Method |   CacheSize | EnableCacheSerialization |      Mean |       Min |       Max |      Gen0 |      Gen1 |     Gen2 |   Allocated |
|---------------------- |------------ |------------------------- |----------:|----------:|----------:|----------:|----------:|---------:|------------:|
| AcquireTokenForClient |     (1, 10) |                    False |  2.990 ms |  2.907 ms |  3.063 ms |    7.8125 |         - |        - |    50.23 KB |
| AcquireTokenForClient |     (1, 10) |                     True |  3.092 ms |  3.024 ms |  3.180 ms |   42.9688 |   11.7188 |        - |   287.06 KB |
| AcquireTokenForClient |   (1, 1000) |                    False |  3.012 ms |  2.932 ms |  3.064 ms |   23.4375 |         - |        - |   158.98 KB |
| AcquireTokenForClient |   (1, 1000) |                     True | 29.365 ms | 27.786 ms | 31.487 ms | 2625.0000 | 1250.0000 | 625.0000 | 17065.87 KB |
| AcquireTokenForClient | (10000, 10) |                    False |  2.966 ms |  2.916 ms |  3.019 ms |    7.8125 |         - |        - |    50.24 KB |
| AcquireTokenForClient | (10000, 10) |                     True |  3.080 ms |  3.010 ms |  3.147 ms |   42.9688 |   11.7188 |        - |   287.03 KB |

**Documentation**
n/a

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
@@ -108,7 +108,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
                 }
                 if (entry == null)
                 {
-                    requestContext.Logger.Info($"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not enabled.");
+                    requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not enabled.");
                     entry = CreateEntryForSingleAuthority(authorityInfo.CanonicalAuthority);
                 }
 
@@ -116,7 +116,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
             }
             else
             {
-                requestContext.Logger.Info($"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not supported.");
+                requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not supported.");
                 return await GetMetadataEntryAsync(authorityInfo, requestContext).ConfigureAwait(false);
             }
         }
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
@@ -108,7 +108,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
                 }
                 if (entry == null)
                 {
-                    requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not enabled.");
+                    requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because it is not enabled.");
                     entry = CreateEntryForSingleAuthority(authorityInfo.CanonicalAuthority);
                 }
 
@@ -116,7 +116,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
             }
             else
             {
-                requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not supported.");
+                requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because it is not supported.");
                 return await GetMetadataEntryAsync(authorityInfo, requestContext).ConfigureAwait(false);
             }
         }
--- a/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
@@ -414,7 +414,7 @@ namespace Microsoft.Identity.Client.Broker
                 {
                     if (readAccountResult.IsSuccess)
                     {
-                        _logger?.Verbose(() => "[WamBroker] WAM Account exist and can be removed.");
+                        _logger?.Verbose(() => "[WamBroker] WAM Account exists and can be removed.");
 
                     }
                     else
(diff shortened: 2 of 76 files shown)


src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -108,15 +108,15 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
                 }
                 if (entry == null)
                 {
-                    requestContext.Logger.Info($"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not enabled.");
+                    requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not enabled.");
```
gladjohn, line 111:
```suggestion
                    requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because it is not enabled.");
```

src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
```diff
@@ -108,15 +108,15 @@ internal class InstanceDiscoveryManager : IInstanceDiscoveryManager
...
                     entry = CreateEntryForSingleAuthority(authorityInfo.CanonicalAuthority);
                 }
 
                 return entry;
             }
             else
             {
-                requestContext.Logger.Info($"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not supported.");
+                requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because is not supported.");
```
gladjohn, line 119:
```suggestion
                requestContext.Logger.Info(() => $"Skipping Instance discovery for {authorityInfo.AuthorityType} authority because it is not supported.");
```

gladjohn (approved):

pmaytak (approved):
Cool stuff. Based on perf tests, seems like it does help to not compute when logging is disabled. Would be nice though if there was a syntactic sugar to not add `() =>` part - parameter is Func, you pass in a string, but compiler figures out to change it to a Func.

PR #3932, opened 2023-02-02
Drop support for Mono.Android 9

bgavrilMS:
As per https://dotnet.microsoft.com/en-us/platform/support/policy/xamarin

Fixes #

**Testing**

**Performance impact**

**Documentation**


SameerK-MSFT:
I would remove just MonoAndroid9.0 and TargetFrameworkAndroid9 and keep StartsWith as is. if we decide to add a version in future, this will becom future proof.
Also, are changes in "AndroidHttpClientFactory.cs" just tab changes? Ok, if they are intentional.

SameerK-MSFT (approved):

neha-bhargava (approved):

pmaytak (approved):
I think it will also be helpful to add the link to that Xamarin lifecycle doc to our Xamarin wiki and any Xamarin samples, with a suggestion to use Maui (unless we already do that).

pmaytak:
> I would remove just MonoAndroid9.0 and TargetFrameworkAndroid9 and keep StartsWith as is. if we decide to add a version in future, this will becom future proof. Also, are changes in "AndroidHttpClientFactory.cs" just tab changes? Ok, if they are intentional.

If you're reviewing via browser, you can hide whitespace:
image

bgavrilMS:
> I would remove just MonoAndroid9.0 and TargetFrameworkAndroid9 and keep StartsWith as is. if we decide to add a version in future, this will becom future proof. Also, are changes in "AndroidHttpClientFactory.cs" just tab changes? Ok, if they are intentional.

In the file yes, just tab changes. I'm guessing it's VS for Mac not applying the rule we defined in .editorconfig 
For the `StartsWith` I don't see the value, as we're never going to add support for a higher level of MonoAndroid. Maui defines Android as netX-android. "monoAndroid" is a xamarin concept and it's obsolete.

PR #3937, opened 2023-02-03
Handle exception for the scenario.

### LINKED ISSUE #3883: [Help]  I am using the method AcquireTokenByUsernamePassword to get the token, but I get the following error. I haven't configured it properly. thank you
I am using the method AcquireTokenByUsernamePassword to get the token, but I get the following error. I haven't configured it properly. thank you

Microsoft.Identity.Client：Version="4.49.1"

code：
 public readonly static string[] ApiScopes = { "openid", "profile", "offline_access", "User.ManageIdentities.All", "User.Read.All","User.ReadBasic.All","User.Invite.All","User.ReadWrite.All","IdentityProvider.Read.All" };

 app = PublicClientApplicationBuilder.Create(ClientId)
                .WithB2CAuthority(AuthoritySignUpSignIn)
                .WithRedirectUri("http://localhost:8080/home")
                .WithExtraQueryParameters(new Dictionary<string, string>() { { "nonce", Guid.NewGuid().ToString() } })
                .WithLogging(Log, LogLevel.Info, false) 
                .Build();

 authResult = await app.AcquireTokenByUsernamePassword(App.ApiScopes, "UserName", "UserPassword").ExecuteAsync();

Error Message：
Error Acquiring Token:
Microsoft.Identity.Json.JsonReaderException: Unexpected character encountered while parsing value: <. Path '', line 0, position 0.
   at Microsoft.Identity.Json.JsonTextReader.ParseValue()
   at Microsoft.Identity.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
   at Microsoft.Identity.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Microsoft.Identity.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Microsoft.Identity.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)
   at Microsoft.Identity.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)
   at Microsoft.Identity.Client.Utils.JsonHelper.DeserializeFromJson[T](String json)
   at Microsoft.Identity.Client.OAuth2.OAuth2Client.CreateResponse[T](HttpResponse response, RequestContext requestContext)
   at Microsoft.Identity.Client.OAuth2.OAuth2Client.ExecuteRequestAsync[T](Uri endPoint, HttpMethod method, RequestContext requestContext, Boolean expectErrorsOn200OK, Boolean addCommonHeaders, Func`2 onBeforePostRequestData)
   at Microsoft.Identity.Client.OAuth2.OAuth2Client.GetTokenAsync(Uri endPoint, RequestContext requestContext, Boolean addCommonHeaders, Func`2 onBeforePostRequestHandler)
   at Microsoft.Identity.Client.OAuth2.TokenClient.SendHttpAndClearTelemetryAsync(String tokenEndpoint, ILoggerAdapter logger)
   at Microsoft.Identity.Client.OAuth2.TokenClient.SendTokenRequestAsync(IDictionary`2 additionalBodyParameters, String scopeOverride, String tokenEndpointOverride, CancellationToken cancellationToken)
   at Microsoft.Identity.Client.Internal.Requests.UsernamePasswordRequest.GetTokenResponseAsync(CancellationToken cancellationToken)
   at Microsoft.Identity.Client.Internal.Requests.UsernamePasswordRequest.ExecuteAsync(CancellationToken cancellationToken)
   at Microsoft.Identity.Client.Internal.Requests.RequestBase.RunAsync(CancellationToken cancellationToken)
   at Microsoft.Identity.Client.ApiConfig.Executors.PublicClientExecutor.ExecuteAsync(AcquireTokenCommonParameters commonParameters, AcquireTokenByUsernamePasswordParameters usernamePasswordParameters, CancellationToken cancellationToken)

neha-bhargava:
Fixes #3883 

Handle exception when json exception occurs in B2C scenario which could be because graph scopes are used which are not supported by B2C.

**Testing**

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -453,5 +453,7 @@ public static string InvalidTokenProviderResponseValue(string invalidValueName)
 
         public const string CombinedUserAppCacheNotSupported = "Using a combined flat storage, like a file, to store both app and user tokens is not supported. Use a partitioned token cache (for ex. distributed cache like Redis) or separate files for app and user token caches. See https://aka.ms/msal-net-token-cache-serialization .";
 
+        public const string JsonParseExceptionB2C = "There might be configuration issues with B2C app or graph scopes are used for B2C application. B2C does not support graph scopes. See inner exception for more details.";
```
gladjohn, line 456: should this be extended to mention what scopes needs to be passed to fix this issue?

src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs
```diff
@@ -45,7 +50,19 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
             await UpdateUsernameAsync().ConfigureAwait(false);
 
-            MsalTokenResponse msalTokenResponse = await GetTokenResponseAsync(cancellationToken).ConfigureAwait(false);
+            MsalTokenResponse msalTokenResponse = null;
+                
+            try
+            {
+                msalTokenResponse = await GetTokenResponseAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (JsonException ex)
```
bgavrilMS, line 59: I'd catch everything.

src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs
```diff
@@ -45,7 +50,19 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            {
+                if (_requestParameters.AuthorityInfo.AuthorityType == AuthorityType.B2C)
```
bgavrilMS, line 61: if it isn't a B2C exception, you should not swallow exception, you need to re-throw it. When rethrowing, make sure to use `throw;` instead of `throw ex;` because `throw;` will preserve the stack trace. This is a blocking comment.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
bgavrilMS, line 456:
```suggestion
        public const string JsonParseExceptionB2C = "An error occurred - see inner exception for more details. Possible cause: you are trying to get a token for Microsoft Graph using a B2C identity, which is not supported.";

```

bgavrilMS:
Unit test for this?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
neha-bhargava: Updated the message.

src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs
```diff
@@ -45,7 +45,21 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
             await UpdateUsernameAsync().ConfigureAwait(false);
 
-            MsalTokenResponse msalTokenResponse = await GetTokenResponseAsync(cancellationToken).ConfigureAwait(false);
+            MsalTokenResponse msalTokenResponse = null;
+                
+            try
+            {
+                msalTokenResponse = await GetTokenResponseAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception ex) when (ex is not MsalException && ex is not TaskCanceledException)
```
pmaytak, line 54:
```suggestion
            catch (JsonException ex)
```

src/client/Microsoft.Identity.Client/Internal/Requests/UsernamePasswordRequest.cs
pmaytak: I would catch a JsonException on line 54, then here throw a more general message, something like "There was an error parsing the response from the token endpoint. Verify that your app is configured correctly. If this is a B2C app, one possible cause is acquiring a token for Microsoft Graph, which is not supported."

Because this Json parsing error can happen in other cases, so we're just future proofing this way.

pmaytak: Since this error happens because instead of Json, AAD returns an HTML error page, we could also log that response into ErrorPii I guess.

neha-bhargava: Thanks for the explanation @pmaytak. Wasn't able to repro this scenario and hence thought catching everything for B2C scenario is better. This makes more sense. Will update the error message and catch just JsonException. CC: @bgavrilMS

pmaytak: > Since this error happens because instead of Json, AAD returns an HTML error page, we could also log that response into ErrorPii I guess, but that would require more work since that code is inside token client classes, so would be out of scope for this work item.

In fact, thinking about this more, that error page will probably be hard to read in the logs anyway, so disregard that comment.

neha-bhargava: Ok, was not planning to log it as it is HTML page.

pmaytak (approved):

Issue #1538, opened 2019-12-06
[Feature Request] MSAL support for other IdP's such as Okta


adefalque:
Hi, is MSAL only compatible with MS Authorization servers or we could use it with other authorization server as well?
I'm trying to use it with Okta and the first error I'm getting is about the ClientID which should be a Guid, because of this code:
```
//Adfs does not require client id to be in the form of a Guid
if (Config.AuthorityInfo?.AuthorityType != AuthorityType.Adfs && !Guid.TryParse(Config.ClientId, out _))
{
    throw new MsalClientException(MsalError.ClientIdMustBeAGuid, MsalErrorMessage.ClientIdMustBeAGuid);
}
```

OAuth / OpenId does not seem to have any particular restriction on ClientID to be a GUID.

bgavrilMS:
MSAL only works with MS Identity Providers (AAD, B2C and ADFS) at this time.

adefalque:
Thanks for the answer Bogdan. Yes that's what I guessed. Would be great to support other IDPs that respect the OpenId/OAuth standard.

jmprieur:
@jmprieur to check with PM team if this would be a goal (scenario multi-clouds)

fowl2:
Would also be useful for 'smaller things' like Adobe Sign.

mliesmons:
Would love it too.
The comment below for a WithAuthority constructor gave me some hope:

> This constructor is mainly used for scenarios where the authority is not a standard
        //     Azure AD authority, nor an ADFS authority, nor an Azure AD B2C authority.

chrisbecke:
Oh crikey.

MSAL seems to be the one dotnet library that manages token refreshing, and is properly integrated with AspNetCore.

But, applications frequently need to integrate with other applications. Its the reason we use open startards alike OpenID and OAuth.

I need to deal with federated auth that only includes AzureAD as part of the stack. Finding out I can't use Microsoft.Identity.* to get and refresh my Keycloakd and Okta tokens just wastes the time I've spend on the AzureAS part of the integration because now there is no general solution that covers all my target IdPs.

I guess I'm back to trying to do everything the hard way with [IdentityModel](https://www.nuget.org/packages/IdentityModel/)

ChrisAWith:
I support this request.

nthurner:
I support this request.

bgavrilMS:
[...]

PR #3948, opened 2023-02-07
Introduce client_credentials flow support with GenericAuthority

g7ed6e:
Relates to #1538 

Introduce a new `GenericAuthority` type.
Introduce ~~2~~1 public API to `AbstractApplicationBuilder<T>`:
~~- `T WithGenericAuthority(string authorityUri, string discoveryDocumentAddress = null)`~~
~~- `T WithDocumentRetriever(IDocumentRetriever documentRetriever)`~~
- `T WithDocumentRetriever(string authorityUri)`

**Testing**
- [X] End to end tests of a client_credentials flow assert an access_token is retrieved.
- [ ] End to end tests of a user flow should assert an exception is thrown.

**Performance impact**

**Documentation**


g7ed6e:
> Which IDPs should I test this with @g7ed6e ?

I wrote unit tests with [the demo instance of Duende IdentityServer](https://demo.duendesoftware.com/). You can pull the branch and run the unit tests of the class `GenericAuthorityTests`.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -0,0 +1,146 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
g7ed6e, line 1: @bgavrilMS These test rely on an online openidconnect server which I think is causing the build fail. I don't have enough permission in azure devops to see what is causing the build fail.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -894,6 +894,38 @@ public T WithB2CAuthority(string authorityUri)
...
+            return (T)this;
+        }
+        
+        /// <summary>
+        /// Configure a <see cref="IDocumentRetriever"/> to retrieve the discovery metadata from the OpenIdConnect Provider.
+        /// </summary>
+        /// <param name="documentRetriever"><see cref="IDocumentRetriever"/> instance responsible for retrieving the discovery metadata</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        /// <exception cref="NotSupportedException"> is thrown when not used with <see cref="WithGenericAuthority"/>.</exception>
+        public T WithDocumentRetriever(IDocumentRetriever documentRetriever)
```
bgavrilMS, line 917: I don't think we want to expose this kind of extensibility at first, i.e. nobody asked for it. I think we should hide this for now?

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -137,17 +146,17 @@ internal class AuthorityInfo
         internal bool IsTenantOverrideSupported => AuthorityType == AuthorityType.Aad;
         internal bool IsMultiTenantSupported => AuthorityType != AuthorityType.Adfs;
         internal bool IsClientInfoSupported => AuthorityType == AuthorityType.Aad || AuthorityType == AuthorityType.Dsts || AuthorityType == AuthorityType.B2C;
+        internal string DiscoveryDocumentAddress { get; set; }
```
bgavrilMS, line 149: AuthorityInfo is a common object for all authorities. Would `DiscoveryDocumentAddress` be better placed in GenericAuthority class?

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
```diff
@@ -0,0 +1,59 @@
...
+           
+        internal GenericAuthority(AuthorityInfo authorityInfo)
+            : base(authorityInfo)
+        {
+
+        }
+
+        internal override string TenantId => null;
+    
+        internal IDocumentRetriever DocumentRetriever { get; set; } = new HttpDocumentRetriever();
```
bgavrilMS, line 25: `HttpDocumentRetriever` is not included.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -894,6 +894,38 @@ public T WithB2CAuthority(string authorityUri)
...
         }
 
+        /// <summary>
+        /// Adds a known authority corresponding to a generic OpenIdConnect Provider (OP).
+        /// See https://openid.net/specs/openid-connect-core-1_0.html#Terminology
+        /// </summary>
+        /// <param name="authorityUri">OpenIdConnect authority</param>
+        /// <param name="discoveryDocumentAddress">Address of the discovery endpoint. See https://openid.net/specs/openid-connect-discovery-1_0.html</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public T WithGenericAuthority(string authorityUri, string discoveryDocumentAddress = null)
```
bgavrilMS, line 904: @jmprieur for public API guidance. 

Context: https://developers.google.com/identity/openid-connect/openid-connect - they suggest that developers hardcode the well-known endpoint in their apps. 

So how should we model this API? Here's what I'm thinking 

## Basic Scenario 

- `WithGenericAuthority(string wellKnownEndpoint)`  

and here MSAL does OIDC discovery and pulls the token endpoint and the authorization endpoint. Internally, for caching, we can say that the `environment` is `wellKnownEndpoint.TrimEnd("/.well-known/openid-configuration")`.

## Advanced scenario

(to be implemented later)

- `WithGenericAuthority(GenericAuthorityInfo.FromOidcDocument(oidcDiscoveryDoc))`  
- `WithGenericAuthority(GenericAuthorityInfo.FromEndpoints(tokenEndpoint, authorizationEndpoint))`  


Thoughts?

bgavrilMS:
@g7ed6e - there seem to be some unit tests failing

image

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
```diff
@@ -0,0 +1,59 @@
...
+        internal override string GetTokenEndpoint()
+        {
+            string cacheKey = AuthorityInfo.CanonicalAuthority.ToString();
+            var configuration = s_openIdConnectConfigurations.GetOrAdd(cacheKey, RetrieveOpenIdConnectConfiguration);
+            return configuration.TokenEndpoint;
+        }
+
+        private OpenIdConnectConfiguration RetrieveOpenIdConnectConfiguration(string canonicalizedAuthority)
+        {
+            var configuration = OpenIdConnectConfigurationRetriever.GetAsync(AuthorityInfo.DiscoveryDocumentAddress, DocumentRetriever, CancellationToken.None)
```
bgavrilMS, line 41: Thoughts about having this operation done in a critical section? To prevent a bunch of requests that are fired at the same time to make many discovery requests? 

We do this elsewhere like:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/a07da18fd6a8ff1340fe9c7ca833c589ad796c3e/src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs#L173

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
bgavrilMS: Alternatively, use `HttpDocumentRetrieve(HttpClient)` constructor, but I am hesitant to take a dependency on those libraries. Will get back to you on this one.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
g7ed6e: When working behind enterprise http proxy you need to be able to customize the underlying `HttpClient`. Exposing `IDocumentRetriever` allow one to configure this providing an `HttpDocumentRetriever` instance with its own HttpClient (please see https://learn.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.protocols.httpdocumentretriever?view=msal-web-dotnet-latest#constructors). I could have exposed `WithHttpClient` to align with `OpenIdConnectConfigurationRetriever.Task<OpenIdConnectConfiguration> GetAsync(string address, HttpClient httpClient, CancellationToken cancel)` (please see https://learn.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.protocols.openidconnect.openidconnectconfigurationretriever.getasync?view=msal-web-dotnet-latest#overloads) but `IDocumentRetriever` is more generic and let one configure MSAL with offline metadata document by providing a `FileDocumentRetriever`.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
g7ed6e: Please see also https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest that describes that metadata document is supposed to be issuer + "/.well-known/openid-configuration"

g7ed6e:
> @g7ed6e - there seem to be some unit tests failing
> 
> image
> 
> You should be able to run unit tests locally, please let me know if that is not the case.
> 
> Also, to speed up development, I suggest you edit the msal.csproj and comment out
> 
> image

@bgavrilMS I fixed the unit tests. Thanks

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
bgavrilMS: Update on this - will discuss with folks on my side about using `HttpDocumentRetriever` dependency (with HttpClient). For now, let's continue to use it.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
g7ed6e: @bgavrilMS it looks like the Authority instance which is set into Config when using the builder is not the same one as the one used when requesting a token causing GenericAuthority properties to be null when requesting a token. Is it the normal behavior ? Should I try to stabilize the Config.Authority instance between builder usage and fetching the token ?

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
g7ed6e: @bgavrilMS Thanks I will. I just spotted a bug related to the fact that the Authority instance used to request the token is not the same one as the one used when using the builder, causing the DocumentRetriever property instance used being defaulted. Please advise about expected lifetimes of Authority instances. My understanding is that an AuthorityInfo instance should be able to new an Authority instance which implies that AuthorityInfo should store all required information to re build a new instance of a concrete Authority.

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Threading;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+
+namespace Microsoft.Identity.Client.Instance
+{
+    internal class GenericAuthority : Authority
+    {
+        internal const string OpenIdConfigurationEndpointSuffix = ".well-known/openid-configuration";
+        private static readonly ConcurrentDictionary<string, OpenIdConnectConfiguration> s_openIdConnectConfigurations = new();
```
brentschmaltz, line 15: What is this dictionary used for?

g7ed6e: @brentschmaltz This dictionary is used to cache discovery responses I'm planning to replace it with either a MemoryCache (but i did not found a mono implementation yet) or a MSAL cache (I did not explored this track yet).

bgavrilMS:
@g7ed6e - I pushed a commit as discussed, to add support for HttpClient. I also wanted to fix the fact the existing methods `GetTokenEndpoint` etc. were sync, but we need them to be async. But as you can imagine, this propagated to a LOT of changes. Apologies for the large commit. Please let me know what you think.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -13,7 +13,7 @@
 using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
 using Microsoft.Identity.Client.Utils;
 using Microsoft.IdentityModel.Abstractions;
-using Microsoft.Identity.Client.Internal;
+using Microsoft.IdentityModel.Protocols;
```
g7ed6e, line 16: remove

g7ed6e:
> @g7ed6e - I pushed a commit as discussed, to add support for HttpClient. I also wanted to fix the fact the existing methods `GetTokenEndpoint` etc. were sync, but we need them to be async. But as you can imagine, this propagated to a LOT of changes. Apologies for the large commit. Please let me know what you think.

Thanks @bgavrilMS ! I see you enabled the authorize endpoint. Do you want this PR to enable interactive flow too ? I did some tests and noticed only minor modifications are required to enable an interactive flow with the code below.

```c#
        [TestMethod]
        public async Task ShouldSupportAuthorizationCodeFlowWithDuendeDemoInstanceAsync()
        {
            var applicationConfiguration = new ApplicationConfiguration(true);
            PublicClientApplicationBuilder builder = new(applicationConfiguration);
            var app = builder.WithGenericAuthority(DemoDuendeSoftwareDotCom)
                .WithClientId("interactive.public")
                .WithRedirectUri("https://notused")
                .Build();
            var response = await app.AcquireTokenInteractive(new[] { "openid", "profile" }).ExecuteAsync().ConfigureAwait(false);
            
            Assert.IsTrue(response.Scopes.Contains("openid"));
            Assert.IsTrue(response.Scopes.Contains("profile"));
            Assert.AreEqual("Bearer", response.TokenType);
            Assert.AreEqual(TokenSource.IdentityProvider, response.AuthenticationResultMetadata.TokenSource);

            var accounts = await app.GetAccountsAsync().ConfigureAwait(false);
            Assert.AreEqual(1, accounts.Count());

            var response2 = await app.AcquireTokenSilent(new[] { "openid", "profile" }, accounts.First()).ExecuteAsync().ConfigureAwait(false);
            
            Assert.IsTrue(response2.Scopes.Contains("openid"));
            Assert.IsTrue(response2.Scopes.Contains("profile"));
            Assert.AreEqual("Bearer", response2.TokenType);
            Assert.AreEqual(TokenSource.Cache, response2.AuthenticationResultMetadata.TokenSource);
        }
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
g7ed6e: removed

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
g7ed6e: `IDocumentRetriever` is no longer exposed in MSAL.NET public API

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
g7ed6e: implemented in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3948/commits/a6454b41b05f539ea89bb5e5bdf12f54eb4651dc

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
g7ed6e: Ok!

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Utils;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.CoreTests.InstanceTests
+{
+    [TestClass]
+    public class GenericAuthorityTests
```
bgavrilMS, line 20: @g7ed6e - still testing this, but hope we can get it in in the next release, in 1-2 weeks. 

Btw, it looks like the HttpDocumentRetriever actually makes 2 HTTP calls: 

image

This makes sense for token validation, but for token acquisition we don't care about the public key of the identity provider.

g7ed6e: > @g7ed6e - still testing this, but hope we can get it in in the next release, in 1-2 weeks.

Thanks @bgavrilMS ! That's good news!

bgavrilMS:
@g7ed6e - I pushed another commit to this where I removed the dependency on Microsoft.identityModel.Protocols for retrieving the OIDC metadata. The reason for this is that MSAL has special handling of all HTTP calls (e.g. a simple inbuilt retry policy, measuring the time spent in HTTP) and also because Microsoft.identityModel.Protocols is making an extra call to well-known/jwk to retrieve the signature from the IdP. This is not needed for token acquisition, only for token validation.

We can always come back to this decision. 

I'm asking folks on my team to review this so that it catches the next release. Let me know if you have any concerns.

g7ed6e:
> I'm asking folks on my team to review this so that it catches the next release. Let me know if you have any concerns.

@bgavrilMS do you still plan to get this into next release ?

bgavrilMS:
@g7ed6e - yes, targeting 29 March

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                    .AcquireTokenForClient(new[] { "api" })
+                    .ExecuteAsync(CancellationToken.None)
+                    .ConfigureAwait(false);
+
+                Assert.IsNotNull(result);
+                Assert.IsNotNull(result.AccessToken);
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+
+                result = await app
+                    .AcquireTokenForClient(new[] { "api" })
```
neha-bhargava, line 60: Update the comment for [AcquireTokenForClient](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/g7ed6e%2Fgeneric-authority-support/src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs#L93) for scopes for generic authority. The scopes in the comment are specific to azure.

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -147,7 +151,7 @@ internal static AuthorityInfo FromAuthorityUri(string authorityUri, bool validat
             var authorityType = GetAuthorityType(canonicalUri);
 
             // If the authority type is B2C, validateAuthority must be false.
```
gladjohn, line 153:
```suggestion
            // If the authority type is B2C or Generic, validateAuthority must be false.
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -611,11 +610,30 @@ public T WithAdfsAuthority(string authorityUri, bool validateAuthority = true)
...
+        /// <summary>
+        /// Adds a known authority corresponding to a generic OpenIdConnect Identity Provider. 
+        /// MSAL will append ".well-known/openid-configuration" to the authority and retrieve the OIDC 
+        /// metadata from there, to figure out the endpoints.
+        /// See https://openid.net/specs/openid-connect-core-1_0.html#Terminology
+        /// </summary>
+        /// <param name="authorityUri">OpenIdConnect authority</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        /// <remarks>This is an experimental API and only AcquireTokenForClient (client_credentials flow) has been implemented</remarks>        
+        public T WithGenericAuthority(string authorityUri)
```
trwalke, line 627: This is placed in the abstract app builder and it is available to the PublicClientApplicationBuilder right? Maybe it is better to put this in ConfidentialClientApplicationBuilder to avoid exposing this to public clients since they dont implement the client_credentials flow

trwalke: Is there a plan to move this to public clients?

g7ed6e: @trwalke I did some tests to use the GenericAuthority with a public client using an authorization_code flow and minor changes were required. 
@bgavrilMS would you be open to a follow up PR that enable GenericAuthority support to public clients ?

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.Identity.Client.OAuth2;
+
+namespace Microsoft.Identity.Client.Instance.Oidc
+{
+    internal static class OidcRetrieverWithCache
+    {
+        private static readonly ConcurrentDictionary<string, OidcMetadata> s_cache = new();
+        private static readonly SemaphoreSlim s_lockOidcRetrieval = new SemaphoreSlim(1);
+
+        internal const string OpenIdConfigurationEndpointSuffix = ".well-known/openid-configuration";
```
trwalke, line 19: Seems like this is already available from Constants.cs. No need to have a duplicate.

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
```diff
@@ -0,0 +1,57 @@
...
+               AuthorityInfo.CanonicalAuthority.AbsoluteUri, requestContext)
+                .ConfigureAwait(false);
+
+            return configuration.AuthorizationEndpoint;
+        }
+
+        internal override Task<string> GetDeviceCodeEndpointAsync(RequestContext requestContext)
+        {
+            // prevents device_code flow which requires knowledge of the device_authorization_endpoint.
+            throw new NotImplementedException();
```
trwalke, line 54: If this is going to be enabled for public clients then I think it makes sense to throw a more actionable error here so developers have a clear understanding of when this should not be used with generic authorities. This can also be handled somewhere higher up the chain of calls when the application is being configured. More of a suggestion for now.

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
bgavrilMS: @g7ed6e - absolutely! Separate PR please, this one is already huge.

src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs
bgavrilMS: I'll move it to CCA. The API is experimental, so not too worried about not being ready everywhere.

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeRequestComponent.cs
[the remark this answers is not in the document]
bgavrilMS: I think for consistency getting the endpoints should be an async action.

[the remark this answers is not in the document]
pmaytak: Ah yea, the method signatures should be updated, I meant we added that extra call to Authority.GetAuthCodeEndpoint.

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public static async Task<OidcMetadata> GetOidcAsync(
+            string authority,
+            RequestContext requestContext)
+        {
+            OidcMetadata configuration = null;
+
+            // Conccurent dictionary get or add
+            if (s_cache.TryGetValue(authority, out configuration))
+                return configuration;
```
pmaytak, line 27:
```suggestion
            if (s_cache.TryGetValue(authority, out configuration))
            {
                requestContext.Logger.Verbose(() => $"[OIDC Discovery] OIDC discovery found a cached entry for {authority}. ");
                return configuration;
             }
```

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            await s_lockOidcRetrieval.WaitAsync().ConfigureAwait(false);
+
+            try
+            {
+                // try again in critical section
+                if (s_cache.TryGetValue(authority, out configuration))
+                    return configuration;
```
pmaytak, line 35:
```suggestion
                if (s_cache.TryGetValue(authority, out configuration))
                {
                requestContext.Logger.Verbose(() => $"[OIDC Discovery] OIDC discovery found a cached entry for {authority}. ");
                    return configuration;
                 }
```

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                Uri oidcMetadataEndpoint = new Uri(authority + Constants.WellKnownOpenIdConfigurationPath);
+
+                var client = new OAuth2Client(requestContext.Logger, requestContext.ServiceBundle.HttpManager);
+                configuration = await client.DiscoverOidcMetadataAsync(oidcMetadataEndpoint, requestContext).ConfigureAwait(false);
+
+                s_cache[authority] = configuration;
+                return configuration;
```
pmaytak, line 43:
```suggestion
                s_cache[authority] = configuration;
                requestContext.Logger.Verbose(() => $"[OIDC Discovery] OIDC discovery retrieved metadata from the network for {authority}. ");
                return configuration;
```

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            catch (Exception ex)
+            {
+                requestContext.Logger.Error(
+                    $"Failed to retrieve OpenId configuration from the OpenId endpoint {authority + Constants.WellKnownOpenIdConfigurationPath} " +
+                    $"due to {ex}");
```
pmaytak, line 49:
```suggestion
                requestContext.Logger.Error(
                    $"[OIDC Discovery] Failed to retrieve OpenId configuration from the OpenId endpoint {authority + Constants.WellKnownOpenIdConfigurationPath} due to {ex}");
```

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -81,10 +81,10 @@ public sealed partial class ConfidentialClientApplication
...
         /// using the client credentials flow. See https://aka.ms/msal-net-client-credentials.
         /// </summary>
-        /// <param name="scopes">scopes requested to access a protected API. For this flow (client credentials), the scopes
-        /// should be of the form "{ResourceIdUri/.default}" for instance <c>https://management.azure.net/.default</c> or, for Microsoft
-        /// Graph, <c>https://graph.microsoft.com/.default</c> as the requested scopes are defined statically with the application registration
-        /// in the portal, and cannot be overridden in the application.</param>
+        /// <param name="scopes">scopes requested to access a protected API. 
+        /// For this flow (client credentials), the scopes of Microsoft APIs protected with AAD tokens are of the form "{ResourceIdUri/.default}". For example <c>https://management.azure.net/.default</c> or, for Microsoft
+        /// Graph, <c>https://graph.microsoft.com/.default</c>. 
+        /// The requested scopes are defined statically with the application registration in the portal, and cannot be overridden in the application.</param>        ///
```
pmaytak, line 87:
```suggestion
        /// The requested scopes are defined statically with the application registration in the portal, and cannot be overridden in the application.</param>
```

pmaytak (approved):

pmaytak:
Generic authority is supported only for client credentials in confidential client apps. See possible solution [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1538#issuecomment-1014480612) and use #1538 for tracking and comments.

PR #3953, opened 2023-02-08
Fix some build warnings

bgavrilMS:


src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
```diff
@@ -140,6 +140,7 @@ public JWTHeader(X509Certificate2 certificate)
                 Certificate = certificate;
             }
 
+            [JsonIgnore]
             protected X509Certificate2 Certificate { get; }
```
bgavrilMS, line 144: SJT is trying to source gen this and this causes a bunch of problems. We don't serialize the cert.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -57,7 +57,7 @@ public virtual async Task<ManagedIdentityResponse> AuthenticateAsync(AppTokenPro
             }
         }
 
-        protected virtual async Task<ManagedIdentityResponse> HandleResponseAsync(
+        protected virtual Task<ManagedIdentityResponse> HandleResponseAsync(
```
bgavrilMS, line 60: There is no `await` in this method.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -92,7 +92,7 @@
     <PathToMsalSources>$(MSBuildThisFileDirectory)</PathToMsalSources>
     <!-- Need to include / exclude / remove cs files manually so that they are present in both assemblies -->
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    <NoWarn>$(NoWarn);CS8002;NU5131;NETSDK1138</NoWarn>
+    <NoWarn>$(NoWarn);CS8002;NU5131;NETSDK1138;SYSLIB0026;SYSLIB0027;SYSLIB0028</NoWarn>
```
bgavrilMS, line 95: Even with `JsonIgnore` on that X509Certificate2, it looks like SJT is issuing warnings about invalid use of obsolete properties from X509Certificate2 (e.g. PublicKey, PrivateKey ... full list here https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/obsoletions-overview). We don't use these properties. 

So I disabled the warnings. I'm on the fence about this one, as disabling the warning may cause MSAL developers to use a deprecated x509cert property.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -202,11 +202,11 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet6Win)'">
     <UseWindowsForms>true</UseWindowsForms>
     <!-- 
-    MSAL will run on Windows 7 and 8 but requires to be built against Win10 to use WinRT APIs for WAM 
+    MSAL will run on Windows 8 but requires to be built against Win10 to use WinRT APIs for WAM 
     See https://docs.microsoft.com/en-us/dotnet/standard/analyzers/platform-compat-analyzer and 
     https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md for details
     -->
-    <SupportedOSPlatformVersion>7</SupportedOSPlatformVersion>
+    <SupportedOSPlatformVersion>8</SupportedOSPlatformVersion>
```
bgavrilMS, line 209: win 7 is deprecated and this is causing a warning.

src/client/Microsoft.Identity.Client/PublicClientApplication.cs
```diff
@@ -91,19 +91,11 @@ public bool IsUserInteractive()
...
-        /// <item><description>On Mac, Linux, and older versions of Windows a broker is not available.</description></item>
-        /// <item><description>In .NET 6 apps, target <c>net6.0-windows10.0.17763.0</c> for all Windows versions and target <c>net6.0</c> for Linux and Mac.</description></item>
-        /// <item><description>In .NET classic or .NET Core 3.1 apps, install Microsoft.Identity.Client.Desktop first and call <c>WithDesktopFeatures()</c>.</description></item>
-        /// <item><description>In mobile apps, the device must be Intune joined and Authenticator or Company Portal must be installed. See https://aka.ms/msal-brokers </description></item>
-        /// </list>
+        /// This method is deprecated
         /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("This method is obsolete. Applications should rely on the library automatically falling back to a browser if the broker is not available. ", false)]
         public bool IsBrokerAvailable()
```
bgavrilMS, line 99: This should have been obsoleted as well.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
neha-bhargava: Right, it is in one of the overridden methods.

bgavrilMS: That's ok. You can omit `async` in the the class hierarchy, i.e. `async` is not part of the contract. Only the return type `Task<foo>` is important.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
neha-bhargava: Can't we do `#pragma warning disable` and then enable it instead?

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
trwalke: I agree with Neha, lets add `#pragma warning disable` or an entry to global suppressions instead so we can have a more targeted approach. Also, we have .editorconfig, global changes like this go in there right?

trwalke (approved):

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
pmaytak: It's still giving an error:
> warning CA1418: Version '8' is not valid for platform 'Windows'. Use a version with 2-4 parts for this platform.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
pmaytak: Pragma doesn't work since this happens in the generated files. Not sure how to disable only for STJ stuff and why JsonIgnore doesn't work. But yea, I wouldn't disable these warnings for everything.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -333,22 +333,11 @@ public PublicClientApplicationBuilder WithKerberosTicketClaim(string servicePrin
...
-        /// <item><description>On Mac, Linux, and older versions of Windows a broker is not available.</description></item>
-        /// <item><description>In .NET 6 apps, target <c>net6.0-windows10.0.17763.0</c> for all Windows versions and target <c>net6.0</c> for Linux and Mac.</description></item>
-        /// <item><description>In .NET classic or .NET Core 3.1 apps, install Microsoft.Identity.Client.Desktop first and call <c>WithDesktopFeatures()</c>.</description></item>
-        /// <item><description>In mobile apps, the device must be Intune joined and Authenticator or Company Portal must be installed. See https://aka.ms/msal-brokers </description></item>
-        /// </list>
+        /// This method is deprecated
         /// </remarks>
-        [Obsolete("This method is obsolete. Applications should rely on the library automatically falling back to a browser if the broker is not available. ", false)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("This method is obsolete. Applications can rely on the library automatically falling back to a browser if the broker is not available. ", false)]
```
pmaytak, line 340: Undo all this.

src/client/Microsoft.Identity.Client/PublicClientApplication.cs
pmaytak: I think this should be undone here and be part of #3958, so it's all in one place.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
pmaytak: Found a related conversation in STJ repo: https://github.com/dotnet/runtime/issues/76829#issuecomment-1349391310

bgavrilMS:
Closing as most solutions were rejected.

Issue #3825, opened 2022-11-25
[Bug] LongOBO. Issues with cache serialization after user session has been revoked


lukasz-zoglowek:
**Which version of MSAL.NET are you using?**
4.48.1

**Platform**
.NET 6.0

**What authentication flow has the issue?**
* Desktop / Mobile
    * [ ] Interactive
    * [ ] Integrated Windows Authentication
    * [ ] Username Password
    * [ ] Device code flow (browserless)
* Web app 
    * [ ] Authorization code
    * [X] On-Behalf-Of
* Daemon app 
    * [ ] Service to Service calls

Other?

**Is this a new or existing app?**
b. The app is in production, I haven't upgraded MSAL, but started seeing this issue.

**Repro**

```csharp


var app = ConfidentialClientApplicationBuilder.Create(clientId)
        .WithAuthority(authorityUri, true)
        .WithCertificate(certificate).Build() as ILongRunningWebApi;

string userTokenCache = string.Empty;

application.UserTokenCache.SetAfterAccess(args =>
{
    if (args.HasStateChanged)
    {
        userTokenCache = args.TokenCache.SerializeMsalV3().FromByteArray();
    }
});

var accessToken = ...
string sessionKey1 = string.Empty;

try
{
    var result = await app.InitiateLongRunningProcessInWebApi(scopes, accessToken, ref sessionKey1)
                                        .WithSendX5C(true)
                                        .ExecuteAsync(); //SUCCESS
    // sessionKey1 ==   "Q97ineyQVRJEjXsoxa24cFMXjOsrD8rg1zUFrINWNyo"                                    
                                        
    // !!!! At this point admin revokes user session
    result = await app.AcquireTokenInLongRunningProcess(scopes, sessionKey1)
                                    .WithForceRefresh(true) // to repro sooner
                                    .WithSendX5C(true)
                                    .ExecuteAsync(); // FAILS (Expected)

}
catch (MsalServiceException e) // AADSTS50173: The provided grant has expired due to it being revoked, a fresh auth token is needed.
{
    // New valid access token
    var newAccessToken = ...

    // The below call doesn't work even though a new valid access token is provided
    //var result = await app.InitiateLongRunningProcessInWebApi(scopes, newAccessToken, ref sessionKey1)
    //                                     .WithSendX5C(true)
    //                                    .ExecuteAsync();   

    // The old session key cannot be used anymore
    var sessionKey2 = string.Empty;

[...]

PR #3956, opened 2023-02-09
LongRunning OBO cache fix for different user assertions

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3825

Enabling obo request to acquire new token if user cached assertion is different from the one provided.
This is done by ensuring `InitiateLongRunningProcessInWebApi` always goes to get a new token rather than using what is already in the cache for a specified cache key.

**Testing**
Unit

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
@@ -128,9 +128,11 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
         private bool IsCachedAssertionTheSame(string userAssertion)
         {
+            //User assertion is null when ILongRunningWebApi.AcquireTokenInLongRunningProcess is called.
+            //In this scenario the cached token should be used.
             if (string.IsNullOrWhiteSpace(userAssertion))
             { 
-                return false;
+                return true;
             }
 
             return AuthenticationRequestParameters.UserAssertion.Assertion.Equals(userAssertion);
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
@@ -40,6 +40,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
             MsalAccessTokenCacheItem cachedAccessToken = null;
             var logger = AuthenticationRequestParameters.RequestContext.Logger;
             AuthenticationResult authResult = null;
+            bool isCachedAccessTokenValid = false;
 
             CacheRefreshReason cacheInfoTelemetry = CacheRefreshReason.NotApplicable;
             if (!_onBehalfOfParameters.ForceRefresh && string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
@@ -55,7 +56,8 @@ namespace Microsoft.Identity.Client.Internal.Requests
                     cachedAccessToken = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
                 }
 
-                if (cachedAccessToken != null)
+                isCachedAccessTokenValid = cachedAccessToken != null && IsCachedAssertionTheSame(cachedAccessToken.Secret);
+                if (isCachedAccessTokenValid)
                 {
                     var cachedIdToken = await CacheManager.GetIdTokenCacheItemAsync(cachedAccessToken).ConfigureAwait(false);
                     var account = await CacheManager.GetAccountAssociatedWithAccessTokenAsync(cachedAccessToken).ConfigureAwait(false);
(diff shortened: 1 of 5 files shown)


trwalke:
Seems like it makes sense to check if the cached assertion is the same when using `InitiateLongRunningProcessInWebApi`.
initialize indicates that you want to start a new long running session and the previous one should not interfere even though the same session key is used. `AcquireTokenInLongRunningProcess` should still respect the cached user assertion however. 

Thoughts @bgavrilMS @pmaytak

pmaytak:
My initial thought - your conclusion is right. If cache key exists, scopes are the same, but assertion is different, we should replace the cached ATs and RTs. Currently we'll return the cached AT if it's not expired. [Here's](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/16b744a4c56efb20073693c081a3512aa385da25/tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs#L229) the test for these scenarios. So we'd need to update it with an extra Initiate call.

Would this have any behavioral breaking changes? Currently if users call Initiate for existing key, they'd expect a cached token; but with the change, each new user assertion would result in a network call, which makes sense - if new user assertion, then give my an updated token.

I'll have to think a bit more about this.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -97,7 +99,7 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             // No AT in the cache or AT needs to be refreshed
             try
             {
-                if (cachedAccessToken == null)
+                if (!isCachedAccessTokenValid)
```
pmaytak, line 102: This won't exactly work. RefreshRt method would try to use the cached RT to get a token and AAD will throw. So need to make sure to just call get new AT method right away.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -55,7 +56,8 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                     cachedAccessToken = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
                 }
 
-                if (cachedAccessToken != null)
+                isCachedAccessTokenValid = cachedAccessToken != null && IsCachedAssertionTheSame(cachedAccessToken.Secret);
```
pmaytak, line 59: Thinking if this filtering could be done in FindAT/FindRT methods in the TokenCache class, but it's probably more efficient to do it here.

pmaytak: After discussion - the filtering should probably be done here. Also a method like IsInitiateLROBO called would also be helpful.

pmaytak:
This test is one that would need to be updated.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/665971b6d514c9e2e3cefc90e9ae3cbcaa06550f/tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs#L262

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -225,6 +225,51 @@ public async Task AcquireTokenByObo_InLongRunningProcess_TestAsync()
...
+
+                Assert.AreEqual(TestConstants.ATSecret2, result.AccessToken);
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+                cachedAccessToken = cca.UserTokenCacheInternal.Accessor.GetAllAccessTokens().Single();
+                cachedRefreshToken = cca.UserTokenCacheInternal.Accessor.GetAllRefreshTokens().Single();
+                Assert.AreEqual(oboCacheKey, cachedAccessToken.OboCacheKey);
+                Assert.AreEqual(oboCacheKey, cachedRefreshToken.OboCacheKey);
+                Assert.AreEqual(TestConstants.ATSecret2, cachedAccessToken.Secret);
+                userCacheAccess.AssertAccessCounts(0, 2);
+            }
```
pmaytak, line 270: I'd also add another test case here for same cache key and scopes but different assertion:
```csharp
result = await cca.InitiateLongRunningProcessInWebApi(TestConstants.s_scope, $"{TestConstants.DefaultAccessToken}2", ref oboCacheKey)
```

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -42,6 +42,15 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             AuthenticationResult authResult = null;
 
             CacheRefreshReason cacheInfoTelemetry = CacheRefreshReason.NotApplicable;
+
+            //Check if initiating a long running process
+            if (AuthenticationRequestParameters.UserAssertion != null && !string.IsNullOrEmpty(AuthenticationRequestParameters.LongRunningOboCacheKey))
```
bgavrilMS, line 47: nit: have a little helper like `IsLongOBOInitialize` kind of helper.

bgavrilMS (approved):

pmaytak (approved):

PR #3969, opened 2023-02-21
Bogavril/latest wilson

bgavrilMS:
Use latest stable Wilson dependency

### DIFF
--- a/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
+++ b/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
@@ -317,6 +317,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <!-- The floating version exlcudes previews, see https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges -->
     <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="6.*" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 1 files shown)


pmaytak (changes requested):
Hmm, not sure that using wildcard versions for dependencies is appropriate for SDKs. This way we force MSAL users to also use the latest version of dependencies. What about regressions? If we use latest, we'll have either release new MSAL version that downgrades or upgrades the dependency, once the fix is in.

I think we should reference the lowest version we can feature-wise or regression-fix-wise.

This is probably fine for the Abstractions package, since it's fairly thin, but still. Maybe good topic to discuss during the design meeting.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -317,6 +317,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="6.22.0" />
+    <!-- The floating version exlcudes previews, see https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges -->
+    <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="6.*" />
```
SameerK-MSFT, line 321: In the context of @pmaytak comment, minimum version can be specified as per this specification. https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

PR #3973, opened 2023-02-22
Fix cache keys in service principal OBO.

### LINKED ISSUE #3913: [Bug] MultipleMatchingTokensDetected in AcquireTokenInLongRunningProcess for Service Principals
**Logs and network traces**
Received logs from partner. Cannot make them public.

4.47.2

**Platform**
Web

* Desktop / Mobile
    * [ ] Interactive
    * [ ] Username Password
* Web app 
    * [ ] Authorization code
    * [X] On-Behalf-Of
* Daemon app 

Existing

**Repro**
In single partition, when user token expires and long running process attempts to acquire token by service principle (SP). The service principle does not have the home account id. As a result, both the expired user token and the new token acquired by SP are stored in the same partition in the cache.
This results in “MultipleMatchingTokensDetected” exception.
This can affect percentage of users in long running process whose token has expired.

**Expected behavior**
The keys should be filtered by home account id.

**Actual behavior**
Filtering by home account id is skipped. 

**Possible solution**
Use Microsoft.Identity.Web Token caching or use different token for custom cache keys.

pmaytak:
Fixes #3913

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedUserTokenCacheAccessor.cs
+++ b/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedUserTokenCacheAccessor.cs
@@ -115,12 +115,15 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
 
         public MsalIdTokenCacheItem GetIdToken(MsalAccessTokenCacheItem accessTokenCacheItem)
         {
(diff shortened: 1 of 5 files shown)


bgavrilMS:
Would it be better to wait for ESTS to fix this  by adding id token in the refresh_token grant?

pmaytak:
> Would it be better to wait for ESTS to fix this by adding id token in the refresh_token grant?

Yes, I think so.

Issue #3346, opened 2022-05-24
[Feature Request] Add RemoveAccount API for long-running OBO


pmaytak:
**Is your feature request related to a problem? Please describe.**
Currently there's no way to remove an account/token in a long running OBO scenario. A real-world example could be using a Graph delta API, where web API subscribes to receive some delta information periodically from Graph about a user. When a user calls the web API to unsubscribe, there should be a way to remove the OBO tokens for that user. Since long-running OBO tokens have a refresh token, there's no suggested cache expiry, so it's not set.

**Describe the solution you'd like**
Create an API something like `RemoveAccountAsync(string longRunningOboKey)` or `StopLongRunningWebApi` to not confuse it with normal OBO.

PR #3975, opened 2023-02-24
Adding api to remove OBO tokens based on provided cache key

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3346

Adding api to remove OBO tokens based on provided cache key

**Testing**
Unit tests

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1168,7 +1179,7 @@ async Task ITokenCacheInternal.RemoveAccountAsync(IAccount account, Authenticati
                             account: account,
```
trwalke, line 1179: It is difficult to acquire the account associated with the tokens since there may be multiple accounts associated with the same cache key. Lets discuss tomorow during our sync.

@bgavrilMS

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1232,6 +1251,22 @@ bool ITokenCacheInternal.HasTokensNoLocks()
             return Accessor.HasAccessOrRefreshTokens();
         }
 
+        internal /* internal for test only */ void RemoveOboAccountInternal(string partitionKey, RequestContext requestContext)
+        {
+            ILoggerAdapter logger = requestContext.Logger;
+
+            //Filter and remove tokens based on OBO Cache Key
+            var refreshTokens = Accessor.GetAllRefreshTokens(partitionKey);
+            refreshTokens.RemoveAll(item => !item.OboCacheKey.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
```
trwalke, line 1260: Filtering by obo cache key rather than home account id

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -166,6 +169,47 @@ public sealed partial class ConfidentialClientApplication
...
         }
 
+        /// <summary>
+        /// Stops an in progress long running OBO session by removing the tokens associated with the provided cache key.
+        /// </summary>
+        /// <param name="longRunningProcessSessionKey">OBO cache key used to remove the tokens</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="longRunningProcessSessionKey"/> is not set.</exception>
+        public async Task StopLongRunningWebApiAsync(string longRunningProcessSessionKey, CancellationToken cancellationToken = default)
```
bgavrilMS, line 179: Not sure about the naming, probably ok. @jmprieur to review.

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -166,6 +169,47 @@ public sealed partial class ConfidentialClientApplication
...
+        {
+            if (string.IsNullOrEmpty(longRunningProcessSessionKey))
+            {
+                throw new ArgumentNullException(nameof(longRunningProcessSessionKey));
+            }
+
+            Guid correlationId = Guid.NewGuid();
```
bgavrilMS, line 186: Why don't we obey the correlation id from the customer passes in?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1146,15 +1147,25 @@ MsalIdTokenCacheItem ITokenCacheInternal.GetIdTokenCacheItem(MsalAccessTokenCach
                 tenantProfiles?.Values);
         }
 
+        async Task ITokenCacheInternal.RemoveLongRunningOboAccountAsync(string longRunningOboCacheKey, AuthenticationRequestParameters requestParameters)
+        {
+            await RemoveAccountAsync(null, longRunningOboCacheKey, requestParameters).ConfigureAwait(false);
```
bgavrilMS, line 1152: It doesn't feel right to reuse `RemoveAccount` like this. I think you need new api.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1232,6 +1251,22 @@ bool ITokenCacheInternal.HasTokensNoLocks()
...
+            var refreshTokens = Accessor.GetAllRefreshTokens(partitionKey);
+            refreshTokens.RemoveAll(item => !item.OboCacheKey.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
+
+            RemoveRefreshTokens(refreshTokens, logger, out bool filterByClientId);
+
+            var accessTokens = Accessor.GetAllAccessTokens(partitionKey);
+            accessTokens.RemoveAll(item => !item.OboCacheKey.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
+
+            RemoveAccessTokens(accessTokens, logger, filterByClientId);
+        }
```
bgavrilMS, line 1268: I think you can now check if there are any access tokens / refresh tokens left in the node. If there aren't, I think it's ok to delete everything else.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -791,6 +792,53 @@ public async Task AcquireTokenByObo_NormalOboThenLongRunningInitiate_WithTheSame
             }
         }
 
+        [TestMethod]
+        public async Task AcquireTokenByObo_LongRunningThenRemoveTokens_TestAsync()
```
bgavrilMS, line 796: You'll need some test with custom caching to see how the TokenCacheNotification args behave.

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
trwalke: Discussed offline: Remove token apis do not communicate with the endpoint

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: Discussed offline: Passing null into notification args

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: Discussed offline: It is fine to leave the tokens there, the user developer expirence is not affected much and the tokenless accounts should be handled by cache eviction which is strongly encouraged by our docs.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1240,14 +1329,31 @@ bool ITokenCacheInternal.HasTokensNoLocks()
...
+            ILoggerAdapter logger = requestContext.Logger;
+
             var refreshTokens = Accessor.GetAllRefreshTokens(partitionKey);
-            refreshTokens.RemoveAll(item => !item.HomeAccountId.Equals(account.HomeAccountId.Identifier, StringComparison.OrdinalIgnoreCase));
+            refreshTokens.RemoveAll(item => !item.HomeAccountId.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
+
+            RemoveRefreshTokens(refreshTokens, logger, out bool filterByClientId);
+
+            var accessTokens = Accessor.GetAllAccessTokens(partitionKey);
+            accessTokens.RemoveAll(item => !item.HomeAccountId.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
```
SameerK-MSFT, line 1343: The three should be like a transaction. What if one of the methods fails? That can lead to inconsistency.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -791,6 +792,82 @@ public async Task AcquireTokenByObo_NormalOboThenLongRunningInitiate_WithTheSame
...
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+                Assert.AreEqual(1, cca.UserTokenCacheInternal.Accessor.GetAllAccessTokens().Count);
+                Assert.AreEqual(1, cca.UserTokenCacheInternal.Accessor.GetAllRefreshTokens().Count);
+                MsalAccessTokenCacheItem cachedAccessToken = cca.UserTokenCacheInternal.Accessor.GetAllAccessTokens().First(t => t.OboCacheKey.Equals(oboCacheKey));
+                MsalRefreshTokenCacheItem cachedRefreshToken = cca.UserTokenCacheInternal.Accessor.GetAllRefreshTokens().First(t => t.OboCacheKey.Equals(oboCacheKey));
+                Assert.AreEqual("access-token-1", result.AccessToken);
+                Assert.AreEqual("access-token-1", cachedAccessToken.Secret);
+                Assert.AreEqual("refresh-token-1", cachedRefreshToken.Secret);
+
+                validateCacheIsEmpty = true;
```
neha-bhargava, line 858: Remove?

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -166,6 +169,42 @@ public sealed partial class ConfidentialClientApplication
                 longRunningProcessSessionKey);
         }
 
+        /// <summary>
+        /// Stops an in progress long running OBO session by removing the tokens associated with the provided cache key.
+        /// </summary>
+        /// <param name="longRunningProcessSessionKey">OBO cache key used to remove the tokens</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns></returns>
```
bgavrilMS, line 177: delete

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -166,6 +169,42 @@ public sealed partial class ConfidentialClientApplication
...
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="longRunningProcessSessionKey"/> is not set.</exception>
+        public async Task StopLongRunningWebApiAsync(string longRunningProcessSessionKey, CancellationToken cancellationToken = default)
```
bgavrilMS, line 179: Should this return a boolean ? Simialr to List.Remove - https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.remove?view=net-7.0

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -166,6 +169,42 @@ public sealed partial class ConfidentialClientApplication
...
+        {
+            if (string.IsNullOrEmpty(longRunningProcessSessionKey))
+            {
+                throw new ArgumentNullException(nameof(longRunningProcessSessionKey));
+            }
+
+            Guid correlationId = Guid.NewGuid();
+            RequestContext requestContext = base.CreateRequestContext(correlationId, cancellationToken);
+            requestContext.ApiEvent = new ApiEvent(correlationId);
+            requestContext.ApiEvent.ApiId = ApiIds.RemoveAccount;
```
bgavrilMS, line 189: Let's please have a new ApiId for this.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1232,6 +1307,20 @@ bool ITokenCacheInternal.HasTokensNoLocks()
             return Accessor.HasAccessOrRefreshTokens();
         }
 
+        internal /* internal for test only */ void RemoveOboTokensInternal(string partitionKey, RequestContext requestContext)
+        {
+            ILoggerAdapter logger = requestContext.Logger;
+
+            //Filter and remove tokens based on OBO Cache Key
+            var refreshTokens = Accessor.GetAllRefreshTokens(partitionKey);
+            refreshTokens.RemoveAll(item => !item.OboCacheKey.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
```
bgavrilMS, line 1316: Possible null ref, as `item.OboCacheKey` can be null.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1146,15 +1147,88 @@ MsalIdTokenCacheItem ITokenCacheInternal.GetIdTokenCacheItem(MsalAccessTokenCach
...
+                }
+                finally
+                {
+                    if (tokenCacheInternal.IsAppSubscribedToSerializationEvents())
+                    {
+                        var args = new TokenCacheNotificationArgs(
+                           tokenCache: this,
+                           clientId: ClientId,
+                           account: null,
+                           hasStateChanged: true,
```
bgavrilMS, line 1196: How do you know the state changed?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1146,15 +1147,88 @@ MsalIdTokenCacheItem ITokenCacheInternal.GetIdTokenCacheItem(MsalAccessTokenCach
...
+                            piiLoggingEnabled: requestParameters.RequestContext.Logger.PiiLoggingEnabled);
+
+                        await tokenCacheInternal.OnAfterAccessAsync(args).ConfigureAwait(false);
+                    }
+                }
+            }
+            finally
+            {
+#pragma warning disable CS0618 // Type or member is obsolete
+                HasStateChanged = false;
```
bgavrilMS, line 1215: The value here should be the same as at line 1196

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1232,6 +1307,20 @@ bool ITokenCacheInternal.HasTokensNoLocks()
...
+            var refreshTokens = Accessor.GetAllRefreshTokens(partitionKey);
+            refreshTokens.RemoveAll(item => !item.OboCacheKey.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
+            RemoveRefreshTokens(refreshTokens, logger, out bool filterByClientId);
+
+            var accessTokens = Accessor.GetAllAccessTokens(partitionKey);
+            accessTokens.RemoveAll(item => !item.OboCacheKey.Equals(partitionKey, StringComparison.OrdinalIgnoreCase));
+            RemoveAccessTokens(accessTokens, logger, filterByClientId);
+        }
+
         internal /* internal for test only */ void RemoveAccountInternal(IAccount account, RequestContext requestContext)
```
bgavrilMS, line 1324: Why do we need to update this method?

bgavrilMS (changes requested):
See comments. Not sure why we need so much logic around RemoveAccount.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
trwalke: This is needed for lines 814 and 824

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
trwalke: Not sure, there are 2 lists where the tokens are removed from so there would be two Boleen values. Also, Remove account does not do this and this code is based on that logic.

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
trwalke: Oh yeah, missed this

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: I extracted the token removal logic to from the method to reuse the logic for the new `RemoveOboTokensInternal` method. If I didnt do this, there would be a bunch of duplicate code. There are minor differences between the two methods because the filtering logic is slightly different. calling the new remove obo api will not call `RemoveAccountInternal` so no duplicate work is done.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: This same value is present in the remove account method below and it looks like it assumes there will be a change in state. not sure if entirely correct actually, this notification args are being senf in the befrore access methods so there should be a state change yet, but I was confused as to why the remove account logic is doing this. Is this a bug in the remove account method?

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: This is the original logic for the remove account method and I am assuming that if one the removal methods fail, then we assume the cache is not formatted properly. Otherwise, we should try to remove everything we can from the cache and if there comes a point where something is missing, the developer can always reacquire a new set of tokens.

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Extensibility methods for <see cref="IConfidentialClientApplication"/>
+    /// </summary>
+    public static class ConfidentialClientApplicationExtensions
+    {
+        /// <summary>
+        /// Stops an in progress long running OBO session by removing the tokens associated with the provided cache key.
+        /// </summary>
+        /// <param name="clientApp">Client app to remove tokens from</param>
+        /// <param name="longRunningProcessSessionKey">OBO cache key used to remove the tokens</param>
```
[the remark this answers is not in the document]
trwalke: Good idea, I will update the docs for this

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -166,6 +169,43 @@ public sealed partial class ConfidentialClientApplication
...
         }
 
+        /// <summary>
+        /// Stops an in progress long running OBO session by removing the tokens associated with the provided cache key.
+        /// See https://aka.ms/msal-net-on-behalf-of.
+        /// </summary>
+        /// <param name="longRunningProcessSessionKey">OBO cache key used to remove the tokens</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="longRunningProcessSessionKey"/> is not set.</exception>
```
jmprieur, line 179: I would add seealso for the 2 APIs that use the long running process session.

jmprieur (approved):
LGTM
Thanks @trwalke

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1232,6 +1307,23 @@ bool ITokenCacheInternal.HasTokensNoLocks()
             return Accessor.HasAccessOrRefreshTokens();
         }
 
+        /// <summary>
+        /// Removes obo tokens stored in the cache. Note that the cache is internally and externally partitioned by the oboKey.
+        /// </summary>
+        internal /* internal for test only */ void RemoveOboTokensInternal(string oboPartitionKey, RequestContext requestContext)
```
bgavrilMS, line 1313: Doesn't seem to be used in tests.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -791,6 +792,82 @@ public async Task AcquireTokenByObo_NormalOboThenLongRunningInitiate_WithTheSame
...
+                UserAssertion userAssertion = new UserAssertion(userToken);
+
+                // InitiateLR - Empty cache - AT via OBO flow (new AT, RT cached)
+                var result = await oboCca.InitiateLongRunningProcessInWebApi(TestConstants.s_scope, userToken, ref oboCacheKey)
+                                        .ExecuteAsync().ConfigureAwait(false);
+
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+                Assert.AreEqual(1, cca.UserTokenCacheInternal.Accessor.GetAllAccessTokens().Count);
+                Assert.AreEqual(1, cca.UserTokenCacheInternal.Accessor.GetAllRefreshTokens().Count);
+                MsalAccessTokenCacheItem cachedAccessToken = cca.UserTokenCacheInternal.Accessor.GetAllAccessTokens().First(t => t.OboCacheKey.Equals(oboCacheKey));
```
bgavrilMS, line 852: Use `Single` instead of `First`

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -791,6 +792,82 @@ public async Task AcquireTokenByObo_NormalOboThenLongRunningInitiate_WithTheSame
             }
         }
 
+        /// <summary>
+        /// This test performs an initiation of a long running obo process, then validated that the remove long running process api deletes the tokens
+        /// in the cache associated with the provided obo cache key.
+        /// </summary>
+        [TestMethod]
+        public async Task AcquireTokenByObo_LongRunningThenRemoveTokens_TestAsync()
```
bgavrilMS, line 800: This test is failing in several places.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -791,6 +792,82 @@ public async Task AcquireTokenByObo_NormalOboThenLongRunningInitiate_WithTheSame
...
+
+                var cca = BuildCCA(httpManager);
+                var oboCca = cca as ILongRunningWebApi;
+
+                cca.UserTokenCache.SetBeforeAccess((args) => 
+                {
+                    if (validateCacheIsEmpty)
+                    {
+                        Assert.AreEqual(true, args.HasTokens);
+                        Assert.AreEqual(true, args.HasStateChanged);
```
bgavrilMS, line 817: before access, the state will have not changed.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -791,6 +792,82 @@ public async Task AcquireTokenByObo_NormalOboThenLongRunningInitiate_WithTheSame
...
+                        Assert.AreEqual(oboCacheKey, args.SuggestedCacheKey);
+                    }
+                });
+
+                cca.UserTokenCache.SetAfterAccess((args) =>
+                {
+                    if (validateCacheIsEmpty)
+                    {
+                        Assert.AreEqual(true, args.HasStateChanged);
+                        Assert.AreEqual(false, args.HasTokens);
```
bgavrilMS, line 827: after RemoveOBO, the state will have changed.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1232,6 +1307,23 @@ bool ITokenCacheInternal.HasTokensNoLocks()
...
+        {
+            ILoggerAdapter logger = requestContext.Logger;
+
+            //Filter and remove tokens based on OBO Cache Key
+            var refreshTokens = Accessor.GetAllRefreshTokens(oboPartitionKey);
```
bgavrilMS, line 1318: pls use the overload with logger, as you have a logger.

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
bgavrilMS: If any token is removed, return true.

bgavrilMS (changes requested):
Test needs to be fixed, otherwise ok. I would return Task<bool>

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: the internal method is used but it makes sense to use the public method as well

src/client/Microsoft.Identity.Client/ConfidentialClientApplication.cs
```diff
@@ -166,6 +169,45 @@ public sealed partial class ConfidentialClientApplication
                 longRunningProcessSessionKey);
         }
 
+        /// <summary>
+        /// Stops an in progress long running OBO session by removing the tokens associated with the provided cache key.
+        /// See https://aka.ms/msal-net-on-behalf-of.
```
pmaytak, line 174:
```suggestion
        /// See https://aka.ms/msal-net-long-running-obo.
```

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1146,15 +1147,92 @@ MsalIdTokenCacheItem ITokenCacheInternal.GetIdTokenCacheItem(MsalAccessTokenCach
...
+        {
+            bool tokensRemoved = false;
+
+            requestParameters.RequestContext.Logger.Verbose(() => $"[StopLongRunningOboProcessAsync] Entering token cache semaphore. Count {_semaphoreSlim.GetCurrentCountLogMessage()}");
+            await _semaphoreSlim.WaitAsync(requestParameters.RequestContext.UserCancellationToken).ConfigureAwait(false);
+            requestParameters.RequestContext.Logger.Verbose(() => "[StopLongRunningOboProcessAsync] Entered token cache semaphore");
+
+            try
+            {
+                requestParameters.RequestContext.Logger.Info("[StopLongRunningOboProcessAsync] Stopping long running OBO process by removing tokens from cache.");
```
gladjohn, line 1160:
```suggestion
                requestParameters.RequestContext.Logger.Info(() => "[StopLongRunningOboProcessAsync] Stopping long running OBO process by removing tokens from cache.");
```

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -1232,6 +1311,25 @@ bool ITokenCacheInternal.HasTokensNoLocks()
             return Accessor.HasAccessOrRefreshTokens();
         }
 
+        /// <summary>
+        /// Removes obo tokens stored in the cache. Note that the cache is internally and externally partitioned by the oboKey.
```
gladjohn, line 1315:
```suggestion
        /// Removes OBO tokens stored in the cache. Note that the cache is internally and externally partitioned by the oboKey.
```

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Stops an in progress long running OBO session by removing the tokens associated with the provided cache key.
+        /// See https://aka.ms/msal-net-on-behalf-of.
+        /// </summary>
+        /// <param name="clientApp">Client app to remove tokens from</param>
+        /// <param name="longRunningProcessSessionKey">OBO cache key used to remove the tokens</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>Returns true if tokens are removed from the cache. False otherwise.</returns>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="longRunningProcessSessionKey"/> is not set.</exception>
+        public static async Task<bool> StopLongRunningWebApiAsync(this ILongRunningWebApi clientApp, string longRunningProcessSessionKey, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
trwalke: Good Idea, I can submit another PR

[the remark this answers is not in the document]
pmaytak: Merged as part of #3998.

PR #3984, opened 2023-03-01
+Legacy in support methods New broker in net6 app

SameerK-MSFT:
Fixes #
Added New broker to .net6

Sample App to update .net 6

**Testing**
manual

**Performance impact**
none

**Documentation**


src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -188,12 +188,14 @@
...
     <Compile Include="$(PathToMsalSources)\Platforms\net6win\**\*.cs" LinkBase="Platforms\net6win" />
     <Compile Include="$(PathToMsalSources)\Platforms\Features\DefaultOSBrowser\**\*.cs" />
-    <Compile Include="$(PathToMsalSources)\Platforms\Features\WamBroker\**\*.cs" />
+    <Compile Include="$(PathToMsalSources)\Platforms\Features\RuntimeBroker\**\*.cs" />
     <Compile Include="$(PathToMsalSources)\Platforms\Features\WebView2WebUi\**\*.cs" />
     <Compile Include="$(PathToMsalSources)\Platforms\Features\DesktopOS\**\*.cs" />
     <Compile Include="$(PathToMsalSources)\Platforms\Features\WinFormsLegacyWebUi\**\*.cs" />
 
     <PackageReference Include="Microsoft.Web.WebView2" Version="1.0.864.35" />
+    <!-- TODO use single variable for the interop package version -->
```
bgavrilMS, line 197: needs to be addressed.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -319,4 +313,8 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="6.22.0" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Platforms\Features\RuntimeBroker\" />
```
bgavrilMS, line 318: Needed?

bgavrilMS (commented):
LGTM, some small comments + tests need resolution.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
SameerK-MSFT: Good catch. Done

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
SameerK-MSFT: Removed.

gladjohn:
should we removed this? [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/WindowsNativeUtils.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/WindowsNativeUtils.cs)

gladjohn:
Running "devapps\WAM\NetFrameworkWam\bin\Debug" and trying to get Work or School accounts shows an error and no accounts are returned 

image

gladjohn:
For existing broker flows that calls into the new broker without a window handle we show this message 

image

should this be re-worded or better handled as this would be a new requirement for the Desktop package?

SameerK-MSFT:
> should we removed this? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/WindowsNativeUtils.cs

This is public API. Not sure, if it used in the wild. Let's leave it alone.

SameerK-MSFT:
> Running "devapps\WAM\NetFrameworkWam\bin\Debug" and trying to get Work or School accounts shows an error and no accounts are returned

This happens as GetTenantProfiles() returns null. There is provision to accommodate null in getting count; but no in the select. Two questions:
1. Did it work earlier?
2. Is it normal to have null for GetTenantProfiles()?

tests/devapps/NetFxConsoleTestApp/Program.cs
```diff
@@ -134,8 +135,9 @@ private static IPublicClientApplication CreatePca()
                 Func<IntPtr> consoleWindowHandleProvider = () => consoleWindowHandle;
 
                 builder = builder
-                    //.WithParentActivityOrWindow(consoleWindowHandleProvider)
-                    .WithExperimentalFeatures()
+                    .WithParentActivityOrWindow(consoleWindowHandleProvider)
+                    // .WithExperimentalFeatures()
```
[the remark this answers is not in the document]
SameerK-MSFT: Good catch

tests/devapps/WAM/MSIX/WPF.Package/WPF.Package.wapproj
```diff
@@ -74,7 +74,7 @@
   </ItemGroup>
   <Import Project="$(WapProjPath)\Microsoft.DesktopBridge.targets" />
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.13.5" />
+    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="$(NativeInteropVersion)" IncludeAssets="all" />
```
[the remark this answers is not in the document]
SameerK-MSFT: We specify version number everywhere. It is unification of version number. Why should it be removed?

bgavrilMS:
> > should we removed this? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/WindowsNativeUtils.cs
> 
> This is public API. Not sure, if it used in the wild. Let's leave it alone.

Please hide it and deprecate it. With new WAM, it should not be needed.

bgavrilMS:
> For existing broker flows that calls into the new broker without a window handle we show this message
> 
> image
> 
> should this be re-worded or better handled as this would be a new requirement for the Desktop package?

Please suggest better wording. I think the discussions around window handle are still happening. We should consider this an incremental improvement and not GA blocking. We can always go from strict to less strict.

bgavrilMS:
> > Running "devapps\WAM\NetFrameworkWam\bin\Debug" and trying to get Work or School accounts shows an error and no accounts are returned
> 
> This happens as GetTenantProfiles() returns null. There is provision to accommodate null in getting count; but no in the select. Two questions:
> 
> 1. Did it work earlier?
> 2. Is it normal to have null for GetTenantProfiles()?

What exactly are the repro steps for this? `GetTenantProfiles` returns a `TenantProfile` for each Id token in the cache associated with that account.

SameerK-MSFT:
> What exactly are the repro steps for this?

- Run NetDesktopWinFormsWAM app
- Select WAMRuntime
- Select List OS Accounts
- Click on ATI
- After token is obtained, eventually it runs `RefreshAccountsAsync()`.

GetTenantProfile returns null in it.

SameerK-MSFT:
> Please hide it and deprecate it. With new WAM, it should not be needed.

Actually, we are using it in one of our test apps. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4434d1d0c8fe03aad7ca7b3d5ac60239da9fd4f3/tests/devapps/WAM/NetCoreWinFormsWam/Program.cs#L19

SameerK-MSFT:
> Please suggest better wording. I think the discussions around window handle are still happening. We should consider this an incremental improvement and not GA blocking. We can always go from strict to less strict.

Changed the wording slightly. "Desktop applications now need to provide parent window handle to the broker. See https://aka.ms/msal-net-wam#parent-window-handles". More suggestions are more than welcome.

gladjohn:
> > Please hide it and deprecate it. With new WAM, it should not be needed.
> 
> Actually, we are using it in one of our test apps.
> 
> https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4434d1d0c8fe03aad7ca7b3d5ac60239da9fd4f3/tests/devapps/WAM/NetCoreWinFormsWam/Program.cs#L19

because the test app was used to test this code with the old broker. We should remove the test app code

pmaytak:
> > > Please hide it and deprecate it. With new WAM, it should not be needed.
> > 
> > 
> > Actually, we are using it in one of our test apps.
> > https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4434d1d0c8fe03aad7ca7b3d5ac60239da9fd4f3/tests/devapps/WAM/NetCoreWinFormsWam/Program.cs#L19
> 
> because the test app was used to test this code with the old broker. We should remove the test app code

It's a workaround for legacy WAM to run in admin apps. Since UWP can't be run as admin, this code is obsolete.

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,124 @@
...
+        public bool ListOperatingSystemAccounts { get; set; }
+
+        /// <summary>
+        /// This is to validate the options
+        /// </summary>
+        internal void Validate()
+        { 
+            if(OSChoices == OperatingSystems.None)
+            {
+                throw new InvalidOperationException($"OS choice must be set.");
```
pmaytak, line 120:
```suggestion
                throw new InvalidOperationException($"Operating system must be specified.");
```

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,124 @@
...
+            return ret;
+        }
+
+        /// <summary>
+        /// Creates BrokerOptions from WindowsBrokerOptions
+        /// </summary>
+        /// <param name="winOptions"></param>
+        /// <param name="osChoice"></param>
+        /// <returns></returns>
+        public static BrokerOptions CreateFromWindowsOptions(WindowsBrokerOptions winOptions, OperatingSystems osChoice = OperatingSystems.Windows)
```
pmaytak, line 79: Should be internal? Users should just create their own BrokerOptions instance, the properties are basically the same.

```suggestion
        internal static BrokerOptions CreateFromWindowsOptions(WindowsBrokerOptions winOptions, OperatingSystems osChoice = OperatingSystems.Windows)
```

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            if (osChoices == OperatingSystems.None)
+            {
+                throw new ArgumentException($"osChoices should not be none");
+            }
+
+            OSChoices = osChoices;
+            Title = title;
+            MsaPassthrough = msaPassthrough;
+        }
```
pmaytak, line 48: I think this constructor should just accept OS enum and properties should be set by themselves because:
- If add new properties later, we'd have to add them to the signature every time.
- Updating the method signature with optional params is a binary breaking change and if we make params required, we'll end up with many overloads.
- The properties are public anyway.

I think this is more simplistic use, users can set options they care about.
```suggestion
        /// <summary>
        /// Constructor
        /// </summary>
        /// <param name="operatingSystems">Choices of OperatingSystems</param>
        public BrokerOptions(OperatingSystems operatingSystems)
        {
            if (operatingSystems == OperatingSystems.None)
            {
                throw new ArgumentException($"Operating system must be specified.");
            }
        }
```

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,124 @@
...
+        public static BrokerOptions CreateDefault(OperatingSystems osChoice = OperatingSystems.Windows, bool listWorkAndSchoolAccts = true)
+        {
+            BrokerOptions ret = new BrokerOptions(osChoice);
+            var winBrokerDefaultOptions = WindowsBrokerOptions.CreateDefault();
+            ret.Title = winBrokerDefaultOptions.HeaderText;
+            ret.MsaPassthrough = winBrokerDefaultOptions.MsaPassthrough;
+            ret.ListOperatingSystemAccounts = listWorkAndSchoolAccts;
+            
+            return ret;
+        }
```
pmaytak, line 71: Why is Windows OS default? I guess it works now, since it's the only broker we support, but if we support more OSs later, then would all of them be enabled by default?

Default options shouldn't accept user parameters because in that case they wouldn't be default.
```suggestion
        public static BrokerOptions CreateDefault()
        {
            return new BrokerOptions(osChoice)
            {
                 OSChoice = OperatingSystems.Windows,
                 ListOperatingSystemAccounts = true,
            };
        }
```

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -139,6 +140,7 @@ public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keycha
         }
 
         /// <summary>
+        /// For new applications, it is recommended to use the new WithBroker that accepts BrokerOptions as parameter
```
pmaytak, line 143: Wouldn't it throw?

`WithBroker(bool)` and `WithBroker(BrokerOptions)` only works for UWP and Net6-Win. For all other targets, it will throw saying to use Msal.Broker package. Plus `WithBroker(BrokerOptions)` is not really applicable to UWP now?

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -193,6 +195,27 @@ public PublicClientApplicationBuilder WithBroker(bool enableBroker = true)
...
+        /// 
+        /// On iOS and Android, Authenticator and Company Portal serve as brokers.
+        /// On Windows, WAM (Windows Account Manager) serves as broker. See https://aka.ms/msal-net-wam
+        /// </summary>
+        /// <param name="brokerOptions">This provides cross platform options for broker.</param>
+        /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
+        /// parameters, and to create a public client application instance</returns>
+        /// <remarks>If your app uses .NET classic or .NET Core 3.x, and you wish to use the Windows broker, 
+        /// please install the NuGet package Microsoft.Identity.Client.Desktop and call .WithDesktopFeatures()</remarks>
+        public PublicClientApplicationBuilder WithBroker(BrokerOptions brokerOptions)
```
pmaytak, line 212: I'm not sure I understand the why this method is needed. If it's called on anything other than UWP and Net6_Win, it will throw.

If I want to specify options, is this the proper calling pattern then:
```csharp
builder.WithBroker(someOptions).WithWindowsBroker()
```
Or
```csharp
builder.WithWindowsBrokerOptions(options).WithWindowsBroker();
```

BrokerOptions also must have OS specified, this means two places where I have to specify OS.

I was under the impression that we would have either:
- a specific method for each OS - `WithWindowsBroker(WindowsBrokerOptions)` or
- a generic one - `WithBroker(brokerOptions)` - user specifies the OS here.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -140,10 +140,11 @@ private enum ResponseStatus
         /// <param name="logger"></param>
         public static NativeInterop.AuthParameters GetCommonAuthParameters(
             AuthenticationRequestParameters authenticationRequestParameters, 
-            WindowsBrokerOptions brokerOptions,
+            BrokerOptions brokerOptions,
             ILoggerAdapter logger)
         {
-            logger.Verbose(() => "[WamBroker] Validating Common Auth Parameters.");
+            logger.Verbose(() => "[RuntimeBroker] Validating Common Auth Parameters.");
+            brokerOptions?.Validate();
```
pmaytak, line 147: I would move this validation to RuntimeBroker constructor. This way we validate once when PCA is build and RuntimeBroker is instantiated. As-is the validation is postponed until any of the AcquireToken methods are called.

Also I think `brokerOptions` is never null.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
```diff
@@ -1,1575 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
pmaytak, line 1: Since we still use old WAM on UWP, any way to keep these tests?

tests/Microsoft.Identity.Test.Unit/pop/PoPTests.cs
```diff
@@ -154,11 +156,15 @@ public async Task POP_WithMissingNonceForPCA_Async()
...
-                                                              .WithBrokerPreview()
-                                                              .WithHttpManager(httpManager)
-                                                              .BuildConcrete();
+                
+                var pcaBuilder = PublicClientApplicationBuilder.Create(TestConstants.ClientId)
+                                                              .WithHttpManager(httpManager);
+#if NET6_WIN
+                pcaBuilder = pcaBuilder.WithBroker(true);
+#else
+                pcaBuilder = Client.Broker.BrokerExtension.WithWindowsBroker(pcaBuilder);
```
pmaytak, line 165: I would invoke the method via the builder, since that's how it will be used and it still tests the extension method.

pmaytak (changes requested):

src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
```diff
@@ -51,13 +51,15 @@
     <DefineConstants>$(DefineConstants);NET_CORE</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="..\Microsoft.Identity.Client\Platforms\Features\RuntimeBroker\RuntimeBroker.cs" Link="RuntimeBroker.cs" />
```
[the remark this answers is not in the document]
bgavrilMS: It has to be non-preview.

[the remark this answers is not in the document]
bgavrilMS: I will change this.

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -23,6 +25,25 @@ public static class BrokerExtension
         /// If a broker does not exist or cannot be used, MSAL will fallback to a browser.
         /// Make sure browser auth is enabled (e.g. if using system browser, register the "http://localhost" redirect URI, etc.)
         /// </remarks>
+        public static PublicClientApplicationBuilder WithWindowsBroker(this PublicClientApplicationBuilder builder, bool enableBroker = true)
```
bgavrilMS, line 28: Why do we still have this API? I thought we settled on the WithBroker(Options)  API?

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        }
+
+        /// <summary>
+        /// Creates default options that can be modified later except the choice of OS
+        /// </summary>
+        /// <param name="osChoice">Choice of OS platforms</param>
+        /// <param name="listWorkAndSchoolAccts">List wokr and school accounts</param>
+        /// <returns></returns>
+        public static BrokerOptions CreateDefault()
```
bgavrilMS, line 56: Not a fan of this API. Making it internal.

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// </summary>
+        /// <remarks>On UWP, accounts are not listed due to privacy concerns</remarks>/// 
+        public bool ListOperatingSystemAccounts { get; set; }
+
+        /// <summary>
+        /// This is to validate the options
+        /// </summary>
+        internal void Validate()
+        { 
+            if(OSChoices == OperatingSystems.None)
```
bgavrilMS, line 110: Why would we fail on this? It simply means broker is not enabled.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -60,6 +61,8 @@ public string ClientVersion
 
         public WindowsBrokerOptions WindowsBrokerOptions { get; set; }
```
bgavrilMS, line 62: We should not keep these around any longer. Internally we should only keep `BrokerOptions `

bgavrilMS (changes requested):
There are too many unresolved comments and issues to close on this. 

In particular, the new API which we closed on Thursday is not implemented. `WithWindowsBroker` is still present and needs to be deelted. 

`WithWindowsBrokerOptions` needs to be soft deprecated in favour of the new `WithBrokerOptions`.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -60,6 +61,8 @@ public string ClientVersion
 
         public WindowsBrokerOptions WindowsBrokerOptions { get; set; }
 
+        public BrokerOptions BrokerOptions { get; set; }
```
bgavrilMS, line 64: Can we inject the options straight into the RuntimeBroker class? And maybe get rid of `WindowsBrokerOptions` as well (since we should always transform them to `RuntimeBroker`)

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
SameerK-MSFT: Makes sense. Not needed as this is assigned in the constructor and set is private. Removed

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
SameerK-MSFT: I would soft deprecate WindowsBrokerOptions as it may be needed for backward compatibility i.e. not to break other apps completely.

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
SameerK-MSFT: This was used by pure net6.0 (not net6.0-windows) apps. If this is removed and moved to WithBroker, only WamAdapter will be left in the package. We can discuss.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
pmaytak: +1 on just deprecate. Plus it's still needed for UWP?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -59,7 +59,8 @@ public string ClientVersion
 
         public bool IsBrokerEnabled { get; internal set; }
 
-        public WindowsBrokerOptions WindowsBrokerOptions { get; set; }
+        // Legacy options for UWP. .NET broker options are in BrokerOptions
+        public WindowsBrokerOptions UwpBrokerOptions { get; set; }
```
bgavrilMS, line 63: TODO: try to get rid of it completely.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
SameerK-MSFT: You are right.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -296,14 +297,6 @@
     <PackageReference Include="Microsoft.CodeAnalysis.RulesetToEditorconfigConverter" Version="3.3.3" PrivateAssets="All" />
   </ItemGroup>
 
-  <ItemGroup>
```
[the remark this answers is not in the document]
SameerK-MSFT: Nope. Not needed. @bgavrilMS please confirm

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
SameerK-MSFT: We are going with  WithBroker(brokerOptions) - user specifies the OS here.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
SameerK-MSFT: It is made sure that broker options will be valid. So removed validate method.

tests/Microsoft.Identity.Test.Unit/BrokerTests/WamTests.cs
SameerK-MSFT: We do not work on it unless there is Sev1. It has not happened for a long time. So no point in keeping them for UWP.

tests/Microsoft.Identity.Test.Unit/pop/PoPTests.cs
SameerK-MSFT: Done.

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
SameerK-MSFT: Removed.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
SameerK-MSFT: Needed for legacy such as MsaPassthrough Header for UWP apps.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
SameerK-MSFT: These have been renamed to UwpBrokerOptions. They are kept for backward compatibility.

SameerK-MSFT:
> should we removed this? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Platforms/Features/DesktopOS/WindowsNativeUtils.cs
> Please hide it and deprecate it. With new WAM, it should not be needed.
> because the test app was used to test this code with the old broker. We should remove the test app code
> It's a workaround for legacy WAM to run in admin apps. Since UWP can't be run as admin, this code is obsolete.

Obsoleted and EditorBrowsableState.Never set.

Directory.Build.props
```diff
@@ -7,7 +7,12 @@
     <DelaySign>false</DelaySign>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <IsPackable>false</IsPackable>    
+    <IsPackable>false</IsPackable>
+    <!-- Version of the native interop package. -->
```
gladjohn, line 11:
```suggestion
    <!-- Version of the Microsoft.Identity.Client.NativeInterop package. -->
```

build/template-build-and-prep-automation.yaml
```diff
@@ -129,11 +129,11 @@ steps:
         $AppsToCheck = @('MSIX\WPF\bin\Release\', 
                          'MSIX\WPF.Package\bin\AnyCPU\Release\WPF\', 
                          'NetCoreWinFormsWam\bin\Release\net6.0-windows10.0.17763.0\', 
-                         'NetCoreWinFormsWam\bin\Release\netcoreapp3.1\',
```
gladjohn, line 132: so we do not need this (netcoreapp3.1) anymore?

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -23,10 +26,30 @@ public static class BrokerExtension
         /// If a broker does not exist or cannot be used, MSAL will fallback to a browser.
         /// Make sure browser auth is enabled (e.g. if using system browser, register the "http://localhost" redirect URI, etc.)
         /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Use WithBroker(BrokerOptions) instead. This API can be found in Microsoft.Identity.Client.Broker package. See https://aka.ms/msal-net-wam for details.", false)]
```
gladjohn, line 30: please have a work item to track doc updates after release

src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
```diff
@@ -13,9 +13,9 @@
 
   <PropertyGroup Label="NuGet and AssemblyInfo metadata">
     <!--This should be passed from the VSTS build-->
-    <MsalClientSemVer Condition="'$(MsalClientSemVer)' == ''">4.46.0-preview</MsalClientSemVer>
+    <MsalClientSemVer Condition="'$(MsalClientSemVer)' == ''">$(MsalInternalVersion)</MsalClientSemVer>
     <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
-    <Version>$(MsalClientSemVer)-preview</Version>
+    <Version>$(MsalClientSemVer)</Version>
```
gladjohn, line 18: YAY

Directory.Build.props
```diff
@@ -7,7 +7,12 @@
...
+    <NativeInteropVersion>0.13.6</NativeInteropVersion>
```
gladjohn, line 12:
```suggestion
    <MSALRuntimeNativeInteropVersion>0.13.6</MSALRuntimeNativeInteropVersion>
```

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -48,7 +49,7 @@
 
   <PropertyGroup Label="NuGet and AssemblyInfo metadata">
     <!--This should be passed from the VSTS build-->
-    <MsalClientSemVer Condition="'$(MsalClientSemVer)' == ''">4.46.0</MsalClientSemVer>
+    <MsalClientSemVer Condition="'$(MsalClientSemVer)' == ''">$(MsalInternalVersion)</MsalClientSemVer>
```
gladjohn, line 52: We need to add a check to see we are not setting a version higher than the NuGet released version

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -135,7 +135,7 @@ private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
             {
                 throw new MsalClientException(
                     "window_handle_required",
-                    "Desktop applications wanting to use the broker need to provide their window handle. See https://aka.ms/msal-net-wam#parent-window-handles");
+                    "Desktop applications now need to provide parent window handle to the broker. See https://aka.ms/msal-net-wam#parent-window-handles");
```
gladjohn, line 138:
```suggestion
                    "The new WAM implementation now needs application developers to provide parent window handle to the broker. See https://aka.ms/msal-net-wam#parent-window-handles");
```

something like this? Or can we get Den to write this up?

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -451,7 +451,7 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
             ICacheSessionManager cacheSessionManager,
             IInstanceDiscoveryManager instanceDiscoveryManager)
         {
-            if (!_wamOptions.ListWindowsWorkAndSchoolAccounts)
+            if (!_wamOptions.ListOperatingSystemAccounts)
```
gladjohn, line 454:
```suggestion
            if (!_wamOptions.ListWindowsWorkAndSchoolAccounts)
```

leave as is, if not every log needs to be changed. see the next few lines and you see two different ways of referencing. Why change this?

src/client/Microsoft.Identity.Client/Platforms/uap/WamBroker/AccountPicker.cs
```diff
@@ -0,0 +1,155 @@
...
+            _synchronizationContext = synchronizationContext;
+            _authority = authority;
+            _isMsaPassthrough = isMsaPassthrough;
+            _optionalHeaderText = optionalHeaderText;
+            _logger.Verbose(()=>"Is MSA passthrough? " + _isMsaPassthrough);
+        }
+
+        public async Task<WebAccountProvider> DetermineAccountInteractivelyAsync()
+        {
+            await ShowPicker_UWPAsync().ConfigureAwait(true);
```
gladjohn, line 44: UAP / UWP, we should stick on to one. But if this old code leave it as is

src/client/Microsoft.Identity.Client/Platforms/uap/WamBroker/WamBroker.cs
```diff
@@ -87,7 +72,7 @@ internal class WamBroker : IBroker
             _msaPassthroughHandler = msaPassthroughHandler ??
                 new MsaPassthroughHandler(_logger, _msaPlugin, _wamProxy, _parentHandle);
 
-            _wamOptions = appConfig.WindowsBrokerOptions ??
+            _wamOptions = appConfig.UwpBrokerOptions ??
```
gladjohn, line 75: UWP / UAP?

tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj
```diff
@@ -16,13 +16,14 @@
...
   <ItemGroup>
-    <ProjectReference Include="..\..\src\client\Microsoft.Identity.Client.Broker\Microsoft.Identity.Client.Broker.csproj" />
+    <ProjectReference Include="..\..\src\client\Microsoft.Identity.Client.Broker\Microsoft.Identity.Client.Broker.csproj" Condition="'$(TargetFramework)' != '$(TargetFrameworkNet6Win)'"/> 
     <ProjectReference Include="..\..\src\client\Microsoft.Identity.Client\Microsoft.Identity.Client.csproj">
       <Project>{3433eb33-114a-4db7-bc57-14f17f55da3c}</Project>
       <Name>Microsoft.Identity.Client</Name>
     </ProjectReference>
     <ProjectReference Include="..\Microsoft.Identity.Test.Common\Microsoft.Identity.Test.Common.csproj" />
     <PackageReference Include="Microsoft.AspNet.WebApi.Client" Version="5.2.9" />
+    <PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="$(NativeInteropVersion)" IncludeAssets="all" />
```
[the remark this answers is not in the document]
gladjohn: please try removing this and see if transitive works

build/template-build-and-prep-automation.yaml
SameerK-MSFT: They have been removed

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
SameerK-MSFT: I want to remove specific operating system from it. What if we add capability to another OS?

src/client/Microsoft.Identity.Client/Platforms/uap/WamBroker/AccountPicker.cs
SameerK-MSFT: Old code.

src/client/Microsoft.Identity.Client/Platforms/uap/WamBroker/WamBroker.cs
SameerK-MSFT: UWP is the latest name of Uap. So UWP is better. However, I would not invest time in changing from Uap to UWP.

src/client/Microsoft.Identity.Client/Platforms/uap/WamBroker/AadPlugin.cs
```diff
@@ -15,11 +15,8 @@
 using Windows.Security.Authentication.Web.Core;
 using Windows.Security.Credentials;
 
-namespace Microsoft.Identity.Client.Platforms.Features.WamBroker
+namespace Microsoft.Identity.Client.Platforms.uap.WamBroker
```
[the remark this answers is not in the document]
SameerK-MSFT: In other uap code, it is all small. Better to keep consistent.

tests/devapps/WAM/TestBrokerApp/TestBrokerApp.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworkNet48>net48</TargetFrameworkNet48>
+    <TargetFrameworkNet6>net6.0</TargetFrameworkNet6>
+    <TargetFrameworkNet6Win>net6.0-windows10.0.17763.0</TargetFrameworkNet6Win>
+    <!--<TargetFrameworkUap>uap10.0.17763</TargetFrameworkUap>-->
```
gladjohn, line 8: remove

tests/devapps/WAM/TestBrokerApp/PCATester.cs
```diff
@@ -0,0 +1,163 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client;
+
+#if NET48 // This for classic net and net core
```
gladjohn, line 10:
```suggestion
#if NET48 // This for classic net 
```

bgavrilMS (approved):

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
SameerK-MSFT: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4015

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
bgavrilMS: This is just a default version string used if the CI isn't injecting the version. Not sure it matters too much.

gladjohn: sounds good

tests/Microsoft.Identity.Test.Unit/pop/PoPTests.cs
```diff
@@ -154,11 +160,14 @@ public async Task POP_WithMissingNonceForPCA_Async()
...
-                                                              .WithHttpManager(httpManager)
-                                                              .BuildConcrete();
+                
+                var pcaBuilder = PublicClientApplicationBuilder.Create(TestConstants.ClientId)
+                                                              .WithHttpManager(httpManager);
+
+                pcaBuilder = pcaBuilder.WithBroker(
+                    new BrokerOptions(BrokerOptions.OperatingSystems.Windows));
+
+                PublicClientApplication app = pcaBuilder.BuildConcrete();
```
pmaytak, line 170: Can this class be cleaned up?
```suggestion         
                PublicClientApplication app = PublicClientApplicationBuilder.Create(TestConstants.ClientId)
                                                              .WithHttpManager(httpManager)
                                                              .WithBroker(
                    new BrokerOptions(BrokerOptions.OperatingSystems.Windows))
                                                              .BuildConcrete();
```

pmaytak (approved):

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+            BrokerOptions ret = new BrokerOptions(OperatingSystems.Windows);
+            ret.Title = winOptions.HeaderText;
+            ret.MsaPassthrough = winOptions.MsaPassthrough;
+            ret.ListOperatingSystemAccounts = winOptions.ListWindowsWorkAndSchoolAccounts;
+
+            return ret;
+        }
+
+        /// <summary>
+        /// On which operating systems the broker is enabled.
```
gladjohn, line 63:
```suggestion
        /// Operating systems on which broker is enabled. 
```

LibsAndSamples.sln
```diff
@@ -169,10 +167,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NetDesktopWpfWAM", "tests\d
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Identity.Test.Integration.NetFx", "tests\Microsoft.Identity.Test.Integration.netfx\Microsoft.Identity.Test.Integration.NetFx.csproj", "{743BEA88-1903-4AF0-A791-705774F4B99D}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Identity.Test.Integration.NetCore", "tests\Microsoft.Identity.Test.Integration.netcore\Microsoft.Identity.Test.Integration.NetCore.csproj", "{7494E7A2-955E-44F4-823D-0601CFD2C966}"
```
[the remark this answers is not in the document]
SameerK-MSFT: we are not supporting netcore3.1 anymore.

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// </summary>
+        public OperatingSystems EnabledOn { get; }
+
+        /// <summary>
+        /// Title of the broker window
+        /// </summary>
+        public string Title { get; set; }
+
+        /// <summary>
+        /// A legacy option available only to Microsoft applications. Should be avoided where possible.
```
gladjohn, line 73:
```suggestion
        /// A legacy option available only to Microsoft First-Party applications. Should be avoided where possible.
```

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// Support is experimental.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)] // 1p feature only, hide it from public API.
+        public bool MsaPassthrough { get; set; } = false;
+
+        /// <summary>
+        /// Currently only supported on Windows
+        /// Allow the Windows broker to list Work and School accounts as part of the <see cref="ClientApplicationBase.GetAccountsAsync()"/>
```
gladjohn, line 81:
```suggestion
        /// Allows the Windows broker to list Work and School accounts as part of the <see cref="ClientApplicationBase.GetAccountsAsync()"/>
```

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -139,6 +140,7 @@ public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keycha
         }
 
         /// <summary>
+        /// For new applications, it is recommended to use the new WithBroker that accepts BrokerOptions as parameter
         /// Brokers enable Single-Sign-On, device identification,
         /// and application identification verification. To enable one of these features,
         /// you need to set the WithBroker() parameters to true. See https://aka.ms/msal-net-brokers
```
gladjohn, line 146:
```suggestion
        /// you need to set the WithBroker() parameters to true. See https://aka.ms/msal-net-wam 

```

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -150,8 +152,9 @@ public PublicClientApplicationBuilder WithIosKeychainSecurityGroup(string keycha
         /// <param name="enableBroker">Determines whether or not to use broker with the default set to true.</param>
         /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
         /// parameters, and to create a public client application instance</returns>
-        /// <remarks>If your app uses .NET classic or .NET Core 3.x, and you wish to use the Windows broker, 
-        /// please install the NuGet package Microsoft.Identity.Client.Desktop and call .WithDesktopFeatures()</remarks>
+        /// <remarks>If your app uses .NET classic or .NET, and you wish to use the Windows broker, 
+        /// please install the NuGet package Microsoft.Identity.Client.Broker and call .WithBroker(BrokerOptions).
+        /// This is not needed if for MAUI apps or for apps targetting net6-windows or higher.</remarks>
```
gladjohn, line 157:
```suggestion
        /// This is not needed for MAUI apps or for apps targeting net6-windows or higher.</remarks>

```

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -166,45 +169,69 @@ public PublicClientApplicationBuilder WithBroker(bool enableBroker = true)
...
         }
 
         /// <summary>
-        /// Allows customization of the Windows 10 Broker experience
+        /// Allows customization of the Windows 10 Broker experience. 
         /// </summary>
-#if !SUPPORTS_BROKER || __MOBILE__
+#if !NET6_WIN && !WINDOWS_APP
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("This API has been replaced with WithBroker(BrokerOptions), which can be found in Microsoft.Identity.Client.Broker package. See https://aka.ms/msal-net-wam for details.", false)]
```
gladjohn, line 202:
```suggestion
        [Obsolete("This API has been replaced with .WithBroker(BrokerOptions), which can be found in Microsoft.Identity.Client.Broker package. See https://aka.ms/msal-net-wam for details.", false)]

```

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -166,45 +169,69 @@ public PublicClientApplicationBuilder WithBroker(bool enableBroker = true)
...
+#endif
+#if NET6_WIN
         [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("This API has been replaced with WithBroker(BrokerOptions). See https://aka.ms/msal-net-wam for details.", false)]
```
gladjohn, line 206:
```suggestion
        [Obsolete("This API has been replaced with .WithBroker(BrokerOptions). See https://aka.ms/msal-net-wam for details.", false)]

```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -178,7 +178,7 @@ private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
                             _logger?.WarningPii(
                                 $"[WamBroker] Could not find a WAM account for the selected user {authenticationRequestParameters.Account.Username}, error: {readAccountResult.Error}",
```
gladjohn, line 179:
```suggestion
                                $"[RuntimeBroker] Could not find a WAM account for the selected user {authenticationRequestParameters.Account.Username}, error: {readAccountResult.Error}",
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -178,7 +178,7 @@ private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
...
                                 $"[WamBroker] Could not find a WAM account for the selected user. Error: {readAccountResult.Error}");
```
gladjohn, line 180:
```suggestion
                                $"[RuntimeBroker] Could not find a WAM account for the selected user. Error: {readAccountResult.Error}");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -178,7 +178,7 @@ private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
...
-                            
+
                             _logger?.Info(
                                 $"[WamBroker] Calling SignInInteractivelyAsync this will show the account picker.");
```
gladjohn, line 183:
```suggestion
                                $"[RuntimeBroker] Calling SignInInteractivelyAsync this will show the account picker.");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -245,7 +245,7 @@ private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
             _logger?.Verbose(() => "[WamBroker] Signing in with the default user account.");
```
gladjohn, line 245:
```suggestion
            _logger?.Verbose(() => "[RuntimeBroker] Signing in with the default user account.");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -286,7 +286,7 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
             _logger?.Verbose(() => "[WamBroker] Acquiring token silently.");
```
gladjohn, line 286:
```suggestion
            _logger?.Verbose(() => "[RuntimeBroker] Acquiring token silently.");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -334,7 +334,7 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
             _logger?.Verbose(() => "[WamBroker] Acquiring token silently for default account.");
```
gladjohn, line 334:
```suggestion
            _logger?.Verbose(() => "[RuntimeBroker] Acquiring token silently for default account.");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -364,7 +364,7 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
             _logger?.Verbose(() => "[WamBroker] Acquiring token with Username Password flow.");
```
gladjohn, line 364:
```suggestion
            _logger?.Verbose(() => "[RuntimeBroker] Acquiring token with Username Password flow.");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -388,7 +388,7 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
         {
             using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
             Debug.Assert(s_lazyCore.Value != null, "Should not call this API if msal runtime init failed");
-            
+
             if (account == null)
             {
                 _logger?.Verbose(() => "[WamBroker] No valid account was passed to RemoveAccountAsync. ");
```
gladjohn, line 394:
```suggestion
                _logger?.Verbose(() => "[RuntimeBroker] No valid account was passed to RemoveAccountAsync. ");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -451,7 +451,7 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
             ICacheSessionManager cacheSessionManager,
             IInstanceDiscoveryManager instanceDiscoveryManager)
         {
-            if (!_wamOptions.ListWindowsWorkAndSchoolAccounts)
+            if (!_wamOptions.ListOperatingSystemAccounts)
             {
                 _logger.Info("[WamBroker] ListWindowsWorkAndSchoolAccounts option was not enabled.");
```
gladjohn, line 456:
```suggestion
                _logger.Info("[RuntimeBroker] ListWindowsWorkAndSchoolAccounts option was not enabled.");
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -515,9 +515,9 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
                         $" Error Code: {discoverAccountsResult.Error.ErrorCode} \n" +
                         $" Error Message: {discoverAccountsResult.Error.Context} \n" +
                         $" Internal Error Code: {discoverAccountsResult.Error.Tag.ToString(CultureInfo.InvariantCulture)} \n" +
-                        $" Telemetry Data: {discoverAccountsResult.TelemetryData } \n";
+                        $" Telemetry Data: {discoverAccountsResult.TelemetryData} \n";
 
-                    _logger.ErrorPii($"[WamBroker] {errorMessagePii}", 
+                    _logger.ErrorPii($"[WamBroker] {errorMessagePii}",
```
gladjohn, line 520:
```suggestion
                    _logger.ErrorPii($"[RuntimeBroker] {errorMessagePii}",
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -515,9 +515,9 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
...
                         $"[WamBroker] DiscoverAccounts Error. " +
```
gladjohn, line 521:
```suggestion
                        $"[RuntimeBroker] DiscoverAccounts Error. " +
```

gladjohn (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -184,24 +181,19 @@ public async Task WamUsernamePasswordRequestAsync_WithPiiAsync()
...
-            {
-                //TODO: See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3916
-                //this failure is occuring outside of MSAL and is being investigated.
-            }
+            MsalAssert.AssertAuthResult(result, TokenSource.Broker, labResponse.Lab.TenantId, expectedScopes);
+
+            await pca.RemoveAsync(account).ConfigureAwait(false);
+            // Assert the account is removed
+            accounts = await pca.GetAccountsAsync().ConfigureAwait(false);
+            Assert.IsNotNull(accounts);
```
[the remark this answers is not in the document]
SameerK-MSFT: Count fails on local machines and is not reliable. So changed it. cc: @gladjohn

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
SameerK-MSFT: Need not put . before the method name in the comment.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
SameerK-MSFT: I disagree with having '.' before the method name in the comment.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
SameerK-MSFT: Done at all the places. Good one.

LibsAndSamples.sln
```diff
@@ -229,15 +229,12 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ManagedIdentityAppVM", "tests\devapps\Managed Identity apps\ManagedIdentityAppVM\ManagedIdentityAppVM.csproj", "{46DF28CE-D680-4E9A-9871-95A1C42DD15D}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "MSIX", "MSIX", "{128D16A2-1B07-4600-A898-71B328375440}"
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MSIHelperService", "tests\devapps\Managed Identity apps\MSIHelperService\MSIHelperService.csproj", "{1E290CD6-58BA-4AD0-A0B5-02458CD3C6C4}"
```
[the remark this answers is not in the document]
SameerK-MSFT: Oops! Good catch

PR #3989, opened 2023-03-03
[DRAFT] Rough POC to assess the impact of ciamlogin.com endpoint

jennyf19:
Rough POC for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3990
cc: @jmprieur @rayluo @bgavrilMS

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityAdapter.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityAdapter.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Identity.Client
+{
+    internal class CiamAuthorityAdapter
+    {
+        public CiamAuthorityAdapter(string authority)
+        {
+            TransformAuthority(authority);
+        }
+
+        public CiamAuthorityAdapter(string instance, string tenant)
+        {
+            instance = instance.Trim().TrimEnd('/');
+            if (string.IsNullOrEmpty(tenant))
+            {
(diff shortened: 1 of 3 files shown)


bgavrilMS:
Yes this is fine as POC, but for the product we'll need to create a new Authority object similar to B2CAuthority, AADAuthority etc. (or refactor the whole lot). 

We can prioritize this, should be ~2-3 days of work, depending on whether tenant changes are possible, device code, S2S and ESTS-R are supported etc.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityAdapter.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+        public string TransformedAuthority { get; private set; }
+
+        public string TransformedTenant { get; private set; }
+
+        public string TransformedInstance { get; private set; }
+
+        public string TransformedMetadata { get; private set; } = "{\"api-version\": \"1.1\",\"metadata\": [{\"preferred_network\": \"login.windows.net\",\"preferred_cache\": \"login.windows.net\",\"aliases\": [\"login.windows.net\",\"login.ciamlogin.com\"]}]}";
```
bgavrilMS, line 36: Are `login.windows.net` and `login.ciamlogin.com` really the same thing?

jmprieur: for the moment, but this will change.

jmprieur: As of 3/24, login.window.net should be replaced by https://login.microsoftonline.com (this is the issuer for the v2.0 endpoint)

PR #4044, opened 2023-03-29
[WAM] Adding more dev apps to test msalruntime existence

gladjohn:
Adding more dev apps to test msalruntime existence

this PR adds the following dev apps 

- a class library calls into the new broker
- a net fwk project that consumes the class lib project and invokes broker
- a MAUI Windows App to test `runtimeidentifiers`


build/template-restore-build-libsandsamples.yaml
```diff
@@ -28,6 +28,14 @@ steps:
     custom: 'workload'
     arguments: 'restore .\src\client\Microsoft.Identity.Client\Microsoft.Identity.Client.csproj'
 
+- task: Bash@3
+  displayName: Install MAUI
```
pmaytak, line 32: This task adds ~6 minutes to the build time. Wondering if we can move these WAM devapps in to a separate sln, which we build and run these tests on?

It's ok for now to improve the tests. Just trying to be mindful each time we update the build pipeline to make sure we don't get into situation again when we have to refactor the pipeline since our builds take long time again.

build/template-restore-build-libsandsamples.yaml
```diff
@@ -28,6 +28,14 @@ steps:
...
+  inputs:
+    targetType: 'inline'
+    script: |
+      dotnet workload install maui --source https://aka.ms/dotnet6/nuget/index.json --source https://api.nuget.org/v3/index.json
+      dotnet workload install android ios maccatalyst tvos macos maui wasm-tools --source https://aka.ms/dotnet6/nuget/index.json --source https://api.nuget.org/v3/index.json
```
pmaytak, line 37: nit: we don't need to install all these workloads.

PR #4051, opened 2023-03-30
Add back netcore tests.

pmaytak:
Adds back running netcore3.1 unit and integration tests.
Fixed some broker tests to call the correct Msal Runtime broker API.

Created issue #4052 as a follow up improvement to this.


gladjohn:
good coverage now 

image

Issue #3990, opened 2023-03-06
[Feature Request] MSAL.NET should support CIAM authorities


jmprieur:
**Is your feature request related to a problem? Please describe.**
- login.ciamlogin.com/tenant
- tenant.ciamlogin,com

bgavrilMS:
Blocked waiting for more details.

PR #4054, opened 2023-03-31
Initial implementation of CIAM support

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3990

Enables support for Ciam
Adds temporary transformation of https://"TenantName".ciamlogin.com into
https://"TenantName".ciamlogin.com/"TenantName".onmicrosoft.com

**Testing**
Unit tests
Integration tests

**Performance impact**
N/A

**Documentation**


src/client/Microsoft.Identity.Client/Instance/CiamAuthority.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Instance
+{
+    internal class CiamAuthority : AadAuthority
+    {
+        internal CiamAuthority(AuthorityInfo authorityInfo) : 
+            base(authorityInfo)
+        {
+
```
trwalke, line 17: Not sure how to differentiate between CIAM and AAD authority yet. They seem to behave very similarly.  Need to investigate further.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    internal class CiamAuthorityHelper
+    {
+        public static readonly string CiamMetadata = "{\"api-version\": \"1.1\",\"metadata\": [{\"preferred_network\": \"login.windows.net\",\"preferred_cache\": \"login.windows.net\",\"aliases\": [\"login.windows.net\",\"login.ciamlogin.com\"]}]}";
```
bgavrilMS, line 14: Not sure it's needed, just disable it?

src/client/Microsoft.Identity.Client/Internal/ServiceBundle.cs
```diff
@@ -35,10 +36,12 @@ internal class ServiceBundle : IServiceBundle
 
             HttpTelemetryManager = new HttpTelemetryManager();
 
+            var ciamMetadata = GetCiamMetadata(config.Authority);
+
             InstanceDiscoveryManager = new InstanceDiscoveryManager(
                 HttpManager,
                 shouldClearCaches,
-                config.CustomInstanceDiscoveryMetadata,
+                ciamMetadata != null ? ciamMetadata : config.CustomInstanceDiscoveryMetadata,
```
bgavrilMS, line 44: why don't we just disable it?

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
trwalke: There is some logic somewhere in the instance discovery manager that prevents this from working without using this metadata. Also, this is part of jennys POC and so I added it here so that the behavior is the same. Seems like there is a discrepency somewhere? Lets sync on this

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -425,9 +432,20 @@ private static AuthorityType GetAuthorityType(string authority)
...
+                return AuthorityType.Ciam;
+            }
+
             return AuthorityType.Aad;
         }
 
+        private static bool isCiamAuthority(string authority)
+        {
+            var uri = new Uri(authority);
+            return uri.Host.Contains("ciamlogin.com");
```
jmprieur, line 446: contains? or ends with?

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
jmprieur, line 14: I'm surprised that we still have `login.windows.net`, and not `login.microsoftonline.com` (it's targetting the v2.0 endpoint). Did you look in OIDC metadata what the issuer will be?

src/client/Microsoft.Identity.Client/Instance/CiamAuthority.cs
jmprieur: What do you mean by differentiate? from which point of view?

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
jmprieur: did you try ? what is the error? can you be more specific, @trwalke ?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
```diff
@@ -61,6 +61,25 @@ public void TestInitialize()
         }
 
         #region Happy Path Tests
+        //[TestMethod]
```
jmprieur, line 64: you didn't move the unit tests on the CiamAuthorityHelper from Jenny's POC?

jmprieur (commented):
thanks @trwalke for sharing this early draft PR
I've added a few remarks and questions

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
trwalke: The error indicates that AAD could not find the account when I type in the user id at the /authorize endpoint. However, when I provide the metadata, the endpoint is able to find the account.

jmprieur:
@trwalke 
you also probably want to have components tests like this one?:

```csharp
static void Test()
        {
            AssertEqual("https://msidlabciam1.ciamlogin.com/", "https://msidlabciam1.ciamlogin.com/msidlabciam1.onmicrosoft.com");
            AssertEqual("https://msidlabciam1.ciamlogin.com/d57fb3d4-4b5a-4144-9328-9c1f7d58179d", "https://msidlabciam1.ciamlogin.com/d57fb3d4-4b5a-4144-9328-9c1f7d58179d");
            AssertEqual("https://msidlabciam1.ciamlogin.com/msidlabciam1.onmicrosoft.com", "https://msidlabciam1.ciamlogin.com/msidlabciam1.onmicrosoft.com");
            AssertEqual("https://msidlabciam1.ciamlogin.com/aDomain", "https://msidlabciam1.ciamlogin.com/aDomain");
        }

        private static void AssertEqual(string authority, string expectedAuthority)
        {
            string clientId = "b8e9d222-c4ee-414c-ac29-b0eff1f32400";
            string effectiveAuthority =
            PublicClientApplicationBuilder.Create(clientId)
                                                    .WithAuthority(authority)
                                                    .WithDefaultRedirectUri()
                                                    .Build()
                                                    .Authority;
            Assert.Equal<string>(expectedAuthority, effectiveAuthority);
        }
```

src/client/Microsoft.Identity.Client/Instance/CiamAuthority.cs
rayluo: FYI:

1. Currently I use "hostname ends with .ciamlogin.com" to differentiate CIAM from AAD.
2. They behave largely the same, except CIAM authority does not have instance discovery endpoint. So, I use `#1` to bypass `#2`.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
pmaytak: +1 to ends with.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
pmaytak: Is login.windows.net a valid alias for ciam anyway, since it's already an alias for login.microsoftonline.com.
If instance discovery is disabled, we would end up hard coding the metadata by host here:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/c8a9937c2b1e98d04553a0889c8760850f9303f4/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs#L210

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
rayluo: It should end with ".ciamlogin.com" (mind the dot at the beginning).

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
rayluo: Hmm, the [OIDC metadata](https://msidlabciam1.ciamlogin.com/msidlabciam1.onmicrosoft.com/v2.0/.well-known/openid-configuration?dc=ESTS-PUB-EUS-AZ1-FD000-TEST1&ciamhost=true) shows the issuer as `https://login.microsoftonline.com/tenant_guid/v2.0` (while the endpoints are `https://tenant.ciamlogin.com/tenant_guid/...`). Such a discrepancy does not affect MSAL, but is it a problem that we shall let ESTS folks know, @jmprieur ?

Regardless, I agree that we shall not need `login.windows.net` here.

src/client/Microsoft.Identity.Client/Instance/CiamAuthority.cs
trwalke: From the point of how MSAL should MSAL use the CIAM authority to generate the various endpoints (authorize, token, device code) as opposed to the AAD Authority. Right now the new `CiamAuthoirty` type is simple inheriting functionality from the `AddAuthority` and the tests seems to be working so I am not sure I need to add anything in the `CiamAuthoirty` class

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
trwalke: Adding the extra query params solved the issue. no longer need instance metadata. this is now removed.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/UsernamePasswordIntegrationTests.cs
trwalke: I migrated them over. They are working as intended.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -428,6 +441,12 @@ private static AuthorityType GetAuthorityType(string authority)
             return AuthorityType.Aad;
         }
 
+        private static bool isCiamAuthority(string authority)
+        {
+            var uri = new Uri(authority);
+            return uri.Host.EndsWith("ciamlogin.com");
```
jmprieur, line 447: @trwalke, as @rayluo said, this needs to be `EndsWith(".ciamlogin.com")`. (note the dot)

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        
+
+        private void TransformAuthority(Uri ciamAuthority)
+        {
+            string host = ciamAuthority.Host + ciamAuthority.AbsolutePath;
+            if (string.Equals(ciamAuthority.AbsolutePath, "/"))
+            {
+                string ciamTenant = host.Substring(0, host.IndexOf(".ciamlogin.com", StringComparison.OrdinalIgnoreCase));
+                TransformedInstance = "https://login.ciamlogin.com/";
```
jmprieur, line 48: @trwalke: this should not be `"https://login.ciamlogin.com/"` any longer (this was in the initial prototype before the endpoint is available), but it should be `$"https://{ciamTenant}.ciamlogin.com/"` now.

tests/Microsoft.Identity.Test.Unit/AppConfigTests/CiamAuthorityHelperTests.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.AppConfig;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.AppConfigTests
+{
+    [TestClass]
+    public class CiamAuthorityHelperTests
+    {
+        private readonly string _ciamInstance = "https://login.ciamlogin.com";
```
jmprieur, line 14:
```suggestion
        private readonly string _ciamInstance = "https://idgciamdemo.ciamlogin.com";
```

tests/Microsoft.Identity.Test.Unit/AppConfigTests/CiamAuthorityHelperTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        private readonly string _ciamTenantGuid = "5e156ef5-9bd2-480c-9de0-d8658f21d3f7";
+        private readonly string _ciamTenant = "idgciamdemo.onmicrosoft.com";
+
+        // Possible CIAM authorities:
+        // https://login.ciamlogin.com/idgciamdemo.onmicrosoft.com
```
jmprieur, line 19:
```suggestion
        // https://idgciamdemo.ciamlogin.com/idgciamdemo.onmicrosoft.com
```

tests/Microsoft.Identity.Test.Unit/AppConfigTests/CiamAuthorityHelperTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        // https://idgciamdemo.ciamlogin.com
+        // https://login.ciamlogin.com/5e156ef5-9bd2-480c-9de0-d8658f21d3f7
```
jmprieur, line 21:
```suggestion
        // https://idgciamdemo.ciamlogin.com/5e156ef5-9bd2-480c-9de0-d8658f21d3f7
```

tests/Microsoft.Identity.Test.Unit/AppConfigTests/CiamAuthorityHelperTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            Assert.AreEqual(_ciamInstance, ciamAuthorityHelper.TransformedInstance);
+            Assert.AreEqual(_ciamTenantGuid, ciamAuthorityHelper.TransformedTenant);
+        }
+
+        [TestMethod]
+        public void CiamAuthorityAdapater_WithCiamLoginTest()
+        {
+            // Arrange
+            string ciamAuthority = "https://idgciamdemo.ciamlogin.com/";
+            string ciamTransformedInstance = "https://login.ciamlogin.com/";
```
jmprieur, line 57: Same thing, the transformed authority should not be "login."

I took the time to express what the behavior should be in this PR comment:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054#issuecomment-1492754089

I don't think that my comment had "login" anywhere?

More below too VV

jmprieur (changes requested):
This is already better but:
- the transformed authority should have the ciamTenant name, not "login"
- the EndsWith parameter should start by a dot.

tests/Microsoft.Identity.Test.Unit/AppConfigTests/CiamAuthorityHelperTests.cs
trwalke: ok, I missed the login part. i was trying too hard to maintain the functionality of the POC.

src/client/Microsoft.Identity.Client/MsalError.cs
```diff
@@ -1141,5 +1141,10 @@ public static class MsalError
         /// Using combined flat storage, like a file, to store both app and user tokens is not supported. Use a partitioned token cache (for ex. distributed cache like Redis) or separate files for app and user token caches. See https://aka.ms/msal-net-token-cache-serialization .
         /// </summary>
         public const string CombinedUserAppCacheNotSupported = "combined_user_app_cache_not_supported";
+
+        /// <summary>
+        /// Setting the CIAM authority (ex. "{tenantName}.ciamlogin.com") at the request level is not supported. CIAM authoritys must be set during application creation.
```
gladjohn, line 1146:
```suggestion
        /// Setting the CIAM authority (ex. "{tenantName}.ciamlogin.com") at the request level is not supported. CIAM authority must be set during application creation.
```

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -457,5 +457,6 @@ public static string InvalidTokenProviderResponseValue(string invalidValueName)
         public const string ManagedIdentityUserAssignedNotConfigurableAtRuntime = "[Managed Identity] Service Fabric user assigned managed identity ClientId or ResourceId is not configurable at runtime.";
         public const string CombinedUserAppCacheNotSupported = "Using a combined flat storage, like a file, to store both app and user tokens is not supported. Use a partitioned token cache (for ex. distributed cache like Redis) or separate files for app and user token caches. See https://aka.ms/msal-net-token-cache-serialization .";
         public const string JsonParseErrorMessage = "There was an error parsing the response from the token endpoint, see inner exception for details. Verify that your app is configured correctly. If this is a B2C app, one possible cause is acquiring a token for Microsoft Graph, which is not supported. See https://aka.ms/msal-net-up";
+        public const string SetCiamAuthorityAtRequestLevelNotSupported = "Setting the CIAM authority (ex. \"{tenantName}.ciamlogin.com\") at the request level is not supported. CIAM authoritys must be set during application creation";
```
gladjohn, line 460:
```suggestion
        public const string SetCiamAuthorityAtRequestLevelNotSupported = "Setting the CIAM authority (ex. \"{tenantName}.ciamlogin.com\") at the request level is not supported. CIAM authority must be set during application creation";

```

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -82,6 +83,10 @@ internal class AuthorityInfo
                             pathSegments[0]),
                         authorityUri.Port);
                     break;
+                case AuthorityType.Ciam:
+                    CiamAuthorityHelper ciamAuthorityAdapter = new CiamAuthorityHelper(authorityUri);
+                    authorityUri = ciamAuthorityAdapter.TransformedAuthority;
+                    goto default;
```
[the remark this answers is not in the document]
pmaytak: I think the intention is to also do the same setup as for AAD? 
I would extract the code into a separate method and call it from AAD and Ciam cases.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -428,6 +441,12 @@ private static AuthorityType GetAuthorityType(string authority)
             return AuthorityType.Aad;
         }
 
+        private static bool isCiamAuthority(string authority)
```
pmaytak, line 444: `IsCiamAuthority`
I think this method is more appropriate in the CiamAuthority class.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// Translates CIAM authorities into a usable form. This is needed only until ESTS is updated to support the north star format
+    /// North star format: https://idgciamdemo.ciamlogin.com
+    /// Transformed format: https://idgciamdemo.ciamlogin.com/idgciamdemo.onmicrosoft.com
+    /// </summary>
+    internal class CiamAuthorityHelper
```
pmaytak, line 17: Can the logic in this class be moved to CiamAuthority class, better enacapsulation?

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -428,6 +441,12 @@ private static AuthorityType GetAuthorityType(string authority)
...
+        {
+            var uri = new Uri(authority);
+            return uri.Host.EndsWith(".ciamlogin.com");
```
jmprieur, line 447: Let's use a constant all over the code. The URL could change.
cc: @bgavrilMS

jmprieur:
Let's have one constant in the codebase for ".ciamlogin.com", to be robust to possible changes of the URL (before Build)

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
trwalke: The problem I ran into was the need to have the transformed authority in the `AuthorityInfo` class, which is needed for the constructor of the `CiamAuthority`. So you would end up with a catch 22 of sorts.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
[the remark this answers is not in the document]
trwalke: Yes, the intention was to fallthrough to the last case but since it is the default it is not supported. I can extract to a method

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
trwalke: same as this https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054#discussion_r1159251858

trwalke: I can take another look at this to see if there is another way to better organize this without too much refactoring. However, we are looking to release by Friday so we can prepare for MS Build so it may be a follow up PR later.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
trwalke: Yes, makes sense

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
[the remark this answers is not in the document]
trwalke: Actually, the properties being set are read only and cant be set in a method other than the constructor..

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
trwalke: Removed `login.windows.net`

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -84,6 +84,11 @@ protected virtual void Validate()
 
         internal void ValidateAndCalculateApiId()
         {
+            if (CommonParameters.AuthorityOverride?.AuthorityType == AuthorityType.Ciam)
```
pmaytak, line 87: Why do we need this? Adds complexity to the authority transform logic?

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
pmaytak: Ah yea, maybe just make it a static helper method - so it would be CiamAuthority.TransformAuthority or something.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Internal;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// Translates CIAM authorities into a usable form. This is needed only until ESTS is updated to support the north star format
```
pmaytak, line 14: Since the users can still pass in the transformed format, would we have to reverse this code then to transform into North Star format or in the future will AAD support both formats?

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+                TransformedTenant = tenant;
+                TransformedAuthority = new Uri(instance + '/' + tenant);
+            }
+        }
+
+        public Uri TransformedAuthority { get; private set; }
+
+        public string TransformedTenant { get; private set; }
+
+        public string TransformedInstance { get; private set; }
```
pmaytak, line 44: I think these two can be made private, since they aren't used anywhere else; and they will be tested transitively though tests on TransformedAuthority.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Transformed format: https://idgciamdemo.ciamlogin.com/idgciamdemo.onmicrosoft.com
+    /// </summary>
+    internal class CiamAuthorityHelper
+    {
+        public CiamAuthorityHelper(Uri authority)
+        {
+            TransformAuthority(authority);
+        }
+
+        public CiamAuthorityHelper(string instance, string tenant)
```
pmaytak, line 25: Remove this, since it's not used anywhere but tests? Just use the other constructor?

tests/Microsoft.Identity.Test.LabInfrastructure/LabServiceParameters.cs
```diff
@@ -14,7 +14,8 @@ public enum FederationProvider
         Shibboleth,
         ADFSv2019,
         B2C,
-        Ping
+        Ping,
+        CIAM
```
pmaytak, line 18: nit: Ciam

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/CiamUsernamePasswordIntegrationTests.cs
```diff
@@ -0,0 +1,88 @@
...
+                .Build();
+
+            var result = await msalPublicClient
+                .AcquireTokenByUsernamePassword(_ciamScopes, labResponse.User.Upn, labResponse.User.GetOrFetchPassword())
+                .WithExtraQueryParameters(_extraQParams)
+                .ExecuteAsync()
+                .ConfigureAwait(false);
+
+            Assert.IsNotNull(result.AccessToken);
+            Assert.AreEqual(result.AuthenticationResultMetadata.TokenSource, TokenSource.IdentityProvider);
```
pmaytak, line 72: switch order here and below:
```suggestion
            Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
```

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/CiamUsernamePasswordIntegrationTests.cs
```diff
@@ -0,0 +1,88 @@
...
+    public class CiamUsernamePasswordIntegrationTests
+    {
+        private const string _extraQParams = "dc=ESTS-PUB-EUS-AZ1-FD000-TEST1";
+        private readonly string[] _ciamScopes = new[] { "openid" };
+        private const string _ciamRedirectUri = "http://localhost";
+
+        [TestMethod]
+        [DataRow("https://{0}" + Constants.CiamAuthorityHostSuffix + "/", 0)]
+        [DataRow("https://{0}" + Constants.CiamAuthorityHostSuffix + "/{1}.onmicrosoft.com", 1)]
+        [DataRow("https://{0}" + Constants.CiamAuthorityHostSuffix + "/{1}", 2)]
```
pmaytak, line 29: Side note. There's a difference in the cache item keys generated for these different authorities. For 1 and 2, cache key uses a domain tenant; while for 3, it uses a GUID tenant. This is because the tenant for this is retrieved from the authority. The partition key is home account ID for all cases, since we use client info for that. This is also the same for B2C scenario. The only scenario which this can affect that I can think of is if the cache is shared between two app instances which use different tenant formats which would result in a cache miss for the users.

cc @bgavrilMS

pmaytak (approved):
Mostly just refactoring suggestions.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
trwalke: It is actually intended to reduce complexity in developer scenarios. Rather than allowing the CIAM authorities to be set at the request level and introduce bugs like we have seen in the past. We decided to only allow CIAM authorities to be set at the app level.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
trwalke: Not sure about that yet. That will depend on the decisions made by management. @jmprieur any thoughts on this?
I would imagine ESTS would allow the transformed versions ('https://tenant.ciamlogin.com/tenant.onmicrosoft.com`) to continue working going forward as to not break the customers who are using the version of MSAL at build.

src/client/Microsoft.Identity.Client/AppConfig/CiamAuthorityHelper.cs
trwalke: Yea, I was planning on using these at some point but I ended up not needing them.

tests/Microsoft.Identity.Test.LabInfrastructure/LabServiceParameters.cs
trwalke: It is mentioned in the docs as CIAM which I am assuming is an acronym for something. 
cc @bgavrilMS

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
bgavrilMS: I don't like this method because it actually does a lot of work and it can be abused. We should parse the string ONCE and only once.

src/client/Microsoft.Identity.Client/Instance/CiamAuthority.cs
```diff
@@ -0,0 +1,18 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Instance
+{
+    internal class CiamAuthority : AadAuthority
```
bgavrilMS, line 12: I would think the "CannonicalAuthority" is the "transformed" authority?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/CiamUsernamePasswordIntegrationTests.cs
bgavrilMS: Not sure we can do anything about it though, we don't have the GUID upfront in MSAL. It's the same for AAD if they choose to use friendy name instead of GUID.

tests/Microsoft.Identity.Test.LabInfrastructure/LabServiceParameters.cs
bgavrilMS: Ciam is fine.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -408,6 +416,11 @@ internal static string GetSecondPathSegment(Uri authority)
 
         private static AuthorityType GetAuthorityType(string authority)
         {
+            if (isCiamAuthority(authority))
```
bgavrilMS, line 419: This leads to double parsing of the string via the Uri constructor, which is expensive (this is on the critical path). I recommend revamping this logic a bit - instead of using `GetFirstPathSegment` do the following: 

1. `(host, firstPathSegment, secondPathSegment) = ParseAuthority(Uri)` - this ensures that we only parse this once. 
2. [Optional, but recommended] Ensure that internally we use exclusively `URI` for the authority. `new Uri(string)` is an expensive operation that needs to be avoided. Not sure how involved this is, seems doable, but the devil is in the details.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -346,7 +354,7 @@ private static void ValidateAuthorityUri(string authority, AuthorityType? author
             }
 
             string path = authorityUri.AbsolutePath.Substring(1);
-            if (string.IsNullOrWhiteSpace(path))
+            if (string.IsNullOrWhiteSpace(path) && !isCiamAuthority(authority))
```
bgavrilMS, line 357: Avoid reparsing this. I would re-write the entire logic, it's not that complex. The logic validation logic needs to avoid parsing at all costs. I would add it as part of `GetAuthorityType` after `ParseAuthority(string)`

bgavrilMS (changes requested):
Too much string parsing in the hot path.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
trwalke: I did a bit of refactoring to reduce the overall number of new Uri() calls. Please take a look and verify. There were a number of cases where we seem to be creating new Uris excessivly and it didnt make sense to me why some of them were there in the first place.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -145,34 +145,47 @@ internal class AuthorityInfo
...
+            }
+
+            return authorityUri;
+        }
+
+        internal static AuthorityInfo FromAadAuthority(string cloudInstanceUri, Guid tenantId, bool validateAuthority)
         {
 #pragma warning disable CA1305 // Specify IFormatProvider
-            return FromAuthorityUri(new Uri(cloudInstanceUri, tenantId.ToString("D")).AbsoluteUri, validateAuthority);
+            return FromAuthorityUri($"{cloudInstanceUri.Trim().TrimEnd('/')}/{tenantId.ToString("D")}", validateAuthority);
```
trwalke, line 178: This method is called by an obsolete api. probably not super impactful but I still changed it. Added a `.Trim().TrimEnd('/')` to avoid a double `/`

@bgavrilMS

src/client/Microsoft.Identity.Client/Instance/CiamAuthority.cs
```diff
@@ -0,0 +1,43 @@
...
+            string transformedInstance;
+            string transformedTenant;
+
+            string host = ciamAuthority.Host + ciamAuthority.AbsolutePath;
+            if (string.Equals(ciamAuthority.AbsolutePath, "/"))
+            {
+                string ciamTenant = host.Substring(0, host.IndexOf(Constants.CiamAuthorityHostSuffix, StringComparison.OrdinalIgnoreCase));
+                transformedInstance = $"https://{ciamTenant}{Constants.CiamAuthorityHostSuffix}/";
+                transformedTenant = ciamTenant + ".onmicrosoft.com";
+                return new Uri(transformedInstance + transformedTenant);
```
trwalke, line 35: Tried to remove the Uri parsing as much as I can but the logic ends up becoming a bit more complex if I remove it from everywhere. I did reduce the number of parses in the hotpash so there should still be a performance improvement overall.

trwalke (commented):
Not sure about this one. Seems like it is just recreating the same URI her. Thoughts on this one?

@bgavril

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/CiamIntegrationTests.cs
```diff
@@ -0,0 +1,156 @@
...
+            Assert.IsNotNull(result.AccessToken);
+            Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+            Assert.AreEqual($"{labResponse.User.LabName}{Constants.CiamAuthorityHostSuffix}".ToLower(), result.Account.Environment);
+        }
+
+        [TestMethod]
+        [DataRow("https://{0}.ciamlogin.com/", 0)] //https://tenantName.ciamlogin.com/
+        [DataRow("https://{0}.ciamlogin.com/{1}.onmicrosoft.com", 1)] //https://tenantName.ciamlogin.com/tenantName.onmicrosoft.com
+        [DataRow("https://{0}.ciamlogin.com/{1}", 2)] //https://tenantName.ciamlogin.com/tenantId
+        public async Task ClientCredentialWithClientSecret_Ciam_Async(string authorityFormat, int authorityVersion)
```
trwalke, line 95: I didnt see a certificate anywhere to use. I will Ask Ryan if it is available so we can also test with the cert.

bgavrilMS: You can also reach out to @siddhijain - she has managed to add E2E tests for client_credentials. 
Also, it seems that acquire token interactive is pretty simple to automate as well. 

But this can be done separately.

PR #4056, opened 2023-04-01
[WAM] Token protection validation with CA Policy

gladjohn:
Token protection (sometimes referred to as token binding in the industry) attempts to reduce attacks using token theft by ensuring a token is usable only from the intended device. When an attacker is able to steal a token, by hijacking or replay, they can impersonate their victim until the token expires or is revoked. Token theft is thought to be a relatively rare event, but the damage from it can be significant.

learn more : [https://learn.microsoft.com/en-us/azure/active-directory/conditional-access/concept-token-protection](https://learn.microsoft.com/en-us/azure/active-directory/conditional-access/concept-token-protection)

In msidlab4 lab, we now have a valid resource (SPO) to test for POP tokens and also a CA Policy for token issuance. AAD will only get a token back for the popUser when requested with a valid resource (i.e. https://msidlab4.sharepoint.com in this case) using the WAM Broker + WithProofOfPossession API.  

**CA Policy in msidlab4.com :**

image

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
@@ -265,6 +265,43 @@ namespace Microsoft.Identity.Test.Integration.Broker
 
             Assert.IsTrue(!string.IsNullOrEmpty(ex.ErrorCode));
         }
+
+        [RunOn(TargetFrameworks.NetStandard | TargetFrameworks.NetCore)]
+        [ExpectedException(typeof(MsalUiRequiredException))]
+        public async Task WamUsernamePasswordPopTokenAsync()
+        {
+            var labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
+            
+            string popUser = "popUser@msidlab4.onmicrosoft.com";
+
+            string[] scopes = { "https://msidlab4.sharepoint.com/user.read" };
+
+            string[] expectedScopes = { 
+                "https://msidlab4.sharepoint.com/Calendars.Read", 
+                "https://msidlab4.sharepoint.com/Sites.Read.All ", 
+                "https://msidlab4.sharepoint.com/UserTimelineActivity.Write.CreatedByApp"
+            };
+
+            IntPtr intPtr = GetForegroundWindow();
+
+            Func<IntPtr> windowHandleProvider = () => intPtr;
+
+            WamLoggerValidator testLogger = new WamLoggerValidator();
+
+            IPublicClientApplication pca = PublicClientApplicationBuilder
+               .Create(labResponse.App.AppId)
+               .WithParentActivityOrWindow(windowHandleProvider)
+               .WithAuthority(labResponse.Lab.Authority, "organizations")
+               .WithLogging(testLogger)
+               .WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows))
+               .Build();
+
+            // Acquire token using username password
+            var result = await pca.AcquireTokenByUsernamePassword(scopes, popUser, labResponse.User.GetOrFetchPassword()).ExecuteAsync().ConfigureAwait(false);
+
+            MsalAssert.AssertAuthResult(result, TokenSource.Broker, labResponse.Lab.TenantId, expectedScopes);
+        }
+
     }
 }
 #endif
--- a/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
@@ -296,13 +296,20 @@ namespace Microsoft.Identity.Test.Integration.Broker
                .WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows))
                .Build();
 
-            // Acquire token using username password
-            var result = await pca.AcquireTokenByUsernamePassword(scopes, popUser, labResponse.User.GetOrFetchPassword())
-                .WithProofOfPossession("some_nonce", System.Net.Http.HttpMethod.Get, new Uri(pca.Authority))
-                .ExecuteAsync()
-                .ConfigureAwait(false);
-
-            MsalAssert.AssertAuthResult(result, TokenSource.Broker, labResponse.Lab.TenantId, expectedScopes);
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -265,6 +264,70 @@ public async Task WamAddDefaultScopesWhenNoScopesArePassedAsync(string scopes)
 
             Assert.IsTrue(!string.IsNullOrEmpty(ex.ErrorCode));
         }
+
+        [RunOn(TargetFrameworks.NetStandard | TargetFrameworks.NetCore)]
+        public async Task WamUsernamePasswordPopTokenEnforcedWithCaOnValidResourceAsync()
```
bgavrilMS, line 269: CC @rayluo - this would show an E2E of using a POP token on an actual resource. 

I think the only difference between this and the North Star is that the nonce is not enforced (and from what I can tell it's not even published by SPO).

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -265,6 +264,70 @@ public async Task WamAddDefaultScopesWhenNoScopesArePassedAsync(string scopes)
...
+        {
+            //Arrange
+            var labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
+            
+            string popUser = "popUser@msidlab4.onmicrosoft.com";
+
+            string[] scopes = { "https://msidlab4.sharepoint.com/user.read" };
```
bgavrilMS, line 276: Are you able to actually make a web request to the actual SPO ? Am curios for: 

1. Does SPO issue server nonce?
2. Does SPO accept a POP token without a server nonce?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -265,6 +264,70 @@ public async Task WamAddDefaultScopesWhenNoScopesArePassedAsync(string scopes)
...
+
+            IPublicClientApplication pca = PublicClientApplicationBuilder
+               .Create(labResponse.App.AppId)
+               .WithParentActivityOrWindow(windowHandleProvider)
+               .WithAuthority(labResponse.Lab.Authority, "organizations")
+               .WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.Windows))
+               .Build();
+
+            // Acquire token using username password with POP on a resource not in the CA policy
+            // CA policy enforces token issuance to popUser only for SPO this call will fail with UI Required Exception
```
rayluo, line 324: Why would attempting a POP token request on a non-POP resource result in a UI Required error? Is that the way AAD tells the app developer "hey you do not need to use POP for this resource"? But even if that is the case, why our negative test case here would need to test that behavior?

I thought the negative test case should be to attempt a bearer token request for the same POP-protected resource, and expect an error, which in turn proves that our happy-path test case did the POP right.

bgavrilMS: Great question. My hypothesis is that you attempt non-interactive flows like ROPC or refresh_token, then AAD will ask you to switch to interactive just to show the end-user a friendly message like "You can't get from here to here" or "This version of the app doesn't meet the security requirements". An end user who uses VS can for example switch between browser and WAM and "fix" the CA policy themselves?

bgavrilMS: Agreed on the nagative test, that's a nice integration scenario.

rayluo: > Great question. My hypothesis is that you attempt non-interactive flows like ROPC or refresh_token, then AAD will ask you to switch to interactive just to show the end-user a friendly message like "You can't get from here to here" or "This version of the app doesn't meet the security requirements". An end user who uses VS can for example switch between browser and WAM and "fix" the CA policy themselves?

As of this writing, the test case here is *sending POP request to a non-POP resource* (MS Graph). Not sure whether ESTS would emit an error "this version of the app EXCEEDS the security requirements (i.e. you are too secure, so I'll block you)". Regardless, this is perhaps not the scenario we need to test here.

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
bgavrilMS: No, ESTS is not aware of RP capability to accept various type of tokens. RP is supposed to issue `WWW-Authenticate Bearer` and `WWW-Authenticate POP` for inform the client of what auth schemes it supports.

trwalke (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
```diff
@@ -265,6 +264,70 @@ public async Task WamAddDefaultScopesWhenNoScopesArePassedAsync(string scopes)
...
+            // https://learn.microsoft.com/en-us/azure/active-directory/conditional-access/concept-token-protection
+            var result = await pca.AcquireTokenByUsernamePassword(scopes, popUser, labResponse.User.GetOrFetchPassword())
+                .WithProofOfPossession("some_nonce", System.Net.Http.HttpMethod.Get, new Uri(pca.Authority))
+                .ExecuteAsync()
+                .ConfigureAwait(false);
```
pmaytak, line 329: Missing assert?

```csharp
            var ex = await AssertException.TaskThrowsAsync<MsalUiRequiredException>(
                 () => pca.AcquireTokenByUsernamePassword(scopes, popUser, labResponse.User.GetOrFetchPassword())
                .WithProofOfPossession("some_nonce", System.Net.Http.HttpMethod.Get, new Uri(pca.Authority))
                .ExecuteAsync())
            .ConfigureAwait(false);
```

gladjohn: It's on line 302

PR #4058, opened 2023-04-03
Fix test loading cert by name when multiple certs with same name exist

bgavrilMS:
Test only. 

On my machine, I have 4 certs with the same name. Current logic loads a random one.

### DIFF
--- a/tests/Microsoft.Identity.Test.LabInfrastructure/CertificateHelper.cs
+++ b/tests/Microsoft.Identity.Test.LabInfrastructure/CertificateHelper.cs
@@ -46,6 +46,7 @@ namespace Microsoft.Identity.Test.LabInfrastructure
                 X509Certificate2Collection collection = store.Certificates.Find(X509FindType.FindBySubjectName, certName, validateCerts);
 
                 X509Certificate2 certToUse = null;
+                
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Identity.Test.LabInfrastructure/CertificateHelper.cs
```diff
@@ -43,11 +43,19 @@ public static X509Certificate2 FindCertificateByName(string certName, StoreLocat
             using (var store = new X509Store(name, location))
             {
                 store.Open(OpenFlags.ReadOnly);
-                var collection = store.Certificates.Find(X509FindType.FindBySubjectName, certName, validateCerts);
+                X509Certificate2Collection collection = store.Certificates.Find(X509FindType.FindBySubjectName, certName, validateCerts);
 
-                return collection.Count == 0
-                    ? null
-                    : collection[0];
+                X509Certificate2 certToUse = null;
```
gladjohn, line 48:
```suggestion
                X509Certificate2 certToUse = null;
                
```

gladjohn (approved):
cool. I was having the same issue on my machine as well. Thanks for the fix!!!

pmaytak (approved):

PR #4068, opened 2023-04-11
Managed Identity - Calculate the refresh_in value based on expires_on

### LINKED ISSUE #4062: [Feature Request] Managed Identity - Calculate the refresh_in value based on expires_in
Managed Identity Endpoints do not give us `refresh_in`. The plan is to calculate this in MSAL 

A sample Managed Identity Token response from IMDS Endpoint is seen below, 

```JSON 
HTTP/1.1 200 OK
Content-Type: application/json
{
  "access_token": "eyJ0eXAi...",
  "refresh_token": "",
  "expires_in": "3599",
  "expires_on": "1506484173",
  "not_before": "1506480273",
  "resource": "https://management.azure.com/",
  "token_type": "Bearer"
}
```

Compute refresh_in as 1/2 expires_in, but only if expires_in > 2h.

```
if expires_in > 2h 
   refresh_in =  expires_in * 0.50 
else 
      refresh_in = null;
```

Where:

- expires_in: The lifetime of the access token in seconds. 
- ~buffer_time: The amount of time (in seconds) you want to subtract from the token lifetime to ensure that you refresh the token before it expires. This is typically a small value, such as 60 seconds, to avoid any chance of the token expiring before it can be refreshed.~
- refresh_in: The time (in seconds) after which you should refresh the access token to ensure that it remains valid.
The value of 0.50 in the formula is a recommended value for the refresh time. This ensures that the token is refreshed before it expires while also minimizing the number of unnecessary refreshes.

For example, if the token lifetime is 3600 seconds (1 hour) ~and the buffer time is 60 seconds~, then the refresh time would be:
```
expires_in = 7200 -> refresh_in  = 3600
expires_in = 3600 -> refresh_in = null
```

gladjohn:
Fixes #4062

- Managed Identity Endpoints do not give us refresh_in. The plan is to calculate this in MSAL

**Testing**
Unit 

**Performance impact**
None

**Documentation**

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
@@ -544,7 +544,7 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
         }
 
         [TestMethod]
-        public async Task ManagedIdentityRefreshInTestAsync()
+        public async Task ManagedIdentityIsProactivelyRefreshedAsync()
         {
             using (new EnvVariableContext())
             using (var httpManager = new MockHttpManager())


tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -443,5 +443,68 @@ public async Task ManagedIdentityCacheTestAsync()
...
+
+                AcquireTokenForManagedIdentityParameterBuilder builder = mi.AcquireTokenForManagedIdentity(Resource);
+
+                switch (hoursToAdd)
+                {
+                    case 0:
+                        MsalClientException ex = await AssertException.TaskThrowsAsync<MsalClientException>(
+                            () => builder.ExecuteAsync()).ConfigureAwait(false);
+                        Assert.AreEqual(ex.ErrorCode, "invalid_token_provider_response_value");
+                        return;
```
gladjohn, line 480: not the greatest way to do this I guess

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -244,15 +244,28 @@ internal static MsalTokenResponse CreateFromManagedIdentityResponse(ManagedIdent
         {
             ValidateManagedIdentityResult(managedIdentityResponse);
 
+            long expiresOn = DateTimeHelpers.GetDurationFromNowInSeconds(managedIdentityResponse.ExpiresOn);
```
neha-bhargava, line 247:
```suggestion
            long expiresIn = DateTimeHelpers.GetDurationFromNowInSeconds(managedIdentityResponse.ExpiresOn);

```

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
```diff
@@ -116,6 +116,14 @@ public static string GetMsiImdsSuccessfulResponse()
           "\"ext_expires_in\":\"12345\",\"token_type\":\"Bearer\"}";
         }
 
+        public static string GetMsiSuccessfulWithExpiresOnResponse(int hoursToAdd)
```
neha-bhargava, line 119: You can reuse https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs#L102. here.

gladjohn:
> This will also need a change here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L82 For reference see: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs#L91

doesn't this call into the same method? 

image


https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L98

neha-bhargava:
> > This will also need a change here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L82 For reference see: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs#L91
> 
> doesn't this call into the same method?
> 
> image
> 
> https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L98

Sorry if that was not clear, we need to add this logic as well:
image

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
```diff
@@ -15,19 +15,45 @@ namespace Microsoft.Identity.Client.ManagedIdentity
...
+        /// <remarks>which is a "Bearer" access token, which means the resource 
+        /// can give access to the bearer of this token.</remarks>
         [JsonProperty("token_type")]
         public string TokenType { get; set; }
 
+        /// <summary>
+        /// A unique identifier generated by Azure AD for the Azure Resource.
+        /// </summary>
+        /// <remarks>The Client ID is a GUID value that uniquely identifies the application 
+        /// and its configuration within the identity platform</remarks>
```
bgavrilMS, line 54: Not a fan of self documenting code I see :)

bgavrilMS (approved):

gladjohn:
> > > This will also need a change here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L82 For reference see: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs#L91

Thanks @neha-bhargava, I have modified it and tested it. I need to add a unit test for the changes. Will do so and confirm

trwalke (approved):

neha-bhargava (approved):

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
```diff
@@ -99,9 +99,9 @@ public static string GetBridgedHybridSpaTokenResponse(string spaAccountId)
             ",\"id_token_expires_in\":\"3600\"}";
         }
 
-        public static string GetMsiSuccessfulResponse()
+        public static string GetMsiSuccessfulResponse(int hoursToAdd = 1)
```
pmaytak, line 102: nit: `expiresInHours`?

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -244,15 +244,28 @@ internal static MsalTokenResponse CreateFromManagedIdentityResponse(ManagedIdent
         {
             ValidateManagedIdentityResult(managedIdentityResponse);
 
+            long expiresIn = DateTimeHelpers.GetDurationFromNowInSeconds(managedIdentityResponse.ExpiresOn);
```
pmaytak, line 247: Managed identity response provides only expires_**on**, not expires_**in**, right? The code sample in the issue #4062 is incorrect then?

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
```diff
@@ -244,15 +244,28 @@ internal static MsalTokenResponse CreateFromManagedIdentityResponse(ManagedIdent
...
-                ExpiresIn = DateTimeHelpers.GetDurationFromNowInSeconds(managedIdentityResponse.ExpiresOn),
+                ExpiresIn = expiresIn,
                 TokenType = managedIdentityResponse.TokenType,
-                TokenSource = TokenSource.IdentityProvider
+                TokenSource = TokenSource.IdentityProvider,
+                RefreshIn = InferManagedIdentityRefreshInValue(expiresIn)
             };
         }
 
+        private static long? InferManagedIdentityRefreshInValue(long expiresIn)
```
pmaytak, line 259:
```suggestion
        // Compute refresh_in as 1/2 expires_in, but only if expires_in > 2h.
        private static long? InferManagedIdentityRefreshInValue(long expiresIn)

```

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -443,5 +443,65 @@ public async Task ManagedIdentityCacheTestAsync()
                 appTokenCacheRecoder.AssertAccessCounts(1, 1);
             }
         }
+
+        [DataTestMethod]
+        [DataRow(0)]
+        [DataRow(1)]
+        [DataRow(2)]
+        [DataRow(3)]
+        public async Task ManagedIdentityExpiresOnTestAsync(int hoursToAdd)
```
pmaytak, line 452: nit: something more specific like `expireInHours`

pmaytak (approved):

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
gladjohn: We map the managed identity response `ExpiresOn` to MSALTokenResponse `Expires_IN`. But some RPs do provide the actual Expires_In value, like in this case IMDS. You can run this [helper service query](https://service.msidlab.com/MSIToken?uri=http%3A%2F%2F169.254.169.254%2Fmetadata%2Fidentity%2Foauth2%2Ftoken%3Fapi-version%3D2018-02-01%26resource%3Dhttps%3A%2F%2Fmanagement.azure.com&azureResource=VM
), to see IMDS does return expires_in

But it leads to the discussion of when we should infer the refresh_on? In IMDS should we use what the RPs provide? @neha-bhargava @bgavrilMS

gladjohn:
> > > This will also need a change here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L82 For reference see: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs#L91
> > 
> > 
> > doesn't this call into the same method?
> > image
> > https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs#L98
> 
> Sorry if that was not clear, we need to add this logic as well: image

added this and also added a unit test for this. Thanks @neha-bhargava

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -37,19 +37,21 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
-                MsalAccessTokenCacheItem cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
+                cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
 
                 if (cachedAccessTokenItem != null)
                 {
                     AuthenticationRequestParameters.RequestContext.ApiEvent.IsAccessTokenCacheHit = true;
 
                     Metrics.IncrementTotalAccessTokensFromCache();
-                    return new AuthenticationResult(
+                    authResult = new AuthenticationResult(
```
gladjohn, line 54: This was the blocker for adding the unit tests. Thanks @pmaytak for your help

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -480,5 +483,168 @@ public async Task ManagedIdentityCacheTestAsync()
...
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+
+                Assert.AreEqual(ApiEvent.ApiIds.AcquireTokenForSystemAssignedManagedIdentity, builder.CommonParameters.ApiId);
+
+                switch (expiresInHours)
+                {
+                    case 0:
+                    case 1:
+                    case 2:
+                        Assert.IsFalse(result.AuthenticationResultMetadata.RefreshOn.HasValue);
```
pmaytak, line 538: nit: this could be a separate parameter in the test - `refreshOnHasValue`
```csharp
        [DataRow(0, false)]
        [DataRow(1, false)]
        [DataRow(2, false)]
        [DataRow(3, true)]
        public async Task ManagedIdentityRefreshOnIsCalculatedAsync(int expiresInHours)
```
and then only
```csharp
Assert.AreEqual(refreshOnHasValue, result.AuthenticationResultMetadata.RefreshOn.HasValue);
```

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -480,5 +483,168 @@ public async Task ManagedIdentityCacheTestAsync()
...
+
+                httpManager.AddManagedIdentityMockHandler(
+                    AppServiceEndpoint,
+                    Resource,
+                    MockHelpers.GetMsiSuccessfulResponse(expiresInHours),
+                    ManagedIdentitySource.AppService);
+
+                AcquireTokenForManagedIdentityParameterBuilder builder = mi.AcquireTokenForManagedIdentity(Resource);
+
+                if (expiresInHours == 0)
```
pmaytak, line 514: nit: this can be moved to a separate test method, something like InvalidRefreshOn_Throws

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -480,5 +483,168 @@ public async Task ManagedIdentityCacheTestAsync()
...
+                {
+                    MsalClientException ex = await AssertException.TaskThrowsAsync<MsalClientException>(
+                        () => builder.ExecuteAsync()).ConfigureAwait(false);
+
+                    Assert.AreEqual(ex.ErrorCode, "invalid_token_provider_response_value");
```
pmaytak, line 519:
```suggestion
                    Assert.AreEqual(ex.ErrorCode, MsalError.InvalidTokenProviderResponseValue);
```

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -480,5 +483,168 @@ public async Task ManagedIdentityCacheTestAsync()
...
+                Assert.IsNotNull(result);
+
+                Assert.AreEqual(0, httpManager.QueueSize,
+                    "MSAL should have refreshed the token because the original AT was marked for refresh");
+                
+                cacheAccess.WaitTo_AssertAcessCounts(1, 1);
+
+                Assert.IsTrue(result.AuthenticationResultMetadata.CacheRefreshReason == CacheRefreshReason.ProactivelyRefreshed);
+                
+                Assert.IsTrue(result.AuthenticationResultMetadata.RefreshOn == refreshOn);
```
pmaytak, line 601: Use Assert.AreEqual.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -480,5 +483,168 @@ public async Task ManagedIdentityCacheTestAsync()
...
+
+                result = await mi.AcquireTokenForManagedIdentity(Resource)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+                
+                Assert.IsTrue(result.AuthenticationResultMetadata.CacheRefreshReason == CacheRefreshReason.NotApplicable);
+            }
+        }
+
+        private bool YieldTillSatisfied(Func<bool> func, int maxTimeInMilliSec = 30000)
```
pmaytak, line 611: This is used in multiple tests. Should be extracted in some common class.

Ideally, long term, we should refactor to not need this. Very flaky. :(

pmaytak (approved):

src/client/Microsoft.Identity.Client/OAuth2/MsalTokenResponse.cs
neha-bhargava: Since the ExpiresOn value is consistent in the response, we can use the same logic. Aside I don't think there will be any difference between us calculating the expiresIn and the one returned from IMDS.

gladjohn: Also I think for consistency it is better for us to infer

PR #4078, opened 2023-04-13
Add retry for specific codes in managed identity

### LINKED ISSUE #4067: Add retry logic for managed identity flow similar to the retry logic in Azure SDK.

neha-bhargava:
Fixes #4067 

Retry for specific status codes for managed identity

**Testing**
Unit tests

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -23,7 +23,7 @@ namespace Microsoft.Identity.Client
 {
     internal sealed class ApplicationConfiguration : IAppConfig
     {
-        public ApplicationConfiguration(bool isConfidentialClient) 
+        public ApplicationConfiguration(bool isConfidentialClient, bool isManagedIdentity = false)
```
bgavrilMS, line 26: Pls use an enum instead of 2 flags. There are 12 references to this, it's not a big refactoring.

bgavrilMS: The concern with 2 flags is that the state [ConfidentialClient = true, ManagedIdentity = true] is not possible, but MSAL developers might forget this.

bgavrilMS: For example https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs#L209 - this is a bug, as the error message is wrong.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -24,17 +24,15 @@ namespace Microsoft.Identity.Client.Http
...
     internal class HttpManager : IHttpManager
     {
-        private readonly IMsalHttpClientFactory _httpClientFactory;
-        //Determines whether or not to retry on 5xx errors. Configurable on application creation. default is true;
-        private readonly bool _retryConfig;
+        protected readonly IMsalHttpClientFactory _httpClientFactory;
         public long LastRequestDurationInMs { get; private set; }
 
-        public HttpManager(IMsalHttpClientFactory httpClientFactory, bool retry = true)
+        protected bool _retryConfig = false;
```
bgavrilMS, line 30: nit: _isRetryConfigured

src/client/Microsoft.Identity.Client/Http/HttpManagerFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static IHttpManager GetHttpManager(IMsalHttpClientFactory httpClientFactory, bool retryConfig, bool isManagedIdentity)
+        {
+            if (!retryConfig)
+            {
+                return new HttpManager(httpClientFactory);
+            }
+
+            return isManagedIdentity ?
+                new HttpManagerManagedIdentity(httpClientFactory) :
+                new HttpManagerWithRetry(httpClientFactory);
```
bgavrilMS, line 26: nit: HttpManagerConfidentialClient ?

src/client/Microsoft.Identity.Client/Http/HttpManagerManagedIdentity.cs
```diff
@@ -0,0 +1,45 @@
...
+    internal class HttpManagerManagedIdentity : HttpManagerWithRetry
+    {
+        public HttpManagerManagedIdentity(IMsalHttpClientFactory httpClientFactory) : 
+            base(httpClientFactory) { }
+
+        /// <summary>
+        /// Retry policy specific to managed identity flow.
+        /// Avoid changing this, as it's breaking change.
+        /// </summary>
+        protected override bool IsRetriableStatusCode(int statusCode)
```
bgavrilMS, line 29: nit:  `Retryable` 

because https://english.stackexchange.com/questions/305273/retriable-or-retryable and because public API `IsRetryable`

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava: Updated. Makes sense.

src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
```diff
@@ -48,5 +83,26 @@ public class MsalManagedIdentityException : MsalServiceException
...
         public ManagedIdentitySource ManagedIdentitySource { get; }
+
+        /// <summary>
+        /// Retry error codes specific to managed identity
+        /// </summary>
+        protected override void UpdateIsRetryable()
+        {
+            switch (StatusCode)
+            {
+                case 404: //Not Found
```
gladjohn, line 94: 404 should not be retriable?

src/client/Microsoft.Identity.Client/MsalServiceException.cs
```diff
@@ -224,7 +224,7 @@ public HttpResponseHeaders Headers
         /// <summary>
         /// As per discussion with Evo, AAD 
         /// </summary>
-        private void UpdateIsRetryable()
+        protected virtual void UpdateIsRetryable()
         {
             IsRetryable =
                 (StatusCode >= 500 && StatusCode < 600) ||
```
gladjohn, line 230: 501 and 505 will also be in this condition, which should not be retried

src/client/Microsoft.Identity.Client/AppConfig/ApplicationType.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// Enum to represent the type of application.
+    /// </summary>
+    internal enum ApplicationType
```
gladjohn, line 15:
```suggestion
    internal enum MSALApplicationType
```
We are dealing with so many apps and app types, would be good to be specific on this App Type

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -383,6 +383,51 @@ await mi.AcquireTokenForManagedIdentity(Resource)
             }
         }
 
+        [DataTestMethod]
+        [DataRow(ManagedIdentitySource.AppService, AppServiceEndpoint, HttpStatusCode.NotFound)]
```
gladjohn, line 387: we should not retry the not found? just a strain on the server. not sure retry will immediately solve the issue, as 404 could be because of various reasons.

src/client/Microsoft.Identity.Client/Http/HttpManagerWithRetry.cs
```diff
@@ -0,0 +1,195 @@
...
+
+                logger.Error("The HTTP request failed. " + exception.Message);
+                isRetriable = true;
+                timeoutException = exception;
+            }
+
+            if (isRetriable && retry)
+            {
+                logger.Info("Retrying one more time..");
+                await Task.Delay(TimeSpan.FromSeconds(1)).ConfigureAwait(false);
```
gladjohn, line 150: suggested wait time is 5 seconds : https://learn.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#retry-usage-guidance

gladjohn: But if this is specific to MI, then should we not do the Exponential back-off we discussed? 1 2 4 ?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -316,7 +277,7 @@ private async Task<HttpContent> CloneHttpContentAsync(HttpContent httpContent)
         /// In HttpManager, the retry policy is based on this simple condition.
         /// Avoid changing this, as it's breaking change.
```
gladjohn, line 278: can we add a test to catch this if someone changes it?

gladjohn (approved):
just a few comments

src/client/Microsoft.Identity.Client/AppConfig/ApplicationType.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        ConfidentialClient,
+        PublicClient,
+        ManagedIdentity
```
pmaytak, line 19: nit for consistency?
```suggestion
        ManagedIdentityClient
```

src/client/Microsoft.Identity.Client/AppConfig/ApplicationType.cs
pmaytak: `MsalClientType` ?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -23,9 +24,18 @@ namespace Microsoft.Identity.Client
 {
     internal sealed class ApplicationConfiguration : IAppConfig
     {
-        public ApplicationConfiguration(bool isConfidentialClient) 
+        public ApplicationConfiguration(ApplicationType applicationType)
```
pmaytak, line 27: I think `IsManagedIdentity` and `IsConfidentialClient` are not needed and should be replaced with `ApplicationType` property.

If that results in many changes, then maybe refactor in a follow up PR, not this one.

pmaytak (changes requested):

pmaytak (changes requested):
I think it'd be good to refactor the ApplicationType usage for clarity.

src/client/Microsoft.Identity.Client/MsalServiceException.cs
neha-bhargava: This is an existing retry logic for ests requests. Changing this logic is a breaking change.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
bgavrilMS: It's a big PR already, just bite the bullet imo. If changes are in a single commit, it's easy to review (or to skip review)

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
bgavrilMS: We surely have a quite a few of these.

src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
bgavrilMS: I think it is. Some scenario where IMDS is getting updated.

src/client/Microsoft.Identity.Client/MsalServiceException.cs
bgavrilMS: Just FYI, ESTS currently throws 500 and 503.

src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
neha-bhargava: Yes, 404 is specifically retried for managed identity scenarios in azure identity repo. https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/ManagedIdentitySource.cs#L153

src/client/Microsoft.Identity.Client/Http/HttpManagerWithRetry.cs
neha-bhargava: Discussed this with @bgavrilMS , we will not implement the retry logic wih exponential back-off since that is a functionality provided by Azure SDK for all the calls. It is by default set to 3 retries but user can configure and change it. MSAL is a low level library and we are not planning to implement all that retry logic. We will instead return IsRetriable true and document that users can choose to implement the retry.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava: I will create another PR for these changes, thanks for the suggestion @pmaytak

neha-bhargava: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4113

src/client/Microsoft.Identity.Client/Http/HttpManagerFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.Http
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    internal sealed class HttpManagerFactory
+    {
+        public static IHttpManager GetHttpManager(IMsalHttpClientFactory httpClientFactory, bool retryConfig, bool isManagedIdentity)
```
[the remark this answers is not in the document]
neha-bhargava: Updating to withRetry as MSAL retries only once.

pmaytak (approved):

src/client/Microsoft.Identity.Client/Http/HttpManagerManagedIdentity.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Retry policy specific to managed identity flow.
+        /// Avoid changing this, as it's breaking change.
+        /// </summary>
+        protected override bool IsRetryableStatusCode(int statusCode)
+        {
+            switch (statusCode)
+            {
+                case 404: //Not Found
```
rayluo, line 33: I'm late to the party. :-)

Why is 404 considered retriable?

PR #4086, opened 2023-04-15
Replace "login.windows.net" with "login.microsoftonline.com" in WwwAuthenticateParameters Tests

gladjohn:
Replace "login.windows.net" with "login.microsoftonline.com"

**Testing**
test

**Performance impact**
none

**Documentation**

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
 
             var authParams = await WwwAuthenticateParameters.CreateFromResourceResponseAsync("https://buildautomation.vault.azure.net/secrets/CertName/CertVersion").ConfigureAwait(false);
 
-            Assert.AreEqual("login.windows.net", new Uri(authParams.Authority).Host);
+            Assert.AreEqual("login.microsoftonline.com", new Uri(authParams.Authority).Host);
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/WwwAuthenticateParametersIntegrationTests.cs
```diff
@@ -32,7 +32,7 @@ public async Task CreateWwwAuthenticateResponseFromKeyVaultUrlAsync()
 
             var authParams = await WwwAuthenticateParameters.CreateFromResourceResponseAsync("https://buildautomation.vault.azure.net/secrets/CertName/CertVersion").ConfigureAwait(false);
 
-            Assert.AreEqual("login.windows.net", new Uri(authParams.Authority).Host);
+            Assert.AreEqual("login.microsoftonline.com", new Uri(authParams.Authority).Host);
```
rayluo, line 35: The change itself looks reasonable. Out of curiosity, why is this test passed (shown as "checks" in github PR?) after we altered an assert value? Was this test case failing before this PR?

bgavrilMS: KeyVault folks have just changed the way they emit this header. We do have unit tests that ensure this functionality is works, but we also added a few integration tests to make sure big customers's headers work. 

This change is not problem and it doesn't break any app, but it does break our assertions (and it's fine, a change like this happens rarery).

gladjohn: Yes that was my finding as well. Folks have been asked to move to use `login.microsoftonline.com` for a while now (4 years I think), so this should be Ok

bgavrilMS (approved):

PR #4105, opened 2023-04-25
Add telemetry API ID for long-running OBO.

### LINKED ISSUE #4099: [Feature Request] Add telemetry API ID for long-running OBO
- Add [API ID](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e2728384232d19bb02258c297696e83d01d8a1fd/src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs#L12) value for acquire token in long running OBO.
- Set this API ID for long-running OBO request.
- Go over any `ApiIds.AcquireTokenOnBehalfOf` usage and update to use the long-running OBO if needed.
- Maybe create a helper, if needed, in ApiEvent - `IsOnBehalfOfRequest(ApiId) => ApiId is normal OBO or LR obo` - for cases that apply to both OBO methods.
- Update tests.

pmaytak:
Fixes #4099

- Added telemetry ApiID for long-running OBO

**Testing**
Added tests for other ApiIDs also.


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
```diff
@@ -181,7 +181,7 @@ protected override void Validate()
         /// <inheritdoc />
         internal override ApiEvent.ApiIds CalculateApiEventId()
         {
-            return ApiEvent.ApiIds.AcquireTokenOnBehalfOf;
+            return string.IsNullOrEmpty(Parameters.LongRunningOboCacheKey) ? ApiEvent.ApiIds.AcquireTokenOnBehalfOf : ApiEvent.ApiIds.AcquireTokenInLongRunningObo;
```
trwalke, line 184: Did we want to identify the difference between initiate and AcquireTokenInLongRunningObo? We were talking about determining who is calling Initiate over and over again in telemetry right?

src/client/Microsoft.Identity.Client/TelemetryCore/Internal/Events/ApiEvent.cs
```diff
@@ -30,7 +30,8 @@ public enum ApiIds
             // The API Ids for managed identity will not be found in Http telemetry,
             // as we don't hit eSTS for managed identity calls.
             AcquireTokenForSystemAssignedManagedIdentity = 1015,
-            AcquireTokenForUserAssignedManagedIdentity = 1016
+            AcquireTokenForUserAssignedManagedIdentity = 1016,
```
bgavrilMS, line 33: Please add a commetn with the version of when this is introduced. It'll help when querying the data.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
pmaytak: Yea, makes sense, will update.
Currently we can figure out the difference because AcquireTokenInLrObo will use a refresh grant, while InitiateLrObo will use obo grant. But once we update the Initiate method to check the cached assertion, this won't work, since initiate will be able to do refresh grant also.

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/RequestsTests/IntegratedWindowsAuthUsernamePasswordTests.cs
```diff
@@ -362,6 +365,8 @@ public async Task AcquireTokenByIntegratedWindowsAuthMetadataTestAsync()
                 StringAssert.Contains(realmDiscoveryHandler.ActualRequestMessage.Headers.ToString(), TestConstants.XClientVer,
                     "Client info header should contain " + TestConstants.XClientVer,
                     StringComparison.OrdinalIgnoreCase);
+                // Assert telemetry ApiId
```
gladjohn, line 368:
```suggestion

                // Assert telemetry ApiId
```

PR #4107, opened 2023-04-26
Update API for managed identity

### LINKED ISSUE #4114: Update managed identity API to specify the identity when creating the application.
Update API to implement proposal 6: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/8040?_a=files&path=/ManagedIdentity/ManagedIdentity.md

neha-bhargava:
Fixes #4114 

* Update managed identity api to specify managed identity with .Create() method.
* Remove cache extensibility APIs from managed identity builder.

**Testing**
Update unit and integration tests

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/AppConfig/IManagedIdentity.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// Interface for managed identity configuration supported by MSAL.
+    /// </summary>
+    public interface IManagedIdentity
```
siddhijain, line 15: Wondering if we can name this something different (IManagedIdentityConfiguration) as another interface IManagedIdentityApplication already exists that looks confusing.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationOptions.cs
```diff
@@ -10,9 +12,10 @@ namespace Microsoft.Identity.Client
     public class ManagedIdentityApplicationOptions : BaseApplicationOptions
     {
         /// <summary>
-        /// User assigned client id or resource id of the managed identity resource. 
+        /// Managed identity assigned to the resource, for system assigned managed identity <see cref="SystemAssignedManagedIdentity"/>
+        /// and for user assigned managed identity <see cref="UserAssignedManagedIdentity"/>.
         /// </summary>
-        public string UserAssignedClientId { get; set; }
+        public IManagedIdentity ManagedIdentity { get; set; }
```
bgavrilMS, line 18: Pls remove `EnableCacheSynchronization`

src/client/Microsoft.Identity.Client/AppConfig/SystemAssignedManagedIdentity.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Class to define a system assigned managed identity.
+    /// </summary>
+    public class SystemAssignedManagedIdentity : IManagedIdentity
+    {
+        /// <summary>
+        /// Returns a default instance of <see cref="SystemAssignedManagedIdentity"/>.
+        /// </summary>
+        /// <returns></returns>
+        public static IManagedIdentity Default()
```
bgavrilMS, line 21: How about 

```csharp

private static SystemAssignedManagedIdentity s_default = new ..
public static IManagedIdentity Default 
{
     get { return s_default; }
}
```
```

File: src/client/Microsoft.Identity.Client/AppConfig/IManagedIdentity.cs (hunk 1 above)
Comment (bgavrilMS) [inline src/client/Microsoft.Identity.Client/AppConfig/IManagedIdentity.cs:15]: Empty interface here seems odd. And in general interfaces are difficult to:

1. version, because you can't add methods to a public interface
2. reason with, because app developers think they have to create an object which implements this interface. And sometimes they actually do (e.g. we had a number of bugs from ppl trying to implement `IAccount`, which is a bad interface)


I'm thinking more of a data class:

```csharp
public class ManagedIdentityDescription
    {
        private bool _isSystemAssigned = false;
        
        private ManagedIdentityDescription(bool isSystemAssigned)
        {
            _isSystemAssigned = isSystemAssigned;
        }

        public static ManagedIdentityDescription SystemAssigned { get; } 
            = new ManagedIdentityDescription(true); // this ensures that we don't create a new instance every time

        public static ManagedIdentityDescription UserAssignedFromClientId(string clientId)
        {
            return new ManagedIdentityDescription(false) { ClientId = clientId };
        }

        public static ManagedIdentityDescription UserAssignedFromResourceId(string resourceId)
        {
            return new ManagedIdentityDescription(false) { ResourceId = resourceId };
            
        }

        public string ClientId { get; private set; }
        public string ResourceId { get; private set; }

    }
```
CC @mjcheetham  as we discussed this

src/client/Microsoft.Identity.Client/AppConfig/UserAssignedIdType.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// Enum to define types of user assigned identities supported by MSAL.
+    /// </summary>
+    public enum UserAssignedIdType
```
bgavrilMS, line 15: See my comment about using objects. I don't think this is needed, or at least not publicly.

src/client/Microsoft.Identity.Client/AppConfig/IManagedIdentity.cs
rayluo: Like the data class approach above. One more suggestion is that the data attributes shall better be `UserAssignedIdType` and `Id` which always have valid value, rather than a bunch of different IDs that might default to NULL. Such a "type + id" data model also matches to the proposal 6 in our ApiReview PR.

src/client/Microsoft.Identity.Client/AppConfig/UserAssignedIdType.cs
neha-bhargava: Updated scope to internal

src/client/Microsoft.Identity.Client/AppConfig/IManagedIdentity.cs
neha-bhargava: Updated to ManagedIdentityConfiguration. Thanks!

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
```diff
@@ -46,92 +47,55 @@ internal ManagedIdentityApplicationBuilder(ApplicationConfiguration configuratio
...
+            var config = new ApplicationConfiguration(isConfidentialClient: false);
 
-            if (Guid.TryParse(userAssignedId, out _))
+            if (managedIdentityConfiguration == null)
             {
-                Config.ManagedIdentityUserAssignedClientId = userAssignedId;
-            } 
-            else
+                throw new ArgumentNullException(nameof(managedIdentityConfiguration));
+            }
```
pmaytak, line 81: Check should be moved above app config init.

Can be also written as:
```csharp
_ = managedIdentityConfiguration ?? throw new ArgumentNullException(nameof(managedIdentityConfiguration));
```

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationOptions.cs
```diff
@@ -10,19 +12,8 @@ namespace Microsoft.Identity.Client
     public class ManagedIdentityApplicationOptions : BaseApplicationOptions
```
pmaytak, line 12: Top of mind thought - it's kind of confusing that there're `ManagedIdentityApplicationOptions` and `ManagedIdentityConfiguration`; and both are public. Our other AppOptions classes are public while AppConfig classes are internal.
Feels like the Options class should include the functionality of the config class.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
```diff
@@ -46,92 +47,55 @@ internal ManagedIdentityApplicationBuilder(ApplicationConfiguration configuratio
...
-            var config = new ApplicationConfiguration(isConfidentialClient: false);
-            return new ManagedIdentityApplicationBuilder(config)
-                .WithCacheSynchronization(false);
-        }
-
         /// <summary>
         /// Creates a ManagedIdentityApplicationBuilder from a user assigned managed identity clientID / resourceId.
-        /// See https://aka.ms/msal-net-managed-identity
+        /// For example, for a system assigned managed identity use ManagedIdentityApplicationBuilder.Create(SystemAssignedManagedIdentity.Default())
+        /// and for a user assigned managed identity use ManagedIdentityApplicationBuilder.Create(UserAssignedManagedIdentity.FromClientId(clientId)).
```
rayluo, line 58: Need to change these comments to match the new API naming accordingly.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+        public static ManagedIdentityConfiguration SystemAssigned { get; } = 
+            new ManagedIdentityConfiguration(ManagedIdentityIdType.SystemAssigned);
+
+        /// <summary>
+        /// Create an instance of ManagedIdentityConfiguration for a user assigned managed identity from a client id.
+        /// </summary>
+        /// <param name="clientId">Client id of the user assigned managed identity assigned to azure resource.</param>
+        /// <returns>Instance of ManagedIdentityConfiguration.</returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static ManagedIdentityConfiguration UserAssignedFromClientId(string clientId)
```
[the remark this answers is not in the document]
siddhijain: Yeah, the second one is clearer.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
```diff
@@ -47,92 +47,51 @@ internal ManagedIdentityApplicationBuilder(ApplicationConfiguration configuratio
...
 
-        private ManagedIdentityApplicationBuilder WithUserAssignedManagedIdentity(string userAssignedId)
+        private static ApplicationConfiguration BuildConfiguration(ManagedIdentityConfiguration managedIdentityConfiguration)
         {
-            Config.IsUserAssignedManagedIdentity = true;
+            _ = managedIdentityConfiguration ?? throw new ArgumentNullException(nameof(managedIdentityConfiguration));
+            var config = new ApplicationConfiguration(MsalClientType.ManagedIdentityClient);
 
-            if (Guid.TryParse(userAssignedId, out _))
+            switch (managedIdentityConfiguration.IdType)
```
trwalke, line 79: I feel like this switch statement can be simplified to an if statement with some duplicate lines removed. For example `config.IsUserAssignedManagedIdentity = true;` probably doesnt need to be in 2 places

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        private ManagedIdentityConfiguration(ManagedIdentityIdType idType)
+        {
+            IdType = idType;
+        }
+
+        /// <summary>
+        /// Create an instance of ManagedIdentityConfiguration for a system assigned managed identity.
+        /// </summary>
+        public static ManagedIdentityConfiguration SystemAssigned { get; } =
```
trwalke, line 32: Just curious, what the reason is for making these methods static?

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationOptions.cs
trwalke: Yeah, I agree. There should only be 1 class that is responsible for configuring the application.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityIdType.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    internal enum ManagedIdentityIdType
+    {
+        SystemAssigned,
```
trwalke, line 14: nit: Does 'SystemAssignedId' make sense here? or does SystemAssigned mean the system will chose between clientId or resourceId?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/AzureArcTests.cs
```diff
@@ -24,17 +25,24 @@ public class AzureArcTests : TestBase
...
-        public async Task AzureArcUserAssignedManagedIdentityNotSupportedAsync(string userAssignedClientId, UserAssignedIdentityId userAssignedIdentityId)
+        public async Task AzureArcUserAssignedManagedIdentityNotSupportedAsync(string userAssignedId, UserAssignedIdentityId userAssignedIdentityId)
         {
             using (new EnvVariableContext())
             using (var httpManager = new MockHttpManager(isManagedIdentity: true))
             {
                 SetEnvironmentVariables(ManagedIdentitySource.AzureArc, ManagedIdentityTests.AzureArcEndpoint);
 
-                IManagedIdentityApplication mi = ManagedIdentityApplicationBuilder.Create(userAssignedClientId)
+                var miBuilder = ManagedIdentityApplicationBuilder
```
trwalke, line 35: Maybe you can create one method that creates an appication for all of these tests?

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationOptions.cs
neha-bhargava: As discussed in the standup, removing the ManagedIdentityApplicationOptions for now and will add as needed later.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityIdType.cs
pmaytak: SystemAssigned means system assigned identity, not user assigned ID. I see how some new to MSI may think abovel. I would assume people who work with MSI will understand the difference.

`SystemAssignedId` could work for consistency. 

We should make sure the enum/method comments are clear, though.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
```diff
@@ -201,7 +140,7 @@ public IManagedIdentityApplication Build()
         /// <returns></returns>
         internal ManagedIdentityApplication BuildConcrete()
         {
-            ValidateUseOfExperimentalFeature("ManagedIdentityClient");
+            ValidateUseOfExperimentalFeature("ManagedIdentity");
```
pmaytak, line 143: The GA will be part of a separate PR, I assume? (If not, I think it should be separate.)

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationOptions.cs
pmaytak, line 1: Didn't realize we already released with app options. I guess it shouldn't break anyone?

Should we keep the options, but let's say not add the Id related config to them?

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityConfiguration.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// Class to store configuration for a managed identity enabled on a resource.
+    /// For a system assigned managed identity use ManagedIdentityConfiguration.SystemAssigned.
+    /// For user assigned managed identity use ManagedIdentityConfiguration.WithUserAssignedClientId("clientId") or 
+    /// ManagedIdentityConfiguration.WithUserAssignedResourceId("resourceId").
+    /// For more details see https://aka.ms/msal-net-managed-identity
+    /// </summary>
+    public class ManagedIdentityConfiguration
```
pmaytak, line 19: Is the plan for this class to have any other non-Id related config also?

Since we have these static builders here, if we add more config to this class, I'd think we need to add more static builders for each config? That can get confusing...

My proposal:
- rename this class to `ManagedIdentityId`.
- It will still have the same logic - string ID, ID type, etc.
- It can also be added to MIAppOptions as a property - `public ManagedIdentityId ManagedIdentityId { get; set; }`
- Since it's a simple class, it can still be used in options when someone loads options from config or something.
- If there are more MI related config, we can add it to some other new class?

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
```diff
@@ -32,107 +32,46 @@ internal ManagedIdentityApplicationBuilder(ApplicationConfiguration configuratio
...
-            {
-                Config.ManagedIdentityUserAssignedResourceId = userAssignedId;
+                case ManagedIdentityIdType.ClientId:
+                    config.IsUserAssignedManagedIdentity = true;
+                    config.ManagedIdentityUserAssignedClientId = managedIdentityConfiguration.UserAssignedId;
+                    break;
+                case ManagedIdentityIdType.ResourceId:
+                    config.IsUserAssignedManagedIdentity = true;
+                    config.ManagedIdentityUserAssignedResourceId = managedIdentityConfiguration.UserAssignedId;
+                    break;
```
pmaytak, line 68: Can we just add a `ManagedIdentityConfiguration` property to `ApplicationConfiguration`

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityIdType.cs
neha-bhargava: Since there is no Id associated with System assigned managed identity, it is SystemAssigned. Client id and resource id are associated with user assigned managed identity.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
neha-bhargava: Yes, GA will be separate PR

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationOptions.cs
neha-bhargava: It was part of experimental Managed identity API right? The users we have talked to are not using the AppicationOptions for MI. I think it will be OK to remove.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityConfiguration.cs
neha-bhargava: My initial thought was also to add this to MIAppOptions as you mentioned above. I think the only concern here is the naming. 
Regarding the question, will this class be used to store anything other than id? I don't think so, it does not replace the MIAppBuilder. More config will be added there. The `WithUserAssignedClientId` / `WithUserAssignedResourceId` methods are helper methods to let the user create an instance of this class using client id or resource id. This class will be used to handle more Ids if needed, 

@bgavrilMS

pmaytak (approved):
I would just rename `ManagedIdentityConfiguration` to `ManagedIdentityId`.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityConfiguration.cs
bgavrilMS: -  `WithUserAssignedClientId` `WithUserAssignedResourceId` should be removed. Let's not have 2 ways to configure the same thing, it leads to dev confusion. 
- We decided to remove `MIAppOptions` for now. It needs a bit more thought. We can add it later if we need it (pls open a issue).
- Naming is hard. `ManagedIdentityId` sounds ok to me.

neha-bhargava: Discussed offline with @bgavrilMS, renamed to `ManagedIdentityId` and removed MIAppOptions

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
neha-bhargava: Good point. We dont need all these separately in config. Removed.

PR #4111, opened 2023-05-01
Update InitiateLongRunningObo method to retrieve tokens from the cache if request assertion matches.

### LINKED ISSUE #4100: [Bug] InitiateLongRunningProcessInWebApi should return cached token if request assertion hash matches cached one
4.51.0+

**Actual behavior**
#3956 updated the `InitiateLongRunningProcessInWebApi` to never check the cache and get a token straight from AAD. Misuse of this can create COGS increase and get throttled by AAD.

**Expected behavior**
- Add `AssertionHash` property to `MsalAccessTokenCacheItem`.
- Update [OnBehalfOfRequest](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/e2728384232d19bb02258c297696e83d01d8a1fd/src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs#L47) or [FindAccessTokenAsync](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L433) to take assertion hash into account for  `InitiateLongRunningProcessInWebApi` method.
- Update tests.
- Update `InitiateLongRunningProcessInWebApi` method comment and docs so it's accurate.

pmaytak:
Fixes #4100

Updates InitiateLongRunningObo method to return valid cached tokens if cached assertion matches the request assertion; otherwise, the assertion is used to retrieve new tokens.
Updates access token cache item .

**Testing**
Updated tests.

**Performance impact**
None.

**Documentation**
- [ ] All relevant documentation is updated. - After release.


src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -138,10 +137,11 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
         private async Task<AuthenticationResult> RefreshRtOrFetchNewAccessTokenAsync(CancellationToken cancellationToken)
         {
             var logger = AuthenticationRequestParameters.RequestContext.Logger;
-            
-            if (ApiEvent.IsLongRunningObo(AuthenticationRequestParameters.ApiId))
+
+            // InitiateLongRunningProcessInWebApi retrieves tokens only with assertion, not by refresh token
```
pmaytak, line 141: The reason why we don't use RT here is because of #3825. If tokens are revoked in AAD, refreshing with this cached RT will throw an AAD exception. Since we already have a user assertion, use it to fetch new tokens via OBO grant. The expectation is that this use assertion is newer than the one that was used to acquire cached RT.

bgavrilMS:
> Fixes #4100
> 
> **Changes proposed in this request** Updates InitiateLongRunningObo method to return valid cached tokens if cached assertion matches the request assertion; otherwise, the assertion is used to retrieve new tokens. Updates access token cache item .
> 
> **Testing** Updated tests.
> 
> **Performance impact**
> 
> **Documentation**
> 
> * [ ]  All relevant documentation is updated.

Pls ensure that our OBO spec / api review is updated.

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -45,11 +46,12 @@ internal class MsalAccessTokenCacheItem : MsalCredentialCacheItemBase
             RawClientInfo = response.ClientInfo;
             HomeAccountId = homeAccountId;
             OboCacheKey = oboCacheKey;
+            UserAssertionHash = userAssertionHash;
```
bgavrilMS, line 49: OboAssertionHash?

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
bgavrilMS: Pls add this comment in the code.

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -240,23 +240,44 @@ public async Task InitiateLongRunningProcessWithExistingKeyAndToken_TestAsync()
...
                 AddMockHandlerAadSuccess(httpManager,
-                    responseMessage: MockHelpers.CreateSuccessTokenResponseMessage(accessToken: TestConstants.ATSecret2));
+                    responseMessage: MockHelpers.CreateSuccessTokenResponseMessage(accessToken: TestConstants.ATSecret2, refreshToken: TestConstants.RTSecret2),
+                    expectedPostData: new Dictionary<string, string> { 
+                        { OAuth2Parameter.RequestedTokenUse, OAuth2RequestedTokenUse.OnBehalfOf }, { OAuth2Parameter.GrantType, OAuth2GrantType.JwtBearer } });
 
-                //Initiate another process using the same cache key
-                //MSAL should ignore the token in the cache, fetch a new token and overrite the existing one
+                // Initiate another process using the same cache key and assertion
+                // Access token is expired, uses assertion, not a refresh token, to retrieve new tokens from AAD
```
trwalke, line 280: Not sure this is the right call or if I am missing something. If the AT is expired and the same OBO key and assertion is used, AAD wouldnt return a new token because the original user assertion would have expired along with the cached OBO AT. So, in order for this to work, wouldnt we need to use the RT of the expired AT that has the same OBO key and assertion? Tenically it will still be possible to refresh with the long running OBO. Im not saying that we should make the Initiate method use the RT. but I dont think AAD will return a new token in this case as the test suggests.

trwalke (commented):
The logic here makes sense overall, just one question about the test.

trwalke (approved):

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
trwalke: Now that I think about this some more, would't this break a customer who is using the initiate method over and over again once the original assertion has expired?

pmaytak: > Now that I think about this some more, would't this break a customer who is using the initiate method over and over again once the original assertion has expired?

Yes, if they upgrade from MSAL 4.50 or below. But as soon as they upgrade to 4.51, then it would break since Initiate would go to AAD each time with expired assertion (which will throw).

pmaytak: Hmm, yea I think I can update to also check for RT with the same assertion - if found use RT to refresh, otherwise use request assertion?

In the scenario when the cached tokens are revoked, we'd expect Initiate to be called with a new / uncached assertion.

src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs
```diff
@@ -64,6 +56,13 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                     cachedAccessToken = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
                 }
 
+                if (AuthenticationRequestParameters.ApiId == ApiEvent.ApiIds.InitiateLongRunningObo && cachedAccessToken != null &&
```
pmaytak, line 59: Currently cached/serialized token items (aka legacy tokens) don't have this new assertion hash property. So there's a scenario if app is updated to this new MSAL version that checks the hash now.

An option here is to add additional check - `&& !string.IsNullOrEmpty(cachedAccessToken.OboAssertionHash)`.

1. As-is, without the above check, `InitiateObo` will never match these cached legacy tokens, and will always use the passed-in assertion to get tokens from AAD, even if the cached tokens are unexpired. This actually mimics current behavior (4.51.0+).
2. With the additional check, `InitiateObo` will return the valid matched legacy access tokens and, if expired, will use legacy refresh token to refresh. This mimics the 4.50.0- behavior but only for legacy tokens.

In the scenario when AAD revokes AT and RT, method 2 reintroduces the previous issue of returning revoked tokens, so devs would have to use the `StopLongObo` method.

`AcquireLongObo` is not affected since it doesn't have the assertion in the request.

A more general question, do we even want to have an assertion hash match for refresh tokens?

Thoughts @bgavrilMS, @trwalke? Maybe I missed something?

trwalke: > do we even want to have an assertion hash match for refresh tokens?

The reason we discussed adding this was to prevent the scenario where someone is using initiate long running OBO over and over again from breaking after 4.51.0+. The initial implementation for this PR lead us to the realization that in order to actually do that while maintaining the logic that fixes the revocation issue is to add another check to the refresh OBO logic to check for a matching token hash. So, if we want to prevent breaking customers, we should keep the hash match for refresh tokens. 

> As-is, without the above check, InitiateObo will never match these cached legacy tokens, and will always use the passed-in assertion to get tokens from AAD, even if the cached tokens are unexpired. This actually mimics current behavior (4.51.0+).

I think this will only mimic the behavior of 4.51.0+ for the first user assertion expiration. So, once the user is required to sign in again, the legacy tokens should be overwritten and assertion will be cached with the new tokens. I think this is fine. The other option (#2) will require the developer to add the `StopLongRunningApi` method to remove the revoked tokens. That is a much more difficult state to recover from than simply asking the user to sign in again.

So, I dont think we need to add the null check. The end user experience is affected, but all they need to do is sign in again to recover and then everything will work fine.

Adding this null check will put the user in a state they cant recover from if the token are revoked without the app developer changing the app or deleting tokens manually.

Lets sync on this tomorrow @pmaytak incase I am missing something.

pmaytak:
Per discussion, will not implement this at this time.

PR #4112, opened 2023-05-01
Add property bag to exceptions and populate it for WAM

### LINKED ISSUE #4106: Require Tag and internal error code for WAM errors when thrown as an Msal exception
Right now in 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4e8df7a04cdec8a21a5a8dc57439f23561ec23e1/src/client/Microsoft.Identity.Client/Platforms/uap/WamBroker/WamAdapters.cs

When an unknown broker error is encountered the error code is just "unknownbroker error".  Can you attach the internal error code and the tag to the msal exception, this information is critical to diagnose wam related errors.   The information is in the message, but we need it on the object in order to use it in remote telemetry.

For example the string says

	"Unknown Status: Unexpected Error: 0xffffffff80070520 Context: Unexpected exception while waiting for accounts control to finish: '(pii)' Tag: 0x1f7d734b (error code -2147023584) (internal error code 528315211)"

We need the error code, internal error code and tag.

bgavrilMS:
Fixes #4106 

1. Add property bag for exceptions
2. Populate property bag with WAM properties

CC @rayluo @Avery-Dunn as this change would also make sense for PY / Java implemenations.

**Testing**

integration with ROPC

**Performance impact**

**Documentation**


tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
```diff
@@ -451,8 +452,35 @@ public async Task NewPOP_WithKeyIdOnly_Async()
...
+                    "noUser",
+                    "badPassword")                
+                .ExecuteAsync())
+                .ConfigureAwait(false);
+
+            Assert.AreEqual("557973642", ex.AdditionalResponseParameters["RuntimeTag"]);
+            Assert.AreEqual("User name is malformed.", ex.AdditionalResponseParameters["RuntimeContext"]); // message might change. not a big deal
+            Assert.AreEqual("ApiContractViolation", ex.AdditionalResponseParameters["RuntimeStatus"]);
+            Assert.AreEqual("3399811229", ex.AdditionalResponseParameters["RuntimeErrorCode"]);
+        }
```
bgavrilMS, line 481: @crmann1 - How about something like this? Not sure if property bag is better suited here as opposed to having new exception hierarchy - `MsalRuntimeServiceException`, `MsalRuntineUiRequiredServiceException` etc. Dedicated exceptions seem a bit overkill.

crmann1:
That will be ok, seems like we will need more for POP ect.   As long as you expose the well known property names somewhere that should work.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
bgavrilMS: CC @localden for an opinion

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -206,6 +211,19 @@ private enum ResponseStatus
             return authParams;
         }
 
+        private static void DecorateExceptionWithRuntimeErrorProperties(MsalException exception, Error runtimeError)
+        {            
+            var result = new Dictionary<string, string>()
+            {
+                { "RuntimeContext", runtimeError.Context },
+                { "RuntimeTag", runtimeError.Tag.ToString(CultureInfo.InvariantCulture)},
```
gladjohn, line 219: should we return `0x{Tag:x}` so it matches the runtime tag value? if that would be helpful for msalruntime team to debug, not sure if they are converting the string we return to the tag value when they debug issues

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
gladjohn: +1 on the property bug for extensibility of the exception handling and to be consistent across

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -98,15 +102,13 @@ private enum ResponseStatus
                         $" Error Message: {authResult.Error.Status} \n" +
                         $" WAM Error Message: {authResult.Error.Context} \n" +
                         $" Internal Error Code: {internalErrorCode} \n" +
-                        $" Is Retryable: false \n" +
                         $" Possible causes: \n" +
-                        $"- Invalid redirect uri - ensure you have configured the following url in the AAD portal App Registration: " +
-                        $"{GetExpectedRedirectUri(authenticationRequestParameters.AppConfig.ClientId)} \n" +
-                        $"- No Internet connection : \n" +
```
gladjohn, line 105: we do not need this anymore? i.e. the aka.ms link for unknown exceptions?

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -117,17 +119,20 @@ private enum ResponseStatus
...
+                        $" Internal Error Code: {internalErrorCode} \n";                        
                     logger.Error($"[WamBroker] WAM_network_error_{errorCode} {errorMessage}");
                     serviceException = new MsalServiceException(errorCode.ToString(), errorMessage);
+                    
                     serviceException.IsRetryable = true;
+                    DecorateExceptionWithRuntimeErrorProperties(serviceException, authResult.Error);
                     throw serviceException;
 
                 default:
                     errorMessage = $"Unknown {authResult.Error} (error code {errorCode}) (internal error code {internalErrorCode})";
```
gladjohn, line 131: seems like we already have all this info, should we just append the existing error codes and tags appropriately to avoid duplicating and adding a new property bag? 

image

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -135,7 +135,7 @@ private void LogEventRaised(NativeInterop.Core sender, LogEventArgs args)
             {
                 throw new MsalClientException(
                     "window_handle_required",
-                    "The new WAM implementation now needs application developers to provide parent window handle to the broker. See https://aka.ms/msal-net-wam#parent-window-handles");
+                    "A window handle must be configured. See https://aka.ms/msal-net-wam#parent-window-handles");
```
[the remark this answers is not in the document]
bgavrilMS: But the wiki page has a link to docs

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
pmaytak, line 1: Should this also be handled in `WamAdapters.HandleResponse`?

cc @gladjohn 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9e4bf560970fb5be9a72b83b3a446ce49c545576/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs#L300-L306

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -78,8 +79,9 @@ private enum ResponseStatus
             {
                 case ResponseStatus.UserCanceled:
                     logger.Error($"[WamBroker] {MsalError.AuthenticationCanceledError} {MsalErrorMessage.AuthenticationCanceled}");
-                    throw new MsalClientException(MsalError.AuthenticationCanceledError, MsalErrorMessage.AuthenticationCanceled);
-
+                    var clientEx = new MsalClientException(MsalError.AuthenticationCanceledError, MsalErrorMessage.AuthenticationCanceled);
+                    DecorateExceptionWithRuntimeErrorProperties(clientEx, authResult.Error);
+                    throw clientEx;
```
pmaytak, line 84: Bit of a code smell. Easy to miss calling Decorate method. Can this be combined into one call? An internal constructor on exception class? Although maybe an issue with Interop reference... 
Or an extension method on Exception class? Something like 
```csharp
throw new MsalClientException(MsalError.AuthenticationCanceledError, MsalErrorMessage.AuthenticationCanceled).WithAdditionalErrorData(authResultError);
```

src/client/Microsoft.Identity.Client/MsalException.cs
```diff
@@ -113,6 +114,13 @@ private set
             }
         }
 
+        /// <summary>
+        /// A property bag with extra details for this exception.
+        /// </summary>
+        public IReadOnlyDictionary<string, string> AdditionalResponseParameters { get; set; }
```
pmaytak, line 120: `AdditionalErrorData`?
`Response` implies to me that there's an error response from somewhere else (AAD, broker) but if this is intended to be a general property bag, it could be filled by MSAL itself.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -206,6 +211,19 @@ private enum ResponseStatus
             return authParams;
         }
 
+        private static void DecorateExceptionWithRuntimeErrorProperties(MsalException exception, Error runtimeError)
+        {            
+            var result = new Dictionary<string, string>()
+            {
+                { "RuntimeContext", runtimeError.Context },
```
pmaytak, line 218: What's the guideline for property bag key names? Should these be in some constant class? I assume changing the names of these afterwards will also be a breaking change?

pmaytak, line 218: In public facing API, I don't think we should use the word `Runtime` - it's a bit generic and means something specific to us. Makes sense to use it in internal code, like Runtime WAM vs Legacy WAM; but in public facing code this should be something like `BrokerErrorContext`, `BrokerErrorTag`.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
```diff
@@ -451,8 +452,35 @@ public async Task NewPOP_WithKeyIdOnly_Async()
...
+                    scopes,
+                    "noUser",
+                    "badPassword")                
+                .ExecuteAsync())
+                .ConfigureAwait(false);
+
+            Assert.AreEqual("557973642", ex.AdditionalResponseParameters["RuntimeTag"]);
+            Assert.AreEqual("User name is malformed.", ex.AdditionalResponseParameters["RuntimeContext"]); // message might change. not a big deal
+            Assert.AreEqual("ApiContractViolation", ex.AdditionalResponseParameters["RuntimeStatus"]);
+            Assert.AreEqual("3399811229", ex.AdditionalResponseParameters["RuntimeErrorCode"]);
```
pmaytak, line 480: Magic strings > flaky test? Just assert that keys/values exist?
Similar to my other comment, property bag keys should be constants.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
```diff
@@ -451,8 +452,35 @@ public async Task NewPOP_WithKeyIdOnly_Async()
                 result2.AuthenticationResultMetadata.TokenSource);
         }
 
-#if NET_CORE
-        [TestMethod]
+        [RunOn(TargetFrameworks.NetCore)]
+        public async Task WamBadROPCAsync()
```
pmaytak, line 456: nit:
```suggestion
        public async Task WamInvalidROPC_ThrowsException_TestAsync()
```

pmaytak (commented):
Design related questions.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
bgavrilMS: We can't expect partners to fetch this data by parsing error messages.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
bgavrilMS: But we could use the runtime for non-broker flows.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
bgavrilMS: I can see the point for the message, but the error codes / tag should be stable. Status as well, although new members may appear rarely.

bgavrilMS: I'm gonna leave it is as is. Makes the test easy to read and acts like a spec.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
bgavrilMS: Moved to MsalException as constants.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/PoPTests.cs
pmaytak: > I can see the point for the message,

Yea, I think per testing guidelines, changing error messages is not a breaking change, so shouldn't be asserted.

src/client/Microsoft.Identity.Client/MsalServiceException.cs
```diff
@@ -214,6 +214,7 @@ public HttpResponseHeaders Headers
         /// </summary>
         public string CorrelationId { get; set; }
 
+
```
trwalke, line 217: nit extra line

gladjohn (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -72,23 +64,24 @@ private enum ResponseStatus
...
+            logger.Verbose(() => $"[WamBroker] TelemetryData: {authResult.TelemetryData}");
 
             switch ((ResponseStatus)authResult.Error.Status)
             {
                 case ResponseStatus.UserCanceled:
                     logger.Error($"[WamBroker] {MsalError.AuthenticationCanceledError} {MsalErrorMessage.AuthenticationCanceled}");
-                    throw new MsalClientException(MsalError.AuthenticationCanceledError, MsalErrorMessage.AuthenticationCanceled);
-
+                    var clientEx = new MsalClientException(MsalError.AuthenticationCanceledError, MsalErrorMessage.AuthenticationCanceled);
+                    return clientEx;
```
pmaytak, line 74:
```suggestion
                    return new MsalClientException(MsalError.AuthenticationCanceledError, MsalErrorMessage.AuthenticationCanceled);
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -72,23 +64,24 @@ private enum ResponseStatus
...
-                    errorMessage = 
+                    errorMessage =
                         $"{WamErrorPrefix} \n" +
                         $" Error Code: {errorCode} \n" +
                         $" Error Message: {authResult.Error.Context} \n" +
                         $" Internal Error Code: {internalErrorCode} \n";
                     logger.Error($"[WamBroker] {MsalError.FailedToAcquireTokenSilentlyFromBroker} {errorMessage}");
-                    throw new MsalUiRequiredException(MsalError.FailedToAcquireTokenSilentlyFromBroker, errorMessage);
+                    var ex = new MsalUiRequiredException(MsalError.FailedToAcquireTokenSilentlyFromBroker, errorMessage);
+                    return ex;
```
pmaytak, line 84:
```suggestion
                    return new MsalUiRequiredException(MsalError.FailedToAcquireTokenSilentlyFromBroker, errorMessage);
```

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -118,16 +108,18 @@ private enum ResponseStatus
...
-                    throw serviceException;
+                    return serviceException;
 
                 default:
                     errorMessage = $"Unknown {authResult.Error} (error code {errorCode}) (internal error code {internalErrorCode})";
-                    logger.Verbose(()=>$"[WamBroker] {MsalError.UnknownBrokerError} {errorMessage}");
-                    throw new MsalServiceException(MsalError.UnknownBrokerError, errorMessage);
+                    logger.Verbose(() => $"[WamBroker] {MsalError.UnknownBrokerError} {errorMessage}");
+                    var ex2 = new MsalServiceException(MsalError.UnknownBrokerError, errorMessage);
+                    return ex2;
```
pmaytak, line 122:
```suggestion
                    return new MsalServiceException(MsalError.UnknownBrokerError, errorMessage);
```

pmaytak (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
bgavrilMS: The compiler can optimize this. I like to separate them, makes it easier to debug.

PR #4128, opened 2023-05-10
Add API and logic for ignoring cached OBO assertion for Initiate long OBO

### LINKED ISSUE #4124: [Feature Request] Add API for InitiateLongRunningProcessInWebApi to ignore cached assertion
In 4.50.0 and below, `InitiateLongRunningProcessInWebApi` always checks the cache for access or refresh tokens and when matching it ignores the assertion hash in the request. In 4.51.0 - 4.53.0, `InitiateLongRunningProcessInWebApi`was updated to not check the cache at all and just use the request user assertion to retrieve tokens from AAD. In 4.54.0 the proposed change is to make `InitiateLongRunningProcessInWebApi`check the cache again but also match based on request and cached user assertions in token. There's a request to add the legacy behavior.

Add `WithIgnoreCachedAssertion(bool)` API. In `OnBehalfOfRequest` class, use the flag to ignore the assertion check for access and refresh tokens.

pmaytak:
Fixes #4124

Adds new API `WithIgnoreCachedAssertion`.
Only affects `InitiateLongRunningProcessInWebApi`.
When enabled, mimics MSAL 4.50.0 and below behavior - does not check cached tokens based on OBO assertions.
When disabled (default behavior), cached tokens will only be returned if the OBO assertion in the request matched the assertion of the cached token.

**Testing**
Adding.

**Performance impact**
None.

**Documentation**
- [ ] All relevant documentation is updated. - After release

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
@@ -163,6 +163,19 @@ namespace Microsoft.Identity.Client
             return this;
         }
 
+        /// <summary>
+        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - does not check cached tokens based on OBO assertions.
+        /// When disabled (default behavior), cached tokens will only be returned if the OBO assertion in the request matched the assertion of the cached token.
+        /// </summary>
+        /// <param name="ignoreCachedAssertion">Whether to match on cached OBO assertion hash.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenOnBehalfOfParameterBuilder WithIgnoreCachedAssertion(bool ignoreCachedAssertion = true)
+        {
+            Parameters.IgnoreCachedOboAssertion = ignoreCachedAssertion;
+            return this;
+        }
+
         /// <inheritdoc />
         internal override Task<AuthenticationResult> ExecuteInternalAsync(CancellationToken cancellationToken)
         {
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
```diff
@@ -161,6 +161,19 @@ public AcquireTokenOnBehalfOfParameterBuilder WithCcsRoutingHint(string userName
...
+        }
+
+        /// <summary>
+        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - does not check cached tokens based on OBO assertions.
+        /// When disabled (default behavior), cached tokens will only be returned if the OBO assertion in the request matched the assertion of the cached token.
+        /// </summary>
+        /// <param name="ignoreCachedAssertion">Whether to match on cached OBO assertion hash.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenOnBehalfOfParameterBuilder WithIgnoreCachedAssertion(bool ignoreCachedAssertion = true)
```
pmaytak, line 173: This ends up being exposed on `AcquireTokenOnBehalfOfParameterBuilder` and is visible for normal OBO.

bgavrilMS (approved):

bgavrilMS:
Do we need the other fix for OBO as well, where we check the actual token assertion?

pmaytak:
> Do we need the other fix for OBO as well, where we check the actual token assertion?

Yes, because currently we don't check the cache at all and that PR adds cache checking.

trwalke:
I am still trying to wrap my head around why this is ultimately needed. I dont think we should add additional api to `InitiateLongRunningOBO` just to maintain its previous functionality when it was used incorrectly. Lets discuss in our sync tomorrow.

pmaytak:
Closing in favour of #4135

PR #4135, opened 2023-05-12
Add public API to search in cache for Initiate long-running OBO

### LINKED ISSUE #4124: [Feature Request] Add API for InitiateLongRunningProcessInWebApi to ignore cached assertion

pmaytak:
Fixes #4124
Related to #4111

Adds new API `WithSearchInCacheForLongRunningProcess`.
Only affects `InitiateLongRunningProcessInWebApi`.
When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens.
When disabled (default behavior), doesn't search in cache, but uses the user assertion to retrieve tokens from AAD.

**Testing**
Unit test. Manual.

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Identity.Client
     public sealed class AcquireTokenOnBehalfOfParameterBuilder :
         AbstractConfidentialClientAcquireTokenParameterBuilder<AcquireTokenOnBehalfOfParameterBuilder>
     {
-        private AcquireTokenOnBehalfOfParameters Parameters { get; } = new AcquireTokenOnBehalfOfParameters();
+        internal AcquireTokenOnBehalfOfParameters Parameters { get; } = new AcquireTokenOnBehalfOfParameters();
 
         /// <inheritdoc />
         internal AcquireTokenOnBehalfOfParameterBuilder(IConfidentialClientApplicationExecutor confidentialClientApplicationExecutor)
@@ -160,20 +160,6 @@ namespace Microsoft.Identity.Client
             };
 
             this.WithExtraHttpHeaders(ccsRoutingHeader);
-            return this;
-        }
-
-        /// <summary>
-        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
-        /// When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens first, 
-        /// and if not found, then uses user assertion to request new tokens from AAD.
-        /// When disabled (default behavior), doesn't search in cache, but uses the user assertion to retrieve tokens from AAD.
-        /// </summary>
-        /// <param name="searchInCache">Whether to match on cached OBO assertion hash.</param>
-        /// <returns>The builder to chain the .With methods</returns>
-        public AcquireTokenOnBehalfOfParameterBuilder WithSearchInCacheForLongRunningProcess(bool searchInCache = true)
-        {
-            Parameters.SearchInCacheForLongRunningObo = searchInCache;
             return this;
         }
 
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
@@ -163,6 +163,19 @@ namespace Microsoft.Identity.Client
             return this;
         }
 
+        /// <summary>
+        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens.
+        /// When disabled (default behavior), doesn't search in cache, but uses the user assertion to retrieve tokens from AAD.
+        /// </summary>
+        /// <param name="searchInCache">Whether to match on cached OBO assertion hash.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenOnBehalfOfParameterBuilder WithSearchInCache(bool searchInCache = true)
+        {
+            Parameters.SearchInCacheForLongRunningObo = searchInCache;
+            return this;
+        }
+
         /// <inheritdoc />
         internal override Task<AuthenticationResult> ExecuteInternalAsync(CancellationToken cancellationToken)
         {
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
@@ -165,12 +165,13 @@ namespace Microsoft.Identity.Client
 
         /// <summary>
         /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
-        /// When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens first, 
+        /// and if not found, then uses user assertion to request new tokens from AAD.
         /// When disabled (default behavior), doesn't search in cache, but uses the user assertion to retrieve tokens from AAD.
         /// </summary>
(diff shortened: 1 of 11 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
```diff
@@ -161,6 +161,19 @@ public AcquireTokenOnBehalfOfParameterBuilder WithCcsRoutingHint(string userName
...
+        }
+
+        /// <summary>
+        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens.
+        /// When disabled (default behavior), doesn't search in cache, but uses the user assertion to retrieve tokens from AAD.
+        /// </summary>
+        /// <param name="searchInCache">Whether to match on cached OBO assertion hash.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenOnBehalfOfParameterBuilder WithSearchInCache(bool searchInCache = true)
```
pmaytak, line 173: This ends up being exposed on `AcquireTokenOnBehalfOfParameterBuilder` and is visible for normal OBO.

Better naming?
`WithSearchInCacheForLongRunningObo`?
Something else?

bgavrilMS: WithSearchInCacheForLongRunningObo lgtm

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
```diff
@@ -161,6 +161,20 @@ public AcquireTokenOnBehalfOfParameterBuilder WithCcsRoutingHint(string userName
...
+
+        /// <summary>
+        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens first, 
+        /// and if not found, then uses user assertion to request new tokens from AAD.
+        /// When disabled (default behavior), doesn't search in cache, but uses the user assertion to retrieve tokens from AAD.
+        /// </summary>
+        /// <param name="searchInCache">Whether to match on cached OBO assertion hash.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenOnBehalfOfParameterBuilder WithSearchInCacheForLongRunningProcess(bool searchInCache = true)
```
trwalke, line 174: Not sure this is something we should advertise since it makes the oth api obsolete. I think it is better to hide this with EditorBrowsable = false

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenOnBehalfOfParameterBuilder.cs
```diff
@@ -161,6 +161,20 @@ public AcquireTokenOnBehalfOfParameterBuilder WithCcsRoutingHint(string userName
 
             this.WithExtraHttpHeaders(ccsRoutingHeader);
             return this;
+        }
+
+        /// <summary>
```
trwalke, line 166: The comments for these new apis should be updated to clearly indicate how the apis should be used. Basically warn against the api for anything other than backwards compatability. ALso, the wiki should be updated to warn against using `InitiateLongRunningProcessInWebApi` over and over again.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -347,6 +347,103 @@ public async Task AcquireTokenByObo_InitiateLongRunningProcessInWebApi_CacheKeyA
             }
         }
 
+        [TestMethod]
```
bgavrilMS, line 350: Assume this is regression test. Pls add a comment with the a link to the bug.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/OBOTests.cs
```diff
@@ -347,6 +347,103 @@ public async Task AcquireTokenByObo_InitiateLongRunningProcessInWebApi_CacheKeyA
...
+                // Token's not in the cache, searched by user assertion hash, retrieved from IdP, saved with the provided OBO cache key
+                var result = await cca.InitiateLongRunningProcessInWebApi(TestConstants.s_scope, TestConstants.DefaultAccessToken, ref oboCacheKey)
+                    .ExecuteAsync().ConfigureAwait(false);
+
+                Assert.AreEqual(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+                Assert.AreEqual(CacheRefreshReason.NotApplicable, result.AuthenticationResultMetadata.CacheRefreshReason);
+                userCacheAccess.AssertAccessCounts(0, 1);
+
+                // Initiate with different user assertion and IgnoreCachedAssertion flag
+                // MSAL will not match on the assertion and return cached token
```
neha-bhargava, line 374:
```suggestion
                // MSAL will match on the assertion and return cached token
```
Shouldn't the assertion match and return cached token?

neha-bhargava: Ignore this comment.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenOnBehalfOfParameters.cs
```diff
@@ -9,13 +10,21 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
...
         public UserAssertion UserAssertion { get; set; }
         /// <summary>
         /// User-provided cache key for long-running OBO flow.
         /// </summary>
         public string LongRunningOboCacheKey { get; set; }
+
+        /// <summary>
+        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - does not check cached tokens based on OBO assertions.
+        /// When disabled (default behavior), cached tokens will only be returned if the OBO assertion in the request matched the assertion of the cached token.
```
neha-bhargava, line 24: So the cache will always be searched? The difference is whether the assertion is matched or not. I think if this is the case we can name it something like `SearchInCacheUsingAssertion`?

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenOnBehalfOfParameterBuilderExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    /// Extension methods for the <see cref="AcquireTokenOnBehalfOfParameterBuilder" />
+    /// </summary>
+    public static class AcquireTokenOnBehalfOfParameterBuilderExtensions
+    {
+        /// <summary>
+        /// Only affects <see cref="ILongRunningWebApi.InitiateLongRunningProcessInWebApi(IEnumerable{string}, string, ref string)"/>.
+        /// When enabled, mimics MSAL 4.50.0 and below behavior - checks in cache for cached tokens first, 
+        /// and if not found, then uses user assertion to request new tokens from AAD.
+        /// When disabled (default behavior), doesn't search in cache, but uses the user assertion to retrieve tokens from AAD.
```
neha-bhargava, line 18: This does not match with what is mentioned above. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4135/files#diff-8510905684539ef8e938512ce363f8a33ab5f441a268981b07638be16bd3385aR24

neha-bhargava (approved):
Looks good. I think the comment here needs to be updated: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4135/files#diff-8510905684539ef8e938512ce363f8a33ab5f441a268981b07638be16bd3385aR24

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenOnBehalfOfParameters.cs
pmaytak: This is an outdated comment. Updated.

PR #4141, opened 2023-05-18
Use sub claim if oid is missing

### LINKED ISSUE #4140: [Bug] SerializeAdalV3 returns null when Id Token has no oid claim
Scenario: web site

This is an ADAL 2 MSAL migration issue. Partner is using `SerializeAdalV3` logic which in some cases returns null. This caused the partner to revert back from MSAL to ADAL.

 It looks like in rare cases the IdToken does not have `oid` claim. ADAL uses `sub` claim when `oid` is not available.

bgavrilMS:
Fixes #4140

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthenticationResult.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResult.cs
@@ -211,9 +211,16 @@ namespace Microsoft.Identity.Client
         public bool IsExtendedLifeTimeToken { get; }
 
         /// <summary>
-        /// Gets the Unique Id of the account. It can be null. When the <see cref="IdToken"/> is not <c>null</c>, this is its ID, that
-        /// is its ObjectId claim, or if that claim is <c>null</c>, the Subject claim.
+        /// Gets the Unique Id of the account in this <see cref="TenantId">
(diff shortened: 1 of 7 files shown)


jmprieur (commented):
LGTM
Was GetUniqueId() already defined?

bgavrilMS:
> LGTM Was GetUniqueId() already defined?

Yeah, it's used to compute `AuthenticationResult.UniqueId` which is public API. I changed the comments on it.

gladjohn:
```cs
public string GetUniqueId()
{
    return ObjectId ?? Subject;
}
```

what does subject return?

neha-bhargava (approved):

pmaytak (approved):

trwalke (approved):

rayluo (approved):

jmprieur (approved):

bgavrilMS:
> ```cs
> public string GetUniqueId()
> {
>     return ObjectId ?? Subject;
> }
> ```
> 
> what does subject return?

sub claim. I added comments everywhere on this. Seems to be mandatory as per OIDC spec. Should be an ID for app + user, but  can theoretically be different for app2 + user. Stable enough for token caching for sure.

PR #4169, opened 2023-06-06
Update docs #4166

### LINKED ISSUE #4166: [Feature Request] WWWAuthenticate Resource property obsolete with warning instead of error
In #3160 the `Resource` property in `WWWAuthenticateParameters` was marked as obsolete. There was a question on the PR as to whether the `ObsoleteAttribute.IsError` property should be set to `true`, and the decision was made to do it, with the rationale that callers should know the resource they are requesting.

The issue is that there are scenarios, such as [this security advisory from Azure Key Vault](https://devblogs.microsoft.com/azure-sdk/guidance-for-applications-using-the-key-vault-libraries/#additional-recommended-actions), where clients are supposed to inspect the `resource` parameter in the response to ensure it matches the set of well known domains.

Because the `ObsoleteAttribute` now throws a compiler error, it doesn't allow consumers to decide on how they want to address this issue - it can't be suppressed as a warning could be.

I'd like the `IsError` property to be set to the default of `false`.

**Describe alternatives you've considered**
I've tried various ways of suppressing the error - other than conditional compilation I don't know of a way, and that certainly seems like overkill.

**Additional context**

bgavrilMS:
Fixes #4166

Docs only

### DIFF
--- a/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
+++ b/src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
@@ -34,6 +34,13 @@ namespace Microsoft.Identity.Client
         /// Resource for which to request scopes.
         /// This is the App ID URI of the API that returned the WWW-Authenticate header.
         /// </summary>
+        /// <remarks>
+        /// Clients that perform resource validation (e.g. by comparing the host part of the resource against a list of known good hosts), 
+        /// can still use the indexer to retrieve the raw value of the resource / scope.
+        /// 
+        /// If a resource is used, add "/.default" to it to transform it into a scope, e.g. "https://graph.microsoft.com/.default" is the OAuth2 scope for "https://graph.microsoft.com" resource.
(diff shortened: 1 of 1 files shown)


trwalke (approved):

src/client/Microsoft.Identity.Client/WwwAuthenticateParameters.cs
```diff
@@ -34,13 +34,27 @@ public class WwwAuthenticateParameters
         /// Resource for which to request scopes.
         /// This is the App ID URI of the API that returned the WWW-Authenticate header.
         /// </summary>
+        /// <remarks>
+        /// Clients that perform resource validation (e.g. by comparing the host part of the resource against a list of known good hosts),
```
pmaytak, line 38: Should this be added as a code example to the [docs ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/WWW-Authenticate-parameters#how-to---use-the-wwwauthenticateparameters-class)too?

PR #4183, opened 2023-06-21
Minor updates to WAM logging.

pmaytak:
Fixes #

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Identity.Client.Platforms.Features.RuntimeBroker
             }
             catch (MsalRuntimeException ex) when (ex.Status == ResponseStatus.ApiContractViolation)
             {
-                // failed to initialize msal runtime - can happen on older versions of Windows. Means broker is not available.
+                // failed to initialize MSAL runtime - can happen on older versions of Windows. Means broker is not available.
                 // We will never get here with our current OS version check. Instead in this scenario we will fallback to the browser
                 // but MSALRuntime does it's internal check for OS compatibility and throws an ApiContractViolation MsalRuntimeException.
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -536,25 +535,25 @@ public bool IsBrokerInstalledAndInvokable(AuthorityType authorityType)
         {
             if (!DesktopOsHelper.IsWin10OrServerEquivalent())
             {
-                _logger?.Warning("[WAM Broker] Not a supported operating system. WAM broker is not available. ");
+                _logger?.Warning("[RuntimeBroker] Not a supported operating system. WAM broker is not available. ");
```
rayluo, line 538: FWIW, we hide both WAM and Runtime concepts in MSAL Python, so the parameter and logs all mention "broker" only.

This PR LGTM.

pmaytak: Earlier I found it a bit difficult sometimes to figure out from logs which WAM implementation was used. Now that WAM runtime is used everywhere except UWP, it's less of an issue.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -67,7 +67,7 @@ internal class RuntimeBroker : IBroker
             }
             catch (Exception ex)
             {
-                // When MSAL Runtime dlls fails to load then we catch the exception and throw with a meaningful
+                // When MSAL Runtime DLL fails to load then we catch the exception and throw with a meaningful
```
trwalke, line 70:
```suggestion
                // When the MSAL Runtime DLL fails to load then we catch the exception and throw with a meaningful
```

trwalke (approved):

gladjohn (approved):

Issue #4060, opened 2023-04-03
[Feature Request] Add support for STI 


bgavrilMS:
STI is a Microsoft internal token issuer. We want to support it client_credentials flow only, via extensibility. 

Proposed dev ex:

```csharp
var cca = ConfidentialClientApplicationBuilder.Create(client_id)
                        .WithGenericAuthority("https://sti_authority", tokenRequestContentType: ContentType.JSON ) 
                        .WithCacheOptions( extraTokenParamsToCache: new[] { "topology", "trusted_keys" }) 
                        .Build();

var authResult = cca.AcquireTokenForClient(new[] { scope })
     .OnBeforeTokenRequest( data => data.BodyParameters.Add("param1", "val1") ) // API exists today
     .ExecuteAsync();

```

This will require a few changes in MSAL: 

- [x] Ability to change the way MSAL does a POST request from UrlForms to JSON. Can be a parameter on `WithGenericAuthority`

Idea: could be interesting to have this on all authority. AAD should support this, MSAL did it at some point but stopped for reasons we don't know.

- [x] MSAL must be able to cache additional properties from the /token response. See `WithCacheOptions` proposal.
- [x] `AuthenticationResult.AdditionalResponseParameters` should expose the additional cached properties
- [x] Add support for non-scalar properties in `AuthenticationResult.AdditionalResponseParameters` (values can be JSON strings)

References: [fiddler trace of an STI req](https://microsofteur-my.sharepoint.com/:u:/g/personal/bogavril_microsoft_com/ESkuu7FCm_pLkb-IUagMrX8BUIBiCsWQbFnwKi6fIAKDAw?e=oxXza2)

bgavrilMS:
@jmprieur @sruke - this is a design for the changes needed for MSAL to support STI. Please have a look and see what needs to be done for `ITokenAcquisition`.

jmprieur:
LGTM, @bgavrilMS 
From the point of view of IdWeb, I would say there could propose a new interface in Id.Web like the following (it's just an idea, the types might not be right). This interface would be optional, but understood by TokenAcquisition.

```csharp
interface IIdentityProviderAdapter
{
 bool AppliesTo(string authority);
 void AdaptAppCreation(ConfidentialClientApplicationBuilder builder);
 void AdaptTokenAcquisition(AcquireTokenForClientBuilder builder);
}
```

[...]

PR #4192, opened 2023-06-28
Sti support

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4060

**Testing**

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -39,6 +41,17 @@ public CacheOptions(bool useSharedCache)
             UseSharedCache = useSharedCache;
         }
 
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="useSharedCache">Set to true to share the cache between all ClientApplication objects. The cache becomes static. <see cref="UseSharedCache"/> for a detailed description. </param>
+        /// <param name="extraTokenParamsToCache">Specifies the extra parameters MSAL should cache</param>
+        public CacheOptions(bool useSharedCache, IEnumerable<string> extraTokenParamsToCache)
```
bgavrilMS, line 49: We don't need a ctor for every public property.

src/client/Microsoft.Identity.Client/AppConfig/TokenRequestContentType.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// The media content type for post requests made to the identity provider.
+    /// </summary>
+    public class TokenRequestContentType
+    {
+        internal string _value;
```
bgavrilMS, line 18: private readonly

src/client/Microsoft.Identity.Client/AppConfig/TokenRequestContentType.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        internal TokenRequestContentType(string value) { _value = value; }
+
+        /// <summary>
+        /// JSON media content type
+        /// </summary>
+        public static TokenRequestContentType JSON { get { return new TokenRequestContentType("application/json"); } }
```
bgavrilMS, line 25: I think you should add an option for x-www-form-url which is the default.

https://stackoverflow.com/questions/9870523/what-are-the-differences-between-application-json-and-application-x-www-form-url

src/client/Microsoft.Identity.Client/AppConfig/TokenRequestContentType.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.AppConfig
+{
+    /// <summary>
+    /// The media content type for post requests made to the identity provider.
+    /// </summary>
+    public class TokenRequestContentType
```
bgavrilMS, line 16: Why a class and not an enum?

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
```diff
@@ -50,5 +63,9 @@ public CacheOptions(bool useSharedCache)
         /// </remarks>
         public bool UseSharedCache { get; set; }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        public IEnumerable<string> ExtraTokenParamsToCache { get; set; }
```
bgavrilMS, line 69: I suggest you use a topic branch e.g. "travis/sti" and create small individual PRs for all of these improvements. ExtraTokenParams should be separated from JSON.

src/client/Microsoft.Identity.Client/AppConfig/TokenRequestContentType.cs
```diff
@@ -12,11 +13,20 @@ namespace Microsoft.Identity.Client.AppConfig
...
     {
+        internal string _value;
+
+        internal TokenRequestContentType(string value) { _value = value; }
+
         /// <summary>
         /// JSON media content type
         /// </summary>
-        public const string JSON = "application/json";
+        public static TokenRequestContentType JSON { get { return new TokenRequestContentType("application/json"); } }
```
bgavrilMS, line 25: Avoid creating new objects (strings) where possible. This  a string constant that can be reused. 

In any case, an enum is much simpler to reason with here.

src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs
bgavrilMS, line 69: Naming - "ExtraPayloadParameters" to be consistent with "ExtraQueryParameters" ? (and just a comment that it's only applied to /token). 

Please involve @localden in all public API discussions.

bgavrilMS, line 69: I recommend using `IReadonlyList` instead of `IEnumerable`. It's readonly and operations such as "Count" are guarateed to be fast, whereas IEnumerable.Count() is more complicated.

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -283,6 +288,22 @@ public sealed partial class TokenCache : ITokenCacheInternal
             }
         }
 
+        private Dictionary<string, string> GetExtraTokenParametersFromResponse(IEnumerable<string> extraTokenParams, MsalTokenResponse response)
+        {
+            Dictionary<string, string> extraTokenParamsDict = new Dictionary<string, string>();
+
+            foreach (string tokenParam in extraTokenParams)
+            {
+                if (response.ExtensionData.Keys.Contains(tokenParam))
```
bgavrilMS, line 297: Use 

```csharp
if (dict.TryGetValue(keyName, out var value))
{
   // do something with the value
}

File: src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs [hunk 8]
```diff
@@ -283,6 +288,22 @@ public sealed partial class TokenCache : ITokenCacheInternal
             }
         }
 
+        private Dictionary<string, string> GetExtraTokenParametersFromResponse(IEnumerable<string> extraTokenParams, MsalTokenResponse response)
+        {
+            Dictionary<string, string> extraTokenParamsDict = new Dictionary<string, string>();
```
bgavrilMS, line 293: If you check if `extraTokenParams` is not empty first, you can return an empty dictionary (see CollectionHelper for that). This will avoid allocating a new dictionary every time.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -21,28 +22,37 @@ namespace Microsoft.Identity.Test.Unit.CoreTests.InstanceTests
     public class GenericAuthorityTests : TestBase
     {
         [DataTestMethod]
-        [DataRow(true)]
-        [DataRow(false)]
+        [DataRow(true, false)]
```
bgavrilMS, line 25: Consider creating new tests. Adding more and more options to existing tests makes them hard to read.

bgavrilMS, line 25: We should have a unit test explicitly done based on the STI trace.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -21,28 +22,37 @@ namespace Microsoft.Identity.Test.Unit.CoreTests.InstanceTests
     public class GenericAuthorityTests : TestBase
     {
         [DataTestMethod]
```
bgavrilMS, line 24: You haven't implemented / tested the fact that the `ExtraParamsToCache` should appear in the `AuthenticationResult.AdditionalData`

PR #4243, opened 2023-07-13
Add open telemetry metrics to MSAL

### LINKED ISSUE #4229: [Epic]  Supportability: Add support for open telemetry to log client side data.
- [ ]  Log data points to open telemetry which are currently [logged with ITelemetryClient](https://microsoft-my.sharepoint-df.com/:w:/p/nebharg/EWrRAFHR6VdGvx5i_-uCZiUBROlxzCXZzLdRUJ8ePTKkgw?e=RDxtgw)
- [ ] Capture metrics for success counter with tags client id, token source and cache refresh reason.
- [ ] Capture metrics for failure counter with tags client id, error code.
- [ ]  Add API to opt in to open telemetry.
- [ ] Add database to log the data points to.
- [ ] Create dashboards using the data logged.

neha-bhargava:
Fixes #4229 
Add open telemetry 

* Add success counter with tags client id, token source and cache refresh reason
* Add failure counter with tags client id and error code or exception name in case the exception is not MsalException

**Testing**

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,20 +101,40 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                     LogMetricsFromAuthResult(authenticationResult, AuthenticationRequestParameters.RequestContext.Logger);
                     LogSuccessfulTelemetryToClient(authenticationResult, telemetryEventDetails, telemetryClients);
 
+#if NET6_0_OR_GREATER
```
bgavrilMS, line 104: You can avoid having these compliation flags in the code by putting everything in a class, smth like `OpenTelemetryWrapper`

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,20 +101,40 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+#endif
+
                     return authenticationResult;
                 }
                 catch (MsalException ex)
                 {
                     apiEvent.ApiErrorCode = ex.ErrorCode;
                     AuthenticationRequestParameters.RequestContext.Logger.ErrorPii(ex);
                     LogMsalErrorTelemetryToClient(ex, telemetryEventDetails, telemetryClients);
+#if NET6_0_OR_GREATER
```
bgavrilMS, line 119: You can move this into the OpenTelemetry class or into some wrapper to avoid having `#if xyz` everywhere

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -303,6 +303,10 @@
     <PackageReference Include="System.Security.SecureString" Version="4.3.0" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(TargetFrameworkNet6)' ">
```
bgavrilMS, line 306: It will be interesting to see what happens on net461, as many apps will be using net461. If you reference System.Diagnostics.DiagnosticSource from net461, it will resolve to the netstandard version of this package. This may or may not actually have an implemenation ... it may just do nothing...

If this is the case, it's a good argument to bump MSAL support from 4.6.1 to 4.6.2

neha-bhargava: Open telemetry does not work with net461 but works with net462 onwards. But there is no compile time error if we keep the net461 and add the package `System.Diagnostics.DiagnosticSource` version 7.0.2. It just gives a warning that this package is not tested on net461. But when a developer will add an exporter, open telemetry might not log anything for net461

neha-bhargava: image

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +106,106 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
                 catch (MsalException ex)
                 {
                     apiEvent.ApiErrorCode = ex.ErrorCode;
                     AuthenticationRequestParameters.RequestContext.Logger.ErrorPii(ex);
                     LogMsalErrorTelemetryToClient(ex, telemetryEventDetails, telemetryClients);
+
+                    OpenTelemetryClient.LogToFailureCounter(
+                        ServiceBundle.PlatformProxy.GetProductName(), 
+                        AuthenticationRequestParameters.AppConfig.ClientId, 
+                        ex.ErrorCode);
```
jmprieur, line 128: What will it be in the case of an MsalClientException? can there be such an exception?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +106,106 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+
                     return authenticationResult;
                 }
                 catch (MsalException ex)
                 {
                     apiEvent.ApiErrorCode = ex.ErrorCode;
                     AuthenticationRequestParameters.RequestContext.Logger.ErrorPii(ex);
                     LogMsalErrorTelemetryToClient(ex, telemetryEventDetails, telemetryClients);
+
+                    OpenTelemetryClient.LogToFailureCounter(
```
jmprieur, line 125: Don't you want to use the same name (Metric instead of Counter)?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +106,106 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
                 }
                 finally
                 {
+                    activity?.Stop();
                     telemetryClients.TrackEvent(telemetryEventDetails);
                 }
             }
         }
 
+        private void LogMsalSuccessTelemetryToOTel(AuthenticationResult authenticationResult)
```
jmprieur, line 152: This is pretty verbose, and probably not something we'd send to the ID4S store?
is it more for partners?

jmprieur (commented):
Thanks @neha-bhargava 
I've added a couple of questions/comments

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: It should be like this; 

- MsalServiceException -> the suberror code (e.g. 500015) or if not present the error code
- MsalClientException -> the error code (e.g. invalid_certificate)
- other exceptions -> the type of the exception (e.g. SocketTimeout)

neha-bhargava: All the MsalExceptions have error code so the behavior will be as mentioned above by Bogdan.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: The counters are built using the meter. So all the counters and histograms that we have currently will be registered to the meter `ID4S_Msal.Net.Meter`. So the meter will have a success counter, failure counter and histograms based on total duration, duration in cache and duration in http.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: Correct, this is not for ID4S and developers can add an exporter to get this data in their own database. This is all the data we log with ITelemetryClient and was discussed with E&D team.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup Label="Targets required for unit tests to run">
     <TargetFrameworkNetDesktop461>net461</TargetFrameworkNetDesktop461>
-    <TargetFrameworkNetCore>netcoreapp2.1</TargetFrameworkNetCore>
+    <TargetFrameworkNetCore>netcoreapp3.1</TargetFrameworkNetCore>
```
GeoK, line 5: Is there a conflict between System.Diagnostics.DiagnosticSource and netcoreapp2.1 or bump to 3.1 is unrelated?

tests/devapps/TestTelemetry/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+using OpenTelemetry.Trace;
+
+internal class Program
+{
+    static async Task Main(string[] args)
+    {
+        Console.WriteLine("Hello World!");
+
+        using var meterProvider = Sdk.CreateMeterProviderBuilder()
+            .AddMeter("MSAL.Net.Meter")
```
GeoK, line 19: Does it still work considering that this name doesn't match the name on the actual meter?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -79,6 +82,11 @@ private void ValidateScopeInput(ISet<string> scopesToValidate)
         public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationToken = default)
         {
             Stopwatch sw = Stopwatch.StartNew();
+            activity = OpenTelemetryClient.AcquireTokenActivity.StartActivity("Token Acquisition", ActivityKind.Internal);
```
GeoK, line 85: TBD: We'd need to prefix activity/counter names and also satisfy schema requirements 
_Name must use alphanumeric characters ([a-zA-Z0-9]) and must start with a capital letter ([A-Z])._

GeoK: same for counters, e.g. acquire_token_success.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
neha-bhargava: Yes, the System.Diagnostics.DiagnosticSource 7.0.2 version needed is only supported for 3.1 onwards.

tests/devapps/TestTelemetry/Program.cs
neha-bhargava: Missed to check in the updated file. It will not work with another name.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
bgavrilMS: In any case, we should: 

- bump to net fwk 4.6.2
- remove netcoreapp 2.1 (rely on netstandard and net6)
- remove net45

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
jmprieur: Who would that be for? customers?
We would not want to send activities to the centralized telemetry. Only meters/histograms

neha-bhargava: Yes, this will be for the customers. For example E&D if they want to capture this data from MSAL which is currently exposed using ITelemetryClient.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
neha-bhargava: Created tasks to track this work

[[Engineering Task] Remove support for netcoreapp 2.1 · Issue #4313 · AzureAD/microsoft-authentication-library-for-dotnet (github.com)](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4313)
[[Engineering Task] Remove support for net45 · Issue #4314 · AzureAD/microsoft-authentication-library-for-dotnet (github.com)](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4314)
[[Engineering Task] Bump net 461 to net 462 · Issue #4315 · AzureAD/microsoft-authentication-library-for-dotnet (github.com)](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4315)

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/OpenTelemetryClient.cs
```diff
@@ -70,25 +76,57 @@ public class OpenTelemetryClient
...
+            foreach (KeyValuePair<string, object> tag in tags)
+            {
+                s_activity.Value?.AddTag(tag.Key, tag.Value);
+            }
+#endif
+        }
+
+        void IOpenTelemetryClient.LogActivityStatus(bool success)
+        {
+#if NETSTANDARD || NET6_0 || NET462
```
bgavrilMS, line 94: There is a more elegant way of doing this - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/build/platform_and_feature_flags.props

(also, enable it on NET6-Windows)

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -55,6 +53,8 @@ internal abstract class RequestBase
                 throw new ArgumentNullException(nameof(acquireTokenParameters));
             }
 
+            OpenTelemetryClient = (OpenTelemetryClient)OpenTelemetryFactory.CreateClient();
```
bgavrilMS, line 56: Can you use the platform proxy instead? It's already a factory.

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/OpenTelemetryClient.cs
```diff
@@ -1,20 +1,20 @@
...
-using Microsoft.Identity.Client.Internal.Requests;
-//#endif
 
 namespace Microsoft.Identity.Client.TelemetryCore.OpenTelemetry
 {
     /// <summary>
     /// Class to hold the OpenTelemetry objects used by MSAL.
     /// </summary>
-    public class OpenTelemetryClient
+    internal class OpenTelemetryClient : IOpenTelemetryClient
```
bgavrilMS, line 17: A more interesting pattern is to have the a `NullOtelClient` and this one, both implementing `IOpenTelemetryClient`. 

Then PlatformProxy would either create a `NullOtelClient` or a `OpenTelemetryClient` depending on the target

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
bgavrilMS: Also, downcasting is a code smell. Why do you need the actual obj?

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/IOpenTelemetryClient.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.TelemetryCore.OpenTelemetry
+{
+    internal interface IOpenTelemetryClient
```
bgavrilMS, line 10: Is "Client" a bit overloaded word here? Telemetry is more a consumer producer pattern.

neha-bhargava: Updated the name to OTelInstrumentation

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -270,4 +270,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.IdentityModel.Abstractions" Version="6.22.0" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(TargetFrameworkNet6)' Or '$(TargetFramework)' == '$(TargetFrameworkNet6Win)' Or '$(TargetFramework)' == '$(TargetFrameworkNetStandard)' Or '$(TargetFramework)' == '$(TargetFrameworkNetDesktop462)' ">
```
bgavrilMS, line 273: I would just add it each individual ItemGroup (there is one already for each of them).

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,194 @@
...
+        public const string ActivitySourceName = "MicrosoftIdentityClient_Activity";
+
+        private const string SuccessCounterName = "MsalSuccess";
+        private const string FailedCounterName = "MsalFailed";
+        private const string TotalDurationHistogramName = "MsalTotalDuration.1A";
+        private const string TotalDurationInUsHistogramName = "MsalTotalDurationInUs.1A";
+        private const string DurationInCacheHistogramName = "MsalDurationInCache.1A";
+        private const string DurationInHttpHistogramName = "MsalDurationInHttp.1A";
+
+#if SUPPORTS_OTEL
```
bgavrilMS, line 37: A different way of achieving this is to create a feature in the Feature directory. Then each TFM that supports OTel inclues this. You probably won't even need "SUPPORTS_OTEL" flag. In code, Only the PlatformProxy will know who supports OTEL and who doesn't.

bgavrilMS (approved):

bgavrilMS:
I can only imagine that the error that you're getting when running unit tests is due to a bug in how the test framework discovers the test runner. We use MSTest, but the test runner attempts to run tests with NUnit first. In the past, we managed to "fix" these type of isseus by upgrading the test infrastructure packages e.g. Microsoft.NET.Test.Sdk. 

Have a look at Microsoft.Identity.Test.Integration.NetCore.csproj and Microsoft.Identity.Test.Integration.NetFx.csproj

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/OtelInstrumentation.cs
neha-bhargava: What is feature directory? How can this be achieved? It does make sense to have this only for supported versions.

bgavrilMS: Look at MSAL/platforms/features. A good example is the `DefaultOSBrowser` feature, which is available on desktop versions of .NET only and not on mobile, or `RuntimeBroker` feature, which is available in the package only on net6-win

tests/Microsoft.Identity.Test.Common/Core/Helpers/OTelInstrumentationUtil.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Text;
+using Microsoft.Identity.Client.Platforms.Features.OpenTelemetry;
+using Microsoft.Identity.Client.TelemetryCore;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using OpenTelemetry;
+using OpenTelemetry.Metrics;
+
+namespace Microsoft.Identity.Test.Common.Core.Helpers
+{
+    public class OTelInstrumentationUtil
```
jmprieur, line 17: why is it public? is it for the tests?

jmprieur (approved):
LGTM
Thanks @neha-bhargava

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/OtelInstrumentation.cs
neha-bhargava: Updated. Thanks!

localden (approved):

localden:
Thank you @neha-bhargava!

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+    /// </summary>
+    internal class OtelInstrumentation : IOtelInstrumentation
+    {
+        /// <summary>
+        /// Constant to hold the name of the Meter.
+        /// </summary>
+        public const string MeterName = "MicrosoftIdentityClient_Common_Meter";
+
+        /// <summary>
+        /// Constant to holt the name of the ActivitySource.
```
gladjohn, line 26:
```suggestion
        /// Constant to hold the name of the ActivitySource.
```

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// </summary>
+        public const string ActivitySourceName = "MicrosoftIdentityClient_Activity";
+
+        private const string SuccessCounterName = "MsalSuccess";
+        private const string FailedCounterName = "MsalFailed";
+        private const string TotalDurationHistogramName = "MsalTotalDuration.1A";
```
gladjohn, line 32: Could these be enums instead?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -170,6 +179,9 @@
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == '$(TargetFrameworkNetDesktop462)' ">
+    <Compile Include="$(PathToMsalSources)\Platforms\Features\OpenTelemetry\**\*.cs" />
+    
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == '$(TargetFrameworkNetDesktop45)' ">
```
gladjohn, line 187: We bumped net461 to net462 to support OTEL, are there any special considerations for net45?

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// </summary>
+        public const string MeterName = "MicrosoftIdentityClient_Common_Meter";
+
+        /// <summary>
+        /// Constant to holt the name of the ActivitySource.
+        /// </summary>
+        public const string ActivitySourceName = "MicrosoftIdentityClient_Activity";
+
+        private const string SuccessCounterName = "MsalSuccess";
+        private const string FailedCounterName = "MsalFailed";
```
gladjohn, line 31: nit: Failure would have been more appropriate since we call Success

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+            description: "Performance of token acquisition calls total latency");
+
+        /// <summary>
+        /// Histogram to record total duration of token acquisition calls in microseconds(us) when total duration in ms is less than 5.
+        /// This will capture the duration in us when the L1 cache is used to get the token.
+        /// </summary>
+        internal static readonly Histogram<long> s_durationTotalInUs = Meter.CreateHistogram<long>(
+            TotalDurationInUsHistogramName,
+            unit: "us",
+            description: "Performance of token acquisition calls total latency in micro seconds when L1 cache is used.");
```
gladjohn, line 76:
```suggestion
            description: "Performance of token acquisition calls total latency in microseconds when L1 cache is used.");
```

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <summary>
+        /// Histogram to record duration in cache of token acquisition calls.
+        /// </summary>
+        internal static readonly Histogram<long> s_durationInCache = Meter.CreateHistogram<long>(
+            DurationInCacheHistogramName,
+            unit: "ms",
+            description: "Performance of token acquisition calls cache latency");
+
+        /// <summary>
+        /// Histogram to record duration in http of token acquisition calls.
```
gladjohn, line 87:
```suggestion
        /// Histogram capturing the duration, in milliseconds, of token acquisition calls via HTTP.
```

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// </summary>
+        internal static readonly Histogram<long> s_durationInHttp = Meter.CreateHistogram<long>(
+            DurationInHttpHistogramName,
+            unit: "ms",
+            description: "Performance of token acquisition calls network latency");
+
+        internal static readonly Lazy<Activity> s_activity = new Lazy<Activity>(() => s_acquireTokenActivity.StartActivity("Token Acquisition", ActivityKind.Internal));
```
gladjohn, line 94:
```suggestion
        internal static readonly Lazy<Activity> s_activity = new Lazy<Activity>(() => s_acquireTokenActivity.StartActivity("Token acquisition", ActivityKind.Internal));
```

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            foreach (KeyValuePair<string, object> tag in tags)
+            {
+                s_activity.Value?.AddTag(tag.Key, tag.Value);
+            }
+        }
+
+        void IOtelInstrumentation.LogActivityStatus(bool success)
+        {
+            if (success)
```
gladjohn, line 106: will there be a perf improvement if we checked for null? `activity != null`

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/IOtelInstrumentation.cs
```diff
@@ -0,0 +1,25 @@
...
+        internal void LogActivity(Dictionary<string, object> tags);
+        internal void LogActivityStatus(bool success);
+        internal void StopActivity();
+        internal void LogSuccessMetrics(
+            string platform,
+            AuthenticationResultMetadata authResultMetadata,
+            string apiId,
+            string cacheLevel,
+            ILoggerAdapter logger, 
+            long totalDurationInUs);
```
gladjohn, line 21: nit: tricky one, we can simply call this `totalDurationInMs`

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
```diff
@@ -0,0 +1,145 @@
...
+            // Acquire token for client with scope
+            var result = await _cca.AcquireTokenForClient(TestConstants.s_scope)
+                .ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
+            Assert.IsNotNull(result);
+
+            // Acquire token silently
+            var account = (await _cca.GetAccountsAsync().ConfigureAwait(false)).Single();
+            result = await _cca.AcquireTokenSilent(TestConstants.s_scope, account)
+                .ExecuteAsync().ConfigureAwait(false);
+            Assert.IsNotNull(result);
```
gladjohn, line 95: did we miss checking any successful metric here?

tests/devapps/TestTelemetry/Program.cs
gladjohn: @neha-bhargava still not fixed?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+                { TelemetryConstants.ErrorMessage, exception.Message }
+            };
+
+            if (exception is MsalClientException clientException)
+            {
+                tags.Add(TelemetryConstants.ErrorCode, clientException.ErrorCode);
+                return;
+            }
+
+            if (exception is MsalServiceException serviceException)
```
gladjohn, line 214: should we consider adding managedIdentityException?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -165,21 +267,36 @@ private void LogSuccessfulTelemetryToClient(AuthenticationResult authenticationR
...
         }
 
-        private void ParseScopesForTelemetry(MsalTelemetryEventDetails telemetryEventDetails)
+        private Tuple<string, string> ParseScopesForTelemetry()
         {
+            string resource = null;
+            string scopes = null;
             if (AuthenticationRequestParameters.Scope.Count > 0)
             {
                 string firstScope = AuthenticationRequestParameters.Scope.First();
```
gladjohn, line 294: can scopes ever be null? like in a failure case? do we then add metrics for it

gladjohn (approved):
looks good Neha, just some minor comments

tests/devapps/TestTelemetry/Program.cs
neha-bhargava: Missed. Thanks for pointing out

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
neha-bhargava: No, these are strings for names of counters and histograms. Does not make sense as enum.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
neha-bhargava: Net 45 is excluded from logging any telemetry through otel. We did drop net core 2.1 and bump to net 462 but net 45 is still supported as there may be dependencies on it.

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/IOtelInstrumentation.cs
neha-bhargava: It is microseconds(us). I will add a comment here.

neha-bhargava: The totalDurationInMs is present in authResultMetadata. We needed microseconds to record duration in case of L1 cache

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
neha-bhargava: Since otel metrics measures aggregation, I have coupled the success and failure tests and verifying the metrics together. So this is a method to only acquire token and assert that we get a successful response. Metrics is verified at line 76.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: ManagedIdentityException is inherited from MsalServiceException so will fall in this category.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: When the scopes are null, there is a MSAL failure with client exception. So this will not be logged. This is logged for success scenarios.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -125,16 +125,22 @@
     <Compile Include="$(PathToMsalSources)\Platforms\netstandard\**\*.cs" LinkBase="Platforms\netstandard" />
     <Compile Include="$(PathToMsalSources)\Platforms\Features\DesktopOS\**\*.cs" />
     <Compile Include="$(PathToMsalSources)\Platforms\Features\DefaultOSBrowser\**\*.cs" />
+    <Compile Include="$(PathToMsalSources)\Platforms\Features\OpenTelemetry\**\*.cs" />
+
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />
```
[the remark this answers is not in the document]
neha-bhargava: [Here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/nebharg/openTelemetry/src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs) for the metrics and activity

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -80,6 +79,13 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+            ServiceBundle.PlatformProxy.OtelInstrumentation?.LogActivity(new Dictionary<string, object> { 
+                { TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion() },
+                { TelemetryConstants.Platform, ServiceBundle.PlatformProxy.GetProductName() },
+                { TelemetryConstants.ClientId, AuthenticationRequestParameters.AppConfig.ClientId },
+                { TelemetryConstants.ActivityId, AuthenticationRequestParameters.RequestContext.CorrelationId }
+            });
+
             ApiEvent apiEvent = InitializeApiEvent(AuthenticationRequestParameters.Account?.HomeAccountId?.Identifier);
             AuthenticationRequestParameters.RequestContext.ApiEvent = apiEvent;
             MsalTelemetryEventDetails telemetryEventDetails = new MsalTelemetryEventDetails(TelemetryConstants.AcquireTokenEventName);
```
pmaytak, line 91: Will the old ITelemetryClient logic be obsoleted or removed?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
                 }
                 finally
                 {
+                    ServiceBundle.PlatformProxy.OtelInstrumentation?.StopActivity();
                     telemetryClients.TrackEvent(telemetryEventDetails);
                 }
             }
         }
 
+        private void LogMsalSuccessTelemetryToOTel(AuthenticationResult authenticationResult)
```
pmaytak, line 150: `LogMsalSuccessTelemetryToOtel`

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+            }
+            if (resourceAndScopes.Item2 != null)
+            {
+                tags.Add(TelemetryConstants.Scopes, resourceAndScopes.Item2);
+            }
+
+            ServiceBundle.PlatformProxy.OtelInstrumentation.LogActivity(tags);
+        }
+
+        private void LogMsalFailedTelemetryToOTel(Exception exception)
```
pmaytak, line 191: `LogMsalFailedTelemetryToOtel`

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -80,6 +79,13 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
         {
             Stopwatch sw = Stopwatch.StartNew();
 
+            ServiceBundle.PlatformProxy.OtelInstrumentation?.LogActivity(new Dictionary<string, object> {
```
pmaytak, line 82: `OtelInstrumentation` is never null. It'll always get instantiated in AbstractPlatformProxy after it's called here for the first time.

In the scenario, when a platform doesn't support Otel, NullOtelInstrumentation is created with no-op implementation, however, these parameters and dictionaries are still allocated and methods are called unnecessarily. I would maybe just make OtelInstrumentation null in this case and do a null check? @bgavrilMS does this sound right, or is the compiler smart enough to not create these param objects if it sees that they are not used down the line?

Second scenario is similar but for platforms that support Otel but it's just not enabled. Is there a way to instantiate OtelInstrumentation only if it's enabled or have some boolean flag like we have for the ITelemetryClient?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
                     LogMetricsFromAuthResult(authenticationResult, AuthenticationRequestParameters.RequestContext.Logger);
                     LogSuccessfulTelemetryToClient(authenticationResult, telemetryEventDetails, telemetryClients);
 
+                    ServiceBundle.PlatformProxy.OtelInstrumentation?.LogSuccessMetrics(
```
pmaytak, line 107: Can this be moved into LogMsalSuccessTelemetryToOTel? This RunAsync method is getting a bit long, just thinking that extracting code into separate methods will make it more readable and maintainable.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+
                     return authenticationResult;
                 }
                 catch (MsalException ex)
                 {
                     apiEvent.ApiErrorCode = ex.ErrorCode;
                     AuthenticationRequestParameters.RequestContext.Logger.ErrorPii(ex);
                     LogMsalErrorTelemetryToClient(ex, telemetryEventDetails, telemetryClients);
+
+                        ServiceBundle.PlatformProxy.OtelInstrumentation.LogFailedMetrics(
```
pmaytak, line 125: Same comment as above - moving this into the below method.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
                 {
+                    ServiceBundle.PlatformProxy.OtelInstrumentation?.StopActivity();
                     telemetryClients.TrackEvent(telemetryEventDetails);
                 }
             }
         }
 
+        private void LogMsalSuccessTelemetryToOTel(AuthenticationResult authenticationResult)
+        {
+            if (ServiceBundle.PlatformProxy.OtelInstrumentation == null)
```
pmaytak, line 152: Same comment as above, it's never null.

src/client/Microsoft.Identity.Client/TelemetryCore/OpenTelemetry/IOtelInstrumentation.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.TelemetryCore.OpenTelemetry
+{
+    internal interface IOtelInstrumentation
+    {
+        internal void LogActivity(Dictionary<string, object> tags);
+        internal void LogActivityStatus(bool success);
```
pmaytak, line 13:
```suggestion
        internal void LogActivityStatus(bool isSuccessful);
```

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+            ServiceBundle.PlatformProxy.OtelInstrumentation.LogActivityStatus(true);
+
+            Dictionary<string, object> tags = new Dictionary<string, object> { 
+                { TelemetryConstants.CacheInfoTelemetry, Convert.ToInt64(authenticationResult.AuthenticationResultMetadata.CacheRefreshReason) },
+                { TelemetryConstants.TokenSource, Convert.ToInt64(authenticationResult.AuthenticationResultMetadata.TokenSource) },
+                { TelemetryConstants.Duration, authenticationResult.AuthenticationResultMetadata.DurationTotalInMs },
+                { TelemetryConstants.DurationInCache, authenticationResult.AuthenticationResultMetadata.DurationInCacheInMs },
+                { TelemetryConstants.DurationInHttp, authenticationResult.AuthenticationResultMetadata.DurationInHttpInMs },
+                { TelemetryConstants.TokenType, (int)AuthenticationRequestParameters.RequestContext.ApiEvent.TokenType },
+                { TelemetryConstants.RemainingLifetime, (authenticationResult.ExpiresOn - DateTime.Now).TotalMilliseconds }
```
pmaytak, line 166: Since this is a new telemetry store, can we improve on the key name constants and use more specific ones?
For example:
`CacheInfoTelemetry` > `CacheRefreshReason`
`Duration` > `RequestDurationInMs`
`DurationInCache` > `DurationInCacheInMs`
`DurationInHttp` > `DurationInHttpInMs`
`RemainingLifetime` > `RemainingTokenLifetimeInMs`

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -98,29 +104,125 @@ public async Task<AuthenticationResult> RunAsync(CancellationToken cancellationT
...
+
+            if (authenticationResult.AuthenticationResultMetadata.RefreshOn.HasValue)
+            {
+                tags.Add(TelemetryConstants.RefreshOn, DateTimeHelpers.DateTimeToUnixTimestampMilliseconds(authenticationResult.AuthenticationResultMetadata.RefreshOn.Value));
+            }
+            tags.Add(TelemetryConstants.AssertionType, (int)AuthenticationRequestParameters.RequestContext.ApiEvent.AssertionType);
+            tags.Add(TelemetryConstants.Endpoint, AuthenticationRequestParameters.Authority.AuthorityInfo.CanonicalAuthority.ToString());
+            tags.Add(TelemetryConstants.CacheLevel, (int)authenticationResult.AuthenticationResultMetadata.CacheLevel);
+
+            var resourceAndScopes = ParseScopesForTelemetry();
```
pmaytak, line 177: I think this can be deconstructed something like `(string resouce, string scopes) = ParseScopesForTelemetry();` for better readability.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -237,6 +356,7 @@ private void LogMetricsFromAuthResult(AuthenticationResult authenticationResult,
 
         private void UpdateTelemetry(Stopwatch sw, ApiEvent apiEvent, AuthenticationResult authenticationResult)
         {
+            sw.Stop();
```
pmaytak, line 359: Call this in the RunAsync method, so that sw Start and Stop are at the same level and easier to see.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// Constant to holt the name of the ActivitySource.
+        /// </summary>
+        public const string ActivitySourceName = "MicrosoftIdentityClient_Activity";
+
+        private const string SuccessCounterName = "MsalSuccess";
+        private const string FailedCounterName = "MsalFailed";
+        private const string TotalDurationHistogramName = "MsalTotalDuration.1A";
+        private const string TotalDurationInUsHistogramName = "MsalTotalDurationInUs.1B";
+        private const string DurationInCacheHistogramName = "MsalDurationInCache.1A";
+        private const string DurationInHttpHistogramName = "MsalDurationInHttp.1A";
```
pmaytak: Make variable names more specific and also the histogram name itself (idk if it can be changed now.
```suggestion
        private const string RequestDurationInMsHistogramName = "MsalRequestDurationInMs.1A";
        private const string RequestDurationInUsHistogramName = "MsalRequestDurationInUs.1B";
        private const string DurationInCacheInMsHistogramName = "MsalDurationInCacheInMs.1A";
        private const string DurationInHttpInMsHistogramName = "MsalDurationInHttpInMs.1A";
```

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        /// <summary>
+        /// Meter to hold the MSAL metrics.
+        /// </summary>
+        internal static readonly Meter Meter = new Meter(MeterName, "1.0.0");
+
+        /// <summary>
+        /// ActivitySource to hold the MSAL activities.
+        /// </summary>
+        internal static readonly ActivitySource s_acquireTokenActivity = new ActivitySource(ActivitySourceName, "1.0.0");
```
pmaytak, line 45: I assume this is a temporary collection, which gets cleared out once the exporter exports the data? This doesn't grow indefinitely? Can a limit be set on the size of this or does it depend on the export frequency set?
Also to my previous point, if platform supports Otel but it's not enabled, this will just get populated but never cleared, right?

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -0,0 +1,182 @@
...
+            {
+                s_activity.Value?.SetStatus(ActivityStatusCode.Ok, "Success");
+            }
+            else
+            {
+                s_activity.Value?.SetStatus(ActivityStatusCode.Error, "Request failed");
+            }
+        }
+
+        void IOtelInstrumentation.StopActivity()
```
pmaytak, line 116: This `StopActivity` has no corresponding `Start` method, so it's unclear when and if this has to be called. I suggest to rename `LogActivity` to `StartActivity`.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
pmaytak, line 94: What is the intended lifetime of Otel `Activity`? Is it equivalent to a request, ie per unique correlation ID? If so , this shouldn't be static? Otherwise, there would be race conditions, since multiple parallel requests could set their own tags on here.

pmaytak (changes requested):
Questions related to performance and race conditions.

src/client/Microsoft.Identity.Client/TelemetryCore/TelemetryConstants.cs
```diff
@@ -29,6 +29,7 @@ internal static class TelemetryConstants
         public const string StsErrorCode = "StsErrorCode";
         public const string ErrorMessage = "ErrorMessage";
         public const string Duration = "Duration";
+        public const string DurationInUs = "DurationInUs";
```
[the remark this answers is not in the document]
neha-bhargava: Updated the names. This is duration in L1 cache

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: Will be eventually in future.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
pmaytak: Also same question about counters and histogram objects below - when are they cleared.

pmaytak: Based on sync - the exporter clears the collection after exporting data.

tests/Microsoft.Identity.Test.Common/Core/Helpers/OTelInstrumentationUtil.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        public static void VerifyActivity(int expectedTagCount, List<Activity> _exportedActivities)
+        {
+            Assert.AreEqual(1, _exportedActivities.Count);
```
pmaytak, line 21: Shouldn't there be 5 activities in the collection - 2 successful and 3 failed ones because there were 5 requests?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
neha-bhargava: Discussed offline

tests/Microsoft.Identity.Test.Common/Core/Helpers/OTelInstrumentationUtil.cs
neha-bhargava: Right, the activity should start at the beginning of each request. Updated.

pmaytak (approved):

PR #4248, opened 2023-07-14
Enable CodeQL with Autobuild for javascript' and 'python' and custom for csharp

gladjohn:
run codeql in github. to check if S360 is affected in anyway

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -28,8 +28,8 @@ jobs:
     #   - https://gh.io/supported-runners-and-hardware-resources
     #   - https://gh.io/using-larger-runners
     # Consider using larger runners for possible analysis time improvements.
-    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
-    timeout-minutes: ${{ (matrix.language == 'swift' && 120) || 360 }}
+    runs-on: 'ubuntu-latest'
+    timeout-minutes: 360
     permissions:
       actions: read
       contents: read
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'csharp', 'java', 'javascript', 'python' ]
+        language: [ 'javascript', 'python' ]
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby', 'swift' ]
         # Use only 'java' to analyze code written in Java, Kotlin or both
         # Use only 'javascript' to analyze code written in JavaScript, TypeScript or both
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,82 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+  schedule:
+    - cron: '40 12 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,82 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners
+    # Consider using larger runners for possible analysis time improvements.
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
```
bgavrilMS, line 31: delete the matrix language?

.github/workflows/codeql_csharp.yml
```diff
@@ -0,0 +1,91 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners
+    # Consider using larger runners for possible analysis time improvements.
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
```
bgavrilMS, line 31: same as above, remove all references to "switf"

bgavrilMS (changes requested):
Do not run on JS and Py

.github/workflows/codeql_csharp.yml
```diff
@@ -0,0 +1,91 @@
...
+    #     ./location_of_script_within_repo/buildscript.sh
+
+    - name: Set up .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: '6.0.x'
+        include-prerelease: true
+  
+    - name: Build with dotnet
+      run: dotnet build 'src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj' --configuration Release
```
pmaytak, line 86: Should run on the Desktop and Broker too.

.github/workflows/codeql.yml
pmaytak, line 1: You can also look at how Identity Web is set up for reference:

https://github.com/AzureAD/microsoft-identity-web/blob/master/.github/workflows/codeql-analysis.yml

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,82 @@
...
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
```
pmaytak, line 17: Ideally we should only run C# code for main projects, otherwise, it'll add more time to running all these PR checks.

.github/workflows/codeql.yml
gladjohn: do you want us to replicate this from idweb? 

```
on:
  push:
    branches:
      - master
  pull_request:
    types:
      - opened
      - synchronize
      - reopened
      - ready_for_review
    branches:
      - master
```

.github/workflows/codeql.yml
gladjohn: think we can do that since those are test projects. Or should we run this only during our nightly? and PRs

.github/workflows/codeql.yml
pmaytak: What's the default if you don't specify a type? I think it's ok to run after each update to the PR. If I remember correctly, the scans are incremental, only on the changed code. Not sure though. Up to you though.

.github/workflows/codeql.yml
pmaytak: I think for test projects we should only run secret scanning and security vulnerability scans, which we already do. CodeQL is not necessary for tests/devapps; I think it'll add more noise - like telling to put comments on public members or use ConfigureAwait and such.

trwalke: >we should only run secret scanning and security vulnerability scans

At the moment, we would like to maintain the same level of analysis for our test projects that we have on our main source project. I think it makes sense so not run it for JS and PY code, but C# should be covered. We are running the same level of Roslyn analysis on our test projects today.

pmaytak: Ok. Yea, we can adjust later if needed.

gladjohn:
removing this as we have OneBranch working now

PR #4250, opened 2023-07-14
New Feature request template

gladjohn:
New Feature request template

PR #4251, opened 2023-07-14
Create new engineering task template

gladjohn:
Create engineering task template

check the file tab preview

image


bgavrilMS (dismissed):

trwalke:
Not sure the engineering tasks should fall into these categories
image

It should be something like (build, compliance, component gov, etc)

jennyf19:
I would remove the Microsoft internal part, as this is an open source library.

.github/ISSUE_TEMPLATE/engineering_task.yaml
pmaytak, line 1: I'm not sure about this template. I would just use a blank template for any miscellaneous issues. I think these fields don't apply to some tasks like "Create missing tests for blah", "Update expired cert.", "Fix the build.".

trwalke: Wouldn't the "Create missing tests for blah" fall under the test category and the other two will fall under build? So long as the categories are broad enough it should be ok.

trwalke:
Seems like the changes are not reflected in the preview

.github/ISSUE_TEMPLATE/engineering_task.yaml
gladjohn: very minimal info capturing now

gladjohn:
> Seems like the changes are not reflected in the preview

check the file tab preview

rayluo:
Questions:

1. This file is currently different than [the demo](https://github.com/gladjohn/templates/issues/new/choose). Is that expected?
2. What is the significance of having this dedicate "new engineering task" template, while we will have a Feature Request template? Functionality wise, can't we combine them into the #4250?
3. Having two templates may also send a signal to the 3rd party that we would treat 1P and 3P differently. Do we want that?

gladjohn:
> Questions:
> 
> 1. This file is currently different than [the demo](https://github.com/gladjohn/templates/issues/new/choose). Is that expected?
> 2. What is the significance of having this dedicate "new engineering task" template, while we will have a Feature Request template? Functionality wise, can't we combine them into the [New Feature request template #4250](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4250)?
> 3. Having two templates may also send a signal to the 3rd party that we would treat 1P and 3P differently. Do we want that?

This is for MSAL .NET team to file tasks, like engineering improvement or perf improvements, build issues etc

localden:
+1 to @jennyf19's comment. The "Internal Use Only" part should not be really used in an open source project. That is, it's fine to have the template, but I am not sure if we need to state "for internal use only" - it could be rephrased to something like "For MSAL team use only" instead if we want to be explicit about the fact that this is just for the team.

neha-bhargava (approved):

.github/ISSUE_TEMPLATE/engineering_task.yaml
pmaytak: I think it still needs to be more general. A lot of these internal tasks are very simple - fix/update/add a test, fix/update build, change dev apps, add/update governance, design spike / prototype an issue.

- I would remove the category selection. This is implied in the description.
- Remove Solution section, this is also implied in the description.
- Remove mentions of "internal only", non-MSAL folks can create these too.

PR #4252, opened 2023-07-17
Updating OneBranch to add a CI build

trwalke:
Adding CI build so that we dont pack and sign packages for each OneBranch build.

CI Build: https://dev.azure.com/IdentityDivision/IDDP/_build/results?buildId=1140507&view=results
Fixes #

**Testing**

**Performance impact**

**Documentation**


build/template-OneBranch-Release-libsandsamples.yaml
```diff
@@ -0,0 +1,19 @@
...
+  MsalClientSemVer: '4.0.0-devopsbuild'
+  MsalSourceDir: 'microsoft-authentication-library-for-dotnet\' #Default MSAL Location
+
+steps:
+
+- template: template-OneBranch-CI-libsandsamples.yaml
+  parameters:
+    BuildConfiguration: ${{ parameters.BuildConfiguration }}
+    ProjectRootPath: '$(Build.SourcesDirectory)\$(MsalSourceDir)src\client'
+    AssemblyName: 'Microsoft.Identity.Client.Desktop'
```
pmaytak, line 17: What's this for?

gladjohn (commented):
file is referenced [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/6e5953131ebe0b2715761caf8d92bcd6832c4bb4/LibsAndSamples.sln#L126)

gladjohn (commented):
Can we have one skeleton template for all pipelines and derive what gets executed based on which yaml is calling into it?

build/template-OneBranch-Release-libsandsamples.yaml
trwalke: This was a copy and paste from somewhere else in the code. I think I copied the wrong thing and this param should not be there. will remove it.

trwalke:
> Can we have one skeleton template for all pipelines and derive what gets executed based on which yaml is calling into it?

I looked into this and there is nt an easy way to do this because the classic pipelines are so different from the onebranch pipelines since they also run tests. In the future, if we manage to get tests running in OneBranch, we can make a more generic skeleton.

gladjohn:
@trwalke can you please add the test pipeline link?

pmaytak:
> Can we have one skeleton template for all pipelines and derive what gets executed based on which yaml is calling into it?

Or just pass different variables for CI/PR vs Release.

> > Can we have one skeleton template for all pipelines and derive what gets executed based on which yaml is calling into it?
> 
> I looked into this and there is nt an easy way to do this because the classic pipelines are so different from the onebranch pipelines since they also run tests. In the future, if we manage to get tests running in OneBranch, we can make a more generic skeleton.

We should add value incrementally. Maybe can refactor pipelines into one later.

trwalke:
> > Can we have one skeleton template for all pipelines and derive what gets executed based on which yaml is calling into it?
> 
> Or just pass different variables for CI/PR vs Release.
> 
> > > Can we have one skeleton template for all pipelines and derive what gets executed based on which yaml is calling into it?
> > 
> > 
> > I looked into this and there is nt an easy way to do this because the classic pipelines are so different from the onebranch pipelines since they also run tests. In the future, if we manage to get tests running in OneBranch, we can make a more generic skeleton.
> 
> We should add value incrementally. Maybe can refactor pipelines into one later.

Yes, once everything is working well, we can refactor

gladjohn (approved):
Looks good @trwalke. Do we plan to include tests as well?

bgavrilMS:
@trwalke - I though all OneBranch work is now finished? What is this PR?

trwalke:
> @trwalke - I though all OneBranch work is now finished? What is this PR?

This is for the CI build for onebranch. Just needs to be merged in. part of overall migration. Next step is to try to integrate tests into CI build

build/template-OneBranch-Release-libsandsamples.yaml
```diff
@@ -0,0 +1,20 @@
+# template-OneBranch-Release-libsandsamples.yaml
+# Performs Nuget Restore and Build of LibsAndSamples.sln based on BuildPlatform and BuildConfiguration
+# This also includes a pack and sign of the MSAL.NET packages
+
+parameters:
+  BuildPlatform: 'any cpu'
+  BuildConfiguration: 'release'
+  MsalClientSemVer: '4.0.0-devopsbuild'
```
bgavrilMS, line 8: Pls use smth like `4.55.0-devopsbuild`, because there is ESTS logic that is blocked on MSAL version and `4.0.0` is too old, i.e. tests will actually behave differently.

bgavrilMS (approved):

PR #4262, opened 2023-07-19
Add All reviewers for PRs

gladjohn:
Add All reviewers for PRs


.github/auto_assign.yml
```diff
@@ -14,4 +14,4 @@ reviewers:
 
 # A number of reviewers added to the pull request
 # Set 0 to add all the reviewers (default: 0)
-# numberOfReviewers: 2
+numberOfReviewers: 0
```
pmaytak, line 17: If previously it was 2, but it only added me, is it off-by-one error? lol

pmaytak (approved):

.github/auto_assign.yml
gladjohn: per docs default is 0. I just left the commented property so we know such a thing exist, now uncommenting and forcing 0

trwalke (approved):

PR #4271, opened 2023-07-24
Fix some static analyzer issues in tests around certs and TLS

bgavrilMS:
Fixes some Roslyn analyzer failures.


tests/Microsoft.Identity.Test.Unit/Helpers/CertHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Microsoft.Identity.Test.Common.Core.Helpers
+{
+    internal class CertHelper
+    {
+        public static X509Certificate2 CreateTestCert()
```
pmaytak, line 14: Can this be cached, so it's only created once per test run? Otherwise, each test creates a new cert each time.

gladjohn (approved):
looks good, we may need to remove the .pfx file from the [build/credscan-exclusion.json](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/2439c36581cc25e0ba4f7e942fbeeeef10d46318/build/credscan-exclusion.json#L5) file. There are few more pfx exclusions there though.

PR #4273, opened 2023-07-25
Trigger OneBranch CI when github main gets updated

gladjohn:
Trigger OneBranch CI when github main gets updated. this uses the `AZURE_DEVOPS_TOKEN` variable. `AZURE_DEVOPS_TOKEN` holds the ADO PAT

Reference : https://learn.microsoft.com/en-us/azure/devops/pipelines/ecosystems/github-actions?view=azure-devops

### DIFF
--- a/.github/workflows/trigger_onebranch_ci.yml
+++ b/.github/workflows/trigger_onebranch_ci.yml
@@ -1,0 +1,20 @@
+name: CI
+
+# Run this workflow every time a commit gets pushed to main
+# This triggers the ADO OneBranch CI Pipeline
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+    build:
+        name: Call OneBranch ADO Pipeline (CI)
+        runs-on: ubuntu-latest
+        steps:
+        - name: Azure Pipelines Action
+          uses: Azure/pipelines@v1
+          with:
+            azure-devops-project-url: https://identitydivision.visualstudio.com/IDDP
+            azure-pipeline-name: 'MSAL.NET-OneBranch-Release-Official'
+            azure-devops-token: ${{ secrets.AZURE_DEVOPS_TOKEN }}


trwalke:
I believe this is already happening. it is configured in ADO:
image


What we want ultimately is the ability to trigger a OneBranch build from a PR

gladjohn:
> I believe this is already happening. [it is configured in ADO:](https://identitydivision.visualstudio.com/IDDP/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=1545&view=Tab_Triggers) image
> 
> What we want ultimately is the ability to trigger a OneBranch build from a PR

This runs on a schedule or monitors the old pipeline for build completion 

image

Yes, the PR builds are a little complex where we need to pass few variables to the pipeline, working on it

.github/workflows/trigger_onebranch_ci.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+    build:
+        name: Call OneBranch ADO Pipeline (CI)
+        runs-on: ubuntu-latest
+        steps:
+        - name: Azure Pipelines Action
+          uses: Azure/pipelines@v1
+          with:
+            azure-devops-project-url: https://identitydivision.visualstudio.com/IDDP
+            azure-pipeline-name: 'MSAL.NET-OneBranch-Release-Official'
```
pmaytak, line 19: Do we want to have two pipelines - Release and CI/PR? The CI/PR should be faster, won't sign packages and such.

pmaytak (approved):

.github/workflows/trigger_onebranch_ci.yml
gladjohn: Ned to work with Travis on the naming, this was called CI, but now changed to Release.

Issue #4237, opened 2023-07-12
[Fundamentals] Implement performance tests and testing infrastructure


pmaytak:
Used to track July work. Further remaining work will be tracked in specific issues and repos.

Each MSAL (.NET, Python, Node, Java)
- [x] Find a benchmarking tool to use.
- [x] Implement 1 client credential benchmark with the tool.
- [x] Trigger to run the benchmark project as CI build.
- [x] Update the project to detect regressions.
- [x] Break the CI build if perf regression detected.
- [x] Upload benchmark results into a data store.
- [x] Verify dashboard shows test run results.

Common
- [x] Choose metrics.
- [x] Choose benchmark test cases.
- [x] Choose real-world test scenarios.
- [ ] Investigate and choose a common data store and data schema. (Further tracker in #4295.)
- [ ] Investigate and choose common reporting dashboard method. (Further tracker in #4295.)

PRs:
- Node: https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/6276
- Python: https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/580
- .NET: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4285

Dashboards:
- .NET: https://azuread.github.io/microsoft-authentication-library-for-dotnet/benchmarks/
- Node: https://azuread.github.io/microsoft-authentication-library-for-js/dev/bench/
- Python: https://azuread.github.io/microsoft-authentication-library-for-python/dev/bench/

PR #4285, opened 2023-08-03
Add Action to run performance benchmarks

pmaytak:
Completes partially #4237

Removes test case (cache size 1/1000) because it's redundant, as discussed.
Adds yml to run the benchmarks in the perf test project using a GitHub Action.
Publishes simple results to GitHub pages: https://azuread.github.io/microsoft-authentication-library-for-dotnet/benchmarks/
Breaks if detects 30% regression from last perf run.
Currently only runs for commits to the main branch.

**Documentation**


.github/workflows/benchmark-action.yml
```diff
@@ -0,0 +1,112 @@
...
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: AcquireTokenNoCache
+          tool: 'benchmarkdotnet'
+          output-file-path: tests/Microsoft.Identity.Test.Performance/BenchmarkDotNet.Artifacts/results/Microsoft.Identity.Test.Performance.AcquireTokenNoCacheTests-report-full-compressed.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          #external-data-json-path: ./cache/benchmark-data.json
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '140%'
```
bgavrilMS, line 48: In the PR description you mention 30% regression, but here I see 140%

.github/workflows/benchmark-action.yml
```diff
@@ -0,0 +1,112 @@
...
+          comment-on-alert: true
+          fail-on-alert: true
+          alert-comment-cc-users: '@pmaytak'
+          summary-always: true
+          gh-pages-branch: gh-pages
+          benchmark-data-dir-path: benchmarks
```
bgavrilMS, line 72: In the graph page, can you change the name of each test to something more meaningful (currently it's the test FQN)?

It would also be interesting to combine some of those graphs toghether, for example "GetAccounts" + "ATS" + "AcquireTokenByCode" or "EnableCacheSync = True / False"

bgavrilMS (approved):

bgavrilMS:
I can't see any trace of the commit "211047c" in the history of the main branch, but I found it to be one of the 31 commits in here. 

However, am I right to understand that we will only track commits that make it into main (i.e. CI runs)?

.github/workflows/benchmark-action.yml
```diff
@@ -0,0 +1,112 @@
+name: Run performance benchmarks
+on:
+  workflow_dispatch: # Allows manual triggering of workflow
+  push:
+    branches:
+      - main
+    paths:
+      - src/client/Microsoft.Identity.Client/**/*.cs
+  #pull_request: # Will re-enable once the code to not push to gh-pages is fixed
```
bgavrilMS, line 9: I would think 2 more scenarios are relevant, as a follow-up

1. Creation of CCA object 
2. Using cache serialization

.github/workflows/benchmark-action.yml
pmaytak: It has to be 100%+ to count as a regression.
https://github.com/benchmark-action/github-action-benchmark#alert-threshold-optional

.github/workflows/benchmark-action.yml
pmaytak: I don't think so. We specify a user-friendly description, but seems like it's not picked up.

.github/workflows/benchmark-action.yml
pmaytak: 1. We already have a test for that, just not running it for automation. Just need to add.
2. We already have that.

pmaytak:
> I can't see any trace of the commit "211047c" in the history of the main branch, but I found it to be one of the 31 commits in here.
> 
> However, am I right to understand that we will only track commits that make it into main (i.e. CI runs)?

The commits in the chart are from my feature branch. This yaml is currently set to run on merge to main. If we set it to run on each commit in the PR, I don't know if that's necessary. It will add noise to the graphs too; although there's an option to exclude these type of runs from updating the graphs, but I wasn't able to make it work.

.github/workflows/benchmark-action.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Store AcquireTokenNoCacheTests result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: AcquireTokenNoCache
+          tool: 'benchmarkdotnet'
+          output-file-path: tests/Microsoft.Identity.Test.Performance/BenchmarkDotNet.Artifacts/results/Microsoft.Identity.Test.Performance.AcquireTokenNoCacheTests-report-full-compressed.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '130%'
```
gladjohn, line 43: have we tested the alerts?

.github/workflows/benchmark-action.yml
```diff
@@ -0,0 +1,96 @@
...
+          tool: 'benchmarkdotnet'
+          output-file-path: tests/Microsoft.Identity.Test.Performance/BenchmarkDotNet.Artifacts/results/Microsoft.Identity.Test.Performance.AcquireTokenForOboCacheTests-report-full-compressed.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '130%'
+          comment-on-alert: true
+          fail-on-alert: true
+          summary-always: true
+          gh-pages-branch: gh-pages
```
gladjohn, line 79: What will happen if this branch got deleted?

.github/workflows/benchmark-action.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: Store AcquireTokenForClientCacheTests result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: AcquireTokenForClientWithCache
+          tool: 'benchmarkdotnet'
+          output-file-path: tests/Microsoft.Identity.Test.Performance/BenchmarkDotNet.Artifacts/results/Microsoft.Identity.Test.Performance.AcquireTokenForClientCacheTests-report-full-compressed.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '130%'
```
gladjohn, line 59: 130 good enough?

gladjohn:
> > I can't see any trace of the commit "211047c" in the history of the main branch, but I found it to be one of the 31 commits in here.
> > However, am I right to understand that we will only track commits that make it into main (i.e. CI runs)?
> 
> The commits in the chart are from my feature branch. This yaml is currently set to run on merge to main. If we set it to run on each commit in the PR, I don't know if that's necessary. It will add noise to the graphs too; although there's an option to exclude these type of runs from updating the graphs, but I wasn't able to make it work.

per [this](https://github.com/benchmark-action/github-action-benchmark#charts-on-github-pages-1),  `Do not run this workflow on pull request since this workflow has permission to modify contents.`

gladjohn (approved):

.github/workflows/benchmark-action.yml
pmaytak: Yes.

.github/workflows/benchmark-action.yml
pmaytak: I believe it fails at the upload stage, since it won't find the branch.

.github/workflows/benchmark-action.yml
pmaytak: Seems like it's just slightly above the variation between test runs. We can adjust later.

pmaytak:
> > > I can't see any trace of the commit "211047c" in the history of the main branch, but I found it to be one of the 31 commits in here.
> > > However, am I right to understand that we will only track commits that make it into main (i.e. CI runs)?
> > 
> > 
> > The commits in the chart are from my feature branch. This yaml is currently set to run on merge to main. If we set it to run on each commit in the PR, I don't know if that's necessary. It will add noise to the graphs too; although there's an option to exclude these type of runs from updating the graphs, but I wasn't able to make it work.
> 
> per [this](https://github.com/benchmark-action/github-action-benchmark#charts-on-github-pages-1), `Do not run this workflow on pull request since this workflow has permission to modify contents.`

https://github.com/benchmark-action/github-action-benchmark#run-only-on-your-branches

They suggest not running on PR since it will push results to dashboard for each commit for anyone who creates a PR. I wasn't able to exclude the dashboard update in that case. The only option is something like duplicating the whole YAML - one for PR, one for CI with different variables. This can be done later if needed.

.github/workflows/benchmark-action.yml
gladjohn: maybe we should have alerts at 110/115, and not fail for alerts and add failure threshold at 130? that way we can fine tune it better?

pmaytak: Good point. I'll do a couple of test runs to see what the variation is and will do that.

PR #4309, opened 2023-08-22
concurrent token refresh fix for managed identity and app token provider (cc)

### LINKED ISSUE #4196: [Bug] Large numbers of concurrent token refresh attempts cause a cache refresh convoy resulting in chronic 429 errors
### Describe the bug

In a scenario in which an application does the following:
- utilizes a `WithAppTokenProvider` with a callback configured to fetch tokens from a managed identity endpoint
- Issues hundreds of `GetToken` requests simultaneously
- Optionally increases `MaxRetries` in the HttpClient pipeline used to fetch tokens

When such an application encounters a 429 response from the MI endpoint, this can result in a storm of requests and retry requests making the 429 problem worse.
In addition, given the current behavior in MSAL for retries and cache access, all retries are guaranteed to result in a cache miss and will continue to fail as long as the MI endpoint does not return a successful token response.

### Expected behavior

Retry attempts after the token cache is successfully refreshed should succeed via a cache hit rather than through a network request to the MI endpoint or authority. Only one request should be made to the endpoint to refresh the cache for any given cache entry and all other concurrent requests should consume that single result.

### Actual behavior

Once the initial request fails with a retriable status code, all subsequent token requests do not attempt to read the cache and always result in an additional network request.

### Reproduction Steps

Issue a large number of simultaneous `GetToken` requests with a `ManagedIdentityCredential` to induce a 429 response from the MI endpoint

### Environment

Customer example was in Service Fabric, but this should reproduce in any managed identity environment in which a 429 response is possible.

gladjohn:
Fixes #4196

Allow only one request to the IDP

**Testing**
unit testing 

**Performance impact**
should be improved 

**Documentation**


MI429/MI429.csproj
```diff
@@ -0,0 +1,59 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
bgavrilMS, line 2: Please create all new projets with the latest .NET not on .NET FWk

MI429/Program.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Abstractions;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.AppConfig;
+
+namespace MI429
```
bgavrilMS, line 11: Could this be integrated into an existing console application? Or generalized. All of these add to the build time.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -118,20 +107,40 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             }
         }
 
-        private async Task<AuthenticationResult> FetchNewAccessTokenAsync(CancellationToken cancellationToken)
+        private async Task<AuthenticationResult> FetchNewAccessTokenAsync(CancellationToken cancellationToken, bool proactivelyRefreshed = false)
```
bgavrilMS, line 110: Consider not providing a default for `proactivelyRefreshed`. Makes the calling code easier to read.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -118,20 +107,40 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
 
-            return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
+            //allow only one call to the provider 
+            await s_semaphoreSlim.WaitAsync().ConfigureAwait(false);
+
+            try
+            {
+                MsalAccessTokenCacheItem cachedAccessTokenItem = await TryGetCachedAccessTokenAsync().ConfigureAwait(false);
+
+                if (cachedAccessTokenItem != null && !_clientParameters.ForceRefresh &&
```
bgavrilMS, line 128: Add a comment here with the 4 conditions for bypassing the cache: 

- force refersh
- claims
- no AT found in cache
- proactive token refresh needed

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -118,20 +107,40 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+                }
+                else
+                {
+                    msalTokenResponse = await SendTokenRequestToProviderAsync(cancellationToken).ConfigureAwait(false);
+                    return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
+                }
+            }
+            finally
+            {
+                s_semaphoreSlim.Release();
```
bgavrilMS, line 142: Are you adding this semaphore to vanilla `client_credentials` as well? I don't think this is right, as it can dramatically change the behavior of existing apps.

tests/Microsoft.Identity.Test.Unit/AppTokenProviderTests.cs
```diff
@@ -158,5 +159,60 @@ private AppTokenProviderResult GetAppTokenProviderResult(string differentScopesF
...
+
+            var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
+                .WithAuthority("https://login.microsoftonline.com/tid")
+                .WithAppTokenProvider((AppTokenProviderParameters parameters) =>
+                {
+                    return Task.FromResult(GetAppTokenProviderResult(differentScopesForAt));
+                })
+                .BuildConcrete();
+
+#pragma warning disable VSTHRD101
```
bgavrilMS, line 179: Why do you need async delegates?

bgavrilMS: See https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD101.md

bgavrilMS, line 179: Pragma warning disable should always be followed by #pragma warning enable.

tests/Microsoft.Identity.Test.Unit/AppTokenProviderTests.cs
```diff
@@ -158,5 +159,60 @@ private AppTokenProviderResult GetAppTokenProviderResult(string differentScopesF
             return token;
         }
 
+        [TestMethod]
+        public async Task MultiThreadedAppTokenProviderCallHasOnlyOneTokenEndpointRequestAsync()
+        {
+            string differentScopesForAt = string.Empty;
+            int totalThreads = 10; // Number of threads to run concurrently
```
bgavrilMS, line 166: nit `const int ...`

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -99,7 +86,7 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
                         SilentRequestHelper.ProcessFetchInBackground(
                         cachedAccessTokenItem,
-                        () => FetchNewAccessTokenAsync(cancellationToken), logger);
+                        () => FetchNewAccessTokenAsync(cancellationToken, shouldRefresh), logger);
```
bgavrilMS, line 89: try to keep the name of the params the same.

bgavrilMS (approved):
Approved with comments.

tests/Microsoft.Identity.Test.Unit/AppTokenProviderTests.cs
```diff
@@ -158,5 +159,60 @@ private AppTokenProviderResult GetAppTokenProviderResult(string differentScopesF
...
+                }
+                finally
+                {
+                    Interlocked.Decrement(ref threadCount);
+                }
+            });
+
+            while (threadCount != 0)
+            {
+                Thread.Sleep(100);
```
bgavrilMS, line 206: As per https://www.oreilly.com/library/view/concurrency-in-c/9781492054498/ - "The moment you use Thread, you have a legacy application". Modern apps should only use Task. 

Consider rewriting this to make use of `Task.WhenAll` https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenall?view=net-7.0

Also, a failing test should not wait forever, so pls make sure to add a timeout (e.g. via CancellationToken)_

tests/Microsoft.Identity.Test.Unit/AppTokenProviderTests.cs
```diff
@@ -158,5 +159,60 @@ private AppTokenProviderResult GetAppTokenProviderResult(string differentScopesF
...
+            {
+                try
+                {
+                    var authResult = await app.AcquireTokenForClient(TestConstants.s_scope)
+                                                    .ExecuteAsync(new CancellationToken()).ConfigureAwait(false);
+
+                    if (authResult.AuthenticationResultMetadata.TokenSource == TokenSource.IdentityProvider)
+                    {
+                        // Increment identity hits count
+                        Interlocked.Increment(ref identityProviderHits);
```
bgavrilMS, line 190: You should assert in here. These numbers should not go over 1.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: not being added to the `client_credentials`, this gets executed only when `.WithAuthority("https://login.microsoftonline.com/tid")` is used on the cca flow

tests/Microsoft.Identity.Test.Unit/AppTokenProviderTests.cs
gladjohn: removed

gladjohn: removed

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -153,6 +162,35 @@ private async Task<MsalTokenResponse> SendTokenRequestToProviderAsync(Cancellati
             return tokenResponse;
         }
 
+        private async Task<MsalAccessTokenCacheItem> TryGetCachedAccessTokenAsync()
+        {
+            MsalAccessTokenCacheItem cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
+
+            if (cachedAccessTokenItem != null && !_clientParameters.ForceRefresh)
```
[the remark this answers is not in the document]
gladjohn: so the first cache check is done in line 56, and when cache is empty then we sent the call to IDP. In a multi threaded scenario, the calls would have already gone past line 56 and waiting to hit the IDP call. now we only allow one thread to execute the IDP call, fill the cache. The next thread that already passed 56, will now execute the IDP call, before it does we are checking cache one more time, so the cache that was filed by the first thread can be used,.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -700,5 +700,74 @@ public async Task ManagedIdentityIsProactivelyRefreshedAsync()
...
+                var mi = miBuilder.BuildConcrete();
+
+                httpManager.AddManagedIdentityMockHandler(
+                        AppServiceEndpoint,
+                        Resource,
+                        MockHelpers.GetMsiSuccessfulResponse(),
+                        ManagedIdentitySource.AppService);
+
+#pragma warning disable VSTHRD101 
+                ParallelLoopResult result = Parallel.For(0, totalThreads, async (i) =>
```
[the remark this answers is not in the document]
gladjohn: fixed it

MI429/Program.cs
gladjohn: @pmaytak can you please point me to one, i know I have seen one but not able to find it

MI429/MI429.csproj
gladjohn: will replace it per below suggestion

MI429/Program.cs
pmaytak: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/devapps/NetCoreTestApp/Program.cs

gladjohn: added now, this did not show up in the sln

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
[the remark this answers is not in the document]
pmaytak: > does we are checking cache one more time, so the cache that was filed by the first thread can be used,.

Yea, and my point is that if ForceRefresh is true, we don't need to check the cache, just go to IDP

What we have now in this method (if cached token is found):
- Check cache
- If ForceRefresh > go to IDP.
- If !ForceRefresh > return found token.

Proposed:
- If ForceRefresh > go to IDP.
- If !ForceRefresh > check cache, then return found token or go to IDP.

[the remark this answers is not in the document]
pmaytak: This could also be changed to were the parent method checks ForceRefresh flag and calls GetFromCache if it's false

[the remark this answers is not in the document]
gladjohn: > This could also be changed to were the parent method checks ForceRefresh flag and calls GetCachedAT if it's false. So GetCachedAT doesn't need extra ForceRefresh check.

Yes, now I get it because I did not change that logic. Let me modify it.

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -38,87 +38,117 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+
+            ILoggerAdapter logger = AuthenticationRequestParameters.RequestContext.Logger;
             CacheRefreshReason cacheInfoTelemetry = CacheRefreshReason.NotApplicable;
 
-            if (!_managedIdentityParameters.ForceRefresh)
+            //skip checking cache for force refresh
+            if (_managedIdentityParameters.ForceRefresh)
             {
-                cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
+                cacheInfoTelemetry = CacheRefreshReason.ForceRefreshOrClaims;
```
pmaytak, line 48: `ApiEvent.CacheInfo` needs to be set here instead.

nit refactor: remove `cacheInfoTelemetry` altogether, just set `ApiEvent.CacheInfo` everywhere instead. (`ApiEvent.CacheInfo` is NotApplicable by default.)

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -36,105 +37,144 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
 
-            if (!_clientParameters.ForceRefresh &&
-                string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
+            AuthenticationResult authResult = null;
+
+            //skip checking cache for force refresh or when claims are present
+            if (_clientParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
             {
-                cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
+                cacheInfoTelemetry = CacheRefreshReason.ForceRefreshOrClaims;
```
pmaytak, line 55: Same comment here as in the other file.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -36,105 +37,144 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+                !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
             {
-                AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = cacheInfoTelemetry;
+                authResult = await SendTokenRequestToProviderAsync(logger, cancellationToken).ConfigureAwait(false);
+                return authResult;
             }
 
-            // No AT in the cache or AT needs to be refreshed
+            //allow only one call to the provider 
+            logger.Verbose(() => "[GetAccessTokenAsync] Entering token acquire for client credential request semaphore.");
```
pmaytak, line 138: nit: I would replace `[GetAccessTokenAsync]` log prefix with `[ClientCredentialRequest]` because it's a bit more specific when looking through logs.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -36,105 +37,144 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
 
-            return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
+            AuthenticationResult authResult = await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse)
+                .ConfigureAwait(false);
+
+            return authResult;
         }
 
-        private async Task<MsalTokenResponse> SendTokenRequestToProviderAsync(CancellationToken cancellationToken)
+        private async Task<MsalTokenResponse> SendRequestToProviderAsync(CancellationToken cancellationToken)
```
pmaytak, line 177: nit: I would merge this method into `SendTokenRequestToProviderAsync`. The naming on the methods is too similar.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -36,105 +37,144 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            finally
             {
-                return await HandleTokenRefreshErrorAsync(e, cachedAccessTokenItem).ConfigureAwait(false);
+                s_semaphoreSlim.Release();
+                logger.Verbose(() => "[GetAccessTokenAsync] Released token acquire for client credential request semaphore.");
             }
         }
 
-        private async Task<AuthenticationResult> FetchNewAccessTokenAsync(CancellationToken cancellationToken)
+        private async Task<AuthenticationResult> SendTokenRequestToProviderAsync(ILoggerAdapter logger,
```
pmaytak, line 165:
```suggestion
        private async Task<AuthenticationResult> SendTokenRequestToAppTokenProviderAsync(ILoggerAdapter logger, 
```

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -36,105 +37,144 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+
+            //calls to AAD
+            if (ServiceBundle.Config.AppTokenProvider == null)
+            {
+                msalTokenResponse = await SendTokenRequestAsync(GetBodyParameters(), cancellationToken).ConfigureAwait(false);
+                return await CacheTokenResponseAndCreateAuthenticationResultAsync(msalTokenResponse).ConfigureAwait(false);
             }
 
-            if (AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo == CacheRefreshReason.NotApplicable)
+            // Bypass cache and send request to token endpoint, when
```
pmaytak, line 125:
```suggestion
            // Requests to the app token provider delegate must be throttled.
            // Bypass cache and send request to app token provider, when 
```

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -36,105 +37,144 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
             }
 
-            if (AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo == CacheRefreshReason.NotApplicable)
+            // Bypass cache and send request to token endpoint, when 
+            // 1. Force refresh is requested, or
+            // 2. Claims are passed, or 
+            // 3. If the AT needs to be refreshed pro-actively 
+            if (_clientParameters.ForceRefresh ||
+                proactivelyRefresh || 
+                !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
```
pmaytak, line 131: Don't we want to throttle these too since they can cause 429?

pmaytak (changes requested):
The request classes look better 👍 Just some suggestions.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: @bgavrilMS should we?

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -38,87 +38,117 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+
+                    // may fire a request to get a new token in the background when AT needs to be refreshed
+                    if (proactivelyRefresh)
                     {
-                        cacheInfoTelemetry = CacheRefreshReason.NoCachedAccessToken;
+                        cacheInfoTelemetry = CacheRefreshReason.ProactivelyRefreshed;
+
+                        SilentRequestHelper.ProcessFetchInBackground(
+                        cachedAccessTokenItem,
+                        () => SendTokenRequestForManagedIdentityAsync(cancellationToken), logger);
```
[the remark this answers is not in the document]
gladjohn: @bgavrilMS to comment on this

LibsAndSamples.sln
```diff
@@ -2649,6 +2651,48 @@ Global
 		{C5902FA4-CC11-43E8-B7E3-192B5FFA1C40}.Release|x64.Build.0 = Release|Any CPU
 		{C5902FA4-CC11-43E8-B7E3-192B5FFA1C40}.Release|x86.ActiveCfg = Release|Any CPU
 		{C5902FA4-CC11-43E8-B7E3-192B5FFA1C40}.Release|x86.Build.0 = Release|Any CPU
+		{4BEDCC7F-C027-4326-BAE4-AD62C1CD5572}.Debug + MobileApps|Any CPU.ActiveCfg = Debug + MobileApps|Any CPU
```
bgavrilMS, line 2654: No files are added, is there a need to change the sln?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
bgavrilMS: Yes, these will make HTTP request so they should be behind barrier.

I'm concerned about the background thread though. I wonder if the app can kill it or set it to low pri somehow which will leave the semaphore blocked. I wonder if we should set a timeout on that semaphore? 3-5 seconds...

LibsAndSamples.sln
gladjohn: this project was not part of the solution, so just adding the csproj

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: added `await s_semaphoreSlim.WaitAsync(cancellationToken).ConfigureAwait(false);` per our conversation

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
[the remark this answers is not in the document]
gladjohn: now moved inside the semaphore

pmaytak (approved):

PR #4310, opened 2023-08-24
Set CacheSynchronization to true by default

### LINKED ISSUE #4268: [Bug] CacheSynchronizationEnabled flag default is inconsistent
MSAL 4.54.1

## Expected 

CacheSynchronizationEnabled should be `true` by default. It should have negligible impact over when CCA is constructed on a request by request basis. But when CCA is singleton, cache correctness cannot be achieved without this flag set to true (i.e. without locking).

(It should also be hidden / soft deprecated API, as it is too low level)

## Actual

If using `ApplicationOptions` when constructing the CCA object, the `CacheSynchronizationEnabled`  is set to `false` if not explicitly declared.
If not using `ApplicationsOptions`, the `CacheSynchronizationEnabled` is set to `true`

pmaytak:
Fixes #4268

Set CacheSynchronization to true by default

**Testing**
Yes.

**Performance impact**
About the same.

**Documentation**


bgavrilMS (approved):
Consider going into Id.Web and setting this flag explicitly, so that they are isolated by this change.

trwalke (approved):

PR #4316, opened 2023-08-29
Simplify bug template

bgavrilMS:
Simplify bug template

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -43,6 +43,7 @@ body:
       - "ConfidentialClient - web api (AcquireTokenOnBehalfOf)"
       - "ConfidentialClient - service to service (AcquireTokenForClient)"
       - "ManagedIdentityClient - managed identity"
+      - "Other - please specify"
   validations:
     required: true
    
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -33,50 +33,19 @@ body:
 
 - type: dropdown
   attributes:
-    label: MSAL client type
+    label: Scenario
     description: "Are you using PublicClientApplication,  ConfidentialClientApplication or ManagedIdentityApplication?"
     multiple: true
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -33,50 +33,19 @@ body:
...
   attributes:
-    label: MSAL client type
+    label: Scenario
     description: "Are you using PublicClientApplication,  ConfidentialClientApplication or ManagedIdentityApplication?"
     multiple: true
     options: 
-      - "Public"
-      - "Confidential"
-      - "Managed identity"
+      - "PublicClient - desktop app"
```
gladjohn, line 40: the plan was to use these values `Public`, `Confidential` and `Managed identity` and add auto tags

localden (approved):

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -33,50 +33,19 @@ body:
...
+      - "PublicClient - mobile app"
+      - "ConfidentialClient - web site (AcquireTokenByAuthCode)"
```
trwalke, line 42: nit: Isn't it possible for someone to have two of the CCA scenarios? perhaps it should be one generic option since we would ultimately treat them the same?

trwalke:
Is there a way to demo what this would look like?

.github/ISSUE_TEMPLATE/bug_report.yaml
bgavrilMS: Why can't we add tags based on these?

gladjohn: I guess we could

gladjohn:
> Is there a way to demo what this would look like?

view from [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7fe97f6a39b697716f7d4fc2cfedafc3a0de25b3/.github/ISSUE_TEMPLATE/bug_report.yaml) and use the preview

gladjohn (approved):

PR #4318, opened 2023-08-30
Update logging for Runtime Broker

gladjohn:
While investigating an issue for the photo app, the logs were misleading. Updating to `RuntimeBroker`

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
@@ -63,13 +63,13 @@ namespace Microsoft.Identity.Client.Platforms.Features.RuntimeBroker
             long errorCode = authResult.Error.ErrorCode;
             string errorMessage;
 
-            logger.Info("[WamBroker] Processing WAM exception");
-            logger.Verbose(() => $"[WamBroker] TelemetryData: {authResult.TelemetryData}");
+            logger.Info("[RuntimeBroker] Processing WAM exception");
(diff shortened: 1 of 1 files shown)


pmaytak (approved):

trwalke (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -63,13 +63,13 @@ private enum ResponseStatus
             long errorCode = authResult.Error.ErrorCode;
             string errorMessage;
 
-            logger.Info("[WamBroker] Processing WAM exception");
-            logger.Verbose(() => $"[WamBroker] TelemetryData: {authResult.TelemetryData}");
+            logger.Info("[RuntimeBroker] Processing WAM exception");
```
bgavrilMS, line 66: Is it because the "old broker" had [WamBroker] ?

I'm not a fan of this - changing log statements like this will make it harder to find the statements.

gladjohn: Yes the old broker has [WamBroker] and the new one has [[RuntimeBroker]](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs), just the WamAdapter.cs on the new one still has [WamBroker], while looking at logs this leads to a confusion

bgavrilMS (approved):

PR #4328, opened 2023-08-31
Check for null and empty scopes in managed identity

### LINKED ISSUE #4332: [Bug] Managed identity crashes when resource is null
### Library version used

4.54.1

### .NET version

Any.

### Scenario

None

Managed identity code crashes when [default suffix is removed from the resource](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1b56a35f5add14c9bb439d50b3ae95368f5c107d/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs#L43) if the resource is null.

### Relevant code snippets

_No response_

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

gladjohn:
Fixes #4332.
Check for null and empty scopes in managed identity

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
@@ -47,9 +47,7 @@ namespace Microsoft.Identity.Client
         {
             if (string.IsNullOrEmpty(resource))
             {
-                throw new MsalClientException(
-                    MsalError.ManagedIdentityScopesRequired,
-                    MsalErrorMessage.ScopesRequired);
+                throw new ArgumentNullException(nameof(resource));
             }
 
             return AcquireTokenForManagedIdentityParameterBuilder.Create(
--- a/src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
@@ -45,6 +45,13 @@ namespace Microsoft.Identity.Client
         /// <inheritdoc/>
         public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIdentity(string resource)
         {
+            if (string.IsNullOrEmpty(resource))
+            {
+                throw new MsalClientException(
+                    MsalError.ManagedIdentityScopesRequired,
+                    MsalErrorMessage.ScopesRequired);
+            }
+
             return AcquireTokenForManagedIdentityParameterBuilder.Create(
                 ClientExecutorFactory.CreateManagedIdentityExecutor(this),
                 resource);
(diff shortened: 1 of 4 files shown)


trwalke (approved):

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -45,6 +45,13 @@ public sealed class ManagedIdentityApplication
         /// <inheritdoc/>
         public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIdentity(string resource)
         {
+            if (string.IsNullOrEmpty(resource))
+            {
+                throw new MsalClientException(
```
bgavrilMS, line 50: Throw ArgumentNullException instead. We shouldn't invent exceptions for this.

bgavrilMS (changes requested):
NPE is fine. We should not throw MSAL Client Ex for simple arg validation

gladjohn:
> ArgNullEx is better here. We should not throw MSAL Client Ex for simple arg validation

updated

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
pmaytak: I agree that Arg Ex is better, but this is breaking change? (Although I'd consider throwing an MsalEx a bug...)

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -300,6 +300,29 @@ await mi.AcquireTokenForManagedIdentity(resource)
...
+                var miBuilder = ManagedIdentityApplicationBuilder.Create(ManagedIdentityId.SystemAssigned)
+                    .WithHttpManager(httpManager);
+
+                var mi = miBuilder.Build();
+
+                ArgumentNullException ex = await Assert.ThrowsExceptionAsync<ArgumentNullException>(async () =>
+                    await mi.AcquireTokenForManagedIdentity(resource)
+                    .ExecuteAsync().ConfigureAwait(false)).ConfigureAwait(false);
+
+                Assert.IsNotNull(ex);
```
pmaytak, line 322: nit: not needed. The Assert above will throw if error is null.

pmaytak (approved):

bgavrilMS (approved):

pmaytak:
@gladjohn What does moving the null check accomplish? Is null resource not caught right now?

gladjohn:
> @gladjohn What does moving the null check accomplish? Is null resource not caught right now?

For MI, we need to [remove the default suffix](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1b56a35f5add14c9bb439d50b3ae95368f5c107d/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs#L43) if present. This is done before the null check. So moving the null check will not cause this to crash.

neha-bhargava (approved):

PR #4337, opened 2023-09-07
Disabling x-client-CPU

### LINKED ISSUE #4167: [Feature Request] Remove unnecessary x-client info
Currently MSAL sends the following: 

| Name             | Example Value              | Notes                                                              |
|------------------|----------------------------|--------------------------------------------------------------------|
| x-client-sku     | msal-netcore               | Identifies MSAL library (e.g. MSAL.NET on NET fwk, MSAL ObjC etc.) |
| x-client-version | 4.45.0                     |                                                                    |
| x-client-CPU     | x64                        |  |
| x-client-OS      | Win Server 2016 Datacenter | RuntimeInformation.OSDescription                                   |
| x-client-DM      | Apple Iphone 14            | Only on mobile and UWP                                             |

Understand if the service uses all this information. We should cut as much as possible irrelevant information. Even if it is low-impact from a privacy perspective, it is not 0-impact.

Proposal: cut the following

- x-client-CPU

trwalke:
Fixes #4167

Removing x-client-CPU

**Testing**
N/A

**Performance impact**
N/A

**Documentation**


trwalke:
So the initial proposal is to remove the CPU. I am wondering if it also makes sense to remove x-client-DM? We already have the OS on mobile/UWP. I have not encountered an issue on mobile where the device model is relevant. Usually the OS will suffice.

Thoughts @bgavrilMS

pmaytak:
> So the initial proposal is to remove the CPU. I am wondering if it also makes sense to remove x-client-DM? We already have the OS on mobile/UWP. I have not encountered an issue on mobile where the device model is relevant. Usually, the OS will suffice. There were some rare issues on Samsung phones related broker but those are primarily due to power management affecting broker auth.
> 
> Thoughts @bgavrilMS

We have to check with ESTS. I think they use some of these device / OS telemetry to drive what login UI experience to show, and like whether to require broker install, etc.

bgavrilMS:
> So the initial proposal is to remove the CPU. I am wondering if it also makes sense to remove x-client-DM? We already have the OS on mobile/UWP. I have not encountered an issue on mobile where the device model is relevant. Usually, the OS will suffice. There were some rare issues on Samsung phones related broker but those are primarily due to power management affecting broker auth.
> 
> Thoughts @bgavrilMS

Feel free to ping our friends from MSAL Android and iOS on this. We shouldn't be sending it on desktop.

gladjohn:
Would Processor Architecture information be useful in MSALRuntime WAM scenarios? @MSamWils ?

bgavrilMS (approved):

bgavrilMS:
> Would Processor Architecture information be useful in MSALRuntime WAM scenarios? @MSamWils ?

They can read that themselves if needed.

gladjohn (approved):

pmaytak (approved):

PR #4342, opened 2023-09-11
Removing netcore 2.1 target

### LINKED ISSUE #4313: [Engineering Task] Remove support for netcoreapp 2.1
### Type of task?

Other(s)

### Description

Remove support for netcoreapp 2.1

### Solution

_No response_

trwalke:
Fixes #4313

Removing target framework "net core 2.1"

**Testing**
Ensured tests are running on net 6 rather than net standard

**Performance impact**
N/A

**Documentation**

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-#if !NET48 && !NET7_0
+#if !NET6_0
+#if !NET48
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -12,7 +13,7 @@ using System.Threading.Tasks;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 
-#if NET6_WIN || NETCOREAPP3_1
+#if NET6_WIN
 using Microsoft.Identity.Client.Platforms.Features.RuntimeBroker;
 #endif
 
@@ -201,4 +202,5 @@ namespace Microsoft.Identity.Test.Unit.BrokerTests
         }
     }
 }
+#endif
 #endif
--- a/tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-#if !NET6_0
-#if !NET48
+#if NET6_WIN
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -12,11 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
-
-#if NET6_WIN
 using Microsoft.Identity.Client.Platforms.Features.RuntimeBroker;
-#endif
-
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Instance;
 using Microsoft.Identity.Client.Internal;
@@ -34,9 +29,6 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NSubstitute;
 using Account = Microsoft.Identity.Client.Account;
 using Microsoft.Identity.Client.AppConfig;
-#if !NET6_WIN
-using Microsoft.Identity.Client.Desktop;
-#endif
 
 namespace Microsoft.Identity.Test.Unit.BrokerTests
 {
@@ -202,5 +194,4 @@ namespace Microsoft.Identity.Test.Unit.BrokerTests
         }
     }
 }
-#endif
 #endif
--- a/tests/Microsoft.Identity.Test.Unit/CoreTests/CollectionHelperTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/CoreTests/CollectionHelperTests.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Identity.Test.Unit.CoreTests
     [TestClass]
     public class CollectionHelperTests
     {
-#if NET6_WIN
+#if NET6_WIN || NET6_0
 
         [TestMethod]
         public void EmptyStringDictionaryIsImmutable()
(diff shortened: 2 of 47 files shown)


bgavrilMS:
LGTM so far, CC @pmaytak and @neha-bhargava for closing on this.

neha-bhargava:
Remove the net core target framework from the unit tests as well? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj#L5 
Or change it to target net 6?

pmaytak:
> Remove the net core target framework from the unit tests as well? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/Microsoft.Identity.Test.Unit.csproj#L5 Or change it to target net 6?

Change it to net6.0.

trwalke:
@bgavrilMS @neha-bhargava The issue I encountered with this PR is once the net core 2 target was removed, some of the netcore 3 tests started using the netstandard runtime rather than the net 6 runtime. This resulted in some of the tests failing which indicated a breaking change in the behavior of apps targeting net core 3 now. @pmaytak and I were trying to workout the right steps to take here but I dont think we came to a final conclusion. Perhaps there were updates since I been OOF?

pmaytak:
> @bgavrilMS @neha-bhargava The issue I encountered with this PR is once the net core 2 target was removed, some of the netcore 3 tests started using the netstandard runtime rather than the net 6 runtime. This resulted in some of the tests failing which indicated a breaking change in the behavior of apps targeting net core 3 now. @pmaytak and I were trying to workout the right steps to take here but I dont think we came to a final conclusion. Perhaps there were updates since I been OOF?

I think the issue was because of tests themselves. One test was `EmptyStringDictionaryIsImmutable` and I don't remember the other. But they broke because they were testing internal methods, which are different in NET Core and NET Standard; and the tests assumed what the implementation is.

This shouldn't be an issue with the public apps using the public APIs. Removing netcoreapp2.1 target should only affect apps targeting from .NET Core 2.1 to .NET Core 3.1 which will drop to use NET Standard, which should have public API and almost all (?) logic the same as NET Core. So updating these tests should be okay.

The main TFMs we have are net461, netstandard2.0, net6.0, net6.0-windows. So, we need to (1) figure out what the code differences between the platforms are, (2) which platform has the most code (probably net6.0, this will run most tests), (3) make sure the test projects target the relevant platforms to ensure coverage or all the code.

bgavrilMS:
Once we remove net core 2 from MSAL, we should ensure there are no more references to unsupported .NET versions such as netcoreapp3.1 anywhere in the solution, including tests.

bgavrilMS:
@trwalke - please wrap this up, as it's blocking @neha-bhargava from shipping the OTel update.

trwalke:
Replacing netcore app 3 target with net 6 in tests seems to have opened up another can of worms that are a result of errors related to net6-windows. I opted to resolve the issues with the netcore 3 tests instead since they are the fastest way to unblock OTEL.

tests/Microsoft.Identity.Test.Unit/RequestsTests/FociTests.cs
```diff
@@ -116,8 +116,12 @@ public async Task FociDoesNotHideRTRefreshErrorsAsync()
                 // B cannot acquire a token interactivelty, but will try to use FRT
                 var ex = await AssertException.TaskThrowsAsync<MsalUiRequiredException>(
                     () => SilentAsync(_appB, ServerTokenResponse.OtherError)).ConfigureAwait(false);
+#if NET_CORE
+                Assert.AreEqual(MsalError.NoTokensFoundError, ex.ErrorCode);
```
trwalke, line 120: Since this is using the netstandard runtime, there is a slight difference in the error response from ESTS due to netstandard 2.0 not being on the allowed list

tests/Microsoft.Identity.Test.Unit/Kerberos/KerberosSupplementalTicketManagerTests.cs
```diff
@@ -124,6 +124,7 @@ public void GetKrbCred()
             Assert.IsNotNull(krbCred);
         }
 
+#if !NETCOREAPP3_1
```
trwalke, line 127: This test is no longer functional on netstandard due to missing types needed for the test

tests/Microsoft.Identity.Test.Unit/CoreTests/CollectionHelperTests.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Identity.Test.Unit.CoreTests
     [TestClass]
     public class CollectionHelperTests
     {
-#if NETCOREAPP
+#if NET6_WIN
```
trwalke, line 14: This test is no longer functional on netstandard due to missing types needed for the test

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-#if !NET48 && !NET7_0
+#if !NET6_0
+#if !NET48
```
bgavrilMS, line 4: if NET6_WIN ?

tests/Microsoft.Identity.Test.Unit/CoreTests/CollectionHelperTests.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Identity.Test.Unit.CoreTests
     [TestClass]
     public class CollectionHelperTests
     {
-#if NETCOREAPP
+#if NET6_WIN || NET6_0
```
bgavrilMS, line 14: Can you replace with `if NET6_OR_HIGHER` ?

tests/Microsoft.Identity.Test.Unit/Kerberos/KerberosSupplementalTicketManagerTests.cs
bgavrilMS, line 127: NETCOREAPP3_1 should no longer be defined.

bgavrilMS (approved):
Few comments, but generally LGTM

tests/Microsoft.Identity.Test.Unit/BrokerTests/RuntimeBrokerTests.cs
trwalke: Has to be net6, Broker stuff seems to not be available on Net6

tests/Microsoft.Identity.Test.Unit/Kerberos/KerberosSupplementalTicketManagerTests.cs
trwalke: yeah, removing

tests/Microsoft.Identity.Test.Unit/CoreTests/CollectionHelperTests.cs
trwalke: hmm, this directive doesn't seem to be functional
image

pmaytak: It's NET6_0_OR_GREATER

pmaytak (approved):

PR #4347, opened 2023-09-15
Throw ArgumentException for malformed authorities

### LINKED ISSUE #4280: [Bug] .WithTenantId started to throw ArgumentException
### Library version used

4.55.0

### .NET version

4.8

### MSAL client type

Confidential

### Application type

Web API

### Authentication flow type

Client credentials (service-to-service calls)

Well calling on the ConfidentialClientApplication builder .WithTenantId with invalid tenant, a new argument exception regarding authorityurl is being throw, although we do not explictly set authorityurl by ourselves, and this is not declared as a thrown exception.

Until now we handled MsalException for those scenario. As discussed offline with Bogdan suggest to change to MsalClientException not to break flows.

For now we have rolled back to previous MSAL version.

### Relevant code snippets

_No response_

### Expected behavior

MsalClientException should be thrown , better on the Execute and not in the builder (even if MSAL itself will not go our for STS)

### Identity provider

### Regression

4.50

### Solution and workarounds

_No response_

pmaytak:
Fixes #4280

Currently we have 17 ways to set authority or tenant (plus 7 deprecated). If the authority is malformed (like having a space) we don't throw the same exception - mostly `ArgumentException` but sometimes we just send the request to AAD and propagate `MsalServiceException` back. We've refactored authority related logic in 4.55.0 changing some exception types. See table in the appendix. 

The proposal is to always throw an `ArgumentException` here.
- [ArgumentException ](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception?view=net-7.0#remarks)is thrown when a method is invoked and at least one of the passed arguments does not meet the parameter specification of the called method. Also according to guidelines - ["The exception class thrown is the most specific exception available that fits the error conditions."](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/creating-and-throwing-exceptions). So, `ArgumentException` is more appropriate here, [`MsalClientException`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identity.client.msalclientexception?view=msal-dotnet-latest) should be used for more general error scenarios.
- We shouldn't send a request to AAD, if we know the arguments are invalid.
- Throwing the exception at the correct level - app or request - is more accurate.

While changing the exception type of a scenario is considered a breaking change, I would consider this a bug fix instead. We already throw `ArgumentException`s for other errors, so these types of user errors and configuration errors should already be caught in a general `catch` filter.

The code that checked for malformation already was in the older versions. However, there were two ways to reach it - one that used an original authority string and another which used a `Uri.ToString` on a URI authority. Different authority setters used these two different methods. A `Uri` class URL encodes by default.  So in the former case, the string would have a space in it and would get caught in the check; in the latter case, the string would have encoded spaces, and would not get caught, and would send this malformed URI to AAD, getting back an exception. The solution is to always call `ValidateAndCreateAuthorityUri` in both `AuthorityInfo` constructors.
image

- Updates authority handling code to throw `ArgumentException` instead of `MsalClientException` or `MsalServiceException`.
- Updates comments
- Updates and adds tests

**Testing**
Updated and added unit tests. 
Manual testing.

**Performance impact**
N/A

**Documentation**

**Appendix**

This table shows the differences between versions 4.50 and 4.55 in which exceptions are thrown and at which level - app or request. Some show `N/A` because the authority/tenant is not string, so can't be malformed. In 4.50, we didn't allow using `WithTenant` and `WithAuthority` at app level at the same time; in 4.55, it was changed to `WithTenant` overriding `WithAuthority`. The last columns show proposed changed in this PR.

image

We also have these methods that are deprecated, but I kept out of scope for this PR.
```csharp
AbstractAcquireTokenParameterBuilder.WithAuthority(AadAuthorityAudience, validateAuthority)
AbstractAcquireTokenParameterBuilder.WithAuthority(authorityUri, validateAuthority)
AbstractAcquireTokenParameterBuilder.WithAuthority(cloudInstanceUri, tenantId, validateAuthority)
AbstractAcquireTokenParameterBuilder.WithAuthority(cloudInstanceUri, tenant, validateAuthority)
AbstractAcquireTokenParameterBuilder.WithAuthority(azureCloudInstance, tenantId, validateAuthority)
AbstractAcquireTokenParameterBuilder.WithAuthority(azureCloudInstance, tenant, validateAuthority)
AbstractAcquireTokenParameterBuilder.WithAuthority(azureCloudInstance, authorityAudience, validateAuthority)
```


tests/Microsoft.Identity.Test.Unit/MetricsTests.cs
```diff
@@ -162,49 +162,6 @@ public async Task RefreshReasonExpired_ConfidentialClient_Async()
             }
         }
 
-        [TestMethod]
-        public void IdWebAccountExtension()
```
pmaytak, line 166: Moved to AccountTests

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
```diff
@@ -22,7 +22,7 @@ internal class AuthorityInfo
             AuthorityType authorityType,
             string authority,
             bool validateAuthority)
-            : this(authorityType, new Uri(authority), validateAuthority)
+            : this(authorityType, ValidateAndCreateAuthorityUri(authority, authorityType), validateAuthority)
```
pmaytak, line 25: This is the main change.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -254,13 +254,16 @@ public T WithAuthority(AadAuthorityAudience authorityAudience, bool validateAuth
         /// 
         /// If an authority was not specified at the application level, the default used is https://login.microsoftonline.com/common.
         /// </summary>
-        /// <param name="tenantId">The tenant ID, which can be either in GUID format or a domain name. Also known as the Directory ID.</param>
+        /// <param name="tenantId">tenant ID of the Azure AD tenant
```
localden, line 257:
```suggestion
        /// <param name="tenantId">Tenant ID of the Microsoft Entra ID tenant
```

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -254,13 +254,16 @@ public T WithAuthority(AadAuthorityAudience authorityAudience, bool validateAuth
         /// 
         /// If an authority was not specified at the application level, the default used is https://login.microsoftonline.com/common.
         /// </summary>
-        /// <param name="tenantId">The tenant ID, which can be either in GUID format or a domain name. Also known as the Directory ID.</param>
+        /// <param name="tenantId">tenant ID of the Azure AD tenant
+        /// or a domain associated with this Azure AD tenant, in order to sign-in a user of a specific organization only</param>
```
localden, line 258:
```suggestion
        /// or a domain associated with this Microsoft Entra ID tenant, in order to sign-in a user of a specific organization only.</param>
```

src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs
```diff
@@ -280,17 +283,17 @@ public T WithTenantId(string tenantId)
...
         /// If an authority was not provided to the application builder, this method will replace the tenant ID in the default authority - https://login.microsoftonline.com/common.
         /// </summary>
         /// <param name="authorityUri">URI from which to extract the tenant ID</param>
-
         /// <returns>The builder to chain the .With methods.</returns>
-        /// <exception cref="ArgumentNullException">Thrown if <paramref name="authorityUri"/> is null or an empty string</exception>
-        /// <exception cref="MsalClientException">Thrown if the application was configured with an authority that is not AAD specific (e.g. ADFS or B2C).</exception>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="authorityUri"/> is null or an empty string.</exception>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="authorityUri"/> is not well-formatted (for example, has spaces).</exception>
+        /// <exception cref="MsalClientException">Thrown in more general exception scenarios (for ex. if the application was configured with an authority that does not allow tenants).</exception>
```
localden, line 292:
```suggestion
        /// <exception cref="MsalClientException">Thrown in general exception scenarios (for example if the application was configured with an authority that does not allow tenants).</exception>
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -428,16 +428,19 @@ internal override ApplicationConfiguration BuildConfiguration()
         }
 
         #region Authority
-        
+
 
         /// <summary>
-        /// Adds a known authority to the application from its Uri. See https://aka.ms/msal-net-application-configuration.
+        /// Adds a known authority to the application. See https://aka.ms/msal-net-application-configuration.
```
localden, line 434:
```suggestion
        /// Adds a known authority to the application. See <see href="https://aka.ms/msal-net-application-configuration">Application configuration options</see>.
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -428,16 +428,19 @@ internal override ApplicationConfiguration BuildConfiguration()
...
         /// This constructor is mainly used for scenarios where the authority is not a standard Azure AD authority,
         /// nor an ADFS authority, nor an Azure AD B2C authority. For Azure AD, even in sovereign clouds, prefer
         /// using other overrides such as <see cref="WithAuthority(AzureCloudInstance, AadAuthorityAudience, bool)"/>
         /// </summary>
-        /// <param name="authorityUri">Uri of the authority</param>
+        /// <param name="authorityUri">URI of the authority</param>
         /// <param name="validateAuthority">Whether the authority should be validated against the server metadata.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="authorityUri"/> is null.</exception>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="authorityUri"/> is not well-formatted (for example, has spaces).</exception>
+        /// <exception cref="MsalClientException">Thrown in more general exception scenarios (for ex. if the application was configured with multiple different authority hosts).</exception>
```
localden, line 443:
```suggestion
        /// <exception cref="MsalClientException">Thrown in general exception scenarios (for example if the application was configured with multiple different authority hosts).</exception>
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -451,19 +454,22 @@ public T WithAuthority(Uri authorityUri, bool validateAuthority = true)
 
         /// <summary>
         /// Adds a known Azure AD authority to the application to sign-in users specifying
-        /// the full authority Uri. See https://aka.ms/msal-net-application-configuration.
+        /// the full authority URI. See https://aka.ms/msal-net-application-configuration.
```
localden, line 457:
```suggestion
        /// the full authority URI. See <see href="https://aka.ms/msal-net-application-configuration">Application configuration options</see>.
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -479,11 +485,14 @@ public T WithAuthority(string authorityUri, bool validateAuthority = true)
 
         /// <summary>
         /// Adds a known Azure AD authority to the application to sign-in users from a single
-        /// organization (single tenant application) specified by its tenant ID. See https://aka.ms/msal-net-application-configuration.
+        /// organization (single-tenant application) specified by its tenant ID. See https://aka.ms/msal-net-application-configuration.
```
localden, line 488:
```suggestion
        /// organization (single-tenant application) specified by its tenant ID. See <see href="https://aka.ms/msal-net-application-configuration">Application configuration options</see>.
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -451,19 +454,22 @@ public T WithAuthority(Uri authorityUri, bool validateAuthority = true)
...
         ///  <item><description><c>https://login.microsoftonline.com/organizations/</c> to sign-in users with any work and school accounts</description></item>
         ///  <item><description><c>https://login.microsoftonline.com/consumers/</c> to sign-in users with only personal Microsoft accounts (live)</description></item>
         ///  </list>
         ///  Note that this setting needs to be consistent with what is declared in the application registration portal</param>
         /// <param name="validateAuthority">Whether the authority should be validated against the server metadata.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="authorityUri"/> is null or empty.</exception>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="authorityUri"/> is not well-formatted (for example, has spaces).</exception>
+        /// <exception cref="MsalClientException">Thrown in more general exception scenarios (for ex. if the application was configured with multiple different authority hosts).</exception>
```
localden, line 472:
```suggestion
        /// <exception cref="MsalClientException">Thrown in general exception scenarios (for example if the application was configured with multiple different authority hosts).</exception>
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -479,11 +485,14 @@ public T WithAuthority(string authorityUri, bool validateAuthority = true)
...
-        /// organization (single tenant application) specified by its tenant ID. See https://aka.ms/msal-net-application-configuration.
+        /// organization (single-tenant application) specified by its tenant ID. See https://aka.ms/msal-net-application-configuration.
         /// </summary>
-        /// <param name="cloudInstanceUri">Azure Cloud instance.</param>
+        /// <param name="cloudInstanceUri">Azure cloud instance.</param>
         /// <param name="tenantId">GUID of the tenant from which to sign-in users.</param>
         /// <param name="validateAuthority">Whether the authority should be validated against the server metadata.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="cloudInstanceUri"/> is null or empty.</exception>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="cloudInstanceUri"/> is not well-formatted (for example, has spaces).</exception>
+        /// <exception cref="MsalClientException">Thrown in more general exception scenarios (for ex. if the application was configured with multiple different authority hosts).</exception>
```
localden, line 495:
```suggestion
        /// <exception cref="MsalClientException">Thrown in more general exception scenarios (for example if the application was configured with multiple different authority hosts).</exception>
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -508,6 +517,9 @@ public T WithAuthority(string authorityUri, bool validateAuthority = true)
         /// it's recommended to use another override (<see cref="WithAuthority(AzureCloudInstance, Guid, bool)"/>
         /// and <see cref="WithAuthority(AzureCloudInstance, AadAuthorityAudience, bool)"/>
         /// </remarks>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="cloudInstanceUri"/> or <paramref name="tenant"/> is null or empty.</exception>
+        /// <exception cref="ArgumentException">Thrown if <paramref name="cloudInstanceUri"/> or <paramref name="tenant"/> is not well-formatted (for example, has spaces).</exception>
+        /// <exception cref="MsalClientException">Thrown in more general exception scenarios (for ex. if the application was configured with multiple different authority hosts).</exception>
```
localden, line 522:
```suggestion
        /// <exception cref="MsalClientException">Thrown in more general exception scenarios (for example if the application was configured with multiple different authority hosts).</exception>
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -537,8 +549,8 @@ public T WithAuthority(string authorityUri, bool validateAuthority = true)
         /// organization (single tenant application) described by its cloud instance and its tenant ID.
         /// See https://aka.ms/msal-net-application-configuration.
         /// </summary>
-        /// <param name="azureCloudInstance">Instance of Azure Cloud (for instance Azure
-        /// worldwide cloud, Azure German Cloud, US government ...)</param>
+        /// <param name="azureCloudInstance">Instance of Azure cloud (for instance, Azure
```
localden, line 552:
```suggestion
        /// <param name="azureCloudInstance">Instance of Azure cloud (for example, Azure
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -556,11 +568,12 @@ public T WithAuthority(string authorityUri, bool validateAuthority = true)
         /// organization (single tenant application) described by its cloud instance and its domain
         /// name or tenant ID. See https://aka.ms/msal-net-application-configuration.
         /// </summary>
-        /// <param name="azureCloudInstance">Instance of Azure Cloud (for instance Azure
-        /// worldwide cloud, Azure German Cloud, US government ...).</param>
+        /// <param name="azureCloudInstance">Instance of Azure cloud (for instance, Azure
```
localden, line 571:
```suggestion
        /// <param name="azureCloudInstance">Instance of Azure cloud (for example, Azure
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -556,11 +568,12 @@ public T WithAuthority(string authorityUri, bool validateAuthority = true)
...
+        /// worldwide cloud, Azure German Cloud, US government).</param>
```
localden, line 572:
```suggestion
        /// public cloud, Azure Germany, or Azure Government).</param>
```

src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs
```diff
@@ -537,8 +549,8 @@ public T WithAuthority(string authorityUri, bool validateAuthority = true)
         /// organization (single tenant application) described by its cloud instance and its tenant ID.
         /// See https://aka.ms/msal-net-application-configuration.
         /// </summary>
-        /// <param name="azureCloudInstance">Instance of Azure Cloud (for instance Azure
-        /// worldwide cloud, Azure German Cloud, US government ...)</param>
+        /// <param name="azureCloudInstance">Instance of Azure cloud (for instance, Azure
+        /// worldwide cloud, Azure German Cloud, US government).</param>
```
localden, line 553:
```suggestion
        /// public cloud, Azure Germany, or Azure Government).</param>
```

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/B2cAuthorityTests.cs
```diff
@@ -17,42 +15,34 @@ namespace Microsoft.Identity.Test.Unit.CoreTests.InstanceTests
...
-                Assert.AreEqual(instance.AuthorityInfo.AuthorityType, AuthorityType.B2C);
-
-                Assert.Fail("test should have failed");
-            }
-            catch (Exception exc)
-            {
-                Assert.IsInstanceOfType(exc, typeof(ArgumentException));
-                Assert.AreEqual(MsalErrorMessage.B2cAuthorityUriInvalidPath, exc.Message);
-            }
+            var ex = Assert.ThrowsException<ArgumentException>(() =>
```
localden, line 20: Explicit type?

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/B2cAuthorityTests.cs
```diff
@@ -95,5 +85,27 @@ public void CanonicalAuthorityInitTest()
...
+        public void MalformedAuthority_ThrowsException(string malformedAuthority)
+        {
+            Assert.ThrowsException<ArgumentException>(() =>
+                ConfidentialClientApplicationBuilder
+                    .Create(TestConstants.ClientId)
+                    .WithB2CAuthority(malformedAuthority)
+                    .WithClientSecret(TestConstants.ClientSecret)
+                    .Build());
+
+            var app = ConfidentialClientApplicationBuilder
```
localden, line 101: For consistency, worth making this an explicit type.

localden (changes requested):

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/B2cAuthorityTests.cs
pmaytak: We don't really have a strict rule of implicit vs explicit variable typing. Generally we use `var`s unless it's more readable to use an explicit type. Here it's clear that it's `ArgumentException`.

src/client/Microsoft.Identity.Client/AppConfig/AuthorityInfo.cs
gladjohn: Adding some comments in the `ValidateAndCreateAuthorityUri` method for each of the checks we are doing will be good to know the validation coverage we have.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AccountTests.cs
```diff
@@ -369,6 +370,48 @@ public void GetAccountsAndSignThemOutTest()
...
+            var accId = new AccountId($"{oid}.{tid}", oid, tid);
+            account.HomeAccountId.Returns(accId);
+
+            var claimsIdentityResult = ToClaimsPrincipal(account).Identity as ClaimsIdentity;
+
+            Assert.IsNotNull(claimsIdentityResult);
+            Assert.AreEqual(3, claimsIdentityResult.Claims.Count());
+            Assert.AreEqual(username, claimsIdentityResult.FindFirst(ClaimTypes.Upn)?.Value);
+            Assert.AreEqual(oid, claimsIdentityResult.FindFirst("oid")?.Value);
+            Assert.AreEqual(tid, claimsIdentityResult.FindFirst("tid")?.Value);
```
gladjohn, line 412:
```suggestion
            Assert.IsNotNull(claimsIdentityResult, "The ClaimsIdentity should not be null.");
Assert.AreEqual(3, claimsIdentityResult.Claims.Count(), "Expected 3 claims in the ClaimsIdentity.");
Assert.AreEqual(username, claimsIdentityResult.FindFirst(ClaimTypes.Upn)?.Value, "UPN claim value should match.");
Assert.AreEqual(oid, claimsIdentityResult.FindFirst("oid")?.Value, "OID claim value should match.");
Assert.AreEqual(tid, claimsIdentityResult.FindFirst("tid")?.Value, "TID claim value should match.");

```

gladjohn (approved):

bgavrilMS (approved):

bgavrilMS (changes requested):
Let's discuss about this, I'd like to revert back to previous exceptions as this is breaking Kusto Explorer.

pmaytak:
@rabenshm What is your opinion on the proposed changes?

The basics of my opinion above is that our authority handling code was incorrect (I consider it a bug) and throwing an `ArgumentException` is more correct. Also we throw `ArgumentException`s in other places, so catching only `MsalException` will not catch all mis-configuration errors anyway.

bgavrilMS (approved):

PR #4350, opened 2023-09-17
[Prototype] Remove net6-win

pmaytak:
Small prototype of changes that will be made if net6-win target is removed. This updates just the main MSAL project; will need to update the tests also by removing the net6-win target there and running the broker tests on netcoreapp.


bgavrilMS:
Did we commit to this?

pmaytak:
> Did we commit to this?

No. This is prototype work from the investigation phase.

PR #4375, opened 2023-10-12
Print refresh token cache keys to logs

### LINKED ISSUE #3957: [Feature Request] When verbose logging is enabled, print cache keys of all refresh tokens in the cache (same as we do for access tokens
When verbose logging is enabled, print cache keys of all refresh tokens in the cache (same as we do for access tokens [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/16b744a4c56efb20073693c081a3512aa385da25/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L286)).

pmaytak:
Fixes #3957

Print refresh token cache keys to logs

**Testing**
Unit tests. Manual testing

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/Items/MsalRefreshTokenCacheItem.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/Items/MsalRefreshTokenCacheItem.cs
@@ -81,6 +81,17 @@ namespace Microsoft.Identity.Client.Cache.Items
             iOSCacheKeyLazy = new Lazy<IiOSKey>(() => InitiOSKey());
         }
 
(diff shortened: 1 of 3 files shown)


trwalke (approved):

neha-bhargava (approved):

gladjohn (approved):
should we add test coverage for pii cases?

pmaytak:
> should we add test coverage for pii cases? can we?

Good catch. Added tests for ToLogString in AT/RT cache items.

gladjohn (approved):

PR #4376, opened 2023-10-13
Add serialization of broker error data in exceptions

### LINKED ISSUE #4371: [Bug] ToJsonString does not serialize all fields of MSAL exception leading to data loss when deserialized
### Library version used

4.54

### .NET version

.net framework 4.7 .net 6, net 8

### Scenario

PublicClient - desktop app

When the broker additional fields and telemetry fields were added the tojson and fromjson methods were not updated to include serializing the additional fields. 

### Relevant code snippets

```csharp
MSAL Exception
    public string ToJsonString()
    public static MsalException FromJsonString(string json)

are missing
        public const string BrokerErrorContext = "BrokerErrorContext";
        public const string BrokerErrorTag = "BrokerErrorTag";
        public const string BrokerErrorStatus = "BrokerErrorStatus";
        public const string BrokerErrorCode = "BrokerErrorCode";
        public const string BrokerTelemetry = "BrokerTelemetry";
```

### Expected behavior

Those fields should also be serialized. 

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

pmaytak:
Fixes #4371

- Adds additional broker exception data to exception de/serialization.
- Added a flag for less escaping in STJ.
   - System.Text.Json, by default, escapes more characters than Newtonsoft. This makes broker telemetry, which is a JSON string, less readable (see [doc](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-7-0#minimal-character-escaping) and [doc](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/character-encoding#serialize-all-characters)). Something like: 
```text
"broker_telemetry": "{\u0022additional_query_parameters_count\u0022:\u00220\u0022,\u0022all_error_tags\u0022:\u00229zdfv\u0022,\u0022api_error_code\u0022:\u00220\u0022,\u0022api_error_context\u0022:\u0022
```
- When serializing broker data, using snake case for property names for consistency. Like this:
```json
{
  "type": "MsalException",
  "error_code": "some_error_code",
  "error_description": "Some error message.",
  "additional_exception_data": {
    "broker_error_context": "broker error context",
    "broker_error_tag": "0x123456",
    "broker_error_status": "broker error status",
    "broker_error_code": "broker error code",
    "broker_telemetry": "broker telemetry"
  }
}
```
Otherwise, if using property name from the `AdditionalExceptionData` dictionary, JSON would look like this:
```json
{
	"type": "MsalException",
	"error_code": "some_error_code",
	"error_description": "Some error message.",
	"AdditionalExceptionData": {
	  "BrokerErrorContext": "broker error context",
	  "BrokerErrorTag": "0x123456",
	  "BrokerErrorStatus": "broker error status",
	  "BrokerErrorCode": "broker error code",
	  "BrokerTelemetry": "broker telemetry"
	}
```

**Testing**
Manual + unit tests.

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalException.cs
+++ b/src/client/Microsoft.Identity.Client/MsalException.cs
@@ -168,10 +168,7 @@ namespace Microsoft.Identity.Client
                 innerExceptionContents);
         }
 
-        #region SERIALIZATION
-
-        // DEPRECATE / OBSOLETE - this functionality is not used and should be removed in a next major version
-
+        #region Serialization
         private class ExceptionSerializationKey
         {
             internal const string ExceptionTypeKey = "type";
@@ -191,6 +188,7 @@ namespace Microsoft.Identity.Client
             jObject[ExceptionSerializationKey.ErrorCodeKey] = ErrorCode;
             jObject[ExceptionSerializationKey.ErrorDescriptionKey] = Message;
 
+            // Populate JSON string with broker exception data
             var exceptionData = new JObject();
 
             if (AdditionalExceptionData.TryGetValue(BrokerErrorContext, out string brokerErrorContext))
@@ -219,6 +217,7 @@ namespace Microsoft.Identity.Client
 
         internal virtual void PopulateObjectFromJson(JObject jObject)
         {
+            // Populate this exception instance with broker exception data from JSON
             var exceptionData = JsonHelper.ExtractInnerJsonAsDictionary(jObject, ExceptionSerializationKey.AdditionalExceptionData);
 
             if (exceptionData.TryGetValue(ExceptionSerializationKey.BrokerErrorContext, out string brokerErrorContext))
@@ -299,6 +298,6 @@ namespace Microsoft.Identity.Client
             return ex;
         }
 
-        #endregion // SERIALIZATION
+        #endregion
     }
 }
--- a/src/client/Microsoft.Identity.Client/MsalException.cs
+++ b/src/client/Microsoft.Identity.Client/MsalException.cs
@@ -8,7 +8,9 @@ using Microsoft.Identity.Client.Internal;
 using Microsoft.Identity.Client.PlatformsCommon.Factories;
 using Microsoft.Identity.Client.Utils;
 #if SUPPORTS_SYSTEM_TEXT_JSON
-using System.Text.Json.Nodes;
+using System.Text.Json;
+using System.Text.Encodings.Web;
+using System.Text.Unicode;
 using JObject = System.Text.Json.Nodes.JsonObject;
 #else
 using Microsoft.Identity.Json;
@@ -256,7 +258,19 @@ namespace Microsoft.Identity.Client
         {
             JObject jObject = new JObject();
             PopulateJson(jObject);
+
+#if SUPPORTS_SYSTEM_TEXT_JSON
+            // By default, STJ is more restrictive (escapes more characters) than Newtonsoft,
+            // so the telemetry string is less readable.
+            // Relax the encoding rules to match Newtonsoft behavior.
+            return jObject.ToJsonString(new JsonSerializerOptions()
+            {
+                WriteIndented = true,
+                Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping,
+            });
+#else
             return jObject.ToString();
+#endif
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/MsalException.cs
```diff
@@ -170,19 +172,82 @@ public override string ToString()
 
         // DEPRECATE / OBSOLETE - this functionality is not used and should be removed in a next major version
 
-        private const string ExceptionTypeKey = "type";
-        private const string ErrorCodeKey = "error_code";
-        private const string ErrorDescriptionKey = "error_description";
+        private class ExceptionSerializationKey
+        {
+            internal const string ExceptionTypeKey = "type";
```
gladjohn, line 177: would it make the below if statement to be cleaner with a foreach of these constants were added to a dict? 

```cs
if (AdditionalExceptionData.TryGetValue(key, out string value))
{
     exceptionData[key] = value;
}
```

src/client/Microsoft.Identity.Client/MsalException.cs
```diff
@@ -170,19 +172,82 @@ public override string ToString()
 
         // DEPRECATE / OBSOLETE - this functionality is not used and should be removed in a next major version
```
gladjohn, line 173:
```suggestion
        // DEPRECATE / OBSOLETE - this functionality is not used and should be removed in a next major version
```

should remove this, since we know it is being used

src/client/Microsoft.Identity.Client/MsalException.cs
```diff
@@ -170,19 +172,82 @@ public override string ToString()
...
+        internal virtual void PopulateJson(JObject jObject)
         {
-            jobj[ExceptionTypeKey] = GetType().Name;
-            jobj[ErrorCodeKey] = ErrorCode;
-            jobj[ErrorDescriptionKey] = Message;
+            jObject[ExceptionSerializationKey.ExceptionTypeKey] = GetType().Name;
+            jObject[ExceptionSerializationKey.ErrorCodeKey] = ErrorCode;
+            jObject[ExceptionSerializationKey.ErrorDescriptionKey] = Message;
+
+            var exceptionData = new JObject();
```
trwalke, line 194: nit:For this method and maybe the one below it, I would put these if statements in a seperate method and call them something like `PopulateWithBrokerExceptionData`. Especially since we may end up adding more stuff here in the future.

trwalke (approved):
minor comment. LGTM

src/client/Microsoft.Identity.Client/MsalException.cs
pmaytak: The property names in object and in JSON are different. So simple loop won't suffice - has to use some dictionary mapping, so this seems simpler.

src/client/Microsoft.Identity.Client/MsalException.cs
pmaytak: I added comments before these code blocks instead. Tried extracting into a method but method names become too long to read.

tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionSerializationTests.cs
```diff
@@ -18,36 +19,43 @@ public class MsalExceptionSerializationTests
...
 
-            var exDeserialized = MsalException.FromJsonString(json);
-
-            Assert.AreEqual(expectedType, exDeserialized.GetType());
-            Assert.AreEqual(ex.ErrorCode, exDeserialized.ErrorCode);
-            Assert.AreEqual(ex.Message, exDeserialized.Message);
+        [DataTestMethod]
+        [DataRow(false)]
+        [DataRow(true)]
+        public void MsalException_CanSerializeAndDeserializeRoundTrip(bool includeAdditionalData)
```
gladjohn, line 31:
```suggestion
        public void MsalException_CanSerializeAndDeserializeRoundTrip(bool includeAdditionalExceptionData)
```

tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionSerializationTests.cs
```diff
@@ -18,36 +19,43 @@ public class MsalExceptionSerializationTests
...
+        {
+            var ex = new MsalException(SomeErrorCode, SomeErrorMessage);
 
-            if (isServiceExceptionDerived)
+            if (includeAdditionalData)
```
gladjohn, line 35:
```suggestion
            if (includeAdditionalExceptionData)
```

gladjohn (approved):

PR #4379, opened 2023-10-16
Add a Design Proposal issue template

kellyyangsong:


jennyf19:
@pmaytak we are working on an all up contribution guide so would like to have a similar approach for all our libraries. This is following with aspnetcore does.

jennyf19 (approved):

westin-m (approved):

pmaytak:
> @pmaytak we are working on an all up contribution guide so would like to have a similar approach for all our libraries. This is following with aspnetcore does.

If this is for consistency, then let's just merge. 

My first thought is that this has an overlap with New Feature and Bug issue types and we do design discussion within those issues. Looking through aspnetcore issues, seems like this Design proposal is more of an "enhancement to an existing feature" rather than "add new feature" or "existing feature doesn't completely works". To add clarity my feedback is to specify the above, this is an enhancement proposal (if the assumption is correct).

PR #4384, opened 2023-10-17
Remove Net45 

gladjohn:
for a private preview drop to VS for testing without net45


gladjohn:
here is the build retained if we ever need to test it : https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1193344&view=results

PR #4395, opened 2023-10-26
Fall back to local cache if broker fails to return result for silent token call 

### LINKED ISSUE #4209: [Bug] Broker AcquireTokenInteractive() doesn't cache when it fallbacks
**Logs and network traces**
{MSAL.Xamarin.Android.4.54.1.0.MsalUiRequiredException: 
	ErrorCode: failed_to_acquire_token_silently_from_broker
Microsoft.Identity.Client.MsalUiRequiredException: Broker could not satisfy the silent request.
  at Microsoft.Identity.Client.Internal.Requests.Silent.SilentRequest.ExecuteAsync (System.Threading.CancellationToken cancellationToken) [0x002d7] in <f5e0daf1cb4c43c889fca0f3b0814e85>:0 
  at Microsoft.Identity.Client.Internal.Requests.RequestBase.RunAsync (System.Threading.CancellationToken cancellationToken) [0x001c5] in <f5e0daf1cb4c43c889fca0f3b0814e85>:0 
  at Microsoft.Identity.Client.ApiConfig.Executors.ClientApplicationBaseExecutor.ExecuteAsync (Microsoft.Identity.Client.ApiConfig.Parameters.AcquireTokenCommonParameters commonParameters, Microsoft.Identity.Client.ApiConfig.Parameters.AcquireTokenSilentParameters silentParameters, System.Threading.CancellationToken cancellationToken) [0x00144] in <f5e0daf1cb4c43c889fca0f3b0814e85>:0 
  at WiFiMessenger.Services.Communication.OIDCToken.StartRefreshingTokensAsync () [0x0010a] in F:\Code\Work\Wi-Fi Messenger\Client\Client\WiFiMessenger\Services\Communication\OIDCToken.cs:189 
	StatusCode: 0 
	ResponseBody:  
	Headers: }

4.54.1

**Platform**
Xamarin.Android

* Desktop / Mobile
    * [x] Interactive (broker)
    * [x] Interactive (embedded or system browser)
    * [ ] Username / Password
    * [ ] Device code (browserless)
* Web app 
    * [ ] Authorization code
    * [ ] On-Behalf-Of
* Daemon app 
    * [ ] Client credentials (service-to-service calls)

new

**Actual behavior**
When I get a token using AcquireTokenInteractive with my PublicClientApplication built WithBroker(), and the device does not have a broker installed, so it should fall back to webview. When I get that token I try to refresh it periodically using AcquireTokenSilent and the account hint that was in the token. The acquire fails and gives above exception. When I try the same with the WithBroker() commented out, so not using broker flow. Everything works as expected.

**Expected behavior**
use the cached token that was gathered seconds ago using the interactive login

**Possible solution**
If i call 
```csharp
.WithBroker( PublicClientApplicationBuilder.Create( applicationId ).IsBrokerAvailable());
```
I understand this is deprecated but it works for me in the meantime until there is an official fix

ashok672:
Fixes #4209

WithBroker = true, interactive token call falls back correctly to browser and completes the flow. With silent token call for the same account, currently there is no fallback to local cache if broker fails to return result. BrokerStrategy can fail for example if the broker is not available in the machine. The fix is to fall back to local cache for silent token call as well similar to interactive call. 

**Testing**
The existing unit tests were modified to account for the new bahavior
Ran through the scenario manually in the WinForms test app and verified its working as expected

**Performance impact**
None

**Documentation**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
@@ -66,19 +66,11 @@ namespace Microsoft.Identity.Client.Internal.Requests.Silent
                     _logger.Info("Broker is configured and enabled, attempting to use broker instead.");
                     var brokerResult = await _brokerStrategyLazy.Value.ExecuteAsync(cancellationToken).ConfigureAwait(false);
 
+                    // fallback to local cache if broker fails
                     if (brokerResult != null)
                     {
                         _logger.Verbose(() => "Broker responded to silent request.");
                         return brokerResult;
-                    }
-                    else
-                    {
-                        _logger.Verbose(() => "Broker could not satisfy the silent request.");
-                        throw new MsalUiRequiredException(
-                           MsalError.FailedToAcquireTokenSilentlyFromBroker,
-                           "Broker could not satisfy the silent request.",
-                           null,
-                           UiRequiredExceptionClassification.AcquireTokenSilentFailed);
                     }
                 }
 
--- a/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/BrokerTests/BrokerRequestTests.cs
@@ -526,7 +526,7 @@ namespace Microsoft.Identity.Test.Unit.BrokerTests
                 }
                 catch (MsalUiRequiredException e)
                 {
-                    Assert.AreEqual("failed_to_acquire_token_silently_from_broker", e.ErrorCode);
+                    Assert.AreEqual("no_tokens_found", e.ErrorCode);
 
                     harness.HttpManager.AddInstanceDiscoveryMockHandler();
                     harness.HttpManager.AddSuccessTokenResponseMockHandlerForPost();
--- a/tests/Microsoft.Identity.Test.Unit/RequestsTests/SilentRequestTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/RequestsTests/SilentRequestTests.cs
@@ -130,25 +130,15 @@ namespace Microsoft.Identity.Test.Unit.RequestsTests
 
                 var request = new SilentRequest(harness.ServiceBundle, parameters, silentParameters);
 
-                if (brokerConfiguredByUser) //When broker is enabled, local cache should not be used
-                {
-                    var exception = await AssertException.TaskThrowsAsync<MsalUiRequiredException>(() =>
-                        request.RunAsync(default)
-                        ,false
-                        ).ConfigureAwait(false);
+                var result = await request.RunAsync(default).ConfigureAwait(false);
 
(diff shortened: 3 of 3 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -66,20 +66,12 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                     _logger.Info("Broker is configured and enabled, attempting to use broker instead.");
                     var brokerResult = await _brokerStrategyLazy.Value.ExecuteAsync(cancellationToken).ConfigureAwait(false);
 
+                    // fallback to local cache if broker fails
```
bgavrilMS, line 69: if broker is not installed, why is the flag `isBrokerConfigured` set to true?

rayluo: Just my 2 cents. Inferred from [LennoxP90's repro steps](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4209#issuecomment-1632998180), the flag `isBrokerConfigured` in MSAL .Net is presumably only determined by whether the app developer opted in to use broker, not by whether the broker is installed.

Regardless of the implementation detail above, my higher-level question is, does the fact that we are reviewing implementation details imply that we all agree with the direction of this PR, quoted again from the PR's description? @bgavrilMS 

> Changes proposed in this request
WithBroker = true, interactive token call falls back correctly to browser and completes the flow. **With silent token call for the same account, currently there is no fallback to local cache if broker fails to return result. BrokerStrategy can fail for example if the broker is not available in the machine. The fix is to fall back to local cache for silent token call** as well similar to interactive call.

bgavrilMS: @rayluo - actually `isBrokerConfigured` in this context is `isBrokerEnabled by app developer` AND `is broker available on the OS`. I would have expected the `is broker available on the OS` to fail. 

I agree that the flag is poorly named, but it's just a local variable.

bgavrilMS: Otherwise yes, I agree that silent auth should try to use MSAL's token cache ("client strategy") if broker is not available. 

This is different from the following scenario: 

- app users browser and this puts an RT into the cache
- app then moves to use broker
- silent auth is called

Expected: app MUST ignore the cache as long as the broker is available

The reason for this is that the browser RT is not bound (it's not PRT) and it will break the CA policy related to broker.

pmaytak: > I agree that the flag is poorly named, but it's just a local variable.

Should be renamed then. #CleanCode

rayluo (approved):

bgavrilMS (approved):

PR #4413, opened 2023-11-08
[DRAFT] Add CAE support for Managed Identity

gladjohn:
[DRAFT] Add CAE support for Managed Identity


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs
```diff
@@ -59,6 +59,25 @@ public AcquireTokenForManagedIdentityParameterBuilder WithForceRefresh(bool forc
             return this;
         }
 
+        /// <summary>
+        /// Sets claims in the query. Use when the AAD admin has enabled conditional access evaluation. Acquiring the token normally will result in a
```
bgavrilMS, line 63: Add a remark that this API is not always available, depending on the client and that apps can monitor this by using `msi.IsClaimsAvailable()`

bgavrilMS, line 63: Proposed comment

`Adds a claims challenge to the token request. The SDK will bypass the token cache when a claims challenge is specified. 

Remarks: 

this API is not always available on the client, in which case MSAL will not inject the claims. Apps wanting to monitor availability may use <see ...>

A claims challenge typically arises when calling the protected downstream API, for example when the tenant administrator wants to revokes credentials. Apps are required to look for a 401 Unauthorized response from the protected api and to parse the WWW-Authenticate response header in order to extract the claims. For more details see https://learn.microsoft.com/entra/identity/conditional-access/concept-continuous-access-evaluation

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/MtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Microsoft.Identity.Client.PlatformsCommon.Shared
+{
+    /// <summary>
+    /// A simple implementation of the HttpClient factory that uses a managed HttpClientHandler
+    /// </summary>
+    /// <remarks>
+    /// .NET should use the IHttpClientFactory, but MSAL cannot take a dependency on it.
+    /// .NET should use SocketHandler, but UseDefaultCredentials doesn't work with it 
+    /// </remarks>
+    internal class MtlsHttpClientFactory : IMsalMtlsHttpClientFactory
```
bgavrilMS, line 18: Does not seem to be used? Duplicate to `SimpleHttpClientFactory` ?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -29,13 +30,23 @@ internal class HttpManager : IHttpManager
...
+            {
+                return _httpClientFactory.GetHttpClient();
+            }
+
+            if (_httpClientFactory is IMsalMtlsHttpClientFactory msalMtlsHttpClientFactory)
+            {
+                return msalMtlsHttpClientFactory.GetHttpClient(x509Certificate2);
+            }
+
+            throw new MsalClientException("http_client_factory", "You customized HttpClient factory but you are using APIs which require provisioning of a client certificate in HttpClient. Implement IMsalMtlsHttpClientFactory");
```
bgavrilMS, line 49:
```suggestion
            throw new MsalClientException("http_client_factory", "You customized HttpClient factory but you are using APIs which require provisioning of a client certificate in HttpClient. Implement IMsalMtlsHttpClientFactory. For details see https://aka.ms/msal-net-msi-claims");
```

src/client/Microsoft.Identity.Client/Http/HttpManagerWithRetry.cs
```diff
@@ -84,9 +97,30 @@ internal class HttpManagerWithRetry : HttpManager
...
+            IDictionary<string, string> headers,
+            StringContent body,
+            X509Certificate2 bindingCert,
+            ILoggerAdapter logger,
+            CancellationToken cancellationToken = default)
+        {
+            return await SendRequestAsync(uri, headers, body, HttpMethod.Post, logger, bindingCert, doNotThrow: true, cancellationToken: cancellationToken).ConfigureAwait(false);
+        }
+
+        protected override HttpClient GetHttpClient(X509Certificate2 x509Certificate2)
```
bgavrilMS, line 111: This impl seems identical to the virtual method in the base class. Why did you override it?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -154,7 +194,7 @@ private HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string
 
                 using (logger.LogBlockDuration("[HttpManager] ExecuteAsync"))
```
bgavrilMS, line 195:
```suggestion
                using (logger.LogBlockDuration($"[HttpManager] ExecuteAsync (MTLS: {bindingCertificate!=null})"))
```

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -217,13 +258,13 @@ private HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string
 
                 Stopwatch sw = Stopwatch.StartNew();
 
-                HttpClient client = GetHttpClient();
+                HttpClient client = GetHttpClient(bindingCertificate);
```
bgavrilMS, line 261: Pls add more logging to the logging statement, i.e. if `bindignCertificate` is null or not.

src/client/Microsoft.Identity.Client/IManagedIdentityApplication.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Security.Cryptography.X509Certificates;
```
bgavrilMS, line 7: not needed?

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/MsiProxyHttpManager.cs
```diff
@@ -158,5 +159,15 @@ public MsiProxyHttpManager(string testWebServiceEndpoint)
 
             return response;
         }
+
+        public Task<HttpResponse> SendPostForceResponseAsync(Uri uri, IDictionary<string, string> headers, StringContent body, X509Certificate2 bindingCertificate, ILoggerAdapter logger, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
```
bgavrilMS, line 165: Hah, that's gonna be fun :)

tests/devapps/Managed Identity apps/ManagedIdentityAppVM/Program.cs
```diff
@@ -19,24 +24,62 @@
 
     try
     {
-        var result = await mi.AcquireTokenForManagedIdentity(scope)
-            .ExecuteAsync().ConfigureAwait(false);
+        for (int i = 1; i <= 3; i++)
```
bgavrilMS, line 27: not a fan of this, makes the code hard to read. Consider duplicating the `AcquireToken` code.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CredentialTests.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using static Microsoft.Identity.Test.Common.Core.Helpers.ManagedIdentityTestUtil;
+
+namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
+{
+    [TestClass]
+    public class CredentialTests : TestBase
+    {
+        [TestMethod]
+        public async Task CredentialEndpointBadRequestTestAsync()
```
bgavrilMS, line 23: This test fails on my machine with

Exception Message: Assert.AreEqual failed. Expected:<http://169.254.169.254/metadata/identity/credential>. Actual:<http://169.254.169.254/metadata/identity/oauth2/token>.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CredentialTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            using (var httpManager = new MockHttpManager(isManagedIdentity: true))
+            {
+                var miBuilder = ManagedIdentityApplicationBuilder.Create(ManagedIdentityId.SystemAssigned)
+                    .WithExperimentalFeatures()
+                    .WithClientCapabilities(new[] { "cp1" }) //calling this api in MI will trigger the cred flow
+                    .WithHttpManager(httpManager);
+
+                // Disabling shared cache options to avoid cross test pollution.
+                miBuilder.Config.AccessorOptions = null;
```
bgavrilMS, line 33: No need for this. The TestBase class has a helper that will reset all of MSAL's internal caches. See TestCommon.ResetInternalStaticCaches

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs
```diff
@@ -59,6 +59,25 @@ public AcquireTokenForManagedIdentityParameterBuilder WithForceRefresh(bool forc
...
+        /// <param name="claims">A string with one or multiple claims.</param>
+        /// <returns>The builder to chain .With methods.</returns>
+        public AcquireTokenForManagedIdentityParameterBuilder WithClaims(string claims)
+        {
+            if (string.IsNullOrEmpty(claims))
+            {
+                throw new ArgumentNullException(nameof(claims));
+            }
+
+            Parameters.Claims = claims;
```
bgavrilMS, line 77: You forgot to update `ManagedIdentityAuthRequest` (the one that derives from `RequestBase`) to take into account `Claims`. So it does not bypass teh token cache.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -118,13 +118,33 @@ public string ClientVersion
...
         public bool RetryOnServerErrors { get; set; } = true;
 
-        public ManagedIdentityId ManagedIdentityId { get; internal set; }
+        public Func<AppTokenProviderParameters, Task<AppTokenProviderResult>> AppTokenProvider;
 
+        #region ManagedIdentity
+        public ManagedIdentityId ManagedIdentityId { get; internal set; }
         public bool IsManagedIdentity { get; }
+#if TRA
+        public Credential.KeyMaterialInfo KeyMaterialInfo { get; internal set; }
```
bgavrilMS, line 127: I don't think this is the right place for this. ApplicationConfiguration is for whatever the user configures. KeyMaterialInfo is not something we expose. This is only used in `CredentialSource.cs` (for READ) and in the builder (for WRITE) so I would argue it should all go to `CredentialSource`.

Single Responsability Principle.

src/client/Microsoft.Identity.Client/Credential/KeyMaterialInfo.cs
```diff
@@ -0,0 +1,121 @@
...
+        internal const string KeyProviderName = "Microsoft Software Key Storage Provider";
+        internal const string KeyName = "ManagedIdentityCredentialKey";
+        private readonly ECDsaCng _eCDsaCngKey;
+        private readonly RSA _rsaKey;
+
+        public KeyMaterialInfo(bool clientCapabilitiesRequested)
+        {
+            _eCDsaCngKey = GetMachineKey(KeyProviderName, KeyName);
+            
+            if (clientCapabilitiesRequested && _eCDsaCngKey == null)
```
bgavrilMS, line 32: Design question: Are we dropping this logic? i.e. if key is not provisioned, then bail.

src/client/Microsoft.Identity.Client/Credential/KeyMaterialInfo.cs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                _rsaKey = CreateRsaKey();
+            }
+        }
+
+        public static bool IsPoPSupported => _isPopSupported;
```
bgavrilMS, line 38: SRP again. 

The "KeyMaterialInfo" class deals with keys and so its properties names should be in accordance with the key domain. Things like "IsHardwareKey", "IsSoftwareKey" etc. 

Some other class should be responsible for translating key capabilities into MSAL capabiltiies.

src/client/Microsoft.Identity.Client/Credential/KeyMaterialInfo.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace Microsoft.Identity.Client.Credential
+{
+    /// <summary>
+    /// Class to store crypto key information for a Managed Identity supported Azure resource.
+    /// For more details see https://aka.ms/msal-net-managed-identity
+    /// </summary>
+    internal class KeyMaterialInfo
```
bgavrilMS, line 17: Consider injecting the logger into this class to be able to get details about the key.

src/client/Microsoft.Identity.Client/Credential/KeyMaterialInfo.cs
gladjohn: yes

src/client/Microsoft.Identity.Client/Credential/KeyMaterialInfo.cs
```diff
@@ -0,0 +1,121 @@
...
+        private readonly ECDsaCng _eCDsaCngKey;
+        private readonly RSA _rsaKey;
+
+        public KeyMaterialInfo(bool clientCapabilitiesRequested)
+        {
+            _eCDsaCngKey = GetMachineKey(KeyProviderName, KeyName);
+            
+            if (clientCapabilitiesRequested && _eCDsaCngKey == null)
+            {
+                _rsaKey = CreateRsaKey();
```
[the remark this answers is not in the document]
gladjohn: this was to give CAE team a temp app, removed it

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
gladjohn: removed but will discuss if the new approach (PlatformProxy) is the correct approach

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -107,24 +110,26 @@ public string ClientVersion
...
-
         public bool RetryOnServerErrors { get; set; } = true;
 
-        public ManagedIdentityId ManagedIdentityId { get; internal set; }
+        public Func<AppTokenProviderParameters, Task<AppTokenProviderResult>> AppTokenProvider;
 
+        #region ManagedIdentity
+        public ManagedIdentityId ManagedIdentityId { get; internal set; }
         public bool IsManagedIdentity { get; }
+        public bool IsManagedIdentityTokenRequestInfoAvailabe { get; internal set; }
```
chlowell, line 126:
```suggestion
        public bool IsManagedIdentityTokenRequestInfoAvailable { get; internal set; }
```

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </remarks>
+    public interface IMsalMtlsHttpClientFactory : IMsalHttpClientFactory
+    {
+        /// <summary>
+        /// Method returning an HTTP client that will be used to
+        /// communicate with Azure AD over MTLS. This enables advanced scenarios.
+        /// See https://aka.ms/msal-net-application-configuration.
+        /// </summary>
+        /// <returns>An HTTP client with an ceritificate collection on the handler.</returns>
+        HttpClient GetHttpClient(X509Certificate2 x509Certificate2);
```
christothes, line 26: It would be great if this PR contained a sample of how this API would be used in combination with the `ManagedIdentityBindingCertificate` property.

gladjohn: Sure @christothes I can extend the current dev app that is part of the PR to showcase that

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
```diff
@@ -109,6 +110,16 @@ public async Task<HttpResponse> SendGetForceResponseAsync(Uri endpoint, IDiction
...
         }
+
+        public Task<HttpResponse> SendPostForceResponseAsync(Uri uri, IDictionary<string, string> headers, StringContent body, X509Certificate2 bindingCertificate, ILoggerAdapter logger, CancellationToken cancellationToken = default)
+        {
+            throw new NotImplementedException();
+        }
+
+        public async Task<HttpResponse> SendPostForceResponseAsync(Uri uri, IDictionary<string, string> headers, IDictionary<string, string> bodyParameters, X509Certificate2 bindingCertificate, ILoggerAdapter logger, CancellationToken cancellationToken = default)
+        {
+            return await _httpManager.SendPostForceResponseAsync(uri, headers, bodyParameters, logger, cancellationToken).ConfigureAwait(false);
```
gladjohn, line 121: need to add logic for passing the bindingCertificate

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CredentialTests.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        internal const string CredentialEndpoint = "http://169.254.169.254/metadata/identity/credential";
+        internal const string MtlsEndpoint = "https://centraluseuap.mtlsauth.microsoft.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/v2.0/token";
+        internal const string AzureResource = "http://vault.azure.net";
+        
+        [TestMethod]
+        public async Task CredentialHappyPathAsync()
+        {
+            using (new EnvVariableContext())
+            using (new SoftwareKeyProvider())
```
gladjohn, line 30: test runs on local when just the single test is run, but fails when run in the suite.

build/platform_and_feature_flags.props
```diff
@@ -26,4 +26,7 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <DefineConstants>$(DefineConstants);NETSTANDARD;SUPPORTS_CONFIDENTIAL_CLIENT;SUPPORTS_BROKER;SUPPORTS_CUSTOM_CACHE;SUPPORTS_WIN32;</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet6)' or '$(TargetFramework)' == '$(TargetFrameworkNet6Win)'">
+    <DefineConstants>$(DefineConstants);SUPPORTS_MANAGED_IDENTITY_V2;</DefineConstants>
```
gladjohn, line 30: Will need to add net472 target and use this Constant for target specifics

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
```diff
@@ -32,6 +32,14 @@ internal class ManagedIdentityResponse
         [JsonProperty("expires_on")]
         public string ExpiresOn { get; set; }
 
+        /// <summary>
+        /// The timespan when the access token expires.
+        /// </summary>
+        /// <remarks>The date is represented as the number of seconds from "1970-01-01T0:0:0Z UTC" 
+        /// (corresponds to the token's exp claim).</remarks>
+        [JsonProperty("expires_in")]
```
gladjohn, line 40: SLC team has a bug opened for this to expose Expires_On

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
bgavrilMS: Or a unit test should work ? We just mock the HTTP responses.

tests/Microsoft.Identity.Test.Unit/Helpers/SoftwareKeyProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                        Provider = new CngProvider(providerName),
+                        ExportPolicy = CngExportPolicies.AllowPlaintextExport,
+                        KeyCreationOptions = CngKeyCreationOptions.None | CngKeyCreationOptions.OverwriteExistingKey,
+                    };
+
+                    // Create the key
+                    _createdKey = CngKey.Create(CngAlgorithm.ECDsaP256, keyName, keyParams);
+
+                    // Add a 10-second delay
+                    Task.Delay(10000).Wait();
```
bgavrilMS, line 46: This is not great. Can you poll instead until the key is created?

tests/Microsoft.Identity.Test.Unit/Helpers/SoftwareKeyProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+
+                    Debug.WriteLine("Key created successfully!");
+                }
+            }
+            catch (CryptographicException ex)
+            {
+                Debug.WriteLine($"Error creating key: {ex.Message}");
```
bgavrilMS, line 53: Since this is a test helper, it should assert fail or throw an exception?

src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialSource.cs
```diff
@@ -0,0 +1,185 @@
...
+            return await base.HandleResponseAsync(parameters, response, cancellationToken).ConfigureAwait(false);
+        }
+
+        private CredentialResponse GetCredentialResponse(HttpResponse response)
+        {
+            CredentialResponse credentialResponse = JsonHelper.DeserializeFromJson<CredentialResponse>(response?.Body);
+
+            if (credentialResponse == null || credentialResponse.Credential.IsNullOrEmpty())
+            {
+                _requestContext.Logger.Error("[Managed Identity] Response is either null or insufficient for authentication.");
```
bgavrilMS, line 156: If it's null, state that it is null. "Insufficient for appliation" doesn't make sense - can you be more specific?

src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialSource.cs
```diff
@@ -0,0 +1,185 @@
...
+                    MsalError.ManagedIdentityRequestFailed,
+                    MsalErrorMessage.ManagedIdentityInvalidResponse,
+                    _sourceType);
+            }
+
+            return credentialResponse;
+        }
+
+        private static string CreateCredentialPayload(X509Certificate2 x509Certificate2)
+        {
```
bgavrilMS, line 167: HOw expensive is this? Please add more logging and use the `LogBlockDuration` to have some stats.

tests/Microsoft.Identity.Test.Unit/Helpers/SoftwareKeyProvider.cs
gladjohn: this is not needed actually, left it in there while debugging the static issue, will remove

src/client/Microsoft.Identity.Client/LegacyManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -43,39 +42,63 @@ protected AbstractManagedIdentity(RequestContext requestContext, ManagedIdentity
...
+                            logger: _requestContext.Logger,
+                            doNotThrow: true,
+                            retry: true,
+                            mtlsCertificate: null,
+                            cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    
+                    Debug.WriteLine("_sourceType in AuthenticateAsync" + _sourceType);
```
gladjohn, line 73: need to remove

bgavrilMS: Whatever, the compiler will remove it from the final product....

src/client/Microsoft.Identity.Client/Credential/TokenRequestAssertionInfo.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            _certificateCache = CertificateCache.Instance();
+
+            _bindingCertificate = _certificateCache.GetOrAddCertificate(
+                () => CreateCertificateFromCryptoKeyInfo(keyMaterialManager));
+        }
+
+        public X509Certificate2 BindingCertificate => _bindingCertificate;
+
+        private X509Certificate2 CreateCertificateFromCryptoKeyInfo(IKeyMaterialManager keyMaterialManager)
```
bgavrilMS, line 40: GetOrCreate

src/client/Microsoft.Identity.Client/Credential/TokenRequestAssertionInfo.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            lock (_keyInfoLock) // Lock to ensure thread safety
```
bgavrilMS, line 42: Consider using the pattern below, which avoids locking in most cases.

`if (_bindingCertificate == null)
       lock
            if (_bindingCertificate == null)
                  _bindingCertificate  = create()`

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// If your app uses Integrated Windows Authentication, ensure <see cref="HttpClientHandler.UseDefaultCredentials"/> is set to true.
+    /// </remarks>
+    public interface IMsalMtlsHttpClientFactory : IMsalHttpClientFactory
+    {
+        /// <summary>
+        /// Method returning an HTTP client that will be used to
+        /// communicate with Azure AD over MTLS. This enables advanced scenarios.
+        /// See https://aka.ms/msal-net-application-configuration.
+        /// </summary>
+        /// <returns>An HTTP client with an ceritificate collection on the handler.</returns>
```
gladjohn, line 25:
```suggestion
        /// <returns>An HTTP client with an certificate collection on the handler.</returns>
```

bgavrilMS: I recommend you make all the changes directly, don't add PR comments.

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NullKeyMaterialManager.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+using Microsoft.Identity.Client.PlatformsCommon.Shared;
+
+namespace Microsoft.Identity.Client.Platforms.net45
+{
+    /// <summary>
+    /// Class to store crypto key information for a Managed Identity supported Azure resource.
+    /// For more details see https://aka.ms/msal-net-managed-identity
+    /// </summary>
+    internal class NullKeyMaterialManager : IKeyMaterialManager
```
bgavrilMS, line 19: No need to copy this, put in `PlatformsCommon\Shared`

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IKeyMaterialManager.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        CryptoKeyType CryptoKeyType { get; }
+
+        X509Certificate2 BindingCertificate { get; }
+
+        bool IsBindingCertificateExpired();
+        TimeSpan GetTimeUntilCertificateExpiration();
+
+        bool IsKeyGuardProtected(); // Check if the key is KeyGuard protected
+        bool CertificateHasPrivateKey(); // Check if the binding certificate has a private key
```
bgavrilMS, line 20: It does not seem to be used anywhere?

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IKeyMaterialManager.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal interface IKeyMaterialManager
+    {
+        CryptoKeyType CryptoKeyType { get; }
+
+        X509Certificate2 BindingCertificate { get; }
+
+        bool IsBindingCertificateExpired();
+        TimeSpan GetTimeUntilCertificateExpiration();
+
+        bool IsKeyGuardProtected(); // Check if the key is KeyGuard protected
```
bgavrilMS, line 19: Isn't KeyType sufficient?

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IKeyMaterialManager.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client.PlatformsCommon.Shared;
+
+namespace Microsoft.Identity.Client.PlatformsCommon.Interfaces
+{
+    internal interface IKeyMaterialManager
```
bgavrilMS, line 10: Please add comments on what this class is supposed to own. Does it own certificate rotation?

src/client/Microsoft.Identity.Client/Credential/CertificateCache.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        private static readonly Lazy<CertificateCache> s_instance = new Lazy<CertificateCache>(() => new CertificateCache());
+        private X509Certificate2 _cachedCertificate;
+        private readonly object _certificateLock = new();
+
+        private CertificateCache()
+        {
+        }
+
+        public static CertificateCache Instance() => s_instance.Value;
```
bgavrilMS, line 23: Generally avoid singletons. Might as well be all static. 
If a class needs to be mocked, use interfaces.

src/client/Microsoft.Identity.Client/Platforms/net6/KeyMaterialManager.cs
```diff
@@ -0,0 +1,270 @@
...
+        private readonly object _keyInfoLock = new(); // Lock object
+        private readonly ILoggerAdapter _logger;
+
+        public KeyMaterialManager(ILoggerAdapter logger)
+        {
+            _logger = logger;
+
+            _certificateCache = CertificateCache.Instance();
+            
+            _bindingCertificate = _certificateCache.GetOrAddCertificate(
```
bgavrilMS, line 37: I don't think this is correct. The lifetime of this object is the same as the lifetime of the ManagedIdentityApplication, which can be the same as the lifetime of the process. 

I think you should check if the cert needs to be rotated on every request. So instead of having a property `BidingCertificate` you have a method `GetOrCreateBindingCertificate`  which does the rotation?

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IKeyMaterialManager.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        CryptoKeyType CryptoKeyType { get; }
+
+        X509Certificate2 BindingCertificate { get; }
+
+        bool IsBindingCertificateExpired();
```
bgavrilMS, line 16: This interface leaks too much data. Why isn't the `BindingCertificate` enough?

src/client/Microsoft.Identity.Client/Platforms/net6/KeyMaterialManager.cs
bgavrilMS: (this will also eliminate the need to leak cert expiration outside this class)

src/client/Microsoft.Identity.Client/Credential/CertificateCache.cs
gladjohn: deleted the cert cache

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CredentialTests.cs
gladjohn: will explore but did not work when I remove this

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -107,24 +107,26 @@ public string ClientVersion
...
-
         public bool RetryOnServerErrors { get; set; } = true;
 
-        public ManagedIdentityId ManagedIdentityId { get; internal set; }
+        public Func<AppTokenProviderParameters, Task<AppTokenProviderResult>> AppTokenProvider;
 
+        #region ManagedIdentity
+        public ManagedIdentityId ManagedIdentityId { get; internal set; }
         public bool IsManagedIdentity { get; }
+        public bool IsManagedIdentityTokenRequestInfoAvailable { get; internal set; }
```
bgavrilMS, line 123: These are not part of the configuration. Please remove them. `IsManagedIdentityTokenRequestInfoAvailable` is not even used and  `IsManagedIdentityTokenRequestInfoAvailable` is used once and can easily be avoided.

src/client/Microsoft.Identity.Client/Credential/IManagedIdentityCredentialResponseCache.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.ManagedIdentity;
+
+namespace Microsoft.Identity.Client.Credential
+{
+    internal interface IManagedIdentityCredentialResponseCache
```
bgavrilMS, line 11: We should not cache responses. Only certs.

src/client/Microsoft.Identity.Client/Internal/Requests/CredentialBasedMsiAuthRequest.cs
```diff
@@ -0,0 +1,300 @@
...
+using Microsoft.Identity.Client.Cache.Items;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Credential;
+using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.Internal.Requests
+{
+    internal class CredentialBasedMsiAuthRequest : RequestBase
```
bgavrilMS, line 19: I think we need to avoid having another request. There is just too much dup code. 

We can either reuse the ClientCredentials request, which we decided is not easily possible because we can't use CCA here. Or go back to your original design which used "LegacyMsiAuthRequest".

bgavrilMS: CC @pmaytak for an opinion on this.

src/client/Microsoft.Identity.Client/PlatformsCommon/Interfaces/IKeyMaterialManager.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client.PlatformsCommon.Shared;
+
+namespace Microsoft.Identity.Client.PlatformsCommon.Interfaces
+{
+    internal interface IKeyMaterialManager
```
bgavrilMS, line 10: Since this interface is strongly tied to MSI, I would rename it, maybe `SlcKeyAndCertManager` or smth.

src/client/Microsoft.Identity.Client/Internal/Requests/CredentialBasedMsiAuthRequest.cs
pmaytak: Create a base managed identity request class with common code, legacy and this class can extend it? Can also aid in readability to show what's specific to each MSI request implementation.

src/client/Microsoft.Identity.Client/Credential/ManagedIdentityCredentialResponseCache.cs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        /// <summary>
+        /// Creates an OAuth2 client request for fetching the managed identity credential.
+        /// </summary>
+        /// <param name="httpManager"></param>
+        /// <returns></returns>
+        private OAuth2Client CreateClientRequest(IHttpManager httpManager)
+        {
+            var client = new OAuth2Client(_requestContext.Logger, httpManager, null);
```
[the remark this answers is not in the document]
bgavrilMS: Have a look at the MSI tests, see the type of exceptions they use.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
gladjohn: tests added for [httpclient customizations](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/edc4d7236d4e01315738e2710c6ad5d00db98bc7/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CredentialHttpManagerTests.cs)

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ManagedIdentityExecutor.cs
```diff
@@ -35,19 +38,44 @@ public ManagedIdentityExecutor(IServiceBundle serviceBundle, ManagedIdentityAppl
...
+                    !managedIdentityParameters.Resource.EndsWith("/.default", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Add "/.default" to the scopes
+                    commonParameters.Scopes = new SortedSet<string>
+                    {
+                        managedIdentityParameters.Resource + "/.default"
+                    };
+                }
+
+                requestContext.Logger.Info(LogMessages.CredentialScopeUpdated);
```
bgavrilMS, line 55: Generally we preferer not have have constants when logging. Makes the code easier to read.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -107,24 +107,27 @@ public string ClientVersion
...
 
-        public ManagedIdentityId ManagedIdentityId { get; internal set; }
+        public Func<AppTokenProviderParameters, Task<AppTokenProviderResult>> AppTokenProvider;
 
+        #region ManagedIdentity
+        public ManagedIdentityId ManagedIdentityId { get; internal set; }
         public bool IsManagedIdentity { get; }
 
-        public Func<AppTokenProviderParameters, Task<AppTokenProviderResult>> AppTokenProvider;
+        public CryptoKeyType ManagedIdentityCredentialKeyType { get; internal set; }
```
bgavrilMS, line 124: Not sure if this is the right place for these, since app developer can't configure them.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Factory responsible for creating HttpClient with a cert handler for MTLS. 
+    /// See https://learn.microsoft.com/dotnet/api/system.net.http.httpclient?view=net-7.0#instancing for more details.
+    /// </summary>
+    /// <remarks>
+    /// Implementations must be thread safe. 
+    /// Do not create a new HttpClient for each call to <see cref="GetHttpClient(X509Certificate2)"/> - this leads to socket exhaustion.
+    /// If your app uses Integrated Windows Authentication, ensure <see cref="HttpClientHandler.UseDefaultCredentials"/> is set to true.
+    /// </remarks>
+    public interface IMsalMtlsHttpClientFactory : IMsalHttpClientFactory
```
bgavrilMS, line 18: I believe in the end we decided with @christothes and Vinay to leave the current extensibility mechanism as is. App developers who need to extens HttpClient will have to call the "GetMtlsCertificate" upfront.

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
```diff
@@ -100,6 +99,27 @@ public ManagedIdentityApplicationBuilder WithTelemetryClient(params ITelemetryCl
...
+        /// <remarks>
+        /// MSAL will transform these into special claims request. See https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter for
+        /// details on claim requests. This is an experimental API. The method signature may change in the future without involving a major version upgrade.
+        /// For more details see https://aka.ms/msal-net-claims-request
+        /// </remarks>
+        public ManagedIdentityApplicationBuilder WithClientCapabilities(IEnumerable<string> clientCapabilities)
+        {
+            ValidateUseOfExperimentalFeature("WithClientCapabilities");
+
+            if (clientCapabilities != null && clientCapabilities.Any())
```
bgavrilMS, line 115:
```suggestion
            if (clientCapabilities?.Any())
```

src/client/Microsoft.Identity.Client/Platforms/Features/SLC/KeyMaterialManager.cs
```diff
@@ -0,0 +1,342 @@
...
+    /// <summary>
+    /// Provides X509_2 certificates and cryptographic key information for a Credential based 
+    /// Managed Identity-supported Azure resource. 
+    /// This class handles the retrieval or creation of X.509_2 certificates for authentication purposes,
+    /// including the determination of the cryptographic key type.
+    /// For more details, see https://aka.ms/msal-net-managed-identity.
+    /// </summary>
+    internal class ManagedIdentityCertificateProvider : IKeyMaterialManager
+    {
+        // Property to hold the name of the key guard isolation property
```
bgavrilMS, line 21:
```suggestion
        // The name of the key guard isolation property
```

src/client/Microsoft.Identity.Client/Platforms/Features/SLC/KeyMaterialManager.cs
```diff
@@ -0,0 +1,342 @@
...
+        private readonly object _keyInfoLock = new();
+
+        // Logger instance for capturing log information
+        private readonly ILoggerAdapter _logger;
+
+        // Property to get or create the binding certificate from crypto key information
+        public X509Certificate2 BindingCertificate => GetOrCreateCertificateFromCryptoKeyInfo();
+
+        // Property to expose the current crypto key type
+        public CryptoKeyType CryptoKeyType => _cryptoKeyType;
```
bgavrilMS, line 45: No need for a backing field ` _cryptoKeyType`, just use auto property `CryptoKeyType { get; private set;}`

bgavrilMS, line 45: This property is not initialized properly. MSAL dev has to know to call `GetOrCreateCertificateFromCryptoKeyInfo` before calling this property. 

Solutions: 

1. Initialize everything on ctor (best / cleanest way, but you have to be careful about lazy loading, i.e. if customer uses CCA and not MIA, then you mustn't call this ctor).
2. Maintain a private "_isInitialized" field.
a. Option1: throw ex if app dev calls GetKeyType and `_isInaitialized=false`
b. Option2: call `GetOrCreateCertificateFromCryptoKeyInfo` if `_isInaitialized=false`

src/client/Microsoft.Identity.Client/Platforms/Features/SLC/KeyMaterialManager.cs
```diff
@@ -0,0 +1,342 @@
...
+            _logger = logger;
+        }
+
+        /// <summary>
+        /// Retrieves or creates an X.509 certificate from crypto key information.
+        /// </summary>
+        /// <returns>
+        /// The X.509 certificate if available and still valid in the cache; otherwise, a new certificate is created.
+        /// </returns>
+        public X509Certificate2 GetOrCreateCertificateFromCryptoKeyInfo()
```
bgavrilMS, line 58: There is no need for this to be public (i.e. internal). Please review all "internal" API and try to make it private.

src/client/Microsoft.Identity.Client/Platforms/Features/SLC/KeyMaterialManager.cs
```diff
@@ -0,0 +1,342 @@
...
+
+        // Field to store the current crypto key type
+        private CryptoKeyType _cryptoKeyType;
+
+        // Constants specifying the names for the key storage provider and key names
+        private const string KeyProviderName = "Microsoft Software Key Storage Provider";
+        private const string MachineKeyName = "ManagedIdentityCredentialKey";
+        private const string SoftwareKeyName = "ResourceBindingKey";
+
+        // Static field to cache the binding certificate across instances
```
bgavrilMS, line 32:
```suggestion
        // Cache the binding certificate across instances 
```

MSAL dev can see that this is a static field, no need to document the obvious.

src/client/Microsoft.Identity.Client/Platforms/Features/SLC/KeyMaterialManager.cs
```diff
@@ -0,0 +1,342 @@
...
+
+        /// <summary>
+        /// Determines if a given X.509 certificate needs rotation based on a percentage threshold.
+        /// </summary>
+        /// <param name="certificate">The X.509 certificate to evaluate.</param>
+        /// <param name="rotationPercentageThreshold">The threshold percentage for considering rotation (default is 70%).</param>
+        /// <returns>
+        /// True if the certificate needs rotation, false otherwise.
+        /// </returns>
+        public static bool CertificateNeedsRotation(X509Certificate2 certificate, double rotationPercentageThreshold = 70)
```
bgavrilMS, line 95: private

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ManagedIdentityExecutor.cs
```diff
@@ -35,19 +38,48 @@ public ManagedIdentityExecutor(IServiceBundle serviceBundle, ManagedIdentityAppl
...
+                bool endsWithDefault = miResource.EndsWith("/.default", StringComparison.OrdinalIgnoreCase);
+
+                // Add "/.default" only if it doesn't end with "/.default" and doesn't contain "/.somethingelse"
+                if (!endsWithDefault)
+                {
+                    // Add "/.default" to the scopes
+                    commonParameters.Scopes = new SortedSet<string>
+                    {
+                        managedIdentityParameters.Resource + "/.default"
+                    };
```
neha-bhargava, line 56: You can add this as a method to check and add ./default suffix in class [ScopeHelper.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Utils/ScopeHelper.cs).

neha-bhargava: Is this a requirement for the SLC endpoint to have the ./default suffix?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava, line 124: Why do you need these 2 variables here instead of single object for KeyMaterialManager which have both the cert and CryptoKeyType?

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ManagedIdentityExecutor.cs
gladjohn: yes, since we are calling ests directly

PR #4423, opened 2023-11-16
Cache level should be L1 in case of default in memory cache.

### LINKED ISSUE #4414: [Bug] Cache level should be L1 when MSAL's in memory cache is used.
### Library version used

4.57.0

### .NET version

Net 6

### Scenario

None

Cache level when MSAL's cache is used is Unknown, should be L1.

image

As captured from https://portal.aria.ms/kusto-explorer?projectId=4d8bb0b77fd041b4b12cb041d0d84ff6&isEU=false

### Relevant code snippets

_No response_

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

neha-bhargava:
Fixes #4414 

Fix the cache level, when in memory cache is used cache level will be L1Cache.

**Testing**
Updated tests to verify

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -177,6 +177,9 @@ namespace Microsoft.Identity.Client.Cache
                         RequestContext.ApiEvent.CacheLevel = telemetryData.CacheLevel;
                     }
                 }
+            } else
+            {
+                RequestContext.ApiEvent.CacheLevel = CacheLevel.L1Cache;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


gladjohn:
looks good @neha-bhargava. In the RequestBase there is also a check to see if cache level is set if not we set to None. But, is no test coverage for that scenario I guess. Maybe you can file a work item to track that.

gladjohn (approved):

neha-bhargava:
> looks good @neha-bhargava. In the RequestBase there is also a check to see if cache level is set if not we set to None. But, is no test coverage for that scenario I guess. Maybe you can file a work item to track that.

We do have tests to cover all the cache levels here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/nebharg/CacheLevelBug/tests/Microsoft.Identity.Test.Unit/PublicApiTests/TelemetryClientTests.cs#L245. So it is all covered.

jmprieur (approved):
LGTM
Thanks @neha-bhargava

PR #4443, opened 2023-11-30
Updating error messages for non RSA certificates messages

### LINKED ISSUE #4407: [Bug] Error messages for using non-RSA certificates for certificate based auth could be better
Currently, if a non-RSA certificate is used for certificate auth, the resulting error is a null reference exception (see below). Ideally, [SignWithCertificate](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4b0b6cdc73e079cc6a73c20f3ab01058da29e2b0/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/CommonCryptographyManager.cs#L62-L63) would check whether the PublicKey.Oid is RSA and fail with a more descriptive exception.

```
NullReferenceException: Object reference not set to an instance of an object.

This exception was originally thrown at this call stack:
    Microsoft.Identity.Client.PlatformsCommon.Shared.CommonCryptographyManager.SignWithCertificate(string, System.Security.Cryptography.X509Certificates.X509Certificate2)
    Microsoft.Identity.Client.Internal.JsonWebToken.Sign(System.Security.Cryptography.X509Certificates.X509Certificate2, string, bool)
    Microsoft.Identity.Client.Internal.ClientCredential.CertificateAndClaimsClientCredential.AddConfidentialClientParametersAsync(Microsoft.Identity.Client.OAuth2.OAuth2Client, Microsoft.Identity.Client.Core.ILoggerAdapter, Microsoft.Identity.Client.PlatformsCommon.Interfaces.ICryptographyManager, string, string, bool, System.Threading.CancellationToken)
    Microsoft.Identity.Client.OAuth2.TokenClient.AddBodyParamsAndHeadersAsync(System.Collections.Generic.IDictionary<string, string>, string, System.Threading.CancellationToken)
    Microsoft.Identity.Client.OAuth2.TokenClient.SendTokenRequestAsync(System.Collections.Generic.IDictionary<string, string>, string, string, System.Threading.CancellationToken)
    Microsoft.Identity.Client.Internal.Requests.RequestBase.SendTokenRequestAsync(System.Collections.Generic.IDictionary<string, string>, System.Threading.CancellationToken)
    Microsoft.Identity.Client.Internal.Requests.ClientCredentialRequest.GetAccessTokenAsync(System.Threading.CancellationToken, Microsoft.Identity.Client.Core.ILoggerAdapter)
    Microsoft.Identity.Client.Internal.Requests.ClientCredentialRequest.ExecuteAsync(System.Threading.CancellationToken)
    Microsoft.Identity.Client.Internal.Requests.RequestBase.RunAsync(System.Threading.CancellationToken)
    Microsoft.Identity.Client.ApiConfig.Executors.ConfidentialClientExecutor.ExecuteAsync(Microsoft.Identity.Client.ApiConfig.Parameters.AcquireTokenCommonParameters, Microsoft.Identity.Client.ApiConfig.Parameters.AcquireTokenForClientParameters, System.Threading.CancellationToken)
```

related to https://github.com/Azure/azure-sdk-for-net/issues/39711

trwalke:
Fixes #4407

Updating error messages for non RSA certificates messages

**Testing**
Added unit test

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/CommonCryptographyManager.cs
```diff
@@ -62,9 +62,14 @@ public byte[] CreateSha256HashBytes(string input)
         /// <remarks>AAD only supports RSA certs for client credentials </remarks>
         public virtual byte[] SignWithCertificate(string message, X509Certificate2 certificate)
         {
+            //Ensure certificate is of type RSA.
+            if (!certificate.PublicKey.Oid.FriendlyName.Equals(Constants.RsaFriendlyName))
```
trwalke, line 66: not sure if `Oid.FriendlyName` is the best way to do this. looking into if there is abetter way to check for RSA

tests/Microsoft.Identity.Test.Unit/CryptographyTests.cs
```diff
@@ -34,6 +38,42 @@ public void SignWithCertificate()
...
+            var serviceBundle = TestCommon.CreateDefaultServiceBundle();
+
+            X509Certificate2 cert = GenerateSelfSignedCertificate();
+            var crypto = serviceBundle.PlatformProxy.CryptographyManager;
+
+            MsalClientException ex = AssertException.Throws<MsalClientException>(() =>
+            {
+                crypto.SignWithCertificate("TEST", cert);
+            });
+
```
bgavrilMS, line 56: nit: assert that the test cert type is ECD

tests/Microsoft.Identity.Test.Unit/CryptographyTests.cs
```diff
@@ -34,6 +38,42 @@ public void SignWithCertificate()
...
+            Assert.AreEqual(ex.ErrorCode, MsalError.CertificateNotRsa);
+            Assert.AreEqual(ex.Message, MsalErrorMessage.CertMustBeRsa);
+        }
+
+        public static X509Certificate2 GenerateSelfSignedCertificate()
```
bgavrilMS, line 61: nit: There is already a CertificateHelper in MSAL, can you use that?

bgavrilMS (changes requested):
Improve error messages, account for missing Key as well.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/CommonCryptographyManager.cs
```diff
@@ -73,6 +72,12 @@ public virtual byte[] SignWithCertificate(string message, X509Certificate2 certi
                 rsa = certificate.GetRSAPrivateKey();
             }
 
+            //Ensure certificate is of type RSA.
+            if (rsa == null)
```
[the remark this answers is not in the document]
trwalke: Oh yes, I missed the fact that the key can be missing. I used the `Oid.FriendlyName` in my previous commit but then I read somewhere on stack overflow that not all certificates will have the `FriendlyName` field populated so I changed the approach. I will do some more digging to see if we can rely on the `FriendlyName`.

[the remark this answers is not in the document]
pmaytak: Yea, would probably be better to check on a type, if possible, rather than a string FriendlyName?

tests/Microsoft.Identity.Test.Unit/CryptographyTests.cs
```diff
@@ -34,6 +38,42 @@ public void SignWithCertificate()
             Assert.IsNotNull(value);
             Assert.AreEqual("MrknKHbOAVu2iuLHMFSk2SK773H1ysxaAjAPcTXYSfH4P2fUfvzP6aIb9MkBknjoE_aBYtTnQ7jOAvyQETvogdeSH7pRDPhCk2aX_8VIQw0bjo_zBZj5yJYVWQDLIu8XvbuzIGEvVaXKz4jJ1nYM6toun4tM74rEHvwa0ferafmqHWOd5puPhlKH1VVK2RPuNOoKNLWBprVBaAQVJVFOdRcd3iR0INBHykxtOsG0pgo0Q2uQBlKP7KQb7Ox8i_sw-M21BuUzdIdGs_oeUYh0B8s-eIGf34JmHRWMwWCnRWzZgY9YuIjRoaWNqlWYb8ASjKOxzvk99x8eFEYKOjgAcA", value);
         }
+
+        [TestMethod]
+        [TestCategory("CryptographyTests")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Internal.Analyzers", "IA5352:DoNotMisuseCryptographicApi", Justification = "Suppressing RoslynAnalyzers: Rule: IA5352 - Do Not Misuse Cryptographic APIs in test only code")]
+        public void SignWithNonRsaCertificate()
```
pmaytak, line 45: For test naming, good practice is to add what the expected result is, so it's easier to see at a glance.
```suggestion
        public void SignWithNonRsaCertificate_ThrowsException()
```

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/CommonCryptographyManager.cs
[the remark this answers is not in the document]
trwalke: So, I did some more digging on this and it looks like we cannot guarantee that the `FriendlyName` field will be populated or even contain the word `RSA`. Here is a list of the known friendly names for .net core. As you can see, not all of them have `RSA` but it is not clear if those configs still use RSA.
https://github.com/dotnet/corefx/blob/b8b81a66738bb10ef0790023598396861d92b2c4/src/System.Security.Cryptography.Encoding/src/Internal/Cryptography/OidLookup.cs#L113


Also, [we already have a check for a certificate without a private key](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/0eecb41515a816eb3860f656aae575042935fc1e/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L121) so this code will never have to deal with that scenario. We can rely upon the fact `GetRSAPrivateKey` will return null if the key is not of type RSA.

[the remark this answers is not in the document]
bgavrilMS: Discussed that it's not practical to check the friendly name. However, if RSAKey == null, we can put the OID Friendly Name in the error message,.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -341,6 +341,13 @@ public static string RedirectUriMismatch(string expectedUri, string actualUri)
 
         public const string ScopesRequired = "At least one scope needs to be requested for this authentication flow. ";
         public const string InvalidAdalCacheMultipleRTs = "The ADAL cache is invalid as it contains multiple refresh token entries for one user. Deleting invalid ADAL cache. ";
+        public const string CryptoNet45 =
```
bgavrilMS, line 344: bad rebase? This has been deleted.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
gladjohn, line 344:
```suggestion
        public const string CryptoNet45 = "Error: Certificate signing failed. Please check the inner exception for detailed information.\n" +
    "- Possible Cause: The app may be built against .NET Desktop 4.6 or lower.\n" +
    "- Suggested Actions: \n" +
    "   1. Target a higher version of .NET Desktop (4.6.2 or above).\n" +
    "   2. Use a different type of certificate (non-CNG).\n" +
    "   3. Consider signing your own assertion. For more details, visit: https://aka.ms/msal-net-signed-assertion";

```

src/client/Microsoft.Identity.Client/Internal/Constants.cs
```diff
@@ -46,6 +46,8 @@ internal static class Constants
         public const string ManagedIdentityDefaultTenant = "managed_identity";
         public const string CiamAuthorityHostSuffix = ".ciamlogin.com";
 
+        public const string RsaFriendlyName = "RSA";
```
gladjohn, line 49: is this used?

gladjohn (approved):

src/client/Microsoft.Identity.Client/Internal/Constants.cs
trwalke: not needed

pmaytak (approved):

PR #4449, opened 2023-12-04
Fix for #4444 - emit simple OS descriptor

### LINKED ISSUE #4444: [Bug] AcquireTokenSilent call is failing on Mac because of missing OperatingSystem information
### Library version used

4.54.0.0

### .NET version

.NET Core

### Scenario

PublicClient - desktop app

More information available in this ICM, but essentially the user having this issue is seeing Conditional Access policies block him from calls to AcquireTokenSilent because the Operating System (Mac) is not being sent by MSAL when making the network calls.

https://portal.microsofticm.com/imp/v3/incidents/incident/434595820/summary

### Relevant code snippets

_No response_

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

bgavrilMS:
Fixes #4444


src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -78,8 +78,32 @@ protected override string InternalGetProcessorArchitecture()
             return DesktopOsHelper.IsWindows() ? WindowsNativeMethods.GetProcessorArchitecture() : null;
         }
 
+        /// <summary>
+        /// The name of the operating system is important to the STS, as some CA policies 
+        /// will look at x-client-os; as such the name of the OS should be par-sable by the STS.
```
jbzdarkid, line 83:
```suggestion
        /// will look at x-client-os; as such the name of the OS should be parseable by the STS. 
```

[nit] I don't think there should be a hyphen but if you want one then it should be parse-able.

bgavrilMS: Thx, the VS spell check that I use suggested this and I blindly trusted it ...

bgavrilMS:
Let me know if the [new values](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4449/files#diff-4fcf5882b533ffda6e20c803013a8e994e5d781db0c376842c0300585abb4138R23) "Windows", "MacOS", "Linux" are ok, happy to change to anything is easy ESTS. 

Note that for Windows the current string is like "Microsoft Windows 10.0.16620.0" (at least .NET Core), but I would think ESTS doesn't look at the Win Version in x-client-os.

rayluo:
> Let me know if the [new values](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4449/files#diff-4fcf5882b533ffda6e20c803013a8e994e5d781db0c376842c0300585abb4138R23) "Windows", "MacOS", "Linux" are ok, happy to change to anything is easy ESTS, @jbzdarkid
> 
> Note that for Windows the current string is like "Microsoft Windows 10.0.16620.0" (at least .NET Core), but I would think ESTS doesn't look at the Win Version in x-client-os.

Do those values need to be precisely defined by eSTS? I thought they are kind of freestyle. For example, [in MSAL Python it is the output of Python's built-in `sys.platform` value](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.25.0/msal/application.py#L649), whose [values are described here](https://docs.python.org/3/library/sys.html#sys.platform). Do we have to let all MSALs use same set of values?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
```diff
@@ -20,6 +20,10 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
 {
     internal abstract class AbstractPlatformProxy : IPlatformProxy
     {
+        public const string WindowsOS = "Windows";
+        public const string MacOs = "MacOS";
+        public const string LinuxOS = "Linux";
```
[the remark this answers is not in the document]
jbzdarkid: Ah, excuse me -- we currently parse all .Net MSAL/ADAL strings with this logic:
https://msazure.visualstudio.com/One/_git/ESTS-Main?path=/src/Product/Microsoft.AzureAD.ESTS/Sts/ClientInfo.cs&version=GBmaster&line=451&lineEnd=451&lineStartColumn=17&lineEndColumn=44&lineStyle=plain&_a=contents

Again, we can change it to whatever, but we'll want to try to keep parity with existing code. E.g. it looks like we expect to be able to pull windows version out of this string with a simple regex.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
```diff
@@ -20,6 +20,10 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
 {
     internal abstract class AbstractPlatformProxy : IPlatformProxy
     {
+        public const string WindowsOS = "Windows";
```
pmaytak, line 23: Make these names a bit more specific/clear what they are for - that these are sent to STS. WindowsOsPlatformName or something, or put in a nested class? Or I guess property comment also works.

src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs
```diff
@@ -78,8 +78,32 @@ protected override string InternalGetProcessorArchitecture()
             return DesktopOsHelper.IsWindows() ? WindowsNativeMethods.GetProcessorArchitecture() : null;
         }
 
+        /// <summary>
+        /// The name of the operating system is important to the STS, as some CA policies 
+        /// will look at x-client-os; as such the name of the OS should be parseable by the STS. 
+        /// Do not use RID, as the format is not standardized across platforms.
+        /// Do not OSDescription, as it can be very long and non-standard, e.g.
```
pmaytak, line 85:
```suggestion
        /// Do not use OSDescription, as it can be very long and non-standard, e.g. 

```

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopPlatformProxy.cs
```diff
@@ -120,7 +120,10 @@ protected override string InternalGetDeviceModel()
         /// <inheritdoc/>
         protected override string InternalGetOperatingSystem()
         {
-            return DesktopOsHelper.GetWindowsVersionString();
+            return IsWindows ?
+                AbstractPlatformProxy.WindowsOS                
+                : Environment.OSVersion.Platform.ToString(); // probably for Mono, which we don't support
```
pmaytak, line 125: Why  does Mono matter here? Shouldn't NetDesktopPlatformProxy be only used on net462 now?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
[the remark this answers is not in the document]
pmaytak: @bgavrilMS Wouldn't it be also helpful for us to know the specific Windows version in our case, like to put in logs or telemetry?

pmaytak (changes requested):

tests/Microsoft.Identity.Test.Unit/CoreTests/MsalIdHelperTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            Assert.IsTrue(parameters.ContainsKey(MsalIdParameter.OS));
+            Assert.IsTrue(parameters.ContainsKey(MsalIdParameter.Product)); // sku
+            Assert.IsTrue(parameters.ContainsKey(MsalIdParameter.Version)); // version
+            Assert.IsFalse(parameters.ContainsKey(MsalIdParameter.DeviceModel)); // device - we don't send this on .NET and .NET fwk
+
+            string os = parameters[MsalIdParameter.OS];
+
+#if NET48_OR_GREATER
+            Assert.AreEqual(AbstractPlatformProxy.WindowsOS, os);
+#else
```
pmaytak, line 31: Why is this needed?

tests/Microsoft.Identity.Test.Unit/CoreTests/MsalIdHelperTests.cs
```diff
@@ -0,0 +1,51 @@
...
+                Assert.AreEqual(AbstractPlatformProxy.WindowsOS, os);
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                Assert.AreEqual(AbstractPlatformProxy.MacOs, os);
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                Assert.AreEqual(AbstractPlatformProxy.LinuxOS, os);
+            }
```
pmaytak, line 43: We run our tests on Windows, this would never execute, right?

tests/Microsoft.Identity.Test.Unit/CoreTests/MsalIdHelperTests.cs
```diff
@@ -0,0 +1,51 @@
...
+    public class MsalIdHelperTests
+    {
+        [TestMethod]        
+        public void XClientOSTest()
+        {
+            // Act
+            IDictionary<string, string> parameters = MsalIdHelper.GetMsalIdParameters(null);
+
+            // Assert
+            Assert.IsTrue(parameters.ContainsKey(MsalIdParameter.OS));
```
iulico-1, line 22: >OS [](http://example.com/codeflow?start=65&length=2)

if the goal is to synchronize with our xplat library here are the values MSAL.CPP sends today.

x-client-SKU: MSAL.xplat.Win32
x-client-Ver: 1.1.0+ebfe9dc1
x-client-src-SKU: MSAL.xplat.Win32
x-client-OS: 10.0.19041.3636

src/client/Microsoft.Identity.Client/Platforms/netdesktop/NetDesktopPlatformProxy.cs
bgavrilMS: It's just wierd that we have `IsWindows` helper on net462 tfm. But it exists and it should be used due to Mono. When I'm surprised, I add a comment.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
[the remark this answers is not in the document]
bgavrilMS: @joelklabo - I had a look at the ESTS code and it seems you try to parse x-client-os - look for Windows + version. And if that fails, you will look at the UserAgent string (which is not reliable, particularly during refresh_token flow). I see that the os-version is only used for Windows.

I will make the change back to send the OSDescription on Windows, which seems to be "Microsoft + Windows + version". And send "MacOS" and "Linux" as x-client-os. 

I am very hesitant to make changes to `UserAgent`, because Azure SDK actually hijacks that.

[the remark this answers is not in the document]
jbzdarkid: Agreed that we should not change the UserAgent. That also has impact on the UI rendering logic.
The customer issue here was entirely about the SKU OS header.

pmaytak (approved):

iulico-1 (approved):
:shipit:

Issue #4447, opened 2023-12-02
MSI CAE Claims handling - claims merging issue


gladjohn:
With SLC CAE, the claims we get back from ESTS is, 

```json
{
  "access_token": {
    "nbf": {
      "essential": true,
      "value": "1701477303"
    }
  }
}
```

And the merged claims and capab should be like this, 

```json
{
  "access_token": {
    "xms_cc": {
      "values": ["cp1", "cp2"]
    },
    "nbf": {
      "essential": true,
      "value": "1701477303"
    }
  }
}
```

In MSAL .NET, we build claims and capab with `access_token` and `xms_cc`, but with the new incoming claim, we fail to do a proper merge, instead just return the incoming claim without the capab. 

All MSAL's need to check if this is being properly handled. 

```cs
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

using System.Collections.Generic;
using System.Linq;
using Microsoft.Identity.Client.Utils;
#if SUPPORTS_SYSTEM_TEXT_JSON
using System.Text.Json;
using System.Text.Json.Nodes;
using JObject = System.Text.Json.Nodes.JsonObject;
#else
using Microsoft.Identity.Json;
using Microsoft.Identity.Json.Linq;
#endif

namespace Microsoft.Identity.Client.Internal
{
    internal static class ClaimsHelper
    {
        private const string AccessTokenClaim = "access_token";
        private const string XmsClientCapability = "xms_cc";

        internal static string GetMergedClaimsAndClientCapabilities(
            string claims,
            IEnumerable<string> clientCapabilities)
        {
            if (clientCapabilities != null && clientCapabilities.Any())
            {
                JObject capabilitiesJson = CreateClientCapabilitiesRequestJson(clientCapabilities);
                JObject mergedClaimsAndCapabilities = MergeClaimsIntoCapabilityJson(claims, capabilitiesJson);

                return JsonHelper.JsonObjectToString(mergedClaimsAndCapabilities);
            }

            return claims;
        }

        internal static JObject MergeClaimsIntoCapabilityJson(string claims, JObject capabilitiesJson)
        {
            if (!string.IsNullOrEmpty(claims))
            {
                JObject claimsJson;
                try
                {
                    claimsJson = JsonHelper.ParseIntoJsonObject(claims);
                }
                catch (JsonException ex)
                {
                    throw new MsalClientException(
[...]

PR #4452, opened 2023-12-04
Fix for claims and client capabilities JSON merge bug in net6

gladjohn:
Fixes : #4447

Identified a [bug](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4447) in MSAL's JSON merging logic affecting applications targeting .NET 6 and above. JSON objects with properties of the same name but different structures are not merged correctly, leading to data loss

```json
// JSON 1
"access_token": {
    "xms_cc": {
        "values": ["cp1", "cp2"]
    }
}

// JSON 2
"access_token": {
    "nbf": {
        "essential": true,
        "value": "1701540593"
    }
}
```
problem : After merging, xms_cc is lost, retaining only the nbf property.

fix : This fix ensures that JSON merging is performed correctly, avoiding data loss when properties have the same name but different structures. The implementation considers various value kinds (objects, arrays) and recursively merges nested structures. It addresses the identified issue and provides a reliable merging mechanism for JSON objects.

Testing : Unit tests and manual verification that we pass capabilities now in net6. 

Other notes : Unit tests also has coverage for net48 that uses newtonsoft code path, and we do not have this issue in that code path


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/ClaimsHelper.cs
```diff
@@ -52,10 +55,10 @@ internal static JObject MergeClaimsIntoCapabilityJson(string claims, JObject cap
...
                 }
 #if SUPPORTS_SYSTEM_TEXT_JSON
-                foreach (var claim in claimsJson)
-                {
-                    capabilitiesJson[claim.Key] = claim.Value != null ? JsonNode.Parse(claim.Value.ToJsonString()) : null;
-                }
+
+                var mergedJson = JsonHelper.Merge(capabilitiesJson.ToJsonString(), claimsJson.ToJsonString());
+
+                capabilitiesJson = JsonNode.Parse(mergedJson) as JObject;
```
pmaytak, line 61:
```suggestion
                capabilitiesJson = JsonHelper.ParseIntoJsonObject(claims);
```

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
```diff
@@ -155,6 +157,125 @@ internal static long ExtractParsedIntOrZero(JObject json, string key)
...
         internal static T GetValue<T>(JsonNode json) => json != null ? json.GetValue<T>() : default;
+
+        /// <summary>
+        /// Combines two JSON strings into a single JSON string with indentation.
+        /// </summary>
+        /// <param name="originalJson">The original JSON string to merge.</param>
+        /// <param name="newContent">The additional JSON content to merge.</param>
+        /// <returns>A UTF-8 encoded string representing the merged JSON.</returns>
+        /// Code reference : https://github.com/dotnet/runtime/issues/31433
+        internal static string Merge(string originalJson, string newContent)
```
pmaytak, line 168: My first thought was that this creates quite a few allocations. We start with 2 JsonObjects, which are converted to string, then parsed to JsonDocument, then outputBuffer is allocated, which at the end is written to string, and then parsed to JsonObject, then converted to string again. So maybe this can be made more efficient by updating one of the JsonObject s in place or at the very least updating the parameters and return values of these methods so we don't unnecessarily convert to string and back.

Now, this isn't a hot path and the JSON is expected to be small, so not that big of a deal to keep as-as, but we should do simple improvements if possible.

tests/Microsoft.Identity.Test.Unit/CoreTests/OAuth2Tests/ClaimsTest.cs
```diff
@@ -251,6 +251,56 @@ public void ClaimsMerge_Test()
             Assert.AreEqual(TestConstants.ClientCapabilitiesAndClaimsJson, mergedJson);
         }
 
+        [TestMethod]
```
pmaytak, line 254: Can these be merged into one data test method with params? The const names are descriptive enough.

tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -152,6 +152,13 @@ public static HashSet<string> s_scope
...
+        public const string ClaimsWithAccessToken = @"{""access_token"":{""nbf"":{""essential"":true, ""value"":""1701477303""}}}";
+        public const string ClientCapabilitiesAndClaimsJsonWithAccessToken = @"{""access_token"":{""xms_cc"":{""values"":[""cp1"",""cp2""]},""nbf"":{""essential"":true,""value"":""1701477303""}}}";
+        public const string EmptyClaimsJson = @"{}";
+        public const string ClaimsWithAdditionalClaim = @"{""access_token"":{""nbf"":{""essential"":true, ""value"":""1701477303""},""additional_claim"":{""key"":""value""}}}";
+        public const string MergedJsonWithAdditionalClaim = @"{""access_token"":{""xms_cc"":{""values"":[""cp1"",""cp2""]},""nbf"":{""essential"":true,""value"":""1701477303""},""additional_claim"":{""key"":""value""}}}";
```
pmaytak, line 159: Good test cases.
Since in these claims and client capabilities JSONs, `access_token` node is the first one, should we add a test case where `access_token` is on different levels?
Like so:
```csharp
        public const string ClaimsWithAdditionalClaim2 = @"{""additional_claim"":{""key"":""value""}, ""access_token"":{""nbf"":{""essential"":true, ""value"":""1701477303""}}}";
```

pmaytak (changes requested):

tests/Microsoft.Identity.Test.Common/TestConstants.cs
gladjohn: added

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
gladjohn: agree that JSON would be small and not sure if there is any thing much we can improve, code seems to be well structured, just made sure to add a good test coverage so all paths are tested

tests/Microsoft.Identity.Test.Unit/CoreTests/OAuth2Tests/ClaimsTest.cs
gladjohn: was focused on the coverage and missed reducing lines, good catch. but for whatever reason VS does not seem to like datatestmethod, think with the long input data something gets messed up and VS starts throwing exceptions

bgavrilMS (approved):
Approve, but I agree with Peter's comment

src/client/Microsoft.Identity.Client/Utils/JsonHelper.cs
gladjohn: @pmaytak updated please review

tests/Microsoft.Identity.Test.Unit/CoreTests/OAuth2Tests/ClaimsTest.cs
gladjohn: tricky, updated

PR #4454, opened 2023-12-06
Adding exception help message

### LINKED ISSUE #4408: [Bug] Investigate absence of AAD error code in logs
### Library version used

4.56.0.0

### .NET version

.NET 6

### Scenario

While working with an internal partner, we noticed that while the server was responding with an error code, AADSTS700027 in this case, it was not in the logs. The only relevant information in the logs was as followed:

ErrorCode: invalid_client
HTTP StatusCode 401

This is using Microsoft.Identity.Web.TokenAcquisition.

The root cause of the actual issue was that the client need to include the sendX5C option.

### Relevant code snippets

_No response_

### Expected behavior

The error code from AAD is included in the logs.

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

trwalke:
Fixes #4408

Updating service exception error message with AAD Error code

**Testing**
Manual testing

**Performance impact**

**Documentation**


trwalke:
Seems like the issue above is by design, perhaps we can do this?
Or maybe enable the full exception by default?
I'm assuming there was a reason to make it Pii in the first place.

bgavrilMS:
@trwalke - I see that SAL also treats exceptions as not having PII. I tested a bit and I can see that most exceptions have PII obscured. Can we please have a consistent approach across all SDKs for this?

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -119,6 +119,8 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                     sb.AppendLine("=== End of inner exception stack trace ===");
                 }
 
+                sb.AppendLine("To see full exception details, enable Pii Logging.");
```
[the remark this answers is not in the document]
localden: Nit: Should this be `PII logging` given that PII is an acronym?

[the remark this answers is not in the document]
localden: Additionally, @trwalke - we should include a link to where we document how PII logging can be enabled. Assuming the doc page? https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/exceptions/msal-logging

trwalke:
@bgavrilMS Pii, such as customer content (usernames/passwords) are not there but messages will contain Oii and EUPI which ESTS is now hiding in their logs by default. I think it makes sense to add the ESTS error code and a message indicating how one can enable Pii logs.

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -110,6 +113,8 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                 {
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "HTTP StatusCode {0}", msalServiceException.StatusCode));
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "CorrelationId {0}", msalServiceException.CorrelationId));
+                    var aadError = JsonHelper.GetExistingOrEmptyString(JsonHelper.ParseIntoJsonObject(msalServiceException.ResponseBody), OAuth2ResponseBaseClaim.ErrorDescription).Split(':').FirstOrDefault();
+                    sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "AAD Error Code {0}", aadError));
```
trwalke, line 117: Updated PR to add AAD error code.

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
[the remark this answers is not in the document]
trwalke: > Additionally, @trwalke - we should include a link to where we document how PII logging can be enabled. Assuming the doc page? https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/exceptions/msal-logging

Good idea. Added it

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -110,6 +113,8 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                 {
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "HTTP StatusCode {0}", msalServiceException.StatusCode));
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "CorrelationId {0}", msalServiceException.CorrelationId));
+                    var aadError = JsonHelper.GetExistingOrEmptyString(JsonHelper.ParseIntoJsonObject(msalServiceException.ResponseBody), OAuth2ResponseBaseClaim.ErrorDescription).Split(':').FirstOrDefault();
```
gladjohn, line 116: maybe you can move this into a `ParseAadErrorCode` function?

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -119,6 +124,8 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                     sb.AppendLine("=== End of inner exception stack trace ===");
                 }
 
+                sb.AppendLine("To see full exception details, enable Pii Logging. See https://aka.ms/msal-net-logging");
+
                 if (ex.StackTrace != null)
                 {
                     sb.Append(Environment.NewLine + ex.StackTrace);
```
gladjohn, line 131:
```suggestion
                   sb.AppendLine(ex.StackTrace);
```

gladjohn:
just a little improvement in line 102, not part of your change it's the old code, but we can simply check `if (ex == null) return string.Empty;` at the top, that would avoid an unnecessary variable when it is not really required

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
gladjohn: also string interpolation would help with readability

gladjohn (approved):

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -119,6 +124,8 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                     sb.AppendLine("=== End of inner exception stack trace ===");
                 }
 
+                sb.AppendLine("To see full exception details, enable Pii Logging. See https://aka.ms/msal-net-logging");
```
pmaytak, line 127:
```suggestion
                sb.AppendLine("To see full exception details, enable PII logging. See https://aka.ms/msal-net-logging");
```

pmaytak (approved):
- Did you test with net6 and non-net6 app to test both JSON parsing flows?
- Can you please add an example of how `msalServiceException.ResponseBody` JSON looks in the description.

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
bgavrilMS, line 116: We shouldn't do parsing here but on `MsalServiceException` where extensive parsing already takes place. Btw there already exists "SubError" and "ErrorCodes" which may be what you want.

bgavrilMS (changes requested):
Parsing should be done in only one place (on the exception)

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
trwalke: Ah, yes I can use ErrorCodes

trwalke:
> * Did you test with net6 and non-net6 app to test both JSON parsing flows?
> * Can you please add an example of how `msalServiceException.ResponseBody` JSON looks in the description.

I removed the parsing. Just realized what I needed didn't need parsing.

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
trwalke: removed parsing

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -110,6 +113,7 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                 {
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "HTTP StatusCode {0}", msalServiceException.StatusCode));
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "CorrelationId {0}", msalServiceException.CorrelationId));
+                    sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "AAD Error Code AADSTS{0}", msalServiceException.ErrorCodes.FirstOrDefault()));
```
trwalke, line 116: So I see that `ErrorCodes` is an array and I am wondering if it makes sense for there to be more that one AAD error code? I've never seen it so I just picked the first one.

bgavrilMS: This is the contract with AAD. The JSON returned is like this: 

image

Why pick just the first one ? How about `string.join(ErrorCodes, " ");`

bgavrilMS (approved):
Approved with comments.

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -110,6 +113,7 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                 {
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "HTTP StatusCode {0}", msalServiceException.StatusCode));
                     sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "CorrelationId {0}", msalServiceException.CorrelationId));
+                    sb.AppendLine(string.Format(CultureInfo.InvariantCulture, "AAD Error Code AADSTS{0}", string.Join(" ", msalServiceException.ErrorCodes)));
```
localden, line 116: @trwalke we should probably start calling this "Microsoft Entra ID" instead of AAD 😀

localden (changes requested):

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
trwalke: will do

PR #4458, opened 2023-12-07
Add .WithForceRefresh support for silent broker flows

### LINKED ISSUE #4457: Add .WithForceRefresh support for silent broker flows
### Library version used

4.57.0

### .NET version

all

### Scenario

PublicClient - desktop app

None

Currently Broker flows don't allow bypassing broker cache when option  .WithForceRefresh(true) is set. 

### Relevant code snippets

```csharp
.WithForceRefresh(true);
```

### Expected behavior

When using builder .WithForceRefresh(true) MSAL.runtime and WAM cache is bypassed.

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

iulico-1:
Fixes #4457

Currently Broker flows don't allow bypassing broker cache when option .WithForceRefresh(true) is set. This PR proposes to fix that.

**Testing**
No e2e test infra seem to exist for e2e WAM tests. 
Manual validation was done. Test app has been updated to support this option.

**Performance impact**
There's possibility that using builder  .WithForceRefresh(true) will result in 2 network calls if the cache is already expired.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
@@ -312,6 +312,21 @@ namespace Microsoft.Identity.Client.Platforms.Features.RuntimeBroker
                         var errorMessage = "Could not acquire token silently.";
                         msalTokenResponse = WamAdapters.HandleResponse(result, authenticationRequestParameters, _logger, errorMessage);
                     }
+
+                    if (acquireTokenSilentParameters.ForceRefresh && !string.IsNullOrEmpty(msalTokenResponse.AccessToken))
+                    {
+                        authParams.AccessTokenToRenew = msalTokenResponse.AccessToken;
+
+                        using (NativeInterop.AuthResult result = await s_lazyCore.Value.AcquireTokenSilentlyAsync(
+                        authParams,
+                        authenticationRequestParameters.CorrelationId.ToString("D"),
+                        readAccountResult.Account,
+                        cancellationToken).ConfigureAwait(false))
+                        {
+                            var errorMessage = "Could not acquire token silently with AccessTokenToRenew option.";
+                            msalTokenResponse = WamAdapters.HandleResponse(result, authenticationRequestParameters, _logger, errorMessage);
+                        }
+                    }
                 }
             }
 
@@ -343,6 +358,21 @@ namespace Microsoft.Identity.Client.Platforms.Features.RuntimeBroker
                     var errorMessage = "Could not acquire token silently for the default user.";
                     msalTokenResponse = WamAdapters.HandleResponse(result, authenticationRequestParameters, _logger, errorMessage);
                 }
+
+                if (acquireTokenSilentParameters.ForceRefresh && !string.IsNullOrEmpty(msalTokenResponse.AccessToken))
+                {
+                    authParams.AccessTokenToRenew = msalTokenResponse.AccessToken;
+
+                    using (NativeInterop.AuthResult result = await s_lazyCore.Value.SignInSilentlyAsync(
+                       authParams,
+                       authenticationRequestParameters.CorrelationId.ToString("D"),
+                       cancellationToken).ConfigureAwait(false))
+                    {
+                        var errorMessage = "Could not acquire token silently for the default user.";
+                        msalTokenResponse = WamAdapters.HandleResponse(result, authenticationRequestParameters, _logger, errorMessage);
+                    }
+                }
+
             }
 
             return msalTokenResponse;
--- a/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs
@@ -169,8 +169,7 @@ namespace Microsoft.Identity.Test.Integration.Broker
             }
         }
 
-        [RunOn(TargetFrameworks.NetCore)]
-        [ExpectedException(typeof(MsalUiRequiredException))]
+        [RunOn(TargetFrameworks.NetCore)]        
         public async Task WamUsernamePasswordRequestAsync()
         {
             var labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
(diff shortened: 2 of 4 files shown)


src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -312,6 +312,21 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
                         var errorMessage = "Could not acquire token silently.";
                         msalTokenResponse = WamAdapters.HandleResponse(result, authenticationRequestParameters, _logger, errorMessage);
                     }
+
+                    if (acquireTokenSilentParameters.ForceRefresh && !string.IsNullOrEmpty(msalTokenResponse.AccessToken))
```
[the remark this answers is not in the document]
rayluo: The AuthResult of recent version of native MSALs will have an attribute indicating the token source whose value can be "IdentityProvider" or "cache" etc.. Unless MsalRuntime also provide a similar mechanism, the only way to approximately know whether a token is new is to run the same token acquisition code path immediately and assert the new token will be different than the old token.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -312,6 +312,21 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
...
+                    {
+                        authParams.AccessTokenToRenew = msalTokenResponse.AccessToken;
```
rayluo, line 318: Out of curiosity, under what circumstance will a mandatory refresh become necessary?

All native MSALs have `.WithForceRefresh(...)` but such an api risks being abuse to always bypass token cache. OneAuth/MsalRuntime's `AccessTokenToRenew` is a better design which does not have the same risk. If we are making a change here anyway, I would suggest to consider adding a new `.WithAccessTokenToRenew(...)` design to native MSALs.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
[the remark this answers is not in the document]
trwalke: You can look at how we test MSAL Runtime broker [here.](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/79f951a079ce2db3290874c7358d5c3242fbbe05/tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/RuntimeBrokerTests.cs#L174) As you can see, we only have the one `TokenSource.Broker` result that represents both from IDP and broker cache so there is no way to tell from that property if it is form the cache with that property. I suggest making a second silent call with the new api and verifying that the second access token does not match the first one as @rayluo suggests.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -312,6 +312,21 @@ private IntPtr GetParentWindow(CoreUIParent uiParent)
...
+
+                        using (NativeInterop.AuthResult result = await s_lazyCore.Value.AcquireTokenSilentlyAsync(
```
trwalke, line 320: So this is calling `AcquireTokenSilentlyAsync` on MSAL Runtime a second time but we are now setting the `AccessTokenToRenew` property. Are there plans to enable MSAL runtime to respond to the `ForceRefresh` value the first time so we dont need to make 2 calls?

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
ashok672: The abuse of this force refresh was discussed today and concluded eSTS will throttle in care of abuse. That said, this option exists today. Are you suggesting to remove it?

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
iulico-1: The plan is to expose AccessTokenToRenew() to reduce the chance of WithForceRefresh() being abused.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
iulico-1: Yes, I totally agree with abuse part and currently the plan is to unify the programing model between MSAL.CPP and native MSALs in future releases but the decision for this particular fix was to use the existing paradigm.

For the investigation I am working through a fresh AT is needed when user went and removed the account from broker and added it again to the broker this way re-provisioning the device with a new device id. The issue is the cached AT has the wrong device-id claim.

rayluo: > For the investigation I am working through a fresh AT is needed when user went and removed the account from broker and added it again to the broker this way re-provisioning the device with a new device id. The issue is the cached AT has the wrong device-id claim.

Legit use case. And it is indeed a case that the existing native MSAL's WithForceRefresh() is difficult to handle. It relies on the app developers to call WithForceRefreesh() only when necessary.


> this option exists already in public API surface. Are you suggesting to remove it?

No. Never remove existing API unless we are willing to bump major version number.


> decision for this particular fix was to use the existing paradigm.

If this time is not a good timing and good reason to naturally add the new paradigm, I don't know when it will be. :-)

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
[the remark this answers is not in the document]
bgavrilMS: @iulico-1 - here's my attempt at an integration test for this. It uses ROPC, because we don't run interactive auth in MSAL.NET

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4459/files

(feel free to merge it into your branch)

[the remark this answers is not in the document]
iulico-1: Excellent. Thank you @bgavrilMS !

bgavrilMS:
@pmaytak - as you are on-call, can you please review this? The HOW has been discussed internally.

PR #4471, opened 2023-12-14
Use linked cancellation token source during proactive refresh.

### LINKED ISSUE #4473: [Bug] Background proactive token refresh is not properly canceled
### Library version used

4.58.1

### .NET version

Any

### Scenario

ConfidentialClient - web api (AcquireTokenOnBehalfOf), ConfidentialClient - service to service (AcquireTokenForClient), ManagedIdentityClient - managed identity

None

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/546d00f3cbf52e8ca035a789736d9d802be0e5d7/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs#L78

Proactive token refresh is done on a separate thread. A cancellation token provided by the user is passed into it. If the parent token source is disposed after the original token request completes, then any usage of that cancellation token in the proactive refresh task will cause a "object is disposed" exception. 

### Relevant code snippets

_No response_

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

Create a linked cancellation token source using the parent user-provided cancellation token to be used in the background refresh task.

pmaytak:
Fixes #4473

- Create a linked cancellation token source for the proactive refresh task.

**Testing**
- Manually
- Unit tests

**Performance impact**
N/A

**Documentation**


pmaytak:
Alternative solution to passing a cancellation token is not to propagate the cancellation to the proactive refresh since it's kind of a fire-and-forget background operation. The reasoning for passing the cancellation is since the original operation was canceled, perhaps all related operations should be canceled too.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -75,7 +75,11 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
                         SilentRequestHelper.ProcessFetchInBackground(
                         cachedAccessTokenItem,
-                        () => GetAccessTokenAsync(cancellationToken, logger), logger);
+                        () =>
+                        {
+                            using var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```
bgavrilMS, line 80: Pls add some comments on why this is done, i.e. to avoid the app disposing of the ct. Or have a helper method that is named appropriately.

gladjohn (approved):

PR #4476, opened 2023-12-20
Refactor Exception Handling: Transition from ManagedIdentityException to MsalServiceException

gladjohn:
Fixes : #4483

Update Managed Identity Exceptions to MSALServiceExceptions

Previously, **ManagedIdentityException** only derived from **ServiceException** and not from **UIRequiredException**. This created a disconnect in our hierarchy of exceptions, as it didn't encompass all the necessary types that our application deals with, particularly in scenarios requiring user interaction (CAE with SLC). By moving to **MsalServiceException**, we ensure that all exceptions fit neatly into our established hierarchy, making it more predictable and easier to manage.

The introduction of **ManagedIdentityException** was initially intended to add extra data about the source leading to this exception. However, this approach led to a fragmentation in our exception handling logic. In scenarios involving shared logic between Managed Service Identity (MSI) and Security Token Service (STS), the code needed to handle or throw the correct exception type, leading to unnecessary complexity (SLC). By standardizing on **MsalServiceException**, we eliminate this fragmentation, allowing for more streamlined and reusable exception handling code.

The previous approach often required catching **MsalServiceException** and rethrowing it as **ManagedIdentityException** in certain contexts. This added an extra layer of complexity to our exception handling logic. With this refactor, we eliminate such patterns, simplifying the flow and making the code more maintainable.


src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -94,17 +84,17 @@ await _requestContext.ServiceBundle.HttpManager
...
             {
-                _requestContext.Logger.Error($"[Managed Identity] Exception: {e.Message} Http status code: {response?.StatusCode}");
-                exception = e;
-                message = MsalErrorMessage.ManagedIdentityUnexpectedResponse;
+                HandleException(e);
+                throw;
             }
 
-            throw new MsalManagedIdentityException(MsalError.ManagedIdentityRequestFailed, message, exception, _sourceType, (int)response.StatusCode);
+            return Task.FromResult<ManagedIdentityResponse>(null);
```
bgavrilMS, line 97: Why are you changing the logic to return null instead of throwing?

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -141,5 +130,97 @@ internal string GetMessageFromErrorResponse(HttpResponse response)
 
             return $"[Managed Identity] Error Code: {managedIdentityErrorResponse.Error} Error Message: {managedIdentityErrorResponse.ErrorDescription}";
         }
+
+        internal void HandleException(Exception ex, ManagedIdentitySource source = ManagedIdentitySource.None, string additionalInfo = null)
```
bgavrilMS, line 134: We have an MsalServiceExceptionFactory which is better suited for this?

src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.Identity.Client
     /// This exception class is for exceptions generated from Managed Identity sources.
     /// For more details, see https://aka.ms/msal-net-managed-identity
     /// </summary>
+    [Obsolete("MsalManagedIdentityException is deprecated. Use MsalServiceException instead.")]
```
bgavrilMS, line 14: Add the flag that breaks the build and set it to true - this can be a hard deprecation.

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -141,5 +130,97 @@ internal string GetMessageFromErrorResponse(HttpResponse response)
...
+            if (ex is HttpRequestException httpRequestException)
+            {
+                ThrowManagedIdentityException(
+                    MsalError.ManagedIdentityUnreachableNetwork,
+                    httpRequestException.Message,
+                    httpRequestException.InnerException);
+            }
+            else if (ex is TaskCanceledException taskCanceledException)
+            {
+                _requestContext.Logger.Error(TimeoutError);
```
bgavrilMS, line 145: This is not an error. User might have set CancellationToken. I recommend you don't treat this at all and just let it buble up.

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -141,5 +130,97 @@ internal string GetMessageFromErrorResponse(HttpResponse response)
...
+            Exception innerException = null,
+            ManagedIdentitySource managedIdentitySource = ManagedIdentitySource.None,
+            int? statusCode = null)
+        {
+            ManagedIdentitySource source = managedIdentitySource != ManagedIdentitySource.None ? managedIdentitySource : _sourceType;
+
+            ThrowServiceException(errorCode, errorMessage, innerException, source, statusCode);
+        }
+
+        internal static void ThrowServiceException(
```
bgavrilMS, line 186: We really should not have new logic that deals with this. The factory is for throwing service exceptions. The source is just "additional data".

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
bgavrilMS, line 134: None of the new methods need to be internal. All private.

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -141,5 +130,97 @@ internal string GetMessageFromErrorResponse(HttpResponse response)
...
+                
+                ThrowManagedIdentityException(
+                    MsalError.ManagedIdentityRequestFailed,
+                    ex.Message,
+                    ex);
+            }
+            else
+            {
+                // If it's already a MsalServiceException, rethrow it
+                throw ex;
```
bgavrilMS, line 170: When you do this, we lose the stack trace. See https://stackoverflow.com/questions/730250/is-there-a-difference-between-throw-and-throw-ex

I recommend something like

try
{
// msi logic
}
catch (Exception ex) when ex is not MsalServiceException or TaskCancelledException
{
     // repackage exception.
}

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -141,5 +130,97 @@ internal string GetMessageFromErrorResponse(HttpResponse response)
...
+
+            ThrowServiceException(errorCode, errorMessage, innerException, source, statusCode);
+        }
+
+        internal static void ThrowServiceException(
+            string errorCode, 
+            string errorMessage,
+            Exception innerException,
+            ManagedIdentitySource managedIdentitySource,
+            int? statusCode = null)
```
bgavrilMS, line 191: The factory has a better way of treating this. Why not just add something similar to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs#L116

src/client/Microsoft.Identity.Client/ManagedIdentity/AzureArcManagedIdentitySource.cs
```diff
@@ -40,8 +40,15 @@ public static AbstractManagedIdentity TryCreate(RequestContext requestContext)
...
             if (!Uri.TryCreate(identityEndpoint, UriKind.Absolute, out Uri endpointUri))
             {
-                throw new MsalManagedIdentityException(MsalError.InvalidManagedIdentityEndpoint, string.Format(
-                    CultureInfo.InvariantCulture, MsalErrorMessage.ManagedIdentityEndpointInvalidUriError, "IDENTITY_ENDPOINT", identityEndpoint, AzureArc), 
+                string errorMessage = string.Format(
+                    CultureInfo.InvariantCulture,
+                    MsalErrorMessage.ManagedIdentityEndpointInvalidUriError,
+                    "IDENTITY_ENDPOINT", identityEndpoint, AzureArc);
+
+                ThrowServiceException(
```
bgavrilMS, line 48: I think this is an MsalClientException not a service exception.

src/client/Microsoft.Identity.Client/ManagedIdentity/AzureArcManagedIdentitySource.cs
```diff
@@ -56,8 +63,12 @@ public static AbstractManagedIdentity TryCreate(RequestContext requestContext)
 
             if (requestContext.ServiceBundle.Config.ManagedIdentityId._isUserAssigned)
             {
-                throw new MsalManagedIdentityException(MsalError.UserAssignedManagedIdentityNotSupported, 
-                    string.Format(CultureInfo.InvariantCulture, MsalErrorMessage.ManagedIdentityUserAssignedNotSupported, AzureArc),
+                string errorMessage = string.Format(CultureInfo.InvariantCulture, MsalErrorMessage.ManagedIdentityUserAssignedNotSupported, AzureArc);
+                
+                ThrowManagedIdentityException(
```
bgavrilMS, line 68: client?

src/client/Microsoft.Identity.Client/ManagedIdentity/AzureArcManagedIdentitySource.cs
```diff
@@ -56,8 +63,12 @@ public static AbstractManagedIdentity TryCreate(RequestContext requestContext)
 
             if (requestContext.ServiceBundle.Config.ManagedIdentityId._isUserAssigned)
```
bgavrilMS, line 64: nit: _isUserAssigned should not be private field.

pmaytak:
What is the reason for changing this behavior?

bgavrilMS:
> What is the reason for changing this behavior?

The `ManagedIdentityException` didn't play well with our hierarchy of exceptions, since it only derviced from ServiceException. But not from UIRequiredException. The only reason we added this exception was to be able to add extra bit of data about what source leads to this exception.

It also prevents code reuse, since any shared logic between MSI and STS now has to deal with the problem of throwing the correct exception (or catching MsalServiceEx and rethrowing MSIException, which is also complex).

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
gladjohn: we already do this, are you suggesting we repack the exception if it is MSAL Service exception? 

```cs
else if (ex is not MsalServiceException or TaskCanceledException)
{
    _requestContext.Logger.Error($"[Managed Identity] Exception: {ex.Message}");
    CreateAndThrowException(MsalError.ManagedIdentityRequestFailed, ex.Message, ex, source);
}
else
{
    // If it's already a MsalServiceException, rethrow it
    throw ex;
}
```

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityId.cs
```diff
@@ -21,23 +21,22 @@ public class ManagedIdentityId
...
+        /// <summary>
+        /// Gets a value indicating whether the managed identity is user-assigned.
+        /// </summary>
+        /// <remarks>
+        /// This property is true if the managed identity is user-assigned, and false if it is system-assigned.
+        /// </remarks>
+        /// <value>
+        /// True if the managed identity is user-assigned; otherwise, false.
+        /// </value>
+        internal bool IsUserAssigned { get; private set; }
```
bgavrilMS, line 34: You can delete the "private set" becasue this is onyl set in the ctor. THis is a read-only property. 
Also update IdType to be read-only.

nit: update all properties to be read-only - this makes the class immutable. Immutable objects are very easy to reason with and we should keep our objects immutable as much as possible.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
gladjohn: In Line 91, exception MSALServiceException is thrown when the Managed Identity response is not a success, this is a custom exception based on the failure message from line 87, 

```
string message = GetMessageFromErrorResponse(response);
```

and the custom MSAL Service Exception is thrown in Line 98, 

```
throw exception;
```

this check here, is done for any failure that happens after getting a successful response, 

```
catch (Exception e)
{
    HandleException(e);
    throw;
}
```

src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityId.cs
gladjohn: updated where applicable

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
gladjohn: updated @pmaytak will do some more manual tests

pmaytak (approved):

gladjohn:
updated an [error message](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f5aa83c2a943466e963b2361a9f5291d15c89699/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs#L30) to add more details

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -141,5 +135,52 @@ internal string GetMessageFromErrorResponse(HttpResponse response)
...
+            }
+            else if (ex is not MsalServiceException or TaskCanceledException)
+            {
+                _requestContext.Logger.Error($"[Managed Identity] Exception: {ex.Message}");
+                CreateAndThrowException(MsalError.ManagedIdentityRequestFailed, ex.Message, ex, source);
+            }
+            else
+            {
+                // If it's already a MsalServiceException, rethrow it
+                throw ex;
```
neha-bhargava, line 167: You don't need to rethrow the exception here since you are already doing it at line 70 after the HandleException is called.

neha-bhargava (approved):

PR #4479, opened 2023-12-21
Order scopes on save to avoid access token duplication in the cache

### LINKED ISSUE #4474: TokenCache expects that token providers send pre-sorted (alphabetically) scopes - Auth0 does not - which results in multiple_matching_tokens_detected
### Library version used

4.5.7

### .NET version

net4.8

### Scenario

[MsalAccessTokenCacheItem](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs#L33C45-L33C99) seems to expect that the token returned from the token endpoint has the scopes ordered alphabetically.
The scopes are included in the cache-key - which results in cache miss if the scopes are sorted different in the response to `AcquireTokenByAuthorizationCode `vs the response to `msal.AcquireTokenSilent(scopes, account).WithForceRefresh(true)` 

This in turn results in multiple access-tokens added to the cache - which becomes a problem when trying to obtain a token silently from the cache.
> MSAL.Desktop.4.57.0.0.MsalClientException: ErrorCode: multiple_matching_tokens_detected Microsoft.Identity.Client.MsalClientException: The cache contains multiple tokens satisfying the requirements. Try to clear token cache. at Microsoft.Identity.Client.Internal.Requests.Silent.SilentRequest.d__5.MoveNext()

I have read through the Oauth2 spec without finding any requirements for how scopes should be ordered in the token response.

The picture below shows the responses from the code exchange and the refreshtoken exchange. The result of these two requests are two items in the token cache - with the same scopes, but with different order.

image
image

### Relevant code snippets

```csharp
//Setup application with "generic autority (auth0)
var msal = ConfidentialClientApplicationBuilder
                .Create(ClientId)
                .WithRedirectUri(RedirectUri)
                .WithClientSecret(ClientSecret)
                .WithLegacyCacheCompatibility(false)
                .WithGenericAuthority($"https://{Tenant}")
                .WithLogging(new MyIdentityLogger(),true)
                .Build();

//Configure token caching using SQL server
msal.AddDistributedTokenCache(services => {
                services.AddDistributedTokenCaches();
                services.AddDistributedSqlServerCache(options => {
                    options.SchemaName = "dbo";
                    options.TableName = "TokenCache";
                    options.ConnectionString = TokenCacheSqlServerConnectionString;
                    options.DefaultSlidingExpiration = TimeSpan.FromDays(30);
                });
            });

//Exchange authcode with access-, id- and refreshtoken
await msal.AcquireTokenByAuthorizationCode(scopes, notification.ProtocolMessage.Code)
                        .ExecuteAsync();

//Force usage of refresh token
await msal.AcquireTokenSilent(scopes, account).WithForceRefresh(true).ExecuteAsync();

//try to aquire token from cache - this fails because of multiple access tokens in cache which have the same scopes
await msal.AcquireTokenSilent(scopes, account).ExecuteAsync()
```

### Expected behavior

_No response_

### Identity provider

Other

### Regression

_No response_

### Solution and workarounds

Not sure how this might affect existing cache-items, but either this library has to sort the returned scopes before creating the cache key, or the identity provider has to be consistent on the order of returned scopes.

Changing the order of the requested scopes does not seem to make any difference on the response from Auth0

bgavrilMS:
Fixes #4474 

I verified that I am not touching the hot path (getting token from the cache), the reorder happens on saving token response from HTTP, which is already heavy operation.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Identity.Client.Cache.Items
             string keyId = null,
             string oboCacheKey = null)
             : this(
-                scopes: response.Scope, // token providers send pre-sorted (alphabetically) scopes
+                scopes: ScopeHelper.OrderScopesAlphabetically(response.Scope), // order scopes to avoid cache duplication. This is not in the hot path.
                 cachedAt: DateTimeOffset.UtcNow,
                 expiresOn: DateTimeHelpers.DateTimeOffsetFromDuration(response.ExpiresIn),
                 extendedExpiresOn: DateTimeHelpers.DateTimeOffsetFromDuration(response.ExtendedExpiresIn),
@@ -49,6 +49,7 @@ namespace Microsoft.Identity.Client.Cache.Items
             InitCacheKey();
         }
 
+      
         internal /* for test */ MsalAccessTokenCacheItem(
             string preferredCacheEnv,
             string clientId,
--- a/tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
+++ b/tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
@@ -6,6 +6,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Utils;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
@@ -23,6 +24,24 @@ namespace Microsoft.Identity.Test.Common.Core.Helpers
             foreach (string expectedScope in expectedScopes)
             {
                 Assert.IsTrue(actualScopes.Contains(expectedScope));
+            }
+        }
+
+        internal static void AreAccountsEqual(
+            string expectedUsername,
+            string expectedEnv,
+            string expectedId,
+            string expectedTid,
+            string expectedOid,
+            params IAccount[] accounts)
+        {            
+            foreach (var account in accounts)
+            {
+                Assert.AreEqual(expectedUsername, account.Username);
+                Assert.AreEqual(expectedEnv, account.Environment);
+                Assert.AreEqual(expectedId, account.HomeAccountId.Identifier);
+                Assert.AreEqual(expectedTid, account.HomeAccountId.TenantId);
+                Assert.AreEqual(expectedOid, account.HomeAccountId.ObjectId);
             }
         }
 
--- a/src/client/Microsoft.Identity.Client/Internal/IdToken.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/IdToken.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Identity.Client.Internal
         public const string Version = "ver";
         public const string PreferredUsername = "preferred_username";
         public const string Name = "name";
+        public const string Email = "email";
         public const string HomeObjectId = "home_oid";
         public const string GivenName = "given_name";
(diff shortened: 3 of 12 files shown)


tests/Microsoft.Identity.Test.Common/Core/Helpers/CoreAssert.cs
```diff
@@ -26,6 +27,25 @@ public static void AreScopesEqual(string scopesExpected, string scopesActual)
...
 
+        internal static void AreAccountsEqual(
+            string expectedUsername,
+            string expectedEnv,
+            string expectedId,
+            string expectedTid,
+            string expectedOid,
+            params IAccount[] accounts)
+        {
+            // implement this!
```
pmaytak, line 38: What does this mean?

pmaytak (approved):

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -30,7 +30,7 @@ internal class MsalAccessTokenCacheItem : MsalCredentialCacheItemBase
             string keyId = null,
             string oboCacheKey = null)
             : this(
-                scopes: response.Scope, // token providers send pre-sorted (alphabetically) scopes
+                scopes: ScopeHelper.OrderScopesAlphabetically(response.Scope), // order scopes to avoid cache duplication. This is not in the hot path.
```
rayluo, line 33: Does it solve #4474 by normalizing the token cache key to use an ordered scope string?

For what it's worth, I brought up this topic before, but it was considered too late to change the specs. That being said, I still think the token cache key shall use normalized/ordered scope, which [will give token search a performance boost](https://github.com/AzureAD/microsoft-authentication-library-for-python/blame/1.26.0/msal/token_cache.py#L96-L98). Token cache compatibility will not really be affected by normalizing AT's key.

pmaytak: > Does it solve https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4474 by normalizing the token cache key to use an ordered scope string?

Yes. 

The assumption was that AAD always returns sorted scopes, while that seems to usually happen, it's not a strict contact (and order is only for non-OAuth scopes).

bgavrilMS: The scopes in the token response are ordered for Azure related IdP (AAD, B2C, CIAM), but not for other identity providers.

rayluo: > The scopes in the token response are ordered for Azure related IdP (AAD, B2C, CIAM), but not for other identity providers.

In that case, MSAL might as well just normalize (i.e., sort) the scopes before writing the key into token cache. Is that this PR already doing? Other MSALs can follow suit. @bgavrilMS

bgavrilMS: Yes, this is what this fix is doing. But I don't think you can optimize the token lookup logic, because of older token caches.

bgavrilMS:
@gladjohn - can I please get a review on this?

gladjohn (approved):
looks good, but based on this comment from the old code `token providers send pre-sorted (alphabetically) scopes`, if this is mostly true, then would it be performant to check if scopes are sorted already before sorting them?

bgavrilMS:
> looks good, but based on this comment from the old code `token providers send pre-sorted (alphabetically) scopes`, if this is mostly true, then would it be performant to check if scopes are sorted already before sorting them?

I am not sure that checking if scopes are ordered is slower or not. The important thing here is that it's not in the hot path - this is only invoked when MSAL saves tokens in its cache.

PR #4486, opened 2024-01-04
Adds a check if OpenTelemetry instrumentation can be used.

### LINKED ISSUE #4456: [Bug] Dependency on System.Diagnostics.DiagnosticSource (>= 7.0.2)  in version 4.58.0 missing in .NET 6 Azure Functions
### Discussed in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/discussions/4455

<div type='discussions-op-text'>

<sup>Originally posted by **Ozindar** December  6, 2023</sup>
We've got a .NET 6.0 Azure Function using the Identity.Client. This worked fine with 4.57. With a recently deploy we've upgraded to 4.58 but in Azure this triggers a File not found when it's searching for the DiagnosticsSource version 7 assembly.  
"dotnet publish" does not seem put that dll in the output folder. For the moment I've pinned the Identity.Client to 4.58.  
Is there a viable solution to keep using the latest Identity.Client in combination with a .NET6 Azure Function and what would be the best approach?</div>

pmaytak:
Fixes #4456. An alternative to #4487 solution.

Catching error when instantiating Open Telemetry fails (means it's not available in that environment) and using null telemetry implementation instead.

**Testing**
Manual.

This solution seems to be better than #4487, since there may be a need to bump DiagnosticsSource version if we need to use the newer APIs.

### DIFF
--- a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
+++ b/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.AuthScheme.PoP;
 using Microsoft.Identity.Client.Cache;
@@ -55,7 +54,14 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
         private IOtelInstrumentation InternalGetOtelInstrumentation()
         {
 #if SUPPORTS_OTEL
-            return new OtelInstrumentation();
+            try
+            {
+                return new OtelInstrumentation();
+            } catch (Exception ex)
+            {
+                Logger.Warning("Failed instantiating OpenTelemetry instrumentation. Exception: " + ex);
+                return new NullOtelInstrumentation();
+            }
 #else
             return new NullOtelInstrumentation();
 #endif
--- a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
+++ b/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.IO;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.AuthScheme.PoP;
 using Microsoft.Identity.Client.Cache;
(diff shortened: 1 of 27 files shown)


src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
```diff
@@ -55,7 +54,14 @@ protected AbstractPlatformProxy(ILoggerAdapter logger)
         private IOtelInstrumentation InternalGetOtelInstrumentation()
         {
 #if SUPPORTS_OTEL
-            return new OtelInstrumentation();
+            try
+            {
+                return new OtelInstrumentation();
+            } catch (Exception ex)
```
bgavrilMS, line 60: Is there a more specific exception that should be caught?

pmaytak: FileNotFoundException.

bgavrilMS: I'd start with that. We can always add more types.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
```diff
@@ -55,7 +55,15 @@ protected AbstractPlatformProxy(ILoggerAdapter logger)
         private IOtelInstrumentation InternalGetOtelInstrumentation()
         {
 #if SUPPORTS_OTEL
-            return new OtelInstrumentation();
+            try
+            {
+                return new OtelInstrumentation();
+            } catch (FileNotFoundException ex)
```
gladjohn, line 61: would it be safe to also include `TypeLoadException`?

gladjohn (approved):

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/AbstractPlatformProxy.cs
pmaytak: Hmm, not sure. This was the only exception that was thrown in this scenario. I would catch just this one, in case `TypeLoadException` is thrown in some unrelated scenarios?

pmaytak:
> I would recommend using this solution rather than #4487 since the problem will still occur in case anyone uses otel exporter version >=1.4 and the workarounds for the versions will still be needed.

I'll combine the two PRs actually. Yes, if client app uses OpenTelemetry 1.4+, that'll still break but it won't be because of MSAL. The main issue here is client apps that **don't** use any telemetry should work and not break because of MSAL.

PR #4487, opened 2024-01-04
Downgrade System.Diagnostics.DiagnosticSource to v6 and add a check if OpenTelemetry instrumentation can be used

### LINKED ISSUE #4456: [Bug] Dependency on System.Diagnostics.DiagnosticSource (>= 7.0.2)  in version 4.58.0 missing in .NET 6 Azure Functions

pmaytak:
Fixes #4456. An alternative to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4486 solution.

- Downgrade System.Diagnostics.DiagnosticSource to v6, which is the same version as what Azure Functions runtime uses.
- Catching error when instantiating Open Telemetry fails (means it's not available in that environment) and using null telemetry implementation instead.
- See issue for more details.

**Testing**
- Unit tests pass.
- Tested with a dev app.
- Also tested that Activity code from https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4243/commits/5dc10b22d580254e86c683fc05396c91eb1e9010 works.


gladjohn:
thanks for referencing this [issue](https://github.com/Azure/azure-functions-host/issues/8938), so even downgrading still causes some issues?

pmaytak:
> thanks for referencing this [issue](https://github.com/Azure/azure-functions-host/issues/8938), so even downgrading still causes some issues?

Currently downgrading doesn't cause issues. However, it's not future-proof if we ever need newer APIs. Additionally, if the client app enables exporters, OpenTelemetry libraries 1.4.0 and above depend on DiagnosticSource 7.0.0+, so they'll have to use one of the workarounds anyway.

bgavrilMS (approved):

Directory.Packages.props
```diff
@@ -15,7 +15,7 @@
     <PackageVersion Include="Microsoft.IdentityModel.Abstractions" Version="6.22.0" />
     <PackageVersion Include="Microsoft.Web.WebView2" Version="1.0.864.35" />
     <PackageVersion Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
-    <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />
+    <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="6.0.1" />
```
bgavrilMS, line 18: Maybe add a comment here about Az Functions issue.

bgavrilMS:
Have you thought about merging both PRs? So that it fixes the assembly load and fallbacks when Az Functions moves from 6 to 8 (?). Alternatively, this PR is enough and when Az Functions updates, the upgrade will break some folks. This should be a strong signal that they need to figure out a better solution. Thoughts?

neha-bhargava:
Were you able to test this with the exporter manually?

pmaytak:
> Have you thought about merging both PRs? So that it fixes the assembly load and fallbacks when Az Functions moves from 6 to 8 (?). Alternatively, this PR is enough and when Az Functions updates, the upgrade will break some folks. This should be a strong signal that they need to figure out a better solution. Thoughts?

Yea, I'll merge the PRs.

You mean when Azure Functions update their runtime to v8? I tested with a v5 library and everything works. So if they upgrade to v9 and we use v6/7, it's okay.

> Were you able to test this with the exporter manually?

Yes, with the dev app. Workarounds still needed for in-process Function, but works otherwise.

gladjohn (approved):

neha-bhargava (approved):
I did a quick test with MISE exporter too and logs look fine. Thanks!

PR #4491, opened 2024-01-09
Fix performance tests project and improvements

pmaytak:
- Disable Central Package Management for the performance tests project since Benchmark.Net doesn't support it.
- Throw an exception if no tests ran to break the build.
- Run perf project on .NET 6. (Tried on .NET 8 but the whole MSAL solution fails to build, so this will do for now.)
- Upgrade package dependencies.
- Delete some unused yaml files.
- Clean up the YAML files a bit.


build/template-desktop-unit-and-automation.yaml
```diff
@@ -1,18 +1,8 @@
...
-  displayName: 'Use .Net Core SDK 5.x'
-  inputs:
-    version: 5.x    
-
-- task: UseDotNet@2
-  displayName: 'Use .Net Core SDK 6.x'
-  inputs:
-    version: 6.x
+    version: 8.x
     includePreviewVersions: true
```
jennyf19, line 7: You probably don't need this anymore

tests/Microsoft.Identity.Test.Performance/Microsoft.Identity.Test.Performance.csproj
```diff
@@ -2,15 +2,16 @@
...
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
-    <Configurations>Debug;Release;Debug + MobileApps</Configurations>
+    <TargetFramework>net8.0</TargetFramework>
+    <Configurations>Debug;Release;</Configurations>
     <!-- Uncomment only when running EtwProfiler diagnoser on Release-->
     <!-- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation#debugtype -->
     <!--<DebugType>pdbonly</DebugType>
     <DebugSymbols>true</DebugSymbols>-->
     <LangVersion>latest</LangVersion>
```
jennyf19, line 11: >latest [](http://example.com/codeflow?start=17&length=6)

learned from Stephen Toub that you should define the lang version -> see [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2184)

also, do you have a Directory.Build.Props file for shared content across projects?

tests/Microsoft.Identity.Test.Performance/Program.cs
```diff
@@ -19,33 +21,40 @@ public static void Main(string[] args)
...
                         //.AddDiagnoser(new EtwProfiler()) // https://adamsitnik.com/ETW-Profiler/
-                .AddJob(
-                    Job.Default
-                        .WithId("Job-PerfTests")), args);
+                 , args);
+
+                // If no tests ran for whatever reason, throw an exception to break the build.
+                Summary summary = results?.FirstOrDefault();
+                BenchmarkReport report = summary?.Reports.FirstOrDefault();
+                if (summary == null || report == null || !report.Success)
```
jennyf19, line 49: what do the results look like?

build/template-desktop-unit-and-automation.yaml
pmaytak: Yea, that was probably for when Maui first came out.

tests/Microsoft.Identity.Test.Performance/Microsoft.Identity.Test.Performance.csproj
pmaytak: I don't quite understand why `latest` mapped to pre-11? [`latest` is based on the compiler version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#c-language-version-reference), so should map to 12 in this case? If the `LangVersion` is absent, then the TFM dictates the language version.

Yes, we have three Directory.Build.props: https://github.com/search?q=repo%3AAzureAD%2Fmicrosoft-authentication-library-for-dotnet+path%3A**%2Fdirectory.build.props&type=code

tests/Microsoft.Identity.Test.Performance/Microsoft.Identity.Test.Performance.csproj
```diff
@@ -8,7 +8,7 @@
     <!-- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation#debugtype -->
     <!--<DebugType>pdbonly</DebugType>
     <DebugSymbols>true</DebugSymbols>-->
-    <LangVersion>latest</LangVersion>
+    <LangVersion>12</LangVersion>
```
bgavrilMS, line 11: Does this need to be specified here? We should use a Directory.Builds.props for common stuff.

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Performance/Program.cs
pmaytak: It's everything that's printed out to the console during the run - collections of benchmark cases, all the metrics, environment info, settings, etc.

tests/Microsoft.Identity.Test.Performance/Microsoft.Identity.Test.Performance.csproj
pmaytak: Indeed, but seems like changing this stuff is finicky, so I'll leave it out of this PR.

build/template-ci-and-pr.yaml
trwalke, line 1: This file was not being used by anything right?

trwalke (approved):

build/template-ci-and-pr.yaml
pmaytak: Yea, the build pipelines use different files and these do not seem to be referenced anywhere.

gladjohn (approved):

PR #4494, opened 2024-01-10
update langVersion to 12

kellyyangsong:
Related to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2184#issue-1824984342

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -27,7 +27,7 @@
     <IsPackable>true</IsPackable>
     
     <PlatformTarget>AnyCPU</PlatformTarget>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>12</LangVersion>
   </PropertyGroup>
   
   <ItemGroup Label="Build Tools" Condition="$([MSBuild]::IsOsPlatform('Windows'))">
--- a/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
+++ b/src/client/Microsoft.Identity.Client.Broker/Microsoft.Identity.Client.Broker.csproj
@@ -4,7 +4,6 @@
     <TargetFrameworkNetStandard>netstandard2.0</TargetFrameworkNetStandard>
     <TargetFrameworkNetDesktop Condition="$([MSBuild]::IsOsPlatform('Windows'))">net462</TargetFrameworkNetDesktop>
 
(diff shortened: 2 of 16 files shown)


westin-m (approved):

pmaytak:
I don't quite understand why `latest` mapped to pre-11? (The issue doesn't explain this.) [`latest` is based on the compiler version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#c-language-version-reference), so should map to 12 in this case? If the `LangVersion` is absent, then the TFM dictates the language version.

kellyyangsong:
> I don't quite understand why `latest` mapped to pre-11? (The issue doesn't explain this.) [`latest` is based on the compiler version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#c-language-version-reference), so should map to 12 in this case? If the `LangVersion` is absent, then the TFM dictates the language version.

@pmaytak fair question - using `latest` for the `langversion` setting can be problematic because it can lead to unexpected behavior when new versions of C# are released. for example, if you use `latest` and then upgrade to a new version of .NET that includes a new version of C#, your code may start to behave differently than it did before. this can cause issues if you have code that relies on specific language features or behavior. specifying a specific version, like `12`, can help ensure that the code behaves consistently across different versions of .NET.

pmaytak:
> > I don't quite understand why `latest` mapped to pre-11? (The issue doesn't explain this.) [`latest` is based on the compiler version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#c-language-version-reference), so should map to 12 in this case? If the `LangVersion` is absent, then the TFM dictates the language version.
> 
> @pmaytak fair question - using `latest` for the `langversion` setting can be problematic because it can lead to unexpected behavior when new versions of C# are released. for example, if you use `latest` and then upgrade to a new version of .NET that includes a new version of C#, your code may start to behave differently than it did before. this can cause issues if you have code that relies on specific language features or behavior. specifying a specific version, like `12`, can help ensure that the code behaves consistently across different versions of .NET.

Thanks for clarifying.

pmaytak (approved):

bgavrilMS (approved):

PR #4497, opened 2024-01-11
Low level Linux libsecret operations should throw exceptions

### LINKED ISSUE #4493: Error message when token cache persistence fails could be more descriptive
[This error](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/75dce4e212ae11ecc9e5ed11f40554ff87c9b7f0/src/client/Microsoft.Identity.Client.Extensions.Msal/Storage.cs#L207-L209) 

`"Persistence check failed. Data was written but it could not be read. Possible cause: on Linux, LibSecret is installed but D-Bus isn't running because it cannot be started over SSH."` 

could be more descriptive and accurate as it applies to various platforms. In addition the exception behavior is inconsistent (on Linux we don't throw exceptions, we just log them).

bgavrilMS:
Fixes #4493 

**Testing**

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client.Extensions.Msal/Accessors/LinuxKeyRingAccessor.cs
```diff
@@ -103,11 +104,14 @@ public void Clear()
                 try
                 {
                     GError err = (GError)Marshal.PtrToStructure(error, typeof(GError));
-                    _logger.LogError($"An error was encountered while clearing secret from keyring in the {nameof(Storage)} domain:'{err.Domain}' code:'{err.Code}' message:'{err.Message}'");
+                    throw new InteropException(
```
bgavrilMS, line 107: Note that this exception is used for the MacOS accessor. Most APIs either ignore exceptions.

src/client/Microsoft.Identity.Client.Extensions.Msal/Accessors/LinuxKeyRingAccessor.cs
```diff
@@ -103,11 +104,14 @@ public void Clear()
...
+                        $"An error was encountered while clearing secret from keyring in the {nameof(Storage)} domain:'{err.Domain}' code:'{err.Code}' message:'{err.Message}'", 
+                        err.Code);
                 }
                 catch (Exception e)
                 {
-                    _logger.LogError($"An exception was encountered while processing libsecret error information during clearing secret in the {nameof(Storage)} ex:'{e}'");
+                    throw new InteropException(
+                        $"An exception was encountered while processing libsecret error information during clearing secret in the {nameof(Storage)} ex:'{e}'", 0);
```
gladjohn, line 114: can we just add the exception message but add the full exception to the inner exception?

gladjohn (approved):

trwalke:
I am trying to understand what is happening here so my question is how does this address the lack of clarity mentioned in the original issue in regards to [This error](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/75dce4e212ae11ecc9e5ed11f40554ff87c9b7f0/src/client/Microsoft.Identity.Client.Extensions.Msal/Storage.cs#L207-L209)? Im guessing these new exceptions be thrown instead of the one mentioned above to provide more clarity that way?

bgavrilMS:
> I am trying to understand what is happening here so my question is how does this address the lack of clarity mentioned in the original issue in regards to [This error](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/75dce4e212ae11ecc9e5ed11f40554ff87c9b7f0/src/client/Microsoft.Identity.Client.Extensions.Msal/Storage.cs#L207-L209)? Im guessing these new exceptions be thrown instead of the one mentioned above to provide more clarity that way?

Correct, the accessor should throw meaningful exceptions with details, it should not hide errors. This is what Mac and Windows accessors do today - they bubble up exceptions, but on Linux it looks like we're just logging. 

It's the responsability of higher level APIs in this library to hide exceptions, if needed.

gladjohn (approved):

pmaytak (approved):

PR #4500, opened 2024-01-11
Resolve CS0108 Warning by Explicitly Hiding Base Member in MsalManage…

gladjohn:
This pull request addresses the CS0108 compiler warning encountered in MsalManagedIdentityException.

Fixes #

**Testing**
existing unit, integration

**Performance impact**
none

**Documentation**
- [ ] All relevant documentation is updated. (N/A)

### DIFF
--- a/src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
+++ b/src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
@@ -87,7 +87,7 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// Specifies the managed identity source from which the exception initiates.
         /// </summary>
-        public ManagedIdentitySource ManagedIdentitySource { get; }
+        public new ManagedIdentitySource ManagedIdentitySource { get; }
 
         /// <summary>
         /// Retry error codes specific to managed identity


src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
```diff
@@ -87,7 +87,7 @@ public class MsalManagedIdentityException : MsalServiceException
         /// <summary>
         /// Specifies the managed identity source from which the exception initiates.
         /// </summary>
-        public ManagedIdentitySource ManagedIdentitySource { get; }
+        public new ManagedIdentitySource ManagedIdentitySource { get; }
```
pmaytak, line 90: The way we did this is a bit funky. `MsalException.ManagedIdentitySource` is a const string key into the AdditionalExceptionData but this is of `ManagedIdentitySource` type...

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/be489b50f7941add10166c454bf109fbd602a1fa/src/client/Microsoft.Identity.Client/MsalException.cs#L54C29-L54C50

pmaytak (approved):

trwalke (approved):

src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
bgavrilMS: It's a fair point. `ManagedIdentitySource` is still a public enum btw. But I think our Exception's "AdditionalData" is a string 2 string dictionary, so it kind of makes sense to have a string property here. Do you see a better solution?

bgavrilMS (approved):

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/MsalManagedIdentityException.cs
pmaytak: Both properties are public; so can't change them and we're obsoleting this exception type. So I suppose just keep as is.

PR #4501, opened 2024-01-12
remove non existing maui status solution item

SimonCropp:


bgavrilMS:
Thanks for the contribution @SimonCropp

PR #4503, opened 2024-01-14
pass some missing cancellation tokens

SimonCropp:


src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -170,7 +170,7 @@ private async Task<RegionInfo> DiscoverAsync(ILoggerAdapter logger, Cancellation
             RegionInfo result = null;
             try
             {
-                await _lockDiscover.WaitAsync().ConfigureAwait(false);
+                await _lockDiscover.WaitAsync(requestCancellationToken).ConfigureAwait(false);
```
pmaytak, line 173: Should move this cancellable semaphore outside of the try.

https://stackoverflow.com/questions/72983452/canceling-a-semaphoreslim-waitasynccancellationtoken-does-not-always-throw-ope

pmaytak, line 173: Should move this cancellable semaphore outside of the try.

https://stackoverflow.com/questions/72983452/canceling-a-semaphoreslim-waitasynccancellationtoken-does-not-always-throw-ope

pmaytak (changes requested):

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
SimonCropp: done

pmaytak (approved):

gladjohn (approved):

PR #4504, opened 2024-01-14
elide some asyncs

SimonCropp:
to avoid the perf hit of async state machines

and eliding was already done in other places, so it seem acceptable approach for this codebase

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/NetworkMetadataProvider.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/NetworkMetadataProvider.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
             return discoveryResponse;
         }
 
-        private async Task<InstanceDiscoveryResponse> SendInstanceDiscoveryRequestAsync(
+        private Task<InstanceDiscoveryResponse> SendInstanceDiscoveryRequestAsync(
           Uri authority,
           RequestContext requestContext)
         {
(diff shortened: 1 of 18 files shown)


src/client/Microsoft.Identity.Client/Instance/Discovery/NetworkMetadataProvider.cs
```diff
@@ -72,7 +72,7 @@ private void CacheInstanceDiscoveryMetadata(InstanceDiscoveryResponse instanceDi
             return discoveryResponse;
         }
 
-        private async Task<InstanceDiscoveryResponse> SendInstanceDiscoveryRequestAsync(
+        private Task<InstanceDiscoveryResponse> SendInstanceDiscoveryRequestAsync(
```
bgavrilMS, line 75: Not sure about this one. Lines 79 - 84 could throw exception and the ex behavior is subtly different. See https://blog.stephencleary.com/2016/12/eliding-async-await.html

src/client/Microsoft.Identity.Client/Instance/Validation/AadAuthorityValidator.cs
```diff
@@ -23,7 +23,7 @@ public AadAuthorityValidator(RequestContext requestContext)
         /// because instance metadata is used for aliasing, and authority validation is orthogonal to that. 
         /// MSAL must figure out aliasing even if ValidateAuthority is set to false.
         /// </summary>
-        public async Task ValidateAuthorityAsync(
+        public Task ValidateAuthorityAsync(
```
bgavrilMS, line 26: Same comment as above.

src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/InteractiveRequest.cs
```diff
@@ -70,7 +70,7 @@ internal class InteractiveRequest : RequestBase
         }
         #endregion
 
-        private async Task<MsalTokenResponse> FetchTokensFromBrokerAsync(string brokerInstallUrl, CancellationToken cancellationToken)
+        private Task<MsalTokenResponse> FetchTokensFromBrokerAsync(string brokerInstallUrl, CancellationToken cancellationToken)
         {
             IBroker broker = _serviceBundle.PlatformProxy.CreateBroker(
```
bgavrilMS, line 75: Same

src/client/Microsoft.Identity.Client/Instance/Discovery/NetworkMetadataProvider.cs
SimonCropp: yeah i am aware of the advice. and after years of eliding all my async on all my projects (https://www.nuget.org/profiles/simoncropp), i am yet to be negatively impact by the diff exception handling

also note that the majority of MS codebases also elide all their asyncs

PR #4506, opened 2024-01-14
discard some lambda parameters

SimonCropp:
https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards


src/client/Microsoft.Identity.Client/json/JsonValidatingReader.cs
```diff
@@ -79,7 +79,7 @@ public SchemaScope(JTokenType tokenType, IList<JsonSchemaModel> schemas)
                 _tokenType = tokenType;
                 _schemas = schemas;
 
-                _requiredProperties = schemas.SelectMany<JsonSchemaModel, string>(GetRequiredProperties).Distinct().ToDictionary(p => p, p => false);
+                _requiredProperties = schemas.SelectMany<JsonSchemaModel, string>(GetRequiredProperties).Distinct().ToDictionary(p => p, _ => false);
```
bgavrilMS, line 82: The logic inside the "json" directory is actually a fully blown copy of Newtonsoft.Json library. We didn't want to have an explicit dependency on Newtonsoft because this lib has too many major versions and ppl end up with warnings about dependencies etc. On net6 we use System.Text.Json

Anyway, we'd like to keep the source code in "json" as is, as it is pinned to an exact version of Newtonsoft.

bgavrilMS (changes requested):
Pls undo all changes to files under "json"

pmaytak (approved):

SimonCropp:
json dir reverted

bgavrilMS (approved):

PR #4508, opened 2024-01-14
Update ConfidentialClientApplicationExtensions.cs

SimonCropp:
clientApp must be a ConfidentialClientApplication. if it wasnt it would be a null ref exception


bgavrilMS:
Makes sense, better to get some casting exception than an null ref ex

pmaytak (approved):

gladjohn (approved):

PR #4509, opened 2024-01-14
replace combo of where and FirstOrDefault with only FirstOrDefault

SimonCropp:
less code and faster since it it only on lambda allocation

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithCertTest.cs
+++ b/tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithCertTest.cs
@@ -60,11 +60,11 @@ namespace Microsoft.Identity.Test.Unit
                     // Check presence and value of x5c cert claim.
                     var handler = new JwtSecurityTokenHandler();
                     var jsonToken = handler.ReadJwtToken(encodedJwt);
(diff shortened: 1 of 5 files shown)


bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithCertTest.cs
```diff
@@ -60,11 +60,11 @@ private static MockHttpMessageHandler CreateTokenResponseHttpHandler(bool client
                     // Check presence and value of x5c cert claim.
                     var handler = new JwtSecurityTokenHandler();
                     var jsonToken = handler.ReadJwtToken(encodedJwt);
-                    var x5c = jsonToken.Header.Where(header => header.Key == "x5c").FirstOrDefault();
                     if (expectedX5C != null)
                     {
                         Assert.AreEqual("x5c", x5c.Key, "x5c should be present");
                         Assert.AreEqual(x5c.Value.ToString(), expectedX5C);
+                    var x5c = jsonToken.Header.FirstOrDefault(header => header.Key == "x5c");
```
pmaytak, line 67: Should be moved up to the previous location.

pmaytak (changes requested):

pmaytak (approved):

PR #4534, opened 2024-01-19
add a Guard helper

SimonCropp:


SimonCropp:
@bgavrilMS how do i see the actual compile error?

bgavrilMS (approved):
Love this. Nothing like removing boiler plate code.

CC @neha-bhargava @gladjohn @trwalke for a review

src/client/Microsoft.Identity.Client/Utils/Guard.cs
```diff
@@ -0,0 +1,83 @@
...
+
+#nullable enable
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.Identity.Client.Utils
+{
+    static class Guard
```
bgavrilMS, line 12:
```suggestion
    internal static class Guard
```

bgavrilMS: Explicit is better than implicit...

https://peps.python.org/pep-0020/

CC @rayluo for the inspiration

src/client/Microsoft.Identity.Client/Utils/Guard.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                throw new ArgumentNullException(paramName);
+            }
+
+            return argument;
+        }
+    }
+}
+
+#if NETFRAMEWORK || NETSTANDARD || NETCOREAPP2X
```
bgavrilMS, line 52: not sure if NETCOREAPP2X is relevant? MSAL is compiled against NETSTANDARD and NET6

SimonCropp: sorry. i copied that from one of my projects. will fix

SimonCropp:
@bgavrilMS can u grant me access to the devops project so i can see the build output

SimonCropp:
@pmaytak R9 ?

SimonCropp:
@bgavrilMS can u grant me access to the devops project so i can see the build output

SimonCropp:
@bgavrilMS without access to devops. to see my build errors. it severely limit my capacity to contribute to this project. if access is not possible. you may as well remove me from the project since ti is not worth my time to deal with the friction

bgavrilMS:
Will not take this. Let's revisit when we attempt to use nullable in the library.

PR #4535, opened 2024-01-19
use var where type is obvious

SimonCropp:


bgavrilMS (approved):

pmaytak (approved):

SimonCropp:
so there is no consistency in the existing codebase re var

eg many places use var even when type is not apparent

do u want me to just do a "use var everywhere" approach. since that seems to be the gradual approach moving forward

trwalke (approved):

pmaytak:
> so there is no consistency in the existing codebase re var
> 
> eg many places use var even when type is not apparent
> 
> do u want me to just do a "use var everywhere" approach. since that seems to be the gradual approach moving forward

Yes, we don't have consistency in var use. We should probably match what dotnet repo does. Otherwise, personally, I would take your approach of using the var everywhere unless the type is not clear. Side note - we may need to check/disable analysis rules in the editorconfig; I think currently it warns to use explicit types.

bgavrilMS:
> > so there is no consistency in the existing codebase re var
> > eg many places use var even when type is not apparent
> > do u want me to just do a "use var everywhere" approach. since that seems to be the gradual approach moving forward
> 
> Yes, we don't have consistency in var use. We should probably match what dotnet repo does. Otherwise, personally, I would take your approach of using the var everywhere unless the type is not clear. Side note - we may need to check/disable analysis rules in the editorconfig; I think currently it warns to use explicit types.

+1 

We try to avoid debating too much about code style .... everybody has an opinion and they drag on forever :). So, when in doubt, use what dotnet team use. Which is to permit `var` only when the type is clear.

https://github.com/dotnet/runtime/blob/main/.editorconfig#L47

pmaytak (changes requested):
We should update our root editorconfig with these rules related to var: https://github.com/dotnet/runtime/blob/main/.editorconfig#L47

bgavrilMS:
> We should update our root editorconfig with these rules related to var: https://github.com/dotnet/runtime/blob/main/.editorconfig#L47

Let me know if you want me to do this @SimonCropp . Note that we have 1 top level `.editorconfig` and one special for the `json` directory (as we don't want to do this there).

bgavrilMS:
Let's leave this one out. I think the way to tackle this will be to enable Roslyn analyzer, run the automated fix associated on the entire project and commit without review (there's little point in reviewing automated syntactic changes)

PR #4539, opened 2024-01-21
throw for now iss in uath result

SimonCropp:
since if we dont it is a null ref exception


src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
```diff
@@ -339,6 +339,11 @@ private static void AddPopParams(AuthenticationRequestParameters authenticationR
                 {
                     IdToken idToken = IdToken.Parse(authResult.RawIdToken);
                     authorityUrl = idToken.ClaimsPrincipal.FindFirst("iss")?.Value;
+                    if (authorityUrl == null)
+                    {
+                        throw new Exception("Expected the RawIdToken from the auth result to contain a ClaimsPrincipal named 'iss'.");
```
bgavrilMS, line 344: Well the `iss` claim is mandatory and this code flow is only executed for AAD which respects this.

MSAL doesn't throw generic "Exception", it throws MsalClientException or MsalServiceException. In this case, it's a client exception, but I am not sure how useful it is to an app developer to see it.

Ah well, maybe throw MsalClientException("no_iss_claim", "Id token doesn't have an iss claim in broker flow");

bgavrilMS (changes requested):
Don't throw generic Exception

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs
SimonCropp: is iss populated by this lib? if so i think we should assume it exist. 

so

```
authorityUrl = idToken.ClaimsPrincipal.FindFirst("iss")?.Value;
```

should be 

```
authorityUrl = idToken.ClaimsPrincipal.FindFirst("iss")!.Value;
```

thoughts?

SimonCropp:
@bgavrilMS what exception should be thrown?

bgavrilMS:
> @bgavrilMS what exception should be thrown?

`MsalClientException`

bgavrilMS:
It's acceptable to not throw an pretty exception in this case. closing this PR.

PR #4554, opened 2024-01-22
simplify some string building using raw strings

SimonCropp:


src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -99,39 +99,41 @@ public static string FormatLogMessage(string message, bool piiEnabled, string co
 
         internal static string GetPiiScrubbedExceptionDetails(Exception ex)
         {
-            var sb = new StringBuilder();
-            if (ex != null)
```
SimonCropp, line 103: i inverted this and did a return string empty. so avoid a stringbuilder init

bgavrilMS (approved):

pmaytak (approved):

PR #4559, opened 2024-01-22
Remove Intune Xamarin dev apps

bgavrilMS:
Fixes #

**Testing**

**Performance impact**

**Documentation**


pmaytak (approved):
Any chance that we'd need this for debugging down the road? Maybe we should leave until we formally remove Xamarin targets? Although we do have some docs, so maybe that should be enough.

The doc has 2 links pointing to this sample that should be updated.
https://learn.microsoft.com/en-us/entra/msal/dotnet/how-to/protect-ios-android-mam-intune

bgavrilMS:
> Any chance that we'd need this for debugging down the road? Maybe we should leave until we formally remove Xamarin targets? Although we do have some docs, so maybe that should be enough.
> 
> The doc has 2 links pointing to this sample that should be updated. https://learn.microsoft.com/en-us/entra/msal/dotnet/how-to/protect-ios-android-mam-intune

PR in progress to remove the 2 links. If we need to debug Xamarin, we can simply revert this commit.

gladjohn (approved):

PR #4560, opened 2024-01-22
inline some out variables

SimonCropp:


src/client/Microsoft.Identity.Client/json/Bson/BsonReader.cs
```diff
@@ -496,8 +496,7 @@ private void ReadType(BsonType type)
                     break;
                 }
                 case BsonType.Binary:
-                    BsonBinaryType binaryType;
```
bgavrilMS, line 499: As stated before, we'd like to keep the files in the "json" directory unmodified.

bgavrilMS (changes requested):
No changes in "json" dir

SimonCropp:
json dir fixed

bgavrilMS (approved):

PR #4561, opened 2024-01-23
For proactive refresh log error to otel and change token source to IDP

### LINKED ISSUE #4492: In case of proactive token refresh record telemetry for the background process that refreshes the token.
### Library version used

4.58.1

### .NET version

Net 6

### Scenario

None

When a token is refreshed as part of proactive token refresh, the token is returned to the user from the cache and the token source is recorded as "Cache". The telemetry for the background process is not recorded separately.

### Relevant code snippets

_No response_

### Expected behavior

Record telemetry for the background process as well.

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

neha-bhargava:
Fixes #4492 

Add telemetry for background request for proactive token refresh.
Log error to otel if the background process fails.
Added tags ApiId and IsProactiveRefresh to MsalFailure counter
Added tag CacheRefreshReason to histograms MsalTotalDuration, MsalDurationInL1CacheInUs, MsalDurationInL2Cache

**Testing**
Added unit tests and integration tests

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -24,7 +24,7 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
         /// <summary>
         /// The source of the token in the result.
         /// </summary>
-        public TokenSource TokenSource { get; }
+        public TokenSource TokenSource { get; internal set; }
```
bgavrilMS, line 27: might as well make the setter public, as it will make the object testable.

src/client/Microsoft.Identity.Client/Internal/Requests/SilentRequestHelper.cs
```diff
@@ -101,14 +105,29 @@ internal static bool NeedsRefresh(MsalAccessTokenCacheItem oldAccessToken, out D
...
                 {
                     logger.WarningPiiWithPrefix(ex, ProactiveRefreshCancellationError);
+                    error = ex.GetType().Name;
                 }
                 catch (Exception ex)
                 {
                     logger.ErrorPiiWithPrefix(ex, ProactiveRefreshGeneralError);
+                    error = ex.GetType().Name;
+                }
+                finally
```
bgavrilMS, line 121: I don't think a finally block is needed. Error var is only set in catch clauses. Finally executes for non-error cases for nothing.

Why not have a private method `LogOtelError(error, apidId..)` and call it in every `catch` clause?

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -95,13 +95,20 @@ public OtelInstrumentation()
             AuthenticationResultMetadata authResultMetadata,
             ILoggerAdapter logger)
         {
+            var tokenSource = authResultMetadata.TokenSource;
+
+            //if (authResultMetadata.CacheRefreshReason == CacheRefreshReason.ProactivelyRefreshed)
+            //{
+            //    tokenSource = TokenSource.IdentityProvider;
```
bgavrilMS, line 102: This dead code seems like a nicer way to infer the token source. Can it be used instead of updateing every background request logic?

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -127,7 +134,7 @@ public OtelInstrumentation()
             }
 
             // Only log duration in HTTP when token is fetched from IDP
-            if (s_durationInHttp.Value.Enabled && authResultMetadata.TokenSource.Equals(TokenSource.IdentityProvider))
+            if (s_durationInHttp.Value.Enabled && tokenSource.Equals(TokenSource.IdentityProvider))
```
bgavrilMS, line 137: a.Equals(b) is equivalent to a==b for strings.
Use a.Equals(b, Culture.CaseInsensitiveInvariant) for case insensitive op. But I would think tokenSource will always have the same case so a==b is sufficient?

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
neha-bhargava: I added this one first and then used the other way. Thinking again I think this will be a better way since we only need it as IDP for otel. The only reason I used the current approach is because this way we will send token source as cache in `AuthenticationResultMetadata` but pass it as IDP to otel. The current approach will have it consistent.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -136,26 +143,28 @@ public OtelInstrumentation()
...
+        void IOtelInstrumentation.LogFailedMetrics(string platform, string errorCode, string apiId, bool isProactiveTokenRefresh)
         {
             if (s_failureCounter.Value.Enabled)
             {
                 s_failureCounter.Value.Add(1,
                         new(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion()),
                         new(TelemetryConstants.Platform, platform),
-                        new(TelemetryConstants.ErrorCode, errorCode)); 
+                        new(TelemetryConstants.ErrorCode, errorCode), 
+                        new(TelemetryConstants.ApiId, apiId),
```
pmaytak, line 166: Pass ApiId type everywhere and ToString here only.

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
SimonCropp: it is already testable by passing in TokenSource to the constructor

neha-bhargava: Will be reverting this back and using token source as IDP for proactive token refresh only for client side telemetry through otel.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -136,26 +144,28 @@ public OtelInstrumentation()
...
                 new(TelemetryConstants.Platform, platform),
                 new(TelemetryConstants.ApiId, apiId),
-                new(TelemetryConstants.TokenSource, authResultMetadata.TokenSource),
+                new(TelemetryConstants.TokenSource, tokenSource),
                 new(TelemetryConstants.CacheLevel, cacheLevel));
             }
         }
 
-        void IOtelInstrumentation.LogFailedMetrics(string platform, string errorCode)
+        void IOtelInstrumentation.LogFailedMetrics(string platform, string errorCode, ApiEvent.ApiIds apiId, bool isProactiveTokenRefresh)
```
bgavrilMS, line 159: nit: why did you use explicit interface implemenation? We should default to the non-explicit https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
```diff
@@ -60,7 +61,101 @@ public async Task AcquireTokenOTelTestAsync()
...
+                var refreshOn = TestCommon.UpdateATWithRefreshOn(_cca.AppTokenCacheInternal.Accessor).RefreshOn;
+
+                TokenCacheAccessRecorder cacheAccess = _cca.AppTokenCache.RecordAccess();
+
+                Trace.WriteLine("3. Configure AAD to respond with valid token to the refresh RT flow");
+                _harness.HttpManager.AddAllMocks(TokenResponseType.Valid_ClientCredentials);
+
+                // Act
+                Trace.WriteLine("4. ATS - should perform an RT refresh");
+                AuthenticationResult result = await _cca.AcquireTokenForClient(TestConstants.s_scope)
```
bgavrilMS, line 89: Consider adding unit tests for all the flows that perform background refresh, not just client_credentials:

- refresh_token
- obo
- msi

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
neha-bhargava: Good point, this is not intentional. Might be some intellisense or copilot thing that went unnoticed.

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
```diff
@@ -74,16 +169,15 @@ private async Task AcquireTokenSuccessAsync()
                 .ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
             Assert.IsNotNull(result);
 
-            // Acquire token silently
-            var account = (await _cca.GetAccountsAsync().ConfigureAwait(false)).Single();
-            result = await _cca.AcquireTokenSilent(TestConstants.s_scope, account)
-                .ExecuteAsync().ConfigureAwait(false);
+            // Acquire token from the cache
+            //var account = (await _cca.GetAccountsAsync().ConfigureAwait(false)).Single();
```
trwalke, line 173: dead code?

trwalke (approved):

trwalke:
Isn't this going to skew the cache hit ratio resulting customer concern? Also, while there is a token refresh, the token returned to the user is not the one acquired to from the endpoint. Are we sure this is the right thing to change?

neha-bhargava:
> Isn't this going to skew the cache hit ratio metrics resulting in potential customer concern for those who monitor this? Also, while there is a token refresh, the token returned to the user is not the one acquired to from the endpoint. Are we sure this is the right thing to change?

Yes, I will be changing the token source back to cache, we should be using both token source and cache refresh reason when generating metrics. Thanks

pmaytak (approved):

bgavrilMS:
Hi @neha-bhargava - this has been opened for quite a while. Would you like the team to review again?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ProactiveRefreshTests.cs
```diff
@@ -0,0 +1,105 @@
...
+using OpenTelemetry;
+using OpenTelemetry.Metrics;
+
+namespace Microsoft.Identity.Test.Integration.HeadlessTests
+{
+    [TestClass]
+    public class ProactiveRefreshTests
+    {
+        [TestMethod]
+        public async Task TestProactiveTokenWithTelemetry()
```
pmaytak, line 25:
```suggestion
        public async Task ProactiveRefreshTriggers_WithTelemetry_Test()
```

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
```diff
@@ -60,7 +65,247 @@ public async Task AcquireTokenOTelTestAsync()
                 await AcquireTokenMsalClientExceptionAsync().ConfigureAwait(false);
 
                 s_meterProvider.ForceFlush();
-                VerifyMetrics(5, _exportedMetrics);
+                VerifyMetrics(5, _exportedMetrics, 2, 2);
```
pmaytak, line 68: _exportedMetrics needs to be cleared in TestInitialize?

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
```diff
@@ -60,7 +65,247 @@ public async Task AcquireTokenOTelTestAsync()
...
+
+                Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+                Assert.AreEqual(CacheRefreshReason.NotApplicable, result.AuthenticationResultMetadata.CacheRefreshReason);
+
+                s_meterProvider.ForceFlush();
+                VerifyMetrics(4, _exportedMetrics, 4, 0);
+            }
+        }
+
+        private MockHttpMessageHandler AddMockHandlerAadSuccess(
```
pmaytak, line 197: Duplicate code? I think we already have this in some other test helper. HttpManager maybe.

pmaytak (approved):

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ProactiveRefreshTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                .ExecuteAsync(CancellationToken.None)
+                .ConfigureAwait(false);
+
+            Assert.IsNotNull(authResult);
+            Assert.IsNotNull(authResult.AccessToken);
+            Assert.AreEqual(TokenSource.Cache, authResult.AuthenticationResultMetadata.TokenSource);
+            Assert.AreEqual(CacheRefreshReason.ProactivelyRefreshed, authResult.AuthenticationResultMetadata.CacheRefreshReason);
+            Assert.IsTrue(authResult.AuthenticationResultMetadata.DurationTotalInMs < 50);
+
+            Thread.Sleep(5000);  // Wait for the background process to complete
```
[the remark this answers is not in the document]
neha-bhargava: These are integration tests and we are not waiting for the http manager queue to be empty.
Are you suggesting that we yield for another operation to complete?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ProactiveRefreshTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            authResult = await confidentialApp
+                .AcquireTokenForClient(settings.AppScopes)
+                .ExecuteAsync(CancellationToken.None)
+                .ConfigureAwait(false);
+
+            Assert.IsNotNull(authResult);
+            Assert.IsNotNull(authResult.AccessToken);
+            Assert.AreEqual(TokenSource.Cache, authResult.AuthenticationResultMetadata.TokenSource);
+            Assert.AreEqual(CacheRefreshReason.NotApplicable, authResult.AuthenticationResultMetadata.CacheRefreshReason);
+            Assert.IsTrue(authResult.AuthenticationResultMetadata.DurationTotalInMs < 50);
```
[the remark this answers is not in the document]
neha-bhargava: Added these to keep a check on latencies. Generally the cache read/write is about 0-20 ms for these tests. And the http ones are more than 100ms. There is enough margin. 
@bgavrilMS since we discussed having integration tests for the latencies, do you think these look good?

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ProactiveRefreshTests.cs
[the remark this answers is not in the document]
bgavrilMS: Yes, we should not block for a duration. It's better to do smth like: 

1. wait 10 ms
2. check if exportedMetrics.Count == 4. If not, goto 1

This is what YieldTillSatisfied does.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ProactiveRefreshTests.cs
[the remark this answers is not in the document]
bgavrilMS: I don't have a strong opinion here. 

AAD latency can be quite big when it has problems, so we can't add an assertion for when TokenSource is IdP.

For cached token, with this config, the latency should be way under 1ms.  So 50 isn't going to catch anything, except a major regression. 

How about we leave this in and if the test ever fails we remove it?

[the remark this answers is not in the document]
neha-bhargava: Makes sense. Will remove for IDP and leave for the cache for now.

tests/Microsoft.Identity.Test.Integration.netfx/HeadlessTests/ProactiveRefreshTests.cs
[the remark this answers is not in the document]
neha-bhargava: Updated

PR #4570, opened 2024-01-24
Use net6.0 in GitHub Action benchmarks

pmaytak:


SimonCropp:
why not net8?

pmaytak:
Initially I set the perf project to net8 but there were some issues building the solution related to Xamarin iOS. Meanwhile set it to just net6; using net8 runtime in builds requires a bit more work.

PR #4571, opened 2024-01-24
Adding Claims challenge exception and error messages

### LINKED ISSUE #4496: [Bug] Improve logging and error message when the web api receives claim challenge
### Library version used

4.58.0

### .NET version

all

### Scenario

None

Better logs and error message is needed to guide app developers to implement OBO correctly when it comes to claims challenges, MFA etc. Note that exception messages are not logged by default, so please treat this case. 

See https://portal.microsofticm.com/imp/v3/incidents/incident/457725369/summary for confused app developers 

I propose changes as follows: 

1. If error code is `invalid_grant` and Claims are present and this is an OBO flow If response contains Claims, log and throw non-pii exception with clear message to https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/on-behalf-of-flow#handling-multi-factor-auth-mfa-conditional-access-and-incremental-consent
2. If it's not an OBO flow but Claims are present, add a helper message and a link to https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/exceptions/msal-error-handling#conditional-access-and-claims-challenges

Alternative: create an `MsalClaimsChallengeException` that derives from `MsalUiRequiredException` to make it more clear that Claims are present.

CC @localden @neha-bhargava 

### Relevant code snippets

_No response_

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

trwalke:
Fixes #4496

Adding MsalClaimsChallengeException
Updating error messages for claims challenge with OBO

**Testing**
Unit

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -221,5 +229,18 @@ private static bool IsInvalidGrantSubError(string subError)
 
             return !s_nonUiSubErrors.Contains(subError);
         }
+
+
+        private static string UpdateExceptionForClaimsChallenge(string message, ApiIds apiId)
```
bgavrilMS, line 234: I don't think the exception factory should know about OBO. The OBO flow is better suited to massage the exception message.

Alternatively, provide a general purpose message / link to docs where you can clarify how Claims are handled: 

- in a web app, challenge the user. Ensure the challenge contains the claims, for example by including the Claims string in the authorization URI as a parameter
- in a web api .... 
- in client_credentials and managed identity use WithClaims
- in desktop / mobile apps, call AcquireTokenSilent / AcquireTokenInteractive and change them WithClaims

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -34,6 +37,11 @@ internal class MsalServiceExceptionFactory
                 {
                     ex = new MsalUiRequiredException(errorCode, MsalErrorMessage.AadThrottledError, innerException);
                 }
+                else if (oAuth2Response.Claims != null)
```
bgavrilMS, line 40: I don't see why the throttled exception can't have Claims. It looks like just the error message is different.

tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -161,6 +161,7 @@ public static HashSet<string> s_scope
...
         public const string ClaimWithAdditionalKeyAndAccessKey = @"{""some_other_key"":{""nbf"":{""essential"":true,""value"":""1701477303""},""additional_claim"":{""key"":""value""},""new_claim"":{""new_key"":""new_value""}},""access_token"":{""nbf"":{""essential"":true,""value"":""1701477303""},""additional_claim"":{""key"":""value""},""new_claim"":{""new_key"":""new_value""}}}";
         public const string MergedJsonClaimWithAdditionalKeyAndAccessKey = @"{""access_token"":{""xms_cc"":{""values"":[""cp1"",""cp2""]},""nbf"":{""essential"":true,""value"":""1701477303""},""additional_claim"":{""key"":""value""},""new_claim"":{""new_key"":""new_value""}},""some_other_key"":{""nbf"":{""essential"":true,""value"":""1701477303""},""additional_claim"":{""key"":""value""},""new_claim"":{""new_key"":""new_value""}}}";
+        public const string GenericClaims = "some_claims";
```
bgavrilMS, line 164: Better to have this in JSON format.

src/client/Microsoft.Identity.Client/MsalClaimsChallengeException.cs
```diff
@@ -0,0 +1,107 @@
...
+using static Microsoft.Identity.Client.TelemetryCore.Internal.Events.ApiEvent;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// This exception indicates that claims have been returned from the endpoint that need to be resolved. This can be achieved by using them in the next authentication request.
+    /// See https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/exceptions/msal-error-handling#conditional-access-and-claims-challenges
+    /// See https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/on-behalf-of-flow#handling-multi-factor-auth-mfa-conditional-access-and-incremental-consent
+    /// </summary>
+    internal class MsalClaimsChallengeException : MsalServiceException
```
bgavrilMS, line 19: Why internal?

bgavrilMS (commented):
See if you can decouple OBO logic out of the ExceptionFactory (Single Responsability Principle)
MsalClaimsChallengeException should be public?

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
trwalke: Decided to go with a generic message to remove OBO coupling

tests/Microsoft.Identity.Test.Common/TestConstants.cs
trwalke: The claims challenge comes in as a encoded JSON value rather than a direct JSON string.
See https://learn.microsoft.com/en-us/entra/identity-platform/claims-challenge?tabs=dotnet#claims-challenge-header-format

bgavrilMS (approved):

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -30,13 +30,21 @@ internal class MsalServiceExceptionFactory
...
             {
+                var errorMessageToUse = errorMessage;
+
                 if (IsThrottled(oAuth2Response))
                 {
-                    ex = new MsalUiRequiredException(errorCode, MsalErrorMessage.AadThrottledError, innerException);
+                    errorMessageToUse = MsalErrorMessage.AadThrottledError;
+                }
+
+                if (oAuth2Response.Claims != null)
```
SimonCropp, line 40: invert the if. less negation is easier to reason about

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -457,5 +457,6 @@ public static string InvalidTokenProviderResponseValue(string invalidValueName)
...
         public const string JsonParseErrorMessage = "There was an error parsing the response from the token endpoint, see inner exception for details. Verify that your app is configured correctly. If this is a B2C app, one possible cause is acquiring a token for Microsoft Graph, which is not supported. See https://aka.ms/msal-net-up";
         public const string SetCiamAuthorityAtRequestLevelNotSupported = "Setting the CIAM authority (ex. \"{tenantName}.ciamlogin.com\") at the request level is not supported. The CIAM authority must be set during application creation";
+        public const string ClaimsChallenge = "The error returned contains a claims challenge. For additional info on how to handle claims related to multi-factor auth, conditional access and incremental consent, please see https://aka.ms/msal-conditional-access-claims. If you are using the On-behalf-of flow, please see https://aka.ms/msal-conditional-access-claims-obo";
```
pmaytak, line 460:
```suggestion
        public const string ClaimsChallenge = "The error returned contains a claims challenge. For additional info on how to handle claims related to multifactor authentication, conditional access, and incremental consent, see https://aka.ms/msal-conditional-access-claims. If you are using the On-Behalf-Of flow, see https://aka.ms/msal-conditional-access-claims-obo";

```

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -30,13 +30,21 @@ internal class MsalServiceExceptionFactory
...
+                {
+                    errorMessageToUse = UpdateExceptionForClaimsChallenge(errorMessageToUse);
+                    ex = new MsalClaimsChallengeException(errorCode, errorMessageToUse, innerException);
```
pmaytak, line 43: What is the reason for creating a new exception class for this? Is it actually needed?

I feel like it's cleaner and simpler if we keep the current exceptions and just add an extra message with aka.ms links to it. 

[Here](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/on-behalf-of-flow#the-web-api-replies-with-401--www-authenticate) we say MsalUiRequiredEx has the Claims property and [here](https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/exceptions/msal-error-handling#conditional-access-and-claims-challenges, it's MsalServiceEx with claims and now we add a third exception type. I think people will still catch the parent level exception anyway.

I see that the new exception is more specific, I just wonder if the increase in the public API is a net benefit, compared to just adding to the error message.

trwalke: This scenario seems to be a huge pain for a lot of customers so the goal is to try to be as specific as possible. Adding a new message and updating documentation will be a stronger indication to customers that this is not just another service exception, but something that should be acted on. While it is a bit redundant, it may make things more apparent for some customers. 

Thoughts @bgavrilMS @localden ?

bgavrilMS: It's a fair point. There are 2 reasons for this: 

1. "UIRequiredException" is not a good name because this also occurs for S2S (and soon MSI) as well. It's wierd to get "UiRequiredException" for AcquierTokenForClient or AcquireTokenForMsi

2. The more obvs type name should (hopefully) encourage ppl to use the Claims. Note that ADAL had the same concept (AdalClaimsChallengeException).

gladjohn: We just removed `MSALManagedIdentityException` as it was not a good fit, and ended up using `MsalServiceException`, I understand that claims challenge in S2S scenario does not fit into both `Service` or `UiRequired` exceptions, but instead of adding a new exception class just for claims, would be nice to have a more generalized exception class for continuous access evaluation (CAE) or conditional access (CA) or any other such cases in the future.

bgavrilMS: @trwalke - let's discuss this with the team and @localden - and maybe create a simple class diagram to show the inheritance chain. The reason we removed `MSALManagedIdentityException` is that it was complicating the inheritance chain with a new branch.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -457,6 +457,7 @@ public static string InvalidTokenProviderResponseValue(string invalidValueName)
...
         public const string SetCiamAuthorityAtRequestLevelNotSupported = "Setting the CIAM authority (ex. \"{tenantName}.ciamlogin.com\") at the request level is not supported. The CIAM authority must be set during application creation";
-        public const string ClaimsChallenge = "The error returned contains a claims challenge. For additional info on how to handle claims related to multi-factor auth, conditional access and incremental consent, please see https://aka.ms/msal-conditional-access-claims. If you are using the On-behalf-of flow, please see https://aka.ms/msal-conditional-access-claims-obo";
+        public const string ClaimsChallenge = "The error returned contains a claims challenge. For additional info on how to handle claims related to multifactor authentication, conditional access, and incremental consent, see https://aka.ms/msal-conditional-access-claims. If you are using the On-Behalf-Of flow, see https://aka.ms/msal-conditional-access-claims-obo";
```
localden, line 460:
```suggestion
        public const string ClaimsChallenge = "The returned error contains a claims challenge. For additional info on how to handle claims related to multifactor authentication, Conditional Access, and incremental consent, see https://aka.ms/msal-conditional-access-claims. If you are using the On-Behalf-Of flow, see https://aka.ms/msal-conditional-access-claims-obo.";
```

localden (changes requested):

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
localden: @trwalke @bgavrilMS let's discuss this during the Customer Trust sync this Thursday.

src/client/Microsoft.Identity.Client/Internal/Logger/LoggerHelper.cs
```diff
@@ -126,6 +126,11 @@ internal static string GetPiiScrubbedExceptionDetails(Exception ex)
                     sb.AppendLine("=== End of inner exception stack trace ===");
                 }
 
+                if (ex.Message.Contains(MsalErrorMessage.ClaimsChallenge))
```
pmaytak, line 129:
```suggestion
                if (ex is MsalClaimsChallengeException))
```

src/client/Microsoft.Identity.Client/MsalClaimsChallengeException.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
+using static Microsoft.Identity.Client.TelemetryCore.Internal.Events.ApiEvent;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// This exception indicates that claims have been returned from the endpoint that need to be resolved. This can be achieved by using them in the next authentication request.
+    /// See https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/exceptions/msal-error-handling#conditional-access-and-claims-challenges
+    /// See https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/on-behalf-of-flow#handling-multi-factor-auth-mfa-conditional-access-and-incremental-consent
```
pmaytak, line 17: These can be put into an href, so they're clickable.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/57e840463445e5760b49388a5b3ba0af27eb94e2/src/client/Microsoft.Identity.Client/LogLevel.cs#L8

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
pmaytak: There should be a space between a link and a period. `https://aka.ms/msal-conditional-access-claims-obo .` Otherwise, some pages (like the API reference) end up rendering the period as part of the link.

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
```diff
@@ -30,13 +30,21 @@ internal class MsalServiceExceptionFactory
 
             if (IsInvalidGrant(oAuth2Response?.Error, oAuth2Response?.SubError) || IsInteractionRequired(oAuth2Response?.Error))
             {
+                var errorMessageToUse = errorMessage;
+
                 if (IsThrottled(oAuth2Response))
                 {
-                    ex = new MsalUiRequiredException(errorCode, MsalErrorMessage.AadThrottledError, innerException);
+                    errorMessageToUse = MsalErrorMessage.AadThrottledError;
```
pmaytak, line 37: Set the `errorMessage` and reuse? 
```suggestion
                    errorMessage = MsalErrorMessage.AadThrottledError;
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/LoggerTests.cs
```diff
@@ -369,6 +370,44 @@ public async Task NullExternalMsalLoggerTestAsync()
 
         }
 
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        public async Task ClaimsChallengeErrorLogging_TestAsync(bool piiLogging)
```
pmaytak, line 376: nit: Don't need to add Async to tests anymore.

```suggestion
        public async Task ClaimsChallengeErrorLogging_Test(bool piiLogging)
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/LoggerTests.cs
```diff
@@ -369,6 +370,44 @@ public async Task NullExternalMsalLoggerTestAsync()
...
+                TestIdentityLogger testLogger = new TestIdentityLogger();
+                StringBuilder stringBuilder;
+
+                var appBuilder = ConfidentialClientApplicationBuilder
+                    .Create(TestConstants.ClientId)
+                    .WithClientSecret("secret")
+                    .WithHttpManager(httpManager);
+
+                    stringBuilder = testLogger.StringBuilder;
+                    appBuilder.WithExperimentalFeatures()
```
pmaytak, line 389: Not needed?

pmaytak (approved):
Approved with suggestions.

bgavrilMS:
@trwalke - do you plan to discuss this further? I think we should proceed with updating the message and to leave out the new exception, as that seems to be controversial.

src/client/Microsoft.Identity.Client/MsalClaimsChallengeException.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// </summary>
+    public class MsalClaimsChallengeException : MsalServiceException
```
neha-bhargava, line 19:
```suggestion
    public class MsalClaimsChallengeException : MsalUiRequiredException
```

neha-bhargava (changes requested):
As discussed, inherit from `MsalUiRequiredException` to avoid breaking changes.

src/client/Microsoft.Identity.Client/MsalClaimsChallengeException.cs
localden: Also, as @pmaytak mentioned in the meeting, remove the `/en-us/` identifiers - locales are auto-inferred when the customer navigates to the page.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
localden: Updated the comment @pmaytak - good catch.

localden (changes requested):

src/client/Microsoft.Identity.Client/MsalServiceExceptionFactory.cs
trwalke: Why should I do this? errorMessage  is an incoming parameter

PR #4586, opened 2024-01-29
reduce allocs in TryCreateDeviceAuthChallengeResponse

SimonCropp:
* avoids a double call to `DeviceAuthJWTResponse.GetResponseToSign` which does serializarion, concat and encoding
 * changes to FormatResponseHeader to avoid 2 allocs


bgavrilMS (approved):
This is nice, but not in the hot path. 

If you want to have a look at the host path, check out https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs#L30

particularly the FindToken and SaveToken methods.

bgavrilMS (approved):

trwalke (approved):

gladjohn (approved):

PR #4589, opened 2024-01-29
Fix some unit tests which did not declare their dependency on HTTP

bgavrilMS:
To repro: 

1. Disconnect from Internet
2. Run affected UT

Actual: test fails with Socket exception (the instance discovery fails)
Expected: unit tests should not depend on HTTP

### DIFF
--- a/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
+++ b/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
@@ -15,6 +15,7 @@ using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Http;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NSubstitute;
 
 namespace Microsoft.Identity.Test.Common.Core.Mocks
 {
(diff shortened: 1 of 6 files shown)


tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
```diff
@@ -71,6 +72,15 @@ public MockHttpMessageHandler AddMockHandler(MockHttpMessageHandler handler)
 
         public int QueueSize => _httpMessageHandlerQueue.Count;
 
+        /// <summary>
+        /// For use only in threads that spin many threads. Not thread safe.
```
bgavrilMS, line 76:
```suggestion
        /// For use only in tests that spin many threads. Not thread safe.
```

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Http;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NSubstitute;
```
bgavrilMS, line 18:
```suggestion
```

PR #4590, opened 2024-01-31
improve perf for claims parsing

SimonCropp:
so the previous code used `ClaimsPrincipal.FindFirst`

which is effectively

```
foreach (var identity in claimsPrincipal.Identities)
{
    foreach (Claim claim in identity.Claims)
    {
        if (string.Equals(claim.Type, type, StringComparison.OrdinalIgnoreCase))
        {
            return claim
        }
    }
}
```

but we already know the Principal will only have one Identity. and we know the claims of that Identity will be the `claims` variable

so just do a lookup on  `claims` avoiding the nested looping. 

also less code and more readable code


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/IdToken.cs
```diff
@@ -79,6 +80,28 @@ public string GetUniqueId()
...
+                PreferredUsername = FindClaim(claims, IdTokenClaim.PreferredUsername),
+                Name = FindClaim(claims, IdTokenClaim.Name),
+                Email = FindClaim(claims, IdTokenClaim.Email),
+                Upn = FindClaim(claims, IdTokenClaim.Upn),
+                GivenName = FindClaim(claims, IdTokenClaim.GivenName),
+                FamilyName = FindClaim(claims, IdTokenClaim.FamilyName)
+            };
+
+            static string FindClaim(List<Claim> claims, string type) =>
+                claims.SingleOrDefault(_ => string.Equals(_.Type, type, StringComparison.OrdinalIgnoreCase))?.Value;
```
SimonCropp, line 102: @bgavrilMS do u want a SingleOrDefault or a FirstOrDefault

i guess my real question is hos should duplicate claim types be handled when parsing?

bgavrilMS: Good question. An ID Token must be a JWT token https://openid.net/specs/openid-connect-core-1_0.html#IDToken and in a JWT token the claim names must be unique https://www.rfc-editor.org/rfc/rfc7519.html#page-8

So I think `SingleOrDefault` is appropriate, i.e. it's better to fail in case a duplicate claim exists.

trwalke (approved):

PR #4592, opened 2024-01-31
elide OnBefore

SimonCropp:
given it is the fast path, and they are relatively simple methods, is it worth eliding them?

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCache.Notifications.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCache.Notifications.cs
@@ -41,14 +41,16 @@ namespace Microsoft.Identity.Client
                 (this as ITokenCacheInternal).IsAppSubscribedToSerializationEvents();
         }
 
-        async Task ITokenCacheInternal.OnAfterAccessAsync(TokenCacheNotificationArgs args)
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client/TokenCache.Notifications.cs
```diff
@@ -41,28 +41,31 @@ bool ITokenCacheInternal.IsExternalSerializationConfiguredByUser()
                 (this as ITokenCacheInternal).IsAppSubscribedToSerializationEvents();
         }
 
-        async Task ITokenCacheInternal.OnAfterAccessAsync(TokenCacheNotificationArgs args)
+        Task ITokenCacheInternal.OnAfterAccessAsync(TokenCacheNotificationArgs args)
         {
             AfterAccess?.Invoke(args);
```
bgavrilMS, line 46: `AfterAccess` calls a user defined callback, which can error out. I am not sure it's worth testing all implications for this potential perf gain.

bgavrilMS (changes requested):
Code before elided method can throw exception, and removal of `async` changes the behavior of exceptions. Not sure testing this is worth the ROI

PR #4594, opened 2024-01-31
better perf in MergeDifferentEntries

SimonCropp:
fast path

removes a tolist alloc and a lambda alloc


src/client/Microsoft.Identity.Client/Utils/EnumerableExtensions.cs
```diff
@@ -58,13 +58,18 @@ internal static bool ContainsOrdinalIgnoreCase(this IEnumerable<string> set, str
             this IDictionary<TKey, TValue> source,
             IDictionary<TKey, TValue> other)
         {
-            other?.ToList().ForEach(x =>
+            if (source == null)
```
pmaytak, line 61:
```suggestion
            if (source == null || other == null)
```

bgavrilMS:
Btw, we already have a Performance test project setup which runs Benchmark.NET - feel free to reuse it.

SimonCropp:
>  I have been bitten by linq many times...

but this removes the linq

bgavrilMS (approved):

PR #4613, opened 2024-02-05
#4605 Only include needed packages

### LINKED ISSUE #4605: Only include dependencies not provided by the framework
### Summary
The nuget packages should only have dependencies on libraries which are not provided by the framework

### Proposal
The System.Io.FileSystem.AccessControl is conditionally added based on target framework so that it is not included for the newer frameworks

thompson-tomo:
Fixes #4605 

- Add net 6 as TFM for Microsoft.Identity.Client.Extensions.Msal
- Make System.IO.FileSystem.AccessControl only included for Net standard 2.0 in Microsoft.Identity.Client.Extensions.Msal

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
+++ b/src/client/Microsoft.Identity.Client.Extensions.Msal/Microsoft.Identity.Client.Extensions.Msal.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Label="NuGet and AssemblyInfo metadata">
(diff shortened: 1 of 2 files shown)


bgavrilMS:
Hi @thompson-tomo - I understand your approach now and thank you for the time to create this. We don't want to add a new TFM to this package - it'll increase build time and we'll have to do some testing on it etc. The dependency is low risk as it's owned by the .NET team. 

Aside: I don't understand the need to remove System.Diagnostics.DiagnosticSource. This is used for MSAL to emit some metrics through Open Telemetry

thompson-tomo:
@bgavrilMS in relation to System.Diagnostic.DiagnosticSource the new target framework's natively provide the library hence the explicit package reference is not needed.

bgavrilMS:
> @bgavrilMS in relation to System.Diagnostic.DiagnosticSource the new target framework's natively provide the library hence the explicit package reference is not needed.
> 
> For the additional TFM to eliminate an explicit dependency, I wonder how much testing would be needed given said package is extremely likely to be the same as the provided by the framework. This is the case based on the fact that the package has not been updated in quite a whil3.

Ok, you make a good point. Can you please leave just the TFM changes in there, i.e. do not include the System.Diagnostics removal?

bgavrilMS:
@neha-bhargava - a heads up for System.Diagnostics.

thompson-tomo:
@bgavrilMS change made as requested and have been pushed.

bgavrilMS (approved):

pmaytak (approved):

PR #4614, opened 2024-02-06
Update installEdgeDriver.ps1 to fetch Edge version dynamically

gladjohn:
**Fixes #**
OneBranch build failure

**Changes proposed in this request :** 
gets the edge version of the build machine and uses the version to download the edge driver.

**Testing :** 
builds and local powershell

**Performance impact :** 
none

**Documentation :** 
n/a


build/installEdgeDriver.ps1
```diff
@@ -1,4 +1,7 @@
-$url = "https://msedgedriver.azureedge.net/119.0.2151.72/edgedriver_win64.zip" #Chrome Driver from https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/
+# Get the installed version of Microsoft Edge
+$edgeVersion = (Get-Item "C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe").VersionInfo.ProductVersion
```
bgavrilMS, line 2: Can you error out with some message if this is null?

bgavrilMS, line 2: Use %PROGRAMFILES(x86)% to avoid hardcoding ?

build/installEdgeDriver.ps1
```diff
@@ -1,4 +1,7 @@
...
+
+$url = "https://msedgedriver.azureedge.net/$edgeVersion/edgedriver_win64.zip" #Chrome Driver from https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/
```
bgavrilMS, line 4:
```suggestion
$url = "https://msedgedriver.azureedge.net/$edgeVersion/edgedriver_win64.zip" #Edge Driver from https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/
```

build/installEdgeDriver.ps1
gladjohn: added error checks

pmaytak (approved):

Issue #4187, opened 2023-06-26
[Feature Request] Expose CorrelationId from MsalException 


KurzyukovAndrey:
**Problem**
We need to meter `CorrelationId` in case of acquiring a token. However, we cannot retrieve `CorrelationId` from [MsalException](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identity.client.msalexception?view=msal-dotnet-latest). 

**Use case**
The simplified code of usage is shown below: 

	public async Task<AuthenticationResult> AcquireTokenForClientAsync(
		IEnumerable<string> scopes, 
		Action<AcquireTokenForClientParameterBuilder>? configuration, 
		CancellationToken cancellationToken)
	{
		var acquireTokenBuilder = _confidentialClientApplication.AcquireTokenForClient(scopes);

		// WithCorrelationId might be called here.
		configuration?.Invoke(acquireTokenBuilder);

		try
		{
			var authenticationResult = await acquireTokenBuilder
				.ExecuteAsync(cancellationToken)
				.ConfigureAwait(false);

			// authenticationResult exposes correlationId.
		}
		catch (MsalException ex)
		{
			// MsalException does not expose correlationId. 
			// Cannot use correlationId for metering.
		}
	}

Our consumers call [WithCorrelationId](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identity.client.baseabstractacquiretokenparameterbuilder-1.withcorrelationid?view=msal-dotnet-latest#microsoft-identity-client-baseabstractacquiretokenparameterbuilder-1-withcorrelationid(system-guid)) using configuration parameter. Then, we acquire a token. In case of success, we can retrieve `CorrelationId` from [AuthenticationResult.CorrelationId](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identity.client.authenticationresult.correlationid?view=msal-dotnet-latest#microsoft-identity-client-authenticationresult-correlationid). But we cannot retrieve it from [MsalException](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identity.client.msalexception?view=msal-dotnet-latest).

**Proposal**
Expose `CorrelationId` for all exceptions.

**Additional context**
[...]

PR #4649, opened 2024-02-28
Adding correlation id to MSAL Exception

trwalke:
Fixes #
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4187

Adding correlation id to `MsalException` class

**Testing**
Unit

**Performance impact**

**Documentation**

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionTests.cs
@@ -6,6 +6,7 @@ using System.Data.SqlClient;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Threading.Tasks;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Http;
 using Microsoft.Identity.Client.Internal;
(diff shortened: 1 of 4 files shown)


tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionTests.cs
```diff
@@ -432,6 +433,24 @@ public void ExceptionsPropertiesHavePublicSetters()
...
+        public async Task CorrelationIdInExceptions()
+        {
+            var ex = await AssertException.TaskThrowsAsync<MsalUiRequiredException>(async () =>
+                {
+                    var app = PublicClientApplicationBuilder
+                    .Create(TestConstants.ClientId)
+                    .WithDefaultRedirectUri()
+                    .Build();
+
+                    await app.AcquireTokenSilent(TestConstants.s_graphScopes, TestConstants.s_user).ExecuteAsync().ConfigureAwait(false);
```
bgavrilMS, line 446: Can we have a test where you add a "WithCorrelationId(xyz)" and match that up?

tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionTests.cs
```diff
@@ -432,6 +433,24 @@ public void ExceptionsPropertiesHavePublicSetters()
             AssertPropertyHasPublicGetAndSet(typeof(MsalServiceException), "CorrelationId");
         }
 
+        [TestMethod]
+        public async Task CorrelationIdInExceptions()
+        {
+            var ex = await AssertException.TaskThrowsAsync<MsalUiRequiredException>(async () =>
```
bgavrilMS, line 439: This is fine, but MsalUIRequiredException already had this correlation ID. 

How about adding (or changing) a test that results in a real MsalClientException? Maybe this one - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs#L158

bgavrilMS (changes requested):
Have a test that ends up in MsalClientException

tests/Microsoft.Identity.Test.Unit/ExceptionTests/MsalExceptionTests.cs
trwalke: Yeah, I was looking for a way to trigger the client exception. this will work. thanks

bgavrilMS (approved):

PR #4659, opened 2024-03-12
SsoPolicy support for PCA.

ashok672:
Add support for SsoPolicy enforcement when broker is not used. This is only needed for Microsoft 1st party applications.

1. Introduce WithSsoPolicy in PCABuilder.
2. Append the appropriate SsoPolicy headers to authorize and token endpoints.

**Testing**
Tested using the NetCoreWinFormsWAM app by adding .WithSsoPolicy to embedded web view option.

**Performance impact**
There will be an WinRT API call now for each token and authorize endpoint calls. Performance impact should be negligible.

**Documentation**


src/client/Microsoft.Identity.Client/Properties/InternalsVisibleTo.cs
```diff
@@ -6,6 +6,7 @@
 
 [assembly: InternalsVisibleTo("Microsoft.Identity.Client.Desktop" + KeyTokens.MSAL)]
 [assembly: InternalsVisibleTo("Microsoft.Identity.Client.Broker" + KeyTokens.MSAL)]
+[assembly: InternalsVisibleTo("Microsoft.Identity.Client.SsoPolicy" + KeyTokens.MSAL)]
```
bgavrilMS, line 9: Oh, interesting. You want to create a new DLL / package for this? Is there a reason why you can't reuse the "Broker" DLL / package for this ?

src/client/Microsoft.Identity.Client/Internal/SsoPolicy/ISsoPolicy.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+
+namespace Microsoft.Identity.Client.Internal.SsoPolicy
+{
+    internal interface ISsoPolicy
```
bgavrilMS, line 7: empty interfaces are strange :)

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeSsoPolicy/RuntimeSsoPolicy.cs
```diff
@@ -0,0 +1,190 @@
...
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Cache;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Internal.SsoPolicy;
+using Microsoft.Identity.Client.NativeInterop;
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.Platforms.Features.RuntimeSsoPolicy
+{
+    internal class RuntimeSsoPolicy : ISsoPolicy
```
bgavrilMS, line 17: There's a lot of duplicate code with the broker package....

src/client/Microsoft.Identity.Client.SsoPolicy/Microsoft.Identity.Client.SsoPolicy.csproj
```diff
@@ -0,0 +1,45 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
ashok672: I wanted this be separate not knowing all the requirements. If you think reusing broker will have not any issues, I think thats a better path. I will move this to broker pacakge.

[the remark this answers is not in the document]
ashok672: Also I considered this point initially. SSOPolicy is only applicable only when not using broker so putting this in broker might be confusing.  Atleast the way we named it is confusing.

src/client/Microsoft.Identity.Client/Internal/SsoPolicy/ISsoPolicy.cs
ashok672: this was draft and anways the new approach is different reusing the same IBroker interface.

ashok672:
I am still trying to figure out how to add extra headers for System WebView. Apparently based on my initial research there is no way to add headers when launching the default browser with an Uri. I will investigate more. 
WebView 2 is also not done yet. I will get to that soon.

bgavrilMS:
> I am still trying to figure out how to add extra headers for System WebView. Based on my initial research, apparently there is no way to add headers when launching the default browser with an Uri. I will investigate more. WebView 2 is also not done yet. I will get to that soon.

@ashok672 

- WebView2 is not usable with AAD authorities. We disabled it completely because we couldn't get the SSO to work (and the future is WAM, so we stopped investigating). 
- There is no way to inject headers in a system browser. 

I would focus on getting WebView1 to work.

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -47,29 +47,55 @@ public static PublicClientApplicationBuilder WithBrokerPreview(this PublicClient
...
 
-        private static void AddRuntimeSupportForWam(PublicClientApplicationBuilder builder)
+        /// <summary>
+        /// Use this API to enable SsoPolicy enforcement when not using broker.
+        /// 
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <returns>A <see cref="PublicClientApplicationBuilder"/> from which to set more
+        /// parameters, and to create a public client application instance</returns>
+        public static PublicClientApplicationBuilder WithSsoPolicy(this PublicClientApplicationBuilder builder)
```
bgavrilMS, line 63: @localden @iulico-1 for public API.

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -47,29 +47,55 @@ public static PublicClientApplicationBuilder WithBrokerPreview(this PublicClient
...
+                             return new RuntimeBroker(uiParent, appConfig, logger);
+                         };
+                }
+                else
+                {
+                    builder.Config.BrokerCreatorFunc =
+                         (uiParent, appConfig, logger) =>
+                         {
+                             logger.Info("[Runtime] SsoPolicy supported OS.");
+                             return new RuntimeBroker(appConfig, logger);
```
bgavrilMS, line 89: Why did you need to create a new ctor? The ctor does not validate the widown handle (uiParent), only `broker.AcquireTokenInteractive` does.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -444,6 +444,11 @@ internal override void Validate()
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (Config.IsBrokerEnabled && Config.IsSsoPolicyEnabled)
+            {
+                throw new InvalidOperationException(MsalErrorMessage.BrokerAndSsoPolicyConfigurationError);
```
bgavrilMS, line 450: Please throw `MsalClientException` exception instead.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -407,6 +407,8 @@ public static string CertMustBeRsa(string certificateFriendlyName)
 
         public const string MultiCloudSupportUnavailable = "Multi cloud support unavailable with broker.";
 
+        public const string BrokerAndSsoPolicyConfigurationError = "WithBroker and WithSsoPolicy cannot be used together. Broker by default supports SsoPolicy. Use WithSsoPolicy only when using embedded webview or system webview";
```
bgavrilMS, line 410:
```suggestion
        public const string BrokerAndSsoPolicyConfigurationError = "WithBroker and WithSsoPolicy cannot be used together. Broker by default supports SsoPolicy. Use WithSsoPolicy only when using embedded webview";

```

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
```diff
@@ -80,5 +80,10 @@ public Task<MsalTokenResponse> AcquireTokenByUsernamePasswordAsync(Authenticatio
             _logger.Info("NullBroker - returning null on ROPC request.");
             return Task.FromResult<MsalTokenResponse>(null);
         }
+
+        public Dictionary<string, string> GetSsoPolicyHeaders()
+        {
+            return new Dictionary<string, string>();
```
bgavrilMS, line 86: Creating a new empty object each time is not good for perf. See `CollectionHelpers.GetEmptyDictionary` instead.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
```diff
@@ -108,6 +110,7 @@ public Task<OidcMetadata> DiscoverOidcMetadataAsync(Uri endpoint, RequestContext
             {
                 AddCommonHeaders(requestContext);
             }
+            AddSsoPolicyHeaders(requestContext);
```
bgavrilMS, line 113: This is too deep / affects too many flows. OAuth2Client knows about headers and other HTTP concepts, but not specifically about "SSO policy headers". 

Let's think of a way to set this header only for public client flows.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -66,6 +66,8 @@ public string ClientVersion
 
         public bool IsBrokerEnabled { get; internal set; }
 
+        public bool IsSsoPolicyEnabled { get; internal set; }
```
bgavrilMS, line 69: Naming: `SSO` is quite overloaded term. Maybe `IsWebviewSsoPolicyEnabled` ? And a comment that this affects only PCA.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -431,13 +433,32 @@ protected Task ResolveAuthorityAsync()
...
+            InjectPcaSsoPolicyHeader(tokenClient);
+
             return tokenClient.SendTokenRequestAsync(
                 additionalBodyParameters,
                 scopes,
                 tokenEndpoint,
                 cancellationToken);
         }
 
+        private void InjectPcaSsoPolicyHeader(TokenClient tokenClient)
```
bgavrilMS, line 445: @ashok672 - I played a bit with the code and I think this is better place for the injection of the header. Let me know what you think. If you disagree, just revert my commit and we can discuss alternatives. 

OAuth2Client is too much in the hot path.

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -47,29 +47,55 @@ public static PublicClientApplicationBuilder WithBrokerPreview(this PublicClient
...
+        public static PublicClientApplicationBuilder WithSsoPolicy(this PublicClientApplicationBuilder builder)
+        {
+            AddRuntimeSupport(builder, false);
+            builder.Config.IsSsoPolicyEnabled = true;
+            return builder;
+        }
+
+        private static void AddRuntimeSupport(PublicClientApplicationBuilder builder, bool isWamBrokerScenario)
         {
             if (DesktopOsHelper.IsWin10OrServerEquivalent())
```
gladjohn, line 72: So we only care about 1P apps on Windows?

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -552,6 +564,37 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
...
+                byte[] bytes = System.Text.Encoding.UTF8.GetBytes(subStatusValue);
+                string base64UrlEncoded = System.Convert.ToBase64String(bytes)
+                                                        .Replace('+', '-')
+                                                        .Replace('/', '_')
+                                                        .TrimEnd('=');
+                ssoPolicyHeaders.Add("x-ms-SsoFlagsSubstatus", base64UrlEncoded);
+            }
+            else if (ssoPolicy._ssoPolicyType == SsoPolicyType.Unknown)
+            {
+                ssoPolicyHeaders.Add("x-ms-SsoFlags", "SsoRestrUndefined ");
```
gladjohn, line 593:
```suggestion
                ssoPolicyHeaders.Add("x-ms-SsoFlags", "SsoRestrUndefined");
```

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
```diff
@@ -125,11 +125,11 @@ private IPublicClientApplication CreatePca(AuthMethod? authMethod)
                     break;
                 case AuthMethod.SystemBrowser:
                     builder.WithBroker(new BrokerOptions(BrokerOptions.OperatingSystems.None));
-                    builder = builder.WithBroker(false);
+                    builder = builder.WithBroker(false).WithSsoPolicy();
```
gladjohn, line 128: suggest adding a check box for this in the form

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -552,6 +564,37 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
...
+            }
+
+            return ssoPolicyHeaders;
```
gladjohn, line 596: would it be ok to log what headers were added based on the policy type?

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
```diff
@@ -31,6 +31,7 @@ internal interface IBroker
             AuthenticationRequestParameters authenticationRequestParameters,
             AcquireTokenByUsernamePasswordParameters acquireTokenByUsernamePasswordParameters);
 
+        Dictionary<string, string> GetSsoPolicyHeaders();
```
gladjohn, line 34: Would `DetermineSsoPolicyHeaders` be an appropriate name? given that the method's purpose is to resolve or determine which headers to use based on the SSO policy type Runtime returns?

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
ashok672: Yes thats the plan for now. Runtime does support cross plat and if there is need for this in other platforms, we can extend it

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
ashok672: Done

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
ashok672: I actually realized later this was  not needed when I created the broker instance.  I will remove this.

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
ashok672: Done

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
ashok672: I do like this. I previously forgot to check if policy is enabeld in this particular flow and I added that on top of your change.

src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs
ashok672: Resolved in another comment

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -407,6 +407,9 @@ public static string CertMustBeRsa(string certificateFriendlyName)
 
         public const string MultiCloudSupportUnavailable = "Multi cloud support unavailable with broker.";
 
+        public const string BrokerAndSsoPolicyConfigurationError = "WithBroker and WithSsoPolicy cannot be used together. Broker by default supports SsoPolicy. Use WithSsoPolicy only when using embedded webview";
```
bgavrilMS, line 410:
```suggestion
        public const string BrokerAndSsoPolicyConfigurationError = "WithBroker and WithSsoPolicy cannot be used together. Broker by default supports SsoPolicy. Use WithSsoPolicy only when using embedded or system webview";

```

Sorry for going back and forwards on this. I now realize that since you inject this in the "refresh_token" flow, it implicitly covers all webviews.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -552,6 +552,37 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
...
+            if (ssoPolicy._ssoPolicyType == SsoPolicyType.PermissionRequired)
+            {
+                ssoPolicyHeaders.Add("x-ms-SsoFlags", "SsoRestr");
+            }
+            else if (ssoPolicy._ssoPolicyType == SsoPolicyType.Error)
+            {
+                ssoPolicyHeaders.Add("x-ms-SsoFlags", "SsoPolicyError");
+                string subStatusValue = "SsoRestrError:" + ssoPolicy._errorCode.ToString();
+                byte[] bytes = System.Text.Encoding.UTF8.GetBytes(subStatusValue);
+                string base64UrlEncoded = System.Convert.ToBase64String(bytes)
```
bgavrilMS, line 573: nit: See Base64UrlHelpers for equivalent functionality.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -552,6 +564,37 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
                 }
             }
         }
+        public Dictionary<string, string> GetSsoPolicyHeaders()
+        {
+            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
+            Debug.Assert(s_lazyCore.Value != null, "Should not call this API if MSAL runtime init failed");
```
MSamWils, line 558: >Should not call this API if MSAL runtime init failed [](http://example.com/codeflow?start=52&length=52)

I don't believe this is a requirement. 
And we should exit out if we know we cannot call the policy api here.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -552,6 +564,37 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
                 }
             }
         }
+        public Dictionary<string, string> GetSsoPolicyHeaders()
```
MSamWils: >GetSsoPolicyHeaders [](http://example.com/codeflow?start=42&length=19)

Please make sure to test both MSA and AAD.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -552,6 +564,37 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
...
+        {
+            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
+            Debug.Assert(s_lazyCore.Value != null, "Should not call this API if MSAL runtime init failed");
+
+            NativeInterop.SsoPolicy ssoPolicy = new SsoPolicy();
+            _logger.Info(() => $"[RuntimeBroker] Broker returned SsoPolicyType {ssoPolicy._ssoPolicyType} and errorCode {ssoPolicy._errorCode}.");
+
+            var ssoPolicyHeaders = new Dictionary<string, string>();
```
MSamWils, line 563: >= new Dictionary<string, string>(); [](http://example.com/codeflow?start=33&length=35)

Out side of EU will be always allowed. 
So we should just call new when we need to.

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -444,6 +444,11 @@ internal override void Validate()
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (Config.IsBrokerEnabled && Config.IsSsoPolicyEnabled)
```
MSamWils: >IsBrokerEnabled [](http://example.com/codeflow?start=23&length=15)

I don't believe we need this check.
Setting broker enable is not contradict with the sso policy. (Especially in non-windows platform)

I see this check has more harm than good.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
MSamWils: This policy may not only scope to webview.

We can use IsPlatformAutomaticAppSignInPolicyEnabled. which is coming from the public API:
[https://learn.microsoft.com/en-us/uwp/api/windows.system.profile.platformautomaticappsigninpolicy?view=winrt-insider](https://learn.microsoft.com/en-us/uwp/api/windows.system.profile.platformautomaticappsigninpolicy?view=winrt-insider)

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
MSamWils: I would propose something like GetSsoPolicy, AcquireSsoPoilcy or along those line.

The determination is coming from windows api. But it shouldn't specific to the header since it could also be in cookie in some cases.

pmaytak:
GitHub issue should be created/associated with this PR.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -121,6 +126,9 @@ public string ClientVersion
         public ManagedIdentityId ManagedIdentityId { get; internal set; }
 
         public bool IsManagedIdentity { get; }
+        public bool IsConfidentialClient { get; }
+
+        public bool IsPublicClient => !IsConfidentialClient && !IsManagedIdentity;
```
pmaytak, line 131:
```suggestion
        public bool IsConfidentialClient { get; }
        public bool IsPublicClient => !IsConfidentialClient && !IsManagedIdentity;

```

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -47,20 +47,35 @@ public static PublicClientApplicationBuilder WithBrokerPreview(this PublicClient
...
-            AddRuntimeSupportForWam(builder);
+            AddRuntimeSupport(builder);
             builder.Config.BrokerOptions = brokerOptions;
             builder.Config.IsBrokerEnabled = brokerOptions.IsBrokerEnabledOnCurrentOs();
             return builder;
         }
 
-        private static void AddRuntimeSupportForWam(PublicClientApplicationBuilder builder)
+        /// <summary>
+        /// Use this API to enable SsoPolicy enforcement.
```
pmaytak, line 57: nit: Should the comment be more descriptive - what the SSO Policy is or what it actually does?

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
```diff
@@ -444,6 +444,11 @@ internal override void Validate()
             {
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
+
+            if (Config.IsBrokerEnabled && Config.IsWebviewSsoPolicyEnabled)
+            {
+                throw new MsalClientException(MsalErrorMessage.BrokerAndSsoPolicyConfigurationError);
+            }
```
pmaytak, line 451: Should this be removed? If someone specifies WithBroker and WithSsoPolicy, let the broker handle the SSO stuff.

src/client/Microsoft.Identity.Client/Platforms/Features/WinFormsLegacyWebUi/WindowsFormsWebAuthenticationDialogBase.cs
```diff
@@ -276,7 +281,21 @@ internal AuthorizationResult AuthenticateAAD(Uri requestUri, Uri callbackUri, Ca
...
 
-            _webBrowser.Navigate(requestUri);
+            if(RequestContext.ServiceBundle.Config.IsWebviewSsoPolicyEnabled)
+            {
+                IBroker broker = RequestContext.ServiceBundle.Config.BrokerCreatorFunc(null, RequestContext.ServiceBundle.Config, RequestContext.Logger);
+                var ssoPolicyHeaders = broker.GetSsoPolicyHeaders();
+                string ssoPolicyHeadersString="";
+                foreach (KeyValuePair<string, string> kvp in ssoPolicyHeaders)
+                {
+                    ssoPolicyHeadersString += kvp.Key + ":" + kvp.Value + "\r\n";
```
pmaytak, line 291:
```suggestion
                    ssoPolicyHeadersString += kvp.Key + ":" + kvp.Value + Environment.NewLine;
```

src/client/Microsoft.Identity.Client/Internal/Broker/NullBroker.cs
```diff
@@ -80,5 +81,10 @@ public Task<MsalTokenResponse> AcquireTokenByUsernamePasswordAsync(Authenticatio
             _logger.Info("NullBroker - returning null on ROPC request.");
             return Task.FromResult<MsalTokenResponse>(null);
         }
+
+        public IReadOnlyDictionary<string, string> GetSsoPolicyHeaders()
```
pmaytak, line 85: nit: It's more performant to iterate over concrete types rather than the interfaces.
```suggestion
        public ReadOnlyDictionary<string, string> GetSsoPolicyHeaders()
```

src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs
bgavrilMS: interface is internal, don't dwell on this too long :)

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
bgavrilMS: This only executes in DEBUG mode (so it's removed by compiler in the final bits). It's a message for the MSAL team dev, not for app deveveloper.

tests/devapps/WAM/NetCoreWinFormsWam/Form1.cs
ashok672: This can be done later.

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
```diff
@@ -552,6 +552,37 @@ public async Task RemoveAccountAsync(ApplicationConfiguration appConfig, IAccoun
...
+        public IReadOnlyDictionary<string, string> GetSsoPolicyHeaders()
+        {
+            using LogEventWrapper logEventWrapper = new LogEventWrapper(this);
+            Debug.Assert(s_lazyCore.Value != null, "Should not call this API if MSAL runtime init failed");
+
+            NativeInterop.SsoPolicy ssoPolicy = new SsoPolicy();
+            _logger.Info(() => $"[RuntimeBroker] Broker returned SsoPolicyType {ssoPolicy._ssoPolicyType} and errorCode {ssoPolicy._errorCode}.");
+
+            var ssoPolicyHeaders = new Dictionary<string, string>();
+            if (ssoPolicy._ssoPolicyType == SsoPolicyType.PermissionRequired)
```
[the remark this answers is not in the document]
ashok672: Fixed

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
ashok672: Its only for 1st party applications that I am going to closely work with. I did update the comment with little more details

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
ashok672: Removed this now after our offline discussion

src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs
ashok672: yes, we had a discussion about this offline with WAM. This indeed is not needed because we might need to provide fallback logic in case broker is not installed even when apps use broker

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
ashok672: Not critical at this moment. we did log the overall response

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
ashok672: Done

src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/RuntimeBroker.cs
ashok672: not critical now. can look into this later

bgavrilMS:
@ashok672 - this is now merged.

PR #4672, opened 2024-03-21
HttpManager updates to include support for MTLS

gladjohn:
HttpManager udates to include support for MTLS


bgavrilMS:
@gladjohn - we discussed with Neha that it would be easier to just deliver all your changes for MSIv2 as part of the public preview.

neha-bhargava:
@gladjohn can this PR be closed?

PR #4674, opened 2024-03-22
Use MSI based Resource Manager for Builds / Tests

gladjohn:
Use MSI based Resource Manager for Builds / Tests

### DIFF
--- a/build/template-install-keyvault-secrets.yaml
+++ b/build/template-install-keyvault-secrets.yaml
@@ -2,13 +2,14 @@
 # Install all secrets needed from KeyVault onto the build machine.
 
 steps:
-- task: AzureKeyVault@1
-  displayName: 'Azure Key Vault: buildautomation'
+# Azure Key Vault v2
+# Download Azure Key Vault secrets.
+- task: AzureKeyVault@2
   inputs:
-    azureSubscription: '.NET Keyvault'
-    KeyVaultName: buildautomation
-    SecretsFilter: 'AzureADIdentityDivisionTestAgentCert'
-
+    azureSubscription: 'AuthSdkResourceManager' # string. Alias: ConnectedServiceName. Required. Azure subscription. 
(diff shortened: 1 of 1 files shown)


build/template-install-keyvault-secrets.yaml
```diff
@@ -2,13 +2,14 @@
...
+# Download Azure Key Vault secrets.
+- task: AzureKeyVault@2
   inputs:
-    azureSubscription: '.NET Keyvault'
-    KeyVaultName: buildautomation
-    SecretsFilter: 'AzureADIdentityDivisionTestAgentCert'
-
+    azureSubscription: 'AuthSdkResourceManager' # string. Alias: ConnectedServiceName. Required. Azure subscription. 
+    KeyVaultName: 'buildautomation' # string. Required. Key vault. 
+    SecretsFilter: '*' # string. Required. Secrets filter. Default: *.
```
bgavrilMS, line 11: Why is the SecretsFilter changed?

build/template-install-keyvault-secrets.yaml
```diff
@@ -2,13 +2,14 @@
 # Install all secrets needed from KeyVault onto the build machine.
 
 steps:
-- task: AzureKeyVault@1
-  displayName: 'Azure Key Vault: buildautomation'
+# Azure Key Vault v2
+# Download Azure Key Vault secrets.
+- task: AzureKeyVault@2
```
bgavrilMS, line 7: Can you add some comments / details that the connection is done via MSI

bgavrilMS (approved):

build/template-install-keyvault-secrets.yaml
gladjohn: updated it. will need to move away from this double hop of KeyVaults and directly call into the labs vault next. Will file a task to understand the work involved and how we can get rid of the cert

neha-bhargava (approved):

PR #4679, opened 2024-03-23
Use internal feed for OneBranch

gladjohn:


NuGet.Config
```diff
@@ -2,6 +2,6 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="authsdkinternalfeed" value="https://identitydivision.pkgs.visualstudio.com/IDDP/_packaging/authsdkinternalfeed/nuget/v3/index.json" />
```
jmprieur, line 5: @gladjohn: We can't do that. This is an open source project. It needs to be buildable by customers.
What you can do is add a build step to replace the NuGet.Config

```
dotnet nuget disable source NuGet
dotnet nuget add source  "https://identitydivision.pkgs.visualstudio.com/IDDP/_packaging/authsdkinternalfeed/nuget/v3/index.json" --name authsdkinternalfeed
```

gladjohn: Thanks @jmprieur updated it for the OneBranch pipeline only

PR #4700, opened 2024-04-03
Update to use lab app for automation

gladjohn:
Update to use lab app for automation

OneBranch Pipeline run : https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1280123&view=results


build/template-install-keyvault-secrets.yaml
```diff
@@ -8,12 +8,12 @@ steps:
   inputs:
     azureSubscription: 'AuthSdkResourceManager' # string. Workload identity service connection to use managed identity authentication
     KeyVaultName: 'buildautomation' # string. Required. The name of the Key Vault containing the secrets.
-    #setting secrets filter to fetch only AzureADIdentityDivisionTestAgentCert cert from the vault
```
trwalke, line 11: Why do we need to change the cert?

gladjohn: because the `LabVaultAccessCert` is maintained by the lab

gladjohn: to avoid confusion, I now directly refer to the lab cert from the lab vault

neha-bhargava (approved):

pmaytak (approved):

PR #4709, opened 2024-04-12
Update Selenium dependencies

pmaytak:
Update Selenium and VS Test dependencies to make tests more stable.
Remove non-existent project from from solution.


tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/SeleniumExtensions.cs
```diff
@@ -70,7 +70,7 @@ public static void SaveScreenshot(this IWebDriver driver, TestContext testContex
 #endif
 
             Trace.WriteLine($"Saving picture to {failurePicturePath}");
-            ss.SaveAsFile(failurePicturePath, ScreenshotImageFormat.Png);
```
trwalke, line 73: Do we not want screenshots?

trwalke (approved):

tests/Microsoft.Identity.Test.Integration.netfx/Infrastructure/SeleniumExtensions.cs
pmaytak: It still saves screenshots as PNG by default. In the latest version, they just removed the overload to specify the image type.

bgavrilMS: A bit of a breaking change on their part huh? :)

PR #4731, opened 2024-04-25
Add cert validation for service fabric

neha-bhargava:
Fixes #4655 

**Testing**

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -28,6 +28,7 @@ internal class HttpManager : IHttpManager
         protected readonly IMsalHttpClientFactory _httpClientFactory;
         private readonly Func<HttpResponse, bool> _retryCondition;
         public long LastRequestDurationInMs { get; private set; }
+        public HttpClientHandler HttpClientHandler { get; set; }
```
bgavrilMS, line 31: It's going to be tough to reason about this class if it becomes mutable. Can you make it immutable by moving this to a ctor?

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -48,9 +48,30 @@ public static AbstractManagedIdentity TryCreate(RequestContext requestContext)
...
+            handler.ServerCertificateCustomValidationCallback = (message, certificate, chain, sslPolicyErrors) =>
+            {
+                if (sslPolicyErrors != System.Net.Security.SslPolicyErrors.None)
+                {
+                    return 0 == string.Compare(certificate.Thumbprint, EnvironmentVariables.IdentityServerThumbprint, StringComparison.OrdinalIgnoreCase);
+                }
+                return true;
+            };
+
+            requestContext.ServiceBundle.HttpManager.HttpClientHandler = handler;
```
bgavrilMS, line 71: But then all the other flows will need to set ServiceBundle.HttpManager.HttpClientHandler to null and reset the HttpManager's static cache. 

I think this will be quite messy. 

I'd go for either: 

- make HttpManager immutable, e.g. by adding the special handler on the `RequestAsync` method instead of making it a property
- or get away completely from this and create a new HttpClient (and cache it) here statically

src/client/Microsoft.Identity.Client/Http/IHttpManager.cs
```diff
@@ -14,6 +14,7 @@ namespace Microsoft.Identity.Client.Http
     internal interface IHttpManager
     {
         long LastRequestDurationInMs { get; }
+        HttpClientHandler HttpClientHandler { get; set; }
```
bgavrilMS, line 17: See my comments below. We should not make the class mutable, it will be very difficult to reason with because there is static caching in here. 

How about adding a HttpClientHandler in the SendRequedstAsync instead? Or better yet, add the callback as an input param of that factory? HttpManger already uses HttpClientHandler for mtls purposes.

neha-bhargava:
Since this branch was based off of SLC branch, I am unable to rebase it to http manager branch as it is pulling changes from SLC branch too. I will close this and open another PR addressing these comments and based off of updated http manager.

PR #4752, opened 2024-05-07
Used multiple ARM API versions for subscriptions/get in WwwAuthenticateParametersIntegrationTests

abatishchev:

Currently ARM API for subscriptions/get supports multiple API versions, this PR updates the one used to the currently latest.

**Testing**

Yes, the PR updates the integration tests only.

**Performance impact**

No perf impact.

**Documentation**

<s>- [ ] All relevant documentation is updated.</s>


abatishchev:
@bgavrilMS I can update the tests to try multiple API versions, just to increase the coverage.

bgavrilMS (approved):

bgavrilMS:
> @bgavrilMS I can update the tests to try multiple API versions, just to increase the coverage.

Are there different formats of the WWW-AUthenticate header?

abatishchev:
@bgavrilMS no, the only difference is in the API version when it calls ARM API. Which strives on backward compatibility, so I don't expect a different outcome or behavior per API version

bgavrilMS:
> @bgavrilMS no, the only difference is in the API version when it calls ARM API. Which strives on backward compatibility, so I don't expect a different outcome or behavior per API version

Then I think we are good.

abatishchev:
@bgavrilMS it's ready for a re-review

bgavrilMS (approved):

bgavrilMS:
Tested this locally.

PR #4753, opened 2024-05-08
Fix for #4608- do not throw exception for ADFS + WithTenantId

### LINKED ISSUE #4608: [Bug] Ignore WithTenantId on non-tenated authorities
### Library version used

4.58

### .NET version

all

### Scenario

None

https://github.com/AzureAD/microsoft-identity-web/issues/2509

WithTenantId should be no-op for ADFS, Generic authority etc. so as to not have to write custom logic in Id.Web and other high level APIs. 

### Relevant code snippets

_No response_

### Expected behavior

_No response_

### Identity provider

### Regression

no

### Solution and workarounds

_No response_

bgavrilMS:
Fixes #4608 

This will help partners (Id.Web) keep their code simple and not having to worry about authority type in OBO.

https://github.com/AzureAD/microsoft-identity-web/issues/2509

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
@@ -132,7 +132,6 @@ namespace Microsoft.Identity.Client
         /// If both WithLogging apis are set, this one will override the other
         /// </param>
         /// <returns>The builder to chain the .With methods</returns>
-        /// <remarks>This is an experimental API. The method signature may change in the future without involving a major version upgrade.</remarks>
         public T WithLogging(
             IIdentityLogger identityLogger,
             bool enablePiiLogging = false)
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
```diff
@@ -218,13 +217,6 @@ internal void ResolveAuthority()
                 // Both WithAuthority and WithTenant were used at app config level
                 if (!string.IsNullOrEmpty(Config.TenantId))
                 {
-                    if (!Config.Authority.AuthorityInfo.CanBeTenanted)
```
neha-bhargava, line 221: Replace with a log?

bgavrilMS: We don't have a logger at this stage, since the app is only being built.

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
```diff
@@ -218,13 +217,6 @@ internal void ResolveAuthority()
                 // Both WithAuthority and WithTenant were used at app config level
                 if (!string.IsNullOrEmpty(Config.TenantId))
```
pmaytak, line 218: So the new behavior is ignoring the tenant for ADFS authorities?

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5daef991982e092d2eeb4b6047cf8de585d5e5cd/src/client/Microsoft.Identity.Client/Instance/AdfsAuthority.cs#L58-L61

pmaytak (approved):

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
bgavrilMS: Yes.

bgavrilMS: ADFS team (Bryant) confirm that even if there is a `tid` claim in the ADFS token, ADFS is not tenanted and it should be ignored.

PR #4761, opened 2024-05-13
Add msi source detect logic

### LINKED ISSUE #4751: [Engineering task] Need an API for Managed Identity to detect the current environment
### Task type

Development

### Description

Please refer to the "MSAL implementation" part of [this internal PR](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?version=GBdev&path=/ManagedIdentity/FacilitateImdsProbing.md&_a=preview&anchor=msal-implementations).

### Solution

_No response_

neha-bhargava:
Fixes #4751 

Expose an API to get and return current environment for managed identity

**Testing**
Added unit tests

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -47,5 +47,41 @@ private static AbstractManagedIdentity SelectManagedIdentitySource(RequestContex
                 AzureArcManagedIdentitySource.TryCreate(requestContext) ??
                 new ImdsManagedIdentitySource(requestContext);
         }
+
+        internal static ManagedIdentitySource GetManagedIdentitySource()
+        {
+            string identityEndpoint = EnvironmentVariables.IdentityEndpoint;
```
bgavrilMS, line 53: this logic already exists in MSAL in those "tryCreate" methods, can it be reused?

CC @Robbie-Microsoft

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -47,5 +47,41 @@ private static AbstractManagedIdentity SelectManagedIdentitySource(RequestContex
...
+            {
+                return ManagedIdentitySource.AzureArc;
+            }
+            else if (!string.IsNullOrEmpty(podIdentityEndpoint))
+            {
+                return ManagedIdentitySource.Imds;
+            }
+            else
+            {
+                return ManagedIdentitySource.DefaultToImds;
```
bgavrilMS, line 83: I don't understand the difference between IMDS and DefaultToImds. It's just IMDS. Maybe it's just a naming problem - is the Pod releated to Kubernetes? In which case maybe call it KubernetesImds? 

It does not see to be as per spec: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?version=GBdev&path=/ManagedIdentity/FacilitateImdsProbing.md&_a=preview&anchor=msal-implementations

bgavrilMS:
@Robbie-Microsoft @Avery-Dunn @rayluo - pls make sure you review this.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: Ideally we should find this environment variable POD_IDENTITY on the VM. If that is the case, it definitely is a VM with managed identity configured. But since this environment variable is generally not found on the VM, we Default to VM.  I don't think I have seen the POD_IDENTITY environment variable ever during my testing and I am not sure what is the probability of having this set. CC: @rayluo

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: Refactored a bit to separate the logic for detect and validate.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -54,5 +54,14 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
                 ClientExecutorFactory.CreateManagedIdentityExecutor(this),
                 resource);
         }
+
+        /// <summary>
+        /// Detects and returns the managed identity source available on the environment.
+        /// </summary>
+        /// <returns>Managed identity source detected on the environment if any.</returns>
+        public static ManagedIdentitySource GetManagedIdentitySource()
```
gladjohn, line 62: how will this behave in a pure FIC scenario?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -42,6 +42,12 @@ public enum ManagedIdentitySource
...
         /// The source to acquire token for managed identity is Service Fabric.
         /// </summary>
-        ServiceFabric
+        ServiceFabric,
+
+        /// <summary>
+        /// The source to acquire token for managed identity is defaulted to IMDS as no environment variables are found. 
+        /// This is specifically used when the source is not known.
+        /// </summary>
+        DefaultToImds
```
gladjohn, line 51: We already have an IMDS source, what is the purpose of this one?

neha-bhargava: When no environment variable is found it is not really IMDS but default to IMDS as per the api review https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/12459?_a=files&path=/ManagedIdentity/FacilitateImdsProbing.md. I can leave it as IMDS. But I don't think we should rename the existing enum as it will be a breaking change. So either add a new enum or use exiting one.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
rayluo: I am not aware of the POD_IDENTITY env var either. Out of curiosity, is it defined in some docs, or was it backported from Azure Identity .Net's implementation?

Regardless, I suppose you may keep both the existing `Imds` logic and the new `DefaultToImds` logic. In such a case, the new enum name "*DefaultTo*Imds" actually plays well with your existing "Imds" name, because their different names hint their subtle differences.

Aside: Personally, I would choose the name "...VM" rather than "...Imds". VM is a kind of runtime environment, [IMDS is a service/endpoint available in VM](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux). Many other Managed Identity environments use an IMDS-like endpoint, too.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
rayluo: It is considered FIC-agnostic, because it only detects the underlying managed identity environment. It is the calling app to decide whether use a managed identity or FIC.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: @rayluo, this was backported from Azire .Net's implementation.

I agree VM would have been a better choice but it is already there.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -40,12 +40,63 @@ internal Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityAsync(A
...
-                ServiceFabricManagedIdentitySource.TryCreate(requestContext) ??
-                AppServiceManagedIdentitySource.TryCreate(requestContext) ?? 
-                CloudShellManagedIdentitySource.TryCreate(requestContext) ??
-                AzureArcManagedIdentitySource.TryCreate(requestContext) ??
-                new ImdsManagedIdentitySource(requestContext);
+            ManagedIdentitySource managedIdentitySource = GetManagedIdentitySource();
+
+            switch (managedIdentitySource)
+            {
+                case ManagedIdentitySource.ServiceFabric:
```
gladjohn, line 47: Would be nice to move the source detection logic here and cache it, so we do not detect the source again.

bgavrilMS: +1

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: KubernetesImds?

bgavrilMS (approved):
Approved with comments,.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: That is a good point. These should be consistent across MSALs for telemetry. But the DefaultToImds is only sent back as response to GetManagedIdentitySource and not used for other scenarios.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -40,12 +40,63 @@ internal Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityAsync(A
...
+            }
+            else if (!string.IsNullOrEmpty(msiEndpoint))
+            {
+                return ManagedIdentitySource.CloudShell;
+            }
+            else if (!string.IsNullOrEmpty(identityEndpoint) && !string.IsNullOrEmpty(imdsEndpoint))
+            {
+                return ManagedIdentitySource.AzureArc;
+            }
+            else if (!string.IsNullOrEmpty(podIdentityEndpoint))
```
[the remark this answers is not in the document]
neha-bhargava: I see. Thanks for pointing out. Did not know this.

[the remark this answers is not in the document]
gladjohn: also POD Identity endpoint has been deprecated, per MSI folks. Do we have any telemetry if we have ever detected this env var?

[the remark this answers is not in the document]
neha-bhargava: In that case only IMDS. I will remove the Default and add IMDS

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
neha-bhargava: Gladwin mentioned in the comment below that the pod identity env variable is deprecated. I will use the existing enum

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: @rayluo, using the existing enum IMDS. Do you see any issues with it?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -37,6 +37,27 @@ public class ManagedIdentityTests : TestBase
...
 
+        [DataTestMethod]
+        [DataRow("http://127.0.0.1:41564/msi/token/", ManagedIdentitySource.AppService, ManagedIdentitySource.AppService)]
+        [DataRow(AppServiceEndpoint, ManagedIdentitySource.AppService, ManagedIdentitySource.AppService)]
+        [DataRow(ImdsEndpoint, ManagedIdentitySource.Imds, ManagedIdentitySource.Imds)]
+        [DataRow(null, ManagedIdentitySource.Imds, ManagedIdentitySource.Imds)]
+        [DataRow(AzureArcEndpoint, ManagedIdentitySource.AzureArc, ManagedIdentitySource.AzureArc)]
+        [DataRow(CloudShellEndpoint, ManagedIdentitySource.CloudShell, ManagedIdentitySource.CloudShell)]
+        [DataRow(ServiceFabricEndpoint, ManagedIdentitySource.ServiceFabric, ManagedIdentitySource.ServiceFabric)]
+        public void GetManagedIdentityTests(
```
gladjohn, line 48:
```suggestion
        public void GetManagedIdentityTests(
    string endpoint,
    ManagedIdentitySource expectedManagedIdentitySource)
```

you just need one and then compare it with actual

gladjohn (approved):
Approved. Please cache it so we do not check the source again, this will also be useful for SLC

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: Discussed offline

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
neha-bhargava: The only difference is for IMDS so passing the expected one to

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -40,12 +40,48 @@ internal Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityAsync(A
...
-                AppServiceManagedIdentitySource.TryCreate(requestContext) ?? 
-                CloudShellManagedIdentitySource.TryCreate(requestContext) ??
-                AzureArcManagedIdentitySource.TryCreate(requestContext) ??
-                new ImdsManagedIdentitySource(requestContext);
+            return s_managedIdentitySourceDetected.Value switch
+            {
+                ManagedIdentitySource.ServiceFabric => ServiceFabricManagedIdentitySource.TryCreate(requestContext),
+                ManagedIdentitySource.AppService => AppServiceManagedIdentitySource.TryCreate(requestContext),
+                ManagedIdentitySource.CloudShell => CloudShellManagedIdentitySource.TryCreate(requestContext),
+                ManagedIdentitySource.AzureArc => AzureArcManagedIdentitySource.TryCreate(requestContext),
```
rayluo, line 48: Nit: I guess all those `TryCreate()` can be renamed to `Create()` now. Not a big deal, though, I'm approving.

rayluo (approved):
LGTM. Thank you for the effort!

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -39,6 +39,8 @@ internal class ImdsManagedIdentitySource : AbstractManagedIdentity
         internal ImdsManagedIdentitySource(RequestContext requestContext) : 
             base(requestContext, ManagedIdentitySource.Imds)
         {
+            requestContext.Logger.Info(() => "[Managed Identity] Defaulting to IMDS managed identity.");
```
gladjohn, line 42:
```suggestion
            requestContext.Logger.Info(() => "[Managed Identity] Defaulting to IMDS.");
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -40,12 +40,48 @@ internal Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityAsync(A
...
+            string msiEndpoint = EnvironmentVariables.MsiEndpoint;
+            string imdsEndpoint = EnvironmentVariables.ImdsEndpoint;
+            string podIdentityEndpoint = EnvironmentVariables.PodIdentityEndpoint;
+
+            if (!string.IsNullOrEmpty(identityEndpoint) && !string.IsNullOrEmpty(identityHeader) && 
+                !string.IsNullOrEmpty(identityServerThumbprint))
+            {
+                return ManagedIdentitySource.ServiceFabric;
+            }
+            else if (!string.IsNullOrEmpty(identityEndpoint) && !string.IsNullOrEmpty(identityHeader))
```
gladjohn, line 69: you have already checked this condition, why do we need to check this again? 

```cs
if (!string.IsNullOrEmpty(identityEndpoint) && !string.IsNullOrEmpty(identityHeader))
    {
        // Check for the additional condition if the first two conditions are met
        if (!string.IsNullOrEmpty(identityServerThumbprint))
        {
            return ManagedIdentitySource.ServiceFabric;
        }
        return ManagedIdentitySource.AppService;
    }
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -20,7 +20,7 @@ public enum ManagedIdentitySource
         None,
 
         /// <summary>
-        /// The source to acquire token for managed identity is IMDS.
+        /// The source to acquire token for managed identity is defaulted to IMDS.
```
gladjohn, line 23:
```suggestion
        /// The source to acquire token for managed identity is IMDS.
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -42,6 +42,12 @@ public enum ManagedIdentitySource
...
         /// The source to acquire token for managed identity is Service Fabric.
         /// </summary>
-        ServiceFabric
+        ServiceFabric,
+
+        /// <summary>
+        /// Indicates that the source is defaulted to IMDS since no environment variables are set.
+        /// This is used to detect the managed identity source.
+        /// </summary>
+        DefaultToImds
```
gladjohn, line 51:
```suggestion
        DefaultToImds
```
did we consider `FallbackToImds` or `ImdsAutoDetect` ?

neha-bhargava: `DefaultToImds` sounds like a name that solves the purpose. I don't think the other names you mentioned suits here. Any reason you are suggesting a name change now? We discussed the names earlier too.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: I think we can simply call the constructor now. Thanks for pointing this out.

gladjohn (approved):

Issue #4784, opened 2024-05-23
[Bug] MeasureDurationResult incorrect when running in linux


SingleCopy:
### Library version used

4.60.3

### .NET version

C# .net 8.0

### Scenario

ManagedIdentityClient - managed identity

### Is this a new or an existing app?

The app is in production, I haven't upgraded MSAL, but started seeing this issue

### Issue description and reproduction steps

When our unit tests are run using a linux distro, the AuthenticationResult->AuthenticationResultMetadata->DurationTotalInMs is returning a incorrect time. This is due to the Stopwatch.Frequency being 10000000 on windows and 1000000000 on linux, this is not taken into account when performing the calculation.

[](url)

### Relevant code snippets

```csharp
public MeasureDurationResult(long ticks)
{
    Milliseconds = ticks / TimeSpan.TicksPerMillisecond;
    Microseconds = ticks / (TimeSpan.TicksPerMillisecond / 1000);
    Ticks = ticks;
}
```

### Expected behavior

Regardless of O/S, the DurationTotalInMs should be the same, however, on linux, the DurationTotalInMs is incorrect

Windows:
ElapsedTicks: 8423086
DurationTotalInMs: 842

Linux
ElapsedTicks: 8423086
DurationTotalInMs: 84230

### Identity provider

Other

localden:
Sounds like we need to update the converstion factor (`1000.0 / Stopwatch.Frequency`).

SingleCopy:
@localden Correct, this is for Managed Identity, I have updated my description.

PR #4785, opened 2024-05-23
MeasureDurationResult incorrect when running in linux

SingleCopy:
Fixes: #4784

The MeasureDurationResult time calculations do not work when running in linux as the Stopwatch.Frequency is not taken into account. This change correctly calculates the Milliseconds and Microseconds by taking the Stopwatch.Frequency into account.

**Testing**
Difficult without running the tests on different O/S

**Performance impact**
N/A

**Documentation**
N/A


bgavrilMS:
This is a very good catch @SingleCopy . I found some mention about this here: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.elapsedticks?view=net-8.0&redirectedfrom=MSDN#System_Diagnostics_Stopwatch_ElapsedTicks

@GeoK - I think this code is common with our other SDKs

gladjohn (approved):

bgavrilMS:
CC @jennyf19 as I believe this code was copied from MISE. Also adding @pmaytak for a review

bgavrilMS:
@neha-bhargava - please review this PR. There may be implications in how we interpret some of our telemetry.

neha-bhargava (approved):
Looks good. Thanks!

Issue #4662, opened 2024-03-12
[Feature Request] Update managed identity retry policy


neha-bhargava:
### MSAL client type

Managed identity

### Problem statement

Currently MSAL.Net retry once, update the managed identity retry policy to retry 3 times by default with a delay of 1s between retry.

neha-bhargava:
Completed as part of Http Manager updates

PR #4793, opened 2024-05-31
Updated httpmanager to support mtls

gladjohn:
- Updated httpmanager to support mtls
- Adds Net472 support 
-

Tests : Unit and Integration


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -167,20 +88,45 @@ private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary
...
             {
                 if (cancellationToken.IsCancellationRequested)
                 {
-                    logger.Info("The HTTP request was cancelled. ");
+                    logger.Info("The HTTP request was canceled. ");
                     throw;
                 }
 
                 logger.Error("The HTTP request failed. " + exception.Message);
+                isRetriable = true;
```
bgavrilMS, line 103: Why retry on task cancelled exception? Isn't this cancellation from the user?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -189,7 +135,7 @@ private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary
             }
 
             // package 500 errors in a "service not available" exception
-            if (IsRetryableStatusCode((int)response.StatusCode))
+            if ((int)response.StatusCode >= 500 && (int)response.StatusCode < 600)
             {
                 throw MsalServiceExceptionFactory.FromHttpResponse(
```
bgavrilMS, line 140: Can you please add the status code in the error message and the uri that produced it? (just host and path, not full uri)

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -200,11 +146,39 @@ private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary
...
-        protected async Task<HttpResponse> ExecuteAsync(
+        private HttpClient GetHttpClient(X509Certificate2 x509Certificate2)
+        {
+            if (_httpClientFactory is IMsalMtlsHttpClientFactory msalMtlsHttpClientFactory)
+            {
+                // If the factory is an IMsalMtlsHttpClientFactory, use it to get an HttpClient with the certificate
+                return msalMtlsHttpClientFactory.GetHttpClient(x509Certificate2);
+            }
+
+            // If the factory is not an IMsalMtlsHttpClientFactory, use it to get a default HttpClient
```
bgavrilMS, line 157: you should log a warning here"?

src/client/Microsoft.Identity.Client/Http/HttpManagerFactory.cs
```diff
@@ -14,16 +14,19 @@ namespace Microsoft.Identity.Client.Http
     /// </summary>
     internal sealed class HttpManagerFactory
     {
-        public static IHttpManager GetHttpManager(IMsalHttpClientFactory httpClientFactory, bool withRetry, bool isManagedIdentity)
+        public static IHttpManager GetHttpManager(
+            IMsalHttpClientFactory httpClientFactory,
+            bool withRetry,
```
bgavrilMS, line 19: Can you get rid of these 2 flags? Just reading the code is tricky here. Might as well have the upper layer inject the retry policy directly.

src/client/Microsoft.Identity.Client/Http/HttpRetryCondition.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Net;
+
+namespace Microsoft.Identity.Client.Http
+{
+    internal static class HttpRetryConditions
```
bgavrilMS, line 8: Have you seen @Robbie-Microsoft 's similar abstraction - https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/4d3c84b3f5b633b269e46e03382fa8d834886ab0/lib/msal-node/src/retry/LinearRetryPolicy.ts#L7

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -27,123 +26,38 @@ namespace Microsoft.Identity.Client.Http
...
-        {
-            return SendRequestAsync(endpoint, headers, null, HttpMethod.Get, logger, cancellationToken: cancellationToken);
-        }
-
         /// <summary>
-        /// Performs the GET request just like <see cref="SendGetAsync(Uri, IDictionary{string, string}, ILoggerAdapter, bool, CancellationToken)"/>
-        /// but does not throw a ServiceUnavailable service exception. Instead, it returns the <see cref="HttpResponse"/> associated
-        /// with the request.
+        /// A new instance of the HTTP manager with a retry *condition*. The retry policy hardcodes: 
+        /// - the number of retries (1)
```
bgavrilMS, line 34: See comment below. Can you find an abstraction that eliminates these 2 hardcoded things? See Robbie's work in MSAL JS

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -27,123 +26,38 @@ namespace Microsoft.Identity.Client.Http
...
-            IDictionary<string, string> headers,
+            Dictionary<string, string> headers,
             HttpContent body,
             HttpMethod method,
             ILoggerAdapter logger,
-            bool doNotThrow = false,
-            bool retry = false,
-            CancellationToken cancellationToken = default)
+            bool doNotThrow,
+            bool retry,
```
bgavrilMS, line 53: Why does this still have retry flag? I thught retries were injected by ctor?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -200,11 +146,39 @@ private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary
             return response;
         }
 
-        protected async Task<HttpResponse> ExecuteAsync(
+        private HttpClient GetHttpClient(X509Certificate2 x509Certificate2)
+        {
+            if (_httpClientFactory is IMsalMtlsHttpClientFactory msalMtlsHttpClientFactory)
```
neha-bhargava, line 151: Since the default client is now an implementation of `IMsalMtlsHttpClientFactory`, this is always true but the cert is not always there. You are adding that check later in the flow. Can this be updated to check if the cert is here and then make a call to either? Also adding a check here will make it more obvious which client is used in which case.

Reason: for the service fabric cert validation, I am trying to create a client with handler. As per the discussion we had, my understanding is if the binding cert is there then we use the MtlsClient and not a default handler.

neha-bhargava (changes requested):

src/client/Microsoft.Identity.Client/Instance/Discovery/NetworkMetadataProvider.cs
```diff
@@ -34,17 +34,17 @@ internal class NetworkMetadataProvider : INetworkMetadataProvider
 
         public async Task<InstanceDiscoveryMetadataEntry> GetMetadataAsync(Uri authority, RequestContext requestContext)
         {
-            var logger = requestContext.Logger;
+            ILoggerAdapter logger = requestContext.Logger;
```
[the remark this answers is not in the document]
trwalke: Using explicit types can make it a bit easier to know of what type you are dealing with when it isnt immediatly obvious. However, when the type is obvious:
```
var x = new TokenCache();
```

I think it is fine to use var. Clearly the above variable x is a `TokenCache`

src/client/Microsoft.Identity.Client/Internal/Requests/DeviceCodeRequest.cs
```diff
@@ -29,7 +29,7 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
         {
             await ResolveAuthorityAsync().ConfigureAwait(false);
 
-            var client = new OAuth2Client(ServiceBundle.ApplicationLogger, ServiceBundle.HttpManager);
+            var client = new OAuth2Client(ServiceBundle.ApplicationLogger, ServiceBundle.HttpManager, null);
```
trwalke, line 32: I see that there are a number of instances where this constructor is called with the binding cert `null`. Wondering if it makes sense to add an overload where it is assumed to be null? Or maybe a default value?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -14,23 +16,73 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
...
+            HttpClientConfig.ConfigureRequestHeadersAndSize(httpClient);
+
+            return httpClient;
+        }
+
+        private static HttpClient CreateMtlsHttpClient(X509Certificate2 bindingCertificate)
+        {
+            CheckAndManageCache();
+
+            if (bindingCertificate == null)
```
trwalke, line 43: does it make sense to put this check inside the `#if SUPPORTS_MTLS` block below? If the target platform cant use MTLS the client probably wouldnt pass in a binding cert. Or would it?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava, line 34: For managed identity flow we need to update the retry logic "retry policy to retry 3 times by default with a delay of 1s between retry." 
See #4662

bgavrilMS:
Any updates on this PR @gladjohn ? Is it ready for a re-review?

gladjohn:
> Any updates on this PR @gladjohn ? Is it ready for a re-review?

Had few test failures last week after addressing all the changes. I will start looking into this again this week. Hopefully have something out for re-review by Wednesday

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityRequest.cs
```diff
@@ -17,11 +17,11 @@ internal class ManagedIdentityRequest
 
         public HttpMethod Method { get; }
 
-        public IDictionary<string, string> Headers { get; }
+        public Dictionary<string, string> Headers { get; }
```
neha-bhargava, line 20: Why this change? Shouldn't we program to interface?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: In the line above we check if the cancellation token is cancelled and we throw. If the cancellation token is not canceled, we retry. This has always been like this.

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: Discussed this with Bogdan and if the service fabric is used then server cert validation will be done and exception will be thrown in case an MTLS cert is used.

src/client/Microsoft.Identity.Client/Http/HttpManagerFactory.cs
neha-bhargava: Discussed offline, these flags cannot be eliminated but are needed to use the correct policy

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: Will be done as part of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4820/files. FYI @trwalke

trwalke (approved):

Issue #4809, opened 2024-06-17
[Documentation] - ICustomWebUi.AcquireAuthorizationCodeAsync mistake


NETSphereSoft:
_Make sure that your question or issue is not already covered in [MSAL docs](https://learn.microsoft.com/en-us/entra/msal/dotnet/) or [samples](https://learn.microsoft.com/en-us/azure/active-directory/develop/sample-v2-code?tabs=apptype)._

**Documentation related to component**
Microsoft.Identity.Client

**Please check those that apply**

- [x] typo
- [ ] documentation doesn't exist
- [ ] documentation needs clarification
- [ ] error(s) in example
- [ ] needs example

**Description of the issue**
There is a small mistake in the XML `remarks` of `Microsoft.Identity.Client.Extensibility.ICustomWebUi.AcquireAuthorizationCodeAsync` at [line 34](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4e51d0afea8aab1871e3a5fef7193b87ad43226d/src/client/Microsoft.Identity.Client/Extensibility/ICustomWebUI.cs#L34). There is an additional `"\>` after the `paramref`

bgavrilMS:
Fixed via https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4812

PR #4812, opened 2024-06-18
Doc fix for 4808 and 4809

### LINKED ISSUE #4808: [Feature Request] Allow updating the cached token before expiry
### MSAL client type

Public, Confidential

### Problem statement

MSAL does not currently seem to provide a mechanism to force the refresh of a cached access token before expiry. Such functionality would be useful in cases where the token is known to have changes, such as when group memberships are updated as a result of [PIM for Groups](https://learn.microsoft.com/en-us/entra/id-governance/privileged-identity-management/concept-pim-for-groups).

### Proposed solution

Change the implementation of `WithForceRefresh(true)` to also update the token cache. Updating the token cache would allow use of this function to refresh the currently cached token as used by other invocations utilizing the same cache (using the default `WithForceRefresh(false)`).

### Alternatives

1. Use `WithForceRefresh(true)` for all token acquisitions until cached token expiry.
2. Clearing the cache and reauthenticating, which causes unnecessary effort considering that the refresh token was still valid, just unusable for cached token refresh.

bgavrilMS:
Docs only

Fixes #4808 and #4809

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
@@ -44,7 +44,7 @@ namespace Microsoft.Identity.Client
         /// By default the token is taken from the application token cache (forceRefresh=false)
         /// </summary>
         /// <param name="forceRefresh">
-        /// If <c>true</c>, the request will ignore cached access tokens on read, but will still write them to the cache once obtained from the Identity Provider. The default is <c>false</c>
+        /// If <c>true</c>, the request will ignore cached access tokens on read, but will still write them to the cache once obtained from the identity provider. The default is <c>false</c>
         /// </param>
         /// <remarks>
         /// Do not use this flag except in well understood cases. Identity Providers will throttle clients that make too issue similar token requests.
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
@@ -43,7 +43,9 @@ namespace Microsoft.Identity.Client
         /// New tokens will still be written to the application token cache.
         /// By default the token is taken from the application token cache (forceRefresh=false)
         /// </summary>
-        /// <param name="forceRefresh">If <c>true</c>, the request will ignore cached access tokens on read, but will still write them to the cache once obtained from the Identity Provider. The default is <c>false</c>        /// </param>
+        /// <param name="forceRefresh">
+        /// If <c>true</c>, the request will ignore cached access tokens on read, but will still write them to the cache once obtained from the Identity Provider. The default is <c>false</c>
+        /// </param>
         /// <remarks>
         /// Do not use this flag except in well understood cases. Identity Providers will throttle clients that make too issue similar token requests.
         /// </remarks>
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -39,12 +39,16 @@ internal AcquireTokenForClientParameterBuilder(IConfidentialClientApplicationExe
         }
 
         /// <summary>
-        /// Specifies if the token request will ignore the access token in the application token cache
-        /// and will attempt to acquire a new access token using client credentials.
+        /// Specifies if the client application should ignore access tokens when reading the token cache.
```
trwalke, line 42:
```suggestion
        /// Specifies if the client application should ignore access tokens already present in the token cache and attempt to acquire a new one. 
```
nit: I think it ready better and is a little more clear.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
```diff
@@ -70,9 +70,9 @@ private AcquireTokenSilentParameterBuilder WithLoginHint(string loginHint)
         }
 
         /// <summary>
-        /// Specifies if the client application should force refreshing the
-        /// token from the user token cache. By default the token is taken from the
-        /// the user token cache (forceRefresh=false)
+        /// Specifies if the client application should ignore access tokens when reading the token cache.
+        /// Refresh token will still be used. Any new tokens from the Identity Provider will still be written to the token cache.
```
trwalke, line 74:
```suggestion
        /// Refresh tokens will still be used. Any new tokens from the Identity Provider will still be written to the token cache.
```

trwalke (approved):

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenSilentParameterBuilder.cs
pmaytak, line 74:
```suggestion
        /// Cached refresh token will still be used to acquire fresh tokens from the identity provider, which will still be written to the token cache.
```

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -39,12 +39,16 @@ internal AcquireTokenForClientParameterBuilder(IConfidentialClientApplicationExe
...
         /// <summary>
-        /// Specifies if the token request will ignore the access token in the application token cache
-        /// and will attempt to acquire a new access token using client credentials.
+        /// Specifies if the client application should ignore access tokens when reading the token cache. 
+        /// New tokens will still be written to the application token cache.
         /// By default the token is taken from the application token cache (forceRefresh=false)
         /// </summary>
-        /// <param name="forceRefresh">If <c>true</c>, the request will ignore the token cache. The default is <c>false</c>
+        /// <param name="forceRefresh">
+        /// If <c>true</c>, the request will ignore cached access tokens on read, but will still write them to the cache once obtained from the Identity Provider. The default is <c>false</c>
```
pmaytak, line 47:
```suggestion
        /// If <c>true</c>, the request will ignore cached access tokens on read, but will still write them to the cache once obtained from the identity provider. The default is <c>false</c>
```

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -39,12 +39,16 @@ internal AcquireTokenForClientParameterBuilder(IConfidentialClientApplicationExe
...
         /// </param>
+        /// <remarks>
+        /// Do not use this flag except in well understood cases. Identity Providers will throttle clients that make too issue similar token requests.
```
pmaytak, line 50:
```suggestion
        /// Do not use this flag except in well understood cases. Identity providers will throttle clients that make too many similar token requests.
```

pmaytak (approved):
Approved with suggestions.

PR #4818, opened 2024-06-19
Add validation callback and tests

### LINKED ISSUE #4462: [Feature Request]     Service Fabric MI Auth doesn't validate server cert.
### MSAL client type

Managed identity

### Problem Statement

Currently, MSAL .NET reads the Thumbprint from env but doesn't use it to validate the cert from the server in the transport layer.
This logic needs to be implemented. 

### Proposed solution

.NET Identity SDK does this, this logic can be ported from here:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/ServiceFabricManagedIdentitySource.cs#L55

### Other languages

java: https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java#L1032

Py: (not implemented) https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_credentials/service_fabric.py#L38

JS: (not implemented) https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/identity/identity/src/credentials/managedIdentityCredential/fabricMsi.ts#L128

neha-bhargava:
Fixes #4462 

Add callback for cert validation
Unit test the callback

**Testing**
Unit test

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -45,8 +45,42 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
+            {
+                if (s_httpClient == null)
+                {
+                    s_httpClient = new HttpClient(ValidateServerCertificate(requestContext.Logger));
+                }
+                return s_httpClient;
+            };
+        }
+
+        internal HttpClientHandler ValidateServerCertificate(ILoggerAdapter logger)
```
bgavrilMS, line 60: This method is named as if it were an action, but it is a creator. Maybe `CreateHandlerWithSslValidation`

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -52,6 +52,7 @@ internal class HttpManager : IHttpManager
             bool doNotThrow,
             bool retry,
             X509Certificate2 bindingCertificate,
+            Func<HttpClient> validateServerCertificateCallback,
```
bgavrilMS, line 55: Why do you need a callback instead of an HttpClient? smth like "HttpClient customHttpClient"

The Func is always invoked, so the object is constructed anyway (besides, for lazy optimizations recommendation is to use Lazy<T>)

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -146,14 +149,20 @@ internal class HttpManager : IHttpManager
...
+        private HttpClient GetHttpClient(X509Certificate2 x509Certificate2, Func<HttpClient> validateServerCertificateCallback)
         {
-            if (_httpClientFactory is IMsalMtlsHttpClientFactory msalMtlsHttpClientFactory)
+            if (x509Certificate2 != null && _httpClientFactory is IMsalMtlsHttpClientFactory msalMtlsHttpClientFactory)
             {
                 // If the factory is an IMsalMtlsHttpClientFactory, use it to get an HttpClient with the certificate
                 return msalMtlsHttpClientFactory.GetHttpClient(x509Certificate2);
             }
 
+            HttpClient client = validateServerCertificateCallback?.Invoke();
```
bgavrilMS, line 160: I think you should throw an "unssuported exception" or at least add a Debug.Assert if x509certificate2 is used in combination with this custom httpclient.

The custom httpclient should take precedence.

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -92,6 +94,11 @@ protected AbstractManagedIdentity(RequestContext requestContext, ManagedIdentity
             }
         }
 
+        internal virtual Func<HttpClient> ValidateServerCertificateCallback(RequestContext requestContext)
```
bgavrilMS, line 97: Add a comment that this is internal for test purposes only.

src/client/Microsoft.Identity.Client/Http/IHttpManager.cs
```diff
@@ -24,6 +25,7 @@ internal interface IHttpManager
            bool doNotThrow,
            bool retry,
            X509Certificate2 mtlsCertificate,
+           Func<HttpClient> validateServerCertificateCallback,
```
bgavrilMS, line 28: I'd add a method comment. Some of these params are tricky.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -45,8 +45,42 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
             return new ServiceFabricManagedIdentitySource(requestContext, endpointUri, EnvironmentVariables.IdentityHeader);
         }
 
+        internal override Func<HttpClient> ValidateServerCertificateCallback(RequestContext requestContext)
```
bgavrilMS, line 48: better name might be "CustomHttpClientCreator" but maybe you don't need to use a function pointer / Func here at all?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: I think I discussed something on these lines with @GladwinJohnson and my understanding is that if the x509Certificate2 is there then we do not need this cert validation.

src/client/Microsoft.Identity.Client/Http/IHttpManager.cs
```diff
@@ -8,13 +8,29 @@
...
+        /// </summary>
+        /// <param name="endpoint">The endpoint to send the request to.</param>
+        /// <param name="headers">Headers to send in the http request.</param>
+        /// <param name="body">Body of the request.</param>
+        /// <param name="method">Http method.</param>
+        /// <param name="logger">Logger from the request context.</param>
+        /// <param name="doNotThrow">Flag to decide if MsalServiceException is thrown or the response is returned in case of 5xx errors.</param>
+        /// <param name="retry">Flag to indicate whether the retries are performed in for specific failures.</param>
+        /// <param name="mtlsCertificate">Certificate used for MTLS authentication.</param>
+        /// <param name="customHttpClient">Custom http client which by passes the HttpClientFactory.
```
bgavrilMS, line 30:
```suggestion
        /// <param name="customHttpClient">Custom http client which bypasses the HttpClientFactory. 
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -45,8 +45,35 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
         }
 
+        internal override HttpClient CreateCustomHttpClient(RequestContext requestContext)
+        {
+            return new HttpClient(CreateHandlerWithSslValidation(requestContext.Logger));
+        }
+
+        internal HttpClientHandler CreateHandlerWithSslValidation(ILoggerAdapter logger)
+        {
+#if NET471_OR_GREATER || NETSTANDARD1_3_OR_GREATER || NET
```
bgavrilMS, line 55:
```suggestion
#if NET471_OR_GREATER || NETSTANDARD || NET
```

netstandard2.0 is the only one supported by MSAL. I doubt we will ever support netstandard2.1

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -15,9 +15,9 @@ namespace Microsoft.Identity.Client.ManagedIdentity
     internal class ServiceFabricManagedIdentitySource : AbstractManagedIdentity
     {
         private const string ServiceFabricMsiApiVersion = "2019-07-01-preview";
-
         private readonly Uri _endpoint;
         private readonly string _identityHeaderValue;
+        internal static Lazy<HttpClient> _httpClient = new Lazy<HttpClient>();
```
bgavrilMS, line 20: not used?

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -45,8 +45,35 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
             return new ServiceFabricManagedIdentitySource(requestContext, endpointUri, EnvironmentVariables.IdentityHeader);
         }
 
+        internal override HttpClient CreateCustomHttpClient(RequestContext requestContext)
+        {
+            return new HttpClient(CreateHandlerWithSslValidation(requestContext.Logger));
```
bgavrilMS, line 50: I think you started looking at reusing this HttpClient with the Lazy<T> version, but didn't wrap it up.

bgavrilMS (changes requested):
Current update creates a new HttpClient for each request, which is problematic. Please complete the Lazy<T> approach you started or use a static.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -146,14 +149,18 @@ internal class HttpManager : IHttpManager
             return response;
         }
 
-        private HttpClient GetHttpClient(X509Certificate2 x509Certificate2)
-        {
-            if (_httpClientFactory is IMsalMtlsHttpClientFactory msalMtlsHttpClientFactory)
+        private HttpClient GetHttpClient(X509Certificate2 x509Certificate2, HttpClient customHttpClient) {
```
bgavrilMS, line 152:
```suggestion
        private HttpClient GetHttpClient(X509Certificate2 mtlsCertificate, HttpClient customHttpClient) {
```

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -146,14 +149,18 @@ internal class HttpManager : IHttpManager
...
+            if (x509Certificate2 != null && _httpClientFactory is IMsalMtlsHttpClientFactory msalMtlsHttpClientFactory)
             {
                 // If the factory is an IMsalMtlsHttpClientFactory, use it to get an HttpClient with the certificate
                 return msalMtlsHttpClientFactory.GetHttpClient(x509Certificate2);
             }
```
bgavrilMS, line 158: if both certificate and customHttpClient are configured, I would add a `Debug.Assert("Not supported. Ignoring the customHttpClient`) and log.error(). Or throw an exception like NotImplementedException

bgavrilMS (dismissed):

gladjohn (approved):
looks Neha, were you able to validate this on a service fabric environment?

neha-bhargava:
> looks Neha, were you able to validate this on a service fabric environment?

No, since the setup was not easy it was decided to rely on the unit tests instead

PR #4842, opened 2024-07-15
[Bug Fix] Correct alg to PS256 for SHR POP Tokens

### LINKED ISSUE #4839: [Bug] SHR POP tokens are broken due to invalid alg (it should be PS256 not RS256)
### Library version used

4.60.3

### .NET version

net 8.0

### Scenario

I think the signed pop token produced by Microsoft.Identity.Client is not specifying the correct cryptographic algorithm in the JsonWebToken header.

I have a client application that uses ConfidentialClientApplicationBuilder AcquireTokenForClient to generate a signed pop token. Then my server application uses SignedHttpRequestHandler ValidateSignedHttpRequestAsync to validate the signature.

This code was working fine up through M.I.C version 4.59.1: the token generated by the client could be validated successfully in the server. However, in version 4.60.3 (the next one we tried), this validation starts failing with an error on signature validation (IDX23034).

Debugging deep into the validation code, I have figured out that the failure is because the validation code examines the algorithm from the JWT header and determines that the signature padding is Pkcs1. But with version 4.60 and higher, actually it is Pss.

To reproduce, use 4.60.3 or more recent to generate a signed pop token and then run ValidateSignedHttpRequestAsync on it.

I can provide sample code if necessary--I have a standalone application that generates a pop token and then validates it--but I think this issue can be seen just by reading the project code.

---

**More details**

I am pretty sure the issue stems from a recent fix in InMemoryCryptoProvider to use PSS for the padding. See PR https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4633.

The code now reads:

        public byte[] Sign(byte[] payload)
        {
            return _signingKey.SignData(
                payload,
                HashAlgorithmName.SHA256,
                **RSASignaturePadding.Pss);**
        }

However, the crypto provider still returns "RS256" for the CryptographicAlgorithm:

        public string CryptographicAlgorithm { get => "RS256"; }

I think it needs to return "PS256" to get Pss signature padding (I got that value from https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/wiki/Supported-Algorithms).

Here's how it all goes down:

The value returned from CryptographicAlgorithm gets put into the JsonWebToken header in the "alg" field. Later when the signature is being verified in the server (using SignedHttpRequestHandler ValidateSignatureAsync), the alg field is read and used to determine the padding on the signature field. Padding determination is done in this code (from [AsymmetricAdapter.cs](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/3b3d3ec6417885b686625004adfd8fb7dc3442a2/src/Microsoft.IdentityModel.Tokens/AsymmetricAdapter.cs#L187)):

            if (algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha256) ||
                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha256Signature) ||
                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha384) ||
                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha384Signature) ||
                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha512) ||
                algorithm.Equals(SecurityAlgorithms.RsaSsaPssSha512Signature))
            {
                RSASignaturePadding = RSASignaturePadding.Pss;
            }
            else
            {
                // default RSASignaturePadding for other supported RSA algorithms is Pkcs1
                RSASignaturePadding = RSASignaturePadding.Pkcs1;
            }

Since the header alg field is not any of the "Pss" versions, the padding is set to Pkcs1. Then the signature validation fails because it's not using the correct padding.

### Relevant code snippets

We are building the confidential client app as:
var uamiCcApp = ConfidentialClientApplicationBuilder
    .Create(sniClientId)
    .WithCertificate(cert, true)
    .WithTenantId(tenantId)
    .WithAuthority(new Uri(authConfig.Authority))
    .WithExperimentalFeatures() // for PoP
    .Build();

and the pop token as:
                    popConfig.HttpHost = host.Replace("https://", "");
                    popConfig.HttpMethod = httpMethod;
                    popConfig.HttpPath = path;
                    popConfig.SignHttpRequest = true;

                    result = await ccApp.AcquireTokenForClient(scopes)
                        .WithSendX5C(true)
                        .WithProofOfPossession(popConfig)
                        .ExecuteAsync();

### Expected behavior

The signature validates correctly.

### Identity provider

### Regression

MSAL Version: 4.59.1

### Solution and workarounds

I think the solution is to modify InMemoryCryptoProvider to return PS256 for the algorithm:

public string CryptographicAlgorithm { get => "PS256"; }

I am trying this solution out locally by implementing a custom crypto provider, and so far it appears to work.

gladjohn:
Fixes #4839

1. **Update `InMemoryCryptoProvider`**:
   - Modify the `CryptographicAlgorithm` property to return `PS256` instead of `RS256`.

2. **Update Test Cases**:
   - Add tests to verify that the `alg` in the JWT header is correctly set to `PS256`.
   
**Testing**
- Ran existing unit tests to ensure no regressions.
- Added new tests to specifically check for the correct `alg` value in the SHR POP tokens.
- Did some manual testing
- Gave a private build for partner testing

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/AuthScheme/PoP/InMemoryCryptoProvider.cs
```diff
@@ -58,5 +58,10 @@ private static string ComputeCanonicalJwk(RSAParameters rsaPublicKey)
         {
             return $@"{{""{JsonWebKeyParameterNames.E}"":""{Base64UrlHelpers.Encode(rsaPublicKey.Exponent)}"",""{JsonWebKeyParameterNames.Kty}"":""{JsonWebAlgorithmsKeyTypes.RSA}"",""{JsonWebKeyParameterNames.N}"":""{Base64UrlHelpers.Encode(rsaPublicKey.Modulus)}""}}";
         }
+
+        internal RSAParameters GetPublicKeyParameters()
```
neha-bhargava, line 62: Add a comment that this method is used for testing only

src/client/Microsoft.Identity.Client/AuthScheme/PoP/InMemoryCryptoProvider.cs
```diff
@@ -24,7 +24,7 @@ public InMemoryCryptoProvider()
 
         public string CannonicalPublicKeyJwk { get; private set; }
 
-        public string CryptographicAlgorithm { get => "RS256"; }
+        public string CryptographicAlgorithm { get => "PS256"; }
```
neha-bhargava, line 27: Is this true for both netframework and otherwise?

neha-bhargava, line 27: Can we add tests to verify the "PS256" value of the `CryptographicAlgorithm`?

neha-bhargava: The test got added to the PR as I was adding this comment seems like :)

neha-bhargava (approved):
Some small comments / questions

src/client/Microsoft.Identity.Client/AuthScheme/PoP/InMemoryCryptoProvider.cs
gladjohn: Sorry my bad, added the test but did not push it yet, yes we now have a test

gladjohn: Would really need @trwalke or @bgavrilMS to look at this one and comment

bgavrilMS: I think ComputeCanonicalJwk needs to be updated as well.

gladjohn: let me update and add a test for it as well

gladjohn: @bgavrilMS can you please check to make sure if the update is correct? if not can you please suggest the correct update for ComputeCanonicalJwk  with PS256

bgavrilMS (changes requested):
Let's please add an integration test for this, for both .NET and .NET fwk.

Graph officially supports SHR POP, just try to access https://graph.microsoft.com/v1.0/me/ 
and you will see the WWW-Authenticate headers. Just use the WWW-Authenticate parser to extract the nonce.

image

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/PoPTests.NetFwk.cs
```diff
@@ -541,6 +541,98 @@ public async Task InMemoryCryptoProvider_AlgIsPS256()
...
+            using (var sha256 = SHA256.Create())
+            {
+                var kidBytes = sha256.ComputeHash(Encoding.UTF8.GetBytes(canonicalJwk));
+                expectedKid = Base64UrlHelpers.Encode(kidBytes);
+            }
+
+            // Use InMemoryCryptoProvider
+            var popConfig = new PoPAuthenticationConfiguration(new Uri(ProtectedUrl))
+            {
+                PopCryptoProvider = cryptoProvider,
```
bgavrilMS, line 573: I don't think you need this, you can rely on the in-memory one

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/PoPTests.NetFwk.cs
```diff
@@ -541,6 +541,98 @@ public async Task InMemoryCryptoProvider_AlgIsPS256()
...
+
+            // Integration Test: Call the Graph API and test for success
+            var httpClient = new HttpClient();
+            var response = await httpClient.GetAsync("https://graph.microsoft.com/v1.0/users").ConfigureAwait(false);
+
+            // Check for WWW-Authenticate header
+            Assert.IsTrue(response.StatusCode == HttpStatusCode.Unauthorized, "The response should be Unauthorized (401)");
+
+            // Extract WWW-Authenticate header to get the nonce
+            var authParams = await WwwAuthenticateParameters.CreateFromAuthenticationResponseAsync(
```
bgavrilMS, line 606: I think this makes the Http call again. There is no need for that.

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/PoPTests.NetFwk.cs
bgavrilMS: Oh, I guess you defined one to be able to validate some stuff,

gladjohn: The e2e does not work and the test is failing with the graph endpoint. Looks like and email thread has been started already to unblock

src/client/Microsoft.Identity.Client/AuthScheme/PoP/InMemoryCryptoProvider.cs
trwalke: Yeah, this looks right to me according to the spec

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/PoPTests.NetFwk.cs
trwalke: can you add me to that email thread?

src/client/Microsoft.Identity.Client/AuthScheme/PoP/InMemoryCryptoProvider.cs
trwalke: actually, i dont think we need to add it to the conical JWK because it is not a required member for the computation. This is probably why we didn't have "alg":"RSA256" to begin with. Since PS256 is essentially RS256 with PSS padding, I would imagine the required members for the JWK public key would be the same but not entirely sure. I doudt we need to add the "alg":"PS256" though

image

gladjohn: checking with ESTS

gladjohn: @trwalke per ESTS "Algorithm is not part of JWK, same key can be used to produce multiple different signature, like here RS256 or PS256"

gladjohn: `alg` parameter is not included, as it is not necessary for the canonical representation used for computing the thumbprint

By including only the required parameters (e, kty, and n) in lexicographic order, this method ensures that the JWK is correctly canonicalized for thumbprint computation as specified in RFC 7638.

bgavrilMS (approved):

trwalke (approved):

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/PoPTests.NetFwk.cs
gladjohn: it was a chat let me forward it to you

PR #4848, opened 2024-07-19
Remove AccountSource from public IAccount interface to avoid breaking change

ashok672:
The previous check in to fix a device code flow issue unintentionally introduced a breaking change by adding a new property in IAccount public interface. Remove the new property in IAccount and only depend on the internal Account object for the new property.

**Testing**
The Integration test has been updated to test the change.

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
@@ -61,8 +61,9 @@ namespace Microsoft.Identity.Client.Internal.Requests.Silent
                        UiRequiredExceptionClassification.AcquireTokenSilentFailed);
                 }
 
-                bool isAccountSourceDeviceCodeFlow = !string.IsNullOrEmpty(AuthenticationRequestParameters.Account.AccountSource) &&
-                                               AuthenticationRequestParameters.Account.AccountSource == "device_code_flow";
+                var account = AuthenticationRequestParameters.Account as Account;
+                bool isAccountSourceDeviceCodeFlow = !string.IsNullOrEmpty(account.AccountSource) &&
+                                               account.AccountSource == "device_code_flow";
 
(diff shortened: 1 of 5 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
```diff
@@ -61,8 +61,9 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
                        UiRequiredExceptionClassification.AcquireTokenSilentFailed);
                 }
 
-                bool isAccountSourceDeviceCodeFlow = !string.IsNullOrEmpty(AuthenticationRequestParameters.Account.AccountSource) &&
-                                               AuthenticationRequestParameters.Account.AccountSource == "device_code_flow";
+                var account = AuthenticationRequestParameters.Account as Account;
+                bool isAccountSourceDeviceCodeFlow = !string.IsNullOrEmpty(account.AccountSource) &&
```
bgavrilMS, line 65: There can be a null pointer ex here if account is null. 

Try smth like `bool isAccountSourceDeviceCodeFlow = account != null && account.AccountSource == "device_code_flow"`

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/Silent/SilentRequest.cs
pmaytak, line 65: `AccountSource` property can be changed to an enum of type `ApiIds`

ashok672: I will leave it to a free from string for now. Also MSAL.NET only has to skip broker for device code flow. ROPC is already through broker in MSAL.NET. I in fact want to get rid of this workarounds in the future when device code flow is deprecated and not make it easy to introduce the workarounds for any other flow by mistake.

neha-bhargava (approved):

PR #4856, opened 2024-07-23
Update the logic to add file based detection for azure arc for linux env

### LINKED ISSUE #4358: MSI support for ARC on Linux
### Library version used

4.56.0

### .NET version

net6.0

### Scenario

I'm developing a new [Powershell module](https://github.com/SCOMnewbie/PSMSALNet) based on the MSAL.net library and I'm testing several scenarios. One of them is being able to generate an access token from an enrolled ARC Linux (tested on several Ubuntu 22.04). I don't think the problem comes from the Powershell module itself because thanks to your nice work, the code is pretty obvious. 

On Windows ARC server it's working well. 

The problem is not on the ARC side because I was able to generate tokens using "other methods". Here the error message I received from the prompt:

image

### Relevant code snippets

```csharp
$ClientApplicationBuilder = [Microsoft.Identity.Client.ManagedIdentityApplicationBuilder]::Create([Microsoft.Identity.Client.AppConfig.ManagedIdentityId]::SystemAssigned)
$ClientApplication = $ClientApplicationBuilder.Build()
$AquireTokenParameters = $ClientApplication.AcquireTokenForManagedIdentity($Scopes)
and then generate the call to get the token
```

### Expected behavior

Should return an access token for the KeyVault scope in this case.

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

neha-bhargava:
Fixes #4358 

Add file based detection in addition to env variables when env variables are not found.

**Testing**

**Performance impact**

**Documentation**


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -81,5 +83,31 @@ internal static ManagedIdentitySource GetManagedIdentitySource()
                 return ManagedIdentitySource.DefaultToImds;
             }
         }
+
+        // Method to return true if a file exists and is not empty to validate the Azure arc environment.
+        private static bool ValidateAzureArcEnvironment(string identityEndpoint, string imdsEndpoint)
+        {
+            if (!string.IsNullOrEmpty(identityEndpoint) || !string.IsNullOrEmpty(imdsEndpoint))
```
bgavrilMS, line 90: Shouldn't this be && as per original logic?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -81,5 +83,31 @@ internal static ManagedIdentitySource GetManagedIdentitySource()
                 return ManagedIdentitySource.DefaultToImds;
             }
         }
+
+        // Method to return true if a file exists and is not empty to validate the Azure arc environment.
+        private static bool ValidateAzureArcEnvironment(string identityEndpoint, string imdsEndpoint)
```
bgavrilMS, line 88: Feels like this logic should hvae some log statements in it.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/AzureArcManagedIdentitySource.cs
```diff
@@ -29,7 +29,16 @@ internal class AzureArcManagedIdentitySource : AbstractManagedIdentity
 
         public static AbstractManagedIdentity Create(RequestContext requestContext)
         {
-            string identityEndpoint = EnvironmentVariables.IdentityEndpoint;
+            string identityEndpoint;
+                
+            if (EnvironmentVariables.IdentityEndpoint == null)
```
gladjohn, line 34: When `AzureArcManagedIdentitySource` gets executed, it will always return true because the default endpoint is set and will never execute IMDS?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: This is a static method that can also be called from the static method to get source. The request context is not yet created in that case. I added the logs later in the flow when we are instantiating the azure arc object. I can add an optional variable for a logger here and log only when the logger is available.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: Yes, good catch. Thanks :)

trwalke (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -81,5 +84,38 @@ internal static ManagedIdentitySource GetManagedIdentitySource()
...
+            {
+                logger?.Verbose(() => "[Managed Identity] Azure Arc managed identity is available through environment variables.");
+                return true;
+            }
+
+            string path;
+
+            if (DesktopOsHelper.IsWindows())
+            {
+                path = Environment.ExpandEnvironmentVariables("%Programfiles%\\AzureConnectedMachineAgent\\himds.exe");
```
Robbie-Microsoft, line 101: Q: Should the "f" in "Programfiles" be capitalized?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -46,7 +49,7 @@ private static AbstractManagedIdentity SelectManagedIdentitySource(RequestContex
         // Detect managed identity source based on the availability of environment variables.
         // The result of this method is not cached because reading environment variables is cheap. 
         // This method is perf sensitive any changes should be benchmarked.
-        internal static ManagedIdentitySource GetManagedIdentitySource()
+        internal static ManagedIdentitySource GetManagedIdentitySource(ILoggerAdapter logger = null)
```
bgavrilMS, line 52: not a fan of this "maybe null" approach for objects. Can't the product code guarantee it is non-null all the time? Tests can inject mocks if needed.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -81,5 +84,38 @@ internal static ManagedIdentitySource GetManagedIdentitySource()
...
+            {
+                path = Environment.ExpandEnvironmentVariables("%Programfiles%\\AzureConnectedMachineAgent\\himds.exe");
+            }
+            else if (DesktopOsHelper.IsLinux())
+            {
+                path = "/opt/azcmagent/bin/himds";
+            }
+            else
+            {
+                return false;
```
bgavrilMS, line 109: Consider adding a logging statement, maybe at warning level. As we don\t expect this logic to ever run elsewhere.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
neha-bhargava: This is not for the tests. This method is called from 2 places. One when the request context and everything is available and the other is a static method from ManagedIdentityApplication. The logger is not available at that point.

Issue #4868, opened 2024-08-01
[Bug] SHA2 changes break custom claims


localden:
### Library version used

4.62.0

### .NET version

Cross-framework.

### Scenario

ConfidentialClient - service to service (AcquireTokenForClient)

### Is this a new or an existing app?

The app is in production, and I have upgraded to a new version of MSAL

### Issue description and reproduction steps

In this PR ([Use SHA2 and PSS for client assertion (#4616)](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/commit/d2f015ad1daa38af744f66adda3cc6abec1cdc8b?diff=split&w=0)), MSAL totally revamps how it processes client claims and encodes them into the request token. For the Network Security Perimeter feature team, they use a client claim called "custom_claims" that can look like this:

```json
 “customClaims”: { 
     “xms_az_nwperimid”: ["GUID", "GUID2", "GUID3"]} 
```

The new code explicitly removes handling the client claim values as JSON and instead treats them as strings, putting double quotes around whatever is passed. 

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

bgavrilMS:
Hi @trwalke - any updates on this?

trwalke:
Taking a look @localden

PR #4886, opened 2024-08-09
Changing the handling of client claims to use JSON

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4868

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
@@ -8,6 +8,11 @@ using System.Text;
 using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
 using Microsoft.Identity.Client.Utils;
 using System.Security.Cryptography;
+#if SUPPORTS_SYSTEM_TEXT_JSON
+using JObject = System.Text.Json.Nodes.JsonObject;
+#else
+using Microsoft.Identity.Json.Linq;
+#endif
 
 namespace Microsoft.Identity.Client.Internal
 {
@@ -65,17 +70,16 @@ namespace Microsoft.Identity.Client.Internal
                 payload.Append('{');
             }
 
-            int i = 0;
-            foreach (var kvp in _claimsToSign)
+            var json = new JObject();
+
+            foreach (var claim in _claimsToSign)
             {
-                payload.Append($"\"{kvp.Key}\":\"{kvp.Value}\"");
-                
-                if (i!= _claimsToSign.Count-1)
-                {
-                    i++;
-                    payload.Append(',');
-                }
+                json[claim.Key] = claim.Value;
             }
+
+            var jsonClaims = JsonHelper.JsonObjectToString(json);
+
+            payload.Append(jsonClaims.Substring(1, jsonClaims.Length - 2));
 
             payload.Append('}');
 
--- a/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
@@ -79,6 +79,7 @@ namespace Microsoft.Identity.Client.Internal
 
             var jsonClaims = JsonHelper.JsonObjectToString(json);
 
+            //Remove extra brackets from JSON result
             payload.Append(jsonClaims.Substring(1, jsonClaims.Length - 2));
 
             payload.Append('}');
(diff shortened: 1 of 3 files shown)


trwalke:
@localden @bgavrilMS I switched the parsing for custom claims back to the JSON parsing. Here is the change in parsing JSON strings 

The input KVP is `{ "customClaims", "{\"xms_az_claim\": [\"GUID\", \"GUID2\", \"GUID3\"]}" }`

Current: `"customClaims":"{"xms_az_claim": ["GUID", "GUID2", "GUID3"]}"`
New (fix): `"customClaims":"{\"xms_az_claim\": [\"GUID\", \"GUID2\", \"GUID3\"]}"`

Is this what we are looking for? I want to be sure this is exactly what we need to send to ESTS.

src/client/Microsoft.Identity.Client/Internal/JsonWebToken.cs
```diff
@@ -65,18 +70,17 @@ private string CreateJsonPayload()
                 payload.Append('{');
             }
 
-            int i = 0;
-            foreach (var kvp in _claimsToSign)
+            var json = new JObject();
+
+            foreach (var claim in _claimsToSign)
             {
-                payload.Append($"\"{kvp.Key}\":\"{kvp.Value}\"");
```
trwalke, line 71: What was the reason for switching this to manual string parsing? Performance right? Since we are no longer parsing the default claims with JSON, but only the custom claims, I would imagine this approach will keep some of the performance gains depending on how many claims are added by the client app.

The reason I went with this approach is the JSON api will handle all of the edge cases we may miss trying to parse manually.

bgavrilMS: SOunds good, thanks Travis. Just add some unit tests for this case, they are clearly missing.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithCertTest.cs
```diff
@@ -506,6 +506,8 @@ public void ClientAssertionTests(bool sendX5C, bool useSha2AndPss, bool addExtra
             {
                 Assert.AreEqual("Val1", decodedToken.Payload["Key1"]);
                 Assert.AreEqual("Val2", decodedToken.Payload["Key2"]);
+                //Ensure JSON formatting is preserved
+                Assert.AreEqual("{\"xms_az_claim\": [\"GUID\", \"GUID2\", \"GUID3\"]}", decodedToken.Payload["customClaims"]);
```
bgavrilMS, line 510: LGTM, anything else to add ? The PR is still a draft.

bgavrilMS (approved):

localden (approved):

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithCertTest.cs
trwalke: Nope, I marked as ready

bgavrilMS:
> @localden @bgavrilMS I switched the parsing for custom claims back to the JSON parsing. Here is the change in parsing JSON strings
> 
> The input KVP is `{ "customClaims", "{\"xms_az_claim\": [\"GUID\", \"GUID2\", \"GUID3\"]}" }`
> 
> Current: `"customClaims":"{"xms_az_claim": ["GUID", "GUID2", "GUID3"]}"` New (fix): `"customClaims":"{\"xms_az_claim\": [\"GUID\", \"GUID2\", \"GUID3\"]}"`
> 
> Is this what we are looking for? I want to be sure this is exactly what we need to send to ESTS.

Please see the behavior on an older MSAL to confirm.

rayluo:
> > @localden @bgavrilMS I switched the parsing for custom claims back to the JSON parsing. Here is the change in parsing JSON strings
> > The input KVP is `{ "customClaims", "{\"xms_az_claim\": [\"GUID\", \"GUID2\", \"GUID3\"]}" }`
> > Current: `"customClaims":"{"xms_az_claim": ["GUID", "GUID2", "GUID3"]}"`
> > New (fix): `"customClaims":"{\"xms_az_claim\": [\"GUID\", \"GUID2\", \"GUID3\"]}"`
> > Is this what we are looking for? I want to be sure this is exactly what we need to send to ESTS.
> 
> Please see the behavior on an older MSAL to confirm.

* Since the input value `"{\"xms...\": ...}"` was in this case a string (regardless of whether it contains escaped quotation mark), the output is rightfully also the same string (with valid escaping).
* Out of curiosity, could the input be a json blob `{"xms_az_claim": ["GUID", "GUID2", "GUID3"]}`, and if so, will our implementation generate the output as the same json blob? [It probably will](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4886/files#diff-f0d569d5c86679d53fd6a97ff8fa9030952e9511c986afd750b7c64e3263e506R77-R80), but I just want to call this out and ask.

PR #4898, opened 2024-08-19
Prototype extensiblity with example

bgavrilMS:
Please review directly the tests/Microsoft.Identity.Test.Unit/PublicApiTests/ExtensiblityTests.cs to see an example

The src/client/Microsoft.Identity.Client/Extensibility/MsalAddIn.cs has been added as an agreegation of mulitple extensiblity points in MSAL. 

TODO: telemetry signals like time to create CDT

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
@@ -67,11 +67,12 @@ namespace Microsoft.Identity.Client.AuthScheme.PoP
             };
         }
 
-        public string FormatAccessToken(MsalAccessTokenCacheItem msalAccessTokenCacheItem)
+        public void FormatResult(AuthenticationResult authenticationResult)
         {
             if (!_popAuthenticationConfiguration.SignHttpRequest)
             {
-                return msalAccessTokenCacheItem.Secret;
+                // token will be signed by the caller
+                return;
             }
 
             var header = new JObject();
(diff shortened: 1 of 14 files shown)


src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs
```diff
@@ -67,25 +67,26 @@ public IReadOnlyDictionary<string, string> GetTokenRequestParams()
             };
         }
 
-        public string FormatAccessToken(MsalAccessTokenCacheItem msalAccessTokenCacheItem)
+        public void FormatResult(AuthenticationResult authenticationResult)
```
trwalke, line 70: I think this will work, now that the additionalCacheParameters are stored in the `AuthenticaitonResult.AdditionalResponseParameters`, the ds_req_cnf and the nonce will be available to use.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -22,15 +24,48 @@ public static class AbstractConfidentialClientAcquireTokenParameterBuilderExtens
...
 
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="builder"></param>
+        /// <param name="addIn"></param>
+        /// <returns></returns>
+        public static AbstractAcquireTokenParameterBuilder<T> WithAddIn<T>( // TODO: bogavril - support a list of add-ins ?
```
trwalke, line 49: Not sure a list is needed. The question will be "in what order do we apply the add ins?" Will customers be aware of what order to use? Seems like it can be a pain point down the road. Probably should limit it to only allowing one add in for now?

trwalke:
My biggest question here is, we know that the SHR scheme will come after the provided one (CDT) one, but will that always be the case? Should we default the behavior of add-ins be to always be performed first? Does it make sense to modify the add-in api to allow developers to specify that it should happen after the other internal MSAL schemes

bgavrilMS:
> My biggest question here is, we know that the SHR scheme will come after the provided one (CDT) one, but will that always be the case? Should we default the behavior of add-ins be to always be performed first? Does it make sense to modify the add-in api to allow developers to specify that it should happen after the other internal MSAL schemes

I guess not. We could think of an ordering hint, like "BeforeAuthScheme = pop" or "AuthSchemeOrder = 10"

PR #4917, opened 2024-09-04
Fix for 4913 - sendX5C for ROPC CCA

### LINKED ISSUE #4913: [Bug] X5C not being sent with the API call for ROPC flow with confidential clients
### Library version used

4.64.0

### .NET version

4.7.2

### Scenario

Other - please specify

When making a ROPC API call with confidential clients (using X5C overload with certificates), the certificate chain information is not being sent via MSAL.

Reproduction steps:
1. Create a confidential client with a certificate.
2. Call the ROPC API using the above client.
3. MSAL throws error indicating X5C is not being sent.

### Relevant code snippets

```csharp
Microsoft.Identity.Client.IConfidentialClientApplication authContext = Microsoft.Identity.Client.ConfidentialClientApplicationBuilder.Create(clientId)
    .WithCertificate(certificate, sendX5C: true)
    .WithAuthority(authority)
    .Build();
Microsoft.Identity.Client.AuthenticationResult authResult = await (((Microsoft.Identity.Client.IConfidentialClientApplication)authContext) as Microsoft.Identity.Client.IByUsernameAndPassword).AcquireTokenByUsernamePassword(new string[] { scope }, fullUserPrincipalName, usersPassword).ExecuteAsync();
accessToken = authResult?.AccessToken;
```

### Expected behavior

X5C certificate chain information is sent via MSAL and token can be acquired using SNI.

### Identity provider

### Regression

_No response_

### Solution and workarounds

Ensure X5C is sent or add an extension method to send X5C info similar to acquire token for client.

bgavrilMS:
Fixes #4913

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder.cs
@@ -18,7 +18,7 @@ using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
 namespace Microsoft.Identity.Client
 {
     /// <summary>
-    /// Parameter builder for the <see cref="IConfidentialClientApplication.AcquireTokenByUsernamePassword(IEnumerable{string}, string, string)"/>
+    /// Parameter builder for the <see cref="IByUsernameAndPassword.AcquireTokenByUsernamePassword(IEnumerable{string}, string, string)"/>
     /// operation. See https://aka.ms/msal-net-up
     /// </summary>
     public sealed class AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder :
@@ -66,6 +66,17 @@ namespace Microsoft.Identity.Client
         internal override ApiEvent.ApiIds CalculateApiEventId()
         {
             return ApiEvent.ApiIds.AcquireTokenByUsernamePassword;
+        }
+
+        /// <inheritdoc/>
+        protected override void Validate()
+        {
+            base.Validate();
+
+            if (Parameters.SendX5C == null)
+            {
+                Parameters.SendX5C = ServiceBundle.Config?.SendX5C ?? false;
+            }
         }
     }
 }
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Executors/ConfidentialClientExecutor.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Executors/ConfidentialClientExecutor.cs
@@ -142,6 +142,8 @@ namespace Microsoft.Identity.Client.ApiConfig.Executors
                 commonParameters,
                 requestContext,
                 _confidentialClientApplication.UserTokenCacheInternal).ConfigureAwait(false);
+            
+            requestParams.SendX5C = usernamePasswordParameters.SendX5C ?? false;
 
             var handler = new UsernamePasswordRequest(
                 ServiceBundle,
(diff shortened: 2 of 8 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder.cs
```diff
@@ -67,5 +67,16 @@ internal override ApiEvent.ApiIds CalculateApiEventId()
...
         }
+
+        /// <inheritdoc/>
+        protected override void Validate()
+        {
+            base.Validate();
+
+            if (Parameters.SendX5C == null)
+            {
+                Parameters.SendX5C = ServiceBundle.Config?.SendX5C ?? false;
```
trwalke, line 78: Why wouldnt this be set globally when it is enabled on the `WithCertificate(...)` method?

Also, should we add `WithSendX5C()` to this class?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithCertTest.cs
```diff
@@ -669,6 +669,41 @@ await app.AcquireTokenByAuthorizationCode(TestConstants.s_scope, TestConstants.D
...
+        {
+            using (var harness = CreateTestHarness())
+            {
+                var certificate = CertHelper.GetOrCreateTestCert();
+                var exportedCertificate = Convert.ToBase64String(certificate.Export(X509ContentType.Cert));
+
+                var app = ConfidentialClientApplicationBuilder
+                    .Create(TestConstants.ClientId)
+                    .WithHttpManager(harness.HttpManager)
+                    .WithCertificate(certificate, sendX5C)
```
trwalke, line 687: Should also test `WithSendX5C`

bgavrilMS: It's not exposed.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder.cs
bgavrilMS: As for `Why wouldnt this be set globally when it is enabled on the WithCertificate(...) method?` question. It would require quite a big refactoring to do this and I'm not sure it's worth it, as we don't add new flows except very rarely. It would also introduce (small) breaking changes, which I don't think it's worth risking.

- SendX5C is defined on `AbstractAcquireTokenConfidentialClientParameters`, but ROPCParameters doesn't inherit this
- `AbstractConfidentialClientAcquireTokenParameterBuilder` doesn't have access to  `AbstractAcquireTokenConfidentialClientParameters` to set this in Validate()

src/client/Microsoft.Identity.Client/ApiConfig/Executors/ConfidentialClientExecutor.cs
```diff
@@ -142,6 +142,8 @@ public async Task<AuthenticationResult> ExecuteAsync(
                 commonParameters,
                 requestContext,
                 _confidentialClientApplication.UserTokenCacheInternal).ConfigureAwait(false);
+            
+            requestParams.SendX5C = usernamePasswordParameters.SendX5C ?? false;
```
rayluo, line 146: I was also wondering whether we could/should put this setting into a central place when `WithCertificate(...)` is called, but then I realized that it has been discussed and [the conclusion](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4917/files#r1747523539) makes sense.

Ship it!

rayluo (approved):

PR #4929, opened 2024-09-12
Add more logging to GetAccounts by id

bgavrilMS:
Fixes #

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
+++ b/src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
@@ -38,8 +38,8 @@ namespace Microsoft.Identity.Client
 
         internal ClientApplicationBase(ApplicationConfiguration config) : base(config)
         {
-            UserTokenCacheInternal = 
-                config.UserTokenCacheInternalForTest ?? 
+            UserTokenCacheInternal =
+                config.UserTokenCacheInternalForTest ??
                 new TokenCache(
                     ServiceBundle,
                     false,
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
```diff
@@ -171,6 +171,15 @@ private async Task<IEnumerable<IAccount>> GetAccountsInternalAsync(ApiIds apiId,
         {
             Guid correlationId = Guid.NewGuid();
             RequestContext requestContext = CreateRequestContext(correlationId, cancellationToken);
+            
+            if (requestContext.Logger.IsLoggingEnabled(LogLevel.Info))
```
trwalke, line 175: So we usually check this before creating a logger message that has quite a bit of overhead. This doesn't seem to be too taxing so cant we let the logger adapter determine if this should be logged or not?

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ClientCredentialsTests.WithRegion.cs
```diff
@@ -125,17 +125,18 @@ private IConfidentialClientApplication BuildCCA(
...
             var builder = ConfidentialClientApplicationBuilder.Create(settings.ClientId);
-            if (useClaims)
-            {
-                builder.WithClientAssertion(() => GetSignedClientAssertionUsingMsalInternal(settings.ClientId, GetClaims(settings)));
-            }
-            else
-            {
-                builder.WithCertificate(settings.GetCertificate());
-            }
+            //if (useClaims)
```
trwalke, line 128: Dead code?

trwalke (approved):
Minor comments

src/client/Microsoft.Identity.Client/ClientApplicationBase.cs
bgavrilMS: This is in the hot path ... Any optimization counds. Besides at line 179-180 all those string interpolations are not free and will lead to some allocations etc.

gladjohn (approved):

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ClientCredentialsTests.WithRegion.cs
bgavrilMS: Good catch, thx.

PR #4976, opened 2024-10-23
Unobsolete GetAccountsAsync

neha-bhargava:
Fixes #

Remove the obsolete attribute from the GetAccountsAsync method in the IConfidentialClientApplication interface to make it available for ROPC silent flow in Id web

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
+++ b/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
@@ -94,6 +94,7 @@ namespace Microsoft.Identity.Client
         new AcquireTokenSilentParameterBuilder AcquireTokenSilent(IEnumerable<string> scopes, string loginHint);
 
         /// <summary>
+        /// This method is strictly to be used with ROPC flow in test environments. It is not recommended to use ROPC in production applications.
         /// Use <see cref="IClientApplicationBase.GetAccountAsync(string)"/> in web apps and web APIs, and use a token cache serializer for better security and performance. See https://aka.ms/msal-net-cca-token-cache-serialization.
         /// </summary>
         new Task<IEnumerable<IAccount>> GetAccountsAsync();
--- a/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
+++ b/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
@@ -96,8 +96,6 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// Use <see cref="IClientApplicationBase.GetAccountAsync(string)"/> in web apps and web APIs, and use a token cache serializer for better security and performance. See https://aka.ms/msal-net-cca-token-cache-serialization.
         /// </summary>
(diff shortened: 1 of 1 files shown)


trwalke (approved):

bgavrilMS (approved):

bgavrilMS:
@jmprieur and @localden - can you pls provide a perspective on this? `GetAccounts` is still used safely in all other MSALs in both web app and ROPC scenarios.

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
```diff
@@ -96,8 +96,6 @@ AcquireTokenByAuthorizationCodeParameterBuilder AcquireTokenByAuthorizationCode(
         /// <summary>
         /// Use <see cref="IClientApplicationBase.GetAccountAsync(string)"/> in web apps and web APIs, and use a token cache serializer for better security and performance. See https://aka.ms/msal-net-cca-token-cache-serialization.
```
bgavrilMS, line 97: Pls add a remark that this is only for use with ROPC

localden:
@bgavrilMS @neha-bhargava - given that this is only for ROPC in CCA, no concerns on my part. This will effectively just list out the accounts in the cache?

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
jmprieur: we have no way to enforce that it's only used for ROPC in CCA, though
Are we sure that the cache is always partitioned?

neha-bhargava:
> @bgavrilMS @neha-bhargava - given that this is only for ROPC in CCA, no concerns on my part. This will effectively just list out the accounts in the cache?

Yes, this will list the accounts in the cache. Only recommended for ROPC.

src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs
neha-bhargava: AFAIK the cache is always partitioned.

bgavrilMS: It is partitioned. But please add a unit test (or update the integration test) @neha-bhargava, where you show that TokenCacheNotificationArgs has the correct `SuggestedCacheKey`.

bgavrilMS: @neha-bhargava and I had a chat and concluded that `AcquireTokenForUsernamePassword` is partitioned, but `GetAccounts()` is not partitioned (and can't really be partitioned).

From MSAL perspective, we can ask ppl to use `GetAccount(id)`. 
For Id.Web, this gets a bit tricky.

jmprieur: How would that work in MSAL? where do they get the id from? (assuming this is tid.oid?)?

jmprieur: Proposing that CCA.AcquireTokenByUserPassword first checks the cache, and if not found or error, goes to ESTS.
Like OBO does.

bgavrilMS (changes requested):
We decided against it.

neha-bhargava:
Not needed anymore.

PR #4987, opened 2024-11-07
Remove MSAL v3 migration aid

bgavrilMS:
Internal

Remove some migration AID filed, as MSAL v3 usage is very small.


trwalke (approved):
Im guessing that there is a possibility that some of the apis can be used right? So this would break them?

gladjohn (approved):

PR #5001, opened 2024-11-19
[DO NOT REVIEW] MSI V2 

gladjohn:
[DO NOT REVIEW] MSI V2


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -29,6 +34,17 @@ public ManagedIdentityClient(RequestContext requestContext)
             }
         }
 
+        /// <summary>
+        /// Resets the cached managed identity source. Used only for testing purposes.
+        /// </summary>
+        internal static void ResetManagedIdentitySourceCache()
```
bgavrilMS, line 40: I think this will need to be made public for Azure SDK to test. And it'd be better to have a "ResetStaticCaches" general API, like our tests currently do, which reset everything.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ServiceFabricTests.cs
```diff
@@ -30,6 +30,9 @@ public class ServiceFabricTests
         public void TestInitialize()
         {
             TestCommon.ResetInternalStaticCaches();
```
bgavrilMS, line 32: The TestBase class already calls `TestCommon.ResetInternalStaticCaches();`

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: i.e. see the logic in TestCommon.ResetInternalStaticCaches();

src/client/Microsoft.Identity.Client/ManagedIdentity/IProbe.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal interface IProbe
```
bgavrilMS, line 9: The only reason to introduce an interface is if you're going to mock it, but it doesn't seem to be the case here. You don't need to mock this object, as you can rely on HTTP mocking.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsCredentialProbeManager.cs
```diff
@@ -0,0 +1,111 @@
...
+                _logger.Info($"[Probe] Credential endpoint supported. Server: {serverHeader}");
+                return ProbeResult.Success();
+            }
+
+            _logger.Verbose(() => "[Probe] Credential endpoint not supported.");
+            return ProbeResult.Failure("Credential endpoint not supported.");
+        }
+    }
+
+    internal class ProbeResult
```
bgavrilMS, line 96: Seems to be more like a general purpose class "SuccessAndErrorMessage". I would also not use a class here, but a struct, to avoid heap allocations.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsCredentialProbeManager.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (response.StatusCode == HttpStatusCode.BadRequest &&
+                response.HeadersAsDictionary.TryGetValue("Server", out string serverHeader) &&
+                serverHeader.StartsWith(ImdsHeader, StringComparison.OrdinalIgnoreCase))
+            {
+                _logger.Info($"[Probe] Credential endpoint supported. Server: {serverHeader}");
+                return ProbeResult.Success();
+            }
+
+            _logger.Verbose(() => "[Probe] Credential endpoint not supported.");
+            return ProbeResult.Failure("Credential endpoint not supported.");
```
bgavrilMS, line 92: If you're just using static messages, I don't see the point of "ProbeResult" at all.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsCredentialProbeManager.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            if (response.StatusCode == HttpStatusCode.BadRequest &&
+                response.HeadersAsDictionary.TryGetValue("Server", out string serverHeader) &&
+                serverHeader.StartsWith(ImdsHeader, StringComparison.OrdinalIgnoreCase))
+            {
+                _logger.Info($"[Probe] Credential endpoint supported. Server: {serverHeader}");
+                return ProbeResult.Success();
+            }
+
+            _logger.Verbose(() => "[Probe] Credential endpoint not supported.");
```
bgavrilMS, line 91: You will need a lot more logging here, to find out why the probe failed.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsCredentialProbeManager.cs
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal class ImdsCredentialProbeManager : IProbe
+    {
+        private const string CredentialEndpoint = "http://169.254.169.254/metadata/identity/credential";
+        private const string ProbeBody = ".";
+        private const string ImdsHeader = "IMDS/";
+        private static readonly SemaphoreSlim s_lock = new(1);
+        private static ProbeResult s_cachedResult;
```
bgavrilMS, line 20: We had a lot of problems with static region auto-discovery, which uses a similar concept. 

1. Should the HTTP call fail after some time? (it does in region ... 2 seconds)
2. Are you sure that probe failure really means endpoint not available? Could the probe fail for other reasons? 
3.  should the probe (HTTP call) be retried? How often / for how long?

gladjohn: we use the MSI httpmanager logic with the same retries and error conditions. 

We cannot default to credential endpoint because there are still RPs that use IMDS like AKS etc. So we need to check if credential endpoint exist first and then fallback to IMDS

PR #5023, opened 2024-12-16
Refactor lab API and remove certificate based auth

bgavrilMS:
Fixes #

**Testing**

**Performance impact**

**Documentation**


tests/Microsoft.Identity.Test.LabInfrastructure/LabAuthenticationHelper.cs
```diff
@@ -8,72 +8,103 @@
...
-            confidentialApp = ConfidentialClientApplicationBuilder
-                .Create(clientIdForCertAuth)
-                .WithAuthority(new Uri(authority), true)
-                .WithCacheOptions(CacheOptions.EnableSharedCacheOptions)
-                .WithCertificate(cert, true)
-                .Build();
+        // TODO: test this on MacOs / Linux WSL
+        private static TokenCredential GetAzureCredentialForDevBox()
+        {
+            InteractiveBrowserCredential interactiveBrowserCredential = new InteractiveBrowserCredential(
```
gladjohn, line 89: can we use `VisualStudioCredential` so it works even on a VPN'd dev machine? 

```cs
// Set the Key Vault URL
string keyVaultUrl = "https://msidlabs.vault.azure.net/";

// Create a new Visual Studio Credential
var credential = new VisualStudioCredential();

// Create a new SecretClient using the Visual Studio Credential
var client = new SecretClient(new Uri(keyVaultUrl), credential);

// Retrieve a secret from Azure Key Vault
KeyVaultSecret secret = client.GetSecret("msidlab1");

Console.WriteLine($"Secret: {secret.Value}");
```
this will use the Azure Service Authentication account 

image

tests/Microsoft.Identity.Test.LabInfrastructure/LabApiConstants.cs
```diff
@@ -3,7 +3,15 @@
 
 namespace Microsoft.Identity.Test.LabInfrastructure
 {
-    public class LabApiConstants
+    public static class LabApiConstants
+    {
+        public const string LabClientId = "f62c5ae3-bf3a-4af5-afa8-a68b800396e9";
```
gladjohn, line 8: if we move to use Visual Studio credential for DevBox and UAMI based acccess for CI, we do not need this app flow anymore

tests/Microsoft.Identity.Test.Common/Core/Helpers/CertHelper.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+        /// <summary>
+        /// Try and locate a certificate matching the given <paramref name="certName"/> by searching in
+        /// the in the given <see cref="StoreName"/> and <see cref="StoreLocation"/>.
+        /// </summary>
+        /// <param subjectName="certName">Thumbprint of certificate to locate</param>
+        /// <param subjectName="location"><see cref="StoreLocation"/> in which to search for a matching certificate</param>
+        /// <param subjectName="name"><see cref="StoreName"/> in which to search for a matching certificate</param>
+        /// <returns><see cref="X509Certificate2"/> with <paramref subjectName="certName"/>, or null if no matching certificate was found</returns>
+        public static X509Certificate2 FindCertificateByName(string certName, StoreLocation location, StoreName name)
```
gladjohn, line 40: there seems to be a cert helper already for this

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/CiamIntegrationTests.cs
```diff
@@ -184,7 +171,7 @@ public async Task OBOCiam_CustomDomain_ReturnsValidTokens()
             //Acquire tokens for OBO
             var msalConfidentialClient = ConfidentialClientApplicationBuilder
                 .Create(ciamWebApi)
-                .WithCertificate(CertificateHelper.FindCertificateByName(TestConstants.AutomationTestCertName))
+                .WithCertificate(CertificateFinder.FindCertificateByName(TestConstants.AutomationTestCertName))
```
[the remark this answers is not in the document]
bgavrilMS: It still makes sense to have a single cert secure all our tests apps.

tests/Microsoft.Identity.Test.LabInfrastructure/LabAuthenticationHelper.cs
bgavrilMS: No, we can't use any of the DefaultAzureCredential options, because there is no option to change the clientID. If you use VSCredentail, it'll use VS clientID.

And VS clientID is not authorized to call MSIDlab.

I was thinking that if we were to directly call the KV that might work though. Maybe not VS client ID, but az cli or some tool might do it

gladjohn: `VS clientID` is now authorized to call Lab API for an authorized user

tests/Microsoft.Identity.Test.Common/Core/Helpers/CertHelper.cs
bgavrilMS: There is, but the lab infra doesn't depend on anything.

PR #5027, opened 2024-12-20
Fix for #4860 - ignore tenant id for authorities that do not support …

### LINKED ISSUE #4860: AbstractAcquireTokenParameterBuilder<T>.WithAdfsAuthority throws with a null TenantId when the authority host Uri ends in `/adfs`
Azure.Identity recently changed from calling the deprecated:

```c#
AbstractAcquireTokenParameterBuilder<T>.WithAuthority(AuthorityHost.AbsoluteUri, tenantId);
```
to this: 
```c#
UriBuilder uriBuilder = new UriBuilder(AuthorityHost)
{
    Path = tenantId
};
builder.WithTenantIdFromAuthority(uriBuilder.Uri);
```

When `adfs` is passed as the tenantId, `WithTenantIdFromAuthority` throws because tenantId is null.

I looked at the MSAL code and this is the reason we end up with a null TenantId
 
[This](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4110b05894ef5ab39740ac4733c03f697530cf71/src/client/Microsoft.Identity.Client/ApiConfig/AbstractAcquireTokenParameterBuilder.cs#L311-L313) creates an AdfsAuthority which [hard codes null as the TenantId](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4110b05894ef5ab39740ac4733c03f697530cf71/src/client/Microsoft.Identity.Client/Instance/AdfsAuthority.cs#L63)
 
Ideally `WithTenantIdFromAuthority` should handle the ADFS case and do the same thing as `WithAuthority(AuthorityHost.AbsoluteUri, tenantId)` does.

bgavrilMS:
…tenants

Fixes #4860 

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/UsernamePasswordIntegrationTests.NetFwk.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/UsernamePasswordIntegrationTests.NetFwk.cs
@@ -110,12 +110,14 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
             LabResponse labResponse = await LabUserHelper.GetAdfsUserAsync(FederationProvider.ADFSv2019, true).ConfigureAwait(false);
 
             var user = labResponse.User;
-
+            Uri authorityUri = new Uri(Adfs2019LabConstants.Authority);
+            
             var msalPublicClient = PublicClientApplicationBuilder
                 .Create(Adfs2019LabConstants.PublicClientId)
-                .WithAdfsAuthority(Adfs2019LabConstants.Authority)
+                .WithAuthority(authorityUri)
                 .WithTestLogging()
                 .Build();
+
             AuthenticationResult authResult = await msalPublicClient
                 .AcquireTokenByUsernamePassword(s_scopes, user.Upn, user.GetOrFetchPassword())
                 .ExecuteAsync()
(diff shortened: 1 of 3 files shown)


tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/UsernamePasswordIntegrationTests.NetFwk.cs
```diff
@@ -110,12 +110,14 @@ public async Task AcquireTokenFromAdfsUsernamePasswordAsync()
             LabResponse labResponse = await LabUserHelper.GetAdfsUserAsync(FederationProvider.ADFSv2019, true).ConfigureAwait(false);
 
             var user = labResponse.User;
-
+            Uri authorityUri = new Uri(Adfs2019LabConstants.Authority);
+            
             var msalPublicClient = PublicClientApplicationBuilder
                 .Create(Adfs2019LabConstants.PublicClientId)
-                .WithAdfsAuthority(Adfs2019LabConstants.Authority)
+                .WithAuthority(authorityUri)
```
gladjohn, line 117: do we need to test for `WithTenantIdFromAuthority`?

gladjohn (approved):

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/UsernamePasswordIntegrationTests.NetFwk.cs
bgavrilMS: I added a unit test. I don't think we need integration test, since the protocol is the same. This is just MSAL internal logic.

johnypony3:
Does it make sense that this is included in the CHANGELOG.md?

PR #5029, opened 2024-12-30
Force use specific version of .net SDK

bgavrilMS:
Internal change. 

VS now uses .NET 9 as default. MSAL doesn't compile with .NET due to the mobile targets. This will force-use .NET 6, until we upgrade mobile targets.

Should not affect CI.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "8.0.404",
+        "version": "6.0.428",
         "rollForward": "latestFeature"
     }
 }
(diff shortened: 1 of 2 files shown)


global.json
```diff
@@ -1,5 +1,6 @@
 {
-	"msbuild-sdks": {
```
bgavrilMS, line 2: not used since we dropped UWP

bgavrilMS:
CI is green with the exception of BuildQualityTask . @trwalke - thoughts on this?

bgavrilMS:
Will do as part of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5033

PR #5051, opened 2025-01-06
Add mac broker support and a mac sample app

fengga:

This PR added the macOS broker support via NativeInterop. Need to work with PR https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/4438 to include NativeInterop.proj and get the needed binary files.
Also added a new MAUI sample app calling public APIs to invoke mac broker.

**Testing**

**Performance impact**

**Documentation**


MauiMacAppWithBroker.sln
```diff
@@ -0,0 +1,22 @@
+﻿
```
bgavrilMS, line 1: We already have a Maui dev app that enables the broker on Mac, do we need another one? Maybe delete the old one and keep the new one?

MauiMacAppWithBroker.sln
```diff
@@ -0,0 +1,22 @@
...
+Microsoft Visual Studio Solution File, Format Version 12.00
```
bgavrilMS, line 2: Why do you need a new sln file? We already load a ton of dev apps by default in LibsAndSamples.sln and you can set a solution config to ignore the maui app (happy to show you how to). Is it because our CI build doesn't build your maui app?

Not saying that a new sln file is a bad approach, just trying to figure out what you want to do.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
```diff
@@ -55,6 +55,16 @@ public static bool IsMac()
 #endif
         }
 
+        public static bool IsDarwin()
+        {
+#if MAC
```
bgavrilMS, line 60: This flag isn't in use!

MSAL today can run on MacOS only on `.NET` and `.NETSTANDARD` target frameworks. MSAL doesn't explicitly target `net8-maccatalyst`, so if an app targets net8-maccatalyst it will end up hitting MSAL `net8` version I believe. 

IMO you only need to take care of `net8` support as full `maccatalyst` MSAL support is tracked seperately (lots on interest on that) but is currently blocked due to caching concerns.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
```diff
@@ -55,6 +55,16 @@ public static bool IsMac()
...
+            string OSDescription = RuntimeInformation.OSDescription;
+            return OSDescription.Contains("Darwin", StringComparison.OrdinalIgnoreCase);
+#else
+            return false;
+#endif
+        }
```
bgavrilMS, line 66: Just modify `IsMac` above ? Something like: 

```
   public static bool IsMac()
        {
#if NETFRAMEWORK
            return Environment.OSVersion.Platform == PlatformID.MacOSX;
#if __MOBILE__ 
            return false;
#else
            return RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
#endif
```

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -29,6 +29,10 @@ public enum OperatingSystems
             /// Use broker on Windows OS
             /// </summary>
             Windows = 0b_0000_0001,  // 1
+            /// <summary>
+            /// Use broker on Mac OS
+            /// </summary>
+            Mac = 0b_0000_0010,  // 10
```
bgavrilMS, line 35: nit: consider naming this as `OSX` as this is what the .NET team does in https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.osplatform?view=net-9.0

MauiMacAppWithBroker.sln
fengga: Hi @bgavrilMS , can you point me to the "old" MAUI app with broker on Mac?
I saw in this repo "MauiB2C", "MauiAppBasic", "MauiAppWithBroker" contains "net6.0-maccatalyst", 
but in https://github.com/Azure-Samples/ms-identity-dotnetcore-maui, "maccatalyst" is removed.

MauiMacAppWithBroker.sln
fengga: Per offline discussion, if the new MAUI app added to LibsAndSamples.sln, building on mac will fail due to compatibly issues. Added a new Mac.sln for mac broker sample apps.

src/client/Microsoft.Identity.Client/PublicApi/net472/PublicAPI.Shipped.txt
```diff
@@ -417,6 +417,7 @@ Microsoft.Identity.Client.BrokerOptions.MsaPassthrough.set -> void
 Microsoft.Identity.Client.BrokerOptions.OperatingSystems
 Microsoft.Identity.Client.BrokerOptions.OperatingSystems.None = 0 -> Microsoft.Identity.Client.BrokerOptions.OperatingSystems
 Microsoft.Identity.Client.BrokerOptions.OperatingSystems.Windows = 1 -> Microsoft.Identity.Client.BrokerOptions.OperatingSystems
+Microsoft.Identity.Client.BrokerOptions.OperatingSystems.OSX = 2 -> Microsoft.Identity.Client.BrokerOptions.OperatingSystems
```
bgavrilMS, line 420: These need to go to "Unshipped.txt" and we will move them to "Shipped.txt" when we release.

bgavrilMS (changes requested):
- 1 small issue with IsMac API
- Broker packages need to be updated to reference the new runtime package 

Otherwise LGTM

gautamdsheth:
Very much interested in this :) 

Could you also update the [docs](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/desktop-mobile/wam) page whenever this gets merged ? Also, adding a sample for a console app would be awesome !

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
```diff
@@ -44,12 +44,9 @@ public static bool IsLinux()
 
         public static bool IsMac()
         {
-#if MAC
-            return true;
-#elif NETFRAMEWORK
-            return Environment.OSVersion.Platform == PlatformID.MacOSX;
-#elif !__MOBILE__ 
-            return RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
+#if NET8_0_OR_GREATER
```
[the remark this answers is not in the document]
fengga: @bgavrilMS, I debugged the mac MAUI app, it will not enter `#elif NETFRAMEWORK`. Also, the value of `Environment.OSVersion.Platform` is actually `Unix`.
```
            string OSDescription = RuntimeInformation.OSDescription;
            return OSDescription.Contains("Darwin", StringComparison.OrdinalIgnoreCase);
```
works correctly.

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -80,6 +80,15 @@ private static void AddRuntimeSupport(PublicClientApplicationBuilder builder)
                          return new RuntimeBroker(uiParent, appConfig, logger);
                      };
             }
+            else if (DesktopOsHelper.IsMac())
+            {
+                 builder.Config.BrokerCreatorFunc =
+                     (uiParent, appConfig, logger) =>
+                     {
+                         logger.Info("[Runtime] macOS supports broker.");
```
bgavrilMS, line 88: nit: log message at line 97 needs to be updated.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -100,10 +100,18 @@ public RuntimeBroker(
...
             }
 
-            _parentHandle = GetParentWindow(uiParent);
+            if (DesktopOsHelper.IsWindows())
+            {
+                _parentHandle = GetParentWindow(uiParent);
+            }
+            else
+            {
+                // TODO:ADO 3055958 Parent window handle support on mac
```
bgavrilMS, line 109: nit: Just out of curiosity, how will the broker window be handled on mac? (add a comment)

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
```diff
@@ -44,14 +44,15 @@ public static bool IsLinux()
...
-            return true;
+#if __MOBILE__
+            return false;
 #elif NETFRAMEWORK
             return Environment.OSVersion.Platform == PlatformID.MacOSX;
-#elif !__MOBILE__ 
-            return RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
+#elif NET8_0_OR_GREATER
+            string OSDescription = RuntimeInformation.OSDescription;
+            return OSDescription.Contains("Darwin", StringComparison.OrdinalIgnoreCase);
```
bgavrilMS, line 53: Hmm, so is `RuntimeInformation.IsOSPlatform(OSPlatform.OSX)` not working on net8?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
fengga: I updated in the comment. Previously I did some investigation, but don't have a clear solution yet... I have a card about this topic and will do more research in the next quarter.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
fengga: No... at least `RuntimeInformation.IsOSPlatform(OSPlatform.OSX)` does not work on my mac.

bgavrilMS (approved):

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
```diff
@@ -29,6 +29,10 @@ public enum OperatingSystems
             /// Use broker on Windows OS
             /// </summary>
             Windows = 0b_0000_0001,  // 1
+            /// <summary>
+            /// Use broker on OSX
+            /// </summary>
+            OSX = 0b_0000_0010,  // 10
```
bgavrilMS, line 35: The interop package needs to be updated. The current one only has Windows DLLs

https://nuget.info/packages/Microsoft.Identity.Client.NativeInterop/0.17.2

bgavrilMS (changes requested):
Interop package needs a bump. See the other PR about Linux broker too.

Mac.sln
```diff
@@ -0,0 +1,22 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
gladjohn, line 2: why do we need a new sln, can this not be part of the existing one?

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -80,12 +80,21 @@ private static void AddRuntimeSupport(PublicClientApplicationBuilder builder)
                          return new RuntimeBroker(uiParent, appConfig, logger);
                      };
             }
+            else if (DesktopOsHelper.IsMac())
```
gladjohn, line 83: can you please add a helper method to reduce the dupe here?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -100,10 +100,22 @@ public RuntimeBroker(
                 s_lazyCore.Value.EnablePii(_logger.PiiLoggingEnabled);
             }
 
-            _parentHandle = GetParentWindow(uiParent);
+            if (DesktopOsHelper.IsWindows())
```
gladjohn, line 103: Please update the summary of this method

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -100,10 +100,22 @@ public RuntimeBroker(
...
+                _parentHandle = GetParentWindow(uiParent);
+            }
+            else
+            {
+                // TODO:ADO 3055958 Parent window handle support on mac
+                // Without setting parent window on macOS, the mac broker UI will show up in the middle
+                // of the screen, and keep in the foreground until UI dismissed.
+                // Technically, macOS broker only accept an objc pointer as window handle, currently we
+                // do not know how to get such kind of pointer in MAUI. The solution is still unclear.
+                _parentHandle = (IntPtr)1;
```
gladjohn, line 114: What will happen in Linux?

tests/devapps/MauiMacAppWithBroker/MainPage.xaml.cs
```diff
@@ -0,0 +1,119 @@
+﻿namespace MauiMacAppWithBroker;
+
+using Microsoft.Maui.Controls;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Broker;
+using System;
+using System.Runtime.InteropServices;
+
+public partial class MainPage : ContentPage
```
gladjohn, line 9: can you please add this dev app to the pipeline, build it, and make sure the interop files are placed in the appropriate folders

gladjohn: see here : https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/build/template-build-and-prep-automation.yaml#L63

tests/devapps/MauiMacAppWithBroker/MainPage.xaml.cs
```diff
@@ -0,0 +1,119 @@
...
+	{
+		if (inUse)
+		{
+			await Application.Current.MainPage.DisplayAlert("Warning", "Please wait until previous operation finished", "OK").ConfigureAwait(false);
+			return;
+		}
+		inUse = true;
+
+		PublicClientApplicationBuilder builder = PublicClientApplicationBuilder
+			.Create("04b07795-8ddb-461a-bbee-02f9e1bf7b46")
```
gladjohn, line 73: what app is this? can we not use one from the lab?

tests/devapps/MauiMacAppWithBroker/MainPage.xaml.cs
```diff
@@ -0,0 +1,119 @@
...
+			return;
+		}
+		inUse = true;
+
+		SemanticScreenReader.Announce(CACIACSBtn.Text);
+
+        PublicClientApplicationBuilder builder = PublicClientApplicationBuilder
+			.Create("04b07795-8ddb-461a-bbee-02f9e1bf7b46")
+			.WithRedirectUri("msauth.com.msauth.unsignedapp://auth")
+			.WithClientCapabilities(new[] { "cp1" })
```
gladjohn, line 33: do you need `WithClientCapabilities` the request is not even passing any claims or testing the CAE flow

src/client/Microsoft.Identity.Client.Broker/BrokerExtension.cs
```diff
@@ -80,12 +80,21 @@ private static void AddRuntimeSupport(PublicClientApplicationBuilder builder)
...
+                         return new RuntimeBroker(uiParent, appConfig, logger);
+                     };
+            }
             else
             {
                 builder.Config.BrokerCreatorFunc =
                    (uiParent, appConfig, logger) =>
                    {
-                       logger.Info("[RuntimeBroker] Not a Windows 10 or Server equivalent machine. Runtime broker or SsoPolicy support is not available.");
+                       logger.Info("[RuntimeBroker] not available in current platform.");
```
gladjohn, line 97: can you please add the current os type? 

```suggestion
                       logger.Warning($"[RuntimeBroker] Not available on the current platform: {osType}.");
```

gladjohn (commented):
Looks good. 

- CI pipeline integration of the test app is a must. 
- As Bogdan mention need an updated Interop with MAC runtimes
- Highly recommend building a broker app with MAC libs/runtimes and testing from within Visual Studio for MAC. Chris maybe able to help here with some guidance if needed.

Mac.sln
bgavrilMS: I think this is fine. It's the easiest way to exclude projects taht don't build on Mac.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
bgavrilMS: Separate PR is for Linux.

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
bgavrilMS: Looks like interop 0.18.0 was released and it has Linux x64 support. But no mac support https://nuget.info/packages/Microsoft.Identity.Client.NativeInterop/0.18.0

fengga: I published a NativeInterop-0.19.0-alpha to IDDP, which contains the mac binaries.
https://dev.azure.com/IdentityDivision/IDDP/_artifacts/feed/IDDP/NuGet/Microsoft.Identity.Client.NativeInterop/overview/0.19.0-alpha
Our original plan was first publishing all the packages to IDDP, and partners verify everything works fine, then publish to nuget.org. Does this make sense?

tests/devapps/MauiMacAppWithBroker/MainPage.xaml.cs
fengga: Can we track it with another work item? I created one https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3168507
It's in my epic and will not be forgotten

tests/devapps/MauiMacAppWithBroker/MainPage.xaml.cs
fengga: I created an empty MAUI project; this id is generated automatically.

tests/devapps/MauiMacAppWithBroker/MainPage.xaml.cs
bgavrilMS: Does it matter? It's just a test app. It's not enough to test teh full CAE flow, but it is enough to test that CP1 is going through.

tests/devapps/MauiMacAppWithBroker/MainPage.xaml.cs
fengga: Now I'm using a client id of my own app.

Mac.sln
bgavrilMS, line 2: Hi @fengga - looks like we figured out a new way of doing this. It's called solution filters. See https://learn.microsoft.com/en-us/visualstudio/msbuild/solution-filters?view=vs-2022 

There's already an example https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/LibsAndSamples.sdk.slnf

It would be better to just add a `LibsAndSamples.MacDesktop.slnf` or smth like that?

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
bgavrilMS: You can release MSAL **from this branch** to IDDP and validate. We should not merge this branch to master, because we might release at any time.

Mac.sln
fengga: Thanks, @bgavrilMS ! 
The mac maui sample app can be directed added into LibsAndSamples.sln

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
fengga: https://nuget.info/packages/Microsoft.Identity.Client.NativeInterop/0.18.2 already contains mac binaries, and I updated the sample app .csproj to consume 0.18.2 from nuget.
My understanding is if this PR merged to main, UI based app will support broker flows, nobody would be blocked and the main branch is still ready for release any time.
@iulico-1 , can you comment on this?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -108,10 +108,25 @@ public RuntimeBroker(
...
 
-            _parentHandle = GetParentWindow(uiParent);
+            // We need to set parent window handle for MSALRuntime on Windows.
+            // On Windows, WAM UI correctly parented will prevent issues like WAM UI being hidden by terminal app.
+            // The solution will be different on different platforms.
+            if (DesktopOsHelper.IsWindows())
+            {
+                _parentHandle = GetParentWindow(uiParent);
+            }
+            else
```
xinyuxu1026, line 118: How about Linux platform?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
```diff
@@ -499,7 +514,7 @@ public async Task<IReadOnlyList<IAccount>> GetAccountsAsync(
             ICacheSessionManager cacheSessionManager,
             IInstanceDiscoveryManager instanceDiscoveryManager)
         {
-            if (!_wamOptions.ListOperatingSystemAccounts)
+            if (_brokerOptions.EnabledOn == BrokerOptions.OperatingSystems.Windows && !_brokerOptions.ListOperatingSystemAccounts)
```
xinyuxu1026, line 517: This will also change the Linux platform logic

fengga: When debugging on mac, the code here will fail mac flows. I think this part is windows only logic. Isn't it?

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
fengga: It will be directly related to Linux implementation, I'm not sure how will Linux support parent window handle.
I think we can change the code here when begin to support Linux parent window handle.

bgavrilMS (approved):

gladjohn (approved):

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
xinyuxu1026: Please see the summary part of ListOperatingSystemAccounts. if you disable Linux platform here, it might be breaking changes for customers.

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
xinyuxu1026: I see. Could you also add more comment in the summary part? Thanks

xinyuxu1026 (approved):

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
iulico-1: IMO we should add a changelog entry that this is scoped to UI based apps only and next change will address the console apps. Having a single large PR is not ideal I would say. 

Since this is adding new functionality not changing existing flows that should be ok.

iulico-1 (approved):
:shipit:

src/client/Microsoft.Identity.Client.Broker/RuntimeBroker.cs
fengga: Thanks, you are right! It's a bug. Linux and macOS should have the same behavior.

src/client/Microsoft.Identity.Client/ApiConfig/BrokerOptions.cs
fengga: Thanks Iulian, this PR will not affect existing flows.
But I will not touch changelog in this PR since I saw changelog.md is updated centrally per release.

xinyuxu1026 (approved):

Issue #5043, opened 2025-01-02
[Bug] Pipeline failure - Expired PAT


johnypony3:
### Library version used

N/A

### .NET version

N/A

### Scenario

Other - please specify

### Is this a new or an existing app?

None

### Issue description and reproduction steps

This code is failing [the pipeline](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/actions/runs/12529767055), no deployments are happening.

```yaml
        - name: Azure Pipelines Action
          uses: Azure/pipelines@354dddefceb0b503a61338ca81e4091eae3bc84f # v1
          with:
            azure-devops-project-url: https://identitydivision.visualstudio.com/IDDP
            azure-pipeline-name: 'MSAL.NET-OneBranch-Release-Official'
            azure-devops-token: ${{ secrets.AZURE_DEVOPS_TOKEN }}
```

### Relevant code snippets

```csharp
﻿{
    "$id": "1",
    "customProperties": {
        "Descriptor": null,
        "IdentityDisplayName": null,
        "Token": null,
        "RequestedPermissions": 0,
        "NamespaceId": "00000000-0000-0000-0000-000000000000"
    },
    "innerException": null,
    "message": "Access Denied: The Personal Access Token used has expired.",
    "typeName": "Microsoft.VisualStudio.Services.Security.AccessCheckException, Microsoft.VisualStudio.Services.WebApi",
    "typeKey": "AccessCheckException",
    "errorCode": 0,
    "eventId": 3000
}
```

### Expected behavior

The pipeline does not fail.

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

### Solution and workarounds

Generate new PAT and update value. Add monitoring to deployment.

bgavrilMS:
Not an MSAL issue.

johnypony3:
@bgavrilMS Respectfully you are closing bug tickets but the bug fixes are not being deployed because the pipeline is broken. As the owner of this repo, it would fall squarely on you to make sure this is deployed.

bgavrilMS:
Hi @johnypony3 - this is an SDK. We publish a nuget package to https://www.nuget.org/ . Your application references our package explicitly and you manually point it to a version you are interested in. 

In particular, the bug fix you want hasn't even been released. The GH milestone will show you which version of the SDK will have the fix. 

We are not responsible for any pipelines other than our own (and I can tell you we are green)

[...]

PR #5054, opened 2025-01-06
Delete .github/workflows/trigger_onebranch_ci.yml

gladjohn:
This needs a PAT in GitHub to connect to ADO, but with recent security updates the PATs expire every 7 days. Instead we now run OneBranch CI internally on a time trigger that pulls the main from msal source.


bgavrilMS:
Fixes #5043

PR #4965, opened 2024-10-22
Implementation for SNI + MTLS Flow in MSAL

gladjohn:
Fixes : #4986 

**Changes proposed in this request**
- When `.WithCertificate` API is used and if the new `.WithMtlsProofOfPossession()` API is used at request level, we now use that cert over mTLS
- ESTS on the other hand, will use the mTLS cert as the auth cert in this client credentials flow
- MSAL will simply pass the client id and the MTLS cert and will also specify that the `token type = mtls_pop` and ESTS will issue a token that will contain the cnf claim in it. 
- There is a MSAL Client [-preview NuGet package](https://www.nuget.org/packages/Microsoft.Identity.Client/4.66.2-preview) that let's you test the flow 

```cs
string clientId = "163ffef9-a313-45b4-ab2f-c7e2f5e0e23e";
string authority = "https://login.microsoftonline.com/bea21ebe-8b64-4d06-9f6d-6a889b120a7c";
string[] scopes = new string[] { "https://graph.microsoft.com/.default" };

X509Certificate2 certificate = GetCertificateFromStore("Use The Lab Auth Cert");

IConfidentialClientApplication app = ConfidentialClientApplicationBuilder.Create(clientId)
    .WithAuthority(authority)
    .WithAzureRegion("westus")
    .WithCertificate(certificate, true)
    .WithExperimentalFeatures(true)
    .Build();

AuthenticationResult result = await app.AcquireTokenForClient(scopes).WithMtlsProofOfPossession()
    .WithExtraQueryParameters("dc=ESTSR-PUB-WUS3-AZ1-TEST1&slice=TestSlice") //Feature in test slice 
    .WithSendX5C(true)
    .ExecuteAsync();

Console.WriteLine($"Token: {result.AccessToken}");
```

**Testing**
- Unit Tests
- Integration Tests

**Performance impact**

**Documentation**
- [ ] Todo  : need to update docs for MTLS POP

PR #5055, opened 2025-01-06
Code Quality Base branch must point to OB builds

gladjohn:
Code Quality Base branch must point to OB builds

### DIFF
--- a/build/template-build-and-run-all-tests.yaml
+++ b/build/template-build-and-run-all-tests.yaml
@@ -81,5 +81,5 @@ jobs: #Build and stage projects
       coverageUpperThreshold: '90'
       includePartiallySucceeded: false
       fallbackOnPRTargetBranch: false
-      baseDefinitionId: '905' #this is the PR build and it is used as a baseline
+      baseDefinitionId: '1545' #this is the OB build and it is used as a baseline
       baseBranchRef: 'main'


gladjohn:
this might need a little more work in OB, because OB % is being reported very low now. But ideally we need the baseline to be in main and not pointing to PRs. 

image

gladjohn:
when pointing to PR builds, sometimes it ends up picking another PR build like [here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1410581&view=logs&j=5cf383dd-a6bb-5807-f5fb-3befc0552291&t=8fd48d9a-e3c9-5143-e607-20e15dde04ca&l=17)

image

And that PR build did not even run the required step. 

image


Ideally we need to be running this against main, but with OB and main the coverage has a vast difference. 

@trwalke thoughts?

gladjohn:
another issue want to fix is, 

image

gladjohn:
> another issue want to fix is,
> 
> image

we can maybe set 

```
allowCoverageVariance: true
coverageVariance: 0.00001
```

trwalke:
> when pointing to PR builds, sometimes it ends up picking another PR build like [here](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1410581&view=logs&j=5cf383dd-a6bb-5807-f5fb-3befc0552291&t=8fd48d9a-e3c9-5143-e607-20e15dde04ca&l=17)
> 
> image
> 
> And that PR build did not even run the required step.
> 
> image
> 
> Ideally we need to be running this against main, but with OB and main the coverage has a vast difference.
> 
> @trwalke thoughts?

The code coverage works like this:
1. The release build runs on main to create a baseline code coverage number every night. (The check code coverage task does not run on the PR build that targets main because the policy requires a code coverage improvement. There may not be an improvement on the baseline so this task is likely to fail.)
2. The policy scans the PR build history for the latest build that targets main to use as a baseline for the comparison.
3. The policy evaluates the code coverage from the current build and the baseline build to ensure it has improved.

So it is expected that the baseline PR builds do not run the task. Also, if the oneBranch build is reporting different code coverage numbers, it is most likely because there is a change in the code coverage config, or the test tasks that scan for code coverage.

trwalke:
Yeah, there was a change in the Code coverage config for sure. now the wrong dlls are being scanned. This is why the value is so low.
{5D26ED32-B8F3-4E83-BC41-75B0ED2654DA}

build/template-build-and-run-all-tests.yaml
```diff
@@ -81,5 +81,5 @@ jobs: #Build and stage projects
       coverageUpperThreshold: '90'
       includePartiallySucceeded: false
       fallbackOnPRTargetBranch: false
-      baseDefinitionId: '905' #this is the PR build and it is used as a baseline
+      baseDefinitionId: '1545' #this is the OB build and it is used as a baseline
```
trwalke, line 84: We cant use OneBranch build as a baseline because the value calculated is slightly different than on the PR build. Most likely due to underlying VS components.

bgavrilMS:
> Yeah, there was a change in the Code coverage config for sure. now the wrong dlls are being scanned. This is why the value is so low. {5D26ED32-B8F3-4E83-BC41-75B0ED2654DA}

What is the change? All changes are tracked by git.

trwalke:
> > Yeah, there was a change in the Code coverage config for sure. now the wrong dlls are being scanned. This is why the value is so low. {5D26ED32-B8F3-4E83-BC41-75B0ED2654DA}
> 
> What is the change? All changes are tracked by git.

@bgavrilMS @gladjohn I found the offending Commit but it is the massive MTLS PR
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4965
There are no obvious indications of what the issue is so I will need to dive a bit deeper to find the change that broke the build.

build/template-build-and-run-all-tests.yaml
gladjohn: fixed it

gladjohn:
merge this PR : https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5072 

then the current PR

gladjohn:
> Yeah, there was a change in the Code coverage config for sure. now the wrong dlls are being scanned. This is why the value is so low. {5D26ED32-B8F3-4E83-BC41-75B0ED2654DA}

this is because the tests had code coverage enabled. 

PR for fix : https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5072

**After fix :** 

**PR Build :** 

image


**OB Build :** 

**
image
**

build/template-build-and-run-all-tests.yaml
gladjohn: @trwalke now OB and PR build % are the same

neha-bhargava: How was this fixed? Just curious

neha-bhargava (approved):

PR #5072, opened 2025-01-10
Disable OB Test Code Coverage

gladjohn:
Disable OB Test Code Coverage

https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1412185&view=codecoverage-tab

image

### DIFF
--- a/build/template-OneBranch-Tests-libsandsamples.yaml
+++ b/build/template-OneBranch-Tests-libsandsamples.yaml
@@ -25,7 +25,7 @@ steps:
     testSelector: 'testAssemblies'
     testAssemblyVer2: '**\CacheCompat\CommonCache.Test.Unit\bin\**\CommonCache.Test.Unit.dll'
     searchFolder: '$(System.DefaultWorkingDirectory)'
-    codeCoverageEnabled: true
+    codeCoverageEnabled: false
     failOnMinTestsNotRun: true
     minimumExpectedTests: '1'
     runInParallel: true


gladjohn:
> I do not understand the reasoning.

If you go to OB [Code Coverage](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1412355&view=codecoverage-tab) you will see that test dlls are being reported and the overall % is 54. 

With this fix, we now have a matching number with the PR builds. see here : https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1412185&view=codecoverage-tab 

OB runs on main, ideally the PR code coverage baseline must point to the OB and not the PR itself. As PR numbers keep varying.

bgavrilMS (approved):

bgavrilMS:
> > I do not understand the reasoning.
> 
> If you go to OB [Code Coverage](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1412355&view=codecoverage-tab) you will see that test dlls are being reported and the overall % is 54.
> 
> With this fix, we now have a matching number with the PR builds. see here : https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1412185&view=codecoverage-tab
> 
> OB runs on main, ideally the PR code coverage baseline must point to the OB and not the PR itself. As PR numbers keep varying.

It's very strange that **disabling** code coverage collection increases the CC percentage :)

gladjohn:
> > > I do not understand the reasoning.
> > 
> > 
> > If you go to OB [Code Coverage](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1412355&view=codecoverage-tab) you will see that test dlls are being reported and the overall % is 54.
> > With this fix, we now have a matching number with the PR builds. see here : https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1412185&view=codecoverage-tab
> > OB runs on main, ideally the PR code coverage baseline must point to the OB and not the PR itself. As PR numbers keep varying.
> 
> It's very strange that **disabling** code coverage collection increases the CC percentage :)

For OB we have two stages where code is being built, build and test stages. 

image

ideally coverage should be calculated in only one stage. So disabling test reports the correct code coverage % from the build stage. this now perfectly matches the PR build.

PR #5083, opened 2025-01-14
Update code coverage policy

trwalke:
Enabling the build to not fail if the code coverage value is maintained

Fixes #
Basically, this will allow us to make changes to non covered lines without the build failing. the policy is a lot more lenient while still nudging us in the right direction. The build will now fail if the code coverage falls below the baseline which is set by the nightly build that targets main

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/build/template-build-and-run-all-tests.yaml
+++ b/build/template-build-and-run-all-tests.yaml
@@ -74,10 +74,9 @@ jobs: #Build and stage projects
       checkCoverage: '$(CodeCoverageGateEnabled)'
       coverageFailOption: 'build'
       coverageType: 'lines'
-      allowCoverageVariance: true 
-      coverageVariance: 1 # Specify by how much the current amount of covered or uncovered code (depending on the parameter *Use Uncovered Elements*) may deviate from the previous value before the policy fails. Please be aware that the code coverage may slowly but steadily decrease from build to build if you allow a code coverage variance. Thus, you should keep this value as low as possible.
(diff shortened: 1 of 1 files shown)


build/template-build-and-run-all-tests.yaml
```diff
@@ -74,10 +74,9 @@ jobs: #Build and stage projects
       checkCoverage: '$(CodeCoverageGateEnabled)'
       coverageFailOption: 'build'
       coverageType: 'lines'
-      allowCoverageVariance: true 
-      coverageVariance: 1 # Specify by how much the current amount of covered or uncovered code (depending on the parameter *Use Uncovered Elements*) may deviate from the previous value before the policy fails. Please be aware that the code coverage may slowly but steadily decrease from build to build if you allow a code coverage variance. Thus, you should keep this value as low as possible.
+      allowCoverageVariance: false
```
trwalke, line 77: Disabling for now. Can be enabled based on the behavior of the current policy.

neha-bhargava, line 77: With this now the build won't fail but will give it in warning?

neha-bhargava (approved):

build/template-build-and-run-all-tests.yaml
trwalke: It will still fail, but it will be a lot more forgiving to non product code changes

gladjohn (approved):

PR #5091, opened 2025-01-20
Cache key extensibility spec proposal

bgavrilMS:
Spec proposal

### DIFF
--- a/docs/cache_extensibility.md
+++ b/docs/cache_extensibility.md
@@ -86,7 +86,7 @@ Example:
       "cached_at": "1553819803",
       "expires_on": "1553823402",
       "key1": "val1",  // new!
-      "key2": "val2",  // new!
+      "key2": "val2",  // new! }
     }
 }
 ```
--- a/docs/cache_extensibility.md
+++ b/docs/cache_extensibility.md
@@ -48,7 +48,7 @@ After executing the AcquireToken instruction, MSAL shall associate the token wit
 
 #### Higher level APIs
 
-Do no expose this logic in higher level APIs.
+Do not expose this logic in higher level APIs.
 
 ## Token Schema changes
 
--- a/docs/cache_extensibility.md
+++ b/docs/cache_extensibility.md
@@ -87,7 +87,8 @@ Example:
       "expires_on": "1553823402",
       "key1": "val1",  // new!
       "key2": "val2",  // new!
-    },
+    }
+}
 ```
 
 Note: in case the values contain reserved JSON characters, use standard JSON escape rules to serialize.
--- a/docs/cache_extensibility.md
+++ b/docs/cache_extensibility.md
@@ -1,0 +1,124 @@
+# MSAL Cache Key Extensibility Proposal
+
+This document looks at defining a strategy for extending how tokens are cached internally by MSAL, allowing app developers to cache multiple tokens by a custom key.
+
+## Functional Requirements
+
+- Allow higher level SDKs to extend the default cache key semantics that all MSALs come with.
+- Non-breaking changes
+- Implementations MUST be consistent across confidential client MSALs
+- `AcquireTokenForClient` (client_credentials) MUST support this mechanism. Other confidential client flows (web app, web api, ROPC) SHOULD support it.
+
+## Non-Functional Requirements
+
+- Performance of cache look-up operations does not degrade.
+- Forward compatibility strategy - older MSALs must continue to work on the with the same cache as the new ones, to support upgrade scenarios. If this is not possible, an intermediate version of MSAL must be released that supports this.
+- Logging must be extended so that MSAL developers can understand a cache miss.
+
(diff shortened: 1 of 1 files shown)


docs/cache_extensibility.md
```diff
@@ -0,0 +1,125 @@
...
+
+- Allow higher level SDKs to extend the default cache key semantics that all MSALs come with.
+- Non-breaking changes
+- Implementations MUST be consistent across confidential client MSALs
+- `AcquireTokenForClient` (client_credentials) MUST support this mechanism. Other confidential client flows (web app, web api, ROPC) SHOULD support it.
+
+## Non-Functional Requirements
+
+- Performance of cache look-up operations does not degrade.
+- Forward compatibility strategy - older MSALs must continue to work on the with the same cache as the new ones, to support upgrade scenarios. If this is not possible, an intermediate version of MSAL must be released that supports this.
```
trwalke, line 15: I think I understand what is being asked here, but the wording needs to be fixed so that we are 100% sure.

docs/cache_extensibility.md
```diff
@@ -0,0 +1,125 @@
...
+- Logging must be extended so that MSAL developers can understand a cache miss.
```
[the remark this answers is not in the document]
rayluo: My understanding is that the logging can start with including parameter names only, so that we know what parameters contributed into a cache hit/miss. We will revisit this when/if we encounter a need to log the parameter values (hopefully never).

docs/cache_extensibility.md
```diff
@@ -0,0 +1,125 @@
...
+
+var result = await app.AcquireTokenForClient("https://graph.microsoft.com/.default")
+    .WithAdditionalCacheKeyComponents(new Dictionary<string, string>{ // New API
+        {"cert_thubprint", x509Cert.GetSha2THumbprint() }, 
+        {"spiffee_id", "37" }
+    });
+    .ExecuteAsync();   
+```
+
+After executing the AcquireToken instruction, MSAL shall associate the token with the existing components - authority, scope and client_id. In addition, it will assocaite the token with "cert_thumbprint" and "spifee_id".
```
trwalke, line 47: "cert_thumbprint" and "spifee_id" are defined by the customer right?

docs/cache_extensibility.md
```diff
@@ -0,0 +1,125 @@
...
+        {"spiffee_id", "37" }
+    });
+    .ExecuteAsync();   
+```
+
+After executing the AcquireToken instruction, MSAL shall associate the token with the existing components - authority, scope and client_id. In addition, it will assocaite the token with "cert_thumbprint" and "spifee_id". 
+
+#### Higher level APIs
+
+Do no expose this logic in higher level APIs.
```
trwalke, line 51:
```suggestion
Do not expose this logic in higher level APIs.
```

docs/cache_extensibility.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Cache Key
+
+Similar to the [POP](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/SSO/change_proposals/11232019-accesstoken_with_authscheme.md&_a=preview) enhancement, a new credential type will be used - `atext`.
+
+The access token cache key will also add a suffix composed as follows (all operations are ordinal case sensitive):
+
+1. Take the key-value pair list of components and **order** it alphabetically by the key (e.g. "key1": "val1", "key2": "val2")
+1. Concatenate this list  (e.g. "key1val1key2val2")
+1. Hash this using SHA256 - (e.g. `cc252f65706f969930208e4b2403435a95f7f7d9c964bd190ae2c6e032938235`)
```
trwalke, line 63: makes sense. So when we are only searching for one hash value in the cache

docs/cache_extensibility.md
```diff
@@ -0,0 +1,125 @@
...
+## Non-Functional Requirements
+
+- Performance of cache look-up operations does not degrade.
+- Forward compatibility strategy - older MSALs must continue to work on the with the same cache as the new ones, to support upgrade scenarios. If this is not possible, an intermediate version of MSAL must be released that supports this.
+- Logging must be extended so that MSAL developers can understand a cache miss.
+
+### Example scenarios that will use this extensibility point
+
+- Associate tokens with the client certificate used to obtain them.
+- Associtate tokens with the client secrets used to obtain them.
```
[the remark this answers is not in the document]
bgavrilMS: Probably. We don't plan to do this, so we don't care.

docs/cache_extensibility.md
bgavrilMS: yes.

docs/cache_extensibility.md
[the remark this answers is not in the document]
bgavrilMS: We should document that it is expected that only non-PII values are used here.

PR #5107, opened 2025-01-30
Cache key extensibility for MSAL

trwalke:
Fixes #
Implementation of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5091

Enabling MSAL to associate cache key components with cache items (Access tokens only) for later retrival.
Cache components are used when filtering the cache for tokens

**Testing**
Unit tests

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Threading;
 using System.Threading.Tasks;
@@ -92,7 +93,7 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// Modifies the request to acquire a Signed HTTP Request (SHR) Proof-of-Possession (PoP) token, rather than a Bearer.
         /// SHR PoP tokens are bound to the HTTP request and to a cryptographic key, which MSAL manages on Windows.
-       /// SHR PoP tokens are different from mTLS PoP tokens, which are used for Mutual TLS (mTLS) authentication. See <see href="https://aka.ms/mtls-pop"/> for details.
+        /// SHR PoP tokens are different from mTLS PoP tokens, which are used for Mutual TLS (mTLS) authentication. See <see href="https://aka.ms/mtls-pop"/> for details.
         /// </summary>
         /// <param name="popAuthenticationConfiguration">Configuration properties used to construct a Proof-of-Possession request.</param>
         /// <returns>The builder.</returns>
@@ -110,6 +111,26 @@ namespace Microsoft.Identity.Client
             CommonParameters.PopAuthenticationConfiguration = popAuthenticationConfiguration ?? throw new ArgumentNullException(nameof(popAuthenticationConfiguration));
 
             CommonParameters.AuthenticationOperation = new PopAuthenticationOperation(CommonParameters.PopAuthenticationConfiguration, ServiceBundle);
+
+            return this as T;
+        }
+
+        /// <summary>
+        /// Specifies additional cache key components to use when caching and retrieving tokens.
+        /// </summary>
+        /// <param name="cacheKeyComponents">The list of additional cache key components.</param>
+        /// <returns>The builder.</returns>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item><description>This api can be used to associate certificate key identifiers along with other keys with a particular token.</description></item>
+        /// <item><description>In order for the tokens to be succsesfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
+        /// </list>
+        /// </remarks>
+        public T WithAdditionalCacheKeyComponents(IDictionary<string, string> cacheKeyComponents)
+        {
+            ValidateUseOfExperimentalFeature();
+
+            CommonParameters.CacheKeyComponents = cacheKeyComponents ?? throw new ArgumentNullException(nameof(cacheKeyComponents));
 
             return this as T;
         }
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
@@ -2,9 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Generic;
 using System.ComponentModel;
-using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Executors;
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Executors;
@@ -111,26 +112,6 @@ namespace Microsoft.Identity.Client
             CommonParameters.PopAuthenticationConfiguration = popAuthenticationConfiguration ?? throw new ArgumentNullException(nameof(popAuthenticationConfiguration));
 
             CommonParameters.AuthenticationOperation = new PopAuthenticationOperation(CommonParameters.PopAuthenticationConfiguration, ServiceBundle);
-
-            return this as T;
-        }
-
-        /// <summary>
-        /// Specifies additional cache key components to use when caching and retrieving tokens.
-        /// </summary>
-        /// <param name="cacheKeyComponents">The list of additional cache key components.</param>
-        /// <returns>The builder.</returns>
-        /// <remarks>
-        /// <list type="bullet">
-        /// <item><description>This api can be used to associate certificate key identifiers along with other keys with a particular token.</description></item>
-        /// <item><description>In order for the tokens to be succsesfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
-        /// </list>
-        /// </remarks>
-        public T WithAdditionalCacheKeyComponents(IDictionary<string, string> cacheKeyComponents)
-        {
-            ValidateUseOfExperimentalFeature();
-
-            CommonParameters.CacheKeyComponents = cacheKeyComponents ?? throw new ArgumentNullException(nameof(cacheKeyComponents));
 
             return this as T;
         }
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
@@ -1,10 +1,9 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
-using System.Diagnostics.CodeAnalysis;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
@@ -98,6 +99,32 @@ namespace Microsoft.Identity.Client
                 CommonParameters.AuthenticationOperation = new MtlsPopAuthenticationOperation(certificateCredential.Certificate);
                 CommonParameters.MtlsCertificate = certificateCredential.Certificate;
             }
+
+            return this;
+        }
+
+        /// <summary>
+        /// Specifies additional cache key components to use when caching and retrieving tokens.
+        /// </summary>
+        /// <param name="cacheKeyComponents">The list of additional cache key components.</param>
+        /// <returns>The builder.</returns>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item><description>This api can be used to associate certificate key identifiers along with other keys with a particular token.</description></item>
+        /// <item><description>In order for the tokens to be succsesfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
+        /// </list>
+        /// </remarks>
+        [SuppressMessage("Microsoft.ApiDesignGuidelines", "RS0016:Symbol 'WithAdditionalCacheKeyComponents' is not part of the declared public API")]
+        public AcquireTokenForClientParameterBuilder WithAdditionalCacheKeyComponents(IDictionary<string, string> cacheKeyComponents)
+        {
+            ValidateUseOfExperimentalFeature();
+
+            if (cacheKeyComponents == null || cacheKeyComponents.Count == 0)
+            {
+                throw new ArgumentNullException(nameof(cacheKeyComponents));
+            }
+
+            CommonParameters.CacheKeyComponents = cacheKeyComponents;
 
             return this;
         }
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
@@ -104,32 +104,6 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
-        /// Specifies additional cache key components to use when caching and retrieving tokens.
-        /// </summary>
-        /// <param name="cacheKeyComponents">The list of additional cache key components.</param>
-        /// <returns>The builder.</returns>
-        /// <remarks>
-        /// <list type="bullet">
-        /// <item><description>This api can be used to associate certificate key identifiers along with other keys with a particular token.</description></item>
-        /// <item><description>In order for the tokens to be succsesfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
-        /// </list>
-        /// </remarks>
-        [SuppressMessage("Microsoft.ApiDesignGuidelines", "RS0016:Symbol 'WithAdditionalCacheKeyComponents' is not part of the declared public API")]
-        public AcquireTokenForClientParameterBuilder WithAdditionalCacheKeyComponents(IDictionary<string, string> cacheKeyComponents)
-        {
-            ValidateUseOfExperimentalFeature();
-
-            if (cacheKeyComponents == null || cacheKeyComponents.Count == 0)
-            {
-                throw new ArgumentNullException(nameof(cacheKeyComponents));
-            }
-
-            CommonParameters.CacheKeyComponents = cacheKeyComponents;
-
-            return this;
-        }
-
-        /// <summary>
         /// Please use WithAzureRegion on the ConfidentialClientApplicationBuilder object
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
@@ -30,6 +30,6 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
         public Func<OnBeforeTokenRequestData, Task> OnBeforeTokenRequestHandler { get; internal set; }
         public X509Certificate2 MtlsCertificate { get; internal set; }
         public List<string> AdditionalCacheParameters { get; set; }
-        public SortedDictionary<string, string> CacheKeyComponents { get; internal set; }
+        public SortedList<string, string> CacheKeyComponents { get; internal set; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
@@ -30,6 +30,6 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
         public Func<OnBeforeTokenRequestData, Task> OnBeforeTokenRequestHandler { get; internal set; }
         public X509Certificate2 MtlsCertificate { get; internal set; }
         public List<string> AdditionalCacheParameters { get; set; }
-        public IDictionary<string, string> CacheKeyComponents { get; internal set; }
+        public SortedDictionary<string, string> CacheKeyComponents { get; internal set; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
@@ -30,5 +30,6 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
         public Func<OnBeforeTokenRequestData, Task> OnBeforeTokenRequestHandler { get; internal set; }
         public X509Certificate2 MtlsCertificate { get; internal set; }
         public List<string> AdditionalCacheParameters { get; set; }
+        public IDictionary<string, string> CacheKeyComponents { get; internal set; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Identity.Client
         public LogCallback LoggingCallback { get; internal set; }
         public IIdentityLogger IdentityLogger { get; internal set; }
         public string Component { get; internal set; }
-        public IDictionary<string, string> ExtraQueryParameters { get; internal set; } = new Dictionary<string, string>();
+        public IDictionary<string, string> ExtraQueryParameters { get; internal set; }
         public bool UseRecommendedDefaultRedirectUri { get; internal set; }
 
         public bool ExperimentalFeaturesEnabled { get; set; } = false;
--- a/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
@@ -161,7 +161,7 @@ namespace Microsoft.Identity.Client
             }
         }
 
-        public IDictionary<string, string> CacheKeyComponents { get; internal set; }
+        public SortedList<string, string> CacheKeyComponents { get; internal set; }
 #endregion
 
 #region Region
--- a/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
@@ -160,6 +160,8 @@ namespace Microsoft.Identity.Client
                 return null;
             }
         }
+
+        public IDictionary<string, string> CacheKeyComponents { get; internal set; }
 #endregion
 
 #region Region
--- a/src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
-using System.Globalization;
+using System.Linq;
 using Microsoft.Identity.Client.Cache.Items;
 using Microsoft.Identity.Client.Internal.Requests;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
@@ -52,9 +52,7 @@ namespace Microsoft.Identity.Client.Cache
 
             if (requestParameters.AppConfig.IsConfidentialClient)
             {
-                return requestParameters.CacheKeyComponents != null
-                    ? homeAccountIdFromResponse + CoreHelpers.ComputeKeyFromComponents((Dictionary<string, string>)requestParameters.CacheKeyComponents)
-                    : homeAccountIdFromResponse;
+                return homeAccountIdFromResponse;
             }
             if (requestParameters.ApiId == ApiEvent.ApiIds.AcquireTokenSilent)
             {
@@ -85,8 +83,7 @@ namespace Microsoft.Identity.Client.Cache
                 requestParameters.ApiId == ApiEvent.ApiIds.AcquireTokenForUserAssignedManagedIdentity)
             {
                 string tenantId = requestParameters.Authority.TenantId ?? "";
-                key = GetClientCredentialKey(requestParameters.AppConfig.ClientId, tenantId, requestParameters.AuthenticationScheme?.KeyId);
-
+                key = GetClientCredentialKey(requestParameters.AppConfig.ClientId, tenantId, requestParameters.AuthenticationScheme?.KeyId, requestParameters.CacheKeyComponents);
                 return true;
             }
 
@@ -94,8 +91,17 @@ namespace Microsoft.Identity.Client.Cache
             return false;
         }
 
-        public static string GetClientCredentialKey(string clientId, string tenantId, string popKid)
+        public static string GetClientCredentialKey(
+            string clientId,
+            string tenantId,
+            string popKid,
+            SortedDictionary<string, string> CacheKeyComponents = null)
         {
+            if (CacheKeyComponents != null && CacheKeyComponents.Any())
+            {
+                return $"{popKid}{clientId}_{tenantId}_AppTokenCache_{CoreHelpers.ComputeAccessTokenExtCacheKey(CacheKeyComponents)}";
+            }
+
             return $"{popKid}{clientId}_{tenantId}_AppTokenCache";
         }
 
--- a/src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Identity.Client.Cache
                 requestParameters.ApiId == ApiEvent.ApiIds.AcquireTokenForUserAssignedManagedIdentity)
             {
                 string tenantId = requestParameters.Authority.TenantId ?? "";
-                key = GetClientCredentialKey(requestParameters.AppConfig.ClientId, tenantId, requestParameters.AuthenticationScheme?.KeyId, requestParameters.CacheKeyComponents);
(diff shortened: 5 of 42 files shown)


trwalke:
Just need to update tests a little bit

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -324,6 +367,24 @@ internal override JObject ToJObject()
             // previous versions of MSAL used "ext_expires_on" instead of the correct "extended_expires_on".
             // this is here for back compatibility
             SetItemIfValueNotNull(json, StorageJsonKeys.ExtendedExpiresOn_MsalCompat, extExpiresUnixTimestamp);
+#if SUPPORTS_SYSTEM_TEXT_JSON
```
trwalke, line 370: So I updated the implementation on this because I encountered an issue where, based on the original Access token payload design in the spec, where all of the key value pairs are added to the payload, It can be challenging and or prone to failure when reading the additional keys from the user because there is nothing stopping them from adding keys with the same name as our already well defined elements. Such as "client_id" for example. This is especially problematic when reading from a serialized cache because we don't have a good way of knowing which keys to look for. So, I am proposing to simply add an array of elements to the layout of the access token so we can add a well known name to it.

```json
{
  "AccessToken": {
    "-login.microsoftonline.com-atext-client_id-tenant_id-https://graph.microsoft.com/.deafult-cc252f65706f969930208e4b2403435a95f7f7d9c964bd190ae2c6e032938235": {
      "home_account_id": "6afc833f-49c0-4fd5-b685-2998a6cc8d8d.469fdeb4-d4fd-4fde-991e-308a78e4bea4",
      "environment": "login.microsoftonline.de",     
      "client_id": "0615b6ca-88d4-4884-8729-b178178f7c27",
      "secret": "omitted",
      "credential_type": "AccessToken_Extended", // new!
      "realm": "469fdeb4-d4fd-4fde-991e-308a78e4bea4",
      "target": "https://graph.cloudapi.de/62e90394-69f5-4237-9190-012177145e10 https://graph.cloudapi.de/.default",      
      "cached_at": "1553819803",
      "expires_on": "1553823402",
      "additional_cache_key_components": { //new!
                 "key1": "val1",  // new!
                 "key2": "val2",  // new! }
       }
    }
}
```

I also updated the atest to "AccessToken_Extended" as it is more inline with the other token types we have. 

What are your thoughts @rayluo?

src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
```diff
@@ -2,7 +2,9 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
```
bgavrilMS, line 7: No changes in this file?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -102,6 +103,32 @@ public AcquireTokenForClientParameterBuilder WithMtlsProofOfPossession()
...
+        /// <param name="cacheKeyComponents">The list of additional cache key components.</param>
+        /// <returns>The builder.</returns>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item><description>This api can be used to associate certificate key identifiers along with other keys with a particular token.</description></item>
+        /// <item><description>In order for the tokens to be succsesfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
+        /// </list>
+        /// </remarks>
+        [SuppressMessage("Microsoft.ApiDesignGuidelines", "RS0016:Symbol 'WithAdditionalCacheKeyComponents' is not part of the declared public API")]
+        public AcquireTokenForClientParameterBuilder WithAdditionalCacheKeyComponents(IDictionary<string, string> cacheKeyComponents)
```
bgavrilMS, line 118: There is an extensibility namespace and directory that this should go into.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -102,6 +103,32 @@ public AcquireTokenForClientParameterBuilder WithMtlsProofOfPossession()
...
+        {
+            ValidateUseOfExperimentalFeature();
+
+            if (cacheKeyComponents == null || cacheKeyComponents.Count == 0)
+            {
+                throw new ArgumentNullException(nameof(cacheKeyComponents));
```
bgavrilMS, line 124: It's safer to just no-op

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -31,7 +36,8 @@ internal MsalAccessTokenCacheItem(
             string homeAccountId,
             string keyId = null,
             string oboCacheKey = null,
-            IEnumerable<string> persistedCacheParameters = null)
+            IEnumerable<string> persistedCacheParameters = null,
```
bgavrilMS, line 39: Do the cacheKeyComponents behave like the persistedCacheParameters? Should they?

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -48,6 +54,7 @@ internal MsalAccessTokenCacheItem(
             RawClientInfo = response.ClientInfo;
             HomeAccountId = homeAccountId;
             OboCacheKey = oboCacheKey;
+            AdditionalCacheKeyComponents = cacheKeyComponents?.OrderBy(kvp => kvp.Key).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
 #if !MOBILE
```
bgavrilMS, line 58: nit: we shouldn't do things like this. Just the public APIs are hidden.

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -48,6 +54,7 @@ internal MsalAccessTokenCacheItem(
             RawClientInfo = response.ClientInfo;
             HomeAccountId = homeAccountId;
             OboCacheKey = oboCacheKey;
+            AdditionalCacheKeyComponents = cacheKeyComponents?.OrderBy(kvp => kvp.Key).ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
bgavrilMS, line 57: [Blocking] This returns a dictionary. Dictionaries are not sorted by keys.

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
```diff
@@ -179,5 +180,22 @@ internal static string GetCcsUpnHint(string upn)
         {
             return string.IsNullOrEmpty(upn)? string.Empty : $@"upn:{upn}";
         }
+
+        internal static string ComputeKeyFromComponents(Dictionary<string, string> cacheKeyComponents)
```
bgavrilMS, line 184: Why does this need to be a "core helper". It seems very limited in use.

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -48,12 +50,17 @@ public static string GetExternalCacheKeyFromResponse(
                 return key;
             }
 
-            if (requestParameters.AppConfig.IsConfidentialClient ||
-                requestParameters.ApiId == ApiEvent.ApiIds.AcquireTokenSilent)
+            if (requestParameters.AppConfig.IsConfidentialClient)
```
bgavrilMS, line 53: This is incorrect. For `AcquireTokenForClient` it will return at line 50. 

It also shows that there is insufficient testing. The `TokenNotificationsArgs` have a `SuggestedCacheKey` property that needs to checked.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
bgavrilMS, line 118: How many times will we have to order this collection? Would it be faster to accept a SortedList or a SortedDictionary instead? 

https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.sorteddictionary-2?view=net-9.0&redirectedfrom=MSDN#remarks

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -48,12 +50,17 @@ public static string GetExternalCacheKeyFromResponse(
                 return key;
             }
 
-            if (requestParameters.AppConfig.IsConfidentialClient ||
```
bgavrilMS, line 51: I recommend that you update the "internal" partition key as well. No point in having internal and external keys be different.

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
bgavrilMS, line 184: ComputeAccessTokenExtCacheKey

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
```diff
@@ -36,5 +36,23 @@ public static IReadOnlyDictionary<TKey, TValue> GetEmptyDictionary<TKey, TValue>
             return new Dictionary<TKey, TValue>();
 #endif
         }
+
+        public static bool AreDictionariesEqual<TKey, TValue>(IDictionary<TKey, TValue> dict1, IDictionary<TKey, TValue> dict2)
```
bgavrilMS, line 40: Is the generic version slow / faster than the string, string version ?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/CacheKeyExtensionTests.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            //Ensure that the order of the keys does not matter
+            result = await app.AcquireTokenForClient(TestConstants.s_scope.ToArray())
+                                .WithAdditionalCacheKeyComponents(_additionalCacheKeys3)
+                                .ExecuteAsync(CancellationToken.None)
+                                .ConfigureAwait(false);
+
+            Assert.IsNotNull(result);
+            Assert.AreEqual("header.payload.signature", result.AccessToken);
+            Assert.AreEqual(1, app.AppTokenCacheInternal.Accessor.GetAllAccessTokens().Count);
```
bgavrilMS, line 98: You need to assert the exact cache key and the payload entries in the cache.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/CacheKeyExtensionTests.cs
```diff
@@ -0,0 +1,207 @@
...
+
+                var exception = await Assert.ThrowsExceptionAsync<MsalClientException>(async () =>
+                {
+                    await app.AcquireTokenForClient(TestConstants.s_scope.ToArray())
+                                            .WithAdditionalCacheKeyComponents(null)
+                                            .ExecuteAsync(CancellationToken.None)
+                                            .ConfigureAwait(false);
+                }).ConfigureAwait(false);
+
+                Assert.AreEqual(MsalErrorMessage.CryptographicError, exception.Message);
```
bgavrilMS, line 136: That's very poor devex.

bgavrilMS (changes requested):
Cache keys are not correct.

src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
trwalke: There was a missing commit, this is removed

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
trwalke: Moved to proper namespace

trwalke: I realize there are quite a few casts. I changed it so that we still accept an `IDictionary` but the set is ordered right in the beginning and is consistent throughout the code.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
trwalke: Done

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
trwalke: resolved

tests/Microsoft.Identity.Test.Unit/PublicApiTests/CacheKeyExtensionTests.cs
trwalke: I think this is actually incorrect to begin with, this was supposed to be an `ArgumentNull` exception instead of a CryptographicError. Probably pressed ctrl z too many times.

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
trwalke: Under the hood for the generic method, when comparing stings, this method would use `String.Equals`, which would be the same method I would use if writing this for a string comparison. So I would imagine the performance would be very similar but slightly more performant for a string specific method. I figured this might be useful for other scenarios so I used the generic version. The other methods in the class are also generic

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
trwalke: Yes, both are stored in the payload and should be treated the same.

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
trwalke: Updated

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
trwalke: Yeah, I needed to validate the external key. Updated

tests/Microsoft.Identity.Test.Unit/PublicApiTests/CacheKeyExtensionTests.cs
trwalke: Done

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
```diff
@@ -495,6 +497,23 @@ async Task<MsalAccessTokenCacheItem> ITokenCacheInternal.FindAccessTokenAsync(
             return msalAccessTokenCacheItem;
         }
 
+        private void FilterTokensByAdditionalKeyComponents(List<MsalAccessTokenCacheItem> accessTokens, AuthenticationRequestParameters requestParams)
+        {
+            if (requestParams.CacheKeyComponents != null)
+            {
+                accessTokens.FilterWithLogging(item =>
+                    item.AdditionalCacheKeyComponents != null &&
+                    CollectionHelpers.AreDictionariesEqual(item.AdditionalCacheKeyComponents, requestParams.CacheKeyComponents),
```
gladjohn, line 506: did we consider benchmarking this?

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -11,6 +13,34 @@ namespace Microsoft.Identity.Client.Extensibility
...
+        /// <summary>
+        /// Specifies additional cache key components to use when caching and retrieving tokens.
+        /// </summary>
+        /// <param name="cacheKeyComponents">The list of additional cache key components.</param>
+        /// <param name="builder"></param>
+        /// <returns>The builder.</returns>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item><description>This api can be used to associate certificate key identifiers along with other keys with a particular token.</description></item>
+        /// <item><description>In order for the tokens to be succsesfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
```
gladjohn, line 25:
```suggestion
        /// <item><description>In order for the tokens to be successfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
```

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -78,17 +83,25 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
...
             key = null;
             return false;
         }
 
-        public static string GetClientCredentialKey(string clientId, string tenantId, string popKid)
+        public static string GetClientCredentialKey(
+            string clientId,
+            string tenantId,
+            string popKid,
+            SortedDictionary<string, string> CacheKeyComponents = null)
```
gladjohn, line 98:
```suggestion
            SortedDictionary<string, string> cacheKeyComponents = null)
```

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -78,17 +83,25 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
...
         {
+            if (CacheKeyComponents != null && CacheKeyComponents.Any())
+            {
+                return $"{popKid}{clientId}_{tenantId}_AppTokenCache_{CoreHelpers.ComputeAccessTokenExtCacheKey(CacheKeyComponents)}";
```
gladjohn, line 102:
```suggestion
                return $"{popKid}{clientId}_{tenantId}_AppTokenCache_{CoreHelpers.ComputeAccessTokenExtCacheKey(cacheKeyComponents)}";
```

tests/Microsoft.Identity.Test.Unit/PublicApiTests/CacheKeyExtensionTests.cs
```diff
@@ -0,0 +1,218 @@
...
+                                .ExecuteAsync(CancellationToken.None)
+                                .ConfigureAwait(false);
+
+            Assert.IsNotNull(result);
+            Assert.AreEqual("header.payload.signature", result.AccessToken);
+            Assert.AreEqual(1, app.AppTokenCacheInternal.Accessor.GetAllAccessTokens().Count);
+            Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+            ValidateCacheKeyComponents(app.AppTokenCacheInternal.Accessor.GetAllAccessTokens().First(), _additionalCacheKeys3, expectedCacheKey1);
+
+            //Ensure that tokens are not ovverriden
```
gladjohn, line 118:
```suggestion
            //Ensure that tokens are not overridden
```

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
```diff
@@ -36,5 +36,23 @@ public static IReadOnlyDictionary<TKey, TValue> GetEmptyDictionary<TKey, TValue>
...
+
+            foreach (var kvp in dict1)
+            {
+                if (!dict2.TryGetValue(kvp.Key, out var value) || !EqualityComparer<TValue>.Default.Equals(kvp.Value, value))
+                {
+                    return false;
+                }
+            }
+
+            return true;
```
gladjohn, line 55: can either be null?

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -78,17 +83,25 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
...
+                key = GetClientCredentialKey(requestParameters.AppConfig.ClientId, tenantId, requestParameters.AuthenticationScheme?.KeyId, requestParameters.CacheKeyComponents);
                 return true;
             }
 
             key = null;
             return false;
         }
 
-        public static string GetClientCredentialKey(string clientId, string tenantId, string popKid)
+        public static string GetClientCredentialKey(
```
gladjohn, line 94: can we rename this? In MSI V2, credential means something else

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -48,6 +54,12 @@ internal MsalAccessTokenCacheItem(
             RawClientInfo = response.ClientInfo;
             HomeAccountId = homeAccountId;
             OboCacheKey = oboCacheKey;
+
+            if (cacheKeyComponents != null && cacheKeyComponents.Any())
```
gladjohn, line 58: can this go into a helper? `InitializeAdditionalCacheKeyComponents`

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
```diff
@@ -179,5 +180,22 @@ internal static string GetCcsUpnHint(string upn)
         {
             return string.IsNullOrEmpty(upn)? string.Empty : $@"upn:{upn}";
         }
+
+        internal static string ComputeAccessTokenExtCacheKey(SortedDictionary<string, string> cacheKeyComponents)
```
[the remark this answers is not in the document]
trwalke: Returning an empty string now

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
trwalke: Yes, adding case for that

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
trwalke: Yeah

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
trwalke: It looks like the string comparison version is faster by a small margin.

This is 1000 comparisons measured in ms
{8DB7E5F2-2345-4BF3-BBEB-B10C4E22FF4E}
{7EC91036-B35D-4DC6-8E73-8C5094EA7AFC}
{32B38F2F-81BE-4106-B980-500099E45FE5}

src/client/Microsoft.Identity.Client/TokenCache.ITokenCacheInternal.cs
trwalke: @gladjohn please see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5107#discussion_r1942270849

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
gladjohn: good idea

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
gladjohn: Thanks @trwalke

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
rayluo: First and foremost, I consider it acceptable to adjust the design one way or another, on the grounds that SDKs with same cache algorithm can hit the same token in cache, and SDKs with different algorithm (meaning, either with an older version of cache algorithm or with a future newer version of cache algorithm) will merely populate a token with their own key and stick with it. In other words, interoperability across MSALs with different cache algorithm is less of a concern; MSALs with different cache algorithm simply work alongside with each other.

Because of that, the following comments are food for thought, rather than hard requirements.

```json
      ...,
      "cached_at": "1553819803",
      "expires_on": "1553823402",
      "additional_cache_key_components": { //new!
                 "key1": "val1",  // new!
                 "key2": "val2",  // new! 
       }
```

Back then when the Unified Cache Schema was drafted, there was an effort trying to get the schema use integer for those `expires_on` timestamp, and/or use json's native `null` instead of `""`. That effort was rejected, because the schema insisted to keep all values as strings for "consistency". So, introducing a compound key-value pair of "additional_cache_key_components": {...} now will feel like a further deviation from the schema.

Let's revisit your reasoning,

> ... where all of the key value pairs are added to the payload, It can be challenging and or prone to failure when reading the additional keys from the user because there is nothing stopping them from adding keys with the same name as our already well defined elements. Such as "client_id" for example.

Perhaps the current API design gives you a perception that it is a first-class citizen so than app developer can use this api to add arbitrary keys, potentially including "client_id".

```csharp
var result = await app.AcquireTokenForClient("https://graph.microsoft.com/.default")
    .WithAdditionalCacheKeyComponents(new Dictionary<string, string>{ // New API
        {"cert_thubprint", x509Cert.GetSha2THumbprint() }, 
        {"spiffee_id", "37" }
    })
    .ExecuteAsync();
```

In reality, this API shall ideally be internal and is expected to be used by ourselves or MSAL's extension's developers, who will develop an extension to support for example `.WithFmiPath("foo")`, and that extension will internally call `.WithAdditionalCacheKeyComponents(new Dictionary<string, string>{"fmi_path", "foo"})`. There shall never be an encounter of attempting to save client_id again as additional component.

> ... This is especially problematic when reading from a serialized cache because we don't have a good way of knowing which keys to look for.

Out of curiosity, will we still run into trouble with flatten cache entry structure, if the additional keys do not collide with basic fields such as client_id?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -160,6 +160,8 @@ public X509Certificate2 ClientCredentialCertificate
                 return null;
             }
         }
+
+        public IDictionary<string, string> CacheKeyComponents { get; internal set; }
```
bgavrilMS, line 164: There is no advantage of using a interface here. It will lead to upcasting later. Use SortedDictionary.

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -78,17 +83,25 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
...
-        public static string GetClientCredentialKey(string clientId, string tenantId, string popKid)
+        public static string GetTokenCacheItemKey(
+            string clientId,
+            string tenantId,
+            string popKid,
+            SortedDictionary<string, string> cacheKeyComponents = null)
         {
+            if (cacheKeyComponents != null && cacheKeyComponents.Any())
+            {
+                return $"{popKid}{clientId}_{tenantId}_AppTokenCache_{CoreHelpers.ComputeAccessTokenExtCacheKey(cacheKeyComponents)}";
```
bgavrilMS, line 102: Please add this after the {tenantId} part, to keep the "_AppTokenCache" suffix as is.

src/client/Microsoft.Identity.Client/Cache/CacheKeyFactory.cs
```diff
@@ -78,17 +83,25 @@ private static bool GetOboOrAppKey(AuthenticationRequestParameters requestParame
...
+                key = GetTokenCacheItemKey(requestParameters.AppConfig.ClientId, tenantId, requestParameters.AuthenticationScheme?.KeyId, requestParameters.CacheKeyComponents);
                 return true;
             }
 
             key = null;
             return false;
         }
 
-        public static string GetClientCredentialKey(string clientId, string tenantId, string popKid)
+        public static string GetTokenCacheItemKey(
```
bgavrilMS, line 94: The naming of this method is misleading. The previous name was better - as this only applies to app tokens.

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
bgavrilMS: Thanks @rayluo for careful review. 

I like that `additional_cache_key_components` is an extra node because it saves us from having to worry about custom keys conflicting with well-known keys. For example if app developer uses "client_id"="foo" via this new API, it can lead to some rather unexpected behavior. This is totally avoided by having a separate node.

This API is only enabled for client_credentials and in general extensibility points we only enable on CCA. This will limit possible cache sharing issues with PCA. 

.NET doesn't really have "friendly assembly" semantics. So we can't expose this method only to higher level SDKs. We use namespacing to keep the new APIs "hidden".

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
```diff
@@ -30,5 +30,6 @@ internal class AcquireTokenCommonParameters
         public Func<OnBeforeTokenRequestData, Task> OnBeforeTokenRequestHandler { get; internal set; }
         public X509Certificate2 MtlsCertificate { get; internal set; }
         public List<string> AdditionalCacheParameters { get; set; }
+        public SortedDictionary<string, string> CacheKeyComponents { get; internal set; }
```
bgavrilMS, line 33: Why did you choose SortedDictionary over SortedList ? The docs say that SortedList is faster 

image

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
bgavrilMS: 1. But now you use SortedDictionary (and I suggest seeing SortedList too)
2. Why quick and dirty. Here's what Copilot showed me when I asked `How to compare 2 SortedDictionaries with strings for equality?` and `Are there any benchmarks` ?

using System;
using System.Linq;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

public class SortedDictionaryComparisonBenchmark
{
    private SortedDictionary<string, string> dict1;
    private SortedDictionary<string, string> dict2;

    [GlobalSetup]
    public void Setup()
    {
        dict1 = new SortedDictionary<string, string>
        {
            { "one", "1" },
            { "two", "2" },
            { "three", "3" }
        };

        dict2 = new SortedDictionary<string, string>
        {
            { "one", "1" },
            { "two", "2" },
            { "three", "3" }
        };
    }

    [Benchmark]
    public bool SequenceEqualComparison()
    {
        return dict1.SequenceEqual(dict2);
    }

    [Benchmark]
    public boolManualComparison()
    {
        if (dict1.Count != dict2.Count)
            return false;

        foreach (var kvp in dict1)
        {
            if (!dict2.TryGetValue(kvp.Key, out var value) || value != kvp.Value)
                return false;
        }

        return true;
    }

    public static void Main(string[] args)
    {
        var summary = BenchmarkRunner.Run<SortedDictionaryComparisonBenchmark>();
    }
}

src/client/Microsoft.Identity.Client/Cache/StorageJsonValues.cs
```diff
@@ -12,6 +12,7 @@ internal static class StorageJsonValues
         public const string TokenTypeBearer = "Bearer";
         public const string CredentialTypeRefreshToken = "RefreshToken";
         public const string CredentialTypeAccessToken = "AccessToken";
+        public const string CredentialTypeAccessTokenExtended = "AccessToken_Extended";
```
bgavrilMS, line 15: can we have ATExt  instead? Cache items should be as small as possible.

src/client/Microsoft.Identity.Client/Cache/StorageJsonKeys.cs
```diff
@@ -39,5 +39,7 @@ internal static class StorageJsonKeys
         // previous versions of MSAL used "ext_expires_on" instead of the correct "extended_expires_on".
         // this is here for back compatibility
         public const string ExtendedExpiresOn_MsalCompat = "ext_expires_on";
+
+        public const string AdditionalCacheKeyComponents = "additional_cache_key_components";
```
bgavrilMS, line 43: Can we have this be just "Ext" ?

src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs
```diff
@@ -170,6 +191,19 @@ internal void InitCacheKey()
                 _credentialDescriptor = StorageJsonValues.CredentialTypeAccessTokenWithAuthScheme;
             }
 
+            if (AdditionalCacheKeyComponents != null)
+            {
+                _credentialDescriptor = StorageJsonValues.CredentialTypeAccessTokenExtended;
+                if (_extraKeyParts != null)
+                {
+                    _extraKeyParts.Concat(new[] { CoreHelpers.ComputeAccessTokenExtCacheKey(AdditionalCacheKeyComponents) });
```
bgavrilMS, line 199: This is not covered by unit tests. YOu need a test where you combine POP with extra cache and validate the cache keys.

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
bgavrilMS: Here are the results: 

image

`ManualComparison` is the <string, string> version. I used SortedList. It looks statistically relevant.

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -11,6 +13,34 @@ namespace Microsoft.Identity.Client.Extensibility
...
+        /// <param name="cacheKeyComponents">The list of additional cache key components.</param>
+        /// <param name="builder"></param>
+        /// <returns>The builder.</returns>
+        /// <remarks>
+        /// <list type="bullet">
+        /// <item><description>This api can be used to associate certificate key identifiers along with other keys with a particular token.</description></item>
+        /// <item><description>In order for the tokens to be successfully retrieved from the cache, all components used to cache the token must be provided.</description></item>
+        /// </list>
+        /// </remarks>
+        public static AcquireTokenForClientParameterBuilder WithAdditionalCacheKeyComponents(this AcquireTokenForClientParameterBuilder builder,
```
bgavrilMS, line 28: Let's go with @rayluo 's suggestion and keep this internal for now

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
trwalke: - The spec uses a dictionary
- The the cache key components are a key value pair
So with that, a list didn't seem like the best data type to use here.

trwalke: Updated to sortedList

src/client/Microsoft.Identity.Client/Utils/CollectionHelpers.cs
trwalke: Switching to SortedList

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -11,6 +13,34 @@ namespace Microsoft.Identity.Client.Extensibility
...
+        {
+            builder.ValidateUseOfExperimentalFeature();
+
+            if (cacheKeyComponents == null || cacheKeyComponents.Count == 0)
+            {
+                //no-op
+                return builder;
+            }
+
+            builder.CommonParameters.CacheKeyComponents = new SortedList<string, string>(cacheKeyComponents);
```
bgavrilMS, line 39: Can you pls make this additive? In case 2 MISE extensions use this. i.e. if CacheKeyComponents  already exists, the entries.

bgavrilMS (approved):
LGTM, 1 minor comment.

gladjohn (approved):
looks good @trwalke 

I do not see any explicit tests ensuring that app developers cannot add standard keys (e.g. "client_id", "tenant_id") to `AdditionalCacheKeyComponents`. Can that be added?

PR #5123, opened 2025-02-07
Removing known key check for cache key extensions

trwalke:
Fixes #
Removing known key check for cache key extension. This is no longer needed because cache key components are added to an additional node which prevents clashing with known JSON keys.

Removing known key check for cache key extension.

**Testing**

**Performance impact**

**Documentation**


bgavrilMS (approved):

gladjohn:
please add a work item

gladjohn (approved):

Issue #5111, opened 2025-02-05
[Feature Request] .WithAccessTokenSha256ToRefresh(string hash) and .WithAccessTokenToRefresh(string oldToken)


rayluo:
### MSAL client type

Confidential, Managed identity

### Problem statement

Currently, MSAL provides only unconditional token refresh. We need to support precisely specifying an old token to be refreshed. Details described in [Precise Token Cache Refresh proposal](https://microsoft.sharepoint-df.com/:w:/r/teams/MSALConfidentialClient9/_layouts/15/Doc2.aspx?action=edit&sourcedoc=%7B3197a27c-2828-4d62-b902-fddcabc7908d%7D&wdOrigin=TEAMS-MAGLEV.teamsSdk_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1734524234579&web=1&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiI0OS8yNDEyMDUxODEwMiJ9&share=IQF8opcxKChiTbkC_dyrx5CNARUCwwE2p3TUZioajYhlQN8).

### Proposed solution

A Minimal Viable Product (MVP) shall implement:

* Chapter 5 of the proposal above
  - `ConfidentialClientApplication.AcquireTokenForClient().WithAccessTokenSha256ToRefresh(string hash)`
  - ~`ManagedIdentityClient.AcquireTokenForClient().WithAccessTokenToRefresh(string token)`~
* `.WithClaim(...)` behavior change described in Appendix 1 of the proposal above

### Acceptance Tests

The following test cases are meant to be tested sequentially.

1. CCA
   1. Prepopulate the token cache of a CCA with a token `token1`.
   2. Test a new `CCA.AcquireTokenForClient().WithAccessTokenSha256ToRefresh("mismatchingHash")` should not trigger refresh.
   3. Test matching hash `CCA.AcquireTokenForClient().WithAccessTokenSha256ToRefresh("hashOfToken1")` should trigger refresh and obtain a new token `token2`.
   4. `CCA.AcquireTokenForClient().WithAccessTokenSha256ToRefresh("hashOfToken1").WithClaims("...")`. This is a client using old token1, with proper hash(token1), even with claims challenge, should NOT trigger refresh, because we can serve it with token2 in cache.

[...]

PR #5137, opened 2025-02-12
MSI v1 token revocation specification

gladjohn:
This pull request adds documentation for MSAL support for MSI v1 token revocation and capabilities. It introduces two optional parameters that developers can use to control token revocation and specify client capabilities when acquiring tokens via Managed Identity.

Key changes include:

* **Documentation Addition**:
  * [`docs/msiv1_token_revocation.md`](diffhunk://#diff-708928b0ae864249d92886b9476987f030e1ffa83a0f529e2f6df380c06a7676R1-R48): Added detailed documentation describing the `bypass_cache` and `xms_cc` parameters, their purposes, behaviors, and use cases.


docs/msiv1_token_revocation.md
```diff
@@ -0,0 +1,48 @@
+# MSAL Support for MSI v1 Token Revocation and Capabilities
+
+This specification describes **two optional parameters** that MSAL can send to the **MSI v1 endpoint** (e.g., `http://169.254.169.254/metadata/identity/oauth2/token`) when acquiring tokens via **Managed Identity**. These parameters empower developers to control **token revocation** and specify **client capabilities** that might alter the token issuance behavior by Azure AD (ESTS).
+
+---
+
+## Overview
+
+When MSAL requests a token from the MSI v1 endpoint for resources like `https://management.azure.com/`, it typically reuses valid, locally cached tokens. However, certain scenarios require explicitly **ignoring** the cache or specifying **client capabilities** that Azure AD uses to enable or disable particular features.
```
bgavrilMS, line 9: I am confused by this spec. `bypass_cache` and `xms_cc` are new internal params for MSIv1. We don't expose these to app developers. What do we expose to app developers ? WithClaims and WithClientCapabilities?

If so, you need the APIs from the app developers to the ppl with Claims. 

We should also have details about how to E2E test this (even if manual) including how to invoke token revocation.

bgavrilMS (changes requested):
No mapping between public APIs and the new MSI params.

docs/msiv1_token_revocation.md
rayluo: > What do we expose to app developers ? WithClaims and WithClientCapabilities?

[`WithClaims` has already been exposed](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4.68.0/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs#L74).  `WithClientCapabilities` seems unnecessary because CAE is automatically enabled for S2S traffic, possibly also for MSI.

As to bypass_cache and xms_cc, they are internal for sure. And it seems no service is currently supporting them. We can follow up the conversation on Teams.

gladjohn: addressed comments

gladjohn: @rayluo we do have a way to test this now. I've added the doc link

docs/msiv1_token_revocation.md
```diff
@@ -0,0 +1,74 @@
...
+var result = await mi.AcquireTokenForManagedIdentity(new[] { "https://management.azure.com/.default" })
+    .WithClaims(Claims)
+    .ExecuteAsync()
+    .ConfigureAwait(false);
+
+```
+
+## End to End testing 
+
+Given the complexity of the scenario, it may not be easy to automate this. Here is the [guideline](https://microsoft.sharepoint.com/:w:/t/AzureMSI/ESBeuafJLZdNlSxkBKvjcswBD4FGVz0o6YJcf4mfDRSH-Q?e=2hJRUt) to test this feature manually using a Virtual Machine. For further details, please contact Gladwin.
```
rayluo, line 70: > ... to test this feature manually using a Virtual Machine

Did you mean Azure VM already implemented this? My impression is that they did not; they went with MSIv2.

gladjohn: let discuss on teams

bgavrilMS (dismissed):
Not the design SF wanted.

docs/msiv1_token_revocation.md
```diff
@@ -0,0 +1,74 @@
+# MSAL Support for MSI v1 Token Revocation and Capabilities
+
+This specification describes **two optional parameters** that MSAL can send to the **MSI v1 endpoint** (e.g., `http://169.254.169.254/metadata/identity/oauth2/token`) when acquiring tokens via **Managed Identity**. These parameters empower developers to control **token revocation** and specify **client capabilities** that might alter the token issuance behavior by Azure AD (ESTS).
+
+---
+
+## Overview
```
bgavrilMS, line 7: Needs more details on the E2E, i.e. how does token revocation happen? i.e. what script do I run as a tenant admin?

docs/msiv1_token_revocation.md
```diff
@@ -0,0 +1,86 @@
...
+
+rect rgb(173, 216, 230)
+    CX->> Resource: 1. Call resource with "bad" token T
+    Resource->>CX: 2.HTTP 401 + claims C
+    CX->>CX: 3. Parse response, extract claims C
+    CX->>MSAL: 4. MSI.AcquireToken <br/> WithClaims(C) <br/> WithClientCapabilities("cp1")
+end
+rect rgb(215, 234, 132)
+    MSAL->>MSAL: 5. Get token T from cache. Assume it is the "bad" token.
+    MSAL->>SF: 6. Call MITS_endpoint?xms_cc=cp1&revoked_token=SHA256(T)
```
rayluo, line 29:
```suggestion
    MSAL->>SF: 6. Call MITS_endpoint?claims=C&token_sha256_to_refresh=SHA256(T)
```

docs/msiv1_token_revocation.md
```diff
@@ -0,0 +1,86 @@
...
+    participant MSAL
+    participant SF        
+    participant eSTS
+
+rect rgb(173, 216, 230)   
+    CX->>MSAL: 1. MSI.AcquireToken <br/> WithClientCapabilities("cp1")
+    MSAL->>MSAL: 2. Find and return token T in cache. <br/>If not found, goto next step.
+end
+rect rgb(215, 234, 132)    
+    MSAL->>SF: 3. Call MITS_endpoint?xms_cc=cp1
```
rayluo, line 58: Either simply remove `xms_cc`

```suggestion
    MSAL->>SF: 3. Call MITS_endpoint
```

or do it in the standard CAE way which is MSAL automatically converts xms_cc=cp1 into a `claims={"access_token": {"xms_cc": {"values": ["cp1"]}}}` and send that blob to the wire.

rayluo (commented):
Based on the original CAE design, the [`WithClientCapabilities` is a per-app parameter in MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/68dc10fe42c8d7cf549c7b5ac274c18d5e4eef91/src/client/Microsoft.Identity.Client/AppConfig/AbstractApplicationBuilder.cs#L326). That is one of the reason why `xms_cc` should not be used in the second last arrow in [both diagrams](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/gladjohn-msi_v1/docs/msiv1_token_revocation.md), otherwise it would force SFRP to create new CCA instance per request, which would be too inefficient.

In the SF's MSIv1 token revocation design, their protocol accepts the `claims` parameter (signed off by eSTS architect), and it does not accept an `xms_cc` as a standalone parameter (which is not in original CAE protocol either). So, MSAL's ManagedIdentityApplication can simply relays the incoming WithClaims to SF.

I proposed suggestions on diagrams, accordingly.

You may also refer to [this feature request](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5111) for the same project.

docs/msiv1_token_revocation.md
bgavrilMS: How will this work Ray? Claims and client capabilities need to be passed separately, because claims bypass cache and cp1 doesn't.

docs/msiv1_token_revocation.md
bgavrilMS: MSAL needs to have that claims and capabilities separately, as they have different behavior. claims bypases cache, cp does not.

bgavrilMS: Also, if memory serves, the call to MITS is a GET call. So the claims parameter could lead to an URL too long issue. Not sure if this is really a problem since it's ussually browsers that refuse long urls.

bgavrilMS (approved):

PR #5157, opened 2025-02-26
Update benchmark-action.yml to use NET 8

### LINKED ISSUE #5159: [Engineering task] Update Benchmark to use Net 8 from deprecated Net 6
### Task type

Builds

### Description

Update Benchmark to use Net 8 from deprecated Net 6

### Solution

Update Benchmark to use Net 8 from deprecated Net 6

gladjohn:
Fixes #5159

This pull request includes updates to the `.github/workflows/benchmark-action.yml` file to ensure compatibility with the latest .NET version.

Updates to `.github/workflows/benchmark-action.yml`:

* Updated the `dotnet-version` from `6.0.418` to `8.0.x` to use the latest .NET version.
* Changed the target framework in the benchmark run command from `net6.0` to `net8.0` to align with the updated .NET version.

**Testing**
build

**Performance impact**
none

**Documentation**

### DIFF
--- a/.github/workflows/benchmark-action.yml
+++ b/.github/workflows/benchmark-action.yml
@@ -26,9 +26,9 @@ jobs:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
       - uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: '6.0.418'
+          dotnet-version: '8.0.x'
       - name: Run benchmark
-        run: cd tests/Microsoft.Identity.Test.Performance && dotnet run -c release -f net6.0 --exporters json 
+        run: cd tests/Microsoft.Identity.Test.Performance && dotnet run -c release -f net8.0 --exporters json 
 
       - name: Download previous benchmark data
         uses: actions/cache@e12d46a63a90f2fae62d114769bbf2a179198b5c # v3.3.3
--- a/.github/workflows/benchmark-action.yml
+++ b/.github/workflows/benchmark-action.yml
@@ -31,7 +31,7 @@ jobs:
         run: cd tests/Microsoft.Identity.Test.Performance && dotnet run -c release -f net8.0 --exporters json 
 
       - name: Download previous benchmark data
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark-action.yml
```diff
@@ -26,9 +26,9 @@ jobs:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
       - uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: '6.0.418'
+          dotnet-version: '8.0.x'
       - name: Run benchmark
-        run: cd tests/Microsoft.Identity.Test.Performance && dotnet run -c release -f net6.0 --exporters json 
+        run: cd tests/Microsoft.Identity.Test.Performance && dotnet run -c release -f net8.0 --exporters json
```
bgavrilMS, line 31: Why don't we delete it instead? Are we using this?

gladjohn: afaik we are using it, tagging @pmaytak to confirm

trwalke: I remember we discussed this very thing. As I understand it, We are comparing the performance of the PRs to a baseline that is acquired regularly. (Daily I believe).

trwalke (approved):

.github/workflows/benchmark-action.yml
gladjohn: I see it as disabled now, when I checked it. 

image

@bgavrilMS is the suggestion to delete? we don't benchmark anymore?

gladjohn: @bgavrilMS can you please suggest

pmaytak: I think our agreement was that while this approach is not super accurate it can be used to show any abnormal changes in performance; so it still has value. This was supposed to be an intermediate step before a more permanent solution.

pmaytak (approved):

.github/workflows/benchmark-action.yml
gladjohn: Thanks for confirming this @pmaytak. It makes sense merge this and turn on the action @bgavrilMS

PR #5173, opened 2025-03-05
Adds NetStandard support for MTLS

### LINKED ISSUE #5172: [Bug] MSAL netstandard should not be excluded from mTLS
### Library version used

latest

### .NET version

In the lastest MSAL (4.69.1), we exclude mTLS operations from netstandard, via the SUPPORTS_MTLS feature. But HttpClientHandler.ClientCertificates underlying API is supported on NETSTANDARD. There is not reason to block this on netstandard. It will lead to "breaking" changes when we deprecate old frameworks like net6.

### Scenario

In the lastest MSAL (4.69.1), we exclude mTLS operations from netstandard, via the SUPPORTS_MTLS feature. But HttpClientHandler.ClientCertificates underlying API is supported on NETSTANDARD. There is not reason to block this on netstandard. It will lead to "breaking" changes when we deprecate old frameworks like net6.

### Relevant code snippets

```csharp

```

### Expected behavior

Add MTLS Support to Net Standard

### Identity provider

### Regression

latest 

### Solution and workarounds

Add MTLS Support to Net Standard

gladjohn:
Fixes #5172

This pull request includes changes to improve platform compatibility and code simplification. The most important changes include updating the target framework conditions in the `platform_and_feature_flags.props` file and simplifying the creation of `HttpClientHandler` in the `SimpleHttpClientFactory` class.

### Platform Compatibility:
* [`build/platform_and_feature_flags.props`](diffhunk://#diff-ab8426b220a9645209f54a0e3e841745bcac9689b4506e8f1eb8ce959b6bd058L8-R8): Updated the condition to include `$(TargetFrameworkNetStandard)` for defining `SUPPORTS_MTLS` constants.

### Code Simplification:
* [`src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs`](diffhunk://#diff-c3cb436aeedc6aedc1e600b05baa129916cb26e430db7f54e4fca6c76fc6cc97L50-R50): Simplified the instantiation of `HttpClientHandler` by using target-typed new expressions.

**Testing**
Tested by adding a Net6 Project. 

image

And modified the 

image

Ensured in net 8, the code path is not active, 

image

And tested by running the console app, and acquired MTLS token successfully

image

**Performance impact**
none

**Documentation**

### DIFF
--- a/build/platform_and_feature_flags.props
+++ b/build/platform_and_feature_flags.props
@@ -5,7 +5,7 @@
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet)' or '$(TargetFramework)' == '$(TargetFrameworkNetDesktop462)' or '$(TargetFramework)' == '$(TargetFrameworkNetDesktop472)' or '$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <DefineConstants>$(DefineConstants);SUPPORTS_OTEL;</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet)' or '$(TargetFramework)' == '$(TargetFrameworkNetDesktop472)'">
+  <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNet)' or '$(TargetFramework)' == '$(TargetFrameworkNetDesktop472)' or '$(TargetFramework)' == '$(TargetFrameworkNetStandard)'">
     <DefineConstants>$(DefineConstants);SUPPORTS_MTLS;</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetAndroid)'">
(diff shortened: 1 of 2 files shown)


neha-bhargava:
Which tests include coverage for NetStandard support?

gladjohn:
> Which tests include coverage for NetStandard support?

will testing in .NET 6.0 to just confirm runtime-level functionality is enough? or do need to confirm to a genuine .NET Standard 2.0 support. That means adding support to our unit testing at a min. Think at some point we used to have it and removed it.

bgavrilMS (approved):

bgavrilMS:
You can do a manual validation in this case, by running a net6 test or just running that piece of code in a console app.
We shouldn't run tests on net6. net6 is out of support. The .NET team guarantees that netstandard behaves like net6.

Please note that MSIv2 cannot replace MSIv1 on net462 if it uses mTLS. We cannot break ppl. But we'll probably drop mTLS so it's fine.

gladjohn:
> Please note that MSIv2 cannot replace MSIv1 on net462 if it uses mTLS. We cannot break ppl. But we'll probably drop mTLS so it's fine.

Should we throw if `net462` is used and if the source is determined as `Credential`? or just let `ests` do it's thing?

trwalke (approved):

gladjohn:
> You can do a manual validation in this case, by running a net6 test or just running that piece of code in a console app.

Validation steps added to the description. And testing was successful.

PR #5175, opened 2025-03-07
Adding a flag to TokenCacheNotificationArgs to signal that the clientId is an RMA node.

trwalke:
Fixes #
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3174516

Adding a flag to `TokenCacheNotificationArgs` to signal that the clientId is an RMA node.

**Testing**
Unit testing

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
@@ -258,7 +258,7 @@ namespace Microsoft.Identity.Client
         public TelemetryData TelemetryData { get; }
 
         /// <summary>
-        /// Determines if the client application authentication is an FMI node under an RMA.
+        /// Determines whether the client application authentication instance is classified as an FMI (Federated Managed Identity) node under a specified RMA (Resource Managed Authority).
         /// </summary>
         public bool IsFmiClientNode 
         {
--- a/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
+++ b/src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Threading;
+using Microsoft.Identity.Client.Internal;
 using Microsoft.Identity.Client.TelemetryCore.TelemetryClient;
 using Microsoft.IdentityModel.Abstractions;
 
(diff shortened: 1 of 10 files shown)


bgavrilMS (approved):

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -255,5 +256,13 @@ public TokenCacheNotificationArgs(    // only use this constructor in product co
         /// Cache Details contains the details of L1/ L2 cache for telemetry logging.
         /// </summary>
         public TelemetryData TelemetryData { get; }
+
+        /// <summary>
+        /// Determines if the client application authentication is an FMI node under an RMA.
```
gladjohn, line 261:
```suggestion
        /// Determines whether the client application authentication instance is classified as an FMI (Federated Managed Identity) node under a specified RMA (Resource Managed Application).
```

gladjohn (approved):

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
bgavrilMS: RMA stands for `Resource Management Authority`

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
```diff
@@ -255,5 +256,13 @@ public TokenCacheNotificationArgs(    // only use this constructor in product co
...
+        /// </summary>
+        public bool IsFmiClientNode
```
bgavrilMS, line 263: I think @jmprieur provided feedback that this is too specific for FMI and a more general purpose mechanism can be used. 

In my opinion, Id.Web Cache project needs to throw a very clear error message of why distributed cache cannot be used with FMI. 

Maybe we want to expose a `string NoDistributedCacheUseReason` with the error message?

jmprieur: Would be a good idea.
The presence of the message would actually act as the bool?

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
gladjohn: updated the comment, thanks @bgavrilMS

src/client/Microsoft.Identity.Client/TokenCacheNotificationArgs.cs
trwalke: Sure, I can do that

trwalke:
Accompanying PR for IdWeb
https://github.com/AzureAD/microsoft-identity-web/pull/3295
@jmprieur @bgavrilMS

PR #5139, opened 2025-02-12
initial

gladjohn:
Fixes #5138

**Changes proposed in this request**
This pull request includes several changes to the `Microsoft.Identity.Client` library to support claims and capabilities in managed identity requests. The most important changes involve adding a new `Claims` property, modifying request creation methods to include this property, and implementing a new method to apply claims and capabilities to requests.

### Support for Claims and Capabilities:

* [`src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForManagedIdentityParameters.cs`](diffhunk://#diff-8a2dae698d340180c3ad81c8c6d92b32dce6ba143a830f70ee279f46eea034b9R19-R20): Added a new `Claims` property to the `AcquireTokenForManagedIdentityParameters` class.

* [`src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs`](diffhunk://#diff-a1260a45bdc4ac7ab60d4a348996be6b69c5a48fa6f484fa41daa390c7e69ca8R39): Updated the `ExecuteAsync` method to set the `Claims` property in `_managedIdentityParameters`.

### Request Creation and Handling:

* [`src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs`](diffhunk://#diff-26925862fc2f25040b857e0a110d355651a21e695340d3683565fba12fa30001L51-R53): Modified the `CreateRequest` method to accept `AcquireTokenForManagedIdentityParameters` and added the `ApplyClaimsAndCapabilities` method to set request parameters based on claims and capabilities. [[1]](diffhunk://#diff-26925862fc2f25040b857e0a110d355651a21e695340d3683565fba12fa30001L51-R53) [[2]](diffhunk://#diff-26925862fc2f25040b857e0a110d355651a21e695340d3683565fba12fa30001L128-R130) [[3]](diffhunk://#diff-26925862fc2f25040b857e0a110d355651a21e695340d3683565fba12fa30001R298-R346)

[...]

PR #5179, opened 2025-03-08
Enhancing Token Refresh Control in MSAL: Introducing .WithAccessTokenSha256ToRefresh() in AcquireTokenForClient Flows

### LINKED ISSUE #5111: [Feature Request] .WithAccessTokenSha256ToRefresh(string hash) and .WithAccessTokenToRefresh(string oldToken)

gladjohn:
Fixes #5111

**Specification -** https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/docs/msiv1_token_revocation.md

This pull request includes several changes to enhance token acquisition and revocation capabilities in the MSAL library. The most important changes focus on adding support for token hash-based refresh, improving cache handling, and updating error handling.

### Enhancements to Token Acquisition and Revocation:

* Added a new method `WithAccessTokenSha256ToRefresh` to the `AcquireTokenForClientParameterBuilder` for configuring the SDK to bypass the cache for a specific token based on its SHA-256 hash.
* Introduced a new property `AccessTokenHashToRefresh` in `AcquireTokenForClientParameters` to store the hash of the token to be refreshed.
* Updated `ClientCredentialRequest` to handle token hash-based refresh logic, including checking the cache for matching tokens and bypassing the cache if necessary. [[1]](diffhunk://#diff-655d5868568553085143b1c4df713cb7baf07c35ab78610a8ac73da1e467ce43R26) [[2]](diffhunk://#diff-655d5868568553085143b1c4df713cb7baf07c35ab78610a8ac73da1e467ce43L47-R68) [[3]](diffhunk://#diff-655d5868568553085143b1c4df713cb7baf07c35ab78610a8ac73da1e467ce43R209-R287)

### Error Handling Improvements:

* Added a new error `ForceRefreshNotCompatibleWithTokenHash` to indicate that force refresh and token hash cannot be used together. [[1]](diffhunk://#diff-e5c43e5cc8296950029f80cce9e2b350341fcdc994737931e50a2d43d310878eR1180-R1190) [[2]](diffhunk://#diff-5f7fcb836912bee0deb52d9dbdedafd36ce6fad65c20327ea497847edb7f2fdeR440) [[3]](diffhunk://#diff-d8d9399ed6a37cdd3bf25cd223b32bfe33318862404ac5bba32f34e213fb6eb5R1-R3)

### Internal Code Changes:

* Changed the visibility of the `Parameters` property in `AcquireTokenForClientParameterBuilder` from private to internal.
* Added validation in `AcquireTokenForClientParameterBuilder` to throw an exception if both force refresh and token hash are used together.

### Documentation Updates:

* Removed outdated documentation on MSI v1 token revocation and capabilities from `docs/msiv1_token_revocation.md`.

These changes collectively improve the flexibility and robustness of token handling in the MSAL library.

**Testing**
unit tests

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
@@ -157,6 +157,26 @@ namespace Microsoft.Identity.Client
             return this;
         }
 
+        /// <summary>
+        /// Specifies that the token refresh should only occur if the specified SHA-256 hash
+        /// of the previously issued access token matches a cached token.
+        /// If the hash does not match, the existing cached token is returned without refresh.
+        /// </summary>
+        /// <param name="hash">The SHA-256 hash of the access token to refresh.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenForClientParameterBuilder WithAccessTokenSha256ToRefresh(string hash)
+        {
+            ValidateUseOfExperimentalFeature();
+
+            if (string.IsNullOrWhiteSpace(hash))
+            {
+                throw new ArgumentNullException(nameof(hash), "Access token hash cannot be null or empty.");
+            }
+
+            Parameters.AccessTokenHashToRefresh = hash;
+            return this;
+        }
+
         /// <inheritdoc/>
         internal override Task<AuthenticationResult> ExecuteInternalAsync(CancellationToken cancellationToken)
         {
@@ -182,6 +202,14 @@ namespace Microsoft.Identity.Client
             }
 
             base.Validate();
+
+            // Force refresh + AccessTokenHashToRefresh APIs cannot be used together
+            if (Parameters.ForceRefresh && !string.IsNullOrEmpty(Parameters.AccessTokenHashToRefresh))
+            {
+                throw new MsalClientException(
+                    MsalError.ForceRefreshNotCompatibleWithTokenHash,
+                    MsalErrorMessage.ForceRefreshAndTokenHasNotCompatible);
+            }
 
             if (Parameters.SendX5C == null)
             {
--- a/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Identity.Client
     public sealed class AcquireTokenForClientParameterBuilder :
         AbstractConfidentialClientAcquireTokenParameterBuilder<AcquireTokenForClientParameterBuilder>
     {
-        private AcquireTokenForClientParameters Parameters { get; } = new AcquireTokenForClientParameters();
+        internal AcquireTokenForClientParameters Parameters { get; } = new AcquireTokenForClientParameters();
 
         /// <inheritdoc/>
         internal AcquireTokenForClientParameterBuilder(IConfidentialClientApplicationExecutor confidentialClientApplicationExecutor)
@@ -154,26 +154,6 @@ namespace Microsoft.Identity.Client
 
             CommonParameters.FmiPathSuffix = pathSuffix;
 
-            return this;
-        }
-
-        /// <summary>
-        /// Specifies that the token refresh should only occur if the specified SHA-256 hash
-        /// of the previously issued access token matches a cached token.
-        /// If the hash does not match, the existing cached token is returned without refresh.
-        /// </summary>
-        /// <param name="hash">The SHA-256 hash of the access token to refresh.</param>
-        /// <returns>The builder to chain the .With methods</returns>
-        public AcquireTokenForClientParameterBuilder WithAccessTokenSha256ToRefresh(string hash)
-        {
-            ValidateUseOfExperimentalFeature();
-
-            if (string.IsNullOrWhiteSpace(hash))
-            {
-                throw new ArgumentNullException(nameof(hash), "Access token hash cannot be null or empty.");
-            }
-
-            Parameters.AccessTokenHashToRefresh = hash;
             return this;
         }
 
--- a/src/client/Microsoft.Identity.Client/Extensibility/RP/AcquireTokenForClientParameterBuilderForResourceProviders.cs
+++ b/src/client/Microsoft.Identity.Client/Extensibility/RP/AcquireTokenForClientParameterBuilderForResourceProviders.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.ApiConfig.Parameters;
+
+namespace Microsoft.Identity.Client.RP
+{
+    /// <summary>
+    /// Resource Provider extensibility methods for AcquireTokenForClientParameterBuilder
+    /// </summary>
+    public static class AcquireTokenForClientParameterBuilderForResourceProviders
+    {
+        /// <summary>
+        /// Configures the SDK to not retrieve a token from the cache if it matches the SHA256 hash 
+        /// of the token configured. Similar to WithForceRefresh(bool) API, but instead of bypassing 
+        /// the cache for all tokens, the cache bypass only occurs for 1 token
+        /// </summary>
+        /// <param name="builder">The existing AcquireTokenForClientParameterBuilder instance.</param>
+        /// <param name="hash">
+        /// A Base64-encoded SHA-256 hash of the token (UTF-8). For example:
+        /// <c>Convert.ToBase64String(SHA256(Encoding.UTF8.GetBytes(accessToken)))</c>.
+        /// </param>
+        /// <returns>The builder to chain the .With methods.</returns>
+        public static AcquireTokenForClientParameterBuilder WithAccessTokenSha256ToRefresh(
+            this AcquireTokenForClientParameterBuilder builder,
+            string hash)
+        {
+            if (!string.IsNullOrWhiteSpace(hash))
+            {
+                builder.Parameters.AccessTokenHashToRefresh = hash;
+            }
+
+            return builder;
+        }
+    }
+}
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -63,8 +63,6 @@ namespace Microsoft.Identity.Client.Internal.Requests
                 (!string.IsNullOrEmpty(AuthenticationRequestParameters.Claims) && 
                 string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh));
 
-            // Skip checking cache when either ForceRefresh is true
-            // or (Claims are present without a token hash).
             if (skipCache)
             {
                 AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.ForceRefreshOrClaims;
@@ -206,79 +204,49 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
         private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
         {
-            // 1) Get the cached item 
-            MsalAccessTokenCacheItem cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
+            // Fetch the cache item (could be null if none found).
+            MsalAccessTokenCacheItem cacheItem =
+                await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
 
-            // 2) If no cached item or force refresh is requested, return null
-            if (!IsValidCachedToken(cachedAccessTokenItem))
+            // If the item fails any checks (null, or hash mismatch),
+            if (!ShouldUseCachedToken(cacheItem))
             {
                 return null;
             }
 
-            // 3) If there is a matching AccessTokenHashToRefresh, ignore this cached token
-            if (IsTokenIgnoredByMatchingHash(cachedAccessTokenItem))
-            {
-                return null;
-            }
-
-            // 4) Otherwise, record a cache hit and return the cached token
+            // Otherwise, record a successful cache hit and return the token.
             MarkAccessTokenAsCacheHit();
-            return cachedAccessTokenItem;
+            return cacheItem;
         }
 
-        /// <summary>
-        /// Checks if the cached access token can be used.
-        /// </summary>
-        private bool IsValidCachedToken(MsalAccessTokenCacheItem cachedAccessTokenItem)
+        private bool ShouldUseCachedToken(MsalAccessTokenCacheItem cacheItem)
         {
-            // Return false if the cache is empty
-            if (cachedAccessTokenItem == null)
+            // 1) No cached item 
+            if (cacheItem == null)
             {
+                return false;
+            }
+
+            // 2) If the token’s hash matches AccessTokenHashToRefresh, ignore it
+            if (!string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh) &&
+                IsMatchingTokenHash(cacheItem.Secret, _clientParameters.AccessTokenHashToRefresh))
+            {
+                AuthenticationRequestParameters.RequestContext.Logger.Info(
+                    "[ClientCredentialRequest] A cached token was found and its hash matches AccessTokenHashToRefresh, so it is ignored.");
                 return false;
             }
 
             return true;
         }
 
-        /// <summary>
-        /// Determines whether the cached token should be ignored due to matching the AccessTokenHashToRefresh.
-        /// </summary>
-        private bool IsTokenIgnoredByMatchingHash(MsalAccessTokenCacheItem cachedAccessTokenItem)
+        private bool IsMatchingTokenHash(string tokenSecret, string accessTokenHashToRefresh)
         {
-            if (string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh))
-            {
-                return false;
-            }
-
-            string cachedTokenHash = _cryptoManager.CreateSha256Hash(cachedAccessTokenItem.Secret);
-
-            // If the hash of the cached token matches the hash to refresh, ignore the cached token
-            bool matchesHash = string.Equals(
-                cachedTokenHash,
-                _clientParameters.AccessTokenHashToRefresh,
-                StringComparison.Ordinal);
-
-            if (matchesHash)
-            {
-                AuthenticationRequestParameters.RequestContext.Logger.Info(
-                    "[ClientCredentialRequest] A cached token was found and its hash matches AccessTokenHashToRefresh, so it is ignored.");
-                return true;
-            }
-            else
-            {
-                AuthenticationRequestParameters.RequestContext.Logger.Info(
-                    "[ClientCredentialRequest] A cached token was found, but its hash does NOT match AccessTokenHashToRefresh. Using the cached token.");
-                return false;
-            }
+            string cachedTokenHash = _cryptoManager.CreateSha256Hash(tokenSecret);
+            return string.Equals(cachedTokenHash, accessTokenHashToRefresh, StringComparison.Ordinal);
         }
 
-        /// <summary>
-        /// Marks the current access token retrieval as a successful cache hit, 
-        /// and increments any relevant telemetry or counters.
-        /// </summary>
         private void MarkAccessTokenAsCacheHit()
         {
-            // Mark the request as a cache hit
             AuthenticationRequestParameters.RequestContext.ApiEvent.IsAccessTokenCacheHit = true;
             Metrics.IncrementTotalAccessTokensFromCache();
         }
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
 using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
@@ -47,8 +50,12 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
             AuthenticationResult authResult;
 
+            bool skipCache = _clientParameters.ForceRefresh || 
+                (!string.IsNullOrEmpty(AuthenticationRequestParameters.Claims) && 
+                string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh));
+
             // Skip checking cache when force refresh or claims are specified
-            if (_clientParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
+            if (skipCache)
             {
                 AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.ForceRefreshOrClaims;
                 logger.Info("[ClientCredentialRequest] Skipped looking for a cached access token because ForceRefresh or Claims were set.");
@@ -193,12 +200,38 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
             if (cachedAccessTokenItem != null && !_clientParameters.ForceRefresh)
             {
+                if (!string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh))
+                {
+                    string cachedTokenHash = ComputeSHA256(cachedAccessTokenItem.Secret);
+
+                    if (string.Equals(cachedTokenHash, _clientParameters.AccessTokenHashToRefresh, StringComparison.Ordinal))
+                    {
+                        AuthenticationRequestParameters.RequestContext.Logger.Info(
+                            "[ClientCredentialRequest] Found the 'bad' token in the cache. Skipping cache usage.");
+                        return null; // triggers a new token acquisition
+                    }
+                }
+
                 AuthenticationRequestParameters.RequestContext.ApiEvent.IsAccessTokenCacheHit = true;
                 Metrics.IncrementTotalAccessTokensFromCache();
                 return cachedAccessTokenItem;
             }
 
             return null;
+        }
+
+        private static string ComputeSHA256(string token)
+        {
+#if NET6_0_OR_GREATER
+            byte[] hashBytes = SHA256.HashData(Encoding.UTF8.GetBytes(token));
+            return Convert.ToBase64String(hashBytes);
+#else
+            using (var sha256 = SHA256.Create())
+            {
+                byte[] hashBytes = sha256.ComputeHash(Encoding.UTF8.GetBytes(token));
+                return Convert.ToBase64String(hashBytes);
+            }
+#endif
         }
 
         private AuthenticationResult CreateAuthenticationResultFromCache(MsalAccessTokenCacheItem cachedAccessTokenItem)
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -203,7 +203,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
         }
 
         /// <summary>
-        /// Checks if the token should be used from the cache.
+        /// Checks if the token should be used from the cache and returns the cached access token if applicable.
         /// </summary>
         /// <returns></returns>
         private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Security.Cryptography;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
@@ -13,6 +14,7 @@ using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Extensibility;
 using Microsoft.Identity.Client.Instance;
 using Microsoft.Identity.Client.OAuth2;
+using Microsoft.Identity.Client.PlatformsCommon.Interfaces;
 using Microsoft.Identity.Client.Utils;
 
 namespace Microsoft.Identity.Client.Internal.Requests
@@ -21,6 +23,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
     {
         private readonly AcquireTokenForClientParameters _clientParameters;
         private static readonly SemaphoreSlim s_semaphoreSlim = new SemaphoreSlim(1, 1);
+        private readonly ICryptographyManager _cryptoManager;
 
         public ClientCredentialRequest(
             IServiceBundle serviceBundle,
@@ -29,6 +32,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
             : base(serviceBundle, authenticationRequestParameters, clientParameters)
         {
             _clientParameters = clientParameters;
+            _cryptoManager = serviceBundle.PlatformProxy.CryptographyManager;
         }
 
         protected override async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
@@ -47,14 +51,20 @@ namespace Microsoft.Identity.Client.Internal.Requests
             {
                 logger.Error(MsalErrorMessage.ClientCredentialWrongAuthority);
             }
-
             AuthenticationResult authResult;
 
+            // Skip cache if either:
+            // 1) ForceRefresh is set, or
+            // 2) Claims are specified and there is no AccessTokenHashToRefresh.
+            // This ensures that when both claims and AccessTokenHashToRefresh are set,
+            // we do NOT skip the cache, allowing MSAL to attempt retrieving a matching
+            // cached token by the provided hash before requesting a new token.
             bool skipCache = _clientParameters.ForceRefresh || 
                 (!string.IsNullOrEmpty(AuthenticationRequestParameters.Claims) && 
                 string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh));
 
-            // Skip checking cache when force refresh or claims are specified
+            // Skip checking cache when either ForceRefresh is true
+            // or (Claims are present without a token hash).
             if (skipCache)
             {
                 AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.ForceRefreshOrClaims;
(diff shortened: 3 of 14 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -158,6 +158,26 @@ public AcquireTokenForClientParameterBuilder WithFmiPath(string pathSuffix)
...
+        /// <param name="hash">The SHA-256 hash of the access token to refresh.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenForClientParameterBuilder WithAccessTokenSha256ToRefresh(string hash)
+        {
+            ValidateUseOfExperimentalFeature();
+
+            if (string.IsNullOrWhiteSpace(hash))
+            {
+                throw new ArgumentNullException(nameof(hash), "Access token hash cannot be null or empty.");
+            }
```
rayluo, line 175: **Thought(non-blocking)**
Alternatively, we may also simply accept null and empty string and even white spaces, and then simply treat them as a mismatching hash. Reason being, at one point, SF folks asked whether it would be OK to unconditionally call MSAL's `WithAccessTokenSha256ToRefresh()` with empty or null value, that way their implementation would be even simpler.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -193,6 +200,18 @@ private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
...
             {
+                if (!string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh))
+                {
+                    string cachedTokenHash = ComputeSHA256(cachedAccessTokenItem.Secret);
+
+                    if (string.Equals(cachedTokenHash, _clientParameters.AccessTokenHashToRefresh, StringComparison.Ordinal))
+                    {
+                        AuthenticationRequestParameters.RequestContext.Logger.Info(
+                            "[ClientCredentialRequest] Found the 'bad' token in the cache. Skipping cache usage.");
+                        return null; // triggers a new token acquisition
```
rayluo, line 211: **Suggestion(non-blocking)**
Shall we consider adding one more token refresh reason for telemetry?
Currently we have a reserved value `5`, so this time we may use `6`.

```suggestion
                        AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.TokenRejected;
                        return null; // triggers a new token acquisition
```

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -437,5 +437,6 @@ public static string InvalidTokenProviderResponseValue(string invalidValueName)
         public const string MtlsInvalidAuthorityTypeMessage = "mTLS PoP is only supported for AAD authority type. See https://aka.ms/msal-net-pop for details.";
         public const string MtlsNonTenantedAuthorityNotAllowedMessage = "mTLS authentication requires a tenanted authority. Using 'common', 'organizations', or similar non-tenanted authorities is not allowed. Please provide an authority with a specific tenant ID (e.g., 'https://login.microsoftonline.com/{tenantId}'). See https://aka.ms/msal-net-pop for details.";
         public const string RegionRequiredForMtlsPopMessage = "Regional auto-detect failed. mTLS Proof-of-Possession requires a region to be specified, as there is no global endpoint for mTLS. See https://aka.ms/msal-net-pop for details.";
+        public const string ForceRefreshAndTokenHasNotCompatible = "Cannot specify ForceRefresh and AccessTokenSha256ToRefresh in the same request.";
```
rayluo, line 440: **praise:**

Good job on coming up with this check which is not defined in the acceptance tests.

**Typo?**

```suggestion
        public const string ForceRefreshAndTokenHashNotCompatible = "Cannot specify ForceRefresh and AccessTokenSha256ToRefresh in the same request.";

```

rayluo (commented):
Overall looks good. Do we plan to implement the managed identity part of functionality in a different PR?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
gladjohn: @bgavrilMS ^^ thoughts?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: Sure @rayluo, we should consider a telemetry ask for all MSALs

gladjohn:
> Overall looks good. Do we plan to implement the managed identity part of functionality in a different PR?

yes @rayluo, here is the MIA PR for Token Revocation, I had marked this as blocked since we had to finalize the design. I can work on this now - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5139

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
rayluo: > a telemetry ask for all MSALs

Do you mean asking all MSALs to at least add a constant `TokenRejected = 6`, even if they don't pick up this WithAccessTokenSha256ToRefresh()? That makes sense.

@bgavrilMS please also comment on whether all ID4S MSALs shall support WithAccessTokenSha256ToRefresh() for consistency. Currently, the first customer of this PR will be SFRP which uses .Net only.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -158,6 +158,26 @@ public AcquireTokenForClientParameterBuilder WithFmiPath(string pathSuffix)
...
         }
 
+        /// <summary>
+        /// Specifies that the token refresh should only occur if the specified SHA-256 hash
+        /// of the previously issued access token matches a cached token.
+        /// If the hash does not match, the existing cached token is returned without refresh.
+        /// </summary>
+        /// <param name="hash">The SHA-256 hash of the access token to refresh.</param>
+        /// <returns>The builder to chain the .With methods</returns>
+        public AcquireTokenForClientParameterBuilder WithAccessTokenSha256ToRefresh(string hash)
```
gladjohn, line 168: @rayluo why did we not choose the name to be `WithAccessTokenToRefresh(string hash)`.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
bgavrilMS: Sure, makes sense to not throw exceptions. Simplest is to have this be a no-op, i.e. hash is null or empty, don't call `AccessTokenHashToRefresh`

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
bgavrilMS: There is an equivalent API in MSAL C++ that takes in the full plaintext token. 

IMO it might a slightly nicer dev-experience to use `WithForceRefresh(string hash)`

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -158,6 +158,26 @@ public AcquireTokenForClientParameterBuilder WithFmiPath(string pathSuffix)
             return this;
         }
 
+        /// <summary>
+        /// Specifies that the token refresh should only occur if the specified SHA-256 hash
```
bgavrilMS, line 162: `Configures the SDK to not retrieve a token from the cache if it matches the SHA256 hash of the token configured. Similar to WithForceRefresh(bool) API, but instead of bypassing the cache for all tokens, the cache bypass only occurs for 1 token`

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -47,8 +50,12 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
 
+            bool skipCache = _clientParameters.ForceRefresh || 
+                (!string.IsNullOrEmpty(AuthenticationRequestParameters.Claims) && 
+                string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh));
+
             // Skip checking cache when force refresh or claims are specified
-            if (_clientParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
+            if (skipCache)
             {
                 AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.ForceRefreshOrClaims;
```
bgavrilMS, line 60: Can we take a moment to also fix this? i.e. have a separate value in CacheRefreshReason for force refresh vs claims.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
bgavrilMS: +1, this API is for RPs. We won't expose it in other MSALs.

@rayluo - thoughts on adding to a separate namespace `Microsoft.Identity.Client.RP` (this already exists)

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -47,8 +50,12 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
                 logger.Info("[ClientCredentialRequest] Skipped looking for a cached access token because ForceRefresh or Claims were set.");
```
bgavrilMS, line 61: Pls fix this comment as well, we should know exactly why the cache was by-passed.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -47,8 +50,12 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
             AuthenticationResult authResult;
 
+            bool skipCache = _clientParameters.ForceRefresh || 
+                (!string.IsNullOrEmpty(AuthenticationRequestParameters.Claims) && 
+                string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh));
```
bgavrilMS, line 55: I don't understand the interaction between claims and `AccessTokenHashToRefresh`. I think we have 2 choices: 

1. Do not allow WithClaims and this API to be used toghether. OR
2. If both are set, bypass cache if claims AND a precise cache hit occurs.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -201,6 +220,20 @@ private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
             return null;
         }
 
+        private static string ComputeSHA256(string token)
+        {
+#if NET6_0_OR_GREATER
```
bgavrilMS, line 225: We don't compile MSAL on net6. Only on net8. Can we do `#if NET` instead?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -193,6 +200,18 @@ private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
 
             if (cachedAccessTokenItem != null && !_clientParameters.ForceRefresh)
             {
+                if (!string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh))
```
bgavrilMS, line 203: The [cyclomatic complexity](https://learn.microsoft.com/en-us/visualstudio/code-quality/code-metrics-cyclomatic-complexity?view=vs-2022) of this block of code just exploded. Pls use some private methods to hide some complexity away.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -193,6 +200,18 @@ private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
...
+                {
+                    string cachedTokenHash = ComputeSHA256(cachedAccessTokenItem.Secret);
+
+                    if (string.Equals(cachedTokenHash, _clientParameters.AccessTokenHashToRefresh, StringComparison.Ordinal))
+                    {
+                        AuthenticationRequestParameters.RequestContext.Logger.Info(
+                            "[ClientCredentialRequest] Found the 'bad' token in the cache. Skipping cache usage.");
```
bgavrilMS, line 210:
```suggestion
                            "[ClientCredentialRequest] A cached token was found, but it matches the AccessTokenHashToRefresh so it will be skipped");
```

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -193,6 +200,18 @@ private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
 
             if (cachedAccessTokenItem != null && !_clientParameters.ForceRefresh)
             {
+                if (!string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh))
+                {
+                    string cachedTokenHash = ComputeSHA256(cachedAccessTokenItem.Secret);
+
+                    if (string.Equals(cachedTokenHash, _clientParameters.AccessTokenHashToRefresh, StringComparison.Ordinal))
```
bgavrilMS, line 207: Pls add an `else` clause with a logging statement.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -201,6 +220,20 @@ private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
             return null;
         }
 
+        private static string ComputeSHA256(string token)
```
bgavrilMS, line 223: nit: consider adding this to CryptoManager instead.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
bgavrilMS, line 162: You need to provide clear guidance in the API docs on the encoding used to transform from byte[] to string, i.e. `Convert.ToBase64String`.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
rayluo: > thoughts on adding to a separate namespace Microsoft.Identity.Client.RP (this already exists)

The design of WithAccessTokenSha256ToRefresh() is generic, and conceptually better than ForceRefresh (functionally comparable and has no risk being misused). So, it does not harm to add it everywhere. MSAL C++ has the similar `AccessTokenToRefresh` for years now.

That being said, it is totally OK to only implement it in MSAL .Net and in its `Microsoft.Identity.Client.RP` space only. This way, we have minimal workload today. We may expose it in broader surface when a need arises in the future. This is also a good decision.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
rayluo: > why did we not choose the name to be `WithAccessTokenToRefresh(string hash)`.

In the early editions of the design, there were actually two APIs planned, `WithAccessTokenToRefresh(string rawToken)` and `WithAccessTokenSha256ToRefresh(string hash)`. Later we realized that for SFRP, they will have only the hash, not the raw old token, so, currently only `WithAccessTokenSha256ToRefresh(string hash)` is required. But still, they are two different things, so, we shall not mix up their naming. This way, we won't paint ourselves into a corner when/if we will need to support `WithAccessTokenToRefresh(string rawToken)` in the future.

> IMO it might a slightly nicer dev-experience to use WithForceRefresh(string hash)

I can understand this perspective of reusing the same name established from `WithForceRefresh(bool b)`, but it feels like a dead end because we won't be able to easily introduce the 3rd `WithForceRefresh(string rawToken)` when a need arises. Besides, the `WithForceRefresh(true)` is intrinsically susceptible to being abused, that's also why conceptually we shall move away from that old api & concept. Hopefully we may deprecate `WithForceRefresh(true)` someday in the future.

gladjohn:
@rayluo @Robbie-Microsoft can I get a review from you as well please

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -44,11 +51,21 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            // 2) Claims are specified and there is no AccessTokenHashToRefresh.
+            // This ensures that when both claims and AccessTokenHashToRefresh are set,
+            // we do NOT skip the cache, allowing MSAL to attempt retrieving a matching
+            // cached token by the provided hash before requesting a new token.
+            bool skipCache = _clientParameters.ForceRefresh || 
+                (!string.IsNullOrEmpty(AuthenticationRequestParameters.Claims) && 
+                string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh));
+
+            // Skip checking cache when either ForceRefresh is true
+            // or (Claims are present without a token hash).
```
rayluo, line 67: **suggestion (non-blocking)**
I suppose we no longer need these 2 lines here, because we already have clearer inline comments above for the `skipCache`.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -189,16 +206,83 @@ private async Task<AuthenticationResult> SendTokenRequestToAppTokenProviderAsync
...
             MsalAccessTokenCacheItem cachedAccessTokenItem = await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
 
-            if (cachedAccessTokenItem != null && !_clientParameters.ForceRefresh)
+            // 2) If no cached item or force refresh is requested, return null
+            if (!IsValidCachedToken(cachedAccessTokenItem))
             {
-                AuthenticationRequestParameters.RequestContext.ApiEvent.IsAccessTokenCacheHit = true;
-                Metrics.IncrementTotalAccessTokensFromCache();
-                return cachedAccessTokenItem;
+                return null;
```
rayluo, line 215: **suggestion**
I was about to ask the following question, "Wouldn't this change make it always bypassing cache when `ForceRefresh` is true, regardless of whether `AccessTokenHashToRefresh` exists or not".

It took me a while to realize that those two parameters won't coexist. But still this `ForceRefresh` check here seems unnecessary and misleading.

How about we simplify this to only check `if(!cachedAccessTokenItem) {return null;}`?

Another way - and perhaps a better way - to think about it is that the new hash check is not supposed to increase the Cyclomatic Complexity by more than 1. We could simply add an `if(isHashMatching(...)) {return null;}` right before the spot that would originally return a cached token, and then it shall just work.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: Agree, just that there is code in between and this stands out a little more

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: in this scenario where we check forceRefresh is not tied to the Sha256 API, we do need to check, for forceRefresh as a standalone scenario, that's what this does

```cs
private bool IsValidCachedToken(MsalAccessTokenCacheItem cachedAccessTokenItem)
{
    // Return false if:
    // - The cache is empty
    // - ForceRefresh is enabled
    if (cachedAccessTokenItem == null || _clientParameters.ForceRefresh)
    {
        return false;
    }

    return true;
}
```

gladjohn: we do have unit tests, that specifically tests forceRefresh alone. And all tests are passing. WithSha25 API, force refresh cannot be used and that is also tested

gladjohn: fixed it @rayluo

kennyoye88:
Stop sending me email again please
Regards,

Oyewale Kehinde Peter
08099740389

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: removed it

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -189,16 +204,51 @@ private async Task<AuthenticationResult> SendTokenRequestToAppTokenProviderAsync
...
+            {
+                return false;
+            }
+
+            // 2) If the token’s hash matches AccessTokenHashToRefresh, ignore it
+            if (!string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh) &&
+                IsMatchingTokenHash(cacheItem.Secret, _clientParameters.AccessTokenHashToRefresh))
+            {
+                AuthenticationRequestParameters.RequestContext.Logger.Info(
+                    "[ClientCredentialRequest] A cached token was found and its hash matches AccessTokenHashToRefresh, so it is ignored.");
```
rayluo, line 235: Given that we now seem to lean towards only implementing this in MSAL .Net, shall we add that [telemetry "TokenRejected"](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5179#discussion_r1987572831) to MSAL .Net, too, preferably in this PR?

rayluo (approved):
Functionality wise, this PR shall work. Adding a couple of minor comments below, but overall LGTM.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -2150,5 +2153,229 @@ public void ConfidentialClient_WithInvalidAuthority_ThrowsArgumentException()
...
+
+                // Assert
+                // We expect to reuse the cached token
+                Assert.AreEqual("access-token", result.AccessToken,
+                    "Should reuse cached token if the hash does not match the 'bad' token's hash.");
+                Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+            }
+        }
+
+        public async Task WithAccessTokenSha256ToRefresh_NullOrEmpty_ThrowsArgumentNullException_Async()
```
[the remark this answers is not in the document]
gladjohn: good catch @rayluo. I removed the [TestMethod] tag, now removed the whole test, as it is no longer needed.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: Sure @rayluo

src/client/Microsoft.Identity.Client/Extensibility/RP/AcquireTokenForClientParameterBuilderForResourceProviders.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <c>Convert.ToBase64String(SHA256(Encoding.UTF8.GetBytes(accessToken)))</c>.
+        /// </param>
+        /// <returns>The builder to chain the .With methods.</returns>
+        public static AcquireTokenForClientParameterBuilder WithAccessTokenSha256ToRefresh(
+            this AcquireTokenForClientParameterBuilder builder,
+            string hash)
+        {
+            if (!string.IsNullOrWhiteSpace(hash))
+            {
+                builder.Parameters.AccessTokenHashToRefresh = hash;
```
Robbie-Microsoft, line 32: Can we display a warning to the dev here and let them know that they didn't enter a hash?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -44,11 +51,19 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
                 logger.Error(MsalErrorMessage.ClientCredentialWrongAuthority);
             }
-
             AuthenticationResult authResult;
 
-            // Skip checking cache when force refresh or claims are specified
-            if (_clientParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
+            // Skip cache if either:
+            // 1) ForceRefresh is set, or
+            // 2) Claims are specified and there is no AccessTokenHashToRefresh.
```
Robbie-Microsoft, line 58: Shouldn't you check here to see if forceRefresh and AccessTokenHashToRefresh are both provided and throw an error if so?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -189,16 +204,51 @@ private async Task<AuthenticationResult> SendTokenRequestToAppTokenProviderAsync
...
+                await CacheManager.FindAccessTokenAsync().ConfigureAwait(false);
 
-            if (cachedAccessTokenItem != null && !_clientParameters.ForceRefresh)
+            // If the item fails any checks (null, or hash mismatch),
+            if (!ShouldUseCachedToken(cacheItem))
             {
-                AuthenticationRequestParameters.RequestContext.ApiEvent.IsAccessTokenCacheHit = true;
-                Metrics.IncrementTotalAccessTokensFromCache();
-                return cachedAccessTokenItem;
+                return null;
```
Robbie-Microsoft, line 214: Should we add a warning to let the user know that the hash they provided was not matched to an existing hash?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -189,16 +204,51 @@ private async Task<AuthenticationResult> SendTokenRequestToAppTokenProviderAsync
...
-            return null;
+            // Otherwise, record a successful cache hit and return the token.
+            MarkAccessTokenAsCacheHit();
+            return cacheItem;
+        }
+
+        private bool ShouldUseCachedToken(MsalAccessTokenCacheItem cacheItem)
+        {
+            // 1) No cached item 
+            if (cacheItem == null)
```
Robbie-Microsoft, line 225: nit: return cacheItem || false

Robbie-Microsoft: Are you able to do this in .net?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -189,16 +204,51 @@ private async Task<AuthenticationResult> SendTokenRequestToAppTokenProviderAsync
 
         private async Task<MsalAccessTokenCacheItem> GetCachedAccessTokenAsync()
```
Robbie-Microsoft, line 205: nit: can you add `/// <summary>` docs for this function?

Robbie-Microsoft: If you choose to implement this feedback, can you also do it for the other new functions you've created in this file?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
Robbie-Microsoft: I see you're checking ^ in Validate inside of AcquireTokenForClientParameterBuilder. Does that check happen before or after this code?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -2150,5 +2153,206 @@ public void ConfidentialClient_WithInvalidAuthority_ThrowsArgumentException()
...
+                    .Create(TestConstants.ClientId)
+                    .WithClientSecret(TestConstants.ClientSecret)
+                    .WithHttpManager(httpManager)
+                    .WithExperimentalFeatures(true)
+                    .BuildConcrete();
+
+                httpManager.AddMockHandlerSuccessfulClientCredentialTokenResponseMessage(token: "old-token");
+                AuthenticationResult firstResult = await app.AcquireTokenForClient(TestConstants.s_scope).ExecuteAsync().ConfigureAwait(false);
+
+                Assert.AreEqual("old-token", firstResult.AccessToken);
```
Robbie-Microsoft, line 2281: nit: Should you define "old-token" constant since it's used more than once? Same feedback to all other constants you're using.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
Robbie-Microsoft: If after, should it happen here instead?

gladjohn: before

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: no we cannot, because `cacheItem` is of type `MsalAccessTokenCacheItem`, which cannot be directly coalesced with a bool. we can surely do `return cacheItem != null;`

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: yes there is one at `AcquireTokenForClientParameters`

src/client/Microsoft.Identity.Client/Extensibility/RP/AcquireTokenForClientParameterBuilderForResourceProviders.cs
gladjohn: yes there is one at `AcquireTokenForClientParameters`

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
rayluo: Actually, we are [accepting null or empty string as a valid input](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5179#discussion_r1989085160). @gladjohn 

Besides, an input hash not matching any existing token is considered normal (especially in Service Fabric's scenario); that is what we designed this WithAccessTokenSha256ToRefresh() for: to silently serve a cached token to a request with mismatching hash. @Robbie-Microsoft

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -29,7 +29,7 @@ namespace Microsoft.Identity.Client
     public sealed class AcquireTokenForClientParameterBuilder :
         AbstractConfidentialClientAcquireTokenParameterBuilder<AcquireTokenForClientParameterBuilder>
     {
-        private AcquireTokenForClientParameters Parameters { get; } = new AcquireTokenForClientParameters();
+        internal AcquireTokenForClientParameters Parameters { get; } = new AcquireTokenForClientParameters();
```
neha-bhargava, line 32: Is this needed?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -187,20 +202,78 @@ private async Task<AuthenticationResult> SendTokenRequestToAppTokenProviderAsync
             return authResult;
         }
 
+        /// <summary>
+        /// Checks if the token should be used from the cache.
```
neha-bhargava, line 206:
```suggestion
        /// Checks if the token should be used from the cache and returns the cached access token if applicable.
```

bgavrilMS (approved):
Approved with comments. Let's discuss about telemetry separately.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: skipping for now

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
neha-bhargava: Can you point me to where? @gladjohn

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
neha-bhargava: This change is not applied and marked resolved. Can you share the reason why?

gladjohn: good find @neha-bhargava , applied now.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
gladjohn: @neha-bhargava The `Parameters` property needs to be `internal` because it is accessed in the `AcquireTokenForClientParameterBuilderForResourceProviders` class to set the `AccessTokenHashToRefresh` property. Here is the relevant code snippet:

```cs
public static AcquireTokenForClientParameterBuilder WithAccessTokenSha256ToRefresh( this AcquireTokenForClientParameterBuilder builder, string hash) { if (!string.IsNullOrWhiteSpace(hash)) { builder.Parameters.AccessTokenHashToRefresh = hash; }
```

Making the `Parameters` property `internal` allows it to be accessed within the same assembly, which is necessary for this extension method to function correctly.

gladjohn: does this explanation resolve your questions @neha-bhargava, I see you closed it without accepting the reasoning. Is there a better way to do this?

gladjohn:
> Approved with comments. Let's discuss about telemetry separately.

All comments addressed and removed experimental flags check from the tests, as we are GA'ing the API.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -44,11 +51,19 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            // we do NOT skip the cache, allowing MSAL to attempt retrieving a matching
+            // cached token by the provided hash before requesting a new token.
+            bool skipCache = _clientParameters.ForceRefresh || 
+                (!string.IsNullOrEmpty(AuthenticationRequestParameters.Claims) && 
+                string.IsNullOrEmpty(_clientParameters.AccessTokenHashToRefresh));
+
+            if (skipCache)
             {
                 AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.ForceRefreshOrClaims;
                 logger.Info("[ClientCredentialRequest] Skipped looking for a cached access token because ForceRefresh or Claims were set.");
```
neha-bhargava, line 69:
```suggestion
                logger.Info("[ClientCredentialRequest] Skipped looking for a cached access token because either of ForceRefresh, Claims or AccessTokenHashToRefresh were set.");
```

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
gladjohn: Committed.

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
rayluo: Nice catch. Actually, the new change was still inaccurate. The accurate statement shall be "skipped ... cache ... because either ForceRefresh was set, or Claims was set without AccessTokenHashToRefresh".

rayluo: Being addressed in #5201

PR #5182, opened 2025-03-10
For Mtls, throw exception when service config region is set to null.

### LINKED ISSUE #5181: [Bug] `WithMtlsProofOfPossession()' if the region in the config is null.
### Library version used

Latest

### .NET version

NET 8

### Scenario

Currently, the WithMtlsProofOfPossession() validation logic ensures that requests with `ServiceBundle.Config.AzureRegion` with an MsalClientException (MsalError.MtlsPopWithoutRegion) for null or empty but not for whitespace.

### Relevant code snippets

```csharp
// Check for Azure region only if the authority is AAD
                if (ServiceBundle.Config.Authority.AuthorityInfo.AuthorityType == AuthorityType.Aad &&
                    string.IsNullOrEmpty(ServiceBundle.Config.AzureRegion))
                {
                        MsalError.MtlsPopWithoutRegion,
                        MsalErrorMessage.MtlsPopWithoutRegion);
                }
```

### Expected behavior

- Requests with `ServiceBundle.Config.AzureRegion` set to whitespace should fail with an MsalClientException (MsalError.MtlsPopWithoutRegion).
- Add unit tests

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

ksaaf:
This PR fixes throwing exception when service config region is set to null for Mtls.

Fixes #5181 

- Replaced `string.IsNullOrEmpty` with `null` in `AcquireTokenForClientParameterBuilder.cs` for Azure region check when the authority is AAD.

**Testing**
- Updates existing unit tests to cover another case for this case when service config region is set to null and fixes existing tests.

**Performance impact**

N/A

**Documentation**

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
@@ -130,28 +130,40 @@ namespace Microsoft.Identity.Test.Unit
         }
 
         [DataTestMethod]
-        [DataRow(null, DisplayName = "Region is null")]
-        [DataRow("", DisplayName = "Region is empty string")]
-        [DataRow("   ", DisplayName = "Region is whitespace")]
-        public async Task MtlsPopWithoutRegionAsync(string region)
+        [DataRow(false)]
+        [DataRow(true)]
+        public async Task MtlsPop_WithoutRegion_ThrowsException(bool setAzureRegion)
         {
             using (var envContext = new EnvVariableContext())
             {
-                Environment.SetEnvironmentVariable("REGION_NAME", region); // Ensure no region is set
+                IConfidentialClientApplication app;
+                if (setAzureRegion)
+                {
+                    app = ConfidentialClientApplicationBuilder
+                                    .Create(TestConstants.ClientId)
+                                    .WithAuthority(TestConstants.AuthorityTenant)
+                                    .WithCertificate(s_testCertificate)
+                                    // Setting Azure region to ConfidentialClientApplicationBuilder.DisableForceRegion overrides the AzureRegion to null.
+                                    .WithAzureRegion(ConfidentialClientApplicationBuilder.DisableForceRegion)
+                                    .WithExperimentalFeatures()
+                                    .Build();
+                }
+                else
+                {
+                    app = ConfidentialClientApplicationBuilder
+                                    .Create(TestConstants.ClientId)
+                                    .WithAuthority(TestConstants.AuthorityTenant)
+                                    .WithCertificate(s_testCertificate)
+                                    .WithExperimentalFeatures()
+                                    .Build();
+                }
 
-                IConfidentialClientApplication app = ConfidentialClientApplicationBuilder
-                                .Create(TestConstants.ClientId)
-                                .WithAuthority(TestConstants.AuthorityTenant)
-                                .WithCertificate(s_testCertificate)
-                                .WithExperimentalFeatures()
-                                .Build();
-
-                // Set WithMtlsProofOfPossession on the request without specifying a region
+                // Set WithMtlsProofOfPossession on the request
                 MsalClientException ex = await AssertException.TaskThrowsAsync<MsalClientException>(() =>
-                    app.AcquireTokenForClient(TestConstants.s_scope)
-                       .WithMtlsProofOfPossession() // Enables MTLS PoP
-                       .ExecuteAsync())
-                    .ConfigureAwait(false);
+                        app.AcquireTokenForClient(TestConstants.s_scope)
+                           .WithMtlsProofOfPossession() // Enables MTLS PoP
+                           .ExecuteAsync())
+                        .ConfigureAwait(false);
 
                 Assert.AreEqual(MsalError.MtlsPopWithoutRegion, ex.ErrorCode);
                 Assert.AreEqual(MsalErrorMessage.MtlsPopWithoutRegion, ex.Message);
--- a/tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
@@ -129,12 +129,15 @@ namespace Microsoft.Identity.Test.Unit
             Assert.AreEqual(MsalError.MtlsCertificateNotProvided, ex.ErrorCode);
         }
 
-        [TestMethod]
-        public async Task MtlsPopWithoutRegionAsync()
+        [DataTestMethod]
+        [DataRow(null, DisplayName = "Region is null")]
+        [DataRow("", DisplayName = "Region is empty string")]
+        [DataRow("   ", DisplayName = "Region is whitespace")]
+        public async Task MtlsPopWithoutRegionAsync(string region)
         {
             using (var envContext = new EnvVariableContext())
             {
-                Environment.SetEnvironmentVariable("REGION_NAME", null); // Ensure no region is set
+                Environment.SetEnvironmentVariable("REGION_NAME", region); // Ensure no region is set
 
                 IConfidentialClientApplication app = ConfidentialClientApplicationBuilder
                                 .Create(TestConstants.ClientId)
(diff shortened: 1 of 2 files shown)


sruke (approved):

gladjohn:
not sure if this is the correct fix. The issue is not in MTLS code path in the the Azure Region API code path. After your fix trying running the test `MsalForceRegionIsSet_WithRegionIsSet_WithRegionWins` with " "

MZOLN (approved):

gladjohn (changes requested):
may not fix all region cases

ksaaf:
> not sure if this is the correct fix. The issue is not in MTLS code path in the the Azure Region API code path. After your fix trying running the test `MsalForceRegionIsSet_WithRegionIsSet_WithRegionWins` with " "

@gladjohn - I ran the tests as original and with your suggestion under debugger putting a breakpoint on the line I fixed. It didn't even hit the breakpoint.

Do you have a different test in mind? Can you run tests against this PR branch and check? Thanks.

ksaaf:
> may not fix all region cases

I am not sure if we I am following you. The fix is simple and adds test cases for string.empty scenario as well as for the new whitespace. And the test passes in all three cases. What am I missing here?

gladjohn:
> I ran the tests as original and with your suggestion under debugger putting a breakpoint on the line I fixed. It didn't even hit the breakpoint.

And what is the outcome of the test?

gladjohn:
> do you have a change in mind as you requested a change here?

The correct fix should happen [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d0a3fc04980174a61c2ec6434dd28dc626674b92/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L294)

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
```diff
@@ -129,12 +129,15 @@ public async Task MtlsPopWithoutCertificateWithClientAssertionAsync()
...
+        [DataTestMethod]
+        [DataRow(null, DisplayName = "Region is null")]
+        [DataRow("", DisplayName = "Region is empty string")]
+        [DataRow("   ", DisplayName = "Region is whitespace")]
+        public async Task MtlsPopWithoutRegionAsync(string region)
         {
             using (var envContext = new EnvVariableContext())
             {
-                Environment.SetEnvironmentVariable("REGION_NAME", null); // Ensure no region is set
+                Environment.SetEnvironmentVariable("REGION_NAME", region); // Ensure no region is set
```
bgavrilMS, line 140: @gladjohn - does adding `WithMtlsProofOfPossession` add a `WithRegion(TRY_AUTO_DISCOVER)` ? I think the regional opt-in should remain explicit via `WithRegion("bla")` or `WithRegion()`

Or in other words, why do you set this env variable here?

bgavrilMS (changes requested):
I believe the test MtlsPopWithoutRegionAsync is misleading. It should not set an env variable. Regional opt-in is done via API.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
gladjohn: that is correct the regional opt-in should be via the API and not tweaking the env variables

ksaaf:
> MsalForceRegionIsSet_WithRegionIsSet_WithRegionWins

It passed.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
ksaaf: @gladjohn, @bgavrilMS 

Thanks for the discussion and code pointers. 

Just to make sure we are all aligned, I created this PR to fix the issue #5181; problematic code [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d0a3fc04980174a61c2ec6434dd28dc626674b92/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs#L191C21-L191C75).

For this issue, there is `MtlsPopWithoutRegionAsync` test which uses `WithMtlsProofOfPossession` which exercises the above code. As you called out setting environment variable has no bearing here in this test. And it turns out `ServiceBundle.Config.AzureRegion` in the above referenced code path will be null even without setting env variable. The above referenced code path can also be exercised by calling `WithAzureRegion(ConfidentialClientApplicationBuilder.DisableForceRegion)` while building application. I will update the test and the code path with a comment by modifying the check from string.isNullOrEmpty to null.


> The correct fix should happen [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d0a3fc04980174a61c2ec6434dd28dc626674b92/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L294)

I think this one is a separate issue. In addition to the above, the fix also needs to be applied in this [function](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d0a3fc04980174a61c2ec6434dd28dc626674b92/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L369). I will open a separate issue to address this.

ksaaf:
> I believe the test MtlsPopWithoutRegionAsync is misleading. It should not set an env variable. Regional opt-in is done via API.

ksaaf:
> > do you have a change in mind as you requested a change here?
> 
> The correct fix should happen [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d0a3fc04980174a61c2ec6434dd28dc626674b92/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L294)

This is a different fix. Please see my response [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5182#discussion_r1989850437).

tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
ksaaf: Opened #5185 to track the separate issue.

gladjohn:
> > > do you have a change in mind as you requested a change here?
> > 
> > 
> > The correct fix should happen [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d0a3fc04980174a61c2ec6434dd28dc626674b92/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L294)
> 
> This is a different issue. Please see my response [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5182#discussion_r1989850437).

@ksaaf the actual issue is not in the AcquireTokenForClient API or flows, The issue you are describing stems from WithAzureRegion API. I am happy to discuss this with you and go over the actual problem. 

As @bgavrilMS, mentioned there could be other bugs that are present in the Region flow that has been exposed by this test. So this needs a little more investigation.

ksaaf:
> > > > do you have a change in mind as you requested a change here?
> > > 
> > > 
> > > The correct fix should happen [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d0a3fc04980174a61c2ec6434dd28dc626674b92/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L294)
> > 
> > 
> > This is a different issue. Please see my response [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5182#discussion_r1989850437).
> 
> @ksaaf the actual issue is not in the AcquireTokenForClient API or flows, The issue you are describing stems from WithAzureRegion API. I am happy to discuss this with you and go over the actual problem.
> 
> As @bgavrilMS, mentioned there could be other bugs that are present in the Region flow that has been exposed by this test. So this needs a little more investigation.

Yes, let's discuss; happy to learn the context as I am new to the codebase.

bgavrilMS (approved):

gladjohn (commented):
looks good.

PR #5188, opened 2025-03-13
Split CacheRefreshReason.ForceRefreshOrClaims into ForceRefresh and WithClaims

### LINKED ISSUE #5186: [Feature Request] Split CacheRefreshReason.ForceRefreshOrClaims
### MSAL client type

Confidential

### Problem statement

[Feature Request] Split CacheRefreshReason.ForceRefreshOrClaims 

### Proposed solution

Split CacheRefreshReason.ForceRefreshOrClaims 

### Alternatives

_No response_

gladjohn:
Fixes #5186

This pull request includes several changes to the `CacheRefreshReason` enum and updates to the related logic and tests across multiple files. The changes focus on refining the reasons for cache refresh and improving the logging and handling of these reasons.

### Enum and Logic Updates:

* [`src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs`](diffhunk://#diff-dd548256ab53963ad65e75170deba2ff00da77c5ca6192415ff6b38f9048d93eL16-R18): Updated `CacheRefreshReason` enum by removing `ForceRefreshOrClaims` and adding `WithClaims` and `TokenRejected` reasons. [[1]](diffhunk://#diff-dd548256ab53963ad65e75170deba2ff00da77c5ca6192415ff6b38f9048d93eL16-R18) [[2]](diffhunk://#diff-dd548256ab53963ad65e75170deba2ff00da77c5ca6192415ff6b38f9048d93eL30-R38)
* [`src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs`](diffhunk://#diff-655d5868568553085143b1c4df713cb7baf07c35ab78610a8ac73da1e467ce43L50-R58): Updated cache refresh reason logic to distinguish between `ForceRefresh` and `WithClaims`.
* [`src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs`](diffhunk://#diff-a1260a45bdc4ac7ab60d4a348996be6b69c5a48fa6f484fa41daa390c7e69ca8L36-R43): Similar update to cache refresh reason logic.
* [`src/client/Microsoft.Identity.Client/Internal/Requests/OnBehalfOfRequest.cs`](diffhunk://#diff-bf467b9439792137a4965d3a9b52423cf088346687d966681557d134a6ef1b10L106-R109): Updated to log specific cache refresh reasons.
* [`src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs`](diffhunk://#diff-04f82f33c9cf802268c7c48a869ca80b26839db606a0d729bd739887c7593bfaL69-R72): Updated to log specific cache refresh reasons.

### Public API Updates:

* [`src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Shipped.txt`](diffhunk://#diff-cbf7ceab1866a87ca05ef66c7a043bce3f21b454700e64c182b403f3884ad195L440): Removed `ForceRefreshOrClaims` from the shipped API.
* [`src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt`](diffhunk://#diff-d8d9399ed6a37cdd3bf25cd223b32bfe33318862404ac5bba32f34e213fb6eb5R1-R3): Added `ForceRefresh`, `WithClaims`, and `TokenRejected` to the unshipped API.
* Similar updates were made to other `PublicAPI.Shipped.txt` and `PublicAPI.Unshipped.txt` files for different frameworks. [[1]](diffhunk://#diff-cbf7ceab1866a87ca05ef66c7a043bce3f21b454700e64c182b403f3884ad195L440) [[2]](diffhunk://#diff-ba02fd692f23584ac367416f93f025aafe3ee9535e714b7ad1f46486a18c81c7L443) [[3]](diffhunk://#diff-521cf91ecf46f87e497e10f087b1a619fa9cdf96c4698ae336c7dd47324092e1L445) [[4]](diffhunk://#diff-ffdc781f3ef8f95fb6c862704d1835c18d30398d487f8bba40c46091311be10aL437) [[5]](diffhunk://#diff-9f88cae51dd278939a6c4dd17a5f89c65e7bd3560ab630ed505b869500463cd5L437)

### Test Updates:

* [`tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/HttpTelemetryRecorder.cs`](diffhunk://#diff-e560c22d3e206c40340cfa15a35dc7c161384a5820ec978a148eb01c062ad807R26-R37): Updated to handle new cache refresh reasons.
* [`tests/Microsoft.Identity.Test.Unit/RequestsTests/LongRunningOnBehalfOfTests.cs`](diffhunk://#diff-1cd803db17cc6016844d5ee7bc809e58ac9b8695236bac89405edda7d5b8bf1aL334-R334): Added a new test for `WithClaims` and updated existing tests to reflect the new enum values. [[1]](diffhunk://#diff-1cd803db17cc6016844d5ee7bc809e58ac9b8695236bac89405edda7d5b8bf1aL334-R334) [[2]](diffhunk://#diff-1cd803db17cc6016844d5ee7bc809e58ac9b8695236bac89405edda7d5b8bf1aR931-R972)
* [`tests/Microsoft.Identity.Test.Unit/TelemetryTests/HttpTelemetryTests.cs`](diffhunk://#diff-606c6cb88f2f7be52cfb5bdeee91051feba0dc6135aca2786dd869867b174004L105-R105): Updated telemetry test to use `ForceRefresh` instead of `ForceRefreshOrClaims`.

These changes collectively enhance the granularity and clarity of cache refresh reasons, improving both logging and functionality.

**Testing**
unit testing

**Performance impact**
none

**Documentation**
- [ ] All relevant documentation is updated. Need to check documentation,

### DIFF
--- a/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
@@ -13,9 +13,9 @@ namespace Microsoft.Identity.Client
         /// </summary>
         NotApplicable = 0,
         /// <summary>
-        /// When the token request goes to the identity provider because force_refresh was set to true. Also occurs if WithClaims() is used.
+        /// When the token request goes to the identity provider because force_refresh was set to true.
         /// </summary>
-        ForceRefreshOrClaims = 1,
+        ForceRefresh = 1,
         /// <summary>
         /// When the token request goes to the identity provider because no cached access token exists
         /// </summary>
@@ -27,6 +27,14 @@ namespace Microsoft.Identity.Client
         /// <summary>
         /// When the token request goes to the identity provider because refresh_in was used and the existing token needs to be refreshed
         /// </summary>
-        ProactivelyRefreshed = 4
+        ProactivelyRefreshed = 4,
+        /// <summary>
+        /// When the token request goes to the identity provider because WithClaims() was used.
+        /// </summary>
+        WithClaims = 5,
+        /// <summary>
+        /// When the token request goes to the identity provider because WithAccessTokenSha256ToRefresh() is used AND the hash matches.
+        /// </summary>
+        TokenRejected = 6
     }
 }
--- a/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+
 namespace Microsoft.Identity.Client
 {
     /// <summary>
@@ -13,28 +15,33 @@ namespace Microsoft.Identity.Client
         /// </summary>
         NotApplicable = 0,
         /// <summary>
+        /// When the token request goes to the identity provider because force_refresh was set to true. Also occurs if WithClaims() is used.
+        /// </summary>
+        [Obsolete("Use ForceRefresh or WithClaims instead.")]
+        ForceRefreshOrClaims = 1,
+        /// <summary>
         /// When the token request goes to the identity provider because force_refresh was set to true.
         /// </summary>
-        ForceRefresh = 1,
+        ForceRefresh = 2,
         /// <summary>
         /// When the token request goes to the identity provider because no cached access token exists
         /// </summary>
-        NoCachedAccessToken = 2,
+        NoCachedAccessToken = 3,
         /// <summary>
         /// When the token request goes to the identity provider because cached access token expired
         /// </summary>
-        Expired = 3,
+        Expired = 4,
         /// <summary>
         /// When the token request goes to the identity provider because refresh_in was used and the existing token needs to be refreshed
         /// </summary>
-        ProactivelyRefreshed = 4,
+        ProactivelyRefreshed = 5,
         /// <summary>
         /// When the token request goes to the identity provider because WithClaims() was used.
         /// </summary>
-        WithClaims = 5,
+        WithClaims = 6,
         /// <summary>
-        /// When the token request goes to the identity provider because WithAccessTokenSha256ToRefresh() is used AND the hash matches.
+        /// Indicates that the resource (e.g., Microsoft Graph) has rejected the provided token. 
         /// </summary>
-        TokenRejected = 6
+        TokenRejectedByResource = 7
     }
 }
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
             AuthenticationResult authResult;
 
-            // Determine the cache refresh reason and log
+            // Determine if there is a cache refresh reason and log
             if (_clientParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
             {
                 AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo =
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
@@ -47,11 +47,15 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
             AuthenticationResult authResult;
 
-            // Skip checking cache when force refresh or claims are specified
+            // Determine the cache refresh reason and log
             if (_clientParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
             {
-                AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.ForceRefreshOrClaims;
-                logger.Info("[ClientCredentialRequest] Skipped looking for a cached access token because ForceRefresh or Claims were set.");
+                AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo =
+                    _clientParameters.ForceRefresh ? CacheRefreshReason.ForceRefresh : CacheRefreshReason.WithClaims;
+
+                logger.Info($"[ClientCredentialRequest] Skipped looking for a cached access token because " +
+                            $"{(_clientParameters.ForceRefresh ? "ForceRefresh" : "Claims")} was set.");
+
                 authResult = await GetAccessTokenAsync(cancellationToken, logger).ConfigureAwait(false);
                 return authResult;
             }
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Identity.Client.Internal.Requests
             AuthenticationResult authResult = null;
             ILoggerAdapter logger = AuthenticationRequestParameters.RequestContext.Logger;
 
-            // Determine the cache refresh reason and log
+            // Determine the cache refresh reason, skip the cache if force refresh or claims is used and log
             if (_managedIdentityParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
             {
                 AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo =
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
@@ -33,13 +33,14 @@ namespace Microsoft.Identity.Client.Internal.Requests
             AuthenticationResult authResult = null;
             ILoggerAdapter logger = AuthenticationRequestParameters.RequestContext.Logger;
 
-            // Skip checking cache when force refresh or claims is specified
+            // Determine the cache refresh reason and log
             if (_managedIdentityParameters.ForceRefresh || !string.IsNullOrEmpty(AuthenticationRequestParameters.Claims))
             {
-                AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo = CacheRefreshReason.ForceRefreshOrClaims;
-                
-                logger.Info("[ManagedIdentityRequest] Skipped looking for a cached access token because ForceRefresh or Claims were set. " +
-                    "This means either a force refresh was requested or claims were present.");
+                AuthenticationRequestParameters.RequestContext.ApiEvent.CacheInfo =
+                    _managedIdentityParameters.ForceRefresh ? CacheRefreshReason.ForceRefresh : CacheRefreshReason.WithClaims;
+
+                logger.Info($"[ManagedIdentityRequest] Skipped looking for a cached access token because " +
+                            $"{(_managedIdentityParameters.ForceRefresh ? "ForceRefresh" : "Claims")} was set.");
 
                 authResult = await GetAccessTokenAsync(cancellationToken, logger).ConfigureAwait(false);
                 return authResult;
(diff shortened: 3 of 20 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -33,13 +33,14 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             AuthenticationResult authResult = null;
             ILoggerAdapter logger = AuthenticationRequestParameters.RequestContext.Logger;
 
-            // Skip checking cache when force refresh or claims is specified
+            // Determine the cache refresh reason and log
```
neha-bhargava, line 36:
```suggestion
            // Determine the cache refresh reason, skip the cache if force refresh or claims is used and log
```

tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/HttpTelemetryRecorder.cs
```diff
@@ -23,9 +23,18 @@ public void SplitCurrentCsv(string telemetryCsv)
         {
             string[] splitCsv = telemetryCsv.Split('|');
             string[] splitApiIdAndCacheInfo = splitCsv[1].Split(',');
+
             ApiId.Add(splitApiIdAndCacheInfo[0]);
-            Enum.TryParse(splitApiIdAndCacheInfo[1], out CacheRefreshReason cacheInfoTelemetry);
-            ForceRefresh = CacheRefreshReason.ForceRefreshOrClaims == cacheInfoTelemetry;
```
neha-bhargava, line 28: You can just replace this line with ```ForceRefresh = cacheInfoTelemetry == CacheRefreshReason.ForceRefresh ||
                               cacheInfoTelemetry == CacheRefreshReason.WithClaims;```

neha-bhargava:
The changes looks good to me. This change will result in change in all the queries that we have that looks at the cache refresh reason. Both the server side and client side. @bgavrilMS do you know if anyone else consumes this data? If they do then this might be breaking change for them. But we should update all the queries in the dashboards and the oneNote once this change goes in.

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
```diff
@@ -27,6 +27,14 @@ public enum CacheRefreshReason
...
-        ProactivelyRefreshed = 4
+        ProactivelyRefreshed = 4,
+        /// <summary>
+        /// When the token request goes to the identity provider because WithClaims() was used.
+        /// </summary>
+        WithClaims = 5,
+        /// <summary>
+        /// When the token request goes to the identity provider because WithAccessTokenSha256ToRefresh() is used AND the hash matches.
+        /// </summary>
+        TokenRejected = 6
```
trwalke, line 38: What does it mean that we are rejecting the token? Who is rejecting it? The comment doesn't seem to fully explain it.

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
```diff
@@ -13,9 +13,9 @@ public enum CacheRefreshReason
         /// </summary>
         NotApplicable = 0,
         /// <summary>
-        /// When the token request goes to the identity provider because force_refresh was set to true. Also occurs if WithClaims() is used.
+        /// When the token request goes to the identity provider because force_refresh was set to true.
         /// </summary>
-        ForceRefreshOrClaims = 1,
```
rayluo, line 18: **question**
Is removing this value a breaking change? If so, we may keep both `ForceRefreshOrClaims = 1` and `ForceRefresh = 1` ([Csharp allows that](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/enums#194-enum-members), right?), and use `<summary>...</summary>` to describe that `ForceRefreshOrClaims` remains only for backward compatibility purposes.

Existing telemetry and dashboards will technically still work as-is; it is just that they won't see the new reasons until they are updated.

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
bgavrilMS: I agree with @trwalke, the naming is misleading. Maybe use `ForceRefreshSha256Token`

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -47,11 +47,15 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
             AuthenticationResult authResult;
 
-            // Skip checking cache when force refresh or claims are specified
+            // Determine the cache refresh reason and log
```
trwalke, line 50:
```suggestion
            // Determine if there is a cache refresh reason and log
```

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
bgavrilMS: Maybe leave value 1 as "not longer used" ? And use new values for everything else?

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/RequestsTests/LongRunningOnBehalfOfTests.cs
```diff
@@ -928,6 +928,48 @@ await app.InitiateLongRunningProcessInWebApi(TestConstants.s_scope, TestConstant
...
+                // STEP 1: Initiate LRO OBO with claims
+                var initResult = await cca
+                    .InitiateLongRunningProcessInWebApi(TestConstants.s_scope, TestConstants.DefaultAccessToken, ref oboCacheKey)
+                    .WithClaims(TestConstants.Claims)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+
+                // Because we used claims, MSAL sets CacheRefreshReason.WithClaims, skipping the cached token
+                Assert.AreEqual(TokenSource.IdentityProvider, initResult.AuthenticationResultMetadata.TokenSource);
+                Assert.AreEqual(CacheRefreshReason.NotApplicable, initResult.AuthenticationResultMetadata.CacheRefreshReason);
```
trwalke, line 955: the comment above says MSAL sets the reason to `CacheRefreshReason.WithClaims`, however, this is checking for CacheRefreshReason.NotApplicable?

trwalke (approved):
Approve contingent on naming being updated.

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
rayluo: `ForceRefreshSha256Token` seems too specific because the "sha256" part is really just an implementation detail; it is likely that in the future we may have another parameter `WithAccessTokenToRefresh(string entireToken)`. We shouldn't need to invent yet another enum at that time.

On the other hand, given that we seem comfortable using longer than TwoWords convention, we can use longer phrase to answer the "rejected by who". For example, `TokenRejectedByCaller` (which is actually quite precise).

Just my $0.02

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
gladjohn: @neha-bhargava ^^ can we do that ?

tests/Microsoft.Identity.Test.Unit/RequestsTests/LongRunningOnBehalfOfTests.cs
gladjohn: for the obo flow yes, the first call is not applicable.

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
gladjohn: @rayluo can you please confirm the final name to be used?

gladjohn: Added as 

```cs
/// <summary>
/// Indicates that the resource (e.g., Microsoft Graph) has rejected the provided token. 
/// </summary>
TokenRejectedByResource = 7
```

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
gladjohn: updated per Bogdan's suggestion

gladjohn:
> Approve contingent on naming being updated.

made it `TokenRejectedByResource` @trwalke @bgavrilMS @rayluo

bgavrilMS:
> The changes looks good to me. This change will result in change in all the queries that we have that looks at the cache refresh reason. Both the server side and client side. @bgavrilMS do you know if anyone else consumes this data? If they do then this might be breaking change for them. But we should update all the queries in the dashboards and the oneNote once this change goes in.

I do not. Let's postpone this change and discuss it separately?

bgavrilMS (changes requested):
Need consensus on telemetry.

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
neha-bhargava: This does not seem right, we cannot change all the values. This will break all the server side queries. If you are adding new values, they get assigned new numbers in the enum.

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
```diff
@@ -15,18 +17,31 @@ public enum CacheRefreshReason
...
-        ProactivelyRefreshed = 4
+        ProactivelyRefreshed = 5,
+        /// <summary>
+        /// When the token request goes to the identity provider because WithClaims() was used.
+        /// </summary>
+        WithClaims = 6,
+        /// <summary>
+        /// Indicates that the resource (e.g., Microsoft Graph) has rejected the provided token. 
+        /// </summary>
+        TokenRejectedByResource = 7
```
rayluo, line 45: > rayluo can you please confirm the final name to be used?

**suggestion (non-blocking)**
From logical standpoint, I would still prefer to use `TokenRejectedByCaller`, because that is really it: the MSAL caller calls the WithAccessTokenSha256ToRefresh(...) for *some* reason; we *hope* it was caused by resource rejecting it, but we won't know for sure. (One real-world example is that [another team at one point attempted proactively refreshing token more aggressively](https://msdata.visualstudio.com/DefaultCollection/Database%20Systems/_git/DsMainDev/pullRequest/1533955#1734345632); even though we subsequently persuaded them not to, we can't rule out someone else would do it via WithAccessTokenSha256ToRefresh() in the future for whatever reason.)

P.S.: Now that the `ForceRefreshOrClaims` split is on-hold, but perhaps we can at least add a `TokenRejectedByCaller = 7 // Or 6`, to unblock the WithAccessTokenSha256ToRefresh() PR.

gladjohn:
> > The changes looks good to me. This change will result in change in all the queries that we have that looks at the cache refresh reason. Both the server side and client side. @bgavrilMS do you know if anyone else consumes this data? If they do then this might be breaking change for them. But we should update all the queries in the dashboards and the oneNote once this change goes in.
> 
> I do not. Let's postpone this change and discuss it separately?

let me come up with a mini spec for this and we can discuss different options of doing this. 

FYI @neha-bhargava @bgavrilMS @rayluo

src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs
```diff
@@ -15,18 +17,31 @@ public enum CacheRefreshReason
...
         ForceRefreshOrClaims = 1,
         /// <summary>
+        /// When the token request goes to the identity provider because force_refresh was set to true.
+        /// </summary>
+        ForceRefresh = 2,
+        /// <summary>
         /// When the token request goes to the identity provider because no cached access token exists
         /// </summary>
-        NoCachedAccessToken = 2,
+        NoCachedAccessToken = 3,
```
rayluo, line 29: @neha-bhargava is [correct](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5188#discussion_r1995897832) that this change does not look right. The original topic of that conversation was to keep an alias on the SDK implementation WITHOUT altering any old values. We shall never alter old values.

Please take this into consideration in the upcoming mini specs.

gladjohn:
Closing this as we plan to revisit the design.

PR #5193, opened 2025-03-16
Fix Machine Learning Source to Use "clientid" Instead of "client_id"

### LINKED ISSUE #5183: [Bug] Managed Identity: In the Machine Learning source, the client id is sent incorrectly in the network request
### Library version used

latest

### .NET version

N/A

### Scenario

None

The Machine Learning source is based off of App Service 2017 (MSALs do not support this App Service 2017 version, only 2019). In the 2017 API, the client id should be passed to the network request as "clientid" instead of "client_id".

This bug was identified recently after Azure-SDK-for-JS removed their App Service 2017 implementation in favor of MSALJS's (msal-node) Machine Learning implementation. An Azure-SDK-for-JS customer is seeing network errors due to the incorrect way client id is being passed to the network request.

### Relevant code snippets

```csharp

```

### Expected behavior

_No response_

### Identity provider

Other

### Regression

_No response_

### Solution and workarounds

_No response_

gladjohn:
Fixes #5183

This pull request includes several changes to the Microsoft Identity Client library to support different managed identity configurations and improve test coverage. The most important changes include adding a new constant for the 2017 managed identity client ID, updating the request creation logic for older machine learning environments, and adding a new unit test to verify the default token type.

### Support for different managed identity configurations:

* [`src/client/Microsoft.Identity.Client/Internal/Constants.cs`](diffhunk://#diff-69a69c54931b0896394c0c0ba76f1642cd6931ce882c112675804793bea7a5daR45): Added a new constant `ManagedIdentityClientId2017` for the 2017 managed identity client ID.
* [`src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs`](diffhunk://#diff-a8294b48db4b9d9920dea44df8f04f7e90d5704aa0e02fd6f60467f5f6ce714fL78-R79): Updated the `CreateRequest` method to use the new `ManagedIdentityClientId2017` constant for older ML-based environments.
* [`tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs`](diffhunk://#diff-b5cbc5bfefa33d39f979adca0369b61dfc99c7100434c6785ce15cab5a1e9bd7R381-R391): Modified the `AddManagedIdentityMockHandler` method to handle both the 2017 and 2019 managed identity client ID parameters based on the source type.

### Improved test coverage:

* [`tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationResultTests.cs`](diffhunk://#diff-a38cae874977b1de652fa165a0c0e7952919eb3eb099fa888c5a2508499a2645R193-R218): Added a new test method `DefaultTokenType_IsBearer_Test` to verify that the default token type is 'Bearer' and that `CreateAuthorizationHeader` uses it correctly.

**Testing**
Unit tests

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Constants.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Constants.cs
@@ -42,6 +42,7 @@ namespace Microsoft.Identity.Client.Internal
         public const string SshCertAuthHeaderPrefix = "SshCert";
 
         public const string ManagedIdentityClientId = "client_id";
+        public const string ManagedIdentityClientId2017 = "clientid";
         public const string ManagedIdentityObjectId = "object_id";
         public const string ManagedIdentityResourceId = "mi_res_id";
         public const string ManagedIdentityDefaultClientId = "system_assigned_managed_identity";
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
@@ -75,7 +75,8 @@ namespace Microsoft.Identity.Client.ManagedIdentity
             {
                 case AppConfig.ManagedIdentityIdType.ClientId:
                     _requestContext.Logger.Info("[Managed Identity] Adding user assigned client id to the request.");
-                    request.QueryParameters[Constants.ManagedIdentityClientId] = _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
+                    // Use the new 2017 constant for older ML-based environment
+                    request.QueryParameters[Constants.ManagedIdentityClientId2017] = _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
                     break;
 
                 case AppConfig.ManagedIdentityIdType.ResourceId:
--- a/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
+++ b/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
@@ -378,7 +378,16 @@ namespace Microsoft.Identity.Test.Common.Core.Mocks
 
             if (userAssignedIdentityId == UserAssignedIdentityId.ClientId)
             {
-                httpMessageHandler.ExpectedQueryParams.Add(Constants.ManagedIdentityClientId, userAssignedId);
+                if (managedIdentitySourceType == ManagedIdentitySource.MachineLearning)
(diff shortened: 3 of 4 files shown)


rayluo (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
```diff
@@ -75,7 +75,8 @@ protected override ManagedIdentityRequest CreateRequest(string resource)
             {
                 case AppConfig.ManagedIdentityIdType.ClientId:
                     _requestContext.Logger.Info("[Managed Identity] Adding user assigned client id to the request.");
-                    request.QueryParameters[Constants.ManagedIdentityClientId] = _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
+                    // Use the new 2017 constant for older ML-based environment
+                    request.QueryParameters[Constants.ManagedIdentityClientId2017] = _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
```
bgavrilMS, line 79: Didn't we add ML support because someone complained about this? Won't we be breaking them?

gladjohn: We added ML support but while doing so broke the source detection order. No one AFAIK used App Service 2017, but if we are adding ML 2017 as a source then it has to be `clientid`

src/client/Microsoft.Identity.Client/Internal/Constants.cs
```diff
@@ -42,6 +42,7 @@ internal static class Constants
         public const string SshCertAuthHeaderPrefix = "SshCert";
 
         public const string ManagedIdentityClientId = "client_id";
```
Robbie-Microsoft, line 44: I think you should change this to "ManagedIdentityClientId2019".

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Internal/Constants.cs
rayluo: I think the key here is to NOT think "api 2017 always uses clientid", "api 2019 always client_id", "managed identity Foo uses api 2019 and managed identity Bar also uses api 2019 so they are likely the same". No. Each managed identity provider *models* their API from a similar template but their implementations are effectively different, possibly not sharing the same code base.

That's what I meant the other day by saying "ideally MSAL shall implement each managed identity protocol in their own code path so that they don't share the same implementation". @Robbie-Microsoft

PR #5208, opened 2025-03-24
Revoguard-hack-2025

gladjohn:
Revoguard-hack-2024


docs/revoguard-hack-2024
```diff
@@ -0,0 +1,97 @@
+# Token Revocation Hackathon: RevoGuard
```
bgavrilMS, line 1: Do you want to put this in an "experimental" or "prototype" folder instead?

gladjohn: created a `prototype` and moved it there.

prototype/revoguard-hack-2025.md
```diff
@@ -0,0 +1,97 @@
+# Token Revocation Hackathon: RevoGuard
```
bgavrilMS, line 1: Pls add this file along with the PR with the dev app.

PR #5210, opened 2025-03-27
use HTTP 2.0

### LINKED ISSUE #5209: [Feature Request] Use HTTP 2.0 protocol
### MSAL client type

Confidential

### Problem statement

HttpManager internally creates `HttpRequestMessage` which by default uses `HTTP 1.1`
Ideally should use `HTTP 2.0`
Even though my factory makes `HTTP 2.0` default it won't work

### Proposed solution

```
        private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string, string> headers)
        {
            HttpRequestMessage requestMessage = new HttpRequestMessage { RequestUri = endpoint };
            requestMessage.Version = HttpVersion.Version20;
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            requestMessage.Headers.Accept.Clear();
            if (headers != null)
            {
                foreach (KeyValuePair<string, string> kvp in headers)
                {
                    requestMessage.Headers.Add(kvp.Key, kvp.Value);
                }
            }

            return requestMessage;
        }
```

### Alternatives

Keep using `HTTP 1.1`

sherlock1982:
Fixes #5209

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -186,6 +186,7 @@ namespace Microsoft.Identity.Client.Http
         private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string, string> headers)
         {
             HttpRequestMessage requestMessage = new HttpRequestMessage { RequestUri = endpoint };
+            requestMessage.Version = HttpVersion.Version20;
             requestMessage.Headers.Accept.Clear();
             if (headers != null)
             {


bgavrilMS:
Hi @sherlock1982 - there is a bit more to it because I discovered that some internal endpoints don't yet support HTTP/2.0 

Here is a PR to fix this issue - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5314, which also includes tests

As such, I'm going to close yours.

sherlock1982:
It's ok. But just to let you know setting `HTTP 2.0` doesn't mean `2.0 only`. It means `2.0 or less if not supported`.
If we can do it different way, optional way - of course that would be better

PR #5224, opened 2025-04-04
Fix for 5223 - env var to disable ESTS-R

### LINKED ISSUE #5223: [Feature Request] Env variable to disable ESTS-R
### MSAL client type

Confidential

### Problem statement

Some Azure system build out happens before ESTS-R and teams have to do a 2 pass buildout. 1st  with regional endpoint disabled and then laster when ESTS-R becomes available with regional endpoint enabled. 
 
MSAL has a flag to do this per instance but there are multiple wrappers that mask it. Ex PDP plugin in MISE wraps it and has its own different variable to set to use global vs regional ests. 
 
Instead of chasing these different configs having a global configuration / env variable would make it easy for services to enable or disable the setting.

### Proposed Solution

If env variable `MSAL_DISABLE_REGION` is set (to any value), then all ESTS-R logic will be ignored. 
This option overrides all other env variables or configuration.

bgavrilMS:
Fixes #5223

### DIFF
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Identity.Client
     public class ConfidentialClientApplicationBuilder : AbstractApplicationBuilder<ConfidentialClientApplicationBuilder>
     {
         internal const string ForceRegionEnvVariable = "MSAL_FORCE_REGION";
+        internal const string DisableRegionEnvVariable = "MSAL_DISABLE_REGION";
         internal const string DisableForceRegion = "DisableMsalForceRegion";
 
         /// <inheritdoc/>
@@ -380,10 +381,16 @@ namespace Microsoft.Identity.Client
                 throw new InvalidOperationException(MsalErrorMessage.InvalidRedirectUriReceived(Config.RedirectUri));
             }
 
-            if (!string.IsNullOrEmpty(Config.AzureRegion) && 
-                (Config.CustomInstanceDiscoveryMetadata != null || Config.CustomInstanceDiscoveryMetadataUri != null))
+            ValidateAndUpdateRegion();
+        }
+
+        private void ValidateAndUpdateRegion()
+        {
+            // master override - do not use region if this env variable is set, as per #5223
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(DisableRegionEnvVariable)))
             {
-                throw new MsalClientException(MsalError.RegionDiscoveryWithCustomInstanceMetadata, MsalErrorMessage.RegionDiscoveryWithCustomInstanceMetadata);
+                Config.AzureRegion = null;
+                return;
             }
 
             // use regional if MSAL_FORCE_REGION is used, as per #4930
@@ -398,6 +405,12 @@ namespace Microsoft.Identity.Client
                 {
                     Config.AzureRegion = forcedRegion;
                 }
+            }
+
+            if (!string.IsNullOrEmpty(Config.AzureRegion) &&
+                (Config.CustomInstanceDiscoveryMetadata != null || Config.CustomInstanceDiscoveryMetadataUri != null))
+            {
+                throw new MsalClientException(MsalError.RegionDiscoveryWithCustomInstanceMetadata, MsalErrorMessage.RegionDiscoveryWithCustomInstanceMetadata);
             }
         }
 
--- a/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
@@ -395,6 +395,12 @@ namespace Microsoft.Identity.Client
                 return;
             }
 
+            if (!string.IsNullOrEmpty(Config.AzureRegion) &&
+              (Config.CustomInstanceDiscoveryMetadata != null || Config.CustomInstanceDiscoveryMetadataUri != null))
+            {
+                throw new MsalClientException(MsalError.RegionDiscoveryWithCustomInstanceMetadata, MsalErrorMessage.RegionDiscoveryWithCustomInstanceMetadata);
+            }
+
             // use regional if MSAL_FORCE_REGION is used, as per #4930
             if (string.Equals(Config.AzureRegion, DisableForceRegion, StringComparison.OrdinalIgnoreCase))
             {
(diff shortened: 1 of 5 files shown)


neha-bhargava (dismissed):
LGTM. Can you share why is this override needed?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
```diff
@@ -329,6 +329,90 @@ public async Task MsalForceRegionIsSet_RegionIsUsed()
...
+                    ConfidentialClientApplicationBuilder.ForceRegionEnvVariable, "someregion");
+                Environment.SetEnvironmentVariable(
+                    ConfidentialClientApplicationBuilder.DisableRegionEnvVariable, "1");
+
+                var httpManager = harness.HttpManager;
+
+                var cca = ConfidentialClientApplicationBuilder
+                                .Create(TestConstants.ClientId)
+                                .WithHttpManager(httpManager)
+                                //.WithAzureRegion(TestConstants.Region)
```
trwalke, line 370: Dead code?

trwalke (approved):

bgavrilMS:
> LGTM. Can you share why is this override needed?

Yes, I updated the issue with more details. THe gist of it is that MSAL is encapsulated by many other SDKs (Azure Identity, MISE) and it's hard to control ESTS-R programatically. There is some cases where you need to disable it and then re-enable it. THis env var will sort it out.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -399,6 +406,12 @@ internal override void Validate()
                     Config.AzureRegion = forcedRegion;
                 }
             }
+
```
Robbie-Microsoft, line 411: Why did you move this check after the if/else instead of keeping it before? Do you need to add `return`s to the if/else if?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
bgavrilMS: yes, resolved, thx.

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
```diff
@@ -380,10 +381,16 @@ internal override void Validate()
...
 
-            if (!string.IsNullOrEmpty(Config.AzureRegion) && 
-                (Config.CustomInstanceDiscoveryMetadata != null || Config.CustomInstanceDiscoveryMetadataUri != null))
+            ValidateAndUpdateRegion();
+        }
+
+        private void ValidateAndUpdateRegion()
+        {
+            // master override - do not use region if this env variable is set, as per #5223
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(DisableRegionEnvVariable)))
```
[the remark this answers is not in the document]
bgavrilMS: Yes, resolved, thx.

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ClientCredentialsTests.NetFwk.cs
```diff
@@ -288,7 +288,7 @@ private async Task RunClientCredsAsync(Cloud cloud, CredentialType credentialTyp
             var appCacheRecorder = confidentialApp.AppTokenCache.RecordAccess();
             Guid correlationId = Guid.NewGuid();
             authResult = await confidentialApp
-                .AcquireTokenForClient(settings.AppScopes)
+                .AcquireTokenForClient(settings.AppScopes)
```
Robbie-Microsoft, line 291: delete this new whitespace that you added?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
```diff
@@ -329,6 +329,89 @@ public async Task MsalForceRegionIsSet_RegionIsUsed()
...
         }
 
+        [TestMethod]
+        public async Task MsalDisableRegionEnvVariable_API()
+        {
+            using (new EnvVariableContext())
+            using (var harness = base.CreateTestHarness())
+            {
+                Environment.SetEnvironmentVariable(
+                    ConfidentialClientApplicationBuilder.DisableRegionEnvVariable, "1");
```
Robbie-Microsoft, line 339: "1" is a magic string. Can you define it as a const? Is it used in other tests and already defined as a const?

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
```diff
@@ -329,6 +329,89 @@ public async Task MsalForceRegionIsSet_RegionIsUsed()
...
+        }
+
+        [TestMethod]
+        public async Task MsalDisableRegionEnvVariable_OtherEnv()
+        {
+            using (new EnvVariableContext())
+            using (var harness = base.CreateTestHarness())
+            {
+                Environment.SetEnvironmentVariable(
+                    ConfidentialClientApplicationBuilder.ForceRegionEnvVariable, "someregion");
```
Robbie-Microsoft, line 361: Do we have a const defined already for test regions?

src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs
bgavrilMS: Fair question, I'll revert.

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ClientCredentialsTests.NetFwk.cs
bgavrilMS: removed.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
bgavrilMS: It's not a magic string, the env variable can be set to anything.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/ClientCredentialWithRegionTests.cs
bgavrilMS: I find the tests to be more readable without constants. I don't see any value in constants in test code.

Robbie-Microsoft (approved):

trwalke (approved):

PR #5231, opened 2025-04-11
Reworked retry policy functionality & Created IMDS retry policy

Robbie-Microsoft:
.NET's retry policy functionality now aligns with MSAL-Node. This PR was already implemented in MSAL-Node, [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/7614).

There are now two retry **strategies**:
LinearRetryStrategy and ExponentialRetryStrategy

There are now two retry **policies**, which use the retry **strategies**:
ImdsRetryPolicy and DefaultRetryPolicy (both ESTS and non-IMDS Managed Identity HttpManagers use DefaultRetryPolicy)

The unit tests in this PR are ported from the MSAL-Node PR:
All retry policy unit tests run "100 times faster" by overriding certain variables; Retry policy tests are no longer "expensive".


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -115,7 +114,7 @@ public async Task<HttpResponse> SendRequestAsync(
                     bindingCertificate,
                     customHttpClient,
                     cancellationToken: cancellationToken,
-                    retryCount) // Pass the updated retry count
+                    retryCount++) // increment the retry count
```
neha-bhargava, line 117: nit: 
```suggestion
                    retryCount++) // increment and pass the retry count
```

src/client/Microsoft.Identity.Client/Http/NoRetryPolicy.cs
```diff
@@ -2,20 +2,16 @@
...
 {
     internal class NoRetryPolicy : IRetryPolicy
     {
-        public int DelayInMilliseconds { get => throw new NotImplementedException(); }
-
-        public bool pauseForRetry(HttpResponse response, Exception exception, int retryCount)
+        public Task<bool> pauseForRetryAsync(HttpResponse response, Exception exception, int retryCount, ILoggerAdapter logger)
         {
-            return false;
+            throw new NotImplementedException();
```
neha-bhargava, line 14: Revert this bacl to return false. This will be used in case the user is passing own http factory and want to have their own retry logic

bgavrilMS: return `Task.FromResult(false)` because you are returning a task (a promise in JS terms)

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.Http
+{
+    class DefaultRetryPolicy : IRetryPolicy
```
bgavrilMS, line 10: nit: `Default` is a pretty overloaded term. Let's use `StsEndpointsRetryPolicy` maybe?

Robbie-Microsoft: I like Default, because this is used by both ESTS **AND** Managed Identity sources without any specific policy

src/client/Microsoft.Identity.Client/Http/IRetryPolicy.cs
```diff
@@ -2,16 +2,13 @@
...
 using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
 
 namespace Microsoft.Identity.Client.Http
 {
     internal interface IRetryPolicy
     {
-        int DelayInMilliseconds { get; }
-        bool pauseForRetry(HttpResponse response, Exception exception, int retryCount);
+        Task<bool> pauseForRetryAsync(HttpResponse response, Exception exception, int retryCount, ILoggerAdapter logger);
```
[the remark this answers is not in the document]
Robbie-Microsoft: Is this standard in .NET?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: I feel that "pass" is already pretty self explanatory since it's used as a parameter here

src/client/Microsoft.Identity.Client/Http/NoRetryPolicy.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/IRetryPolicy.cs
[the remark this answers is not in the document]
Robbie-Microsoft: done. I went back through the whole PR and applied this feedback too. I found this very useful: [Capitalization Conventions](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: I added this since this is the only parameter that changes between the retry requests.

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
neha-bhargava: I don't think this will be used by managed identity sources at all. Its either the ones specific to managed identity or NoRetryPolicy.

Robbie-Microsoft: @neha-bhargava, that's incorrect as I understand it. All Managed Identities default to the DefaultRetryPolicy unless otherwise specified. This is how it works in msal-node.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -102,9 +102,9 @@ public async Task<HttpResponse> SendRequestAsync(
                 timeoutException = exception;
             }
 
-            while (_retryPolicy.pauseForRetry(response, timeoutException, retryCount))
+            while (await _retryPolicy.PauseForRetryAsync(response, timeoutException, retryCount, logger).ConfigureAwait(false))
```
gladjohn, line 105: can we pass the caller’s CancellationToken into PauseForRetryAsync?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -102,9 +102,9 @@ public async Task<HttpResponse> SendRequestAsync(
...
             {
-                logger.Warning($"Retry condition met. Retry count: {retryCount++} after waiting {_retryPolicy.DelayInMilliseconds}ms.");
+                retryCount++;
```
gladjohn, line 107: we now increment once here? and the previous log line that printed the delay is gone? is it now logged inside `PauseForRetryAsync`

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: Why would we do that if it's not being done already?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: correct

Robbie-Microsoft: the delay is calculated inside of each retry policy

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: It's not used in the retry policy

bgavrilMS:
Is this ready for review now?

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        private LinearRetryStrategy linearRetryStrategy = new LinearRetryStrategy();
+
+        // constants that are defined in the constructor
+        public static int DEFAULT_RETRY_DELAY_MS { get; set; } // this will be overridden in the unit tests so that they run faster
```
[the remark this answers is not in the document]
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/ImdsRetryPolicy.cs
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+        private ExponentialRetryStrategy _exponentialRetryStrategy = new ExponentialRetryStrategy(
+            ImdsRetryPolicy.MIN_EXPONENTIAL_BACKOFF_MS,
+            ImdsRetryPolicy.MAX_EXPONENTIAL_BACKOFF_MS,
+            ImdsRetryPolicy.EXPONENTIAL_DELTA_BACKOFF_MS
+        );
+
+        public async Task<bool> PauseForRetryAsync(HttpResponse response, Exception exception, int retryCount, ILoggerAdapter logger)
```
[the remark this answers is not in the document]
bgavrilMS: Can you please detail what you have in mind and why? Concurrency tests are never simple, so it would be good to understand the scenario you think it should test. 

Another good question is if  this object is used in multi-threaded context? I believe a retry policy object is created for each HTTP call, so I am not sure if there is concurrency there.

src/client/Microsoft.Identity.Client/Http/ExponentialRetryStrategy.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Identity.Client.Http
+{
+    internal class ExponentialRetryStrategy
```
bgavrilMS, line 8: Do we really need these "strategy" objects? It seems to me that you can just put all the inputs in the `CalculateDelay` method and make that static and part of the Policy object.

src/client/Microsoft.Identity.Client/Http/LinearRetryStrategy.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Identity.Client.Http
+{
+    internal class LinearRetryStrategy
```
bgavrilMS, line 8: An object with no state is a code smell. It takes up space on the stack. See my previous comment

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,85 @@
...
+        // overridden in the unit tests so that they run faster
+        public static int DefaultStsRetryDelayMs { get; set; } = 1000;
+        public static int DefaultManagedIdentityRetryDelayMs { get; set; } = 1000;
+
+        // used for comparison, in the unit tests
+        // will be reset after every test
+        public static int NumRetries { get; set; } = 0;
+
+        public static int DefaultRetryDelayMs;
+        private int MaxRetries;
```
bgavrilMS, line 33: `_maxRetries`

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
bgavrilMS: @neha-bhargava - I thought all MSI sources use a linear 3 time retry, with 1 second pause? While the STS uses 1 retrty.

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,85 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.Http
```
bgavrilMS, line 8: Is it worth moving all these Retry clasess to their own "Retry" folder?

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
bgavrilMS: @Robbie-Microsoft - I want to challenge this `default` concept again. Relying on a "default" behavior will lead MSAL developers to make mistakes whenever we call new endpoints. We'll likely overuse "default". It's also complicated, as the "default" has a different shape between MSI and STS. But I realize I have looked too much at this and my opinion is tainted.

@neha-bhargava @gladjohn - can you please provide a perspective on this one?

neha-bhargava: @bgavrilMS you are right. All MSI does 3 retry with a linear delay of 1 sec and ests 1 retry with 1 sec delay.

gladjohn: @bgavrilMS @neha-bhargava agree the only difference I've know is the 1 second pause between them. But I think @Robbie-Microsoft did this way because there isn't a one second retry in the MSI flows. I filed a task to add it.

gladjohn: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5260

src/client/Microsoft.Identity.Client/Http/ImdsRetryPolicy.cs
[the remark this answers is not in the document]
gladjohn: mainly to test how the 1 second pause is in effect between sources and IDPs. But I do not see a one second pause, maybe we should change the method name.

src/client/Microsoft.Identity.Client/Http/ExponentialRetryStrategy.cs
Robbie-Microsoft: I made them their own objects so that they could be re-used across retry policies. I think it makes sense to keep them their own objects since a retry policy like IMDS's uses both exponential and linear strategy.

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
Robbie-Microsoft: Should the "Retry" folder be inside of the "Http" folder?

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/LinearRetryStrategy.cs
Robbie-Microsoft: resolved offline during feature presentation to team

src/client/Microsoft.Identity.Client/Http/DefaultRetryPolicy.cs
Robbie-Microsoft: done: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5300

src/client/Microsoft.Identity.Client/Http/Retry/IRetryPolicy.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.Http.Retry
+{
+    internal interface IRetryPolicy
```
neha-bhargava, line 12: Add a comment what this interface is for?

src/client/Microsoft.Identity.Client/Http/Retry/ImdsRetryPolicy.cs
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Net;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.Http.Retry
+{
+    internal class ImdsRetryPolicy : IRetryPolicy
```
neha-bhargava, line 12: Add a comment stating the retry policy specific to IMDS and the link to IMDS docs if they are public

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -81,9 +78,12 @@ public async Task<string> GetAzureRegionAsync(RequestContext requestContext)
                 requestContext.ApiEvent != null,
                 "Do not call GetAzureRegionAsync outside of a request. This can happen if you perform instance discovery outside a request, for example as part of validating input params.");
 
+            IRetryPolicyFactory retryPolicyFactory = requestContext.ServiceBundle.Config.RetryPolicyFactory;
+            IRetryPolicy retryPolicy = retryPolicyFactory.GetRetryPolicy(RequestType.STS);
```
neha-bhargava, line 81: @bgavrilMS should we use the IMDS retry policy here with less number of retries? Maybe in another iteration?

tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/MsiProxyHttpManager.cs
```diff
@@ -3,19 +3,16 @@
...
 using System.Net.Security;
 using System.Security.Cryptography.X509Certificates;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Web;
-using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Http.Retry;
```
neha-bhargava, line 15: Not needed, remove.

tests/Microsoft.Identity.Test.Unit/Helpers/ParallelRequestMockHandler.cs
```diff
@@ -11,6 +11,7 @@
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Http.Retry;
```
neha-bhargava, line 14: Unused?

src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
```diff
@@ -21,12 +21,32 @@ public static bool ManagedIdentity(HttpResponse response, Exception exception)
...
+        {
+            if (exception != null)
+            {
+                return exception is TaskCanceledException ? true : false;
+            }
+
+            return (int)response.StatusCode switch
+            {
+                // Not Found, Request Timeout, Gone, Too Many Requests
+                404 or 408 or 410 or 429 => true,
```
gladjohn, line 43: per docs [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/docs/imds_retry_based_on_errors.md), for 410 -> Retry every 10 seconds (max 70s / 7 attempts).

gladjohn: nm, I see you have set it in `LinearStrategyNumRetries`

tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/MsiProxyHttpManager.cs
Robbie-Microsoft: It is needed, there's a reference to `IRetryPolicy` on line 50

tests/Microsoft.Identity.Test.Unit/Helpers/ParallelRequestMockHandler.cs
Robbie-Microsoft: reference to IRetryPolicy on line 44

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
bgavrilMS: Makes sense, but it should be a separate issue / PR

src/client/Microsoft.Identity.Client/Http/Retry/ImdsRetryPolicy.cs
gladjohn: there is a public doc, but we have a slightly modified version based on the discussion with IMDS team. maybe a link to our spec?

src/client/Microsoft.Identity.Client/Http/Retry/LinearRetryStrategy.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    internal class LinearRetryStrategy
+    {
+        /// <summary>
+        /// Calculates the number of milliseconds to sleep based on the `Retry-After` HTTP header.
+        /// </summary>
+        /// <param name="retryHeader">The value of the `Retry-After` HTTP header. This can be either a number of seconds or an HTTP date string.</param>
+        /// <param name="minimumDelay">The minimum delay in milliseconds to return if the header is not present or invalid.</param>
+        /// <returns>The number of milliseconds to sleep before retrying the request.</returns>
+        public int calculateDelay(string retryHeader, int minimumDelay)
```
gladjohn:
```suggestion
        public int CalculateDelay(string retryHeader, int minimumDelay)
```

src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
```diff
@@ -4,15 +4,15 @@
...
 using System.Threading.Tasks;
 
-namespace Microsoft.Identity.Client.Http
+namespace Microsoft.Identity.Client.Http.Retry
 {
     internal static class HttpRetryConditions
     {
         /// <summary>
         /// Retry policy specific to managed identity flow.
         /// Avoid changing this, as it's breaking change.
```
gladjohn:
```suggestion
        /// Avoid changing this, as it's a breaking change.
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityRequest.cs
```diff
@@ -4,18 +4,13 @@
...
+using static Microsoft.Identity.Client.Http.Retry.IRetryPolicyFactory;
 
 namespace Microsoft.Identity.Client.ManagedIdentity
 {
     internal class ManagedIdentityRequest
     {
-        // referenced in unit tests, cannot be private
-        public const int DefaultManagedIdentityMaxRetries = 3;
-        // this will be overridden in the unit tests so that they run faster
-        public static int DefaultManagedIdentityRetryDelayMs { get; set; } = 1000;
```
gladjohn, line 17: you need to rebase

tests/Microsoft.Identity.Test.Common/TestConstants.cs
```diff
@@ -202,7 +202,17 @@ public static HashSet<string> s_scope
         public const string Region = "centralus";
         public const string InvalidRegion = "invalidregion";
         public const int TimeoutInMs = 2000;
-        public const string ImdsUrl = "http://169.254.169.254/metadata/instance/compute/location";
+        public const string ImdsHost = "169.254.169.254";
+        public const string ImdsUrl = $"http://{ImdsHost}/metadata/instance/compute/location";
+
+        public const double ONE_HUNDRED_TIMES_FASTER = 0.01;
```
gladjohn:
```suggestion
        public const double OneHundredTimesFaster = 0.01;

```

gladjohn (approved):
Nice job modularizing the retry logic, please rebase before merging.

tests/Microsoft.Identity.Test.Common/TestConstants.cs
Robbie-Microsoft: oops, this should've been removed

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -208,5 +209,6 @@ public X509Certificate2 ClientCredentialCertificate
         public bool IsInstanceDiscoveryEnabled { get; internal set; } = true;
         #endregion
 
+        internal IRetryPolicyFactory RetryPolicyFactory { get; set; }
```
trwalke: nit: I would move this above the regions. Maybe line 127 so that the properties are grouped together better.

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
```diff
@@ -323,5 +331,16 @@ internal static string GetValueIfNotEmpty(string original, string value)
         {
             return string.IsNullOrWhiteSpace(value) ? original : value;
         }
+
+        /// <summary>
+        /// Internal only: Allows tests to inject a custom retry policy factory.
+        /// </summary>
+        /// <param name="factory">The retry policy factory to use.</param>
+        /// <returns>The builder for chaining.</returns>
+        internal T WithRetryPolicyFactory(IRetryPolicyFactory factory)
```
trwalke: nit: would move this to line 237 so that "With" methods are grouped together

src/client/Microsoft.Identity.Client/Http/Retry/IRetryPolicy.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        Task<bool> PauseForRetryAsync(HttpResponse response, Exception exception, int retryCount, ILoggerAdapter logger);
```
trwalke, line 23: A comment for the method would also be nice

src/client/Microsoft.Identity.Client/Http/Retry/IRetryPolicyFactory.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client.Http.Retry
+{
+    internal interface IRetryPolicyFactory
+    {
+        public enum RequestType
```
trwalke: So this is defining an enum in an interface? This isn't something that we normally do in the code base. Typically, we define enums in their own file and then reference the enum as a property in the interface. Take a look at TokenSource.cs for example

src/client/Microsoft.Identity.Client/Http/Retry/ImdsRetryPolicy.cs
Robbie-Microsoft: I think the fact that it's named "ImdsRetryPolicy" is specific enough

Robbie-Microsoft: I'll link to our internal spec

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
Robbie-Microsoft: The IMDS retry policy can't have a lesser number of retries; it would have to be a new retry policy. Perhaps it can extend the IMDS retry policy and override the number of retries. I made a GitHub issue for this: #5321

src/client/Microsoft.Identity.Client/Http/Retry/ImdsRetryPolicy.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/Retry/IRetryPolicy.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityRequest.cs
Robbie-Microsoft: Do you mean merge the latest from the main branch? If so, done.

tests/Microsoft.Identity.Test.Common/TestConstants.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/Retry/LinearRetryStrategy.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/Retry/IRetryPolicy.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/Retry/IRetryPolicyFactory.cs
Robbie-Microsoft: Thank you for the explanation. done.

PR #5238, opened 2025-04-18
Change the resource id param for IMDS

### LINKED ISSUE #4911: Correct IMDS resource ID query parameter
Specifying an identity by resource ID doesn't work on Azure Container Instances (and maybe some other platforms) because we specify the resource ID with query parameter "mi_res_id". ACI observes only "m**s**i_res_id", which we should prefer anyway because although IMDS observes both parameters, it [documents only msi_res_id](https://github.com/Azure/azure-rest-api-specs/blob/dba6ed1f03bda88ac6884c0a883246446cc72495/specification/imds/data-plane/Microsoft.InstanceMetadataService/stable/2018-10-01/imds.json#L233-L239).

neha-bhargava:
Fixes #4911 

Update code and mocks

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Constants.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Constants.cs
@@ -45,6 +45,7 @@ namespace Microsoft.Identity.Client.Internal
         public const string ManagedIdentityClientId2017 = "clientid";
         public const string ManagedIdentityObjectId = "object_id";
         public const string ManagedIdentityResourceId = "mi_res_id";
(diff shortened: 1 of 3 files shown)


gladjohn (approved):
suggest we test both work.

gladjohn:
> > suggest we test both work.
> 
> By both do you mean both param names?

Thanks for clarifying it offline @neha-bhargava. Guess we cannot test both, since MSAL adopts to use the new param, we are good.

gladjohn (approved):
looks good, no comments

rayluo (approved):

PR #5252, opened 2025-04-23
Retry policies are now pre request instead of per HttpManager.

Robbie-Microsoft:
The same retry policy was previously being used for all Managed Identity sources and requests.

The retry policy will now be based on the Managed Identity source, and therefore will be per-request.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -26,22 +26,41 @@ namespace Microsoft.Identity.Client.Http
     /// </remarks>
     internal class HttpManager : IHttpManager
     {
+        // referenced in unit tests, cannot be private
+        public const int DEFAULT_ESTS_MAX_RETRIES = 1;
+        // this will be overridden in the unit tests so that they run faster
+        public static int DEFAULT_ESTS_RETRY_DELAY_MS { get; set; } = 1000;
+
         protected readonly IMsalHttpClientFactory _httpClientFactory;
-        private readonly IRetryPolicy _retryPolicy;
+        private readonly bool _isManagedIdentity;
+        private readonly bool _withRetry;
         public long LastRequestDurationInMs { get; private set; }
 
         /// <summary>
-        /// A new instance of the HTTP manager with a retry *condition*. The retry policy hardcodes: 
-        /// - the number of retries (1)
-        /// - the delay between retries (1 second)
+        /// Initializes a new instance of the <see cref="HttpManager"/> class.
         /// </summary>
+        /// <param name="httpClientFactory">
+        /// An instance of <see cref="IMsalHttpClientFactory"/> used to create and manage <see cref="HttpClient"/> instances.
+        /// This factory ensures proper reuse of <see cref="HttpClient"/> to avoid socket exhaustion.
+        /// </param>
+        /// <param name="isManagedIdentity">
+        /// A boolean flag indicating whether the HTTP manager is being used in a managed identity context.
+        /// </param>
+        /// <param name="withRetry">
+        /// A boolean flag indicating whether the HTTP manager should enable retry logic for transient failures.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// Thrown when <paramref name="httpClientFactory"/> is null.
+        /// </exception>
         public HttpManager(
             IMsalHttpClientFactory httpClientFactory,
-            IRetryPolicy retryPolicy)
+            bool isManagedIdentity,
+            bool withRetry)
         {
             _httpClientFactory = httpClientFactory ??
                 throw new ArgumentNullException(nameof(httpClientFactory));
-            _retryPolicy = retryPolicy;
+            _isManagedIdentity = isManagedIdentity;
+            _withRetry = withRetry;
         }
 
         public async Task<HttpResponse> SendRequestAsync(
@@ -54,8 +73,20 @@ namespace Microsoft.Identity.Client.Http
             X509Certificate2 bindingCertificate,
             Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> validateServerCert,
             CancellationToken cancellationToken,
+            IRetryPolicy retryPolicy = null,
             int retryCount = 0)
         {
+            // Use the default STS retry policy if the request is not for managed identity
+            // and a non-default STS retry policy is not provided.
+            // Skip this if statement the dev indicated that they do not want retry logic.
+            if (!_isManagedIdentity && retryPolicy == null && _withRetry)
+            {
+                retryPolicy = new LinearRetryPolicy(
+                    DEFAULT_ESTS_RETRY_DELAY_MS,
+                    DEFAULT_ESTS_MAX_RETRIES,
+                    HttpRetryConditions.Sts);
+            }
+
             Exception timeoutException = null;
             HttpResponse response = null;
 
@@ -102,9 +133,9 @@ namespace Microsoft.Identity.Client.Http
                 timeoutException = exception;
             }
 
-            while (_retryPolicy.pauseForRetry(response, timeoutException, retryCount))
+            while (_withRetry && retryPolicy.pauseForRetry(response, timeoutException, retryCount))
             {
-                logger.Warning($"Retry condition met. Retry count: {retryCount++} after waiting {_retryPolicy.DelayInMilliseconds}ms.");
+                logger.Warning($"Retry condition met. Retry count: {retryCount++} after waiting {retryPolicy.DelayInMilliseconds}ms.");
                 return await SendRequestAsync(
                     endpoint,
                     headers,
@@ -113,8 +144,10 @@ namespace Microsoft.Identity.Client.Http
                     logger,
                     doNotThrow,
                     bindingCertificate,
-                    validateServerCert, cancellationToken: cancellationToken,
-                    retryCount: retryCount) // Pass the updated retry count
+                    validateServerCert,
+                    cancellationToken,
+                    retryPolicy,
+                    retryCount) // Pass the updated retry count
                     .ConfigureAwait(false);
             }
 
--- a/src/client/Microsoft.Identity.Client/Http/LinearRetryPolicy.cs
+++ b/src/client/Microsoft.Identity.Client/Http/LinearRetryPolicy.cs
@@ -9,7 +9,9 @@ namespace Microsoft.Identity.Client.Http
 {
     internal class LinearRetryPolicy : IRetryPolicy
     {
-        
+        // referenced in unit tests, cannot be private
+        public static int numRetries { get; private set; } = 0;
+
         private int _maxRetries;
         private readonly Func<HttpResponse, Exception, bool> _retryCondition;
         public int DelayInMilliseconds { private set; get; }
@@ -23,6 +25,9 @@ namespace Microsoft.Identity.Client.Http
 
         public bool pauseForRetry(HttpResponse response, Exception exception, int retryCount)
         {
+            // referenced in the unit tests
+            numRetries = retryCount + 1;
+
             return retryCount < _maxRetries && _retryCondition(response, exception);
         }
     }
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
@@ -82,6 +82,9 @@ namespace Microsoft.Identity.Client.ManagedIdentity
                     break;
             }
 
+            // uncomment in follow-up IMDS retry policy PR
+            // request.RetryPolicy = new ImdsRetryPolicy();
+
             return request;
         }
 
--- a/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
+++ b/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
@@ -116,6 +116,7 @@ namespace Microsoft.Identity.Test.Common.Core.Mocks
             X509Certificate2 mtlsCertificate,
             Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> validateServerCert,
             CancellationToken cancellationToken,
+            IRetryPolicy retryPolicy = null,
             int retryCount = 0)
         {
             return _httpManager.SendRequestAsync(
@@ -127,6 +128,7 @@ namespace Microsoft.Identity.Test.Common.Core.Mocks
                 doNotThrow,
                 mtlsCertificate,
                 validateServerCert, cancellationToken,
+                retryPolicy,
                 retryCount);
         }
     }
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.AppConfig;
+using Microsoft.Identity.Client.Http;
 using Microsoft.Identity.Client.Internal;
 using Microsoft.Identity.Client.ManagedIdentity;
 using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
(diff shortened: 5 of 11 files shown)


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -26,22 +26,41 @@ namespace Microsoft.Identity.Client.Http
...
+        /// <param name="withRetry">
+        /// A boolean flag indicating whether the HTTP manager should enable retry logic for transient failures.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// Thrown when <paramref name="httpClientFactory"/> is null.
+        /// </exception>
         public HttpManager(
             IMsalHttpClientFactory httpClientFactory,
-            IRetryPolicy retryPolicy)
+            bool isManagedIdentity,
```
bgavrilMS, line 57: This violates the S in SOLID principle, i.e. single responsibility. It's not HTTP Manager's job to know about managed identity.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -26,22 +26,41 @@ namespace Microsoft.Identity.Client.Http
...
+            bool withRetry)
```
bgavrilMS, line 58: rename to `isRetryEnabled`

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -54,8 +73,20 @@ public async Task<HttpResponse> SendRequestAsync(
             X509Certificate2 bindingCertificate,
             Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> validateServerCert,
             CancellationToken cancellationToken,
+            IRetryPolicy retryPolicy = null,
```
bgavrilMS, line 76: Let's please not use default parameters with this one. THis will force other MSAL developers to think about retries, instead of using the "default" which is no-retry.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -54,8 +73,20 @@ public async Task<HttpResponse> SendRequestAsync(
             X509Certificate2 bindingCertificate,
             Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> validateServerCert,
             CancellationToken cancellationToken,
```
bgavrilMS, line 75: nit: since you are changing this method signature anyway, I recommend moving CancellationToken to be the last parameter and to have a default value of `default`

src/client/Microsoft.Identity.Client/Http/LinearRetryPolicy.cs
```diff
@@ -23,6 +25,9 @@ public LinearRetryPolicy(int delayMilliseconds, int maxRetries, Func<HttpRespons
 
         public bool pauseForRetry(HttpResponse response, Exception exception, int retryCount)
```
bgavrilMS, line 26: nit: please rename to `PauseForRetry` as this violates naming convention.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -54,8 +73,20 @@ public async Task<HttpResponse> SendRequestAsync(
...
+            IRetryPolicy retryPolicy = null,
             int retryCount = 0)
         {
+            // Use the default STS retry policy if the request is not for managed identity
+            // and a non-default STS retry policy is not provided.
+            // Skip this if statement the dev indicated that they do not want retry logic.
+            if (!_isManagedIdentity && retryPolicy == null && _withRetry)
+            {
+                retryPolicy = new LinearRetryPolicy(
```
bgavrilMS, line 84: nit: Consider creating a RetryPolicyFactory that is responsible for creating all these objects. It can also encapsulate the `HttpRetryConditions` logic.

src/client/Microsoft.Identity.Client/Http/LinearRetryPolicy.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: Shouldn't the default retry by the LinearRetryPolicy? That's how it was already set up, for STS and MI requests.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: The alternative here is to go to all references for SendRequestAsync and manually pass in the default retry policy. I think it makes more sense to define it here, once, if no retryPolicy was provided.

Currently in .NET, all STS and MI requests have the default LinearRetryPolicy.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: The alternative here is to go to all references for SendRequestAsync and manually pass in the default retry policy. I think it makes more sense to define it here, once, if no retryPolicy was provided.

Currently in .NET, all STS and MI requests have the default LinearRetryPolicy.

bgavrilMS: There are 15 references for SendRequestAsync, it doesn't seem so bad?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
bgavrilMS: But we are changing the MI requests to use different policies right?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -117,6 +117,7 @@ public string ClientVersion
         public bool RetryOnServerErrors { get; set; } = true;
 
         public ManagedIdentityId ManagedIdentityId { get; internal set; }
+        public bool DisableInternalRetries { get; internal set; }
```
bgavrilMS, line 120: This is fine, the name is better. But then, should you not delete `RetryOnServerErrors ` ?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: Ok, I'll do it. I didn't know if you'd recommend it or not.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
neha-bhargava: We used to have default value for the cancellation token and then we removed it to be passed back. What is the reason to have a default value for cancellation token? @bgavrilMS

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -1301,5 +1302,109 @@ public async Task MixedUserAndSystemAssignedManagedIdentityTestAsync()
...
+                var miBuilder = ManagedIdentityApplicationBuilder.Create(ManagedIdentityId.SystemAssigned)
+                    .WithHttpManager(httpManager);
+
+                // Disable cache to avoid pollution
+                miBuilder.Config.AccessorOptions = null;
+
+                var mi = miBuilder.Build();
+
+                // Simulate permanent 500s (to trigger the maximum number of retries)
+                int NUM_500 = ManagedIdentityRequest.DEFAULT_MANAGED_IDENTITY_MAX_RETRIES + 1; // initial request + maximum number of retries (3)
```
neha-bhargava: Nit: rename to num500 since this is a local variable.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -1301,5 +1302,109 @@ public async Task MixedUserAndSystemAssignedManagedIdentityTestAsync()
...
+                        .ExecuteAsync().ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    msalException = ex as MsalServiceException;
+                }
+                Assert.IsNotNull(msalException);
+
+                // ensure that the second request was made and retried 3 times
+                // (numRetries would be x2 if retry policy was NOT per request)
```
neha-bhargava, line 1324: I don't think I understand the 2nd line of this comment

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -1301,5 +1302,109 @@ public async Task MixedUserAndSystemAssignedManagedIdentityTestAsync()
...
+                msalException = null;
+                try
+                {
+                    await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
+                        .ExecuteAsync().ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    msalException = ex as MsalServiceException;
+                }
```
neha-bhargava: This can be rewritten as:
```suggestion
                msalException = await Assert.ThrowsExceptionAsync<MsalServiceException>( async() =>
                    await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
                        .ExecuteAsync().ConfigureAwait(false));               
```

src/client/Microsoft.Identity.Client/Http/LinearRetryPolicy.cs
```diff
@@ -21,8 +23,11 @@ public LinearRetryPolicy(int delayMilliseconds, int maxRetries, Func<HttpRespons
             _retryCondition = retryCondition;
         }
 
-        public bool pauseForRetry(HttpResponse response, Exception exception, int retryCount)
+        public bool PauseForRetry(HttpResponse response, Exception exception, int retryCount)
         {
+            // referenced in the unit tests
+            numRetries = retryCount + 1;
```
gladjohn, line 29: where is the pause happening? or where is `DelayInMilliseconds` used during retries

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
Robbie-Microsoft: 2 x (initial request + 3 retries) = 8 network requests

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
Robbie-Microsoft: Is it better to do it the way you suggested? And why? I like the idea of a try/catch as that's what we do in the msal-node tests. Can you explain why I should change it?

src/client/Microsoft.Identity.Client/Http/LinearRetryPolicy.cs
Robbie-Microsoft: @gladjohn I don't believe I touch any of that functionality in this PR? Where was it being used and where did the pause happen previously?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/LinearRetryPolicy.cs
gladjohn: @Robbie-Microsoft I do not see any pauses anywhere in product code, searched both `Thread.Sleep` and `Task.Delay` and nothing in managed identity flows. maybe we never implemented it. I added a separate work item to track this - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5260

gladjohn: you can resolve this conversation

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
gladjohn: Using Assert.ThrowsExceptionAsync<MsalServiceException>() makes the intent of the test clear at a glance. A try/ catch is still perfectly valid, as long as you assert inside the catch block. There’s also a third option:

```cs
[TestMethod]
[ExpectedException(typeof(MsalServiceException))]
public async Task AcquireToken_ThrowsAsync()
{
    await mi.AcquireTokenForManagedIdentity(Resource)
            .ExecuteAsync();
}
```

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
Robbie-Microsoft: I've since refactored this as a `const`, therefore I believe I'm supposed to make this PascalCase.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
Robbie-Microsoft: done

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: I'll do that in another PR. This one is already big enough.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
bgavrilMS: Convention. But I see your point - force MSAL developers to think about cancellation token. Thx.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
```diff
@@ -46,6 +46,11 @@ public RegionInfo(string region, RegionAutodetectionSource regionSource, string
         private static bool s_failedAutoDiscovery = false;
         private static string s_regionDiscoveryDetails;
 
+        private readonly LinearRetryPolicy _linearRetryPolicy = new LinearRetryPolicy(
```
bgavrilMS, line 49: Instead of repeating all this ctor code everywhere, can you have a static method like `LinearRetryPolicy.CreateStsRetry()` . This way you can hide those constants.

bgavrilMS: Or have a new object `RetryPolicyFactory.CreateStsRetry()`

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -82,6 +82,9 @@ protected override ManagedIdentityRequest CreateRequest(string resource)
                     break;
             }
 
+            // uncomment in follow-up IMDS retry policy PR
```
bgavrilMS, line 85: nit: I recommend you remove this. It's dead code. The new PR will do the right thing.

bgavrilMS (approved):
LGTM! Great PR. Just one minor comment.

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
```diff
@@ -60,8 +60,8 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
         /// or setting the Agent.
         /// </summary>
         /// <param name="httpClientFactory">HTTP client factory</param>
-        /// <param name="retryOnceOn5xx">Configures MSAL to retry on 5xx server errors. When enabled (on by default), MSAL will wait 1 second after receiving
-        /// a 5xx error and then retry the http request again.</param>
+        /// <param name="retryOnceOn5xx">Configures MSAL to ignore the internal retry policy.
+        /// The developer will be responsible for configuring their own retry policy in their custom IMsalHttpClientFactory.</param>
```
neha-bhargava, line 64:
```suggestion
        /// <param name="retryOnceOn5xx">Configures MSAL to retry using an internal retry policy.
        /// The developer will be responsible for configuring their own retry policy in their custom IMsalHttpClientFactory when disabled. This flag is enabled by default. </param>
```

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
```diff
@@ -73,7 +73,7 @@ public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory)
         public T WithHttpClientFactory(IMsalHttpClientFactory httpClientFactory, bool retryOnceOn5xx)
         {
             Config.HttpClientFactory = httpClientFactory;
-            Config.RetryOnServerErrors = retryOnceOn5xx;
+            Config.DisableInternalRetries = retryOnceOn5xx;
```
neha-bhargava, line 76: I think this is inversely related. DisableInternalRetries = !retryOnceOn5xx.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
```diff
@@ -28,31 +28,20 @@ internal sealed class MockHttpManager : IHttpManager,
...
-            Func<MockHttpMessageHandler> messageHandlerFunc = null,
-            Func<HttpClient> validateServerCertificateCallback = null,
-            bool invokeNonMtlsHttpManagerFactory = false) :
-            this(true, testName, isManagedIdentity, messageHandlerFunc, invokeNonMtlsHttpManagerFactory)
-        {
-        }
-
         public MockHttpManager(
-            bool retry,
+            bool disableInternalRetries = false,
```
neha-bhargava, line 32: Let's not add the default value here and rely on what is passed

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
Robbie-Microsoft: resolved offline

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs
Robbie-Microsoft: resolved offline

src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs
Robbie-Microsoft: Great catch. Thanks.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
Robbie-Microsoft: I'll make a separate PR for a RetryPolicyFactory since this PR is so big already.

src/client/Microsoft.Identity.Client/Instance/Region/RegionManager.cs
Robbie-Microsoft: I'll make a separate PR for a RetryPolicyFactory since this PR is so big already.

src/client/Microsoft.Identity.Client/Instance/Validation/AdfsAuthorityValidator.cs
```diff
@@ -28,15 +29,23 @@ public async Task ValidateAuthorityAsync(
                 var resource = $"https://{authorityInfo.Host}";
                 string webFingerUrl = Constants.FormatAdfsWebFingerUrl(authorityInfo.Host, resource);
 
-                Http.HttpResponse httpResponse = await _requestContext.ServiceBundle.HttpManager.SendRequestAsync(
+                LinearRetryPolicy _linearRetryPolicy = new LinearRetryPolicy(
+                    LinearRetryPolicy.DefaultStsRetryDelayMs,
+                    LinearRetryPolicy.DefaultStsMaxRetries,
+                    HttpRetryConditions.Sts);
+
+        Http.HttpResponse httpResponse = await _requestContext.ServiceBundle.HttpManager.SendRequestAsync(
```
neha-bhargava, line 37: nit: indentation

gladjohn (approved):
looks good @Robbie-Microsoft, do we have a test with internal retries disabled to verify that exactly one outbound call is made on 5xx.

Robbie-Microsoft:
> looks good @Robbie-Microsoft, do we have a test with internal retries disabled to verify that exactly one outbound call is made on 5xx.

This functionality already existed, so I modified the existing test to include the updated reference to `disableInternalRetries`. Additionally, I'm overhauling the tests in my follow-up PR, already on GitHub, for Imds + Default Retry Policy.

src/client/Microsoft.Identity.Client/Instance/Validation/AdfsAuthorityValidator.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -102,9 +111,9 @@ public async Task<HttpResponse> SendRequestAsync(
                 timeoutException = exception;
             }
 
-            while (_retryPolicy.pauseForRetry(response, timeoutException, retryCount))
+            while (!_disableInternalRetries && retryPolicy.PauseForRetry(response, timeoutException, retryCount))
```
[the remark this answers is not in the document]
Robbie-Microsoft: I deleted it in the latest commit

neha-bhargava (approved):

Issue #5261, opened 2025-05-02
[Bug] WithAdditionalCacheParameters fails to persist when cache is serialized


trwalke:
### Library version used

4.71.1

### .NET version

net 462 -> net 8.0

### Scenario

ConfidentialClient - service to service (AcquireTokenForClient)

### Is this a new or an existing app?

None

### Issue description and reproduction steps

When setting a serialized cache and using the `WithAdditionalCacheParameters` api, the persisted cache parameters are lost after attempting to acquire them from the cache on the second authentication attempt

### Expected behavior

MSAL should return the persisted parameters from the serialized cache

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

### Solution and workarounds

Update the JSON logic in the `MsalAccessTokenCacheItem` to store the `PersistedCacheParameters` collection in the serialized JSON.

PR #5262, opened 2025-05-02
Ensure additional cache parameters are persisted in cache serialization

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5261

`PersistedCacheParameters` needs to be serialized in the cache JSON Object when storing the cache externally. This is to ensure they can be retrieved the next time the token is acquired from the external cache.

**Testing**
Unit test

**Performance impact**

**Documentation**


tests/Microsoft.Identity.Test.Unit/AuthExtension/AuthenticationOperationTests.cs
```diff
@@ -67,7 +68,9 @@ public async Task Should_UseCustomRequestHeaders_And_StoreAdditionalParameters()
         }
 
         [TestMethod]
-        public async Task Should_UseCustomRequestHeaders_And_StoreAdditionalParametersWithCaching()
+        [DataRow(false)]
+        [DataRow(true)]
+        public async Task Should_UseCustomRequestHeaders_And_StoreAdditionalParametersWithCaching(bool useSerializedCache)
```
bgavrilMS, line 73: Is this the regression test then? If so, pls add a comment.

bgavrilMS (approved):

bgavrilMS (approved):

PR #5269, opened 2025-05-09
Add the CodeQL suppression in the right place for wstrustrequest parameter.

ashok672:
The previous  CodeQL suppression didn't work and adding the suppression comment just before the parameter.


trwalke (approved):

rayluo (approved):
"the only way to know is if the codeQL bug gets auto closed. that takes 48 hours to reflect."

PR #5274, opened 2025-05-13
Add mac broker console app support

fengga:
NativeInterop 0.19.0 from nuget already contains mac console app support code.
This PR enables console app support from MSAL.dotnet side, and add a sample app.

Internally, this PR removed authority validation if it is a broker flow. Brokers have their own authority validation processes, and making http requests will switch to a background thread, which will break the main thread context for broker ineractive flows.

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/tests/devapps/MacConsoleAppWithBroker/MacConsoleAppWithBroker.cs
+++ b/tests/devapps/MacConsoleAppWithBroker/MacConsoleAppWithBroker.cs
@@ -1,0 +1,231 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Broker;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using System.IO;
+
+class MacConsoleAppWithBroker
+{
+    private static readonly int MainThreadId = Thread.CurrentThread.ManagedThreadId;
+
+    private static readonly ConcurrentQueue<(Action Action, TaskCompletionSource<bool> Completion, bool IsAsyncAction)> MainThreadActions = 
+        new ConcurrentQueue<(Action, TaskCompletionSource<bool>, bool)>();
+    
+    private static volatile bool _workerFinished = false;
+    
+    static async Task Main(string[] args)
+    {
+
+        _ = Task.Run(() => BackgroundWorker());
+
+        while (!_workerFinished)
+        {
+            while (MainThreadActions.TryDequeue(out var actionItem))
+            {
+                try
+                {
+                    actionItem.Action();
+                    if (!actionItem.IsAsyncAction)
+                    {
+                        actionItem.Completion.TrySetResult(true);
+                    }
+                }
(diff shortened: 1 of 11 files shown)


src/client/Microsoft.Identity.Client/Utils/MacMainThreadScheduler.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            _mainThreadActions = new ConcurrentQueue<MainThreadActionItem>();
+            _workerFinished = false;
+            _isRunning = false;
+        }
+
+        /// <summary>
+        /// Check if the current thread is the main thread.
+        /// </summary>
+        public bool IsCurrentlyOnMainThread => Environment.CurrentManagedThreadId == 1; // Main thread id is always 1 on amcOS.
```
bgavrilMS:
```suggestion
        public bool IsCurrentlyOnMainThread => Environment.CurrentManagedThreadId == 1; // Main thread id is always 1 on macOS.
```

src/client/Microsoft.Identity.Client/Utils/MacMainThreadScheduler.cs
```diff
@@ -0,0 +1,148 @@
...
+                            {
+                                actionItem.Completion.TrySetResult(true);
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            actionItem.Completion.TrySetException(ex);
+                        }
+                    }
+                    Thread.Sleep(10);
```
iulico-1: >Thread.Sleep(10); [](http://example.com/codeflow?start=20&length=17)

Is this intentional ? Add a comment pls and consider if it should be configurable

iulico-1: Magic numbers like this are difficult to understand.

tests/devapps/MacConsoleAppWithBroker/MacConsoleAppWithBroker.cs
```diff
@@ -0,0 +1,232 @@
...
+
+    private static readonly ConcurrentQueue<(Action Action, TaskCompletionSource<bool> Completion, bool IsAsyncAction)> MainThreadActions = 
+        new ConcurrentQueue<(Action, TaskCompletionSource<bool>, bool)>();
+    
+    private static volatile bool _workerFinished = false;
+    
+    static async Task Main(string[] args)
+    {
+
+        _ = Task.Run(() => BackgroundWorker());
```
iulico-1, line 16: >BackgroundWorker [](http://example.com/codeflow?start=27&length=16)

We'll have to update our public docs to describe the pattern. Are there utils that we could share with the console apps to make interactive pattern easier to implement ?

iulico-1 (dismissed):
:shipit:

tests/devapps/MacConsoleAppWithBroker/MacConsoleAppWithBroker.cs
fengga: Exposed MacMainThreadScheduler to public and using it in the sample app. @bgavrilMS, are you OK with this?

bgavrilMS (approved):

PR #5292, opened 2025-05-19
Fix regression in ManagedIdentity flows

### LINKED ISSUE #5286: [Bug] HttpClientTransport setting is not applied after MSAL upgraded to 4.71.1
### Library version used

4.71.1

### .NET version

.Net 472

### Scenario

The application, running on an Azure VM (tenant xxxx), directs calls to a separate Azure VM (tenant cccc), and it is from this VM that Azure API requests are sent. Access tokens were acquired from the VM (tenant cccc).  After upgraded Microsoft.Identity.Client to 4.71.1, requests start failing with error message like:

Azure.RequestFailedException: The access token is from the wrong issuer 'https://sts.windows.net/xxxx'. It must match one of the tenants 'https://sts.windows.net/cccc/,https://sts.windows.net/bbbb/' associated with this subscription. Please use any authority (URL) from 'https://sts.windows.net/ccc/,https://sts.windows.net/bbbb/ to get the token. Note, if the subscription is transferred to another tenant there is no impact to the services, but information about new tenant could take time to propagate (up to an hour). If you just transferred your subscription and see this error message, please try back later.
Status: 401 (Unauthorized)
ErrorCode: InvalidAuthenticationTokenTenant

### Relevant code snippets

```csharp
var tokenCredentialOptions = new TokenCredentialOptions()
{
    Transport = new HttpClientTransport(customHttpClient);
}

var credential = new ManagedIdentityCredential(null, tokenCredentialOptions);

var options = new ArmClientOptions
{
    Environment = new ArmEnvironment(endpoint, authAudience),    
    Transport = new HttpClientTransport(customHttpClient)
};
var client = new ArmClient(credential, subscriptionId, options)
```

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

neha-bhargava:
Fixes #5286 5286

Update the flow to only set the callback for service fabric

**Testing**
Unit tests

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
@@ -12,6 +12,9 @@ using Microsoft.Identity.Client.Core;
 using System.Net;
 using Microsoft.Identity.Client.ApiConfig.Parameters;
 using System.Text;
+using System.Security.Cryptography.X509Certificates;
+using System.Net.Security;
+
 #if SUPPORTS_SYSTEM_TEXT_JSON
 using System.Text.Json;
 #else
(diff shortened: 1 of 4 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -22,11 +25,14 @@ namespace Microsoft.Identity.Client.ManagedIdentity
...
     {
+        private const string ManagedIdentityPrefix = "[Managed Identity] ";
+
         protected readonly RequestContext _requestContext;
+
         internal const string TimeoutError = "[Managed Identity] Authentication unavailable. The request to the managed identity endpoint timed out.";
         internal readonly ManagedIdentitySource _sourceType;
-        private const string ManagedIdentityPrefix = "[Managed Identity] ";
-
+        internal Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> ValidateServerCertificate = null;
```
bgavrilMS, line 34: I don't think this should be a field. It should be a property or a virtual method `GetValidator()`. I asked copilot "In OOP, in the base class, is it better to have a virtual method to override or to have a protected field that derived classes set" but didn't get a clear answer - maybe it will make it sense to you.

If you go for a property, you can use "protected internal"

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
neha-bhargava: Makes sense. I updated the code to use a virtual method instead.

gladjohn (approved):

PR #5299, opened 2025-05-21
PROTOTYPE: Adding WithServiceFabricFmi()

trwalke:
This PR is adding WithServiceFabricFmi which will be used to enable MI applications to detect the FMI service fabric endpoints.

Fixes #

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
@@ -21,11 +21,33 @@ namespace Microsoft.Identity.Client.ManagedIdentity
 
         public static AbstractManagedIdentity Create(RequestContext requestContext)
         {
+            Uri endpointUri;
             string identityEndpoint = EnvironmentVariables.IdentityEndpoint;
 
             requestContext.Logger.Info(() => "[Managed Identity] Service fabric managed identity is available.");
 
-            if (!Uri.TryCreate(identityEndpoint, UriKind.Absolute, out Uri endpointUri))
+            if (requestContext.ServiceBundle.Config.IsFmiServiceFabric)
+            {
+                identityEndpoint = EnvironmentVariables.FmiServiceFabricEndpoint;
+                requestContext.Logger.Info(() => "[Managed Identity] Using FMI Service fabric endpoint.");
+
+                if (!Uri.TryCreate(identityEndpoint, UriKind.Absolute, out endpointUri))
+                {
+                    string errorMessage = string.Format(CultureInfo.InvariantCulture, MsalErrorMessage.ManagedIdentityEndpointInvalidUriError,
+                            "APP_IDENTITY_ENDPOINT", identityEndpoint, "FMI Service Fabric");
+
+                    // Use the factory to create and throw the exception
+                    var exception = MsalServiceExceptionFactory.CreateManagedIdentityException(
+                        MsalError.InvalidManagedIdentityEndpoint,
+                        errorMessage,
+                        null,
+                        ManagedIdentitySource.ServiceFabric,
+                        null);
+
+                    throw exception;
+                }
+            }
+            else if (!Uri.TryCreate(identityEndpoint, UriKind.Absolute, out endpointUri))
             {
                 string errorMessage = string.Format(CultureInfo.InvariantCulture, MsalErrorMessage.ManagedIdentityEndpointInvalidUriError,
                         "IDENTITY_ENDPOINT", identityEndpoint, "Service Fabric");
@@ -102,6 +124,11 @@ namespace Microsoft.Identity.Client.ManagedIdentity
             }
 
             return request;
+        }
+
+        internal string GetEndpointForTesting()
+        {
+            return _endpoint.ToString();
         }
     }
 }
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ServiceFabricTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ServiceFabricTests.cs
@@ -92,6 +92,29 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
         }
 
         [TestMethod]
+        public void ValidateThatFmiEndpointIsUsed()
+        {
+            using (new EnvVariableContext())
+            using (var httpManager = new MockHttpManager())
+            {
+                SetEnvironmentVariables(ManagedIdentitySource.ServiceFabric, "http://localhost:40342/metadata/identity/oauth2/token", fmiEndpoint: "http://localhost:40343/metadata/identity/oauth2/token");
+
+                var miBuilder = ManagedIdentityApplicationBuilder.Create(ManagedIdentityId.SystemAssigned)
+                    .WithServiceFabricFmi()
+                    .WithHttpManager(httpManager);
+
+                var mi = miBuilder.BuildConcrete();
+
+                RequestContext requestContext = new RequestContext(mi.ServiceBundle, Guid.NewGuid(), null);
+
+                ServiceFabricManagedIdentitySource sf = ServiceFabricManagedIdentitySource.Create(requestContext) as ServiceFabricManagedIdentitySource;
+
+                Assert.IsInstanceOfType(sf, typeof(ServiceFabricManagedIdentitySource));
+                Assert.AreEqual("http://localhost:40343/metadata/identity/oauth2/token", sf.GetEndpointForTesting());
+            }
+        }
+
+        [TestMethod]
         public async Task SFThrowsWhenGetHttpClientWithValidationIsNotImplementedAsync()
         {
             using (new EnvVariableContext())
--- a/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
@@ -123,6 +123,7 @@ namespace Microsoft.Identity.Client
         public string CertificateIdToAssociateWithToken { get; set; }
 
         public Func<AppTokenProviderParameters, Task<AppTokenProviderResult>> AppTokenProvider;
+        public bool IsFmiServiceFabric { get; set; }
 
         #region ClientCredentials
 
--- a/src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs
@@ -103,6 +103,18 @@ namespace Microsoft.Identity.Client
         }
 
         /// <summary>
+        /// Configures the application to use the FMI service fabric managed identity endpoint.
+        /// </summary>
+        /// <remarks>
+        /// This is used only for Service Fabric applications that are using the FMI managed identity endpoint.
+        /// </remarks>
+        public ManagedIdentityApplicationBuilder WithServiceFabricFmi()
+        {
+            Config.IsFmiServiceFabric = true;
+            return this;
+        }
+
+        /// <summary>
         /// Builds an instance of <see cref="IManagedIdentityApplication"/> 
         /// from the parameters set in the <see cref="ManagedIdentityApplicationBuilder"/>.
         /// </summary>
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/EnvironmentVariables.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/EnvironmentVariables.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Identity.Client.ManagedIdentity
     internal class EnvironmentVariables
     {
         public static string IdentityEndpoint => Environment.GetEnvironmentVariable("IDENTITY_ENDPOINT");
+        public static string FmiServiceFabricEndpoint => Environment.GetEnvironmentVariable("APP_IDENTITY_ENDPOINT");
         public static string IdentityHeader => Environment.GetEnvironmentVariable("IDENTITY_HEADER");
         public static string PodIdentityEndpoint => Environment.GetEnvironmentVariable("AZURE_POD_IDENTITY_AUTHORITY_HOST");
         public static string ImdsEndpoint => Environment.GetEnvironmentVariable("IMDS_ENDPOINT");
--- a/src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt
+++ b/src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt
@@ -1,0 +1,1 @@
+Microsoft.Identity.Client.ManagedIdentityApplicationBuilder.WithServiceFabricFmi() -> Microsoft.Identity.Client.ManagedIdentityApplicationBuilder
--- a/src/client/Microsoft.Identity.Client/PublicApi/net472/PublicAPI.Unshipped.txt
+++ b/src/client/Microsoft.Identity.Client/PublicApi/net472/PublicAPI.Unshipped.txt
@@ -1,0 +1,1 @@
+Microsoft.Identity.Client.ManagedIdentityApplicationBuilder.WithServiceFabricFmi() -> Microsoft.Identity.Client.ManagedIdentityApplicationBuilder
--- a/src/client/Microsoft.Identity.Client/PublicApi/net8.0-android/PublicAPI.Unshipped.txt
+++ b/src/client/Microsoft.Identity.Client/PublicApi/net8.0-android/PublicAPI.Unshipped.txt
@@ -1,0 +1,1 @@
+Microsoft.Identity.Client.ManagedIdentityApplicationBuilder.WithServiceFabricFmi() -> Microsoft.Identity.Client.ManagedIdentityApplicationBuilder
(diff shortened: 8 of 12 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -21,11 +21,33 @@ internal class ServiceFabricManagedIdentitySource : AbstractManagedIdentity
 
         public static AbstractManagedIdentity Create(RequestContext requestContext)
         {
+            Uri endpointUri;
             string identityEndpoint = EnvironmentVariables.IdentityEndpoint;
 
             requestContext.Logger.Info(() => "[Managed Identity] Service fabric managed identity is available.");
 
-            if (!Uri.TryCreate(identityEndpoint, UriKind.Absolute, out Uri endpointUri))
+            if (requestContext.ServiceBundle.Config.IsFmiServiceFabric)
```
trwalke, line 29: I was considering letting MSAL autodetect the correct env variable, but I am not sure if it is a good idea. Both env variables may be set (IDENTITY_ENDPOINT and APP_IDENTITY_ENDPOINT), so I don't want to introduce bugs. Adding an api for this that will only be used from MISE will make this more robust.

bgavrilMS: Let's take a step back and think how we'd go about doing this in MSAL if we were to productize getting the FMI credential. 

The FMI credential is just a token with a specific audience. And it can use different endpoints. 

Can we use the existing MSI APIs, and have different logic based on the audience? i.e. if app developer requests token for `api://AzureFMITokenExchange` (or the GUID format) - then detect it?

trwalke: I understand what you are saying, but since this scenario isnt intended to use a resource and is simply acquiring the token from MITS, all of the additional application logic for resources/scopes/caching is not needed. I would also would be concerned if someone wanted to use MI to exchange an FMI credential for an access token using `api://AzureFMITokenExchange`, but instead they just get another FMI Credential.  so, maybe instead of using the fmi token exchange resource, we can create some string like "FmiMitsAcquisition" or something so that no one will trigger this logic in MSAL by accident? But if you feel like no one will ever use `api://AzureFMITokenExchange` with a MI app, this this will work.

bgavrilMS: I'm ok with an experimental API `mi,AcquireFmiCredential()`

bgavrilMS: @trwalke - can you put toghter a small design doc for this and let's discuss with the team?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs
```diff
@@ -42,6 +42,13 @@ private AcquireTokenForManagedIdentityParameterBuilder WithResource(string resou
         {
             Parameters.Resource = ScopeHelper.RemoveDefaultSuffixIfPresent(resource);
             CommonParameters.Scopes = new string[] { Parameters.Resource };
+
+            if (resource.Equals("api://AzureFMITokenExchange/.default"))
```
bgavrilMS: need to use case insensitive equals

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForManagedIdentityParameters.cs
```diff
@@ -16,6 +16,8 @@ internal class AcquireTokenForManagedIdentityParameters : IAcquireTokenParameter
 
         public string Resource { get; set; }
 
+        public bool IsFmiServiceFabric { get; set; }
```
bgavrilMS: IsFmiCredential ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricFederatedManagedIdentitySource.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Globalization;
+using System.Net.Http;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Internal;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal class ServiceFabricFederatedManagedIdentitySource : AbstractManagedIdentity
```
bgavrilMS: Seems like a lot of code for this. Is there no opportunity for code reuse?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -53,6 +53,11 @@ public enum ManagedIdentitySource
         /// <summary>
         /// The source to acquire token for managed identity is Machine Learning Service.
         /// </summary>
-        MachineLearning
+        MachineLearning,
+
+        /// <summary>
+        /// The source to acquire token for managed identity is Service Fabric Federated.
+        /// </summary>
+        ServiceFabricFederated
```
bgavrilMS: I would use the term `FMI` instead of `Federated`, because federated is very overloaded.

bgavrilMS: Do we really need a new source here?

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricFederatedManagedIdentitySource.cs
trwalke: There is, but I wanted to make this completely separate. however, it doesn't hurt to merge the two

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
trwalke: It was only needed when there were 2 sources. Now that they are merged, only one is needed

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -77,31 +123,43 @@ private ServiceFabricManagedIdentitySource(RequestContext requestContext, Uri en
...
-                    break;
+                _requestContext.Logger.Info("[Managed Identity] Request is for FMI, no ids or resource will be added to the request.");
+                request.QueryParameters["api-version"] = EnvironmentVariables.FmiServiceFabricApiVersion;
+            }
+            else
+            {
+                request.QueryParameters["api-version"] = ServiceFabricMsiApiVersion;
+                request.QueryParameters["resource"] = resource;
+
+                switch (_requestContext.ServiceBundle.Config.ManagedIdentityId.IdType)
```
MZOLN, line 145: Shouldnt there be a default case? We could log the unexpected, that could give us additional information when troubleshooting.

trwalke: Technically yes, but this ID is a required parameter for the managed identity application. So it will always be set to something unless there is a catastrophic failure.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -16,34 +16,79 @@ internal class ServiceFabricManagedIdentitySource : AbstractManagedIdentity
         private const string ServiceFabricMsiApiVersion = "2019-07-01-preview";
         private readonly Uri _endpoint;
         private readonly string _identityHeaderValue;
+        private readonly bool _isFederated;
+        private static string _mitsEndpointFmiPath => "/metadata/identity/oauth2/fmi/credential";
 
-        internal static Lazy<HttpClient> _httpClientLazy;
-
-        public static AbstractManagedIdentity Create(RequestContext requestContext)
+        public static AbstractManagedIdentity Create(RequestContext requestContext, bool isFmiServiceFabric = false)
```
bgavrilMS: I don't think this is right. The source detection should not interfere with `isFmi` flag. **The MSI source should be expected to be cached (statically, in memory) by MSAL.** 

In other words, each MSI request can either go to the normal endpoint or to the fmi endpoint.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -62,11 +107,12 @@ private bool ValidateServerCertificateCallback(HttpRequestMessage message, X509C
...
         }
 
-        private ServiceFabricManagedIdentitySource(RequestContext requestContext, Uri endpoint, string identityHeaderValue) : 
-        base(requestContext, ManagedIdentitySource.ServiceFabric)
+        private ServiceFabricManagedIdentitySource(RequestContext requestContext, Uri endpoint, string identityHeaderValue, bool isFmi) :
+            base(requestContext, ManagedIdentitySource.ServiceFabric)
         {
             _endpoint = endpoint;
             _identityHeaderValue = identityHeaderValue;
+            _isFederated = isFmi;
```
bgavrilMS: Pls try to use the same naming consistently. smth along the lines of `isFmiCredentialRequest` or smth that includes the "fmi credntial" term

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ServiceFabricTests.cs
```diff
@@ -91,6 +91,80 @@ public void ValidateServerCertificateCallback_ServerCertificateValidationCallbac
...
+
+                ServiceFabricManagedIdentitySource sf = ServiceFabricManagedIdentitySource.Create(requestContext, true) as ServiceFabricManagedIdentitySource;
+
+                Assert.IsInstanceOfType(sf, typeof(ServiceFabricManagedIdentitySource));
+                Assert.AreEqual("http://localhost:40342/metadata/identity/oauth2/fmi/credential", sf.GetEndpointForTesting());
+            }
+        }
+
+        [TestMethod]
+        public async Task ValidateThatFmiCredentialCanBeAcquiredFromMits()
```
bgavrilMS, line 118: You need a test where you ask for a FMI credential, then for an FMI token (and repeat).

bgavrilMS (changes requested):
source detection should not depend on isFmi

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
trwalke: I dont think this is the case. The source detection is effectively turned off if `IsFmi` is enabled and only the FMI source will be used. 
See [ManagedIdentityClient](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/62095ff69c61eec22ea96a1efd9dd5327dd49e9d/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs#L40)

This parameter specifically is to change this source to only do FMI.

Also, when you say cached (statically, in memory), can more that one of these source objects be cached? meaning, can there be one for regular SF and one for FMI?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ServiceFabricTests.cs
trwalke: I didnt add this because this code is only responsible for acquiring the credential. The full E2E is already tested in MISE. Are we sure we also want to add the full e2e test here?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs
```diff
@@ -42,6 +42,13 @@ private AcquireTokenForManagedIdentityParameterBuilder WithResource(string resou
         {
             Parameters.Resource = ScopeHelper.RemoveDefaultSuffixIfPresent(resource);
             CommonParameters.Scopes = new string[] { Parameters.Resource };
+
+            if (resource.Equals("api://AzureFMITokenExchange/.default", StringComparison.OrdinalIgnoreCase))
+            {
+                Parameters.isFmiCredentialRequest = true;
```
bgavrilMS: Do we need an extra param here? Can't ManagedIdentity classes deal with this?

trwalke: Sure, I can do that

trwalke: Actually, it is better to keep this here because there the check is needed for the experimental features to be enabled.

bgavrilMS: I'm ok with dropping the experimental feature flag if it leads to better design.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
bgavrilMS: It's still a good practice to throw an exception @trwalke on default. It will protect against a case where a new enum parameter is added.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs
bgavrilMS: This is still not resolved.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -21,11 +21,11 @@ internal class ManagedIdentityClient
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         private readonly AbstractManagedIdentity _identitySource;
 
-        public ManagedIdentityClient(RequestContext requestContext)
+        public ManagedIdentityClient(RequestContext requestContext, AcquireTokenForManagedIdentityParameters acquireTokenForManagedIdentityParameters)
```
bgavrilMS, line 24: This design changes too many things. As stated above, you can simply look at the scope, which indicates this is an MSI request for a credential.

trwalke:
Update, this work was put on hold as of July since a higher priority item needed to be completed.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
trwalke: added

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
trwalke: Updated to reduce the additional logic

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -16,34 +16,86 @@ internal class ServiceFabricManagedIdentitySource : AbstractManagedIdentity
         private const string ServiceFabricMsiApiVersion = "2019-07-01-preview";
         private readonly Uri _endpoint;
         private readonly string _identityHeaderValue;
+        private readonly bool _isFmiCredentialRequest;
```
bgavrilMS, line 19: I recommend you park for now this or sync with @Robbie-Microsoft who's refactoring this completely.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
```diff
@@ -16,34 +16,86 @@ internal class ServiceFabricManagedIdentitySource : AbstractManagedIdentity
...
+                        null);
+                }
+            }
+
+            requestContext.Logger.Verbose(() => $"[Managed Identity] Creating Service Fabric {(isFmiCredentialRequest ? "federated" : "")} managed identity. Endpoint URI: {identityEndpoint}");
 
-            requestContext.Logger.Verbose(() => "[Managed Identity] Creating Service Fabric managed identity. Endpoint URI: " + identityEndpoint);
-           
-            return new ServiceFabricManagedIdentitySource(requestContext, endpointUri, EnvironmentVariables.IdentityHeader);
+            return new ServiceFabricManagedIdentitySource(requestContext, endpointUri, EnvironmentVariables.IdentityHeader, isFmiCredentialRequest);
```
bgavrilMS, line 74: YOu can have 1 request for FMI credential and 1 request for normal token. So this won't work.

bgavrilMS:
@Robbie-Microsoft - pls review this.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
trwalke: Not sure what you mean by this. The client will only do one or the other. The FMI version of this will only be used by MISE for the credential.

src/client/Microsoft.Identity.Client/ManagedIdentity/ServiceFabricManagedIdentitySource.cs
trwalke: ack

trwalke:
closing, will recreate with updated implementation

PR #5307, opened 2025-05-29
Make it consistent that both MAJOR and MINOR versions have one year support

rayluo:
Based on an internal discussion.

Fixes # N/A

Clarify that we provide one year support, which means accepting bug reports and providing patches.

**Testing**

N/A

**Performance impact**

N/A

**Documentation**
- [X] All relevant documentation is updated.


MZOLN (approved):

RELEASES.md
```diff
@@ -28,9 +28,9 @@ version="[1.1,1.2)"
...
+When we release a new MINOR version, the previous MINOR versions shipped within one year MAY still accept bug report and receive patches. MINOR versions more than one year old will not receive any support. If you suspect there is an issue in a 1+ year old version that you are using, please upgrade to latest MINOR version and retry, before you send out a bug report.
 
-When we release a new MAJOR version, we will continue to apply bug fixes to the existing features in the previous MAJOR version for up to the 2 year support cycle for Azure.
+When we release a new MAJOR version, we will continue to apply bug fixes to the existing features in the previous MAJOR version for up to the 1 year support cycle for Azure.
 Example: We release MSALiOS 2.0 in the future which supports unified Auth for AAD and MSA. Later, we then have a fix in Conditional Access for MSALiOS. Since that feature exists both in MSALiOS 1.1 and MSALiOS 2.0, we will fix both. It will roll up in a PATCH number for each. Customers that are still locked down on MSALiOS 1.1 will receive the benefit of this fix.
 
 ## Microsoft Authentication SDKs and Azure Active Directory
```
Avery-Dunn, line 36: A couple lines down it also says "MAJOR version will continue to work for 2 years after release", and in the table with descriptions/examples above it mentions "We will continue to service MAJOR version numbers with bug fixes up to the two year support cycle"

Both of those sections are also talking about Azure AD support, so is the 2 years of support there different from MSAL's 1 year support?

rayluo: Good catch. I modified it with a new commit now.

RELEASES.md
```diff
@@ -28,14 +28,14 @@ version="[1.1,1.2)"
...
-Example: We release MSALiOS 2.0 in the future which supports unified Auth for AAD and MSA. Later, we then have a fix in Conditional Access for MSALiOS. Since that feature exists both in MSALiOS 1.1 and MSALiOS 2.0, we will fix both. It will roll up in a PATCH number for each. Customers that are still locked down on MSALiOS 1.1 will receive the benefit of this fix.
+When we release a new MAJOR version, we will continue to apply bug fixes to the existing features in the previous MAJOR version for up to the 1 year support cycle for Azure.
+Example: We release MSALiOS 2.0 in the future which supports unified Auth for Microsoft Entra. Later, we then have a fix in Conditional Access for MSALiOS. Since that feature exists both in MSALiOS 1.1 and MSALiOS 2.0, we will fix both. It will roll up in a PATCH number for each. Customers that are still locked down on MSALiOS 1.1 will receive the benefit of this fix.
 
-## Microsoft Authentication SDKs and Azure Active Directory
+## Microsoft Authentication SDKs and Microsoft Entra
```
gladjohn, line 36:
```suggestion
## Microsoft Authentication SDKs and Microsoft Entra ID
```

RELEASES.md
```diff
@@ -28,14 +28,14 @@ version="[1.1,1.2)"
 
 ## Serviceability
 
-When we release a new MINOR version, the previous MINOR version is abandoned.
+When we release a new MINOR version, the previous MINOR versions shipped within one year MAY still accept bug report and receive patches. MINOR versions more than one year old will not receive any support. If you suspect there is an issue in a 1+ year old version that you are using, please upgrade to the latest MINOR version and retry, before you send out a bug report.
```
jennyf19, line 31: why are you updating this and not the LTS section? Also, we need to align on changes here, the changes proposed were approved across the org, it's not a decision we can make at the team level.

jennyf19 (changes requested):

RELEASES.md
rayluo: > why are you updating this and not the LTS section?

By "LTS section", do you mean we should use the [supportPolicy.md](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/supportPolicy.md#overview) as single source of truth? That one is the only .md file in this repo containing the "LTS" keyword.

If that is the case, we would still need to modify the "Serviceability" section in the current file. Currently, neither the "minor version is abandoned (immediately?)" nor the "2 years" description matches the supportPolicy.md's "180 days" policy. Alternatively, we can completely remove the "Serviceability" section and its next paragraph here.

> Also, we need to align on changes here, the changes proposed in the LTS were approved across the org, it's not a decision we can make at the team level.

I totally agree with you. Do we have an org-wide support policy URL that our SDKs can simply link to, rather than maintaining a local copy in each repo? I would rather have us avoid spending cycles on reviewing/debating changes in this kind of PR.

jennyf19: Each repo is a bit different, so we cannot point to one LTS. There are no cycles to spend on this as the work was already done, https://aka.ms/msal/LTS, and signed off by the MSAL team. The "Serviceabilty" section should just point to the LTS.

rayluo: Great! I guess you mean non-MSAL repos are different (and rightfully so), but all MSALs can and shall point to the https://aka.ms/msal/LTS (the link is even language agnostic). I'll send out a commit that remove the content of the "Serviceability" section with a link to the  https://aka.ms/msal/LTS

UPDATE: PR updated.

bgavrilMS (changes requested):
Please have the PCA team open PR on their scenarios.

PR #5310, opened 2025-05-30
Use HTTP 2

bgavrilMS:
Fixes https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5209 , based on [sherlock1982](https://github.com/sherlock1982)s' feedback

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -194,9 +194,12 @@ namespace Microsoft.Identity.Client.Http
             return _httpClientFactory.GetHttpClient();
         }
 
+        private static readonly Version s_http2Version = new Version(2, 0);
+
         private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string, string> headers)
         {
             HttpRequestMessage requestMessage = new HttpRequestMessage { RequestUri = endpoint };
+            requestMessage.Version = s_http2Version;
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -194,9 +194,12 @@ private HttpClient GetHttpClient(X509Certificate2 x509Certificate2, Func<HttpReq
             return _httpClientFactory.GetHttpClient();
         }
 
+        private static readonly Version s_http2Version = new Version(2, 0);
+
         private static HttpRequestMessage CreateRequestMessage(Uri endpoint, IDictionary<string, string> headers)
         {
             HttpRequestMessage requestMessage = new HttpRequestMessage { RequestUri = endpoint };
+            requestMessage.Version = s_http2Version;
```
gladjohn, line 202: will we automatically drop to 1.1 where not  supported?

gladjohn, line 202:
```suggestion
            requestMessage.Version = HttpVersion.Version20;
```

gladjohn (approved):

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
bgavrilMS: Cool, I didn't know about this.

Frulfump:
This was closed without a comment, would be nice with an explanation :)

PR #5313, opened 2025-05-31
update to langVersion 13

jennyf19:
Downlevel "latest" maps to pre-13. In the future, we'll want to either keep this up-to-date (e.g. update it to 14 when .NET 10 ships) or change it to preview, which will always be up-to-date.

note: I haven't tested anything

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>13</LangVersion>
     <!-- This is strong naming, not signing-->
     <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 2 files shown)


trwalke:
Language version 13 is not available on the linux agents
_"CSC(0,0): Error CS1617: Invalid option '13' for /langversion. Use '/langversion:?' to list supported values."_
We will need to wait for the build agents to be updated to the later versions. So not sure if setting to a specific value will work best here. Different agents we use may have different supported versions at different times. Maybe preview will be better? or not setting it at all so that the compiler can choose the correct version?

Need some additional testing

MZOLN:
Might be MSAL specific, we are finding 13 working well on linux.

Frulfump:
This was closed without a comment, please provide the reason

PR #5315, opened 2025-06-02
Use the actual client IDs for FOCI

bgavrilMS:
Fixes an internal repair item - reduces dependency on a KeyVault which is on a deprecation path.

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/FociTests.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/FociTests.cs
@@ -115,8 +115,8 @@ namespace Microsoft.Identity.Test.Integration.SeleniumTests
         private static void CreateFamilyApps(LabResponse labResponse, string cacheFilePath, out IPublicClientApplication pca_fam1, out IPublicClientApplication pca_fam2, out IPublicClientApplication pca_nonFam)
         {
             var keyvault = new KeyVaultSecretsProvider(KeyVaultInstance.MsalTeam);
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/FociTests.cs
```diff
@@ -115,8 +115,8 @@ await AssertException.TaskThrowsAsync<MsalUiRequiredException>(() => pca_nonFam
         private static void CreateFamilyApps(LabResponse labResponse, string cacheFilePath, out IPublicClientApplication pca_fam1, out IPublicClientApplication pca_fam2, out IPublicClientApplication pca_nonFam)
         {
             var keyvault = new KeyVaultSecretsProvider(KeyVaultInstance.MsalTeam);
-            var clientId1 = keyvault.GetSecretByName(TestConstants.FociApp1KeyVaultSecretName).Value;
-            var clientId2 = keyvault.GetSecretByName(TestConstants.FociApp2KeyVaultSecretName).Value;
+            var clientId1 = "872cd9fa-d31f-45e0-9eab-6e460a02d1f1";
+            var clientId2 = "1950a258-227b-4e31-a9cf-717495945fc2";
```
rayluo, line 119: **question (unblocking)**

Out of curiosity, are these two apps managed by lab? If so, we may consider getting their IDs from a (new?) lab api without needing to hardcode them here.

Regardless, this change LGTM. Approving.

bgavrilMS: Yes, managed by the lab. ClientIDs are not secret.

PR #5327, opened 2025-06-06
Another build fix due to Mac dev apps

bgavrilMS:


trwalke:
We should have these rules in an `editor.config` file. It might be that these rules do not reach these test projects.

trwalke (approved):

PR #5328, opened 2025-06-06
Fix mac broker MAUI issues

fengga:
Fixes #

For mac broker flows, limit the "start new message loop to guarantee  go back to main thread" logic to console app only. Maui app should not start a new message loop since it already has one running.

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/tests/devapps/MacMauiAppWithBroker/MacMauiAppWithBroker.csproj
+++ b/tests/devapps/MacMauiAppWithBroker/MacMauiAppWithBroker.csproj
@@ -90,11 +90,7 @@
     </ItemGroup> -->
 
 	<ItemGroup Condition="$([MSBuild]::IsOSPlatform('OSX'))">
-		<PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.18.2" />
-		<NativeReference Include="$(HOME)/.nuget/packages/microsoft.identity.client.nativeinterop/0.18.2/runtimes/macos-arm64/native/msalruntime_arm64.dylib">
-			<Kind>Dynamic</Kind>
-			<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-		</NativeReference>
+		<PackageReference Include="Microsoft.Identity.Client.NativeInterop" Version="0.19.0" />
 	</ItemGroup>
 
 	<ItemGroup Condition="$([MSBuild]::IsOSPlatform('OSX'))">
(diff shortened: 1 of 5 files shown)


bgavrilMS (approved):

tests/devapps/MacMauiAppWithBroker/MacMauiAppWithBroker.csproj
```diff
@@ -74,10 +77,6 @@
 		<PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiVersion)" />
 		<PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
 		<PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="8.0.1" />
-
-		<PackageReference Include="Microsoft.Maui.Controls" Version="$(MauiVersion)" />
-		<PackageReference Include="Microsoft.Maui.Controls.Compatibility" Version="$(MauiVersion)" />
-		<PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="8.0.1" />
 	</ItemGroup>
 
 	<!-- <ItemGroup>
```
bgavrilMS, line 82: dead code?

fengga: These are duplicated code and will trigger some warnings

bgavrilMS (approved):

Directory.Packages.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <!-- Version of the Microsoft.Identity.Client.NativeInterop package. -->
-    <MSALRuntimeNativeInteropVersion>0.18.1</MSALRuntimeNativeInteropVersion>
+    <MSALRuntimeNativeInteropVersion>0.19.1</MSALRuntimeNativeInteropVersion>
```
xinyuxu1026: Do we need to update the changelog somewhere? since we updated the package version.

xinyuxu1026 (approved):

Directory.Packages.props
gladjohn: not for a simple version bump but if there was a bug fix then we should file a bug and add it here in this PR

gladjohn (approved):
looks good, but can we log what OS was detected before calling the Native Interp?

Directory.Packages.props
fengga: #5343 created and linked

fengga:
If change NativeInterop version to > 0.18.1 some tests will fail. Probably a regression.
Issue created: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5346

bgavrilMS (approved):

PR #5330, opened 2025-06-06
Expose access token cache count

### LINKED ISSUE #5329: [Feature Request] Expose cache access token count
### MSAL client type

Confidential

### Problem statement

To help to isolate memory consumption MSAL vs rest of functionality, apps need to know the size of cache, particularly for the static cache. See https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3283679 for more details.

### Proposed solution

_No response_

### Alternatives

_No response_

bgavrilMS:
Fixes #5329 

Expose the cached access token count.

**Testing**
Unit

**Performance impact**
By keeping a count variable, we do not enumerate dictionaries, so this op is O(1)

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
@@ -97,5 +97,11 @@ namespace Microsoft.Identity.Client
         /// See https://aka.ms/msal-net-logging for more details about logging.
         /// </remarks>
         public string Telemetry { get; set; }
+
+        /// <summary>
+        /// The number of entries in the token cache. For app tokens, this is the number of access tokens cached. 
+        /// For users tokens, an entry contains all tokens for a user, including refresh token and id tokens.
+        /// </summary>
+        public int CachedAccessTokenCount { get; set; }
     }
 }
--- a/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
+++ b/src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
@@ -99,8 +99,7 @@ namespace Microsoft.Identity.Client
         public string Telemetry { get; set; }
 
         /// <summary>
-        /// The number of entries in the token cache. For app tokens, this is the number of access tokens cached. 
-        /// For users tokens, an entry contains all tokens for a user, including refresh token and id tokens.
+        /// The number of access tokens in the cache.       
         /// </summary>
         public int CachedAccessTokenCount { get; set; }
     }
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -48,9 +48,11 @@ namespace Microsoft.Identity.Client.Cache
             return await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
         }
 
-        public Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem, Account>> SaveTokenResponseAsync(MsalTokenResponse tokenResponse)
+        public async Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem, Account>> SaveTokenResponseAsync(MsalTokenResponse tokenResponse)
         {
-            return TokenCacheInternal.SaveTokenResponseAsync(_requestParams, tokenResponse);
+            var result = await TokenCacheInternal.SaveTokenResponseAsync(_requestParams, tokenResponse).ConfigureAwait(false);
+            RequestContext.ApiEvent.CacheAccessTokenCount = TokenCacheInternal.Accessor.EntryCount;
+            return result;
         }
 
         public async Task<Account> GetAccountAssociatedWithAccessTokenAsync(MsalAccessTokenCacheItem msalAccessTokenCacheItem)
@@ -181,6 +183,8 @@ namespace Microsoft.Identity.Client.Cache
             {
                 RequestContext.ApiEvent.CacheLevel = CacheLevel.L1Cache;
             }
+
+            RequestContext.ApiEvent.CacheAccessTokenCount = TokenCacheInternal.Accessor.EntryCount;
         }
     }
 }
--- a/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
+++ b/src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Identity.Client.Cache
         public async Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem, Account>> SaveTokenResponseAsync(MsalTokenResponse tokenResponse)
         {
             var result = await TokenCacheInternal.SaveTokenResponseAsync(_requestParams, tokenResponse).ConfigureAwait(false);
-            RequestContext.ApiEvent.CacheAccessTokenCount = TokenCacheInternal.Accessor.EntryCount;
+            RequestContext.ApiEvent.CachedAccessTokenCount = TokenCacheInternal.Accessor.EntryCount;
             return result;
         }
 
@@ -184,7 +184,7 @@ namespace Microsoft.Identity.Client.Cache
                 RequestContext.ApiEvent.CacheLevel = CacheLevel.L1Cache;
             }
 
-            RequestContext.ApiEvent.CacheAccessTokenCount = TokenCacheInternal.Accessor.EntryCount;
+            RequestContext.ApiEvent.CachedAccessTokenCount = TokenCacheInternal.Accessor.EntryCount;
         }
     }
 }
--- a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
+++ b/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
@@ -38,8 +38,8 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
         private int _entryCount = 0;
         private static int s_entryCount = 0;
 
-        public int EntryCount => _tokenCacheAccessorOptions.UseSharedCache ? s_entryCount : _entryCount;
-
+        public int EntryCount => GetEntryCountRef();
+       
         public InMemoryPartitionedAppTokenCacheAccessor(
             ILoggerAdapter logger,
             CacheOptions tokenCacheAccessorOptions)
@@ -67,21 +67,14 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
 
             var partition = AccessTokenCacheDictionary.GetOrAdd(partitionKey, _ => new ConcurrentDictionary<string, MsalAccessTokenCacheItem>());
             bool added = partition.TryAdd(itemKey, item);
+
+            // only increment the entry count if the item was added, not updated
             if (added)
             {
-                if (_tokenCacheAccessorOptions.UseSharedCache)
-                {
-                    Interlocked.Increment(ref s_entryCount);
-                }
-                else
-                {
-                    Interlocked.Increment(ref _entryCount);
-                }
+                Interlocked.Increment(ref GetEntryCountRef());
             }
             else
             {
-                // If not added, it means it was replaced. Only increment if it was a new key.
-                // If you want to treat replacement as not changing the count, do nothing here.
                 partition[itemKey] = item;
             }
         }
@@ -156,14 +149,7 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
                 bool removed = partition.TryRemove(item.CacheKey, out _);
                 if (removed)
                 {
-                    if (_tokenCacheAccessorOptions.UseSharedCache)
-                    {
-                        Interlocked.Decrement(ref s_entryCount);
-                    }
-                    else
-                    {
-                        Interlocked.Decrement(ref _entryCount);
-                    }
+                    Interlocked.Decrement(ref GetEntryCountRef());
                 }
                 else
                 {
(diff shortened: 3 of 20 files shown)


src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
```diff
@@ -97,5 +97,11 @@ public AuthenticationResultMetadata(TokenSource tokenSource)
         /// See https://aka.ms/msal-net-logging for more details about logging.
         /// </remarks>
         public string Telemetry { get; set; }
+
+        /// <summary>
+        /// The number of entries in the token cache. For app tokens, this is the number of access tokens cached. 
+        /// For users tokens, an entry contains all tokens for a user, including refresh token and id tokens.
+        /// </summary>
+        public int CachedAccessTokenCount { get; set; }
```
Robbie-Microsoft: Do you think we should have two different variables here? Could the developer ONLY want to count cached access tokens and not care about any other token type allowed in the user cache?

Alternatively, you could rename the variable to `CachedTokenCount` since you're currently counting ALL types of tokens.

src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
```diff
@@ -48,9 +48,11 @@ public async Task<MsalAccessTokenCacheItem> FindAccessTokenAsync()
             return await TokenCacheInternal.FindAccessTokenAsync(_requestParams).ConfigureAwait(false);
         }
 
-        public Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem, Account>> SaveTokenResponseAsync(MsalTokenResponse tokenResponse)
+        public async Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem, Account>> SaveTokenResponseAsync(MsalTokenResponse tokenResponse)
         {
-            return TokenCacheInternal.SaveTokenResponseAsync(_requestParams, tokenResponse);
+            var result = await TokenCacheInternal.SaveTokenResponseAsync(_requestParams, tokenResponse).ConfigureAwait(false);
```
Robbie-Microsoft, line 53: Should result be typed with the return value of SaveTokenResponseAsync? Or does it not matter in this case?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -34,6 +35,11 @@ internal class InMemoryPartitionedAppTokenCacheAccessor : ITokenCacheAccessor
         protected readonly ILoggerAdapter _logger;
         private readonly CacheOptions _tokenCacheAccessorOptions;
 
+        private int _entryCount = 0;
+        private static int s_entryCount = 0;
+
+        public int EntryCount => _tokenCacheAccessorOptions.UseSharedCache ? s_entryCount : _entryCount;
```
Robbie-Microsoft: I learned some dotnet concepts just now by analyzing what you did in this file: `ref` and "expression-bodied member syntax". Thanks! You set this up very nicely.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -59,9 +65,25 @@ public void SaveAccessToken(MsalAccessTokenCacheItem item)
...
+                    Interlocked.Increment(ref s_entryCount);
+                }
+                else
+                {
+                    Interlocked.Increment(ref _entryCount);
+                }
+            }
+            else
+            {
+                // If not added, it means it was replaced. Only increment if it was a new key.
```
Robbie-Microsoft: I would move this line above the if and change it to "Only increment if a new key was created"

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -59,9 +65,25 @@ public void SaveAccessToken(MsalAccessTokenCacheItem item)
...
+                // If you want to treat replacement as not changing the count, do nothing here.
```
Robbie-Microsoft:
```suggestion
                // Otherwise, the cache item is being replaced. Do not increment.
```

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -129,12 +151,26 @@ public void DeleteAccessToken(MsalAccessTokenCacheItem item)
...
+                    if (_tokenCacheAccessorOptions.UseSharedCache)
+                    {
+                        Interlocked.Decrement(ref s_entryCount);
+                    }
+                    else
+                    {
+                        Interlocked.Decrement(ref _entryCount);
+                    }
+                }
+                else
```
Robbie-Microsoft: You could add the same comments from above. Or do you think it's more appropropriate to only have them in that one spot in the code?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedUserTokenCacheAccessor.cs
```diff
@@ -73,8 +79,23 @@ public void SaveAccessToken(MsalAccessTokenCacheItem item)
...
+                {
+                    System.Threading.Interlocked.Increment(ref s_entryCount);
+                }
+                else
+                {
+                    System.Threading.Interlocked.Increment(ref _entryCount);
+                }
+            }
+            else
+            {
```
Robbie-Microsoft: add comments like in the other file?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedUserTokenCacheAccessor.cs
```diff
@@ -151,12 +172,26 @@ public void DeleteAccessToken(MsalAccessTokenCacheItem item)
...
+                if (removed)
+                {
+                    if (_tokenCacheAccessorOptions.UseSharedCache)
+                    {
+                        System.Threading.Interlocked.Decrement(ref s_entryCount);
+                    }
+                    else
+                    {
+                        System.Threading.Interlocked.Decrement(ref _entryCount);
+                    }
```
Robbie-Microsoft: add comments like in the other file?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedUserTokenCacheAccessor.cs
```diff
@@ -326,6 +361,14 @@ public virtual void Clear(ILoggerAdapter requestlogger = null)
...
             IdTokenCacheDictionary.Clear();
             AccountCacheDictionary.Clear();
+            if (_tokenCacheAccessorOptions.UseSharedCache)
+            {
+                System.Threading.Interlocked.Exchange(ref s_entryCount, 0);
+            }
+            else
+            {
+                System.Threading.Interlocked.Exchange(ref _entryCount, 0);
+            }
```
Robbie-Microsoft: add comments like in the other file?

tests/Microsoft.Identity.Test.Performance/AcquireTokenForClientCacheTests.cs
```diff
@@ -56,7 +55,7 @@ public async Task GlobalSetupAsync()
             _cca = ConfidentialClientApplicationBuilder
                 .Create(TestConstants.ClientId)
                 .WithRedirectUri(TestConstants.RedirectUri)
-                .WithClientSecret(TestConstants.ClientSecret)
+                .WithClientSecret(TestConstants.ClientSecret)
```
Robbie-Microsoft: undo whitespace changes in this file?

tests/Microsoft.Identity.Test.Unit/CacheTests/LoadingProjectsTests.cs
```diff
@@ -8,7 +8,6 @@
 
 namespace Microsoft.Identity.Test.Unit.CacheTests
 {
-#if !ANDROID && !iOS  // custom token cache serialization not available
```
Robbie-Microsoft, line 11: Should the changes in this file be in a separate PR?

tests/Microsoft.Identity.Test.Unit/ParallelRequestsTests.cs
```diff
@@ -121,9 +121,13 @@ public async Task AcquireTokenForClient_ConcurrentTenantRequests_Test()
 
             // Wait for all tasks to complete
             AuthenticationResult[] results = await Task.WhenAll(tasks).ConfigureAwait(false);
+            int[] accessTokenCounts = results.Select(r => r.AuthenticationResultMetadata.CachedAccessTokenCount).ToArray();
 
             // Assert the total tasks
             Assert.AreEqual(NumberOfRequests, results.Length, "Number of AuthenticationResult objects does not match the number of requests.");
+
+            var expected = Enumerable.Range(1, NumberOfRequests).ToArray();
```
Robbie-Microsoft: Can you rename this for clarity? `expected` what?

Robbie-Microsoft:
Is this new telemetry always enabled? And there's no way to disable it? Is that correct?

src/client/Microsoft.Identity.Client/AuthenticationResultMetadata.cs
bgavrilMS: If you look at InMemoryPartitionedUserTokenAccessor, you'll notice that it only counts AT.

Robbie-Microsoft: yep, thanks!

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
```diff
@@ -59,9 +65,25 @@ public void SaveAccessToken(MsalAccessTokenCacheItem item)
...
             string partitionKey = CacheKeyFactory.GetAppTokenCacheItemKey(item.ClientId, item.TenantId, item.KeyId, item.AdditionalCacheKeyComponents);
 
-            // if a conflict occurs, pick the latest value
-            AccessTokenCacheDictionary
-                .GetOrAdd(partitionKey, new ConcurrentDictionary<string, MsalAccessTokenCacheItem>())[itemKey] = item;
+            var partition = AccessTokenCacheDictionary.GetOrAdd(partitionKey, _ => new ConcurrentDictionary<string, MsalAccessTokenCacheItem>());
+            bool added = partition.TryAdd(itemKey, item);
+            if (added)
+            {
+                if (_tokenCacheAccessorOptions.UseSharedCache)
```
neha-bhargava: You can extract this if else block in a method as 
```
private ref int GetEntryCountRef()
{
    return ref _tokenCacheAccessorOptions.UseSharedCache ? ref s_entryCount : ref _entryCount;
}
```
and then use it as 
```
if (added)
{
    Interlocked.Increment(ref GetEntryCountRef());
}
``` 
here and in the decrement, exchange logic

src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt
```diff
@@ -1,7 +1,9 @@
+Microsoft.Identity.Client.AuthenticationResultMetadata.CachedAccessTokenCount.get -> int
+Microsoft.Identity.Client.AuthenticationResultMetadata.CachedAccessTokenCount.set -> void
```
neha-bhargava: Should this be exposed or setter should be internal?

src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
bgavrilMS: MSAL follows this style rule related to `var` - https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008#csharp_style_var_when_type_is_apparent

Given that the return type is extremely verbose `Task<Tuple<MsalAccessTokenCacheItem, MsalIdTokenCacheItem, Account>>` I think it's fine as is?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
bgavrilMS: Very cool, I had no idea

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs
bgavrilMS: Deleting doesn't happen that often

src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt
bgavrilMS: The other properties in here have public setters , just to allow app developers to unit test stuff around MSAL without complex mocking etc.

tests/Microsoft.Identity.Test.Unit/CacheTests/LoadingProjectsTests.cs
bgavrilMS: Possibly. But since this is test code and I'm 99.9% sure this will break anything, I'd rather just do this. If you feel strongly about this, I can pull this out.

Robbie-Microsoft: I would've another PR since it appears to me to be an unrelated change, but you're right that it's so small that you can include it here.

src/client/Microsoft.Identity.Client/Cache/CacheSessionManager.cs
Robbie-Microsoft: agreed, it's fine. Thanks for that link.

PR #5340, opened 2025-06-12
Remove the new Mac dev apps from build

bgavrilMS:
Fix the build


Robbie-Microsoft (commented):
Could you explain to me what these changes do?

For the addition, it looks you added a line that was previously mistakenly missing?

For the removals, how do you know they are for the Mac dev apps? By the clientIds? (Are those clientIds?)

Robbie-Microsoft (approved):

PR #5351, opened 2025-06-18
Added support for DEFAULT_IDENTITY_CLIENT_ID environment variable in Machine Learning Managed Identity

### LINKED ISSUE #5350: [Feature Request] Support for new Azure ML Managed Identity
### MSAL client type

Managed identity

### Problem statement

Core Library
MSAL Node (@azure/msal-node)

Wrapper Library
Not Applicable

Public or Confidential Client?
Confidential

Description
MSAL client type
Managed identity

Problem statement
The Azure ML Managed Identity API is not the same as the App Service 2017-09-01 API, requiring explicit support to accommodate its unique authentication flow.

Key Differences
Expiration Time Format

The expires_on field is returned as an integer, whereas App Service returns it as a string.
The implementation must correctly handle integer-based expiration times
Mandatory clientid Parameter

Unlike App Service, all token requests must specify a clientid, even for system-assigned managed identities.
The platform provides a default client ID via the environment variable:
DEFAULT_IDENTITY_CLIENT_ID
The system should default to this environment variable when no client ID is explicitly provided.
Proposed solution
Add explicit support for the Azure ML Managed Identity API in MSAL.

Alternatives
No response

Notes
Original ask in Azure SDK for JS: https://github.com/Azure/azure-sdk-for-js/issues/27920

### Proposed solution

_No response_

### Alternatives

_No response_

Robbie-Microsoft:
Fixes #5350 

If the Machine Learning System Assigned Managed Identity is being used, the "clientid" request parameter will be set to the DEFAULT_IDENTITY_CLIENT_ID environment variable instead of being omitted in the network request.


gladjohn:
looks good @Robbie-Microsoft, but there are few MI test failures, are they related?

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
```diff
@@ -73,21 +77,33 @@ protected override ManagedIdentityRequest CreateRequest(string resource)
...
+                    request.QueryParameters[Constants.ManagedIdentityClientId2017] = EnvironmentVariables.MachineLearningDefaultClientId; // this environment variable is always set in an Azure Machine Learning source
+                    break;
+
                 case AppConfig.ManagedIdentityIdType.ClientId:
                     _requestContext.Logger.Info("[Managed Identity] Adding user assigned client id to the request.");
                     // Use the new 2017 constant for older ML-based environment
                     request.QueryParameters[Constants.ManagedIdentityClientId2017] = _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
                     break;
 
-                case AppConfig.ManagedIdentityIdType.ResourceId:
```
bgavrilMS, line 82: Are you sure these can't be used with AppService 2017? We should not break people.

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
```diff
@@ -73,21 +77,33 @@ protected override ManagedIdentityRequest CreateRequest(string resource)
 
             switch (_requestContext.ServiceBundle.Config.ManagedIdentityId.IdType)
             {
+                case AppConfig.ManagedIdentityIdType.SystemAssigned:
+                    _requestContext.Logger.Info("[Managed Identity] Adding system assigned client id to the request.");
+                    // Use the new 2017 constant for older ML-based environment
+                    request.QueryParameters[Constants.ManagedIdentityClientId2017] = EnvironmentVariables.MachineLearningDefaultClientId; // this environment variable is always set in an Azure Machine Learning source
```
bgavrilMS: Are you sure this code path won't be hit for some other AppService2017 forgotten environment? A safer way to implement this is to check that `MachineLearningDefaultClientId` exists.

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
```diff
@@ -73,21 +77,33 @@ protected override ManagedIdentityRequest CreateRequest(string resource)
...
-                    _requestContext.Logger.Info("[Managed Identity] Adding user assigned resource id to the request.");
-                    request.QueryParameters[Constants.ManagedIdentityResourceId] = _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
-                    break;
-
-                case AppConfig.ManagedIdentityIdType.ObjectId:
-                    _requestContext.Logger.Info("[Managed Identity] Adding user assigned object id to the request.");
-                    request.QueryParameters[Constants.ManagedIdentityObjectId] = _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
-                    break;
+                default:
+                    string errorMessage = string.Format(
```
bgavrilMS: This is so hard to read. Please, just use a string here. Why do you need all this templating ?

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
Robbie-Microsoft: Yes, that's what @chlowell told me. @chlowell, can you confirm?

gladjohn:
> Do not make breaking changes to AppService2017, instead just add support for this Default_Client_id stuff

a well explained bug on the changes we are making will be a good place to start with. The reason why I ask this is, It is not clear if Azure ML Source and Azure ML API source are same or are they different? 

My understanding, is Azure ML uses App Service 2017. Does Azure ML API also use App Service 2017?

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
bgavrilMS: So based on discussions with Charles, I think it's ok to assume `MachineLearningDefaultClientId` exists. But let's please  throw some exception in case it doesn't

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
bgavrilMS: We discussed on chat. Sounds good.

chlowell:
> My understanding, is Azure ML uses App Service 2017. Does Azure ML API also use App Service 2017?

Azure ML and App Service have always had different managed identity implementations. Azure ML imitated App Service 2017 when designing their managed identity API, however the implementations never behaved the same and Azure ML's API is now incompatible with App Service 2017 in that it requires `clientid` for all token requests.

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
Robbie-Microsoft: done

neha-bhargava (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/MachineLearningManagedIdentitySource.cs
Robbie-Microsoft: I don't. Using a string now.

PR #5358, opened 2025-06-23
Add issuer validation when making call to OIDC endpoint

Avery-Dunn:
This PR adds issuer validation for OIDC scenarios: [3268765](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3268765)

- When using the WithOidcAuthority() API, a call is made to the OIDC discovery endpoint based on the authority set for the application
- Part of the response is an expected issuer, and if this issuer differs from what we expect based on the authority we will throw an exception.

As per OAuth spec, an issuer is valid if it matches the authority. In addition to the standard spec, there is also special validation for CIAM-style issuers which will mainly be relevant when using custom authorities in a CIAM scenario.


src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -70,6 +75,58 @@ public static async Task<OidcMetadata> GetOidcAsync(
...
+        private static void ValidateIssuer(Uri authority, string issuer)
+        {
+
+            // Primary validation: check if authority begins with the issuer string (exact string comparison)
+            if (authority.AbsoluteUri.StartsWith(issuer, StringComparison.OrdinalIgnoreCase))
+            {
+                return;
+            }
+
+            // Extract tenant for CIAM-like scenarios
```
bgavrilMS, line 93: I don't think this is needed. MSAL knows about normal CIAM authorities and will not perform OIDC discovery. I double checked this by running `ROPC_Ciam_StandardDomains_CompletesSuccessfully`

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/CiamIntegrationTests.cs
```diff
@@ -213,6 +213,31 @@ public async Task OBOCiam_CustomDomain_ReturnsValidTokens()
...
+            {
+                string.Format("https://{0}.ciamlogin.com/", labResponse.User.LabName),
+                string.Format("https://{0}.ciamlogin.com/{1}.onmicrosoft.com", labResponse.User.LabName, labResponse.User.LabName),
+                string.Format("https://{0}.ciamlogin.com/{1}", labResponse.User.LabName, labResponse.Lab.TenantId),
+                string.Format("https://login.msidlabsciam.com/{0}/v2.0/", labResponse.Lab.TenantId)
+            };
+
+            foreach (var authority in authorities)
+            {
+                await RunCiamCCATest(authority, labResponse.App.AppId).ConfigureAwait(false);
```
bgavrilMS: This will mostly use the `WithAuthority` API in MSAL, which gets MSAL to parse the authority and to understand it's a CIAM authority, in which case it won't perform OIDC discovery. 

IMO, you want to force test the `WithOidcAuthority` route, which performs OIDC discovery.

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
bgavrilMS: There are 2 ways to solve this: 

1. Assume that app developers will always use `WithAuthority` API for CIAM and AAD authorities. They would only use `WithOidcAuthority` for CIAM CUD. In which case, you only need block 88-90
2. OR, assume that app developers can mix and match. And that some will use `WithOidcAuthority(a)` with AAD and regular CIAM, in which case your logic needs more code for AAD.

If you go on the first path, then pls make sure to run Id.Web tests, to check that they do the right thing.

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/CiamIntegrationTests.cs
```diff
@@ -213,6 +213,31 @@ public async Task OBOCiam_CustomDomain_ReturnsValidTokens()
...
+            var labResponse = await LabUserHelper.GetLabUserDataAsync(new UserQuery()
+            {
+                FederationProvider = FederationProvider.CIAMCUD,
+                SignInAudience = SignInAudience.AzureAdMyOrg
+            }).ConfigureAwait(false);
+
+            //Test with different CIAM authority formats
+            string[] authorities =
+            {
+                string.Format("https://{0}.ciamlogin.com/", labResponse.User.LabName),
```
trwalke: I think these should be added as `DataRow(...)` and passed into the test as parameters so that the test will indicate which version of the authority failed.

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -70,6 +75,58 @@ public static async Task<OidcMetadata> GetOidcAsync(
...
 
+        /// <summary>
+        /// Validates that the issuer in the OIDC metadata matches the authority.
+        /// </summary>
+        /// <param name="authority">The authority URL.</param>
+        /// <param name="issuer">The issuer from the OIDC metadata - the single source of truth.</param>
+        /// <exception cref="MsalServiceException">Thrown when issuer validation fails.</exception>
+        private static void ValidateIssuer(Uri authority, string issuer)
+        {
+
```
neha-bhargava, line 89: nit: extra space

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
Avery-Dunn: I think we discussed this a while back, however this is needed because in a CIAM scenario the authority can a custom domain but the issuer from the OIDC endpoint will still follow the `{tenant}.ciamlogin.com` format.

I couldn't find any doc officially saying that's always true, however that's the validation described in the work item and the endpoints for the CIAM labs resources have `{tenant}.ciamlogin.com` as the issuer.

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/CiamIntegrationTests.cs
Avery-Dunn: Is there some way to uniquely identify a `DataRow` during a test? Because each string is a different format and populated with different data, so wouldn't that require a bunch of if-else statements to figure out which URL the test is working with?

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/CiamIntegrationTests.cs
Avery-Dunn: In the latest commit that `RunCiamCCATest()` now takes in an extra boolean parameter so the tests themselves control which authority API is used instead of it being decided by having "ciamlogin.com" in the authority.

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
Avery-Dunn: Fixed in the latest commit.

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -70,6 +75,62 @@ public static async Task<OidcMetadata> GetOidcAsync(
...
+            // If tenant extraction failed or returned empty, validation fails
+            if (!string.IsNullOrEmpty(tenant))
+            {
+                // Create a collection of valid CIAM issuer patterns for the tenant
+                string[] validCiamPatterns =
+                {
+                    $"https://{tenant}{Constants.CiamAuthorityHostSuffix}",
+                    $"https://{tenant}{Constants.CiamAuthorityHostSuffix}/{tenant}",
+                    $"https://{tenant}{Constants.CiamAuthorityHostSuffix}/{tenant}/v2.0"
```
Robbie-Microsoft, line 118: nit: should you make a helper string for $"https://{tenant}{Constants.CiamAuthorityHostSuffix}"?

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
```diff
@@ -70,6 +75,62 @@ public static async Task<OidcMetadata> GetOidcAsync(
...
+                {
+                    return;
+                }
+            }
+
+            // Validation failed
+            throw new MsalServiceException(
+                "authority_validation_failed",
+                $"Issuer validation failed for {authority}. Issuer from OIDC endpoint does not match any expected pattern: {issuer}");
+        }
```
Robbie-Microsoft: I think you should actually make a new exception for these.

Look at:
MsalError.cs
MsalErrorMessage.cs

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,44 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
             }
         }
 
+        [TestMethod]
+        public async Task OidcIssuerValidation_ThrowsForNonMatchingIssuer_Async()
+        {
+            using (var httpManager = new MockHttpManager())
+            {
+                string authority = "https://demo.duendesoftware.com";
```
Robbie-Microsoft: Should this be defined in the test constants file? Or can you use an existing test authority?

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,44 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
...
+                    .Create(TestConstants.ClientId)
+                    .WithHttpManager(httpManager)
+                    .WithOidcAuthority(authority)
+                    .WithClientSecret(TestConstants.ClientSecret)
+                    .Build();
+
+                // Create OIDC document with non-matching issuer
+                string validOidcDocumentWithWrongIssuer = TestConstants.GenericOidcResponse.Replace(
+                        "\"issuer\":\"https://demo.duendesoftware.com\"",
+                        "\"issuer\":\"https://wrong.issuer.com\"");
```
Robbie-Microsoft: see above about constants. And can't you re-use the demo url here instead of typing the string again?

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,44 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
...
+
+                // Mock OIDC endpoint response
+                httpManager.AddMockHandler(new MockHttpMessageHandler
+                {
+                    ExpectedMethod = HttpMethod.Get,
+                    ExpectedUrl = authority + "/.well-known/openid-configuration",
```
Robbie-Microsoft: I would think that this ".well-known..." string already exists in the test constants?

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,44 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
...
+
+                var ex = await AssertException.TaskThrowsAsync<MsalServiceException>(() =>
+                    app.AcquireTokenForClient(new[] { "api" }).ExecuteAsync()
+                ).ConfigureAwait(false);
+
+                Assert.AreEqual("authority_validation_failed", ex.ErrorCode);
+                Assert.IsTrue(ex.Message.Contains("Issuer validation failed"),
+                    $"Expected error message to contain 'Issuer validation failed', but was: {ex.Message}");
+                Assert.IsTrue(ex.Message.Contains("https://wrong.issuer.com"),
+                    "Error message should include the non-matching issuer value");
```
Robbie-Microsoft: re-use errors if you create them, like in the comment I mentioned above

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,44 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
...
+                });
+
+                var ex = await AssertException.TaskThrowsAsync<MsalServiceException>(() =>
+                    app.AcquireTokenForClient(new[] { "api" }).ExecuteAsync()
+                ).ConfigureAwait(false);
+
+                Assert.AreEqual("authority_validation_failed", ex.ErrorCode);
+                Assert.IsTrue(ex.Message.Contains("Issuer validation failed"),
+                    $"Expected error message to contain 'Issuer validation failed', but was: {ex.Message}");
+                Assert.IsTrue(ex.Message.Contains("https://wrong.issuer.com"),
```
Robbie-Microsoft: re-use this string from above?

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
Avery-Dunn: Added a new MsalErrorMessage.IssuerValidationFailed string in the latest commit

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
Avery-Dunn: Added a new MsalErrorMessage.IssuerValidationFailed string in the latest commit, and am now using MsalError.AuthorityValidationFailed for the error

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Avery-Dunn: Every existing test in GenericAuthorityTests defined that authority within each test, so I just assumed it was the normal style

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Avery-Dunn: In that latest commit it now re-uses the authority, and a new field for the wrong issuer was added and also re-used.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Avery-Dunn: In the latest commit it's now reused.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Avery-Dunn: I created them, and refactored these asserts a bit to look for the exact error code and message.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Avery-Dunn: Yep, it exists in the Constants file of the main library. That constant is now used in the latest commit.

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
bgavrilMS: Ack, this needs a comment in the code.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,46 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
             }
         }
 
+        [TestMethod]
+        public async Task OidcIssuerValidation_ThrowsForNonMatchingIssuer_Async()
```
bgavrilMS, line 337: What happens if you use `WithOidcAuthority("https://login.microsoftonline.com/common/v2.0")` ?

bgavrilMS (commented):
LGTM, but we should understand if `WithOidcAuthority(Entra)` will work on not (via tests). Id.Web might be using this API.

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
Robbie-Microsoft: I think you could go even further and make a helper string for $"https://{tenant}{Constants.CiamAuthorityHostSuffix}", then re-use that helper all 3 times.

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Robbie-Microsoft: I think replacing them all with a constant would be a good opportunity for another PR that can be merged into this one ;)

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,46 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
...
+                // Create OIDC document with non-matching issuer
+                string validOidcDocumentWithWrongIssuer = TestConstants.GenericOidcResponse.Replace(
+                        $"\"issuer\":\"{authority}\"",
+                        $"\"issuer\":\"{wrongIssuer}\"");
+
+                // Mock OIDC endpoint response
+                httpManager.AddMockHandler(new MockHttpMessageHandler
+                {
+                    ExpectedMethod = HttpMethod.Get,
+                    ExpectedUrl = authority + "/" + Constants.WellKnownOpenIdConfigurationPath,
```
Robbie-Microsoft: should you use template literal here?

$"{authority}/{Constants.WellKnownOpenIdConfigurationPath}"

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/CiamIntegrationTests.cs
Robbie-Microsoft: You can pass in array containing both the string and the corresponding url.

bgavrilMS (approved):

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Avery-Dunn: It depends entirely on whether what OIDC endpoints exist and what issuers it returns. Check out the issuer at https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration : the issuer is literally "https://login.microsoftonline.com/{tenantid}/v2.0" with the tenant ID not actually filled in. 

That doesn't follow the OIDC spec or our more custom validation, so I'm not sure we're meant to handle that sort of edge case. I assume this is one of the reasons we explicitly say not to use those sort of endpoints with the OIDC API: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/90443fa218403b4737ca900be5fe384a1f2765dc/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L342

src/client/Microsoft.Identity.Client/Instance/Oidc/OidcRetrieverWithCache.cs
Avery-Dunn: Improved in the latest commit:
` // Extract tenant for CIAM scenarios. In a CIAM scenario the issuer is expected to have "{tenant}.ciamlogin.com" as the host, even when using a custom domain.`

tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
Avery-Dunn: It now uses that literal, as well as the generic authority value found in TestConstants

gladjohn (approved):

billybooth:
Can anyone point me to the location in the OIDC [discovery spec](https://openid.net/specs/openid-connect-discovery-1_0.html) or [core spec](https://openid.net/specs/openid-connect-core-1_0.html) where it indicates that the authority and the issuer MUST match?

I know that the issuer returned by discovery MUST match the value in the `iss` claim, but I can't find an indication (with a reference to the OIDC or OAuth spec) that the authority URI must contain (start with) the issuer URI.

Of course I ask because this change is breaking discovery for one of the providers I care about, but I am also genuinely curious for the item in the spec that motivated this change.

Avery-Dunn:
@billybooth, I believe section 4.3 of the OIDC discovery spec is the relevant one for that change: 
> If any of the validation procedures defined in this specification fail, any operations requiring the information that failed to correctly validate MUST be aborted and the information that failed to validate MUST NOT be used.
>
>**The issuer value returned MUST be identical to the Issuer URL that was used as the prefix to /.well-known/openid-configuration to retrieve the configuration information.** This MUST also be identical to the iss Claim value in ID Tokens issued from this Issuer.

When using the `WithOidcAuthority` API we use the authority's host as the prefix to the `/.well-known/openid-configuration` endpoint, so it must match the issuer.

billybooth:
> @billybooth, I believe section 4.3 of the OIDC discovery spec is the relevant one for that change: 
> 
> > If any of the validation procedures defined in this specification fail, any operations requiring the information that failed to correctly validate MUST be aborted and the information that failed to validate MUST NOT be used.
> 
> >
> 
> >**The issuer value returned MUST be identical to the Issuer URL that was used as the prefix to /.well-known/openid-configuration to retrieve the configuration information.** This MUST also be identical to the iss Claim value in ID Tokens issued from this Issuer.
> 
> 
> 
> When using the `WithOidcAuthority` API we use the authority's host as the prefix to the `/.well-known/openid-configuration` endpoint, so it must match the issuer.

Thanks for digging that out! Good enough for me. This provider (who shall remain nameless) is diverging from spec in a few places, but this new validation finally busted it. Oh well, good on you guys for doing the right thing here.

Edit: As an aside, I was getting hung up on the `iss` claim and totally missed the bit about the issuer value in the discovery doc needing to prefix the URL from which it was retrieved, and that obviously gives the client an opportunity to reject the configuration out of hand. All makes sense now. 😅

PR #5359, opened 2025-06-23
CSR Metadata request acts as a probe for ImdsV2

Robbie-Microsoft:
This PR introduces ImdsV2 as a Managed Identity source, and focuses on executing a network request to get metadata to be used in the rest of the flow.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
+++ b/src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Identity.Client.Http.Retry
 
         /// <summary>
         /// Retry policy specific to CSR Metadata Probe.
-        /// Extends Imds retry policy but excludes 404 and 408 status codes.
+        /// Extends Imds retry policy but excludes 404 status code.
         /// </summary>
         public static bool CsrMetadataProbe(HttpResponse response, Exception exception)
         {
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Identity.Client.ManagedIdentity
                 }
             }
 
-            if (!ValidateCsrMetadataResponse(response, requestContext.Logger, probeMode))
+            if (!probeMode && !ValidateCsrMetadataResponse(response, requestContext.Logger, probeMode))
             {
                 return null;
             }


src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
```diff
@@ -0,0 +1,82 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    /// <summary>
+    /// Represents metadata required for Certificate Signing Request (CSR) operations.
+    /// </summary>
+    internal interface CsrMetadataInterface
```
bgavrilMS: Avoid creating interfaces for data objects. They serve no purpose.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// Concrete implementation of the CsrMetadata interface.
+    /// </summary>
+    internal class CsrMetadata : CsrMetadataInterface
+    {
+        public string ClientId { get; }
+        public string TenantId { get; }
+        public string Cuid { get; }
+        public string AttestationEndpoint { get; }
+
+        public CsrMetadata(CsrMetadataResponse response)
```
bgavrilMS: Should the constructor be private ? Since you prefer to build it via `CreateOrNull` ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,145 @@
...
+            string uami = null;
+            switch (requestContext.ServiceBundle.Config.ManagedIdentityId.IdType)
+            {
+                case AppConfig.ManagedIdentityIdType.ClientId:
+                    uami = requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId;
+                    requestContext.Logger.Info("[Managed Identity] ImdsV2 supports user-assigned client id.");
+                    break;
+
+                case AppConfig.ManagedIdentityIdType.ResourceId:
+                    requestContext.Logger.Info("[Managed Identity] ImdsV2 doesn't support user-assigned resource id. Either provide a client id, or use a system-assigned managed identity.");
```
bgavrilMS: Very suprised to see this. I thought MSIv2 is a drop-in replacement for MSIv1. @gladjohn ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                requestContext.Logger.Info(() => $"[Managed Identity] IMDSV2 managed identity is not available. 'server' header format/version invalid. Extracted version: {match.Groups[1].Value}");
+                return false;
+            }
+
+            CsrMetadataResponse csrMetadataResponse = JsonHelper.DeserializeFromJson<CsrMetadataResponse>(response.Body);
+            CsrMetadata csrMetadata = CsrMetadata.CreateOrNull(csrMetadataResponse, requestContext.Logger);
+            if (csrMetadata == null)
+            {
+                requestContext.Logger.Info(() => "[Managed Identity] IMDSV2 managed identity is not available. CsrMetadata is null.");
```
bgavrilMS:
```suggestion
                requestContext.Logger.Info(() => "[Managed Identity] IMDSV2 managed identity is not available. Invalid CsrMetadata response.");
```

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -0,0 +1,155 @@
...
+                RequestContext requestContext = new RequestContext(managedIdentityApp.ServiceBundle, Guid.NewGuid(), null);
+
+                httpManager.AddManagedIdentityMockHandler(
+                    CsrMetadataEndpoint,
+                    ManagedIdentityTests.Resource,
+                    MockHelpers.GetCsrMetadataSuccessfulResponse(),
+                    ManagedIdentitySource.ImdsV2,
+                    serverHeaderImdsVersion: serverHeaderImdsVersion);
+
+                bool isValid = await ImdsV2ManagedIdentitySource.GetCsrMetadataAsync(requestContext).ConfigureAwait(false);
```
bgavrilMS: Don't we expose the source in the public API @gladjohn ?

bgavrilMS: These are good tests, but not high level enough. Ideally we'd do an `AcquireToken` call and observe the source somehwere (hence my question above).

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -0,0 +1,155 @@
...
+        private const string serverHeaderImdsVersion = "IMDS/150.870.65.1325";
+        private const string CsrMetadataEndpoint = "http://169.254.169.254/metadata/identity/getPlatformMetadata";
+        private readonly TestRetryPolicyFactory _testRetryPolicyFactory = new TestRetryPolicyFactory();
+
+        [TestMethod]
+        public async Task GetCsrMetadataAsyncSucceeds()
+        {
+            using (var httpManager = new MockHttpManager())
+            {
+                var miBuilder = ManagedIdentityApplicationBuilder.Create(ManagedIdentityId.SystemAssigned)
```
bgavrilMS: Ok, so source discovery became much more expensive. I think we need to go back to caching the source? Not sure this happens ? It should be cached for all ManagedIdentityApplication objects?

Thoughts @gladjohn ?

bgavrilMS: And we need a test to prove that subsequent calls source discovery don't "rediscover".

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: This is directly from Raghavendra.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: Yes, we do. Why are you asking? You want me to call that api here?

Robbie-Microsoft: addressed in other comment

Robbie-Microsoft: Yes, eventually the whole flow will be tested instead of just testing the GetCsrMetadataAsync leg; these tests will change to reflect that, in the PR that adds the other leg of this flow. Otherwise, yes I can check the source via that public API. I noticed that only two other Managed Identity tests use this public API in a test, so I didn't use it. But I'll add it.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: Is making the source static not enough, like I've in this PR? This is what we do in msal-node.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -19,39 +19,46 @@ internal class ManagedIdentityClient
     {
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
-        private readonly AbstractManagedIdentity _identitySource;
+        private static RequestContext _requestContext;
+        private static AbstractManagedIdentity _identitySource;
```
bgavrilMS: naming: s_identitySource

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -19,39 +19,46 @@ internal class ManagedIdentityClient
     {
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
-        private readonly AbstractManagedIdentity _identitySource;
+        private static RequestContext _requestContext;
```
bgavrilMS: the request context must not be static.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
bgavrilMS: It is static, so this resolve it.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
bgavrilMS: `ManagedIdentityBuilder.GetSource()`

src/client/Microsoft.Identity.Client/PublicApi/net472/PublicAPI.Shipped.txt
```diff
@@ -970,7 +970,6 @@ static Microsoft.Identity.Client.Kerberos.KerberosSupplementalTicketManager.GetK
 static Microsoft.Identity.Client.Kerberos.KerberosSupplementalTicketManager.GetKrbCred(Microsoft.Identity.Client.Kerberos.KerberosSupplementalTicket ticket) -> byte[]
 static Microsoft.Identity.Client.Kerberos.KerberosSupplementalTicketManager.SaveToWindowsTicketCache(Microsoft.Identity.Client.Kerberos.KerberosSupplementalTicket ticket) -> void
 static Microsoft.Identity.Client.Kerberos.KerberosSupplementalTicketManager.SaveToWindowsTicketCache(Microsoft.Identity.Client.Kerberos.KerberosSupplementalTicket ticket, long logonId) -> void
-static Microsoft.Identity.Client.ManagedIdentityApplication.GetManagedIdentitySource() -> Microsoft.Identity.Client.ManagedIdentity.ManagedIdentitySource
```
bgavrilMS: Deprecate this instead of deleting it. See the [Obsolete] attribute (set it to soft deprecate). If ppl still use it, call `GetManagedIdentitySourceAsync().GetAwaiter().GetResult()`

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
```diff
@@ -0,0 +1,82 @@
...
+            AttestationEndpoint = response.AttestationEndpoint;
+        }
+
+        /// <summary>
+        /// Creates a CsrMetadata instance from a CsrMetadataResponse, logging warnings and returning null if any field is missing.
+        /// </summary>
+        /// <param name="response">The CsrMetadataResponse object.</param>
+        /// <param name="logger">The ILogger to log warnings.</param>
+        /// <returns>CsrMetadata instance or null if any field is null.</returns>
+        public static CsrMetadata CreateOrNull(CsrMetadataResponse response, ILoggerAdapter logger)
```
trwalke: I think `TryCreate` or `TryCreateCsrMetadata` might be a better name. So, when trying to create the CsrMetadata, if it fails, users will expect a null. It will also make it clear that the creation of this object is not guaranteed.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpMessageHandler.cs
```diff
@@ -147,6 +147,10 @@ private void ValidateHeaders(HttpRequestMessage request)
             {
                 foreach (var kvp in ExpectedRequestHeaders)
                 {
+                    // Skip validation for x-ms-client-request-id since it's random
+                    if (kvp.Key.Equals("x-ms-client-request-id", StringComparison.OrdinalIgnoreCase))
```
trwalke: Is this needed? If you dont add the `x-ms-client-request-id` to the expected headers collection when creating the handler, it wont be checked here.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,156 @@
...
+                    doNotThrow: false,
+                    mtlsCertificate: null,
+                    validateServerCertificate: null,
+                    cancellationToken: requestContext.UserCancellationToken,
+                    retryPolicy: retryPolicy)
+                .ConfigureAwait(false);
+            }
+            catch (Exception)
+            {
+                // try/catch is for testing purposes, to avoid adding a mock for this request
```
trwalke: I would avoid as much as possible adding code to the src projects just for testing purposes if you have another option like adding a mock handler. Is there supposed to be an exception when this fails in production?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -152,5 +139,38 @@ internal static string CreateRequestFailedMessage(HttpResponse response, string
 
             return messageBuilder.ToString();
         }
+
+        public static Uri GetValidatedEndpoint(
+            ILoggerAdapter logger,
+            string subPath,
```
neha-bhargava, line 145: subPath is not needed to passed as a variable as it is declared as a constant above.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -17,10 +17,10 @@ namespace Microsoft.Identity.Client.ManagedIdentity
     internal class ImdsManagedIdentitySource : AbstractManagedIdentity
     {
         // IMDS constants. Docs for IMDS are available here https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http
-        private static readonly Uri s_imdsEndpoint = new("http://169.254.169.254/metadata/identity/oauth2/token");
-
+        private const string DefaultImdsBaseEndpoint= "http://169.254.169.254";
         private const string ImdsTokenPath = "/metadata/identity/oauth2/token";
-        private const string ImdsApiVersion = "2018-02-01";
+        public const string ImdsApiVersion = "2018-02-01";
```
neha-bhargava, line 23: Why did you change it to public?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,156 @@
...
+                requestContext.Logger.Info(() => "[Managed Identity] IMDSV2 managed identity is not available. Exception occurred while sending request to CSR metadata endpoint.");
```
neha-bhargava: Also add the exception message to the log. Also this should be Logger.Error()

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,156 @@
...
+             * Capture the last number in a group (`(\d+)`)
+             * Ensure end of string (`$`)
+             *
+             * Example:
+             * [
+             * "IMDS/150.870.65.1556",  // index 0: full match
+             * "1556"                   // index 1: captured group (\d+)
+             * ]
+             */
+            string serverHeader = response.HeadersAsDictionary.TryGetValue("server", out var value) ? value : null;
```
neha-bhargava, line 131: Extract this logic into a method to process a successful response.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,156 @@
...
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal class ImdsV2ManagedIdentitySource : AbstractManagedIdentity
+    {
+        private const string CsrMetadataPath = "/metadata/identity/getPlatformMetadata";
+
+        private static CsrMetadata _csrMetadata;
+
+        public static async Task<bool> GetCsrMetadataAsync(RequestContext requestContext)
```
neha-bhargava: This method is doing too many things. I would recommend to break it down into validate, send request and process the response to help with readability and maintainability

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -59,9 +59,9 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
         /// Detects and returns the managed identity source available on the environment.
         /// </summary>
         /// <returns>Managed identity source detected on the environment if any.</returns>
-        public static ManagedIdentitySource GetManagedIdentitySource()
+        public static async Task<ManagedIdentitySource> GetManagedIdentitySourceAsync()
```
neha-bhargava: This is a breaking change. We cannot rename the public API.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -91,6 +98,11 @@ internal static ManagedIdentitySource GetManagedIdentitySource(ILoggerAdapter lo
                 logger?.Info("[Managed Identity] Azure Arc detected.");
                 return ManagedIdentitySource.AzureArc;
             }
+            else if (await ImdsV2ManagedIdentitySource.GetCsrMetadataAsync(_requestContext).ConfigureAwait(false))
```
neha-bhargava: Can this source be only detected by sending a request? Should it be part of this method to detect the source? Maybe we can move this under selecting between IMDS V2 or IMDS v1? This change breaks the public API to get the source.

neha-bhargava (changes requested):
Need to undo the changes to public API to get the managed identity source

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: nice, thanks

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

src/client/Microsoft.Identity.Client/PublicApi/net472/PublicAPI.Shipped.txt
Robbie-Microsoft: reverted

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: re-used in new `ImdsV2ManagedIdentitySource` file

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
Robbie-Microsoft: made a new method and made this one obsolete

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: yes and yes. Gladwin and Bogdan have signed off on this design.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpMessageHandler.cs
Robbie-Microsoft: It's needed. It's used in this new ImdsV2 source.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
bgavrilMS: I mean that the interface serves no purpose.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
bgavrilMS: It's still public?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: I'll be removing this when I make a mock handler for this request

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
bgavrilMS: Not resolved.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: What is the action item here then?

gladjohn: We need to support `ResourceId`, else it is a breaking change. As of now per MSI V2 design there is no breaking changes. As Bogdan mentioned this is a drop-in replacement for MSIv1

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
```diff
@@ -0,0 +1,71 @@
...
+        private CsrMetadata(CsrMetadataResponse response)
+        {
+            ClientId = response.ClientId;
+            TenantId = response.TenantId;
+            Cuid = response.Cuid;
+            AttestationEndpoint = response.AttestationEndpoint;
+        }
+
+        /// <summary>
+        /// Tries to create a CsrMetadata instance from a CsrMetadataResponse, logs warnings and returns null if any field is missing.
```
bgavrilMS: I think you'd want to throw an exception instead and bail instead? This is a server bug.

src/client/Microsoft.Identity.Client/IManagedIdentityApplication.cs
```diff
@@ -27,5 +28,11 @@ public interface IManagedIdentityApplication : IApplicationBase
         /// <see cref="AcquireTokenForManagedIdentityParameterBuilder.WithForceRefresh(bool)"/>
         /// </remarks>
         AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIdentity(string resource);
+
+        /// <summary>
+        /// Detects and returns the managed identity source available on the environment.
+        /// </summary>
+        /// <returns>Managed identity source detected on the environment if any.</returns>
+        Task<ManagedIdentitySource> GetManagedIdentitySourceAsync();
```
bgavrilMS: You can't add a method to a public interface, it's a breaking change.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,185 @@
...
```
bgavrilMS: But in general you want to avoid passing in "god objects" like RequestContext around. It violates the Single Responsability Principle. It's better to inject just the minimum info that this object needs, i.e. the logger (it doesn't need anything else)

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: So we should support `ObjectId` as well?

src/client/Microsoft.Identity.Client/IManagedIdentityApplication.cs
Robbie-Microsoft: reverted per our offline discussion

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: Resolving this per our offline discussion. Several items are needed from this requestContext: retry policy, http client, logger, etc.

global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "8.0.404",
+        "version": "9.0.0",
```
bgavrilMS: You probably don't want to check this in.

tests/Microsoft.Identity.Test.Unit/TestBase.cs
```diff
@@ -48,6 +60,15 @@ public virtual void TestInitialize()
         [TestCleanup]
         public virtual void TestCleanup()
         {
+            if (ManagedIdentityEnabled)
+            {
+                ManagedIdentityClient.s_sourceName = ManagedIdentitySource.None;
+                // set it null using reflection, since it is a private static field
+                typeof(Microsoft.Identity.Client.ManagedIdentity.ManagedIdentityClient)
```
bgavrilMS: no reflection pls.

tests/Microsoft.Identity.Test.Unit/TestBase.cs
```diff
@@ -48,6 +60,15 @@ public virtual void TestInitialize()
         [TestCleanup]
         public virtual void TestCleanup()
         {
+            if (ManagedIdentityEnabled)
```
bgavrilMS, line 63: not needed, this should be part of a "ResetForTest" API. @gladjohn - who is writing the ResetForTest API?

tests/Microsoft.Identity.Test.Unit/TestBase.cs
```diff
@@ -48,6 +60,15 @@ public virtual void TestInitialize()
...
+            {
+                ManagedIdentityClient.s_sourceName = ManagedIdentitySource.None;
```
bgavrilMS: There already exists a `TestCommon.ResetInternalStaticCaches()` which should do this all the time.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
```diff
@@ -368,7 +368,8 @@ public static MockHttpMessageHandler AddManagedIdentityMockHandler(
             string userAssignedId = null,
             UserAssignedIdentityId userAssignedIdentityId = UserAssignedIdentityId.None,
             HttpStatusCode statusCode = HttpStatusCode.OK,
-            string retryAfterHeader = null // A number of seconds (e.g., "120"), or an HTTP-date in RFC1123 format (e.g., "Fri, 19 Apr 2025 15:00:00 GMT")
```
bgavrilMS, line 371: This helper is too complex. Why not separate things out? HttpManager.AddMock ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,185 @@
...
+                _csrMetadata = csrMetadata;
+                return true;
+            }
+            else
+            {
+                return false;
+            }
+        }
+
+        private static Boolean ValidateCsrMetadataResponse(
```
bgavrilMS: `bool` not `Boolean`

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,185 @@
...
+using Microsoft.Identity.Client.Internal;
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal class ImdsV2ManagedIdentitySource : AbstractManagedIdentity
+    {
+        private const string CsrMetadataPath = "/metadata/identity/getPlatformMetadata";
+
+        private static CsrMetadata _csrMetadata;
```
bgavrilMS: You cannot cache _csrMetadata because it varies based on the request context?

bgavrilMS: @gladjohn - is the csr metadata cachable? if yes, by which key and for how long?

bgavrilMS: I think we should not cache it. Just rely on token caching.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,185 @@
...
+
+                default:
+                    requestContext.Logger.Info("[Managed Identity] ImdsV2 supports system-assigned managed identity.");
+                    break;
+            }
+
+            string queryParams = $"api-version={ImdsManagedIdentitySource.ImdsApiVersion}";
+            if (!string.IsNullOrEmpty(uami))
+            {
+                queryParams += $"&uaid={uami}";
```
bgavrilMS: This is not per spec https://microsoft.sharepoint.com/:w:/t/AzureMSI/EZj3qHCjswREvX3kephvJs0BhrAjQmmWiZ2gmtk1gaYOpg?e=ZQyD7w

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            Uri csrMetadataEndpoint = ImdsManagedIdentitySource.GetValidatedEndpoint(requestContext.Logger, CsrMetadataPath, queryParams);
+
+            var headers = new Dictionary<string, string>
+            {
+                { "Metadata", "true" },
+                { "x-ms-client-request-id", Guid.NewGuid().ToString() }
```
bgavrilMS: correlation ID is from the request context.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,185 @@
...
+            };
+
+            IRetryPolicyFactory retryPolicyFactory = requestContext.ServiceBundle.Config.RetryPolicyFactory;
+            IRetryPolicy retryPolicy = retryPolicyFactory.GetRetryPolicy(RequestType.ManagedIdentityDefault);
```
bgavrilMS: Should this be IMDS retry policy ?

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadataResponse.cs
```diff
@@ -0,0 +1,42 @@
...
+using JsonProperty = System.Text.Json.Serialization.JsonPropertyNameAttribute;
+#else
+using Microsoft.Identity.Json;
+#endif
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    [JsonObject]
+    [Preserve(AllMembers = true)]
+    internal class CsrMetadataResponse
```
bgavrilMS: Why do we need `CsrMetadata` and `CsrMetadataResponse` ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -19,38 +19,101 @@ internal class ManagedIdentityClient
...
+            }
+
+            return s_sourceName;
+        }
+
         // Detect managed identity source based on the availability of environment variables.
         // The result of this method is not cached because reading environment variables is cheap. 
         // This method is perf sensitive any changes should be benchmarked.
+        [Obsolete("Use GetManagedIdentitySourceAsync(RequestContext) instead.")]
         internal static ManagedIdentitySource GetManagedIdentitySource(ILoggerAdapter logger = null)
```
bgavrilMS: Logic from here needs to be consolidated with `GetManagedIdentitySourceAsync`

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
```diff
@@ -113,6 +113,15 @@ public static string GetBridgedHybridSpaTokenResponse(string spaAccountId)
             ",\"id_token_expires_in\":\"3600\"}";
         }
 
+        public static string GetCsrMetadataSuccessfulResponse()
+        {
+            return
+                "{\"client_id\":\"fake_client_id\"," +
+                "\"tenant_id\":\"fake_tenant_id\"," +
+                "\"CUID\":\"fake_CUID\"," +
```
bgavrilMS: Not per spec.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpMessageHandler.cs
bgavrilMS: No, I agree with @trwalke . Do not weaken existing tests. If you need explicit handling, handle it in the tests that require that handling. Explicit is better than implicit.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -19,38 +19,101 @@ internal class ManagedIdentityClient
     {
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
-        private readonly AbstractManagedIdentity _identitySource;
+        private static AbstractManagedIdentity s_identitySource;
```
bgavrilMS: You cannot cache this statically ! A new request will reuse this object, but a new request may have different params or it may come from a different ManagedIdentityApplication objects!!!

neha-bhargava: This will mess up tests for us and Azure Identity team

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: this try/catch is actually being used now, now just for unit testing

global.json
Robbie-Microsoft: oops

tests/Microsoft.Identity.Test.Unit/TestBase.cs
Robbie-Microsoft: removed in refactor

tests/Microsoft.Identity.Test.Unit/TestBase.cs
Robbie-Microsoft: removed in refactor

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManagerExtensions.cs
Robbie-Microsoft: removed in refactor

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: resolved in refactor

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: resolved in refactor

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: resolved in refactor. Thank you Bogdan for the explanation.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: Yes, it'd be a server error, but we only want to throw an exception if `probeMode` is set to `true`. We throw an exception in `ImdsV2ManagedIdentitySource` if null was returned and `probeMode` is set to `false`.

The way it is now, a bunch of warnings will at least be displayed to the dev so they know what part of the response was bad.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadataResponse.cs
gladjohn: `CsrMetadataResponse` should be sufficient for the next call to `/issuecredential`

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
gladjohn: This can be tricky. We need a modified IMDS retry policy, maybe a new `probe` policy. Because 404 should not be retried on the probe.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
gladjohn: > is the csr metadata cachable 

it is cachable, based on managed identity id. But it is better not to cache it because the endpoint also gives us the MAA attestation URL.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: per offline discussion: extend the IMDS retry policy and don't retry on 404

Robbie-Microsoft:
> Need to undo the changes to public API to get the managed identity source

@neha-bhargava: done, after refactoring some code

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpMessageHandler.cs
Robbie-Microsoft: resolved in latest commit

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: resolved

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: per offline discussion, we will not cache it.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadataResponse.cs
Robbie-Microsoft: resolved, just using `CsrMetadata` now.

tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHelpers.cs
Robbie-Microsoft: fixed

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: I'll make a new PR for this

Robbie-Microsoft: PR is located here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5419

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
```diff
@@ -0,0 +1,78 @@
...
+            }
+
+            return true;
+        }
+    }
+
+    /// <summary>
+    /// Represents VM unique Ids for CSR metadata.
+    /// </summary>
+    internal class CuidInfo
```
bgavrilMS: nit: add this up top?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,210 @@
...
+                    method: System.Net.Http.HttpMethod.Get,
+                    logger: requestContext.Logger,
+                    doNotThrow: false,
+                    mtlsCertificate: null,
+                    validateServerCertificate: null,
+                    cancellationToken: requestContext.UserCancellationToken,
+                    retryPolicy: retryPolicy)
+                .ConfigureAwait(false);
+            }
+            catch (Exception ex)
```
bgavrilMS, line 62: I believe @gladjohn said that probe failure occurs exclusively on 404?

bgavrilMS: (i.e. delete the catch block)

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,210 @@
...
+                }
+                else
+                {
+                    throwProbeFailedException(
+                        "ImdsV2ManagedIdentitySource.GetCsrMetadataAsync failed.",
+                        ex);
+                }
+            }
+
+            if (response.StatusCode != HttpStatusCode.OK)
```
bgavrilMS, line 77: @gladjohn to review this. I believe you said probe failure should only happen on 404

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,210 @@
...
+                    throwProbeFailedException(
+                        $"ImdsV2ManagedIdentitySource.GetCsrMetadataAsync failed due to HTTP error. Status code: {response.StatusCode} Body: {response.Body}",
+                        null,
+                        (int)response.StatusCode);
+                }
+            }
+
+            if (!ValidateCsrMetadataResponse(response, requestContext.Logger, probeMode))
+            {
+                return null;
```
bgavrilMS, line 95: This is not a good way to fail. Just throw exception like InvalidOperationException or smth

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+
+            return TryCreateCsrMetadata(response, requestContext.Logger, probeMode);
+        }
+
+        private static void throwProbeFailedException(
```
bgavrilMS: naming - CamelCase

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,210 @@
...
+             * Example:
+             * [
+             * "IMDS/150.870.65.1556",  // index 0: full match
+             * "1556"                   // index 1: captured group (\d+)
+             * ]
+             */
+            string serverHeader = response.HeadersAsDictionary.TryGetValue("server", out var value) ? value : null;
+            if (serverHeader == null)
+            {
+                if (probeMode)
```
bgavrilMS, line 134: Same comment. ProbeMode only look at 404?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,210 @@
...
+                } 
+            }
+
+            var match = System.Text.RegularExpressions.Regex.Match(
+                serverHeader,
+                @"^IMDS/\d+\.\d+\.\d+\.(\d+)$"
+            );
+            if (!match.Success || !int.TryParse(match.Groups[1].Value, out int version) || version <= 1324)
+            {
+                if (probeMode)
```
bgavrilMS, line 154: same comment

src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
```diff
@@ -62,6 +62,21 @@ public static bool RegionDiscovery(HttpResponse response, Exception exception)
...
+        /// </summary>
+        public static bool CsrMetadataProbe(HttpResponse response, Exception exception)
+        {
+            if (!Imds(response, exception))
+            {
+                return false;
+            }
+
+            // If Imds would retry but the status code is 404, don't retry
+            return (int)response.StatusCode is not 404;
```
neha-bhargava, line 77: So, it will retry on all the failures except for 404? This seems like a wide range. Or should it match the retry for all the status codes for IMDS and exclude the 404? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5359/files#diff-657c8db85ac31ecc939bedba01c7298d8b9f549926c0f52484b20e5a01cca01fR40

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,210 @@
...
+    internal class ImdsV2ManagedIdentitySource : AbstractManagedIdentity
+    {
+        private const string CsrMetadataPath = "/metadata/identity/getPlatformMetadata";
+
+        public static async Task<CsrMetadata> GetCsrMetadataAsync(
+            RequestContext requestContext,
+            bool probeMode)
+        {
+            string queryParams = $"cred-api-version={ImdsManagedIdentitySource.ImdsApiVersion}";
+            switch (requestContext.ServiceBundle.Config.ManagedIdentityId.IdType)
```
neha-bhargava: Can you reuse this block from the IMDS one? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5359/files#diff-7e3cbef86868afc50725e99493ee4aaeebd4ddaf17ac4ef0fde22d078f02848fR57

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -19,39 +19,72 @@ internal class ManagedIdentityClient
     {
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
-        private readonly AbstractManagedIdentity _identitySource;
+        internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
```
neha-bhargava, line 22: Making this static will mess up with the Azure Identity testing and we have seen this issue before. They do not have access to our internal methods as we do in our tests. Why are we switching to make this static?

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
gladjohn: IF probe returns 404 - then no /metadata endpoint, flows continues with other source detection, if none are found we fallback to IMDS (Kubernetes etc.)

We should pass a "." in the body to the endpoint and if it exists we will get a bad request (400), then we should inspect the header, spec [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/docs/msi_v2/vm_vmss_credential_probe.md) 

We should not throw anywhere if probe fails or if we do not find the endpoint. 

something like, 

```cs
        private bool EvaluateProbeResponse(HttpResponse response)
        {
            if (response == null)
            {
                _logger.Info("[Probe] No response received from the server.");
                return false;
            }

            _logger.Info($"[Probe] Evaluating response from credential endpoint. Status Code: {response.StatusCode}");

            if (response.HeadersAsDictionary.TryGetValue("Server", out string serverHeader) &&
                serverHeader.StartsWith(ImdsHeader, StringComparison.OrdinalIgnoreCase))
            {
                _logger.Info($"[Probe] Credential endpoint supported. Server: {serverHeader}");
                return true;
            }

            _logger.Info($"[Probe] Credential endpoint not supported. Status Code: {response.StatusCode}");
            return false;
        }
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: I had a separate PR, merged into this one, that made the ManagedIdentityClient be per ManagedIdentityApplication instead of per request, in order to get the Service Bundle into ImdsV2ManagedIdentitySource.

So as long as Azure Identity makes a new managed identity application for each test, they won't run into a caching issue.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: this is a great idea, thanks

src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
Robbie-Microsoft: > Or should it match the retry for all the status codes for IMDS and exclude the 404?

That's what this does. Just like RegionDiscovery, except this one will still retry on 408.

tests/Microsoft.Identity.Test.Common/TestCommon.cs
```diff
@@ -16,22 +16,23 @@
 using Microsoft.Identity.Client.Cache;
 using Microsoft.Identity.Client.Cache.Items;
 using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Http.Retry;
```
neha-bhargava, line 19: Unused?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/AppServiceTests.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Globalization;
 using System.Net;
+using System.Reflection;
```
neha-bhargava: Not needed anymore

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/AppServiceTests.cs
```diff
@@ -55,16 +56,25 @@ await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
...
             string endpoint,
             ManagedIdentitySource managedIdentitySource,
             ManagedIdentitySource expectedManagedIdentitySource)
         {
             using (new EnvVariableContext())
+            using (var httpManager = new MockHttpManager())
             {
                 SetUpgradeScenarioEnvironmentVariables(managedIdentitySource, endpoint);
 
-                Assert.AreEqual(expectedManagedIdentitySource, ManagedIdentityApplication.GetManagedIdentitySource());
```
neha-bhargava, line 67: I would keep this assert until this method is removed.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -55,13 +55,28 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
                 resource);
         }
 
+        /// <inheritdoc/>
+        public async Task<ManagedIdentitySource> GetManagedIdentitySourceAsync()
```
neha-bhargava, line 59: Why not keep this as static as well. The method to get the source was created static so that Azure Identity can detect if managed identity is available before creating the source. Was this discussed with them?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -0,0 +1,145 @@
...
+                    .WithHttpManager(httpManager)
+                    .WithRetryPolicyFactory(_testRetryPolicyFactory)
+                    .Build();
+
+                var miSource = await (managedIdentityApp as ManagedIdentityApplication).GetManagedIdentitySourceAsync().ConfigureAwait(false);
+                Assert.AreEqual(ManagedIdentitySource.ImdsV2, miSource);
+
+                const int NumRequests = 2; // initial probe request + 1 retry
+                int requestsMade = NumRequests - httpManager.QueueSize;
+                Assert.AreEqual(NumRequests, requestsMade);
```
neha-bhargava: In MSAL.Net the tests fail if there are unused mocks left in the queue. This is not needed.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -0,0 +1,145 @@
...
+                // Second attempt succeeds
+                httpManager.AddMockHandler(MockHelpers.MockCsrResponse());
+
+                var managedIdentityApp = ManagedIdentityApplicationBuilder.Create(ManagedIdentityId.SystemAssigned)
+                    .WithHttpManager(httpManager)
+                    .WithRetryPolicyFactory(_testRetryPolicyFactory)
+                    .Build();
+
+                var miSource = await (managedIdentityApp as ManagedIdentityApplication).GetManagedIdentitySourceAsync().ConfigureAwait(false);
+                Assert.AreEqual(ManagedIdentitySource.ImdsV2, miSource);
```
neha-bhargava, line 58: Another assert would be to validate the token source is IDP

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
neha-bhargava: Or you can simply validate that the QueueSize == 0

neha-bhargava: Same comment for all the tests that follow this logic.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
Robbie-Microsoft: *sigh* this is the first time I've heard that. CC @bgavrilMS what are your thoughts on this?

Robbie-Microsoft: @neha-bhargava, this method passes the Service Bundle through the `ManagedIdentityClient`'s `GetManagedIdentitySourceAsync` method, and into `ImdsV2ManagedIdentitySource`, where the csr metadata probe uses things from the Service Bundle: IdType, UserAssignedId, RetryPolicyFactory, HttpManager, Logger, etc.

Robbie-Microsoft: @gladjohn Can you confirm that this would work with Azure Identity? Did you discuss with them?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: resolving, this was a misunderstanding that we cleared up

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/AppServiceTests.cs
Robbie-Microsoft: The solution won't build if I leave it in, since it's marked as `obsolete`.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/AppServiceTests.cs
Robbie-Microsoft: done

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: It won't always throw an exception though, because of `probeMode`.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: What if a 500 comes back?

src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
```diff
@@ -62,6 +62,21 @@ public static bool RegionDiscovery(HttpResponse response, Exception exception)
             return (int)response.StatusCode is not (404 or 408);
         }
 
+        /// <summary>
+        /// Retry policy specific to CSR Metadata Probe.
+        /// Extends Imds retry policy but excludes 404 and 408 status codes.
```
gladjohn:
```suggestion
        /// Extends Imds retry policy but excludes 404 status code.
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -0,0 +1,202 @@
...
+            IRetryPolicyFactory retryPolicyFactory = requestContext.ServiceBundle.Config.RetryPolicyFactory;
+            IRetryPolicy retryPolicy = retryPolicyFactory.GetRetryPolicy(RequestType.CsrMetadataProbe);
+
+            // CSR metadata GET request
+            HttpResponse response = null;
+
+            try
+            {
+                response = await requestContext.ServiceBundle.HttpManager.SendRequestAsync(
+                    ImdsManagedIdentitySource.GetValidatedEndpoint(requestContext.Logger, CsrMetadataPath, queryParams),
```
gladjohn, line 50: we will never use the IMDS POD Identity endpoint in IMDS V2?

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
gladjohn: thought, we agreed to keep this static @Robbie-Microsoft.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: Just spoke with Gladwin. We'll skip this check in `probeMode`.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: correct

src/client/Microsoft.Identity.Client/Http/Retry/HttpRetryCondition.cs
Robbie-Microsoft: done

gladjohn (approved):

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
Robbie-Microsoft: The dev experience won't change for Azure SDK. They are still creating an instance of Managed Identity Application for their tests - they will be able to call this method from the `managedAidentityApplication`. You, @bgavrilMS and I can discuss over chat.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: @Robbie-Microsoft based on discussion with Azure Identity, this has to be a static helper API as @neha-bhargava mentioned, as this is the contract we have with Azure SDK

Issue #5255, opened 2025-04-25
Add correlation id to the error message when request_timeout exception is thrown


neha-bhargava:
Add correlation id to exception message so that in future the correlation id can be used for troubleshooting on the gateway

PR #5364, opened 2025-06-25
Ensure correlationId is added to request_timeout exception when available

trwalke:
Fixes #
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5255

Adds the correlation id when the request_timeout exception is thrown if it is available

**Testing**
Unit testing

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -14,6 +14,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client.Core;
 using Microsoft.Identity.Client.Http.Retry;
+using Microsoft.Identity.Client.OAuth2;
 
 namespace Microsoft.Identity.Client.Http
 {
@@ -134,9 +135,14 @@ namespace Microsoft.Identity.Client.Http
             logger.Warning("Request retry failed.");
             if (timeoutException != null)
             {
+                //If the correlation id is available, include it in the exception message
+                string correlationId = headers.ContainsKey(OAuth2Header.CorrelationId) ?
+                    $" CorrelationId: {headers[OAuth2Header.CorrelationId]}" :
+                    string.Empty;
+
                 throw new MsalServiceException(
                     MsalError.RequestTimeout,
-                    "Request to the endpoint timed out.",
+                    $"Request to the endpoint timed out.{correlationId}",
                     timeoutException);
             }
 
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -136,7 +136,7 @@ namespace Microsoft.Identity.Client.Http
             if (timeoutException != null)
             {
                 //If the correlation id is available, include it in the exception message
-                string msg = "Request to the endpoint timed out.";
+                string msg = MsalErrorMessage.RequestTimeOut;
 
                 if (headers != null && headers.Count > 0)
                 {
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -140,16 +140,25 @@ namespace Microsoft.Identity.Client.Http
 
                 if (headers != null && headers.Count > 0)
                 {
-                    string correlationId = headers.ContainsKey(OAuth2Header.CorrelationId) ?
-                                                $" CorrelationId: {headers[OAuth2Header.CorrelationId]}" :
+                    var correlationId = headers[OAuth2Header.CorrelationId];
+                    string correlationIdMsg = headers.ContainsKey(OAuth2Header.CorrelationId) ?
+                                                $" CorrelationId: {correlationId}" :
                                                 string.Empty;
-                    msg += correlationId;
+
+                    var ex = new MsalServiceException(
+                                MsalError.RequestTimeout,
+                                msg + correlationIdMsg,
+                                timeoutException);
+
+                    ex.CorrelationId = correlationId;
+
+                    throw ex;
                 }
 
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -134,9 +135,20 @@ public async Task<HttpResponse> SendRequestAsync(
             logger.Warning("Request retry failed.");
             if (timeoutException != null)
             {
+                //If the correlation id is available, include it in the exception message
+                string msg = "Request to the endpoint timed out.";
+
+                if (headers != null && headers.Count > 0)
+                {
+                    string correlationId = headers.ContainsKey(OAuth2Header.CorrelationId) ?
```
neha-bhargava: Do MSAL add correlation id to the headers?

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpManagerTests.cs
```diff
@@ -536,6 +537,56 @@ public async Task TestSendPostWithRetryOnTimeoutFailureAsync()
...
+                {
+                    httpManager.AddRequestTimeoutResponseMessageMockHandler(HttpMethod.Post);
+                }
+
+                Guid correlationId = Guid.NewGuid();
+                var headers = new Dictionary<string, string>();
+
+                if (addCorrelationId)
+                {
+                    headers.Add(OAuth2Header.CorrelationId, correlationId.ToString());
```
neha-bhargava, line 559: Can this be tested with a unit test with complete flow? That way we can also validate that when the token request goes the correlation id is there.

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
trwalke: Yes, see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/07f9c3bb0dcd32a69e7eaaa4cef9dca58fdd0495/src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs#L228

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -134,10 +135,30 @@ public async Task<HttpResponse> SendRequestAsync(
             logger.Warning("Request retry failed.");
             if (timeoutException != null)
             {
+                //If the correlation id is available, include it in the exception message
+                string msg = "Request to the endpoint timed out.";
```
gladjohn: let's add this to the MSALErrorMessage

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -134,10 +135,30 @@ public async Task<HttpResponse> SendRequestAsync(
...
+
+                if (headers != null && headers.Count > 0)
+                {
+                    var correlationId = headers[OAuth2Header.CorrelationId];
+                    string correlationIdMsg = headers.ContainsKey(OAuth2Header.CorrelationId) ?
```
gladjohn, line 144: should we do a tryGet? and also check to make sure the id is not empty or null?

gladjohn: you can write a test with the CorrelationId in the header and see how the code works

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -134,10 +135,30 @@ public async Task<HttpResponse> SendRequestAsync(
...
+                                                $" CorrelationId: {correlationId}" :
+                                                string.Empty;
+
+                    var ex = new MsalServiceException(
+                                MsalError.RequestTimeout,
+                                msg + correlationIdMsg,
```
gladjohn, line 150: suggest adding the " CorrelationId - " as part of the message here. Otherwise it is just an unknown guid.

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpManagerTests.cs
```diff
@@ -536,6 +537,95 @@ public async Task TestSendPostWithRetryOnTimeoutFailureAsync()
...
+                {
+                    httpManager.AddRequestTimeoutResponseMessageMockHandler(HttpMethod.Post);
+                }
+                Guid correlationId = Guid.NewGuid();
+
+                var app = ConfidentialClientApplicationBuilder
+                            .Create(TestConstants.ClientId)
+                            .WithAuthority(TestConstants.AuthorityTestTenant)
+                            .WithHttpManager(httpManager)
+                            .WithExperimentalFeatures(true)
```
gladjohn: why do we need the exp flag?

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
trwalke: The correlationId is always added to the headers so the test added is already testing that
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/07f9c3bb0dcd32a69e7eaaa4cef9dca58fdd0495/src/client/Microsoft.Identity.Client/OAuth2/OAuth2Client.cs#L228

src/client/Microsoft.Identity.Client/Http/HttpManager.cs
trwalke: It is added as part of line 145

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpManagerTests.cs
trwalke: no needed. will remove

gladjohn (approved):

PR #5365, opened 2025-06-26
MSI With MAA + KG - Credential Probe

gladjohn:
Fixes [3187132](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3187132)

Add probe logic to check if Credential source exists

**Testing**
unit tests 

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
@@ -152,8 +152,8 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
             await ResolveAuthorityAsync().ConfigureAwait(false);
 
-            ManagedIdentityClient managedIdentityClient = 
-                new ManagedIdentityClient(AuthenticationRequestParameters.RequestContext);
+            ManagedIdentityClient managedIdentityClient = await ManagedIdentityClient.CreateAsync(
+                AuthenticationRequestParameters.RequestContext, cancellationToken).ConfigureAwait(false);
 
             ManagedIdentityResponse managedIdentityResponse =
                 await managedIdentityClient
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialManagedIdentitySource.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Internal;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal class CredentialManagedIdentitySource : AbstractManagedIdentity
+    {
+        /// <summary>
+        /// Factory method to create an instance of `CredentialManagedIdentitySource`.
+        /// </summary>
+        public static AbstractManagedIdentity Create(RequestContext requestContext)
+        {
+            requestContext.Logger.Info(() => "[Managed Identity] Using credential based managed identity.");
+
+            return new CredentialManagedIdentitySource(requestContext);
+        }
+
+        private CredentialManagedIdentitySource(RequestContext requestContext) :
+            base(requestContext, ManagedIdentitySource.Credential)
+        {
+        }
+
+        /// <summary>
+        /// Even though the Credential flow does not use this request, we need to satisfy the abstract contract.
+        /// Return a minimal, valid ManagedIdentityRequest using the fixed credential endpoint.
+        /// </summary>
+        /// <param name="resource">The resource identifier (ignored in this flow).</param>
+        /// <returns>A ManagedIdentityRequest instance using the credential endpoint.</returns>
+        protected override ManagedIdentityRequest CreateRequest(string resource)
+        {
+            // Return a minimal request with the fixed credential endpoint.
+            return new ManagedIdentityRequest(
+                HttpMethod.Post,
+                new Uri("http://169.254.169.254/metadata/identity/issuecredential?cred-api-version=1.0"));
+        }
+    }
+}
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialProbeManager.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialProbeManager.cs
@@ -1,0 +1,112 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Http.Retry;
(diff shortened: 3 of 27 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialProbeManager.cs
```diff
@@ -0,0 +1,112 @@
...
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Http;
+using Microsoft.Identity.Client.Http.Retry;
+using Microsoft.Identity.Client.Internal;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal class CredentialProbeManager
+    {
+        private const string CredentialEndpoint = "http://169.254.169.254/metadata/identity/issuecredential";
```
gladjohn, line 19: Decided to go with the new endpoint for the probe as it is capable of returning `IMDS` in the header. We can always swap this to any other endpoint, because all we do is get a 400 and check the header.

trwalke:
Can you explain what MAA or KG is?
Is there a work item?

gladjohn:
> Can you explain what MAA or KG is? Is there a work item?

Added the KR link, which should have all the tasks. But maybe next week I can go over the new design.

src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialManagedIdentitySource.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Internal;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal class CredentialManagedIdentitySource : AbstractManagedIdentity
```
bgavrilMS, line 11: I'd call it ImdsV2 or smth, to make it clear that it's a repalcement for Imds.

bgavrilMS, line 11: I'm debating about the name. If we think the CredentialManagedIdentity is going to apply to source eventually, then this is ok. But the more I think about it, the more I doubt it. So maybe ImdsV2 is better name?

src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialManagedIdentitySource.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Even though the Credential flow does not use this request, we need to satisfy the abstract contract.
+        /// Return a minimal, valid ManagedIdentityRequest using the fixed credential endpoint.
+        /// </summary>
+        /// <param name="resource">The resource identifier (ignored in this flow).</param>
+        /// <returns>A ManagedIdentityRequest instance using the credential endpoint.</returns>
+        protected override ManagedIdentityRequest CreateRequest(string resource)
+        {
+            // Return a minimal request with the fixed credential endpoint.
```
bgavrilMS, line 36: Should you return null or a null-like object instead? Or throw an exception?

src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialProbeManager.cs
bgavrilMS: Pls add a TODO to remind of this

src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialProbeManager.cs
```diff
@@ -0,0 +1,112 @@
...
+            HttpContent httpContent = request.CreateHttpContent();
+
+            _logger.Info($"[Probe] Sending request to {CredentialEndpoint}");
+            _logger.Verbose(() => $"[Probe] Request Headers: {string.Join(", ", request.Headers)}");
+            _logger.Verbose(() => $"[Probe] Request Body: {ProbeBody}");
+
+            try
+            {
+                IRetryPolicyFactory retryPolicyFactory = _requestContext.ServiceBundle.Config.RetryPolicyFactory;
+                IRetryPolicy retryPolicy = retryPolicyFactory.GetRetryPolicy(RequestType.ManagedIdentityDefault);
```
bgavrilMS: Is this really the correct retry policy? Maybe add a TODO and capture this question.

src/client/Microsoft.Identity.Client/ManagedIdentity/CredentialProbeManager.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                _logger.Verbose(() => $"[Probe] Response Body: {response.Body}");
+            }
+        }
+
+        private bool EvaluateProbeResponse(HttpResponse response)
+        {
+            if (response == null)
+            {
+                _logger.Error("[Probe] No response received from the server.");
```
bgavrilMS: Do not use `Error` because it is part of the regular control flow. It confused app developers. It is a warning at best.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -20,30 +20,63 @@ internal class ManagedIdentityClient
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         private readonly AbstractManagedIdentity _identitySource;
+        // Cache for the managed identity source
+        private static ManagedIdentitySource? s_cachedManagedIdentitySource;
```
bgavrilMS, line 24: just a reminder that we will need to provide a `ResetInternalStateForTest` API to higher level SDKs / Azure SDK

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -53,6 +53,12 @@ public enum ManagedIdentitySource
...
         /// The source to acquire token for managed identity is Machine Learning Service.
         /// </summary>
-        MachineLearning
+        MachineLearning,
+
+        /// <summary>
+        /// Indicates that the source is credential endpoint based on the probe.
+        /// This is used to detect the new managed identity credential source.
+        /// </summary>
+        Credential
```
bgavrilMS, line 62: Same comment as above, maybe this should be called ImdsV2 ?

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -152,8 +152,8 @@ private async Task<AuthenticationResult> SendTokenRequestForManagedIdentityAsync
 
             await ResolveAuthorityAsync().ConfigureAwait(false);
 
-            ManagedIdentityClient managedIdentityClient = 
-                new ManagedIdentityClient(AuthenticationRequestParameters.RequestContext);
+            ManagedIdentityClient managedIdentityClient = await ManagedIdentityClient.CreateAsync(
```
bgavrilMS, line 155: So far, I only see the probe logic. Did you consider working directly on Robbie's branch which has the same?

gladjohn: This is just a prototype, not the final code Robbie is writing

Issue #5357, opened 2025-06-20
[Engineering task] Investigate flaky tests


gladjohn:
### Task type

Test

### Description

**This one fails in OneBranch Pipeline and not on PR pipeline.** 

Failed Kerberos_Interactive_AADAsync [2 m 23 s]
  Error Message:
   Test method Microsoft.Identity.Test.Integration.SeleniumTests.InteractiveFlowTests.Kerberos_Interactive_AADAsync threw exception: 
OpenQA.Selenium.WebDriverException: The HTTP request to the remote WebDriver server for URL http://localhost:52547/session timed out after 120 seconds. 

Other flaky tests : 

![Image](https://github.com/user-attachments/assets/f262cd64-b2fb-4a8a-bef2-cb118c59ffa4)

### Solution

fix the flaky tests

bgavrilMS:
- ignore Kerberos tests (you can even remove the feature, there is a work item for it)
- remove some of the weak flows like DeviceCodeFlow ... no need to test that so much

gladjohn:
how can we split this work @bgavrilMS. can we do another engineering excellence project?

gladjohn:
more tests 

![Image](https://github.com/user-attachments/assets/f05874c7-3fb4-4f01-b0ef-4e612e6bc80c)

gladjohn:
![Image](https://github.com/user-attachments/assets/84692140-ff0e-4468-9203-bb18e02d3e00)

bgavrilMS:
What's the root cause of all these flaky tests?

gladjohn:
> What's the root cause of all these flaky tests?

needs to be investigated

bgavrilMS:
This was done

PR #5367, opened 2025-06-27
Improve stability of Selenium tests

Avery-Dunn:
Attempts to resolve some of the flaky tests described in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5357

One of the most common failure occurs during interactive tests, occasionally there's a timeout when trying to create the web driver for Selenium:
```
Test method Microsoft.Identity.Test.Integration.SeleniumTests.InteractiveFlowTests.Interactive_AADAsync threw exception:
OpenQA.Selenium.WebDriverException: The HTTP request to the remote WebDriver server for URL http://localhost:52547/session timed out after 120 seconds.

Stack trace
at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
at OpenQA.Selenium.Remote.HttpCommandExecutor.<MakeHttpRequest>d__36.MoveNext()
--- End of stack trace from previous location where exception was thrown ---

// ...


at OpenQA.Selenium.Chrome.ChromeDriver..ctor(String chromeDriverDirectory, ChromeOptions options)
[at Microsoft.Identity.Test.Integration.Infrastructure.SeleniumExtensions.CreateDefaultWebDriver() in /_/tests/Microsoft.Identity.Test.Integration.NetCore/Infrastructure/SeleniumExtensions.cs:line 41](https://identitydivision.visualstudio.com/IDDP/_git/9a0dc2f2-2cc8-42ef-a2af-2b659b9c4530?path=%2F_%2Ftests%2FMicrosoft.Identity.Test.Integration.NetCore%2FInfrastructure%2FSeleniumExtensions.cs&version=GBmain&_a=contents&line=41&lineEnd=42&lineStartColumn=1&lineEndColumn=1&lineStyle=plain)
[at Microsoft.Identity.Test.Integration.Infrastructure.SeleniumWebUI.InitDriverAndGoToUrl(String url) in /_/tests/Microsoft.Identity.Test.Integration.NetCore/Infrastructure/SeleniumWebUI.cs:line 95](https://identitydivision.visualstudio.com/IDDP/_git/9a0dc2f2-2cc8-42ef-a2af-2b659b9c4530?path=%2F_%2Ftests%2FMicrosoft.Identity.Test.Integration.NetCore%2FInfrastructure%2FSeleniumWebUI.cs&version=GBmain&_a=contents&line=95&lineEnd=96&lineStartColumn=1&lineEndColumn=1&lineStyle=plain)
[at Microsoft.Identity.Test.Integration.Infrastructure.SeleniumWebUI.<SeleniumAcquireAuthAsync>d__10.MoveNext() in /_/tests/Microsoft.Identity.Test.Integration.NetCore/Infrastructure/SeleniumWebUI.cs:line 104](https://identitydivision.visualstudio.com/IDDP/_git/9a0dc2f2-2cc8-42ef-a2af-2b659b9c4530?path=%2F_%2Ftests%2FMicrosoft.Identity.Test.Integration.NetCore%2FInfrastructure%2FSeleniumWebUI.cs&version=GBmain&_a=contents&line=104&lineEnd=105&lineStartColumn=1&lineEndColumn=1&lineStyle=plain)
```

This PR attempts to improve the stability of the webdriver creation by implementing some simple retry behavior, and giving it a much shorter timeout period before it gives up and tries again. 

This should reduce the number of failed pipeline runs, and overall make the interactive tests run a bit quicker.

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/SeleniumExtensions.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/SeleniumExtensions.cs
@@ -22,44 +22,81 @@ namespace Microsoft.Identity.Test.Integration.Infrastructure
         private static readonly TimeSpan ExplicitTimespan = TimeSpan.FromSeconds(20);
         private static readonly TimeSpan ShortExplicitTimespan = TimeSpan.FromSeconds(5);
 
-        public static IWebDriver CreateDefaultWebDriver()
+        public static IWebDriver CreateDefaultWebDriver(int maxRetries = 3, int timeoutSeconds = 30)
         {
-            // ---------- Chrome launch flags ----------
-            var options = new ChromeOptions();
-            options.AddArguments(
-                "--headless=new",          // modern headless mode (remove this for debugging)
-                "--disable-gpu",
-                "--window-size=1920,1080",
-                "--remote-allow-origins=*",
-                "--disable-dev-shm-usage"); // avoids crashes in low-memory containers
 
-            // ---------- Pick a driver binary ----------
-            // 1) Prefer explicit env-var so devs can override locally
-            var driverDir = Environment.GetEnvironmentVariable("CHROMEWEBDRIVER");
-
-            // 2) Otherwise use the folder where CI drops a matching chromedriver
-            if (string.IsNullOrEmpty(driverDir))
+            for (int attempt = 1; attempt <= maxRetries; attempt++)
             {
-                driverDir = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "drivers");
+                try
+                {
+                    Trace.WriteLine($"[Selenium] Creating Chrome WebDriver (attempt {attempt}/{maxRetries})");
+
+                    // ---------- Chrome launch flags ----------
+                    var options = new ChromeOptions();
+                    options.AddArguments(
+                        "--headless=new",          // modern headless mode (remove this for debugging)
+                        "--disable-gpu",
+                        "--window-size=1920,1080",
+                        "--remote-allow-origins=*",
+                        "--disable-dev-shm-usage"); // avoids crashes in low-memory containers
+
+                    // Additional options to improve stability
+                    options.PageLoadStrategy = PageLoadStrategy.Eager;
+
+                    // ---------- Pick a driver binary ----------
+                    // 1) Prefer explicit env-var so devs can override locally
+                    var driverDir = Environment.GetEnvironmentVariable("CHROMEWEBDRIVER");
+
+                    // 2) Otherwise use the folder where CI drops a matching chromedriver
+                    if (string.IsNullOrEmpty(driverDir))
+                    {
+                        driverDir = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "drivers");
+                    }
+
+                    // 3) Fallback: let Selenium look on PATH
+                    ChromeDriverService service = string.IsNullOrEmpty(driverDir)
+                        ? ChromeDriverService.CreateDefaultService()
+                        : ChromeDriverService.CreateDefaultService(driverDir);
+
+                    service.HideCommandPromptWindow = true;
+                    service.EnableVerboseLogging = true;
+
+                    var driver = new ChromeDriver(
+                        service,
+                        options,
+                        TimeSpan.FromSeconds(timeoutSeconds));
+
+                    driver.Manage().Timeouts().ImplicitWait = ImplicitTimespan;
+
+                    try
+                    {
+                        driver.Navigate().GoToUrl("about:blank");
+                        // If this succeeds, the browser is responding
+                    }
+                    catch (Exception)
+                    {
+                        driver.Dispose();
+                        throw;
+                    }
+
+                    TryMaximize(driver);
+                    Trace.WriteLine($"[Selenium] Chrome WebDriver created successfully");
+                    return driver;
+                }
+                catch (Exception ex)
+                {
+                    if (attempt == maxRetries)
+                    {
+                        Trace.WriteLine($"[Selenium] Failed to create Chrome WebDriver after {maxRetries} attempts. Last error: {ex.Message}");
+                        throw;
+                    }
+
+                    Thread.Sleep(1000);
+                }
             }
 
-            // 3) Fallback: let Selenium look on PATH
-            ChromeDriverService service = string.IsNullOrEmpty(driverDir)
-                ? ChromeDriverService.CreateDefaultService()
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/SeleniumExtensions.cs
```diff
@@ -22,44 +22,81 @@ public static class SeleniumExtensions
...
-                "--disable-dev-shm-usage"); // avoids crashes in low-memory containers
-
-            // ---------- Pick a driver binary ----------
-            // 1) Prefer explicit env-var so devs can override locally
-            var driverDir = Environment.GetEnvironmentVariable("CHROMEWEBDRIVER");
-
-            // 2) Otherwise use the folder where CI drops a matching chromedriver
-            if (string.IsNullOrEmpty(driverDir))
+
+            for (int attempt = 1; attempt <= maxRetries; attempt++)
```
gladjohn, line 28: retries are good, but do we know why this fails in OneBranch and not on the PR builds?

gladjohn: And I am sure the build pipeline itself does a few retries before bailing

trwalke: There are already retries in the test run itself, it will typically retry about 2 or 3 times. So this will multiply that, causing loner delays if the endpoint is not available so we need to be mindful of that

trwalke: Yes. There are already retries in the test run itself, it will typically retry about 2 or 3 times. So this will multiply that, causing longer delays if the endpoint is not available so we need to be mindful of that.

trwalke: What is supposed to be listening on http://localhost:52547/session? the chrome driver? I wonder if the chrome driver process itself is stuck and needs to be restarted?

Avery-Dunn: > retries are good, but do we know why this fails in OneBranch and not on the PR builds?

I'm not sure, I couldn't figure out any consistency in the failures. Best I could find was a 5 year old SO post with the same problem but no clear reason: https://stackoverflow.com/questions/64500195/sporadic-selenium-chrome-webdriver-timeouts-starting-with-chrome-85

Avery-Dunn: > Yes. There are already retries in the test run itself, it will typically retry about 2 or 3 times. So this will multiply that, causing longer delays if the endpoint is not available so we need to be mindful of that.

I lowered the timeout from one 120 second attempt to three 30 second attempts, so even if all three retry attempts fail the tests will still fail faster than they used to.

Avery-Dunn: > What is supposed to be listening on http://localhost:52547/session? the chrome driver? I wonder if the chrome driver process itself is stuck and needs to be restarted?

I believe that is the random port that gets chosen for tests that use localhost, and is what we go to using the `webdriver.Navigate().GoToUrl()` call. These changes also test the webdriver by going to Chrome's `about:blank` URL to ensure that the webdriver is responsive, and retries it if it's not.

tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/SeleniumExtensions.cs
```diff
@@ -22,44 +22,81 @@ public static class SeleniumExtensions
...
+                    var options = new ChromeOptions();
+                    options.AddArguments(
+                        "--headless=new",          // modern headless mode (remove this for debugging)
+                        "--disable-gpu",
+                        "--window-size=1920,1080",
+                        "--remote-allow-origins=*",
+                        "--disable-dev-shm-usage"); // avoids crashes in low-memory containers
+
+                    // Additional options to improve stability
+                    options.PageLoadStrategy = PageLoadStrategy.Eager;
```
Robbie-Microsoft, line 44: Does all of this need to be in the try/catch? Could you move some of it out?

tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/SeleniumExtensions.cs
```diff
@@ -22,44 +22,81 @@ public static class SeleniumExtensions
...
 
-            driver.Manage().Timeouts().ImplicitWait = ImplicitTimespan;
+                    Thread.Sleep(1000);
+                }
+            }
 
-            TryMaximize(driver);
-            return driver;
+            // This should never be reached due to the throw in the catch block on the last attempt
+            throw new InvalidOperationException("Failed to create Chrome WebDriver");
```
Robbie-Microsoft, line 99: If it should never be reached, is it even needed?

tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/SeleniumExtensions.cs
Avery-Dunn: For simplicity and to help avoid any unintended changes I put pretty everything that used to be in this method into the try/catch. All of the config steps up until the creation of the ChromeDriver can probably be safe outside of it, but since it's just test infrastructure I don't think it matters that much

tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/SeleniumExtensions.cs
Avery-Dunn: All code paths in a method need to end with a return or a throw. An alternative would be to initialize the driver as "null" at the start of the method and return it here, but this seemed like a better way show something went wrong.

Robbie-Microsoft (approved):

neha-bhargava (approved):
Have you tried to run this PR through one branch build?

Avery-Dunn:
> Have you tried to run this PR through one branch build? If not, I would recommend running it before merging this PR.

Yeah, if you look at the run history around June 25-26 you'll see a bunch of manual runs by me. Here's an example, and after about 10 runs I didn't see the flaky tests fail: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1498688&view=results

PR #5369, opened 2025-07-01
msi with slc prototype using a demo app

gladjohn:
Fixes 

msi slc demo app 

**Testing**
using the app

**Performance impact**
none

**Documentation**

### DIFF
--- a/prototype/MsiWithSlc/MsiSlcPublicApis/MsiSlcPublicApis/Program.cs
+++ b/prototype/MsiWithSlc/MsiSlcPublicApis/MsiSlcPublicApis/Program.cs
@@ -1,0 +1,103 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.AppConfig;
+using Microsoft.IdentityModel.Abstractions;
+using System;
+using System.Security.Cryptography.X509Certificates;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+internal class Program
+{
+    private static async Task Main()
+    {
+        // ─── 0. Environment banner ────────────────────────────────────────────────
+        Console.WriteLine(new string('═', 70));
+        
+        // ─── 1. Certificate events ────────────────────────────────────────────────
+        ManagedIdentityApplication.BindingCertificateUpdated += c =>
+        {
+            Console.WriteLine($"\n[{Now()}] Event Fired -  BindingCertificateUpdated:");
+            PrintCertificate(c);
+        };
+
+        // ─── 2. Managed-identity source & certs ───────────────────────────────────
+        var source = await ManagedIdentityApplication.GetManagedIdentitySourceAsync()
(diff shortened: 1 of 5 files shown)


prototype/MsiWithSlc/MsiSlcPublicApis/MsiSlcPublicApis/Program.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        // ─── 0. Environment banner ────────────────────────────────────────────────
+        Console.WriteLine(new string('═', 70));
+        
+        // ─── 1. Certificate events ────────────────────────────────────────────────
+        ManagedIdentityApplication.BindingCertificateUpdated += c =>
+        {
+            Console.WriteLine($"\n[{Now()}] Event Fired -  BindingCertificateUpdated:");
+            PrintCertificate(c);
+        };
```
MZOLN, line 24: This is not the best approach as discussed yesterday. There are known challenges with events cleanup especially in the DI environment and not unsubscribing from this event may lead to multiple object lifetime issues. This sample doesn't demo clean up of this well.

I personally like an approach where we would utilize the method that .NET Core is using for options monitor (as we discussed yesterday on a call). Take a look here -> https://github.com/dotnet/dotnet/blob/ddf39a1b4690fbe23aea79c78da67004a5c31094/src/runtime/src/libraries/Microsoft.Extensions.Options/src/OptionsMonitor.cs#L111C13-L113C31

We could copy that approach and return a disposable object upon subscription. 
The benefits of that are that it can be now seemlesly integrated with Dispose method of any user component and is much cleaner than typical event unsubscribe "-=".

gladjohn: agreed, will implement that with new MSI v2 as discussed.

bgavrilMS (approved):

prototype/MsiWithSlc/MsiSlcPublicApis/MsiSlcPublicApis/Program.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        // ─── 4. Token-acquisition loop ───────────────────────────────────────────
+        string scope = "https://vault.azure.net/";
+        while (true)
+        {
+            Console.WriteLine($"\n[{Now()}] Acquiring token for scope → {scope}");
+            try
+            {
+                var result = await mi.AcquireTokenForManagedIdentity(scope)
+                                     //.WithProofOfPossession()
```
christothes, line 56: Is this intentionally commented out?

gladjohn: you can uncomment and test it when ESTS has the changes ready in PROD, as of now it is not.

christothes: So we are not actually doing anything with the cert unless we use `WithProofOfPossession`, correct? What value does the SLC endpoint provide in this case?

PR #5374, opened 2025-07-09
Delete flaky/deprecated Kerberos test

Robbie-Microsoft:
The Kerberos feature is being removed soon. This PR removes a flaky Kerberos test so that we can get the builds to pass.

See [here](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5357#issuecomment-2992520872) for more discussion.

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/InteractiveFlowTests.NetFwk.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/InteractiveFlowTests.NetFwk.cs
@@ -398,15 +398,6 @@ namespace Microsoft.Identity.Test.Integration.SeleniumTests
             }, TestContext);
         }
 
-        #region Azure AD Kerberos Feature Tests
-        [IgnoreOnOneBranch]
-        public async Task Kerberos_Interactive_AADAsync()
-        {
-            LabResponse labResponse = await LabUserHelper.GetDefaultUserAsync().ConfigureAwait(false);
-            await KerberosRunTestForUserAsync(labResponse, KerberosTicketContainer.IdToken).ConfigureAwait(false);
-            await KerberosRunTestForUserAsync(labResponse, KerberosTicketContainer.AccessToken).ConfigureAwait(false);
-        }
-
         private async Task<AuthenticationResult> KerberosRunTestForUserAsync(
             LabResponse labResponse,
             KerberosTicketContainer ticketContainer)


tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/InteractiveFlowTests.NetFwk.cs
```diff
@@ -398,15 +398,6 @@ private SeleniumWebUI CreateSeleniumCustomWebUIForDuende(string username, string
             }, TestContext);
         }
 
-        #region Azure AD Kerberos Feature Tests
-        [IgnoreOnOneBranch]
```
jennyf19, line 402: why isn't ignoreOnOneBranch enough to ignore the test?

tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/InteractiveFlowTests.NetFwk.cs
```diff
@@ -398,15 +398,6 @@ private SeleniumWebUI CreateSeleniumCustomWebUIForDuende(string username, string
             }, TestContext);
         }
 
-        #region Azure AD Kerberos Feature Tests
```
JoshLozensky, line 401: Did you intend to delete the region? If so there is an end region currently on line 507 that should also be removed.

Robbie-Microsoft: oops, that whole chunk should've been deleted. Thanks.

tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/InteractiveFlowTests.NetFwk.cs
```diff
@@ -397,122 +397,5 @@ private SeleniumWebUI CreateSeleniumCustomWebUIForDuende(string username, string
                 loginBtn?.Click();
             }, TestContext);
         }
-
-        #region Azure AD Kerberos Feature Tests
-        [IgnoreOnOneBranch]
```
pmaytak, line 402: nit: You can also remove this attribute, so this method will not be registered as a test, and keep the test code just in case it's needed, since Kerberos feature is still in MSAL. Once the feature is removed, all tests can be removed at the same time.

bgavrilMS: I'm ok with ignoring (e.g. `[Ignore]`) the test like @pmaytak stated and soft deprecating `[Obsolete("msg", false)]` the `WithKerberosTicketClaim` API. We have telemetry that shows that nobody is using it.

tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/InteractiveFlowTests.NetFwk.cs
pmaytak: Maybe the flag isn't being set now: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f665736e5a07a5fe6953441f390dabfcbbe9543d/tests/Microsoft.Identity.Test.Integration.netcore/Infrastructure/IgnoreOnOneBranch.cs#L17

tests/Microsoft.Identity.Test.Integration.netcore/SeleniumTests/InteractiveFlowTests.NetFwk.cs
Robbie-Microsoft: I think that should be a separate PR since I'll have to delete all 8 references to that method if I mark it as obsolete.

pmaytak (approved):

bgavrilMS (approved):

PR #5381, opened 2025-07-11
ManagedIdentityClient is now per ManagedIdentityApplication instead of per ManagedIdentityAuthRequest

Robbie-Microsoft:
This will be merged into the msiv2_probe PR... I made a separate PR for these changes because it's a refactor of ManagedIdentityClient that is unrelated to the CSR metadata probe.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
@@ -19,23 +19,23 @@ namespace Microsoft.Identity.Client.ManagedIdentity
     {
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
-        private AbstractManagedIdentity identitySource;
-        internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
+        private static AbstractManagedIdentity s_identitySource;
+        public static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
 
         internal async Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityAsync(
             RequestContext requestContext,
             AcquireTokenForManagedIdentityParameters parameters,
             CancellationToken cancellationToken)
         {
-            if (identitySource == null)
+            if (s_identitySource == null)
             {
                 using (requestContext.Logger.LogMethodDuration())
                 {
-                    identitySource = await SelectManagedIdentitySourceAsync(requestContext).ConfigureAwait(false);
+                    s_identitySource = await SelectManagedIdentitySourceAsync(requestContext).ConfigureAwait(false);
                 }
             }
 
-            return await identitySource.AuthenticateAsync(parameters, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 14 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -20,7 +20,7 @@ internal class ManagedIdentityClient
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         private AbstractManagedIdentity identitySource;
-        public static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
```
bgavrilMS: I think the source name should remain static, because there is no guarantee that ppl will use 1 ManagedIdentityApplication

Robbie-Microsoft: That's a great point. I forgot that that was our reasoning for making it static in msal-node. Thank you, I'll change it back.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -28,6 +24,8 @@ public sealed class ManagedIdentityApplication
         : ApplicationBase,
             IManagedIdentityApplication
     {
+        internal ManagedIdentityClient ManagedIdentityClient { get; }
```
Robbie-Microsoft, line 27: referenced on line 69 of this file

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: fixed.

neha-bhargava (approved):
Looks good. Thanks @Robbie-Microsoft

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -39,7 +39,7 @@ internal async Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityA
         }
 
         // This method tries to create managed identity source for different sources, if none is created then defaults to IMDS.
-        private static async Task<AbstractManagedIdentity> SelectManagedIdentitySourceAsync(RequestContext requestContext)
+        private async Task<AbstractManagedIdentity> SelectManagedIdentitySourceAsync(RequestContext requestContext)
```
bgavrilMS, line 42: This private method has no state. So it makes sense for it to be static. Is there a benefit for making this an instance method? Maybe `requestContext` should be part of the state?

gladjohn: agree, we should keep this static

gladjohn (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: It's an internal method, used only once in this file.

Robbie-Microsoft: I think the performance gain in keeping it static is negligible for this scenario.

PR #5399, opened 2025-07-21
Create bound_client_assertion.md

gladjohn:
Fixes - Adds a design to the bound assertion flows 

This pull request introduces a new feature to support PoP-bound client assertions in MSAL .NET. The changes include a new API that allows passing both a signed JWT and an X.509 certificate in a single call, ensuring compatibility with existing APIs while enabling mTLS certificate pinning and JWT forwarding. The update is additive and does not alter the behavior of existing flows.

### New Feature: PoP-Bound Client Assertion

* Added a new method `WithClientAssertion` to `ConfidentialClientApplicationBuilder` for registering an async delegate that returns a tuple containing a JWT and an X.509 certificate. This enables simultaneous handling of the signed JWT and certificate for mTLS scenarios.
* Introduced `CertificateAndAssertionDelegateCredential`, inheriting from `CertificateClientCredential`, to handle the new flow. It pins the certificate and sends the JWT as part of the request.

### Compatibility and Implementation Enhancements

* Ensured compatibility with `net462` by adding support for tuples via `System.ValueTuple`. Updated `Directory.Packages.props` and project files accordingly.
* Updated telemetry to include a new `JwtPop` branch in the `AssertionType` field while maintaining no behavioral changes for existing bearer-assertion or certificate-only flows.

### Documentation and Usage

* Added detailed documentation in

**Testing**
n/a

**Performance impact**
none

**Documentation**

### DIFF
--- a/docs/bound_client_assertion.md
+++ b/docs/bound_client_assertion.md
@@ -1,0 +1,111 @@
+# **PoP-Bound Client Assertion**  
+*(MEAv2 Phase-2 / GitHub #5398)*
+
+---
+
+## 1  Problem Statement
+Managed Identity (dSTS) and future Federated Identity Credential (FIC) flows issue a **first-leg JWT** that
+
+* already contains a `cnf` claim (public-key / thumbprint),
+* must be forwarded to Entra ID (`client_assertion`) using  
+  `client_assertion_type = urn:ietf:params:oauth:client-assertion-type:jwt-pop`,
+* **must** be presented **over an mTLS session** using the *same* X-509 certificate.
+
+MSAL .NET currently exposes **`WithCertificate()`** (mTLS) *and* a bearer-assertion overload; the two cannot coexist.  
+We need a single API that passes **both** the signed JWT *and* the certificate.
+
+---
+
+## 2  Design Goals
+| # | Goal |
+|---|------|
+| G-1 | **Single builder call** – provide JWT *and* cert together. |
+| G-2 | Async path (mirrors existing `Func<CancellationToken,Task<string>>`). |
+| G-3 | Use existing mTLS plumbing (`CertificateClientCredential` ➞ `MtlsPopAuthenticationOperation`). |
+| G-4 | Preserve binary compatibility; no changes to current APIs. |
+| G-5 | Work on **net462** without refactor (tuple via `System.ValueTuple`). |
+
+---
+
+## 3  Public API (additive)
+
+```csharp
+/// <summary>
+/// Registers an <b>async</b> delegate that returns a tuple containing
+///  • a PoP-bound client-assertion (JWT) and
+///  • the X-509 certificate referenced by that JWT’s <c>cnf</c> claim.
+/// The delegate receives a <see cref="CancellationToken"/> so callers can
+/// cancel external signing or network work.
+/// </summary>
+public ConfidentialClientApplicationBuilder WithClientAssertion(
+    Func<CancellationToken,
+         Task<(string Assertion, X509Certificate2 Certificate)>> boundAssertionAsync)
+```
+
+---
+
+## 4  Implementation
+
+### 4.1 Credential
+
+`CertificateAndAssertionDelegateCredential` : `CertificateClientCredential`
+
+```
+﻿┌─AddConfidentialClientParametersAsync───────────────┐
+│  (jwt, cert) = await _delegateAsync(ct);            │
+│  req.AuthenticationOperation = MtlsPop(cert);       │ pins cert
+│  req.MtlsCertificate         = cert;               │
+│  body: client_assertion_type = JwtPop               │ sends JWT
+│        client_assertion      = jwt                  │
+└──────────────────────────────────────────────────────┘
+```
+
+*inherits* mTLS certificate-pinning from base class.
+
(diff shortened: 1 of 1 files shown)


docs/bound_client_assertion.md
```diff
@@ -0,0 +1,111 @@
+# **PoP-Bound Client Assertion**  
+*(MEAv2 Phase-2 / GitHub #5398)*
+
+---
+
+## 1  Problem Statement
+Managed Identity (dSTS) and future Federated Identity Credential (FIC) flows issue a **first-leg JWT** that
```
bgavrilMS: Isn't this just FIC support? Not sure why dSTS is important here?

docs/bound_client_assertion.md
```diff
@@ -0,0 +1,111 @@
...
+```csharp
+/// <summary>
+/// Registers an <b>async</b> delegate that returns a tuple containing
+///  • a PoP-bound client-assertion (JWT) and
+///  • the X-509 certificate referenced by that JWT’s <c>cnf</c> claim.
+/// The delegate receives a <see cref="CancellationToken"/> so callers can
+/// cancel external signing or network work.
+/// </summary>
+public ConfidentialClientApplicationBuilder WithClientAssertion(
+    Func<CancellationToken,
```
bgavrilMS: Generally I agree with you, but the exact API for MSAL.NET needs some thought. We already have N `WithClientAssertion` APIs that are deprecated and the only one that we recommend using is `WithClientAssertion(Func<AssertionRequestOptions, string>)`

Changing the return from string to a tuple (string, X509Certificate2) is not an option, because a function overload cannot differ only by return value. 

So maybe the way to do it is to introduce `WithClientAssertion2<Func<AssertionRequestOptions, AssertionResponse>>`

docs/bound_client_assertion.md
```diff
@@ -0,0 +1,111 @@
...
+         Task<(string Assertion, X509Certificate2 Certificate)>> boundAssertionAsync)
+```
+
+---
+
+## 4  Implementation
```
bgavrilMS: Let's please keep these design docs free of low-level implementation details. Or add this as an appendix.

docs/bound_client_assertion.md
gladjohn: yes that would be ideal.

docs/bound_client_assertion.md
```diff
@@ -0,0 +1,41 @@
...
+public sealed class AssertionResponse
+{
+    public string            Assertion   { get; }            // required
+    public X509Certificate2? Certificate { get; }            // optional
+
+    // Future properties (AssertionType, FmiPath, ExtraClaims, …) can be added
+}
+
+public ConfidentialClientApplicationBuilder WithClientAssertion(
+    Func<AssertionRequestOptions,
```
trwalke, line 29: What would the `AssertionRequestOptions` contain?

gladjohn: this already exist today - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f3810f54b12907cd3e295f3900f05049c216d75a/src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs#L16

docs/bound_client_assertion.md
```diff
@@ -0,0 +1,41 @@
...
+2. *(Optional)* the X-509 certificate to which that JWT is bound (for mTLS PoP).
+
+---
+
+## Public API
+
+```csharp
+public sealed class AssertionResponse
+{
+    public string            Assertion   { get; }            // required
```
bgavrilMS: nit: if it is required, the ctor should initialize it. Or use `init` keyword.
nit2: setters are needed.

docs/bound_client_assertion.md
```diff
@@ -0,0 +1,41 @@
...
+    public X509Certificate2? Certificate { get; }            // optional
```
bgavrilMS: TokenBindingCertificate ?

docs/bound_client_assertion.md
gladjohn: we can use init, we may need to polyfil for netstd2.

docs/bound_client_assertion.md
```diff
@@ -0,0 +1,47 @@
...
+
+- MSAL pins the certificate to the mutual‑TLS channel.
+- MSAL sends the JWT with client_assertion_type = jwt-pop.
+
+If Certificate is null, 
+- MSAL sends the JWT as a bearer assertion (jwt-bearer).
+
+## Future Enhancements & mTLS Support
+
+- AssertionResponse is an extensible container; new assertion flavours (e.g., FMI‑bound tokens, custom claim sets) can surface via additional properties—no new builder overloads required.
```
trwalke:
```suggestion
- AssertionResponse is an extensible container; new assertion flavors (e.g., FMI‑bound tokens, custom claim sets) can surface via additional properties; no new builder overloads are required.
```
I am not familiar with using "-" between 2 seperate words. Im assuming it is a typo?
Anyway, it can be replaced with ";" which means ",(comma) and"

bgavrilMS (approved):

PR #5401, opened 2025-07-23
#5400 Fixing issue that leads to multiple active access tokens in the cache for non-tenanted oidc authority

### LINKED ISSUE #5400: [Bug] The cache contains multiple tokens satisfying the requirements
### Library version used

4.73.1

### .NET version

net8-windows

### Scenario

PublicClient - desktop app

The issue occurs when using non-tenanted OIDC authority (OpenIdDict in our case):
`builder.WithExperimentalFeatures().WithOidcAuthority(xxx)`.

When attempting to silently acquire an access token for the same scopes as before, after it was auto-renewed at least once (before the previous token has expired), we end up in a situation where we have two valid access tokens in cache.

After researching the code, we have noticed that when creating a token cache item for non-tenanted authority, the `tenantId` on the token cache item is set to an empty string:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f3810f54b12907cd3e295f3900f05049c216d75a/src/client/Microsoft.Identity.Client/Cache/Items/MsalAccessTokenCacheItem.cs#L141

While in `DeleteAccessTokensWithIntersectingScopes` method it looks for tokens that have `tenantId` strictly equal to `null` therefore not deleting the previous token which is still active for a few minutes.
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f3810f54b12907cd3e295f3900f05049c216d75a/src/client/Microsoft.Identity.Client/TokenCache.cs#L140-L141

The next time `AcquireTokenSilent` is called for the same authority/account/scope, the following exception will be thrown, which makes sense, since there will indeed be two active tokens in the cache:

```
Microsoft.Identity.Client.MsalClientException
  HResult=0x80131500
  Message=The cache contains multiple tokens satisfying the requirements. Try to clear token cache. 
  Source=Microsoft.Identity.Client
  StackTrace:
   at Microsoft.Identity.Client.TokenCache.GetSingleToken(List`1 tokenCacheItems, AuthenticationRequestParameters requestParams)
   at Microsoft.Identity.Client.TokenCache.<Microsoft-Identity-Client-ITokenCacheInternal-FindAccessTokenAsync>d__41.MoveNext()

  This exception was originally thrown at this call stack:
    Microsoft.Identity.Client.TokenCache.GetSingleToken(System.Collections.Generic.List<Microsoft.Identity.Client.Cache.Items.MsalAccessTokenCacheItem>, Microsoft.Identity.Client.Internal.Requests.AuthenticationRequestParameters)
    Microsoft.Identity.Client.TokenCache.Microsoft.Identity.Client.ITokenCacheInternal.FindAccessTokenAsync(Microsoft.Identity.Client.Internal.Requests.AuthenticationRequestParameters)
```

### Relevant code snippets

```csharp

```

### Expected behavior

Upon renewal of an access token for non-tenanted authority, the old access token should be removed from cache.

### Identity provider

Other

### Regression

_No response_

### Solution and workarounds

_No response_

andkorsh:
Fixes #5400 

During `DeleteAccessTokensWithIntersectingScopes` we now allow tenantId comparison to pass when either of the sides of condition is null or empty string.

**Testing**
When integrated with OpenIdDict, it simplifies testing if access token lifetime is set to a lower value than the default 30 minutes:
`.SetAccessTokenLifetime(TimeSpan.FromMinutes(2))`

**Performance impact**
None

**Documentation**
Just a bugfix, no doc update required.

### DIFF
--- a/prototype/KeyGuardMaa/Program.cs
+++ b/prototype/KeyGuardMaa/Program.cs
@@ -86,7 +86,7 @@ class Program
             byte[] sig = rsa.SignData(
                 System.Text.Encoding.UTF8.GetBytes("Hello KeyGuard"),
                 HashAlgorithmName.SHA256,
-                RSASignaturePadding.Pkcs1);
+                RSASignaturePadding.PSS);
 
             Console.WriteLine($"\nSignature length: {sig.Length} bytes");
         }


bgavrilMS (approved):

gladjohn:
PR build - https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1511407

trwalke (approved):

PR #5404, opened 2025-07-26
Bogavril/aui

bgavrilMS:
Fixes #

**Testing**

**Performance impact**

**Documentation**


tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/Agentic.cs
```diff
@@ -0,0 +1,139 @@
...
+        const string ClientId = "d15884b6-a447-4dd5-a5a5-a668c49f6300"; // agent app
+        const string TenantId = "31a58c3b-ae9c-4448-9e8f-e9e143e800df";
+        const string AgentIdentity = "d84da24a-2ea2-42b8-b5ab-8637ec208024";
+        const string UserUpn = "aui1@msidlabtoint.onmicrosoft.com";
+        private const string TokenExchangeUrl = "api://AzureADTokenExchange/.default";
+
+        [TestMethod]
+        public async Task AgentUserIdentityGetsTokenForGraphTest()
+        {
+            await AgentUserIdentityGetsTokenForGraphAsync().ConfigureAwait(false);
```
neha-bhargava, line 28: Why are these methods called twice? Maybe assert if the result was from IDP or cache if that is the reason.

bgavrilMS: Yes, good point, updated.

bgavrilMS: I added an AcquireTokenSilent option to the user fic call.

4gust (approved):

PR #5406, opened 2025-07-28
Update Program.cs

bgavrilMS:
Don't use PKCS1 padding, it's not secure. PSS is needed as per CodeQL


trwalke:
Should this also be done for the other instances of this in the solution?

PR #5427, opened 2025-08-06
Imdsv2: Generate CSR and execute CSR request

Robbie-Microsoft:
.NET native CSR generation via RSA and ASN.1 Encoding. Tested with net462, net472, net8.0, netstandard2.0


src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequestResponse.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.Identity.Client.Utils;
+using Microsoft.Identity.Json;
+#endif
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    /// <summary>
+    /// Represents the response for a Managed Identity CSR request.
+    /// </summary>
+    internal class CsrRequestResponse
```
gladjohn: there is no CSR Response in the MSI v2 design. CSR is something MSAL forms based on platform metadata

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -16,6 +17,7 @@ namespace Microsoft.Identity.Client.ManagedIdentity
     internal class ImdsV2ManagedIdentitySource : AbstractManagedIdentity
     {
         private const string CsrMetadataPath = "/metadata/identity/getPlatformMetadata";
+        private const string CsrRequestPath = "/metadata/identity/issuecredential";
```
gladjohn: this is issuecredential not csr request

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        /// <summary>
+        /// Generates a CSR for the given client, tenant, and CUID info.
+        /// </summary>
+        /// <param name="clientId">Managed Identity client_id.</param>
+        /// <param name="tenantId">AAD tenant_id.</param>
+        /// <param name="cuid">CuidInfo object containing VMID and VMSSID.</param>
+        /// <returns>CsrRequest containing the PEM CSR.</returns>
+        public static CsrRequest Generate(string clientId, string tenantId, CuidInfo cuid)
```
gladjohn: why is CUID an object of type CuidInfo? What's inside CuidInfo? We get CUID from metadata endpoint and we simply pass that in the CSR.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            if (string.IsNullOrWhiteSpace(clientId))
+                throw new ArgumentException("clientId must not be null or empty.", nameof(clientId));
+            if (string.IsNullOrWhiteSpace(tenantId))
+                throw new ArgumentException("tenantId must not be null or empty.", nameof(tenantId));
+            if (cuid == null)
+                throw new ArgumentNullException(nameof(cuid));
+            if (string.IsNullOrWhiteSpace(cuid.Vmid))
+                throw new ArgumentException("cuid.Vmid must not be null or empty.", nameof(cuid.Vmid));
```
gladjohn: we do not need to validate CUID, simply pass what we get

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
Robbie-Microsoft: How are you asking me what that is? You approved the PR I introduced it in.

`CuidInfo` is an object that contains the `Vmid` and `Vmssid` strings, which is received from the `getPlatformMetadata` endpoint

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: you are incorrect, according to the spec

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
gladjohn: I see what you mean, there were two response objects and I had asked you to remove one, looks like the wrong one got removed. But CUID is part of the metadata response, we simply get it and pass it to /issuecredential. I am really not sure what Vmid and Vmssid are, these are not part of the design.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
```diff
@@ -0,0 +1,439 @@
...
+            var rsa = RSA.Create();
+            rsa.KeySize = 2048;
+            return rsa;
+#endif
+        }
+
+        /// <summary>
+        /// Builds the CertificationRequestInfo structure containing subject, public key, and attributes.
+        /// </summary>
+        private static byte[] BuildCertificationRequestInfo(string clientId, string tenantId, CuidInfo cuid, RSA rsa)
```
gladjohn: We should use CertificateRequest where available, and a thin polyfill for old TFMs like net462 and netstd2

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
Robbie-Microsoft: They are part of the design. The Cuid is a JSON object that contains these fields. The object needs to be validated to ensure at least `Vmid` exists. If the VM is part of a VMSS, then `Vmssid` will exist too.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
Robbie-Microsoft: We need to validate it to ensure the response from `getPlatformMetadata` is valid.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequestResponse.cs
Robbie-Microsoft: There is. This is the response from request `issuecredential` request that sends the CSR.

gladjohn: we should simply call this `issuecredentialresponse` or `managedidentitycredentialresponse`

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
gladjohn: not resolved

Robbie-Microsoft: Do you mean the naming of the variable? I thought you meant the URL itself.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
gladjohn: maybe check to see if CUID is not null/empty. But MSAL does not need to know what value it holds.

Robbie-Microsoft: I'll change these from IsNullOrWhitespace to IsNullOrEmpty

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequestResponse.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
Robbie-Microsoft: `System.Security.Cryptography.X509Certificates.CertificateRequest` cannot be used because:
 
- It only supports X.509 extensions, not PKCS#10 attributes
- The CUID data must be embedded as an attribute (OID 1.2.840.113549.1.9.7) in the CSR attributes section
- This is a fundamental limitation that requires custom ASN.1 encoding

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
```diff
@@ -57,13 +57,12 @@ public CsrMetadata() { }
         /// Validates a JSON decoded CsrMetadata instance.
         /// </summary>
         /// <param name="csrMetadata">The CsrMetadata object.</param>
-        /// <returns>false if any field is null.</returns>
+        /// <returns>false if any required field is null. Note: Vmid is required, Vmssid is optional.</returns>
```
gladjohn: We should not check for anything inside CUID, i.e. no checks to see what CUID holds.

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// </summary>
+    internal class ClientCredentialRequestResponse
+    {
+        [JsonProperty("client_id")]
+        public string ClientId { get; set; }
+
+        [JsonProperty("tenant_id")]
+        public string TenantId { get; set; }
+
+        [JsonProperty("client_credential")]
```
gladjohn: this is the cert, maybe store it as byte[]

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: Raghavendra said to implement a null check.

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        [JsonProperty("expires_in")]
+        public int ExpiresIn { get; set; }
+
+        [JsonProperty("refresh_in")]
+        public int RefreshIn { get; set; }
+
+        public ClientCredentialRequestResponse() { }
+
+        public static bool ValidateCsrRequestResponse(ClientCredentialRequestResponse clientCredentialRequestResponse)
```
bgavrilMS: nit: naming `IsValid`

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -194,8 +196,80 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
+            };
+
+            var payload = new
+            {
+                pem = pem
+            };
+            var body = JsonHelper.SerializeToJson(payload);
+
+            IRetryPolicyFactory retryPolicyFactory = _requestContext.ServiceBundle.Config.RetryPolicyFactory;
+            IRetryPolicy retryPolicy = retryPolicyFactory.GetRetryPolicy(RequestType.Imds);
```
bgavrilMS, line 205: I continue to enjoy seeing this pattern. Very elegant.

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.Identity.Client.Utils;
+using Microsoft.Identity.Json;
+#endif
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    /// <summary>
+    /// Represents the response for a Managed Identity CSR request.
+    /// </summary>
+    internal class ClientCredentialRequestResponse
```
bgavrilMS: naming: `ClientCredential` is pretty overloaded term. Can you find a better name?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -16,6 +17,8 @@ namespace Microsoft.Identity.Client.ManagedIdentity
     internal class ImdsV2ManagedIdentitySource : AbstractManagedIdentity
     {
         private const string CsrMetadataPath = "/metadata/identity/getPlatformMetadata";
+        private const string ClientCredentialRequestPath = "/metadata/identity/issuecredential";
+        private const string AcquireEntraTokenPath = "/oauth2/v2.0/token";
```
bgavrilMS: not used?

bgavrilMS: Also, I think you need to reuse some existing componetns to call the token endpoint. Like TokenClient and such. @gladjohn - are you picking this one up?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
```diff
@@ -194,8 +196,80 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
+                    ManagedIdentitySource.ImdsV2,
+                    (int)response.StatusCode);
+            }
+
+            var clientCredentialRequestResponse = JsonHelper.DeserializeFromJson<ClientCredentialRequestResponse>(response.Body);
+            if (!ClientCredentialRequestResponse.ValidateCsrRequestResponse(clientCredentialRequestResponse))
+            {
+                throw MsalServiceExceptionFactory.CreateManagedIdentityException(
+                    MsalError.ManagedIdentityRequestFailed,
+                    $"[ImdsV2] ImdsV2ManagedIdentitySource.GetCsrMetadataAsync failed because the CsrMetadata response is invalid. Status code: {response.StatusCode} Body: {response.Body}",
```
bgavrilMS: I think you need to handle response.StatusCode separately from response.Body

bgavrilMS: And in case of 200 ok, you should probably not throw the full body in the exception message. It can have certificate and such.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
gladjohn: https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.certificaterequest.otherrequestattributes?view=net-9.0#system-security-cryptography-x509certificates-certificaterequest-otherrequestattributes

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
gladjohn: We need to use tokenclient for the probe as well. Let's discuss how we should proceed.

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
Robbie-Microsoft: `CertificateRequestResponse`?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: oops, this is for the child PR, not sure how this made its way into this one

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,439 @@
...
+#if NET462 || NET472
+            using (var sha256 = SHA256.Create())
+            {
+                byte[] hash = sha256.ComputeHash(certificationRequestInfo);
+                var formatter = new RSAPKCS1SignatureFormatter(rsa);
+                formatter.SetHashAlgorithm("SHA256");
+                return formatter.CreateSignature(hash);
+            }
+#else
+            return rsa.SignData(certificationRequestInfo, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
```
bgavrilMS: PKCS1 is not an allowed padding algorithm and you'll be flagged by Crypto Board static analysis rules. PSS is the only approved alogirhtm. If server side doesn't understand it, pls reach out to them to correct this.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,439 @@
...
+        {
+            Pem = pem ?? throw new ArgumentNullException(nameof(pem));
+        }
+
+        /// <summary>
+        /// Generates a CSR for the given client, tenant, and CUID info.
+        /// </summary>
+        /// <param name="clientId">Managed Identity client_id.</param>
+        /// <param name="tenantId">AAD tenant_id.</param>
+        /// <param name="cuid">CuidInfo object containing required VMID and optional VMSSID.</param>
```
bgavrilMS: VMSSID isn't used anywhere? except in tests?

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,439 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates a 2048-bit RSA key pair compatible with all target frameworks.
+        /// </summary>
+        private static RSA CreateRsaKeyPair()
+        {
+#if NET462 || NET472
+            var rsa = new RSACryptoServiceProvider(2048);
```
bgavrilMS: You need to use the strongest key on the machine i.e. a TPM key. This can be done later, but pls add a TODO here.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,439 @@
...
+
+        /// <summary>
+        /// Builds the X.500 Distinguished Name for the subject.
+        /// </summary>
+        private static byte[] BuildSubjectName(string clientId, string tenantId)
+        {
+            var rdnSequence = new System.Collections.Generic.List<byte[]>();
+
+            // CN=<clientId>
+            byte[] cnOid = EncodeAsn1ObjectIdentifier(new int[] { 2, 5, 4, 3 }); // commonName OID
```
bgavrilMS: I am not sure about owning this type of code. Is there a reference for it? e.g. if you copy it from the .NET FWK, we can at least add a link to it. The problem with this very low level code is that if a security issued is discovered, then we are not notified.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,439 @@
...
+            byte[] cnValue = EncodeAsn1Utf8String(clientId);
+            byte[] cnAttributeValue = EncodeAsn1Sequence(new[] { cnOid, cnValue });
+            rdnSequence.Add(EncodeAsn1Set(new[] { cnAttributeValue }));
+
+            // DC=<tenantId>
+            byte[] dcOid = EncodeAsn1ObjectIdentifier(new int[] { 0, 9, 2342, 19200300, 100, 1, 25 }); // domainComponent OID
```
bgavrilMS, line 120: What are all these magic numbers?

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
neha-bhargava: Can you remind me what does CSR stand for? Maybe CsrRequestResponse?

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
gladjohn: I am working on a e2e flow and that will bring in the key and the attestation info from the PR I opened a while back. Robbie can then replace the work he has been doing into that.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,439 @@
...
+            temp >>= 7;
+
+            while (temp > 0)
+            {
+                bytes.Insert(0, (byte)((temp & 0x7F) | 0x80));
+                temp >>= 7;
+            }
+
+            return bytes.ToArray();
+        }
```
bartonjs: All of these seem like they can/should use System.Formats.Asn1.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
neha-bhargava: If this is something documented can you add a comment here?

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrRequest.cs
Robbie-Microsoft: Resolved after discussion with Managed Crypto Discussion. Current approach is fine.

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
Robbie-Microsoft: Certificate Signing Request. In this case, `ClientCredential` is synonymous with `Certificate`, so we're going to go with my suggestion.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: original comments is resolved. Follow up comment is an action item for another PR.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: discussion to clear this up occurred offline

src/client/Microsoft.Identity.Client/ManagedIdentity/ClientCredentialRequestResponse.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: Resolved per offline discussion with the Managed Crypto Discussion group. There is no existing infrastructure for this in net 462, net472 and netstandard2.0. What I've written is the best way forward.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: This will actually require changes on MSAL's end. PSS will need to be custom implemented for the older versions of .net.

Robbie-Microsoft: For .NET 8.0 and newer frameworks:
- Use built-in RSASignaturePadding.Pss

For .NET Framework 4.6.2/4.7.2 and .NET Standard 2.0:
- Implement manual PSS padding using the PSS specification (RFC 3447)
- Use the RSA private key's raw signing capability
- Handle the PSS padding/unpadding manually

bartonjs: PSS works just fine in .NET Framework 4.6 and higher.  Unless you're using RSACryptoServiceProvider, which would mean "you're holding it wrong" (you might accidentally be getting an RSACSP.  If that's the case, either use `RSA.Create(int)` if the current OS is Windows, use `new RSACng()` instead of RSA.Create()).

You absolutely shouldn't be implementing PSS padding yourself.

RSA-PSS isn't very common in X.509 land.  Before using PSS here because "you're supposed to", you might want to verify that your CA will even correctly process it

Robbie-Microsoft: @bartonjs I think `RSA.Create(int)` is not available in the earlier versions of .net?

Robbie-Microsoft: I did some brief pair-programming with @bartonjs to resolve this issue.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: Looks like this is not available in .NET Framework 462/472 and .NET Standard 2.0. Do you think it's worth it to conditionally use `System.Formats.Asn1` for only 8.0? Or would it be clearer to use the same code for all 4 versions (how it is now).

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: added TODO

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
bartonjs: It's available, you just need to reference the System.Formats.Asn1 NuGet package.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
bgavrilMS: Can you keep the naming consistent? Is this a CsrResponse?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CsrValidator.cs
```diff
@@ -0,0 +1,431 @@
...
+using Microsoft.Identity.Client.ManagedIdentity;
+using Microsoft.Identity.Client.Utils;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
+{
+    /// <summary>
+    /// Test helper to expose CsrValidator methods for testing malformed PEM.
+    /// </summary>
+    internal static class TestCsrValidator
```
bgavrilMS: I can't read any of these tests. At least add a commen on what this does.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -130,5 +131,100 @@ public async Task GetCsrMetadataAsyncFails404WhichIsNonRetriableAndRetryPolicyIs
...
+                Vmssid = vmssid
+            };
+
+            // Should succeed since Vmssid is optional (Vmid is provided and valid)
+            var csrRequest = Csr.Generate(TestConstants.ClientId, TestConstants.TenantId, cuid);
+            Assert.IsNotNull(csrRequest);
+            Assert.IsFalse(string.IsNullOrWhiteSpace(csrRequest.Pem));
+
+            // Validate the CSR contents - this should handle null/empty VMSSID gracefully
+            CsrValidator.ValidateCsrContent(csrRequest.Pem, TestConstants.ClientId, TestConstants.TenantId, cuid);
```
bgavrilMS: I'm not clear what exactly you are testing here. Ideally we would test that the CSR is valid and that it contains all the necessary data (e.g. oid=clientId etc.)

bgavrilMS: (or maybe you are doing this, but I am not able to tell)

bgavrilMS: As a test idea, you can have a  test that runs on net8 only and which uses the rutnime to create the CSR and then compare it with yours. Just an idea, not sure how useful it would be.

src/client/Microsoft.Identity.Client/ManagedIdentity/CsrMetadata.cs
Robbie-Microsoft: It is consistent, and there's no csrResponse - that was renamed to certificateResponse. This is named correctly.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CsrValidator.cs
Robbie-Microsoft: There are no tests in this file. The whole purpose of this file is to decode the PEM so it's values can be validated.

Robbie-Microsoft: I feel that the class summary does a good job explaining what this file is for

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: The purpose of the CsrValidator is to decode the PEM for unit tests. It's too low level to really focus on.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,482 @@
...
+        /// </summary>
+        private static byte[] BuildAttributes(CuidInfo cuid)
+        {
+            var attributes = new List<byte[]>();
+
+            // CUID attribute (OID 1.2.840.113549.1.9.7)
+            // Serialize CuidInfo as JSON object string using existing JSON serialization
+            byte[] cuidOid = EncodeAsn1ObjectIdentifier(new int[] { 1, 2, 840, 113549, 1, 9, 7 });
+            string cuidValue = JsonHelper.SerializeToJson(cuid);
+            byte[] cuidData = EncodeAsn1PrintableString(cuidValue);
```
bartonjs: Assuming `cuidValue` is JSON (previous line) it's illegal as a PrintableString value (PrintableString is defined as restricted to the characters `ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789 '()+,-./:=?`, which doesn't include `{` or `}`).  "Oh, but it works" you say, that just means you are sending it to a lax reader, and you're signing up for a compatibility nightmare.

You probably want UTF8String.  And, as previously mentioned, to be using AsnWriter, not bit-banging it out yourself.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
```diff
@@ -0,0 +1,482 @@
...
+        /// Builds the CertificationRequestInfo structure containing subject, public key, and attributes.
+        /// </summary>
+        private static byte[] BuildCertificationRequestInfo(string clientId, string tenantId, CuidInfo cuid, RSA rsa)
+        {
+            var components = new List<byte[]>();
+
+            // Version (INTEGER 0)
+            components.Add(EncodeAsn1Integer(new byte[] { 0x00 }));
+
+            // Subject: CN=<clientId>, DC=<tenantId>
```
bartonjs: Your comment here implies you want the subject name ToString() to be CN-first.  That actually requires you encode CN last, because the display string is prevented in stack order, instead of queue order.

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: I've addressed this in #5449

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: I've addressed this in #5449

src/client/Microsoft.Identity.Client/ManagedIdentity/Csr.cs
Robbie-Microsoft: I've addressed this in #5449

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        [JsonProperty("expires_in")]
+        public int ExpiresIn { get; set; }
+
+        [JsonProperty("refresh_in")]
+        public int RefreshIn { get; set; }
+
+        public CertificateRequestResponse() { }
+
+        public static bool IsValid(CertificateRequestResponse certificateRequestResponse)
```
bgavrilMS: Maybe you want to do `void Validate()` and throw a more meaningful exception instead?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            if (string.IsNullOrEmpty(certificateRequestResponse.ClientId) ||
+                string.IsNullOrEmpty(certificateRequestResponse.TenantId) ||
+                string.IsNullOrEmpty(certificateRequestResponse.ClientCredential) ||
+                string.IsNullOrEmpty(certificateRequestResponse.RegionalTokenUrl) ||
+                certificateRequestResponse.ExpiresIn <= 0 ||
+                certificateRequestResponse.RefreshIn <= 0)
```
bgavrilMS: I'm pretty sure it won't have refresh in at first, so don't block on that.

Robbie-Microsoft: This is the first I'm hearing about this. Is `refreshIn` the only thing that'll be missing?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
bgavrilMS: Let's assume it is.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
```diff
@@ -0,0 +1,56 @@
...
+        public string ClientCredential { get; set; }
+
+        [JsonProperty("regional_token_url")]
+        public string RegionalTokenUrl { get; set; }
+
+        [JsonProperty("expires_in")]
+        public int ExpiresIn { get; set; }
+
+        [JsonProperty("refresh_in")]
+        public int RefreshIn { get; set; }
```
bgavrilMS: this is optional I think, should it be int?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
Robbie-Microsoft: Nidhi just now told me neither `refreshIn` or `expiresIn` will be returned

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
```diff
@@ -0,0 +1,51 @@
...
+                    rsa,
+                    HashAlgorithmName.SHA256,
+                    RSASignaturePadding.Pss);
+
+                AsnWriter writer = new AsnWriter(AsnEncodingRules.DER);
+                writer.WriteCharacterString(UniversalTagNumber.UTF8String, JsonHelper.SerializeToJson(cuid));
+
+                req.OtherRequestAttributes.Add(
+                    new AsnEncodedData(
+                        "1.2.840.113549.1.9.7",
```
gladjohn: Nidhi mentioned this value is wrong, we may need to check with her on the final string

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -128,12 +130,14 @@ private static bool ValidateCsrMetadataResponse(
              * "1556"                   // index 1: captured group (\d+)
              * ]
              */
-            string serverHeader = response.HeadersAsDictionary.TryGetValue("server", out var value) ? value : null;
+            // Imds bug: headers are missing
+            // TODO: uncomment this when the bug is fixed
```
gladjohn: do we have a bug to IMDS team?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -194,8 +198,77 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
+                    ex,
+                    ManagedIdentitySource.ImdsV2,
+                    (int)response.StatusCode);
+            }
+
+            if (response.StatusCode != HttpStatusCode.OK)
+            {
+                throw MsalServiceExceptionFactory.CreateManagedIdentityException(
+                    MsalError.ManagedIdentityRequestFailed,
+                    $"[ImdsV2] ImdsV2ManagedIdentitySource.ExecuteCertificateRequestAsync failed due to HTTP error. Status code: {response.StatusCode} Body: {response.Body}",
```
gladjohn, line 238: can we use the existing helper `GetMessageFromErrorResponse` to form the message here?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -194,8 +198,77 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
+            var certificateRequestResponse = JsonHelper.DeserializeFromJson<CertificateRequestResponse>(response.Body);
+            CertificateRequestResponse.Validate(certificateRequestResponse);
+
+            return certificateRequestResponse;
+        }
+
         protected override ManagedIdentityRequest CreateRequest(string resource)
         {
+            var csrMetadata = GetCsrMetadataAsync(_requestContext, false).GetAwaiter().GetResult();
+            var csrPem = Csr.Generate(csrMetadata.ClientId, csrMetadata.TenantId, csrMetadata.CuId);
```
gladjohn: we should also pass the key (RSA) here

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Identity.Client.PlatformsCommon.Shared;
 using System.IO;
 using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.ManagedIdentity.V2;
```
gladjohn, line 12: not needed?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
```diff
@@ -0,0 +1,235 @@
...
+            {
+                RSAParameters rsaParameters = _rsa.ExportParameters(false);
+                writer.WriteIntegerUnsigned(rsaParameters.Modulus);
+                writer.WriteIntegerUnsigned(rsaParameters.Exponent);
+            }
+
+            byte[] publicKey = writer.Encode();
+            writer.Reset();
+
+            // CertificationRequestInfo ::= SEQUENCE {
```
gladjohn, line 89: not needed?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -130,5 +136,39 @@ public async Task GetCsrMetadataAsyncFails404WhichIsNonRetriableAndRetryPolicyIs
...
             }
         }
+
+        [TestMethod]
+        public void TestCsrGeneration()
+        {
+            var cuid = new CuidInfo
+            {
+                VmId = TestConstants.VmId,
+                VmssId = TestConstants.VmssId
```
gladjohn, line 154: is there a test without vmssid?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -130,5 +136,39 @@ public async Task GetCsrMetadataAsyncFails404WhichIsNonRetriableAndRetryPolicyIs
...
+
+            // Generate CSR
+            var csrPem = Csr.Generate(TestConstants.ClientId, TestConstants.TenantId, cuid);
+
+            // Validate the CSR contents using the helper
+            CsrValidator.ValidateCsrContent(csrPem, TestConstants.ClientId, TestConstants.TenantId, cuid);
+        }
+
+        [TestMethod]
+        public void TestCsrGeneration_MalformedPem_FormatException()
```
gladjohn, line 162: Can we also add a positive tests for ParseCsrFromPem on well-formed input?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
Robbie-Microsoft: deleted, per spec

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequestResponse.cs
Robbie-Microsoft: deleted, per spec

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: That's being done in the follow up PR that create the mTLS cert

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: it is. ImdsV2 probe referenced in this file

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
Robbie-Microsoft: they're clarifying-comments

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: no, but I can remove from all and just make one vmId + vmssId test

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: Nidhi is aware

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
Robbie-Microsoft: It's now 1.3.6.1.4.1.311.90.2.10

Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -194,9 +189,90 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
+
+            var certificateRequestResponse = JsonHelper.DeserializeFromJson<CertificateRequestResponse>(response.Body);
+            CertificateRequestResponse.Validate(certificateRequestResponse);
+
+            return certificateRequestResponse;
+        }
+
         protected override ManagedIdentityRequest CreateRequest(string resource)
         {
+            var csrMetadata = GetCsrMetadataAsync(_requestContext, false).GetAwaiter().GetResult();
```
gladjohn: please consider refactoring this part to use async/await.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -194,9 +189,90 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
...
+            var csr = Csr.Generate(csrMetadata.ClientId, csrMetadata.TenantId, csrMetadata.CuId);
+
+            var certificateRequestResponse = ExecuteCertificateRequestAsync(csr).GetAwaiter().GetResult();
```
gladjohn: same comment as above - please consider refactoring this part to use async/await.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: maybe it got missed? I do not see the probe code

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
gladjohn: still not updated, maybe a bad rebase? 

image

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CsrMetadata.cs
```diff
@@ -57,13 +57,12 @@ public CsrMetadata() { }
...
-        /// <returns>false if any field is null.</returns>
+        /// <returns>false if any required field is null. Note: VmId is required, VmssId is optional.</returns>
         public static bool ValidateCsrMetadata(CsrMetadata csrMetadata)
         {
             if (csrMetadata == null ||
-                csrMetadata.Cuid == null ||
-                string.IsNullOrEmpty(csrMetadata.Cuid.Vmid) ||
-                string.IsNullOrEmpty(csrMetadata.Cuid.Vmssid) ||
+                csrMetadata.CuId == null ||
+                string.IsNullOrEmpty(csrMetadata.CuId.VmId) ||
```
gladjohn: VMID is not a required field

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: I can't without changing the base method, which is used by every other MI source

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: `ImdsV2ManagedIdentitySource.GetCsrMetadataAsync` line 71.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
Robbie-Microsoft: Haven't pushed latest code yet.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: It was a fair comment @Robbie-Microsoft . The diff shows that you only added a using statement. YOu can't expect people to read code that is not in the diff.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
bgavrilMS: +1 to this. No sync over async.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
bgavrilMS: Can you please point reviewers to that test.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
```diff
@@ -0,0 +1,235 @@
...
+using System;
+using System.Collections.ObjectModel;
+using System.Formats.Asn1;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    internal class CertificateRequest
```
bartonjs, line 13: I don't see this file being conditionally included in a csproj.  If the intent is to be a downlevel polyfill (and use the platform version when it has support for OtherRequestAttributes), then you should wrap the class (or the whole file) in `#if !NET7_0_OR_GREATER`

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: please see this draft PR - in the [MI executor](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5438/files#diff-c173d517556b1e1f07ea7c662830e81eb30fea2ca39c4ad5e7ace1849cf623f0) we can make a decision based on the [probe outcome](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5438/files#diff-ccb7c2c1e90d10b0b0a643086d9e7986d32c6c875ea8f90d93f1b5849ec44708), if it is MSI v2 path then we split from here 

And create a [CredentialBasedManagedIdentity](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5438/files#diff-2834e86de4f69e687e7c58766dd0e095c16e5d5280e8a37fc30066ad580bf174). The existing one can be renamed to `LegacyManagedIdentityAuthRequest` or something and extract common functionality to a new base. 

The north star for all services will be to move towards CredentialBasedManagedIdentity

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
gladjohn: that's a good idea. or you could also use the PEM and create a self signed cert.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CsrMetadata.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: Please show me how you would make this change

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: The existing TestCsrGeneration tests cover this - `CsrValidator.ValidateCsrContent` in that test calls `CsrValidator.ParseCsrFromPem`.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
Robbie-Microsoft: We decided to use the "polyfill" as the main implementation. We thought it would be easier to maintain for all versions of net that way (462, 472, netstandard2.0, 8.0). Do you recommend making this a true polyfill and using existing net8.0 where possible? A concern we have is that we'd have to have many versions of the same unit test.

bartonjs: > Do you recommend making this a true polyfill and using existing net8.0 where possible?

I do, mostly on general principles.  For one thing, as time goes on, maybe you want to add new algorithms... and maybe you can get away with only caring about those new algorithms on .NET 8+... cool, you didn't need to edit your polyfill at all.

> A concern we have is that we'd have to have many versions of the same unit test.

I'm not sure I follow. If you're claiming you work on various versions presumably you're testing on various versions already, and really you want this to be an implementation detail.  "Do the tests pass on .NET Framework? Do the tests pass on .NET 8+?"  That's really all that matters.

Since you're already split compiling, removing the class on runtimes where it's redundant will make your DLL smaller, if nothing else.  Sure, it's only _barely_ smaller, but depending on how much polyfill code you end up accumulating, it could be death by a thousand papercuts.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
gladjohn: Thanks @Robbie-Microsoft! Since `ValidateCsrContent` calls `ParseCsrFromPem`, coverage is indirect. Let’s also add a focused positive test that calls `ParseCsrFromPem` directly with a well-formed CSR and asserts parsed fields (subject CN, SANs if present, pub key alg/size, and signature alg, etc). That way failures in parsing won’t be masked by the outer validator.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: refer to [this ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f0a02aa1021020d19c82ce38de062fd3bfc00e0a/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs#L108)

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: resolved offline. we will retest later and add this back for now.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: closing this per offline discussion

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: closing this per offline discussion

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: per offline discussion, we're changing the CreateRequest signature to an async mthod

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CsrMetadata.cs
```diff
@@ -7,18 +7,24 @@
...
 
-        [JsonProperty("vmssid")]
-        public string Vmssid { get; set; }
+        [JsonProperty("vmssId")]
+        public string VmssId { get; set; }
+
+        public static bool IsNullOrEmpty(CuidInfo cuidInfo)
+        {
+            return cuidInfo == null ||
+                   (string.IsNullOrEmpty(cuidInfo.VmId) && string.IsNullOrEmpty(cuidInfo.VmssId));
```
gladjohn, line 26: do we need the vmid / vmssid validations?

Robbie-Microsoft: Yes, this ensures Cuid is not empty

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
Robbie-Microsoft: Thank you for that feedback. We're creating a work item for this and it will be it's own PR, to be published soon.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
```diff
@@ -0,0 +1,235 @@
...
+                using (writer.PushSequence())
+                {
+                    if (_rsaPadding == RSASignaturePadding.Pss)
+                    {
+                        if (_hashAlgorithmName != HashAlgorithmName.SHA256)
+                        {
+                            throw new NotSupportedException("Only SHA256 is supported with PSS padding.");
+                        }
+
+                        writer.WriteObjectIdentifier("1.2.840.113549.1.1.10");
```
gladjohn, line 171: the OID to be used for CUID is 1.3.6.1.4.1.311.90.2.10, looks like other places have been updated, do we need to update it here? Is there a test that checks for OID?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CsrValidator.cs
```diff
@@ -0,0 +1,144 @@
...
+                        }
+                    }
+                    Assert.IsTrue(foundCuid, "CUID (challengePassword) attribute not found");
+                }
+            }
+
+            // signatureAlgorithm
+            var sigAlgSeq = csrSequence.ReadSequence();
+            string sigAlgOid = sigAlgSeq.ReadObjectIdentifier();
+            Assert.AreEqual("1.2.840.113549.1.1.10", sigAlgOid, "Signature algorithm is not RSASSA-PSS (SHA256withRSA/PSS)");
```
gladjohn, line 136: `1.2.840.113549.1.1.10` this needs to be updated?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CsrMetadata.cs
```diff
@@ -57,13 +63,11 @@ public CsrMetadata() { }
         /// Validates a JSON decoded CsrMetadata instance.
         /// </summary>
         /// <param name="csrMetadata">The CsrMetadata object.</param>
-        /// <returns>false if any field is null.</returns>
+        /// <returns>false if any required field is null. Note: VmId is required, VmssId is optional.</returns>
```
gladjohn: comment can be updated and VmId and CmssId references can be removed

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CsrMetadata.cs
gladjohn, line 26: suggest testing this against the newly created VM. It doesn't return a VMSSID when I checked

gladjohn:
Looks good overall, @Robbie-Microsoft, just a few minor comments remaining. I also see that the PR is already approved, so feel free to address the open comments and proceed with merging. Adding the suggested .NET 8 CSR comparison test would also be valuable for extra confidence and long-term regression coverage.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
Robbie-Microsoft: You're looking at the wrong thing. The Cuid OID is used correctly in Csr.cs. This is a RSASSA-PSS reference.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/CsrValidator.cs
Robbie-Microsoft: You're looking at the wrong thing. The Cuid OID is used correctly in Csr.cs. This is a RSASSA-PSS reference.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CsrMetadata.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CsrMetadata.cs
Robbie-Microsoft: This correctly ensures Cuid is not empty

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateRequest.cs
Robbie-Microsoft: I've created #5454 to address this. @bartonjs, can you please review it if you get a chance?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: This won't be possible, otherwise it will defeat the purpose of #5454.

PR #5435, opened 2025-08-12
Proof-of-concept test service

Avery-Dunn:
This draft PR presents a working version of the work described in KR [3261363](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3261363), and is meant to address feedback after some internal reviews of https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5332

This proof-of-concept demonstrates how we can us a simple test service to simulate services such as STS and provide scenario-specific test configuration. With this set up each MSAL can have more consistent testing experience by centralizing configuration and mocked HTTP responses, all while using an API that can be easily spun up locally for new additions and quick adjustments.

This PR makes the following changes:
- Add a new ConfigService class to retrieve and parse responses from the ID labs API
  - A new controller was made to serve the new endpoints and all the config: https://identitydivision.visualstudio.com/Engineering/_git/IDLABS/pullrequest/19526
- Adjust an existing test in `GenericAuthorityTests` to use the new `ConfigService` rather than local constants and mocked responses.
- Add a new `TokenRevocationTests` class to show how it could be used for a scenario that is hard to test end-to-end (Service Fabric MI and token revocation)

Much of the configuration here and in the test service was created as a first draft, good-enough-for-now implementation. The names of classes/methods/fields and all of the config in the test service are not final or made with any grand design in mind, and is very open to feedback.

N.B: currently these new tests only work when the API is running locally, so they will fail in the pipeline. The related PR in the ID labs repo must be merged in first.

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/TokenRevocationTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/TokenRevocationTests.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
             
             // Get the resource and endpoints from the scenario helper
             string resource = scenario.Resource;
-            string serviceUri = scenario.CreateIdentityProviderUri();
+            string serviceUri = scenario.CreateIdentityProviderUri_SuccessfulToken();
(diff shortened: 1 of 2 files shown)


bgavrilMS:
Please reopen if needed.

PR #5447, opened 2025-08-20
Add asn1 packge conditionally

bgavrilMS:
Fixes #

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
+++ b/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
@@ -82,6 +82,7 @@
     <Compile Remove="$(PathToMsalSources)\PlatformsCommon\PlatformNotSupported\ApiConfig\SystemWebViewOptions.cs" />
     <EmbeddedResource Include="$(PathToMsalSources)\Properties\Microsoft.Identity.Client.rd.xml" />
     <None Include="$(PathToMsalSources)\..\..\..\README.md" Pack="true" PackagePath="\" />
+    <None Include="Platforms\Features\CertificateSigningRequest\CertificateSigningRequest.cs" />
     <None Include="Platforms\net\JsonObjectAttribute.cs" />
     <None Include="Platforms\net\JsonStringConverter.cs" />
     <None Include="Platforms\net\MsalJsonSerializerContext.cs" />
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/PlatformsCommon/Polyfils/CertificateRequest.cs
```diff
@@ -0,0 +1,376 @@
...
+            _hashAlgorithmName = hashAlgorithm;
+            _rsaPadding = padding;
+        }
+
+        internal Collection<AsnEncodedData> OtherRequestAttributes { get; } = new Collection<AsnEncodedData>();
+
+        private static string MakePem(byte[] ber, string header)
+        {
+            const int LineLength = 64;
+
```
bgavrilMS, line 41: nit: remove redundant newlines

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -94,6 +94,7 @@
     <Compile Include="$(PathToMsalSources)\Platforms\Features\OpenTelemetry\**\*.cs" />
 
     <PackageReference Include="System.Diagnostics.DiagnosticSource" />
+    <PackageReference Include="System.Formats.Asn1" />
```
bgavrilMS, line 97: This is how you add a package ref conditionally. Notice the ItemGroup parent and the condition it has.

src/client/Microsoft.Identity.Client/PlatformsCommon/Polyfils/CertificateRequest.cs
```diff
@@ -0,0 +1,376 @@
...
+using System.Formats.Asn1;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+
+namespace Microsoft.Identity.Client.PlatformsCommon.Polyfils
+{
+    // This exists in net8+ but we are using a polifyl 
+    // TODO: evaluate using the net8+ version directly
+    internal class CertificateRequest
```
bgavrilMS, line 15: The nice thing about this class is that it follows the Single Responsability Principle - it only deals with making a cert request.

src/client/Microsoft.Identity.Client/PlatformsCommon/Polyfils/CertificateRequest.cs
```diff
@@ -0,0 +1,376 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.ObjectModel;
+using System.Formats.Asn1;
```
bgavrilMS, line 6: This comes either from the direct package reference on net fwk, or from the implicit reference on net8

PR #5448, opened 2025-08-20
MSI V2 client side keys

gladjohn:
**Fixes - adds the key management logic for MSI flows or aka CSK (Client side keys)** 

**This will not be merged as is. Instead will go into the MSI v2 feature branch.** 

This pull request introduces a new abstraction for Managed Identity key management, allowing for improved flexibility and security in how authentication keys are created and sourced. The main changes include the addition of provider classes for key creation (with platform-specific logic for Windows), a new interface for key providers, and integration of this abstraction into the authentication request flow. The implementation prioritizes secure hardware-backed keys and falls back to in-memory keys as needed.

**Managed Identity Key Provider Abstraction**

* Introduced the `IManagedIdentityKeyProvider` interface (`src/client/Microsoft.Identity.Client/ManagedIdentity/IManagedIdentityKeyProvider.cs`) and the `ManagedIdentityKeyInfo` class to encapsulate key details and provider messages. [[1]](diffhunk://#diff-713f9b581e819b3552a89e68bfe08533f6b6d500d44fa97fefe08c2e1e5d6ccdR1-R14) [[2]](diffhunk://#diff-849683882e3e4bacee7f40fc3e7cbe0b3a578ca4e49a4a688de88e2ed479fd39R1-R32)

**Platform-Specific Key Providers**

* Added `InMemoryManagedIdentityKeyProvider`, which generates and caches an RSA key in memory for authentication scenarios.
* Added `WindowsManagedIdentityKeyProvider`, which attempts to source keys from KeyGuard (VBS isolation), hardware TPM/KSP, or falls back to in-memory keys, with detailed logging and error handling.
* Added supporting helpers for Windows key operations, including KeyGuard and hardware-backed key creation (`WindowsCngKeyOperations`, `KeyGuardHelper`). [[1]](diffhunk://#diff-44405fcab03dded7eb0a56e8a422bc8df5d53029e8cbd76115ada61e2ccf50ebR1-R142) [[2]](diffhunk://#diff-284f66a04bfb70fa6a3c955bd652c9221dade4fd42cfe9b3cb25dbf986e04ae4R1-R78)

**Integration with Authentication Flow**

* Updated `ManagedIdentityAuthRequest` to use the new key provider abstraction, retrieving the best available key for authentication. [[1]](diffhunk://#diff-a1260a45bdc4ac7ab60d4a348996be6b69c5a48fa6f484fa41daa390c7e69ca8R22) [[2]](diffhunk://#diff-a1260a45bdc4ac7ab60d4a348996be6b69c5a48fa6f484fa41daa390c7e69ca8R32-R42)

**Dependency Management**

* Adjusted `Directory.Packages.props` to ensure `System.Security.Cryptography.Cng` is included as a private asset for key provider functionality. [[1]](diffhunk://#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156R25) [[2]](diffhunk://#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156L74)


src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -197,6 +198,16 @@ private async Task<AuthenticationResult> SendTokenRequestForManagedIdentityAsync
             ManagedIdentityClient managedIdentityClient = 
                 new ManagedIdentityClient(AuthenticationRequestParameters.RequestContext);
 
+            var keyProvider = ServiceBundle.PlatformProxy.ManagedIdentityKeyProvider;
+
+            var mi = await keyProvider.GetOrCreateKeyAsync(cancellationToken).ConfigureAwait(false);
+            var rsa = mi.KeyInfo;
+
+            byte[] data = System.Text.Encoding.UTF8.GetBytes("ping");
+            byte[] sig = rsa.SignData(data, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
```
bgavrilMS: Pls stop using PKCS1. PSS is the only allowed padding.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,154 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// src/client/Microsoft.Identity.Client/ManagedIdentity/Providers/WindowsManagedIdentityKeyProvider.cs
+#if !NETSTANDARD2_0
```
bgavrilMS: Leave a comment explaining what API is missing in netstandard2

src/client/Microsoft.Identity.Client/ManagedIdentity/MiKeyTypes.cs
```diff
@@ -0,0 +1,30 @@
...
+        public MiKeyType Type { get; }
+
+        public MiKeyInfo(RSA keyInfo, MiKeyType type)
+        {
+            KeyInfo = keyInfo ?? throw new ArgumentNullException(nameof(keyInfo));
+            Type = type;
+        }
+    }
+
+    internal interface IManagedIdentityKeyProvider
```
bgavrilMS: nit: move to it's own file

src/client/Microsoft.Identity.Client/ManagedIdentity/MiKeyTypes.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal enum MiKeyType { KeyGuard, Hardware, InMemory }
+
+    internal sealed class MiKeyInfo
+    {
+        public RSA KeyInfo { get; }
```
bgavrilMS: nit: naming `ActualKey` ? or just `Key` ?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -80,6 +80,13 @@
     <Compile Include="$(PathToMsalSources)\**\*.cs" Exclude="$(PathToMsalSources)\obj\**\*.*" />
     <Compile Remove="$(PathToMsalSources)\Platforms\**\*.*;$(PathToMsalSources)\Resources\*.cs" />
     <Compile Remove="$(PathToMsalSources)\PlatformsCommon\PlatformNotSupported\ApiConfig\SystemWebViewOptions.cs" />
+    <Compile Remove="C:\msi_vs_keys\src\client\Microsoft.Identity.Client\ManagedIdentity\IManagedIdentityPlatformKeyProvider.cs" />
```
bgavrilMS: undo everything here

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardKey.cs
```diff
@@ -0,0 +1,82 @@
...
+#if !NETSTANDARD2_0
+using System;
+using System.Security.Cryptography;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.KeyGuard
+{
+    /// <summary>
+    /// Helper for creating and validating Key Guard–isolated RSA keys.
+    /// </summary>
+    internal static class KeyGuardKey
```
bgavrilMS: nit: KeyGuardHelper

Although I'm not sure it's worth having this class at all, as it is only used in 1 place.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                CngProvider provider = new CngProvider(KgProviderName);
+
+                CngKey key;
+                if (CngKey.Exists(KgKeyName, provider))
+                {
+                    key = CngKey.Open(KgKeyName, provider);
+
+                    // Ensure actually KeyGuard-protected; if not, recreate as KeyGuard.
+                    if (!KeyGuardKey.IsKeyGuardProtected(key))
```
bgavrilMS: Is it worth adding all the `KeyGuardKey` methods in this class?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyProviderFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        internal static IManagedIdentityKeyProvider GetOrCreateProvider()
+        {
+            var p = Volatile.Read(ref s_provider);
+            if (p != null)
+                return p;
+
+            IManagedIdentityKeyProvider created = CreateProviderCore();
+            Interlocked.CompareExchange(ref s_provider, created, null);
+            return s_provider!;
```
bgavrilMS: nit: we don't have nullable in msal, so `!` isn't needed?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyProviderFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        private static IManagedIdentityKeyProvider CreateProviderCore()
```
bgavrilMS: This could use some logging.

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
gladjohn: this was not needed, used to test the flow. Now we have unit tests. But think we need to add some sort of static analysis as this line comes from CoPilot

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
gladjohn: moved it to to platform flags

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardKey.cs
gladjohn: keeping it as it's own file so I can conditionally exclude this one where not needed

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
gladjohn: CNG support is not available on netstd2.0, so this way I am able to exclude the file.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyProviderFactory.cs
gladjohn: removed

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -80,8 +80,10 @@
     <Compile Include="$(PathToMsalSources)\**\*.cs" Exclude="$(PathToMsalSources)\obj\**\*.*" />
     <Compile Remove="$(PathToMsalSources)\Platforms\**\*.*;$(PathToMsalSources)\Resources\*.cs" />
     <Compile Remove="$(PathToMsalSources)\PlatformsCommon\PlatformNotSupported\ApiConfig\SystemWebViewOptions.cs" />
+    <Compile Remove="$(PathToMsalSources)\ManagedIdentity\KeyProviders\WindowsCngKeyOperations.cs" />
     <EmbeddedResource Include="$(PathToMsalSources)\Properties\Microsoft.Identity.Client.rd.xml" />
     <None Include="$(PathToMsalSources)\..\..\..\README.md" Pack="true" PackagePath="\" />
+    <None Include="ManagedIdentity\KeyProviders\KeyGuardHelper.cs" />
```
gladjohn: rename added this - remove this and the lines way below

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyProviderFactory.cs
gladjohn: added

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
gladjohn: add package reference now and removed the conditions

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        private static RSA CreateRsaKeyPair()
+        {
+            RSA rsa;
+#if NET462 || NET472
+            // .NET Framework (Windows): use RSACng 
+            rsa = new RSACng();
+#else
+            // Cross-platform (.NET Core/8+/Standard)
```
bgavrilMS:
```suggestion
            // Cross-platform; uses Cng on Windows and OpenSSL on Linux
```

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+            finally
+            {
+                s_once.Release();
+            }
+        }
+
+        private static RSA CreateRsaKeyPair()
+        {
+            RSA rsa;
+#if NET462 || NET472
```
bgavrilMS: if NETFRAMEWORK

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
gladjohn: updated

build/template-run-mi-e2e-azurearc.yaml
```diff
@@ -37,4 +37,4 @@ steps:
     codeCoverageEnabled: false
     failOnMinTestsNotRun: true
     minimumExpectedTests: '1'
-    testFiltercriteria: 'TestCategory=MI_E2E_AzureArc'
+    testFiltercriteria: '(TestCategory=MI_E2E_AzureArc|TestCategory=MI_E2E_KeyAcquisition_KeyGuard)'
```
Robbie-Microsoft, line 40: Why did you put this in the AzureArc yaml file? Should we make a new yaml file just for this new e2e test?

src/client/Microsoft.Identity.Client/ManagedIdentity/IManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    internal interface IManagedIdentityKeyProvider
+    {
+        Task<ManagedIdentityKeyInfo> GetOrCreateKeyAsync(ILoggerAdapter logger, CancellationToken ct);
```
Robbie-Microsoft, line 31: Should we also have a `TryGetExistingKeyAsync` separate from creation for rotation scenarios

Robbie-Microsoft, line 31: Can you add XML doc comments to clarify: lifetime, thread-safety, and whether key rotation is expected

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </summary>
+    internal sealed class ManagedIdentityKeyInfo
+    {
+        public RSA Key { get; }
+        public ManagedIdentityKeyType Type { get; }
+        public string ProviderMessage { get; }
+
+        public ManagedIdentityKeyInfo(RSA keyInfo, ManagedIdentityKeyType type, string providerMessage)
+        {
+            Key = keyInfo;
```
Robbie-Microsoft, line 32: Can we add an enum `ProtectionTier` { Isolated, Hardware, SoftwarePersisted, InMemoryEphemeral } for telemetry?

Robbie-Microsoft, line 32: Should we add a `CreatedAt` timestamp for rotation or age-based policies?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+
+                try
+                {
+                    rsa = CreateRsaKeyPair();
+                    message = "In-memory RSA key created for Managed Identity authentication.";
+                    logger?.Info("[MI][InMemoryKeyProvider] RSA key created (2048).");
+                }
+                catch (Exception ex)
+                {
+                    message = $"Failed to create in-memory RSA key: {ex.GetType().Name} - {ex.Message}";
```
Robbie-Microsoft, line 73: Do we already have a helper that forms error messages like this?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+
+namespace Microsoft.Identity.Client.ManagedIdentity.KeyProviders
+{
+    /// <summary>
+    /// In-memory managed identity key provider.
+    /// </summary>
+    internal sealed class InMemoryManagedIdentityKeyProvider : IManagedIdentityKeyProvider
+    {
+        private static readonly SemaphoreSlim s_once = new(1, 1);
+        private volatile ManagedIdentityKeyInfo _cached;
```
Robbie-Microsoft: Please rename to _cachedKey for clarity.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        private static RSA CreateRsaKeyPair()
+        {
+            RSA rsa;
+#if NETFRAMEWORK
+            // .NET Framework (Windows): use RSACng 
+            rsa = new RSACng();
+#else
+            // Cross‑platform: RSA.Create() -> CNG (Windows) / OpenSSL (Linux).
+            rsa = RSA.Create();
+#endif
```
Robbie-Microsoft: This is what we have in the feature branch right now. Is netstandard2.0 included in NETFRAMEWORK? If so, should you change this to look like the feature branch code?
```
#if NET462 || NET472
            // .NET Framework runs only on Windows, so RSACng (Windows-only) is always available
            rsa = new RSACng();
#else
            // Cross-platform .NET - RSA.Create() returns appropriate PSS-capable implementation
            rsa = RSA.Create();
#endif
```

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.KeyProviders
+{
+    /// <summary>
+    /// In-memory managed identity key provider.
+    /// </summary>
+    internal sealed class InMemoryManagedIdentityKeyProvider : IManagedIdentityKeyProvider
+    {
+        private static readonly SemaphoreSlim s_once = new(1, 1);
```
MZOLN: Is this to support double checked locking for async creation flows?
Wouldnt it make sense to extract it to a common component like AsyncLazy<T> structure that we could then reuse?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+using System;
+using System.Security.Cryptography;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.KeyGuard
+{
+    /// <summary>
+    /// Helper that creates or opens a Key Guard–protected RSA key.
+    /// </summary>
+    internal static class KeyGuardHelper
```
MZOLN: I'd think that we should avoid helper classes if we can. Since this is actually creating keys, could it be called a KeyGuardKeyProvider and then have a create method?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            RSA rsa;
+#if NETFRAMEWORK
+            // .NET Framework (Windows): use RSACng 
+            rsa = new RSACng();
+#else
+            // Cross‑platform: RSA.Create() -> CNG (Windows) / OpenSSL (Linux).
+            rsa = RSA.Create();
+#endif
+            rsa.KeySize = 2048;
```
MZOLN: Hardcoding key length can hurt us in the long run. Could it be a constructor variable with a default value? This way if we ever need to route it through config it's already there. In some time 3072 or even 4096 and more would become the minimum. Currently 2048 sets us up only for deployments up to 2030 according to NIST, but already high-assurance deployments are expected to deploy 4096....

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+                KeyUsage = CngKeyUsages.AllUsages,
+                ExportPolicy = CngExportPolicies.None,
+                KeyCreationOptions =
+                      CngKeyCreationOptions.OverwriteExistingKey
+                    | NCryptUseVirtualIsolationFlag
+                    | NCryptUsePerBootKeyFlag
+            };
+
+            p.Parameters.Add(new CngProperty("Length",
+                              BitConverter.GetBytes(2048),
```
MZOLN: In my mind, we should avoid hardcoding key lengths. Just like above, I'd feed it through the constructor to ensure openess to future adaptability.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Microsoft.Identity.Client.ManagedIdentity.KeyGuard
+{
+    /// <summary>
+    /// Helper that creates or opens a Key Guard–protected RSA key.
+    /// </summary>
+    internal static class KeyGuardHelper
+    {
+        private const string ProviderName = "Microsoft Software Key Storage Provider";
+        private const string KeyName = "KeyGuardRSAKey";
```
MZOLN: Is this always const? What if the component changes, would we have to provide a new verison of MSAL? Wouldnt it be more flexible to add optional configurability?

This way if key guard rolls a breaking change we wouldnt have to force every single user to wait for an MSAL update and roll it out.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+                        logger?.Verbose(() => "[MI][WinKeyProvider] Unable to obtain a KeyGuard-protected key.");
+                        return false;
+                    }
+                }
+
+                rsa = new RSACng(key);
+                if (rsa.KeySize < 2048)
+                {
+                    try
+                    { rsa.KeySize = 2048; }
```
MZOLN: Do we know if behavior here is consistent? Does altering the key size force generation of a new key or just adds padding?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+                    catch { /* some providers don't allow */ }
```
MZOLN: Why would we fail silently if we attempt to extend the key size and it fails? Doesnt it create a false sense of security?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+            catch (PlatformNotSupportedException)
+            {
+                // VBS/Core Isolation not available => KeyGuard unavailable
+                logger?.Verbose(() => "[MI][WinKeyProvider] Exception creating KeyGuard key.");
+                return false;
+            }
+            catch (CryptographicException)
+            {
+                return false;
+            }
```
MZOLN: Any reason why we wouldnt want to log those errors and show them to the user? Might be crucial for troubleshooting.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+                    : CreateUserPcpRsa(provider, HardwareKeyName);
+
+                rsa = new RSACng(key);
+
+                if (rsa.KeySize < 2048)
+                {
+                    try
+                    { rsa.KeySize = 2048; }
+                    catch { /* PCP typically ignores post-create change */ }
+                }
```
MZOLN: Same as above

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            // Return cached if available
+            if (_cached != null)
+            {
+                logger?.Info("[MI][WinKeyProvider] Returning cached key.");
+                return _cached;
+            }
+
+            // Ensure only one creation at a time
+            logger?.Verbose(() => "[MI][WinKeyProvider] Waiting on creation semaphore.");
```
Robbie-Microsoft: Do we also need a log for after the wait is finished?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+                    // Check again after recreate; still null or not protected -> give up KeyGuard path
+                    if (key == null || !KeyGuardHelper.IsKeyGuardProtected(key))
+                    {
+                        key?.Dispose();
+                        logger?.Verbose(() => "[MI][WinKeyProvider] Unable to obtain a KeyGuard-protected key.");
+                        return false;
+                    }
+                }
+
+                rsa = new RSACng(key);
```
MZOLN, line 104: I'm not entirely sure, but from what I'm understanding CngKey is disposable and RSACng doesnt take ownership of it -  won't dispose it. Returning just RSACng but not CngKey would never dispose of it (?)
If we dispose of cngkey here and return rsacng we might do it prematurely and at the same time if we never dispose cngkey, we might leak native handles.

There are a few safe patterns to deal with it -> if key is exportable, just copy it when creating the RSACng and another one is to create a disposable wrapper strucutre that would contain both RSACng and CngKey (keep that private) and have the user dispose of them both, then return that wrapper instead of RSACng e.g.

`
   internal sealed class RsaWithKey: IDisposable
   {
       public RSACng Rsa { get; }
       private readonly CngKey _key;
    
       public RsaWithKey(CngKey key)
       { 
          _key = key ?? throw new ArgumentNullException(nameof(key));
           Rsa = new RSACng(key);
         }
      
         // probably a disposable pattern would be better  
         public void Dispose()
         {
            Rsa.Dispose();
            _key.Dispose();
         }
   }
`

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+                    if (WindowsCngKeyOperations.TryGetOrCreateKeyGuard(logger, out RSA kgRsa))
+                    {
+                        messageBuilder.AppendLine("KeyGuard RSA key created successfully.");
+                        _cached = new ManagedIdentityKeyInfo(kgRsa, ManagedIdentityKeyType.KeyGuard, messageBuilder.ToString());
+                        logger?.Info("[MI][WinKeyProvider] Using KeyGuard key (RSA).");
+                        return _cached;
+                    }
+                    else
+                    {
+                        messageBuilder.AppendLine("KeyGuard RSA key creation not available or failed.");
```
Robbie-Microsoft, line 103: You have a catch for "failed" key creation, why is this else necessary?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <summary>
+    /// Windows policy:
+    ///   1) KeyGuard (CVM/TVM) if available
+    ///   2) Hardware (TPM/KSP via Microsoft Platform Crypto Provider)
+    ///   3) In-memory fallback (delegates to InMemoryManagedIdentityKeyProvider)
+    /// </summary>
+    internal sealed class WindowsManagedIdentityKeyProvider : IManagedIdentityKeyProvider
+    {
+        private static readonly SemaphoreSlim s_once = new (1, 1);
+        private volatile ManagedIdentityKeyInfo _cached;
```
MZOLN: This is reusing the pattern from above -> this would be a great component.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
Robbie-Microsoft: Should this else only be for "unavailable"?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyProviderFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+        // Cached singleton instance of the chosen key provider.
+        private static IManagedIdentityKeyProvider s_provider;
+
+        /// <summary>
+        /// Returns the cached provider if available; otherwise creates it in a thread-safe manner.
+        /// </summary>
+        internal static IManagedIdentityKeyProvider GetOrCreateProvider(ILoggerAdapter logger)
+        {
+            // Fast path: read the field once (Volatile ensures latest published value).
+            IManagedIdentityKeyProvider existing = Volatile.Read(ref s_provider);
```
MZOLN, line 50: Isn't it the same/similar pattern to volatile/semaphoreslim from above?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+                    if (WindowsCngKeyOperations.TryGetOrCreateHardwareRsa(logger, out RSA hwRsa))
+                    {
+                        messageBuilder.AppendLine("Hardware RSA key created successfully.");
+                        _cached = new ManagedIdentityKeyInfo(hwRsa, ManagedIdentityKeyType.Hardware, messageBuilder.ToString());
+                        logger?.Info("[MI][WinKeyProvider] Using Hardware key (RSA).");
+                        return _cached;
+                    }
+                    else
+                    {
+                        messageBuilder.AppendLine("Hardware RSA key creation not available or failed.");
```
Robbie-Microsoft, line 128: Same comment: Will the catch cover the "failed" scenarios?

MZOLN (commented):
General feedback

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+                {
+                    messageBuilder.AppendLine($"Hardware RSA key creation threw exception: {ex.GetType().Name}: {ex.Message}");
+                    logger?.WarningPii(
+                        $"[MI][WinKeyProvider] Exception creating Hardware key: {ex}",
+                        $"[MI][WinKeyProvider] Exception creating Hardware key: {ex.GetType().Name}");
+                }
+
+                // 3) In-memory fallback (software RSA)
+                logger?.Info("[MI][WinKeyProvider] Falling back to in-memory RSA key (software).");
+                if (ct.IsCancellationRequested)
```
Robbie-Microsoft, line 142: you are already doing this on line 47. Why are you doing it again here?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+                {
+                    logger?.Verbose(() => "[MI][WinKeyProvider] Cancellation requested before in-memory fallback.");
+                    ct.ThrowIfCancellationRequested();
+                }
+
+                var fallback = new InMemoryManagedIdentityKeyProvider();
```
Robbie-Microsoft: Can you give this a clearer name? `inMemoryMIKP`?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+                if (rsa.KeySize < 2048)
```
Robbie-Microsoft: I'd make a `const` for 2048 since you're using it more than once

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+            }
+            catch (PlatformNotSupportedException)
+            {
+                // VBS/Core Isolation not available => KeyGuard unavailable
+                logger?.Verbose(() => "[MI][WinKeyProvider] Exception creating KeyGuard key.");
+                return false;
+            }
+            catch (CryptographicException)
+            {
+                return false;
```
Robbie-Microsoft: Should you include a logger in this catch?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+                CngProvider provider = new CngProvider(SoftwareKspName);
+                CngKeyOpenOptions openOpts = CngKeyOpenOptions.UserKey | CngKeyOpenOptions.Silent;
+
+                CngKey key = CngKey.Exists(HardwareKeyName, provider, openOpts)
+                    ? CngKey.Open(HardwareKeyName, provider, openOpts)
+                    : CreateUserPcpRsa(provider, HardwareKeyName);
+
+                rsa = new RSACng(key);
+
+                if (rsa.KeySize < 2048)
```
Robbie-Microsoft: use `const` for 2048

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                // Add HResult to make CI diagnostics actionable
+                logger?.Verbose(() => "[MI][WinKeyProvider] Hardware key creation/open failed. " +
+                                       $"HR=0x{e.HResult:X8}. {e.GetType().Name}: {e.Message}");
+                return false;
+            }
+
+            static CngKey CreateUserPcpRsa(CngProvider provider, string name)
+            {
+                var p = new CngKeyCreationParameters
```
Robbie-Microsoft: Can you give this a more intuitive name? `params`?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.ManagedIdentity.KeyGuard;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.KeyProviders
+{
+    /// <summary>
```
Robbie-Microsoft, line 11: Should we add a minimal retry policy? What about `NTE_DEVICE_NOT_READY` exception (rare with TPM but can occur during initialization)?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
Robbie-Microsoft: You're using these constants in more than one file? Can you extract them to a shared Constants file?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+        // KeyGuard + per-boot flags
+        private const CngKeyCreationOptions NCryptUseVirtualIsolationFlag = (CngKeyCreationOptions)0x00020000;
+        private const CngKeyCreationOptions NCryptUsePerBootKeyFlag = (CngKeyCreationOptions)0x00040000;
+
+        /// <summary>
+        /// Creates a new RSA-2048 Key Guard key.
+        /// </summary>
+        public static CngKey CreateFresh(ILoggerAdapter logger)
+        {
+            var p = new CngKeyCreationParameters
```
Robbie-Microsoft: Can you give this a more intuitive name like `params`?

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
```diff
@@ -0,0 +1,78 @@
...
+                    "Inner exception: " + ex.Message);
+                
+                return null;
+            }
+        }
+
+        /// <summary>Returns <c>true</c> if the key has the Key Guard flag.</summary>
+        public static bool IsKeyGuardProtected(CngKey key)
+        {
+            if (!key.HasProperty("Virtual Iso", CngPropertyOptions.None))
```
Robbie-Microsoft: Make a const for "Virtual Iso" since you're using the string more than once.

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -29,13 +29,17 @@ public ManagedIdentityAuthRequest(
...
             _cryptoManager = serviceBundle.PlatformProxy.CryptographyManager;
+            _managedIdentityKeyProvider = serviceBundle.PlatformProxy.ManagedIdentityKeyProvider;
         }
 
         protected override async Task<AuthenticationResult> ExecuteAsync(CancellationToken cancellationToken)
         {
             AuthenticationResult authResult = null;
             ILoggerAdapter logger = AuthenticationRequestParameters.RequestContext.Logger;
 
+            ManagedIdentityKeyInfo keyInfo = await _managedIdentityKeyProvider.GetOrCreateKeyAsync(
```
Robbie-Microsoft: Doesn't `keyInfo` need to be attached to the return object?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyProviderFactory.cs
Robbie-Microsoft, line 50: Why aren't you using semaphoreSlim or declaring a volatile variable like you do in the other files?

tests/Microsoft.Identity.Test.E2e/ManagedIdentityKeyAcquisitionTests.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.Identity.Client.ManagedIdentity.KeyProviders;
+using Microsoft.Identity.Test.Common.Core.Helpers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.E2E
+{
+    [TestClass]
+    public class ManagedIdentityKeyAcquisitionTests
+    {
+        private const string SoftwareKspName = "Microsoft Software Key Storage Provider";
```
Robbie-Microsoft, line 15: Use as const from prod files

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
MZOLN: I'd apply that to all const strings and key lengths. Those are first candidates to potentially become configurable, to prevent customers from having to pull a new build on the instant in case changes arise. Instead they could do it gradually, update the config then pull a new build (roll outs take time).

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
bgavrilMS: This is probably your integration point.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
bgavrilMS: It is not. NETFRAMEWORK is the old .NET FWK which is Windows only. netstandard is like .NET Core

bgavrilMS: Pls add a comment on why you did it this way @gladjohn . Is it because on .NET FWK, RSA.Create() doesn't use RsaCng by default?

build/template-run-mi-e2e-azurearc.yaml
bgavrilMS: +1, needs a comment.

src/client/Microsoft.Identity.Client/ManagedIdentity/IManagedIdentityKeyProvider.cs
bgavrilMS: This is internal interface. I think it's ok as is, but if you integrate this PR into the feature branch and need an extra method, you can add it.,

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
bgavrilMS: https://devblogs.microsoft.com/dotnet/asynclazyt/

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+                try
+                {
+                    rsa = CreateRsaKeyPair();
+                    message = "In-memory RSA key created for Managed Identity authentication.";
+                    logger?.Info("[MI][InMemoryKeyProvider] RSA key created (2048).");
+                }
+                catch (Exception ex)
+                {
+                    message = $"Failed to create in-memory RSA key: {ex.GetType().Name} - {ex.Message}";
+                    logger?.WarningPii(
```
bgavrilMS, line 74: I don't see any PII here. Please log the entire expcetion.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
bgavrilMS: We don't want to expose this sort of low level config to the public. MSAL should be able to handle it on its own. Agree it can be a constant.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
```diff
@@ -0,0 +1,142 @@
...
+                    try
+                    { rsa.KeySize = 2048; }
+                    catch { /* some providers don't allow */ }
+                }
+                return true;
+            }
+            catch (PlatformNotSupportedException)
+            {
+                // VBS/Core Isolation not available => KeyGuard unavailable
+                logger?.Verbose(() => "[MI][WinKeyProvider] Exception creating KeyGuard key.");
```
bgavrilMS: Verbosity should be info.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
gladjohn: good suggestion, but for now I'd simply file a task on MSAL .NET as this was taken from existing logic. We may need to bring this in a separate PR.

src/client/Microsoft.Identity.Client/ManagedIdentity/IManagedIdentityKeyProvider.cs
bgavrilMS: It's thread safe. Lifetime and key rotation are managed internally and are not the responsability of the caller. For now, you can assume keys are long lived / no rotation.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
bgavrilMS: no. However, I do not see any PII. I recommend you log the entire exception. `ex.Tostring()` will do that.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
bgavrilMS: It is because on .net fwk, RSA.Create() doesn't use RsaCng objects by default. RSA is an abstract class and implementations are hidden from callers. But in this case, we nee RsaCng.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
bgavrilMS: A constant for this is ok for now. Making it configurable is over-engineered.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
bgavrilMS: No. Some of this code was copied from a Windows component I believe. They do not use retries.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
bgavrilMS: RsaCng takes ownership. See the remarks in https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsacng.key?view=net-9.0#remarks

Current approach is ok.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
bgavrilMS: Let's log the error? In any case there will be an attestation token involved that proves the key is in hardware.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
bgavrilMS: Should be fine - https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsacng.-ctor?view=net-9.0

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
bgavrilMS: Agreed, this is a pattern used in .NET, a defensive coding technique. `bool TryXYZ(out result)` e.g. `bool TryParse(string input, out int result)`. It should not throw exceptions on failure, just return false.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
bgavrilMS: This is how you cancel a long running operation in .NET. You keep checking if the cancellation was requested. Cancellation is not guarateneed, it is best effort. Generally, you'd check before doing some expensive work that is sync (which is the case here).  See https://devblogs.microsoft.com/premier-developer/recommended-patterns-for-cancellationtoken/

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityKeyInfo.cs
bgavrilMS: None exist. A key doesn't expire. We should only add it when needed.

bgavrilMS: Let's discuss about telemetry separately. For now, don't capture it.

tests/Microsoft.Identity.Test.E2e/ManagedIdentityKeyAcquisitionTests.cs
gladjohn: Sharing the constant turns the assertion into a tautology, so the test can’t catch unintended changes. Good tests need an independent oracle; they should state expected inputs/outputs directly rather than deriving them from the system under test.

Dan North’s advice that tests should optimize for clarity (DAMP), not DRY. xUnit Patterns also frames result verification as comparing actual results to an expected state—not to the same source (i.e. the code from the product under test)

this is a good read - https://dannorth.net/blog/let-your-examples-flow/

build/template-run-mi-e2e-azurearc.yaml
Robbie-Microsoft: ping @gladjohn

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
Robbie-Microsoft: ping @gladjohn

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
Robbie-Microsoft: ping @gladjohn

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/IManagedIdentityKeyProvider.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsManagedIdentityKeyProvider.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
Robbie-Microsoft: done

build/template-run-mi-e2e-azurearc.yaml
gladjohn: yes, this can go in as a new YAML

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
gladjohn: same comment from Bogdan earlier on, We don't want to expose this sort of low level config to the public. MSAL should be able to handle it on its own

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
gladjohn: we can move this to windows key provider class

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <returns>
+        /// An <see cref="RSA"/> instance configured with a 2048-bit key size.
+        /// On .NET Framework, returns <see cref="RSACng"/>; on other platforms, returns the default RSA implementation.
+        /// </returns>
+        /// <remarks>
+        /// This method is public instead of private because it is used in unit tests
+        /// </remarks>
+        public static RSA CreateRsaKeyPair()
+        {
+#if NET462 || NET472 || NET8_0
```
Robbie-Microsoft, line 105: @gladjohn do we need a new strategy to ensure NET8_0 is Windows only?

gladjohn: KeyGuard is Windows only. And for Linux we need to use in memory. Also would be good to test in Linux.

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/KeyGuardHelper.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/WindowsCngKeyOperations.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/KeyProviders/InMemoryManagedIdentityKeyProvider.cs
Robbie-Microsoft: I'm talking specifically about how `(NET8_0 && WINDOWS)` wasn't working for us

PR #5449, opened 2025-08-20
ImdsV2: Reworked Custom ASN1 Encoder to use System.Formats.Asn1 Nuget Package

Robbie-Microsoft:
We decided to use the custom ASN1 Encoder for all versions of .NET, instead of making it a Polyfill for 462, 472 and netstandard2.0, so that we can be sure each version is using the same code. We expect this will reduce confusion when working with or maintaining this code in the future.

- Existing unit tests for CSR PEM creation still function correctly
- Deleted unnecessary unit tests - null checks happen before the CSR PEM is created
- Moved all ImdsV2 files into their own folder within the Managed Identity folder


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Identity.Client.PlatformsCommon.Shared;
 using System.IO;
 using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.ManagedIdentity.V2;
```
bgavrilMS, line 12: needed?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PemPayload.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    internal class PemPayload
```
bgavrilMS, line 6: This is a strange object with only 1 string memebr. Is it needed?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -214,7 +214,7 @@ private async Task<CertificateRequestResponse> ExecuteCertificateRequestAsync(
                 { "x-ms-client-request-id", _requestContext.CorrelationId.ToString() }
             };
             
-            var payload = new PemPayload { pem = pem };
+            var payload = new PemPayload { pem = csrPem };
             var body = JsonHelper.SerializeToJson(payload);
```
bgavrilMS: Do you really need to create a new object just to serialize it in a specific way? Would it be simpler to just use a string template here instaed, e.g. `string body = @"{pemPayload=<payload>}` ? (you'll need to escape { correctly)

bgavrilMS (approved):

bgavrilMS:
Approved with minor comments.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: yes, the ImdsV2ManagedIdentitySource is referenced twice in this file

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: this is great, thanks

PR #5454, opened 2025-08-27
ImdsV2: Implemented downlevel polyfill for CertificateRequest

Robbie-Microsoft:
- Added conditional compilation logic to exclude `CertificateRequest.cs` from all targets by default

- Added conditional include to only compile `CertificateRequest.cs` for downlevel frameworks: `net462`, `net472`, and `netstandard2.0`

- __Excluded from__: `net8.0` (and mobile variants)

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
 using System.Formats.Asn1;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;


src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
```diff
@@ -14,11 +14,21 @@ internal static string Generate(string clientId, string tenantId, CuidInfo cuid)
         {
             using (RSA rsa = CreateRsaKeyPair())
             {
-                CertificateRequest req = new CertificateRequest(
+#if NET8_0_OR_GREATER
```
bartonjs: You shouldn't need the \#if at all, since you've conditionally included the file.

If you're going to add a \#if, I'd do it either a) around the polyfill class, or b) just around the `new` line:

```csharp
CertificateRequest req =
#if NET8_0_OR_GREATER
    // Use a namespace qualified name to ensure we're using the built-in version
    new System.Security.Cryptography.X509Certificates.CertificateRequest(
#else
    new CertificateRequest(
#endif
        new X500DistinguishedName($"CN={clientId}, DC={tenantId}"),
        ... 
```

(you don't need to use `var` since the file was conditionally compiled, there's only one name within reach)

Since dotnet/runtime (and any personal projects I ever work on) disables the implicit "include everything", my personal ranking would be

1) Conditional include the polyfill file, end of change.
2) \#if around the contents of the polyfill, end of change.
3) \#if around just the first line of a multi-line new
4) something like what you have here

In a project where implicit-include is on (or an explicit `**/*` was added), I'd (personally) go for (2).

bgavrilMS: Are you suggesting that we put the polifyl class in the `System.Security.Cryptography.X509Certificates` ? The problem is that on net472, `CertificateRequest` object already exists, but it doesn't have the `OtherRequestAttributes` property. There would be an error on net472 due to naming conflict.

Another option would be to use a proxy pattern - `CertificateRequestProxy` which either calls into the FWK CertificateRequest or into the polyfill CertificateRequest, at the expense of a few extra objects on the stack. This code is not perf critical to matter.

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -80,13 +80,20 @@
...
+    <Compile Remove="$(PathToMsalSources)\ManagedIdentity\V2\CertificateRequest.cs" />
     <EmbeddedResource Include="$(PathToMsalSources)\Properties\Microsoft.Identity.Client.rd.xml" />
     <None Include="$(PathToMsalSources)\..\..\..\README.md" Pack="true" PackagePath="\" />
     <None Include="Platforms\net\JsonObjectAttribute.cs" />
     <None Include="Platforms\net\JsonStringConverter.cs" />
     <None Include="Platforms\net\MsalJsonSerializerContext.cs" />
   </ItemGroup>
 
+  <!-- Include custom CertificateRequest polyfill only for downlevel frameworks that don't have OtherRequestAttributes -->
+  <ItemGroup Condition="'$(TargetFramework)' != '$(TargetFrameworkNet)'">
```
bgavrilMS: This csproj file is messy enough. I think I'd prefer to have #if in the code instead.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
bartonjs: You can put it in the System.Security.Cryptography.X509Certificates namespace if you want... if your caller and the polyfill are in the same assembly, the same assembly version will win.  (though you'll want to suppress CS0436).

[Example, produces CS0436](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAGAAkIEYA6AZRjAFcpsMBPKgYSlYAcMIA5lACGPABbsAGgFZiATg4woGbADNsYYRhgBnANwBYAFDGAdsIC2unsLAxy1Oo2ZtO3PoJHipshUpXqmto6xgDexqSRpNim2lDmuOQATKSKymoaWjAASjAAjgy6GBFR4UZRpAC+xtUmRuZWOjZ2pADyGGJKYSWRMXEJyanCuLhd5aU9FdGxSgOUSOQopACywjEAFACUkxVlU/ukEMAAVvQYh6QAvKSmMADuqf4ZQTn5hToYW4bj+7UVtZUgA===)

If you don't put it in the same namespace, but it's in the same namespace as the calling class (or a parent namespace of the calling class), then (since your `using`-import statements are outside your `namespace` block) the "related namespace" version will win.

[Example, no warnings, and only compiles because it found the local version](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAGAAkIEYA6AZRjAFcpsMBPKgYSlYAcMIA5lACGPABbsAGgFZiATg4woGbADNsYYRhgBnANwBYAFDGAdsIC2unsLAxSAOUvXbMKgEFjAb2Ok/pbFNtKHNccgAmUkVlNQ0tGAAlGABHBl0MX38fI39SAF9jApMjcysdGztHZ3LXD04xPAATb0y/QODQiKjhXFwlVtJs3Nz2pU7KJHIUUgBZYUCACgBKAdyh4Y2IYAAregxSCFIAXlJTGAB3KKUVdU1tJNT05cMcjfyBorygA)

You should be able to revert everything in this change, then go to CertificateRequest.cs and surround the whole file (or the namespace block, or the class block, whatever) with `#if !NET7_0_OR_GREATER` and everything just works.

* If it compiles under net8, you know it's using the built-in version.
* If it compiles under net472, you know it's using your polyfill.

You could prove it to yourself with a temporary `Console.WriteLine(req.GetType().Assembly.FullName);`, and see that it's Microsoft.Identity.Client on older TFMs, and System.Security.Cryptography on newer ones.  (Or print req.GetType().FullName, if the namespaces stay different)

Robbie-Microsoft: @bartonjs I think I'm following you correctly, and I do think your first option is the best choice. If I revert every change in the PR and add `#if !NET7_0_OR_GREATER` around the namespace in CertificateRequest.cs and use `var req = ...` in Csr.cs, I see a CS0246 for CertificateRequest on the `var req = ...` line. Did I misunderstand what you were saying above?

bartonjs: CS0246 would surprise me, since you should see either `Microsoft.Identity.Client.ManagedIdentity.V2.CertificateRequest` (same namespace) for older TFMs, or `System.Security.Cryptography.X509Certificates.CertificateRequest` via the using-import for `System.Security.Cryptography.X509Certificates`.

What TFMs are failing with that error?

Robbie-Microsoft: oops, turns out I just needed to re-build the solution to resolve the errors!

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/Csr.cs
Robbie-Microsoft: I've implemented option 1.

Robbie-Microsoft:
> LGTM

Thanks for all the help Jeremy!

PR #5456, opened 2025-08-28
Remove passing x-client-os as query param in authorization uri

### LINKED ISSUE #5451: [Bug] GetAuthorizationUri should not add x-client-os
### Library version used

4.72.1

### .NET version

all

### Scenario

cca.GetAuthorizationUri generates the auth URI and adds a few query parameters, like `x-client-os`. This is used by ESTS as a hint to determine the OS and if it is not understood, the STS will fallback to UserAgent. 

But since `cca.GetAuthorizationUri` is used on the backend, the OS is that of the server, either Linux or Windows. This is confusing ppl, because they open their website on their ios phone, but the OS shows up as Linux in the Entra logs.

Proposed solution: stop adding x-client-os to the uri
Impact: it will make ests internal logs simpler

### Relevant code snippets

```csharp

```

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

neha-bhargava:
Fixes #5451 

Exclude passing the x-client-os param from GetAuthorizationUri 

**Testing**
Update unit tests to validate that the query param is not passed

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeRequestComponent.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/Interactive/AuthCodeRequestComponent.cs
@@ -214,10 +214,10 @@ namespace Microsoft.Identity.Client.Internal
                     _requestParams.RequestContext.CorrelationId.ToString();
             }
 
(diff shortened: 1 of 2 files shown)


gladjohn:
looks good, thanks @neha-bhargava

PR #5466, opened 2025-09-09
Avdunn/issuer validation adjustment

Avery-Dunn:
This issue in the identity web repo exposed a problem in the issuer validation steps: https://github.com/AzureAD/microsoft-identity-web/issues/3450

Originally the validation step would pass if the authority "started with" the issuer, i.e. it matched the scheme, host, and start of the path. However, there are cases where the path could be different but it still represents a valid authority/issuer pair, such as when the authority uses "organizations" but the issuer uses the actual tenant. 

While it might be better to use the `authority` API instead of `oidcAuthority` in that scenario, we don't strictly need to check the path for the OIDC issuer validation and as https://github.com/AzureAD/microsoft-identity-web/issues/3450 shows this can be a breaking change.

This PR adjust the issuer validation to only check the HTTP scheme and host, and adds a test confirming that behavior. The existing tests should still pass.

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
@@ -338,8 +338,8 @@ namespace Microsoft.Identity.Test.Unit.CoreTests.InstanceTests
         {
             using (var httpManager = new MockHttpManager())
             {
-                string authority = "https://demo.duendesoftware.com/tenant1/2.0/";
-                string issuerWithDifferentPath = "https://demo.duendesoftware.com/oidc/2.0/";
+                string authority = "https://demo.duendesoftware.com/someTenant/2.0/";
+                string issuerWithDifferentPath = "https://demo.duendesoftware.com/organizations/2.0/";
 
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Identity.Test.Unit/CoreTests/InstanceTests/GenericAuthorityTests.cs
```diff
@@ -333,6 +333,50 @@ public async Task BadOidcResponse_ThrowsException_Async(string badOidcResponseTy
             }
         }
 
+        [TestMethod]
+        public async Task OidcIssuerValidation_AcceptsDifferentPath_Async()
+        {
+            using (var httpManager = new MockHttpManager())
+            {
+                string authority = "https://demo.duendesoftware.com/organizations/2.0/";
```
bgavrilMS: Let's use https://login.microsoftonline.com/organizations/v2.0 like the Id.Web bug in the test.

Avery-Dunn: Adjusted in the latest commit. Since the issue is probably only going to happen when using a Microsoft host, the test now uses "login.microsoftonline.com" instead of the custom domain.

PR #5467, opened 2025-09-09
Remove confusing error text as it only applies to one of many possible causes

PetarSDimov:
The text that is removed only applies to one of many possible causes of ApiContractViolation or IncorrectConfigiration.
This causes confusion with developers. 
The recommendation would be to enable PII during development to see the WAM error message which will contain the correct reason for the error. 
A long-term change may be considered, and that will require proper classification of all cases where ApiContractViolation or IncorrectConfigiration may occur.

**Testing**
No new tests are needed

**Performance impact**
No perf impact

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
+++ b/src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
@@ -90,10 +90,7 @@ namespace Microsoft.Identity.Client.Platforms.Features.RuntimeBroker
                         $" Error Code: {errorCode} \n" +
                         $" Error Message: {authResult.Error.Status} \n" +
                         $" WAM Error Message: {authResult.Error.Context} \n" +
-                        $" Internal Error Code: {internalErrorCode} \n" +
-                        $" Possible causes: \n" +
-                        $"- Invalid redirect uri - ensure you have configured the following url in the application registration in Azure Portal: " +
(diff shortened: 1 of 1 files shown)


DharshanBJ (approved):

src/client/Microsoft.Identity.Client.Broker/WamAdapters.cs
```diff
@@ -90,10 +90,7 @@ private static MsalException CreateExceptionFromWamError(
                         $" Error Code: {errorCode} \n" +
                         $" Error Message: {authResult.Error.Status} \n" +
                         $" WAM Error Message: {authResult.Error.Context} \n" +
-                        $" Internal Error Code: {internalErrorCode} \n" +
-                        $" Possible causes: \n" +
-                        $"- Invalid redirect uri - ensure you have configured the following url in the application registration in Azure Portal: " +
-                        $"{GetExpectedRedirectUri(authenticationRequestParameters.AppConfig.ClientId)} \n";
+                        $" Internal Error Code: {internalErrorCode} \n";
```
gladjohn:
```suggestion
                        $" Internal Error Code: {internalErrorCode} \n" + 
                        $" See troubleshooting: https://aka.ms/msal-net-wam \n";
```

gladjohn (commented):
suggest adding a neutral pointer to troubleshooting docs instead of a speculative cause.

PR #5469, opened 2025-09-09
MSI V2 (IMDSv2): Identity‑scoped mTLS certificate cache

gladjohn:
This PR adds a process‑wide, identity‑scoped cache for the IMDSv2 mTLS client certificate, wires it into the MSI V2 flow

- First iteration: cache the certificate

- Next iteration: add tests, plug-in mTLS Authentication Operation to support Pop flows

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
@@ -21,10 +21,12 @@ namespace Microsoft.Identity.Client.ManagedIdentity
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
+        internal static ICertCache s_certCache = new ManagedIdentityCertificateCache();
 
         internal static void ResetSourceForTest()
         {
             s_sourceName = ManagedIdentitySource.None;
+            (s_certCache as ManagedIdentityCertificateCache)?.ClearForTest();
         }
 
         internal async Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityAsync(
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
@@ -1,0 +1,226 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.Internal;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Represents the per-identity binding used by IMDSv2 mTLS flows:
+    /// the issued client certificate plus the data needed to build the
+    /// token request without re-issuing on every call.
+    /// </summary>
+    internal sealed class CertCacheEntry
+    {
+        /// <summary>
+        /// The client certificate (with private key) used to authenticate over mTLS.
+        /// </summary>
+        public X509Certificate2 Certificate { get; private set; }
+
+        /// <summary>
+        /// UTC moment when the cache should proactively refresh this certificate.
+        /// </summary>
+        public DateTimeOffset RefreshAt { get; private set; }
+
+        /// <summary>
+        /// The authoritative client_id for the identity (SAMI/UAMI) as returned by IMDSv2.
+        /// </summary>
+        public string ClientId { get; private set; }
+
+        /// <summary>
+        /// The tenant to use when calling the token endpoint.
+        /// </summary>
+        public string TenantId { get; private set; }
+
+        /// <summary>
+        /// The base endpoint for OAuth2 token acquisition over mTLS.
+        /// </summary>
+        public string MtlsAuthenticationEndpoint { get; private set; }
+
+        /// <summary>
+        /// Initializes a new cache entry.
+        /// </summary>
+        public CertCacheEntry(
+            X509Certificate2 certificate,
+            DateTimeOffset refreshAt,
+            string clientId,
+            string tenantId,
+            string mtlsAuthenticationEndpoint)
+        {
+            Certificate = certificate ?? throw new ArgumentNullException(nameof(certificate));
+            RefreshAt = refreshAt;
+
+            ClientId = clientId ?? string.Empty;
+            TenantId = tenantId ?? string.Empty;
+            MtlsAuthenticationEndpoint = mtlsAuthenticationEndpoint ?? string.Empty;
+        }
+
+        /// <summary>
+        /// Forces a refresh on next access by setting <see cref="RefreshAt"/> to now.
+        /// Useful when the server hints that the certificate was revoked/bad.
+        /// </summary>
+        public void ForceRefreshNow() => RefreshAt = DateTimeOffset.UtcNow;
+    }
+
+    /// <summary>
+    /// Minimal abstraction for the certificate cache.
+    /// the certificate (CSR + /issuecredential + bind private key).
+    /// </summary>
+    internal interface ICertCache
+    {
+        /// <summary>
+        /// Gets a valid binding for <paramref name="clientId"/> or refreshes it if needed.
+        /// Exactly one caller per identity runs <paramref name="acquireFunc"/> under contention.
+        /// </summary>
+        Task<CertCacheEntry> GetAsync(
+            string clientId,
+            Func<CancellationToken, Task<CertCacheEntry>> acquireFunc,
+            RequestContext ctx,
+            CancellationToken ct);
+
+        /// <summary>
+        /// Marks the binding for <paramref name="clientId"/> as stale so that
+        /// the next <see cref="GetAsync"/> will refresh it.
+        /// </summary>
+        void Invalidate(string clientId, string reason);
+
+        /// <summary>
+        /// Clears all entries (test-only; production code should not call this).
+        /// </summary>
+        void ClearForTest();
+    }
+
+    /// <summary>
+    /// Process-wide, identity-scoped, thread-safe certificate cache,
+    /// race-proofed to ensure only one mint/rotation occurs per identity.
+    /// </summary>
+    internal sealed class ManagedIdentityCertificateCache : ICertCache
+    {
+        /// <summary>
+        /// Per-identity container that serializes mint/refresh via a gate.
+        /// </summary>
+        private sealed class Entry
+        {
+            private readonly SemaphoreSlim _gate = new (1, 1);
+            private CertCacheEntry _current;
+
+            internal Entry() { } // First caller will mint behind the gate.
+
+            /// <summary>
+            /// Returns the current binding if still fresh; otherwise mints/rotates exactly once.
+            /// </summary>
+            internal async Task<CertCacheEntry> GetOrRefreshAsync(
+                Func<CancellationToken, Task<CertCacheEntry>> acquireFunc,
+                RequestContext ctx,
+                CancellationToken ct)
+            {
+                var now = DateTimeOffset.UtcNow;
+                if (_current != null && now < _current.RefreshAt)
+                {
+                    return _current;
+                }
+
+                await _gate.WaitAsync(ct).ConfigureAwait(false);
+                try
+                {
+                    now = DateTimeOffset.UtcNow;
+                    if (_current != null && now < _current.RefreshAt)
+                    {
+                        return _current;
+                    }
+
+                    _current = await acquireFunc(ct).ConfigureAwait(false);
+                    ctx.Logger.Info($"[Managed Identity] mTLS cert minted/rotated; expires={_current.Certificate.NotAfter:u}");
+                    return _current;
+                }
+                finally
+                {
+                    _gate.Release();
+                }
+            }
+
+            /// <summary>
+            /// Forces a refresh by marking the current entry as expired.
+            /// </summary>
+            internal void Invalidate()
+            {
+                if (_current != null)
+                {
+                    _current.ForceRefreshNow();
+                }
+            }
+        }
+
+        // Identity map keyed by authoritative client_id (IMDSv2). Case-insensitive for GUID robustness.
+        private readonly ConcurrentDictionary<string, Entry> _entries =
+            new ConcurrentDictionary<string, Entry>(StringComparer.OrdinalIgnoreCase);
+
+        /// <inheritdoc />
+        public async Task<CertCacheEntry> GetAsync(
+            string clientId,
+            Func<CancellationToken, Task<CertCacheEntry>> acquireFunc,
+            RequestContext ctx,
+            CancellationToken ct)
+        {
+            if (string.IsNullOrEmpty(clientId))
+            {
+                clientId = string.Empty; // normalize
+            }
+
+            // Race-proof: create the Entry up-front; the Entry gate ensures only one cert mint occurs.
+            var entry = _entries.GetOrAdd(clientId, _ => new Entry());
+
+            return await entry.GetOrRefreshAsync(acquireFunc, ctx, ct).ConfigureAwait(false);
+        }
+
+        /// <inheritdoc />
+        public void Invalidate(string clientId, string reason)
+        {
+            if (string.IsNullOrEmpty(clientId))
+            {
+                clientId = string.Empty; // normalize
+            }
+
+            Entry entry;
+            if (_entries.TryGetValue(clientId, out entry))
+            {
+                entry.Invalidate();
+                // Optional: hook telemetry here with `reason`.
+            }
+        }
+
+        /// <inheritdoc />
+        public void ClearForTest() => _entries.Clear();
+
+        /// <summary>
+        /// Computes a proactive refresh point for <paramref name="cert"/>: 20% of lifetime
+        /// before expiry, capped at 24 hours. If already expired (or clock skew), returns now.
+        /// </summary>
+        public static DateTimeOffset ComputeRefreshAt(X509Certificate2 cert, DateTimeOffset now)
+        {
+            var notAfter = new DateTimeOffset(cert.NotAfter);
+            var lifetime = notAfter - now;
+            if (lifetime <= TimeSpan.Zero)
+            {
+                // Already expired or clock skew: force immediate refresh
+                return now;
+            }
+
+            // 20% of lifetime, capped at 24h
+            var margin = TimeSpan.FromTicks(lifetime.Ticks / 5);
+            var cap = TimeSpan.FromHours(24);
+            if (margin > cap)
+            {
+                margin = cap;
+            }
+
+            return notAfter - margin;
+        }
+    }
+}
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
@@ -260,20 +260,45 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
             var csrMetadata = await GetCsrMetadataAsync(_requestContext, false).ConfigureAwait(false);
             var (csr, privateKey) = _requestContext.ServiceBundle.Config.CsrFactory.Generate(csrMetadata.ClientId, csrMetadata.TenantId, csrMetadata.CuId);
 
-            var certificateRequestResponse = await ExecuteCertificateRequestAsync(csr).ConfigureAwait(false);
-            
-            // transform certificateRequestResponse.Certificate to x509 with private key
-            var mtlsCertificate = CommonCryptographyManager.AttachPrivateKeyToCert(
-                certificateRequestResponse.Certificate,
-                privateKey);
+            string clientId = csrMetadata.ClientId;
+
+            // Cert from cache (mint/rotate as needed)
+            var certificateRequestResponse = await ManagedIdentityClient.s_certCache.GetAsync(
+                clientId,
+                acquireFunc: async ct =>
+                {
+                    // New keypair + CSR
+                    var (csr, privateKey) = _requestContext.ServiceBundle.Config.CsrFactory.Generate(
+                        csrMetadata.ClientId, csrMetadata.TenantId, csrMetadata.CuId);
+
+                    // Issue client certificate
+                    var resp = await ExecuteCertificateRequestAsync(csr).ConfigureAwait(false);
+
+                    // Attach private key to returned cert
+                    var cert = CommonCryptographyManager.AttachPrivateKeyToCert(resp.Certificate, privateKey);
+
+                    // Compute proactive refresh time (<= 24h or 20% of lifetime)
+                    var refreshAt = ManagedIdentityCertificateCache.ComputeRefreshAt(cert, DateTimeOffset.UtcNow);
+
+                    // Cache entry carries endpoint + tenant + client_id for subsequent token calls
+                    return new CertCacheEntry(
+                        certificate: cert,
+                        refreshAt: refreshAt,
+                        clientId: resp.ClientId,
+                        tenantId: resp.TenantId,
+                        mtlsAuthenticationEndpoint: resp.MtlsAuthenticationEndpoint);
+                },
+                ctx: _requestContext,
+                ct: _requestContext.UserCancellationToken
+            ).ConfigureAwait(false);
 
             ManagedIdentityRequest request = new(HttpMethod.Post, new Uri($"{certificateRequestResponse.MtlsAuthenticationEndpoint}/{certificateRequestResponse.TenantId}{AcquireEntraTokenPath}"));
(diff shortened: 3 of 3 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        /// <inheritdoc />
+        public void ClearForTest() => _entries.Clear();
+
+        /// <summary>
+        /// Computes a proactive refresh point for <paramref name="cert"/>: 20% of lifetime
+        /// before expiry, capped at 24 hours. If already expired (or clock skew), returns now.
+        /// </summary>
+        public static DateTimeOffset ComputeRefreshAt(X509Certificate2 cert, DateTimeOffset now)
```
bgavrilMS: I assume you are injecting a DateTimeOffset here for testing? I think a better solution is to use ITimeService, which is already present in MSAL and provides a way to get the time.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -260,20 +260,45 @@ protected override async Task<ManagedIdentityRequest> CreateRequestAsync(string
...
+                },
+                ctx: _requestContext,
+                ct: _requestContext.UserCancellationToken
+            ).ConfigureAwait(false);
 
             ManagedIdentityRequest request = new(HttpMethod.Post, new Uri($"{certificateRequestResponse.MtlsAuthenticationEndpoint}/{certificateRequestResponse.TenantId}{AcquireEntraTokenPath}"));
             request.Headers.Add("x-ms-client-request-id", _requestContext.CorrelationId.ToString());
             request.BodyParameters.Add("client_id", certificateRequestResponse.ClientId);
-            request.BodyParameters.Add("grant_type", certificateRequestResponse.Certificate);
+            request.BodyParameters.Add("grant_type", ""); //TO DO : pass the correct grant type
```
bgavrilMS:
```suggestion
            request.BodyParameters.Add("grant_type", "client_credentials"); 
```

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+            }
+        }
+
+        // Identity map keyed by authoritative client_id (IMDSv2). Case-insensitive for GUID robustness.
+        private readonly ConcurrentDictionary<string, Entry> _entries =
+            new ConcurrentDictionary<string, Entry>(StringComparer.OrdinalIgnoreCase);
+
+        /// <inheritdoc />
+        public async Task<CertCacheEntry> GetAsync(
+            string clientId,
```
bgavrilMS: If you're building a cache, then this is not the right name. "Key" should be used.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -260,20 +260,45 @@ protected override async Task<ManagedIdentityRequest> CreateRequestAsync(string
...
-            // transform certificateRequestResponse.Certificate to x509 with private key
-            var mtlsCertificate = CommonCryptographyManager.AttachPrivateKeyToCert(
-                certificateRequestResponse.Certificate,
-                privateKey);
+            string clientId = csrMetadata.ClientId;
+
+            // Cert from cache (mint/rotate as needed)
+            var certificateRequestResponse = await ManagedIdentityClient.s_certCache.GetAsync(
+                clientId,
+                acquireFunc: async ct =>
```
bgavrilMS: Curious, do we really need a full blown cache (e.g. dictionary) here? Can't there be a 1 certificate associated with each ManagedIdentityApplication object?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+{
+    /// <summary>
+    /// Represents the per-identity binding used by IMDSv2 mTLS flows:
+    /// the issued client certificate plus the data needed to build the
+    /// token request without re-issuing on every call.
+    /// </summary>
+    internal sealed class CertCacheEntry
+    {
+        /// <summary>
+        /// The client certificate (with private key) used to authenticate over mTLS.
```
Robbie-Microsoft: I would say "The mTLS Certificate"

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+            string clientId,
+            string tenantId,
+            string mtlsAuthenticationEndpoint)
+        {
+            Certificate = certificate ?? throw new ArgumentNullException(nameof(certificate));
+            RefreshAt = refreshAt;
+
+            ClientId = clientId ?? string.Empty;
+            TenantId = tenantId ?? string.Empty;
+            MtlsAuthenticationEndpoint = mtlsAuthenticationEndpoint ?? string.Empty;
```
Robbie-Microsoft: Why are some of these optional? Should you also throw an exception if `refreshAt` wasn't provided?

Robbie-Microsoft: Can you add comments to the definitions above stating which fields are optional?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// Useful when the server hints that the certificate was revoked/bad.
+        /// </summary>
+        public void ForceRefreshNow() => RefreshAt = DateTimeOffset.UtcNow;
+    }
+
+    /// <summary>
+    /// Minimal abstraction for the certificate cache.
+    /// the certificate (CSR + /issuecredential + bind private key).
+    /// </summary>
+    internal interface ICertCache
```
bgavrilMS: nit: I am not sure a wrapper is needed? Can we just use ConcurrentDictionary?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -260,20 +260,45 @@ protected override async Task<ManagedIdentityRequest> CreateRequestAsync(string
             var csrMetadata = await GetCsrMetadataAsync(_requestContext, false).ConfigureAwait(false);
             var (csr, privateKey) = _requestContext.ServiceBundle.Config.CsrFactory.Generate(csrMetadata.ClientId, csrMetadata.TenantId, csrMetadata.CuId);
 
-            var certificateRequestResponse = await ExecuteCertificateRequestAsync(csr).ConfigureAwait(false);
-            
-            // transform certificateRequestResponse.Certificate to x509 with private key
-            var mtlsCertificate = CommonCryptographyManager.AttachPrivateKeyToCert(
-                certificateRequestResponse.Certificate,
-                privateKey);
+            string clientId = csrMetadata.ClientId;
```
bgavrilMS: This is the wrong Key. MSAL needs to pull the cert out of the cache withoput CSR metadata. So the only key is the managed identity ID (uami / sami)

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -260,20 +260,45 @@ protected override async Task<ManagedIdentityRequest> CreateRequestAsync(string
             var csrMetadata = await GetCsrMetadataAsync(_requestContext, false).ConfigureAwait(false);
```
bgavrilMS: If you have a cached cert, do you still need csrMetadata?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+            internal async Task<CertCacheEntry> GetOrRefreshAsync(
+                Func<CancellationToken, Task<CertCacheEntry>> acquireFunc,
+                RequestContext ctx,
+                CancellationToken ct)
+            {
+                var now = DateTimeOffset.UtcNow;
+                if (_current != null && now < _current.RefreshAt)
+                {
+                    return _current;
+                }
```
Robbie-Microsoft: Can you add a one-liner comment here?
`If the cache entry exists and it's not expired, return it`?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+
+                await _gate.WaitAsync(ct).ConfigureAwait(false);
+                try
+                {
+                    now = DateTimeOffset.UtcNow;
+                    if (_current != null && now < _current.RefreshAt)
+                    {
+                        return _current;
+                    }
```
Robbie-Microsoft: Can you add a one-liner comment here?
`After waiting for semaphore entry, if the cache entry exists and it's not expired, return it`?

Robbie-Microsoft: Should the if statement be extracted into a helper function since it's used multiple times?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+    {
+        /// <summary>
+        /// Gets a valid binding for <paramref name="clientId"/> or refreshes it if needed.
+        /// Exactly one caller per identity runs <paramref name="acquireFunc"/> under contention.
+        /// </summary>
+        Task<CertCacheEntry> GetAsync(
+            string clientId,
+            Func<CancellationToken, Task<CertCacheEntry>> acquireFunc,
+            RequestContext ctx,
+            CancellationToken ct);
```
Robbie-Microsoft: Can you please add XML for each parameter so that there can be more clarity? Especially `acquireFunc`.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -260,20 +260,45 @@ protected override async Task<ManagedIdentityRequest> CreateRequestAsync(string
...
+                    // Cache entry carries endpoint + tenant + client_id for subsequent token calls
+                    return new CertCacheEntry(
+                        certificate: cert,
+                        refreshAt: refreshAt,
+                        clientId: resp.ClientId,
+                        tenantId: resp.TenantId,
+                        mtlsAuthenticationEndpoint: resp.MtlsAuthenticationEndpoint);
+                },
+                ctx: _requestContext,
+                ct: _requestContext.UserCancellationToken
```
Robbie-Microsoft: Why can't you only pass in _requestContext? You're passing it in twice here.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
...
+    /// Represents the per-identity binding used by IMDSv2 mTLS flows:
+    /// the issued client certificate plus the data needed to build the
+    /// token request without re-issuing on every call.
+    /// </summary>
+    internal sealed class CertCacheEntry
+    {
+        /// <summary>
+        /// The client certificate (with private key) used to authenticate over mTLS.
+        /// </summary>
+        public X509Certificate2 Certificate { get; private set; }
```
Robbie-Microsoft: Can we rename this to `MtlsCertificate` for clarity?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
```diff
@@ -0,0 +1,226 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
Robbie-Microsoft: Should this file be renamed to `MtlsCertificateCache` for clarity?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
bgavrilMS: @Robbie-Microsoft - you understand object lifetime better. There should be 1 cert per each ManagedIdentityApplication. Can this be added somewhere?

gladjohn: per offline discussion, we cannot as we need the cache to be static.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
gladjohn: changed to use ITimeService

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
gladjohn: modified

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
gladjohn: sure, just using a `ConcurrentDictionary` now

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: yes, because the cert data comes from `/issuecredential` endpoint and we also get the `sSTS-R mTLS endpoint`, `client_id` from the same endpoint.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCache.cs
gladjohn: While addressing an earlier comment, I deleted this file `CertificateCache.cs`. So, closing all comments that are no longer relevant.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: removed it, as we do not use 'CertificateCache.cs` anymore

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -60,6 +60,18 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             MsalAccessTokenCacheItem cachedAccessTokenItem = await GetCachedAccessTokenAsync()
                 .ConfigureAwait(false);
 
+            // If using IMDSv2 (mTLS) and the mTLS cert is near expiry, skip the AT cache
+            if (cachedAccessTokenItem != null && ManagedIdentityClient.s_sourceName == ManagedIdentitySource.ImdsV2)
```
bgavrilMS, line 64: 1. The condition here does not match the comment. You should only look for a certificate if an mtls pop token is requested
2. Let's please avoid peppering the codebase with if (static.sourceName == IMDSv2). First of all, it creates possible race conditions, because you don't know if the source has been evaluated (which is the case here, why would I care what the source is if I have a cached token?). Second, because there should be 1 place (Single Responsability Principle) that decides if MTLS POP is supported or not (e.g. each of the MSI sources should say this)

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -60,6 +60,18 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+            {
+                string identityKey = ServiceBundle.Config.ClientId;
```
bgavrilMS, line 66: Naming is confusing. avoid using `identity` as it has too many meanings. How about `mtlsCertCacheKey` ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -152,10 +163,38 @@ private static bool ValidateAzureArcEnvironment(string identityEndpoint, string
...
-            } 
-            
+            }
+
             logger?.Verbose(() => "[Managed Identity] Azure Arc managed identity is not available.");
             return false;
         }
+
+        // Identity-aware check over the mTLS cert cache
+        internal static bool IsMtlsCertExpiringSoon(string identityKey)
```
bgavrilMS, line 173: THis is not a responsability of the ManagedIdentityClient.

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -60,6 +60,18 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
...
+
+                if (ManagedIdentityClient.IsMtlsCertExpiringSoon(identityKey))
```
bgavrilMS, line 68: How about smth like `if (cert = cache.GetCertifiacate(key) && ~cert.IsExpiringSoon())`? This will avoid having to expose a new complex API that does 2 things at once.

Alternatively, the cache.GetCertificate() can simply return NULL if the certificate is close to expiry. This is what we do for tokens.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -21,10 +25,17 @@ internal class ManagedIdentityClient
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
+        internal static ITimeService s_timeService = new TimeService();
+        internal static readonly TimeSpan MtlsCertRefreshSkew = TimeSpan.FromMinutes(5);
+
+        internal static readonly ConcurrentDictionary<string,
+            (X509Certificate2 Cert, string ClientId, string TenantId, string Endpoint)> s_miCerts = new();
 
         internal static void ResetSourceForTest()
```
bgavrilMS, line 34: Name no longer matches intent

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -21,10 +25,17 @@ internal class ManagedIdentityClient
...
         {
             s_sourceName = ManagedIdentitySource.None;
+            s_miCerts.Clear();
+            s_timeService = new TimeService();
```
bgavrilMS, line 38: The point of the time service is that it should be a singleton. A single instance is injected in all classes that deal with time. That instance can either be the real time or a test time. Instead, you are using many instances, so you are not really sure what you test.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -21,10 +25,17 @@ internal class ManagedIdentityClient
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
+        internal static ITimeService s_timeService = new TimeService();
```
bgavrilMS, line 28: internal variables are not prefixed with s_ (see MtlsCertRefreshSkew )

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -152,10 +163,38 @@ private static bool ValidateAzureArcEnvironment(string identityEndpoint, string
...
+            {
+                // No cert minted yet -> nothing to refresh proactively
+                return false;
+            }
+
+            return IsCertExpiringSoon(binding.Cert, s_timeService.GetUtcNow(), MtlsCertRefreshSkew);
+        }
+
+        // Test-only helpers
+        internal static bool IsCertExpiringSoon(X509Certificate2 cert, DateTime nowUtc, TimeSpan skew)
```
bgavrilMS, line 185: Why did you bother introducing a test service then?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -152,10 +163,38 @@ private static bool ValidateAzureArcEnvironment(string identityEndpoint, string
...
+        {
+            if (cert == null)
+            {
+                return true;
+            }
+            DateTime notAfterUtc = cert.NotAfter.ToUniversalTime();
+            return nowUtc >= (notAfterUtc - skew);
+        }
+
+        internal static void SetTimeServiceForTest(ITimeService timeService) /* internal - test only usage */
```
bgavrilMS, line 195: It's already a internal property, why do you need an extra setter?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -21,10 +25,17 @@ internal class ManagedIdentityClient
         private const string WindowsHimdsFilePath = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
+        internal static ITimeService s_timeService = new TimeService();
+        internal static readonly TimeSpan MtlsCertRefreshSkew = TimeSpan.FromMinutes(5);
+
+        internal static readonly ConcurrentDictionary<string,
```
bgavrilMS, line 31: have you considered moving the cache elsewhere? It seems that this class doesn't even use it?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -257,23 +259,41 @@ private async Task<CertificateRequestResponse> ExecuteCertificateRequestAsync(st
...
 
-            var certificateRequestResponse = await ExecuteCertificateRequestAsync(csr).ConfigureAwait(false);
-            
-            // transform certificateRequestResponse.Certificate to x509 with private key
-            var mtlsCertificate = CommonCryptographyManager.AttachPrivateKeyToCert(
-                certificateRequestResponse.Certificate,
-                privateKey);
+            // 2) Try cached binding; refresh if cert expires within 5 minutes
+            if (!ManagedIdentityClient.s_miCerts.TryGetValue(key, out var binding) ||
+                ManagedIdentityClient.IsMtlsCertExpiringSoon(key))
```
bgavrilMS, line 268: another example why `IsMtlsCertExpiringSoon` is a bad API. You perform a cache GET twice even though you have your value.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -21,10 +25,17 @@ internal class ManagedIdentityClient
...
+            (X509Certificate2 Cert, string ClientId, string TenantId, string Endpoint)> s_miCerts = new();
```
bgavrilMS, line 32: You are caching a lot more than certs, so I am not sure this is a good name.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: But I agree that a big refactor here could take time, maybe add a task to use TimeService everywhere in MSAL where DateTime.now is used.

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -60,6 +60,18 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
             MsalAccessTokenCacheItem cachedAccessTokenItem = await GetCachedAccessTokenAsync()
                 .ConfigureAwait(false);
 
+            // If using IMDSv2 (mTLS) and the mTLS cert is near expiry, skip the AT cache
```
bgavrilMS, line 63: In this file, you do not take care of requesting a POP token form the cache (for which an IAuthenticationOperation is needed)

PR #5489, opened 2025-09-19
Add ResetStateForTest API for test

bgavrilMS:
- Reset API for test
- Add static cache logic
- Refactor tests to use new API

Note: this is a large PR due to refactoring of existing tests. Mostly not worth reviewing. Pls review only product code and code on which I have added comments.

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
@@ -178,7 +178,7 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
 
                 #region Identity 2
                 UserAssignedIdentityId identity2Type = userAssignedIdentityId; // keep the same type, that's the most common scenario
-                string identity2Id = TestConstants.ClientId2;
+                string identity2Id = "some_other_id";
                 var managedIdentityApp2 = await CreateManagedIdentityAsync(httpManager, identity2Type, identity2Id, addProbeMock: false, addSourceCheck: false).ConfigureAwait(false); // source is already cached
 
                 AddMocksToGetEntraToken(httpManager, identity2Type, identity2Id);
@@ -324,7 +324,7 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
 
                 #region Identity 2
                 UserAssignedIdentityId identity2Type = userAssignedIdentityId; // keep the same type, that's the most common scenario
-                string identity2Id = TestConstants.ClientId2;
+                string identity2Id = "some_other_id";
                 var managedIdentityApp2 = await CreateManagedIdentityAsync(
                     httpManager,
                     identity2Type,


src/client/Microsoft.Identity.Client/Instance/Discovery/INetworkCacheMetadataProvider.cs
```diff
@@ -9,6 +9,5 @@ internal interface INetworkCacheMetadataProvider
     {
         void AddMetadata(string environment, InstanceDiscoveryMetadataEntry entry);
         InstanceDiscoveryMetadataEntry GetMetadata(string environment, ILoggerAdapter logger);
-        void /* for test purposes */ Clear();
```
bgavrilMS, line 12: I refactored this because it's just another static cache and so an instance method is not appropriate.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -175,9 +175,11 @@ public async Task BearerTokenTokenIsPerIdentity(
                 #endregion Identity 1
 
                 #region Identity 2
-                var managedIdentityApp2 = await CreateManagedIdentityAsync(httpManager, userAssignedIdentityId, userAssignedId, addProbeMock: false, addSourceCheck: false).ConfigureAwait(false); // source is already cached
```
bgavrilMS, line 180: @Robbie-Microsoft - this test was wrong. It was not using Identity2 as per the comments, but the same identity. It wored just because you had set `miBuilder.Config.AccessorOptions = null;` which nuked the cache in the middle (line 83), but that's not a valid gesture, i.e. app developers can't do that. Resetting the cache should only happen in-between tests.

gladjohn (approved):

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -175,9 +175,11 @@ public async Task BearerTokenTokenIsPerIdentity(
...
+                UserAssignedIdentityId identity2Type = UserAssignedIdentityId.ClientId;
+                string identity2Id = TestConstants.ClientId2;
```
Robbie-Microsoft: Set `identity2Type` equal to the `userAssignedIdentityId` param from the DataRow.

You'll need to include`identity2Id` in the DataRow so that a ResourceId and ObjectId can be used instead of ClientId every time.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -316,9 +318,16 @@ public async Task mTLSPopTokenTokenIsPerIdentity(
                 #endregion Identity 1
 
                 #region Identity 2
-                var managedIdentityApp2 = await CreateManagedIdentityAsync(httpManager, userAssignedIdentityId, userAssignedId, addProbeMock: false, addSourceCheck: false).ConfigureAwait(false); // source is already cached
-
-                AddMocksToGetEntraToken(httpManager, userAssignedIdentityId, userAssignedId/*, mTLSPop: true*/); // TODO: implement mTLS Pop
+                UserAssignedIdentityId identity2Type = UserAssignedIdentityId.ClientId;
+                string identity2Id = TestConstants.ClientId2;
```
Robbie-Microsoft: see above

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
bgavrilMS: Done for the type. 

I don't think the ID needs to set to some specific value based on the type. The ID is just a string, so as long as it's different from Identity1, it fits the purpose of this test.

Robbie-Microsoft: Sure. But maybe we should rename the constant to something else, because using `TestConstants.ClientId2` for the`resourceId` and `objectId` types is confusing. Could you at least include your above comment in the code? 

> The ID is just a string, so as long as it's different from Identity1, it fits the purpose of this test.

bgavrilMS: Done.

PR #5494, opened 2025-09-22
ImdsV2: Entra Request: "expires_in" is now lumped into "expires_on"

Robbie-Microsoft:
- Manually tested on tip node
- All existing unit tests pass

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
@@ -34,14 +34,14 @@ namespace Microsoft.Identity.Client.ManagedIdentity
         public string ExpiresOn { get; set; }  // The actual property consumers use
 
         [JsonProperty("expires_on")]
-        public string ExpiresOnRaw             // Proxy for "expires_on" JSON field
+        private string ExpiresOnRaw             // Proxy for "expires_on" JSON field
         {
             get => ExpiresOn;                  // When serializing, return ExpiresOn value
             set => ExpiresOn = value;          // When deserializing, store in ExpiresOn
         }
 
         [JsonProperty("expires_in")]
-        public string ExpiresInRaw             // Proxy for "expires_in" JSON field
+        private string ExpiresInRaw             // Proxy for "expires_in" JSON field
         {
             get => null;                       // Never serialize this (return null)
             set => ExpiresOn = value;          // When deserializing, store in ExpiresOn
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
@@ -5,6 +5,7 @@ using System;
 #if SUPPORTS_SYSTEM_TEXT_JSON
 using Microsoft.Identity.Client.Platforms.net;
 using JsonProperty = System.Text.Json.Serialization.JsonPropertyNameAttribute;
+using JsonIgnore = System.Text.Json.Serialization.JsonIgnoreAttribute;
 #else
 using Microsoft.Identity.Json;
 #endif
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Utils/DateTimeHelpers.cs
```diff
@@ -85,7 +85,15 @@ public static long GetDurationFromManagedIdentityTimestamp(string dateTimeStamp)
             // Example: "1697490590" (Unix timestamp representing seconds since 1970-01-01)
             if (long.TryParse(dateTimeStamp, out long expiresOnUnixTimestamp))
             {
-                return expiresOnUnixTimestamp - DateTimeHelpers.CurrDateTimeInUnixTimestamp();
+                var timestamp = expiresOnUnixTimestamp - DateTimeHelpers.CurrDateTimeInUnixTimestamp();
```
gladjohn, line 88: should we also update our mocks? so these are unit tested

Robbie-Microsoft: I don't understand. It's still unit tested. All existing tests pass.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
```diff
@@ -5,6 +5,7 @@
 #if SUPPORTS_SYSTEM_TEXT_JSON
 using Microsoft.Identity.Client.Platforms.net;
```
bgavrilMS, line 6: This using should be outside the the "if" ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
```diff
@@ -29,8 +30,22 @@ internal class ManagedIdentityResponse
         /// </summary>
```
bgavrilMS, line 30: How will we keep this object `ManagedIdentityResponse` in sync with `MsalTokenResponse`, e.g. `RefreshIn` is missing?

src/client/Microsoft.Identity.Client/Utils/DateTimeHelpers.cs
```diff
@@ -85,7 +85,15 @@ public static long GetDurationFromManagedIdentityTimestamp(string dateTimeStamp)
...
+
+                // If the timestamp is negative, return the original expiresOnUnixTimestamp. It's format is "seconds from now".
```
bgavrilMS:
```suggestion
                // If the timestamp is negative, return the original expiresOnUnixTimestamp. Its format is "seconds from now".
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
```diff
@@ -29,8 +30,22 @@ internal class ManagedIdentityResponse
...
         /// <remarks>The date is represented as the number of seconds from "1970-01-01T0:0:0Z UTC" 
         /// (corresponds to the token's exp claim).</remarks>
+        [JsonIgnore]
+        public string ExpiresOn { get; set; }  // The actual property consumers use
```
bgavrilMS, line 34: No tests?

src/client/Microsoft.Identity.Client/Utils/DateTimeHelpers.cs
gladjohn: given that ests now returns `expires_in`, our mocks need to be updated as well?

Robbie-Microsoft: yes, done

src/client/Microsoft.Identity.Client/Utils/DateTimeHelpers.cs
Robbie-Microsoft: thx

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
Robbie-Microsoft: All existing tests cover this by changing the mock to send "expires_in" instead of "expires_on" for ImdsV2 entra token request

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityResponse.cs
Robbie-Microsoft: It should be inside where it is, it's only available in net8.0

PR #5502, opened 2025-09-25
ImdsV2: Created Factory to replace Imds GetValidatedEndpoint

Robbie-Microsoft:
The `UnsupportedManagedIdentitySource_ThrowsExceptionDuringTokenAcquisition` in `ManagedIdentityTests.cs` was breaking for the Imds source.

The test sets the Imds endpoint to "unsupported://endpoint". This breaks the probe request before the acquireToken request has a chance to be (correctly) broken by this bad endpoint.

I've added a Factory so we can inject a custom endpoint into the probe request, for testing purposes.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
@@ -178,27 +178,31 @@ namespace Microsoft.Identity.Client.ManagedIdentity
         public static Uri GetValidatedEndpoint(
             ILoggerAdapter logger,
             string subPath,
-            string queryParams = null
+            string queryParams = null,
+            string existingImdsBaseEndpoint = null
             )
         {
-            UriBuilder builder;
-
-            if (!string.IsNullOrEmpty(EnvironmentVariables.PodIdentityEndpoint))
+            string baseEndpoint = null;
+            if (!string.IsNullOrEmpty(existingImdsBaseEndpoint))
+            {
+                logger.Verbose(() => "[Managed Identity] Using IMDS endpoint from an instance of IValidatedProbeEndpointFactory : " + existingImdsBaseEndpoint);
+                baseEndpoint = existingImdsBaseEndpoint;
+            }
+            else if (!string.IsNullOrEmpty(EnvironmentVariables.PodIdentityEndpoint))
             {
                 logger.Verbose(() => "[Managed Identity] Environment variable AZURE_POD_IDENTITY_AUTHORITY_HOST for IMDS returned endpoint: " + EnvironmentVariables.PodIdentityEndpoint);
-                builder = new UriBuilder(EnvironmentVariables.PodIdentityEndpoint)
-                {
-                    Path = subPath
-                };
+                baseEndpoint = EnvironmentVariables.PodIdentityEndpoint;
             }
             else
             {
                 logger.Verbose(() => "[Managed Identity] Unable to find AZURE_POD_IDENTITY_AUTHORITY_HOST environment variable for IMDS, using the default endpoint.");
-                builder = new UriBuilder(DefaultImdsBaseEndpoint)
-                {
-                    Path = subPath
-                };
+                baseEndpoint = DefaultImdsBaseEndpoint;
             }
+
+            UriBuilder builder = new UriBuilder(baseEndpoint)
+            {
+                Path = subPath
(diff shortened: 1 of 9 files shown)


tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -1176,23 +1177,22 @@ public async Task UnsupportedManagedIdentitySource_ThrowsExceptionDuringTokenAcq
             string UnsupportedEndpoint = "unsupported://endpoint";
 
             using (new EnvVariableContext())
+            using (var httpManager = new MockHttpManager())
```
gladjohn, line 1129: simply adding this line fixes the issue for me

```
// Disabling shared cache options to avoid cross test pollution.
miBuilder.Config.AccessorOptions = null;
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -178,28 +178,32 @@ internal static string CreateRequestFailedMessage(HttpResponse response, string
         public static Uri GetValidatedEndpoint(
             ILoggerAdapter logger,
             string subPath,
-            string queryParams = null
+            string queryParams = null,
+            string existingImdsBaseEndpoint = null
```
bgavrilMS, line 182: This seems like a test only parameter? But you already have a factory pattern, so the test factory should not depend on product code.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -25,8 +25,8 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
     [TestClass]
     public class ImdsV2Tests : TestBase
     {
+        private readonly TestCsrFactory _testCsrFactory = new TestCsrFactory();
```
bgavrilMS, line 28: Consider using NSubstitute, the mock framework already used by MSAL.NET, to mock simple objects like this.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -178,28 +178,32 @@ internal static string CreateRequestFailedMessage(HttpResponse response, string
         public static Uri GetValidatedEndpoint(
```
bgavrilMS, line 178: The logic in this class seems to depend on a few env variables. Can you help me understand why we need to introduce a new factory to make this testable? 

My concern is that we are introducing too many abstractions, which are hard to read / reason about. Creating hardware keys and getting attestation tokens are great places where factories should be used, because they encapsulate complex logic. But this seems to just compute a url ?

bgavrilMS (changes requested):
It's not clear why a factory is needed, when the logic depends only on env variables.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
Robbie-Microsoft: That won't fix the issue though because we need a mocked probe failure so that the source can correctly become Imds.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: The problem is that we need the environment variable to change in the middle of the test. There's no way to accomplish that right now, as far as I know.

Robbie-Microsoft: Please refer to the description of this PR.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
Robbie-Microsoft: Also, Bogdan removed all instances of that with his `ResetStateForTest` PR

Robbie-Microsoft:
Discarding PR

PR #5508, opened 2025-09-30
ImdsV2: Fixed Broken "xxxTokenTokenIsPerIdentity" Unit Test

Robbie-Microsoft:
We're now resetting the application state in the (Bearer/mTLSPop)TokenTokenIsPerIdentity unit test since there's two applications. Otherwise, the second application uses the first application's cached token.

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
@@ -49,6 +49,8 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
         public const string Bearer = "Bearer";
         public const string MTLSPoP = "mtls_pop";
 
+        private const string altUserAssignedId = "some_other_id";
+
         private void AddMocksToGetEntraToken(
             MockHttpManager httpManager,
             UserAssignedIdentityId userAssignedIdentityId = UserAssignedIdentityId.None,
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -209,12 +212,13 @@ public async Task BearerTokenTokenIsPerIdentity(
                 Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
                 #endregion Identity 1
 
+                ApplicationBase.ResetStateForTest();
```
bgavrilMS, line 215: I don't think you should call ResetState in the middle of the test. It's not a realistic scenario.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -177,13 +179,14 @@ public async Task BearerTokenHappyPath(
         }
 
         [DataTestMethod]
-        [DataRow(UserAssignedIdentityId.None, null)]                             // SAMI
-        [DataRow(UserAssignedIdentityId.ClientId, TestConstants.ClientId)]       // UAMI
-        [DataRow(UserAssignedIdentityId.ResourceId, TestConstants.MiResourceId)] // UAMI
-        [DataRow(UserAssignedIdentityId.ObjectId, TestConstants.ObjectId)]       // UAMI
+        [DataRow(UserAssignedIdentityId.None, null, null)]                                        // SAMI
```
bgavrilMS, line 182: There is only 1 SAMI. You can have 1 SAMI and 1 UAMI, but not 2 x SAMI

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -177,13 +179,14 @@ public async Task BearerTokenHappyPath(
...
+        [DataRow(UserAssignedIdentityId.ClientId, TestConstants.ClientId, altUserAssignedId)]       // UAMI
```
bgavrilMS, line 183: nit: I find the test harder to read now

PR #5514, opened 2025-10-02
ImdsV2: Added Caching for the IMDS Endpoint Env Variable + Improved Unit Tests

Robbie-Microsoft:
- Added caching so the method that builds a URI for IMDSv1 and IMDSv2 only checks the environment variable once, the first time
    - this removes redundant logs
- Improved IMDSv2 unit tests so that the IMDS test environment variable is now used for IMDSv2

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
@@ -34,6 +34,8 @@ namespace Microsoft.Identity.Client.ManagedIdentity
 
         private readonly Uri _imdsEndpoint;
 
+        private static string s_cachedBaseEndpoint = null;
+
         internal ImdsManagedIdentitySource(RequestContext requestContext) : 
             base(requestContext, ManagedIdentitySource.Imds)
         {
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -34,6 +34,8 @@ internal class ImdsManagedIdentitySource : AbstractManagedIdentity
 
         private readonly Uri _imdsEndpoint;
 
+        private static string s_cachedBaseEndpoint = null;
```
bgavrilMS, line 37: Why would we cache env variables?!

https://martinfowler.com/bliki/TwoHardThings.html

bgavrilMS (changes requested):
No env variable caching.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: @bgavrilMS Can you suggest an alternative approach? We don't want this log to show 3 times per initial ImdsV2 token request, and an additional two times for each subsequent token request. Unless we DO want to show the log? I had thought that the environment variable wouldn't change after app initialization, but I'm seeing now that that could be an incorrect assumption.

gladjohn: Instead of caching, use a `log‑once guard` so we still read the env var on every call, but we only emit the “which endpoint are we using?” message once per process.

gladjohn: something like

```cs
private static int s_loggedImdsEndpoint; // 0 = not logged, 1 = logged

var envEndpoint = EnvironmentVariables.PodIdentityEndpoint; // read every time
var baseEndpoint = string.IsNullOrEmpty(envEndpoint) ? DefaultImdsBaseEndpoint : envEndpoint;

// Log the decision once to avoid noise, but don't cache the value.
if (System.Threading.Interlocked.Exchange(ref s_loggedImdsEndpoint, 1) == 0)
{
    if (string.IsNullOrEmpty(envEndpoint))
    {
        logger.Verbose(() => "[Managed Identity] Using default IMDS endpoint (AZURE_POD_IDENTITY_AUTHORITY_HOST not set).");
    }
    else
    {
        logger.Verbose(() => "[Managed Identity] Using IMDS endpoint from AZURE_POD_IDENTITY_AUTHORITY_HOST: " + envEndpoint);
    }
}

var builder = new UriBuilder(baseEndpoint) { Path = subPath };
```

will this work?

bgavrilMS: @Robbie-Microsoft - I re-read your code. I was confused by your PR description "cache env variables", but in reality you want to cache the endpoint, which is an algorithm based on env variables. This makes sense.

bgavrilMS (approved):

PR #5531, opened 2025-10-14
ImdsV2: Non-mTLS POP Requests Fall Back to ImdsV1

Robbie-Microsoft:
- If an ImdsV2 mTLS PoP token is requested but the ImdsV2 endpoints are not available, an error will be thrown.
- If the app has already returned an ImdsV1 bearer token, an ImdsV2 mTLS PoP cannot be returned unless the app is restarted.

### DIFF
--- a/tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsTests.cs
+++ b/tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsTests.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.AppConfig;
+using Microsoft.Identity.Client.MtlsPop;
 using Microsoft.Identity.Test.Common.Core.Helpers;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
@@ -37,11 +38,11 @@ namespace Microsoft.Identity.Test.E2E
         [RunOnAzureDevOps]
         [TestCategory("MI_E2E_Imds")]
         [DataTestMethod]
-        [DataRow(null /*SAMI*/, null, DisplayName = "SAMI")]
-        [DataRow("4b7a4b0b-ecb2-409e-879a-1e21a15ddaf6", "clientid", DisplayName = "UAMI-ClientId")]
+        [DataRow(null /*SAMI*/, null, DisplayName = "AcquireToken_OnImds_Succeeds-SAMI")]
+        [DataRow("4b7a4b0b-ecb2-409e-879a-1e21a15ddaf6", "clientid", DisplayName = "AcquireToken_OnImds_Succeeds-UAMI-ClientId")]
         [DataRow("/subscriptions/c1686c51-b717-4fe0-9af3-24a20a41fb0c/resourcegroups/MSAL_MSI/providers/Microsoft.ManagedIdentity/userAssignedIdentities/LabVaultAccess_UAMI",
-         "resourceid", DisplayName = "UAMI-ResourceId")]
-        [DataRow("1eee55b7-168a-46be-8d19-30e830ee9611", "objectid", DisplayName = "UAMI-ObjectId")]
+         "resourceid", DisplayName = "AcquireToken_OnImds_Succeeds-UAMI-ResourceId")]
+        [DataRow("1eee55b7-168a-46be-8d19-30e830ee9611", "objectid", DisplayName = "AcquireToken_OnImds_Succeeds-UAMI-ObjectId")]
         public async Task AcquireToken_OnImds_Succeeds(string id, string idType)
         {
             var mi = BuildMi(id, idType);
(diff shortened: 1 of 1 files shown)


src/client/Microsoft.Identity.Client.MtlsPop/ManagedIdentityPopExtensions.cs
```diff
@@ -24,6 +20,22 @@ public static class ManagedIdentityPopExtensions
         public static AcquireTokenForManagedIdentityParameterBuilder WithMtlsProofOfPossession(
             this AcquireTokenForManagedIdentityParameterBuilder builder)
         {
+            void MtlsNotSupportedForManagedIdentity(string message)
```
Robbie-Microsoft, line 23: This logic was previously in the wrong file

bgavrilMS:
> All existing unit tests pass.
> 
> Still requires manual testing.
> 
> Currently having a problem in new unit test, where after a bearer token is requested, requesting an mTLS PoP token will return the bearer token from the cache.

This is because we need Gladwin's big PR around AuthenticationOperation. mtls pop tokens aren't cached properly without it.

gladjohn:
> > All existing unit tests pass.
> > Still requires manual testing.
> > Currently having a problem in new unit test, where after a bearer token is requested, requesting an mTLS PoP token will return the bearer token from the cache.
> 
> This is because we need Gladwin's big PR around AuthenticationOperation. mtls pop tokens aren't cached properly without it.

@Robbie-Microsoft I have a PR out that adds the token type and also adds the cert to the auth result. That will unblock you.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -143,151 +143,6 @@ private async Task<IManagedIdentityApplication> CreateManagedIdentityAsync(
         }
 
         #region Acceptance Tests
-        #region Bearer Token Tests
```
gladjohn, line 146: we should not delete these test cases. Is there a way to add a env_variable to enable MSI v2 on-demand and run these? so we are testing the bearer path as we make changes?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -316,12 +316,10 @@ protected override async Task<ManagedIdentityRequest> CreateRequestAsync(string
             request.Headers.Add(ThrottleCommon.ThrottleRetryAfterHeaderName, ThrottleCommon.ThrottleRetryAfterHeaderValue);
             request.Headers.Add(OAuth2Header.RequestCorrelationIdInResponse, "true");
 
-            var tokenType = _isMtlsPopRequested ? "mtls_pop" : "bearer";
-
             request.BodyParameters.Add("client_id", certificateRequestResponse.ClientId);
             request.BodyParameters.Add("grant_type", OAuth2GrantType.ClientCredentials);
             request.BodyParameters.Add("scope", resource.TrimEnd('/') + "/.default");
-            request.BodyParameters.Add("token_type", tokenType);
+            request.BodyParameters.Add("token_type", "mtls_pop");
```
gladjohn, line 322: Suggest we keep both bearer and mtls_pop for the time being, and we enable MSI v2 bearer based on a feature flag.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -47,15 +48,14 @@ private static readonly Func<AttestationTokenInput, CancellationToken, Task<Atte
...
+        public const string Bearer = "Bearer"; 
         public const string MTLSPoP = "mtls_pop";
 
         private void AddMocksToGetEntraToken(
             MockHttpManager httpManager,
             UserAssignedIdentityId userAssignedIdentityId = UserAssignedIdentityId.None,
             string userAssignedId = null,
-            string certificateRequestCertificate = TestConstants.ValidRawCertificate,
-            bool mTLSPop = false)
+            string certificateRequestCertificate = TestConstants.ValidRawCertificate)
```
gladjohn, line 57: This valid RAW certificate is expired. Can we extend this to 20 years or something?

Robbie-Microsoft: This is not expired, I think you misread it. It expires in 100 years from last August.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: resolved offline

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: resolved offline

PR #5533, opened 2025-10-15
[MSI V2] - Adds Authentication Operation to MSI v2 mTLS flows 

gladjohn:
Fixes - 

•	Purpose: Persist and reuse mTLS binding certificate to enable cache lookups under mtls_pop scheme.
•	Non-goals: Certificate rotation strategy, multi-cert pool, persistence across processes.
•	Follow-up items: (a) Coalescing concurrent first-mint, (b) Attaching binding cert to result (if not in this PR), (c) Tests enabling previously skipped assertions.
•	Risk assessment: Low; internal-only; no behavior change for bearer path.

**Testing**
unit tests

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
@@ -45,14 +45,13 @@ namespace Microsoft.Identity.Client.Internal.Requests
             // Prime the scheme before any cache lookup if we already have a binding cert from a prior mint
             if (AuthenticationRequestParameters.IsMtlsPopRequested)
             {
-                var priorCert = _managedIdentityClient.RuntimeMtlsBindingCertificate;
-                if (priorCert != null)
+                if (_managedIdentityClient.RuntimeMtlsBindingCertificate != null)
                 {
-                    AuthenticationRequestParameters.AuthenticationScheme = new MtlsPopAuthenticationOperation(priorCert);
+                    AuthenticationRequestParameters.AuthenticationScheme = new MtlsPopAuthenticationOperation(_managedIdentityClient.RuntimeMtlsBindingCertificate);
 
                     logger.Info("[ManagedIdentity] Using prior mTLS binding certificate for cache lookup.");
                     logger.InfoPii(
-                        () => $"[ManagedIdentity][PII] Prior mTLS cert thumbprint: {priorCert.Thumbprint}",
+                        () => $"[ManagedIdentity][PII] Prior mTLS cert thumbprint: {_managedIdentityClient.RuntimeMtlsBindingCertificate.Thumbprint}",
                         () => "[ManagedIdentity][PII] Prior mTLS cert thumbprint: ***");
                 }
             }


src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -39,6 +41,18 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
         {
             AuthenticationResult authResult = null;
             ILoggerAdapter logger = AuthenticationRequestParameters.RequestContext.Logger;
+            
+            // Prime the scheme before any cache lookup if we already have a binding cert from a prior mint
+            if (AuthenticationRequestParameters.IsMtlsPopRequested &&
+                !(AuthenticationRequestParameters.AuthenticationScheme is MtlsPopAuthenticationOperation))
+            {
+                var priorCert = _managedIdentityClient.RuntimeMtlsBindingCertificate;
```
Robbie-Microsoft: What if `_managedIdentityClient.RuntimeMtlsBindingCertificate` is null? Should this be inside the if statement?

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -39,6 +41,18 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
         {
             AuthenticationResult authResult = null;
             ILoggerAdapter logger = AuthenticationRequestParameters.RequestContext.Logger;
+            
+            // Prime the scheme before any cache lookup if we already have a binding cert from a prior mint
+            if (AuthenticationRequestParameters.IsMtlsPopRequested &&
+                !(AuthenticationRequestParameters.AuthenticationScheme is MtlsPopAuthenticationOperation))
```
Robbie-Microsoft: Can you please explain to me why this 2nd check is needed? If `AuthenticationRequestParameters.IsMtlsPopRequested` is true, then why do we care if the authentication scheme is not `MtlsPopAuthenticationOperation`?

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
```diff
@@ -61,6 +61,11 @@ public virtual async Task<ManagedIdentityResponse> AuthenticateAsync(
 
             ManagedIdentityRequest request = await CreateRequestAsync(resource).ConfigureAwait(false);
 
+            if (parameters.IsMtlsPopRequested && request?.MtlsCertificate != null)
```
Robbie-Microsoft, line 67: Can you leave a comment here explaining when the `MtlsCertificate` would already be in the request? Would it be there when there was already a mTLS PoP request?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -157,5 +161,13 @@ private static bool ValidateAzureArcEnvironment(string identityEndpoint, string
             logger?.Verbose(() => "[Managed Identity] Azure Arc managed identity is not available.");
             return false;
         }
+
+        internal void SetRuntimeMtlsBindingCertificate(X509Certificate2 cert)
+        {
+            var old = RuntimeMtlsBindingCertificate;
+            RuntimeMtlsBindingCertificate = cert;
+            //dispose prior stored cert on replacement
+            old?.Dispose();
```
Robbie-Microsoft: nit: I would swap these lines

Robbie-Microsoft: How does Dispose() work? Is that part of .NET? Does it communicate to the cert store on the OS?

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
gladjohn: Simplified, you are correct it is not needed

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
gladjohn: Guarded with priorCert != null

src/client/Microsoft.Identity.Client/ManagedIdentity/AbstractManagedIdentity.cs
gladjohn: Added a brief comment in AbstractManagedIdentity.AuthenticateAsync explaining that IMDSv2 mints the binding cert during CSR/attestation and exposes it on the request

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: Added an XML remark clarifying that Dispose() releases the object’s resources and does not remove a cert from the OS store (that’s X509Store.Remove)

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
```diff
@@ -127,7 +127,17 @@ public string Claims
...
+        private IAuthenticationOperation _requestOverrideScheme;
+
+        /// <summary>
+        /// Effective authentication operation (scheme) for this request.
+        /// Defaults to the app's configured operation unless a request-scoped override is applied.
+        /// </summary>
+        public IAuthenticationOperation AuthenticationScheme
+        {
+            get => _requestOverrideScheme ?? _commonParameters.AuthenticationOperation;  // <-- correct fallback
+            internal set => _requestOverrideScheme = value;  // internal set satisfies the “make it settable?” review
```
bgavrilMS: wierd comment

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForManagedIdentityParameters.cs
```diff
@@ -26,6 +27,8 @@ internal class AcquireTokenForManagedIdentityParameters : IAcquireTokenParameter
 
         internal Func<AttestationTokenInput, CancellationToken, Task<AttestationTokenResponse>> AttestationTokenProvider { get; set; }
 
+        internal X509Certificate2 MtlsCertificate { get; set; }
```
bgavrilMS, line 30: Why does this need to be here? Don't we have a caching component for it? This creates a duplicate cache that needs to be maintained.

src/client/Microsoft.Identity.Client/Internal/Requests/AuthenticationRequestParameters.cs
gladjohn: whatever CoPilot says

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -22,6 +23,10 @@ internal class ManagedIdentityClient
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
 
+        // Holds the most recently minted mTLS binding certificate for this application instance.
+        private X509Certificate2 _runtimeMtlsBindingCertificate;
```
bgavrilMS, line 27: This is the second place where you store the cert. There only needs to be 1.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -157,5 +162,24 @@ private static bool ValidateAzureArcEnvironment(string identityEndpoint, string
...
+        /// </remarks>
+        internal void SetRuntimeMtlsBindingCertificate(X509Certificate2 cert)
+        {
+            // Atomically swap the reference and dispose the previous one (if different).
+            var old = System.Threading.Interlocked.Exchange(ref _runtimeMtlsBindingCertificate, cert);
+
+            // If the same instance is passed again, do not dispose it (it is now the current value).
+            if (!object.ReferenceEquals(old, cert))
+            {
+                old?.Dispose();
```
bgavrilMS: Pls don't explicitly call dispose. What if you return the certificate to the caller (via AuthenticationResult) and they use it for calling downstream APIs but you dipose of it explicitly.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -22,6 +23,10 @@ internal class ManagedIdentityClient
         private const string LinuxHimdsFilePath = "/opt/azcmagent/bin/himds";
         internal static ManagedIdentitySource s_sourceName = ManagedIdentitySource.None;
 
+        // Holds the most recently minted mTLS binding certificate for this application instance.
+        private X509Certificate2 _runtimeMtlsBindingCertificate;
+        internal X509Certificate2 RuntimeMtlsBindingCertificate => Volatile.Read(ref _runtimeMtlsBindingCertificate);
```
bgavrilMS, line 28: I understand it's a non-goal to rotate certs etc. But this will return expired certificates.

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForManagedIdentityParameters.cs
gladjohn: Unfortunately, this needs to be there until we add the cache component.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: Unfortunately, this needs to be there until we add the cache component.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: I removed the disposal.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: Yes, it can. The cert rotation logic PR will handle this

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenForManagedIdentityParameters.cs
bgavrilMS: Ok, pls add a //TODO: remove after cache is added (non-blocking comment)

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: @Robbie-Microsoft - pls see https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
neha-bhargava: nit: if the value is null, there is no need to assign a new variable. So the if null should be the first statement.

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
```diff
@@ -209,6 +226,20 @@ await _managedIdentityClient
                 .SendTokenRequestForManagedIdentityAsync(AuthenticationRequestParameters.RequestContext, _managedIdentityParameters, cancellationToken)
                 .ConfigureAwait(false);
 
+            if (AuthenticationRequestParameters.IsMtlsPopRequested && _managedIdentityParameters.MtlsCertificate != null)
```
Robbie-Microsoft, line 228: Why not use a `priorCert` variable here like you did above? Let's keep it consistent. Did you only do it for readability? Alternatively, delete `priorCert` variable above.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: Can you add TODO for deletion?

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
gladjohn: removed it to accommodate to Neha's comment.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3299211

src/client/Microsoft.Identity.Client/Internal/Requests/ManagedIdentityAuthRequest.cs
gladjohn: fixed it

PR #5534, opened 2025-10-15
[MSI v2] MAA token is now cached

gladjohn:
Fixes MAA token caching

- adds in-memory cache 
- 
**Testing**
none

**Performance impact**
better perf as we cache MAA

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/AttestationTokenMemoryCache.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/AttestationTokenMemoryCache.cs
@@ -1,0 +1,270 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/V2/AttestationTokenMemoryCache.cs
```diff
@@ -0,0 +1,270 @@
...
+            {
+                var now = DateTimeOffset.UtcNow;
+
+                // Happy path: valid token in memory
+                if (!string.IsNullOrEmpty(entry.Token) && now + s_expirySkew < entry.ExpiresOnUtc)
+                {
+                    // Past refresh time? Kick a non-blocking background refresh.
+                    if (now >= entry.RefreshOnUtc)
+                    {
+                        KickBackgroundRefresh(entry, input, provider);
```
bgavrilMS, line 69: I thought we discussed to not do any background refresh.

PR #5535, opened 2025-10-16
[POC] Expose secure HTTP client factory and propagate binding certificate in authentication result

cpp11nullptr:


src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -7,18 +7,18 @@
...
     /// <remarks>
     /// Implementations of this interface must be thread-safe.
     /// It is important to reuse HttpClient instances to avoid socket exhaustion.
     /// Do not create a new HttpClient for each call to <see cref="GetHttpClient(X509Certificate2)"/>.
     /// If your application requires Integrated Windows Authentication, set <see cref="HttpClientHandler.UseDefaultCredentials"/> to true.
-    /// This interface is intended for internal use by MSAL only and is designed to support MTLS scenarios.
+    /// This interface is designed to support MTLS scenarios.
     /// </remarks>
-    internal interface IMsalMtlsHttpClientFactory : IMsalHttpClientFactory
+    public interface IMsalMtlsHttpClientFactory : IMsalHttpClientFactory
```
bgavrilMS, line 21: @gladjohn - don't we need this to be made public for MSIv2 as well?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SecureHttpClientFactory.cs
```diff
@@ -12,13 +12,14 @@
...
-    /// A simple implementation of the HttpClient factory that uses a managed HttpClientHandler
+    /// An implementation of the HttpClient factory that uses a managed HttpClientHandler.
+    /// This factory is intended to be used by MTLS scenarios or where server certificate validation is required.
     /// </summary>
     /// <remarks>
     /// .NET should use the IHttpClientFactory, but MSAL cannot take a dependency on it.
     /// .NET should use SocketHandler, but UseDefaultCredentials doesn't work with it 
     /// </remarks>
-    internal class SimpleHttpClientFactory : IMsalMtlsHttpClientFactory, IMsalSFHttpClientFactory
+    public class SecureHttpClientFactory : IMsalMtlsHttpClientFactory, IMsalSFHttpClientFactory
```
bgavrilMS, line 22: Why did you make this public?

cpp11nullptr: Mostly to reuse the logic to create HTTP client with setting up a certificate by Identity.Web but it can be also beneficial for users of MSAL.NET who wants either secure HTTP client factory (with mTLS or server certificate validation).

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
gladjohn: No, we don't need to expose this for MSI v2. We will continue to use the existing factory `IMsalHttpClientFactory` to make calls into IMDS endpoints, this will ensure we use custom factory. But for the mTLS calls to ESTS, we will use the new `IMsalMtlsHttpClientFactory`.  MSAL will be responsible for calls to ESTS.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -7,18 +7,18 @@
 namespace Microsoft.Identity.Client
 {
     /// <summary>
-    /// Internal factory responsible for creating HttpClient instances configured for mutual TLS (MTLS).
-    /// This factory is specifically intended for use within the MSAL library for secure communication with Azure AD using MTLS.
+    /// A factory responsible for creating HttpClient instances configured for mutual TLS (MTLS).
```
MZOLN, line 10: Isn't it an interface not an actual factory?

bgavrilMS:
I believe this is not needed so closing. Please reopen if I am mistaken @cpp11nullptr

Issue #5361, opened 2025-06-24
[Feature Request] Audit all /token parameters


bgavrilMS:
### MSAL client type

Confidential, Managed identity

### Problem statement

Extensiblity APIs can lead to bad token caching, where ppl don't realize that tokens aren't actually cached by all params. We need to limit this by enhancing caching, by making extensiblity APIs not for production (they should be already):

- [ ] make OnBeforeTokenRequest experimental
- [x] if WithExtraQueryParameters, cache tokens by those params
- [x] introduce WithBodyParamters, cache tokens by those params (maybe have an allow-list of params that do not impact caching)
- [ ] audit ID.Web's and other SDKs usage of OnBeforeTokenRequest and how caching is impacted. 
- [ ] See `IAuthenticationOperation GetTokenParameters` - either deprecate it or make it use `WithBodyParamters`

bgavrilMS:
Ideas: 

- make OnBeforeTokenRequest experimental
- make WithExtraQueryParameters experimental
- audit ID.Web's and other SDKs usage of `OnBeforeTokenRequest` and how caching is impacted.
- discuss with eSTS to find out what extra qp apps are using in the wild. Get a list of params and their meaning. Use that to decide if we are to partition the cache or not. Important to differentiate between S2S and C2S.

PR #5536, opened 2025-10-18
Adjust WithExtraQueryParameters APIs and cache key behavior

Avery-Dunn:
This PR is for KR [3310905](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3310905) and issue https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5361.

In short, there is a problem in our existing `WithExtraQueryParameters` behavior: the extra query parameters could affect tokens but were not used as part of the cache keys, so if the `WithExtraQueryParameters` values changed between requests we could return cached tokens that were not valid for the new request. However, simply adding the extra query parameters to the cache key would not work: it would break existing cache lookup behavior, and not all parameters should be cached.

This PR tries to solve that problem with the following changes:
 - Add a new `WithExtraQueryParameters` that takes in a `IDictionary<string, (string value, bool includeInCacheKey)>`
   - Essentially, instead of a key-value pair of strings the value is a tuple, allowing the caller to control which parameters are used in the cache key
   - This new API adds the extra query parameters to the existing `CacheKeyComponents` field, which already does the task of adding extra parameters to the cache key
   - Affects: `BaseAbstractAcquireTokenParameterBuilder`, `AbstractApplicationBuilder`
 - Deprecate the existing `WithExtraQueryParameters` APIs, and have them call the new API to set the `ExtraQueryParameters` fields
   - Affects: `AbstractAcquireTokenParameterBuilder`, `BaseAbstractAcquireTokenParameterBuilder`, `AbstractApplicationBuilder`
  - New internal helper method in `CoreHelpers` to covert the old Dictionary style to the new one
 
 This PR also adds some new tests to cover the new behavior, and makes some small changes to existing tests:
 - `ExtraQueryParametersTests`: A new test class with tests covering the new cache key behavior, as well as demonstrating the behavior of the deprecated `WithExtraQueryParameters` APIs
 - `CacheKeyExtensionTests`: A new test to show the new `WithExtraQueryParameters` API does not conflict with the existing `WithAdditionalCacheKeyComponents` API from the extensibility package
 - The rest of the changes to the test files were just to follow the new style and avoid deprecation warnings, no test behavior was actually changed

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/PublicApiTests/ExtraQueryParametersTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/PublicApiTests/ExtraQueryParametersTests.cs
@@ -291,8 +291,6 @@ namespace Microsoft.Identity.Test.Unit.PublicApiTests
         {
             using (var httpManager = new MockHttpManager())
             {
-                httpManager.AddInstanceDiscoveryMockHandler();
-
                 // Setup app without extra query parameters
                 var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
                                                         .WithAuthority(new Uri(ClientApplicationBase.DefaultAuthority), true)
@@ -352,8 +350,6 @@ namespace Microsoft.Identity.Test.Unit.PublicApiTests
         {
             using (var httpManager = new MockHttpManager())
             {
-                httpManager.AddInstanceDiscoveryMockHandler();
-
                 // Step 1: Create app without extra query parameters
                 var app = ConfidentialClientApplicationBuilder.Create(TestConstants.ClientId)
                                                         .WithAuthority(new Uri(ClientApplicationBase.DefaultAuthority), true)


src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,46 @@ public T WithCorrelationId(Guid correlationId)
...
+        }
+
+        /// <summary>
+        /// Sets Extra Query Parameters for the query string in the HTTP authentication request with control over which parameters are included in the cache key
+        /// </summary>
+        /// <param name="extraQueryParameters">This parameter will be appended as is to the query string in the HTTP authentication request to the authority.
+        /// For each parameter, you can specify whether it should be included in the cache key.
+        /// The parameter can be null.</param>
+        /// <returns>The builder to chain .With methods.</returns>
+        public T WithExtraQueryParameters(IDictionary<string, (string value, bool includeInCacheKey)> extraQueryParameters)
```
bgavrilMS: Having read a bit about tuples, I think these are not good in a public API. You cannot add a member to a tuple. Can we use an object, maybe a record https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,46 @@ public T WithCorrelationId(Guid correlationId)
         /// as a string of segments of the form <c>key=value</c> separated by an ampersand character.
         /// The parameter can be null.</param>
         /// <returns>The builder to chain the .With methods.</returns>
+        [Obsolete("This method is deprecated. Please use the WithExtraQueryParameters(IDictionary<string, (string value, bool includeInCacheKey)>) method instead, which provides control over which parameters are included in the cache key.", false)]
```
bgavrilMS:
```suggestion
        [Obsolete("This method is deprecated. Use the WithExtraQueryParameters(IDictionary<string, (string value, bool includeInCacheKey)>) method instead, which provides control over which parameters are included in the cache key.", false)]
```

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: Yeah, after we talked about using tuples for this I searched through the codebase and didn't see this style used anywhere, so I wasn't sure if it was a common thing in .NET or not. We do use this Tuple class that has three items a few times, and there are versions of the class that take more: https://learn.microsoft.com/en-us/dotnet/api/system.tuple-3?view=net-9.0

I'll look into records, the internal behavior changes are pretty simple so I should be able to try out different data types without much issue.

But is there a reason we'd want to add a new member? What would we need besides the value and whether or not it should be used in the cache key?

bgavrilMS: Ack

bgavrilMS: Naming: (string ParameterValue, bool IncludeInCacheKey)

bgavrilMS (approved):
LGTM, minor option

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
```diff
@@ -140,6 +140,24 @@ public static Dictionary<string, string> ParseKeyValueList(string input, char de
...
+        {
+            if (parameters == null)
+            {
+                return null;
+            }
+
+            var result = new Dictionary<string, (string value, bool includeInCacheKey)>(StringComparer.OrdinalIgnoreCase);
+            foreach (var kvp in parameters)
+            {
+                result[kvp.Key] = (kvp.Value, false); // Include all parameters in cache key by default
```
neha-bhargava: As per the comment, should this not be true?

tests/devapps/DesktopTestApp/PublicClientHandler.cs
```diff
@@ -44,6 +45,7 @@ public async Task<AuthenticationResult> AcquireTokenInteractiveAsync(
             CreateOrUpdatePublicClientApp(InteractiveAuthority, ApplicationId);
 
             AuthenticationResult result;
+#pragma warning disable CS0618 // Type or member is obsolete
```
neha-bhargava, line 48: Maybe replace these with the new method and remove this line everywhere?

Avery-Dunn: I think this is the only file where I suppressed the warnings instead of making the actual change. Unlike all the other places, using the new API here would've meant multiple method signature changes, then changes to everything that referenced those methods, then changes to fields in different classes...

It was already started to be a big PR, and those suppressions seemed to be used in a lot of tests, so I figured I'd take a shortcut on this one test app.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,46 @@ public T WithCorrelationId(Guid correlationId)
...
-            CommonParameters.ExtraQueryParameters = extraQueryParameters ??
-                new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            return WithExtraQueryParameters(CoreHelpers.ConvertToTupleParameters(extraQueryParameters));
+        }
+
+        /// <summary>
+        /// Sets Extra Query Parameters for the query string in the HTTP authentication request with control over which parameters are included in the cache key
+        /// </summary>
+        /// <param name="extraQueryParameters">This parameter will be appended as is to the query string in the HTTP authentication request to the authority.
+        /// For each parameter, you can specify whether it should be included in the cache key.
```
neha-bhargava: I would recommend adding some suggestions or examples when it will make sense to add the param to cache key and when to exclude

Avery-Dunn: In the latest commit I adjusted the comments here and at the app-level API: explained the different parameters in more detail, and gave suggestions on when to set the cache key value to true.

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
Avery-Dunn: Good catch, fix in the latest commit.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: In the latest commit the names start with an uppercase letter, and I improved the comments explaining each parameter.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,49 @@ public T WithCorrelationId(Guid correlationId)
...
+            if (extraQueryParameters == null)
+            {
+                CommonParameters.ExtraQueryParameters = null;
+                return this as T;
+            }
+
+            // Add each parameter to ExtraQueryParameters and, if requested, to CacheKeyComponents
+            foreach (var kvp in extraQueryParameters)
+            {
+                CommonParameters.ExtraQueryParameters = CommonParameters.ExtraQueryParameters ?? new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
```
trwalke: hmm, setting a value to itself seems like an odd way to check if something is null and then initializing it. It can be misleading potentially. I would simply add a normal if check. 
```
if (CommonParameters.ExtraQueryParameters is null)
{...}
```

Avery-Dunn: One the one hand it does seem weird, but on the other hand I couldn't find any guidance saying not to do it like that.

However, after trying out the if statement option I don't think it made it more readable: since CommonParameters.CacheKeyComponents does the same thing a few lines down it'd be 3 if statements in one small method, some of which were wrapped up in a loop.

But I did move the null check/initialization outside of the for loop in the latest commit, since we don't need to do it multiple times.

src/client/Microsoft.Identity.Client/Utils/CoreHelpers.cs
```diff
@@ -140,6 +140,24 @@ public static Dictionary<string, string> ParseKeyValueList(string input, char de
             return ParseKeyValueList(input, delimiter, urlDecode, true, requestContext);
         }
 
+        // Helper method intended to help deprecate some WithExtraQueryParameters APIs.
+        // Convert from Dictionary<string, string> to Dictionary<string, (string value, bool includeInCacheKey)>,
+        // with all includeInCacheKey set to false by default to maintain existing behavior of those older APIs.
+        internal static IDictionary<string, (string value, bool includeInCacheKey)> ConvertToTupleParameters(IDictionary<string, string> parameters)
```
bgavrilMS, line 146: nit: It's not really a "core helper" and could be included in ApiConfig or AppConfig

btw, you can rewrite this in 1-liner

```csharp
 return (parameters ?? Enumerable.Empty<KeyValuePair<string, string>>())
        .ToDictionary(
            kvp => kvp.Key,
            kvp => (kvp.Value, false),
            StringComparer.OrdinalIgnoreCase);
```

src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt
```diff
@@ -7,4 +9,4 @@ Microsoft.Identity.Client.IMsalMtlsHttpClientFactory.GetHttpClient(System.Securi
 Microsoft.Identity.Client.ManagedIdentityApplication.GetManagedIdentitySourceAsync() -> System.Threading.Tasks.Task<Microsoft.Identity.Client.ManagedIdentity.ManagedIdentitySource>
 Microsoft.Identity.Client.ManagedIdentity.ManagedIdentitySource.ImdsV2 = 8 -> Microsoft.Identity.Client.ManagedIdentity.ManagedIdentitySource
 Microsoft.Identity.Client.ManagedIdentityApplicationBuilder.WithExtraQueryParameters(System.Collections.Generic.IDictionary<string, string> extraQueryParameters) -> Microsoft.Identity.Client.ManagedIdentityApplicationBuilder
-static Microsoft.Identity.Client.ApplicationBase.ResetStateForTest() -> void
+static Microsoft.Identity.Client.ApplicationBase.ResetStateForTest() -> void
```
gladjohn, line 12: can this be undone?

src/client/Microsoft.Identity.Client/PublicApi/net8.0-android/PublicAPI.Unshipped.txt
```diff
@@ -7,4 +9,4 @@ Microsoft.Identity.Client.IMsalMtlsHttpClientFactory.GetHttpClient(System.Securi
 Microsoft.Identity.Client.ManagedIdentityApplication.GetManagedIdentitySourceAsync() -> System.Threading.Tasks.Task<Microsoft.Identity.Client.ManagedIdentity.ManagedIdentitySource>
 Microsoft.Identity.Client.ManagedIdentity.ManagedIdentitySource.ImdsV2 = 8 -> Microsoft.Identity.Client.ManagedIdentity.ManagedIdentitySource
 Microsoft.Identity.Client.ManagedIdentityApplicationBuilder.WithExtraQueryParameters(System.Collections.Generic.IDictionary<string, string> extraQueryParameters) -> Microsoft.Identity.Client.ManagedIdentityApplicationBuilder
-static Microsoft.Identity.Client.ApplicationBase.ResetStateForTest() -> void
+static Microsoft.Identity.Client.ApplicationBase.ResetStateForTest() -> void
```
gladjohn, line 12: same comment like before

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,49 @@ public T WithCorrelationId(Guid correlationId)
...
+            CommonParameters.ExtraQueryParameters = CommonParameters.ExtraQueryParameters ?? new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            // Add each parameter to ExtraQueryParameters and, if requested, to CacheKeyComponents
+            foreach (var kvp in extraQueryParameters)
+            {
+                CommonParameters.ExtraQueryParameters[kvp.Key] = kvp.Value.Value;
+
+                if (kvp.Value.IncludeInCacheKey)
+                {
+                    CommonParameters.CacheKeyComponents = CommonParameters.CacheKeyComponents ?? new SortedList<string, Func<CancellationToken, Task<string>>>();
```
gladjohn, line 125: create the SortedList with a stable, case‑insensitive, ordinal comparer

gladjohn: please add a case sensitive test to validate this.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,49 @@ public T WithCorrelationId(Guid correlationId)
...
+
+                    // Capture the value in a local to avoid closure issues
+                    string valueToCache = kvp.Value.Value;
+
+                    // Add to cache key components - uses a func that returns the value as a task
+                    CommonParameters.CacheKeyComponents[kvp.Key] = (CancellationToken _) => Task.FromResult(valueToCache);
+                }
+            }
```
gladjohn, line 133: can someone remove a cache key?

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,49 @@ public T WithCorrelationId(Guid correlationId)
...
+        /// - Value: The parameter value that will be appended to the query string
+        /// - IncludeInCacheKey: Whether this parameter should be included when computing the token's cache key.
+        /// To help ensure the correct token is returned from the cache, IncludeInCacheKey should be true if the parameter affects token content or validity (e.g., resource-specific claims or parameters).
+        /// The parameter can be null.</param>
+        /// <returns>The builder to chain .With methods.</returns>
+        public T WithExtraQueryParameters(IDictionary<string, (string Value, bool IncludeInCacheKey)> extraQueryParameters)
+        {
+            if (extraQueryParameters == null)
+            {
+                CommonParameters.ExtraQueryParameters = null;
```
gladjohn, line 112: should you also remove cache keys that were previously set?

src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt
Avery-Dunn: I'm not sure why Github is marking that as a change, that line's the same as it is in the main branch: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/PublicApi/net462/PublicAPI.Unshipped.txt

Maybe it's just because my lines went to the top instead of the bottom of the file? I just let it merge automatically.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: That is how the CacheKeyComponents field is set in other parts of the library: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7bf0a28d2df77accb50769fa5c9550e9472bd356/src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs#L159

Is there a unique issue here, or should it also be fixed elsewhere? Making it more stable here would mean we each change the existing behavior elsewhere too, or make this version inconsistent.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: This is the same as the current behavior: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7bf0a28d2df77accb50769fa5c9550e9472bd356/src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs#L92

If a null `extraQueryParameters` is passed in then `CommonParameters.ExtraQueryParameters` gets cleared. I just assumed it was a way of overriding the app-level extra query parameters or something like that, but are you saying that behavior might be a bug?

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: What do you mean? Like remove a single, specific entry in the token cache?

src/client/Microsoft.Identity.Client/PublicApi/net8.0-android/PublicAPI.Unshipped.txt
Avery-Dunn: Same as your other comment, that line here is exactly the same as in the main branch so I'm not sure why Github is saying it changed.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn, line 125: is there a test that does something like this, in this specific order? 
```
extra param 1 - include in cache = yes 
extra param 2 - include in cache = no 
extra param 3 - include in cache = yes
```
and asserts?

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn: I would clarify the behavior of this new API. Because this one adds the cache key, the previous one didn't.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn: something like, 

Request 1 - 
extra param 1 - include in cache = yes 
extra param 2 - include in cache = yes

Request 2 - 
extra param 1 - include in cache = yes 
extra param 2 - include in cache = no

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn: adding a case sensitive test should validate if there are existing bugs in the code

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: What sort of clarification do you think it needs? The comments above it explain how to get the query parameters added to the cache key (set that boolean to true).

So if a customer passes in some parameters, we will add those parameters to the cache key. And if a customer passes in nothing, then we will add nothing to the cache key.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn, line 125:
```suggestion
                    CommonParameters.CacheKeyComponents ??=
    new SortedList<string, Func<CancellationToken, Task<string>>>(StringComparer.OrdinalIgnoreCase);
```

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn: so, to remove don't you need to add 

```
else
{
    CommonParameters.CacheKeyComponents?.Remove(kvp.Key);
}
```

again, as I said -  I am not looking into the requirements doc, I am just asking if this is a possibility. If you think this is not a valid case, ignore this comment

neha-bhargava: This dictionary builds per request. I don't think we need remove.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
neha-bhargava: the SortedList will use its default comparer if one is not specified. By default, this is Comparer<TKey>.Default, which for string is StringComparer.Ordinal. The behavior will be same whether or not you add ordinal comparer.

neha-bhargava: You can validate this with test cases.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
neha-bhargava: If someone selects to not add the extra query param to the cache I think the cache key will be different for that request and the request will go to IDP

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
```diff
@@ -89,10 +89,49 @@ public T WithCorrelationId(Guid correlationId)
...
+        /// Sets Extra Query Parameters for the query string in the HTTP authentication request with control over which parameters are included in the cache key
+        /// </summary>
+        /// <param name="extraQueryParameters">This parameter will be appended as is to the query string in the HTTP authentication request to the authority, and merged with those added to the application-level WithExtraQueryParameters API.
+        /// Each dictionary entry maps a parameter name to a tuple containing:
+        /// - Value: The parameter value that will be appended to the query string
+        /// - IncludeInCacheKey: Whether this parameter should be included when computing the token's cache key.
+        /// To help ensure the correct token is returned from the cache, IncludeInCacheKey should be true if the parameter affects token content or validity (e.g., resource-specific claims or parameters).
+        /// The parameter can be null.</param>
+        /// <returns>The builder to chain .With methods.</returns>
+        public T WithExtraQueryParameters(IDictionary<string, (string Value, bool IncludeInCacheKey)> extraQueryParameters)
```
gladjohn, line 108: should this be an experimental API?

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: Just added `WithExtraQueryParameters_ParameterOrderDoesNotAffectCacheKey_TestAsync` to `ExtraQueryParametersTests`.

It covers a variety of different parameter orderings, some that aren't added to cache keys and some are, etc.

Avery-Dunn: I thought about this when I was first writing it, but as best as I can tell (through Copilot) the "??=" was only introduced in newer frameworks than we support, right?

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: Yeah, this is per-request behavior, so if a customer changes what parameters should be used in the cache key then the new request will have a new cache entry, and the cache entry for the other request will be unaffected.

I just added `WithExtraQueryParameters_VaryingCacheKeyInclusion_TestAsync` to `ExtraQueryParametersTests` to show this behavior:
1) Token request with param1 and param2, both included in the cache key
2) Token request with param1 and param2 again, but only param1 in the cache key
3) Token request with just param1, included in the cache key

Requests 1 and 2 result in different tokens even though they would have the same contents, but request 3 finds the token from request 2 even though they have different contents.

In short, the new API allows a caller to recreate the exact behavior that was a problem in the old API, where we aren't uniquely identifying a token based on the query parameters originally used to retrieve it.

However, I don't think this is something we need to fix: the API is meant to give the caller control over how tokens are identified in the cache, and if they want to keep turning the option on and off they then presumably they some reason for it.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5536#discussion_r2500349434 for more context, but I I just added `WithExtraQueryParameters_VaryingCacheKeyInclusion_TestAsync` to `ExtraQueryParametersTests` to cover this behavior:
1) Token request with param1 and param2, both included in the cache key
2) Token request with param1 and param2 again, but only param1 in the cache key
2) Token request with just param1, included in the cache key

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
Avery-Dunn: @bgavrilMS can weigh in on this because I'm not sure how y'all decide what to put behind that experimental flag.

In our initial discussion we also planned on putting certain extensibility APIs behind the experimental flag, but it turned out the relevant API already was using it: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7bf0a28d2df77accb50769fa5c9550e9472bd356/src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs#L52

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn: AFAIK, this should work in all current TFMs we have

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
gladjohn: > and the cache entry for the other request will be unaffected

how do we communicate this to the caller?

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
bgavrilMS: No, I think this is fine. It's for forcing a test slice etc.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
bgavrilMS: These APIs should be additive. So I should be able to call WithExtraParams(p1, p2) and WithExtraParams(p3) and to expect p1, p2, p3. There are no removal semantics.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
bgavrilMS: If it builds, it works :). 

But to your point @Avery-Dunn - the null coalescing operator was introduced in a specific version of C# (the language). The language version and the target framework are different. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning and https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

Today, MSAL uses 

Image which is actually discouraged because it makes builds unreliable. But this has not been our experience :). We can change it if builds break. The actual version is dependent on the SDK.

bgavrilMS: I think @gladjohn is saying to use `new SortedList<string, Func<CancellationToken, Task<string>>>(StringComparer.OrdinalIgnoreCase);`

This is fine, being explicit about string comparison rules is preferred. Ultimately, these params go on the wire and URIs are case insensitive. But it would lead to smth like `login.microsoft.com?fmi=a&fmi=b` which creates confusion.

src/client/Microsoft.Identity.Client/ApiConfig/BaseAbstractAcquireTokenParameterBuilder.cs
bgavrilMS: You cannot remove a cache key.

src/client/Microsoft.Identity.Client/PublicApi/net8.0-android/PublicAPI.Unshipped.txt
bgavrilMS: As long as it doesn't create warnings, let's just use it.

PR #5541, opened 2025-10-23
Updated DesktopOsHelper.IsMac to work properly on net10 + macOS 26

reynoldskr:
Before .NET 10, `RuntimeInformation.OSDescription` [called `uname`](https://github.com/dotnet/runtime/blob/v10.0.0-rc.2.25502.107/src/native/libs/System.Native/pal_runtimeinformation.c#L44) on *nix systems to populate itself. Now it uses a [nicer way](https://github.com/dotnet/runtime/blob/cc5f4c0ed72aad319bf032d03f9a99213845aeaf/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/RuntimeInformation.Unix.cs#L12) using the .NET runtime target information, that means that field no longer contains `Darwin`, so `IsMac` always returned false. This simply lets that method check for `macOs` as well.

**Testing**
I couldn't find any unit tests for this class.

**Performance impact**
None

### DIFF
--- a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
+++ b/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
@@ -52,11 +52,7 @@ namespace Microsoft.Identity.Client.PlatformsCommon.Shared
 #elif NETFRAMEWORK
             return Environment.OSVersion.Platform == PlatformID.MacOSX;
 #elif NET8_0_OR_GREATER
-            string OSDescription = RuntimeInformation.OSDescription;
-
-            // .NET 10 and later changes the OS string in `OSDescription`
-            return OSDescription.Contains("Darwin", StringComparison.OrdinalIgnoreCase) ||
-                   OSDescription.Contains("macOS", StringComparison.OrdinalIgnoreCase);
+            return OperatingSystem.IsMacOS();
 #else
             return RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
 #endif


bgavrilMS:
@fengga - smth for you to review

reynoldskr:
Hey @bgavrilMS / @fengga , any chance of a review here 🙂 ?

reynoldskr:
This is still broken BTW, and net10 is GA now

jrr:
This is a real thing blocking me from updating a project to .NET 10.

For what it's worth, here are some values from my machine:

```csharp
Log.Debug("Platform: {OsVersionPlatform}", Environment.OSVersion.Platform);
Log.Debug("OSDescription: {OsDescription}", RuntimeInformation.OSDescription);
Log.Debug("Is OSX: {IsOsPlatform}", RuntimeInformation.IsOSPlatform(OSPlatform.OSX));
```

```
[22:28:40 DBG] Platform: Unix
[22:28:40 DBG] OSDescription: macOS 26.1.0
[22:28:40 DBG] Is OSX: True
```

(so I think you could either look for the string _"macOS"_ (as in this PR) or call `IsOSPlatform`)

For comparison, here are the values from .NET 9:
```
[23:26:02 DBG] Platform: Unix
[23:26:02 DBG] OSDescription: Darwin 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:41 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6000
[23:26:02 DBG] Is OSX: True
```

_Update:_ the problem isn't specific to MacOS 26! It also happens with .NET 10 on earlier OS versions:
```
Microsoft.Data.SqlClient.SqlException (0x80131904): macOS 15.6.1
 ---> System.PlatformNotSupportedException: macOS 15.6.1
```

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs
```diff
@@ -53,7 +53,10 @@ public static bool IsMac()
             return Environment.OSVersion.Platform == PlatformID.MacOSX;
 #elif NET8_0_OR_GREATER
             string OSDescription = RuntimeInformation.OSDescription;
-            return OSDescription.Contains("Darwin", StringComparison.OrdinalIgnoreCase);
+
+            // .NET 10 and later changes the OS string in `OSDescription`
+            return OSDescription.Contains("Darwin", StringComparison.OrdinalIgnoreCase) ||
```
cheenamalhotra: View this comment as well: https://github.com/dotnet/runtime/pull/113041/files#r1985415935

You should be using `RuntimeInformation.IsOSPlatform(OSPlatform.OSX) or OperatingSystem.IsMacOS()` for .NET 10 onwards - maybe easier to handle it with if-def conditionally.

reynoldskr: I switched to the `IsMacOS` function

cricketthomas:
Would really love for this to get merged and released! It would solve so many issues! Thanks

PR #5549, opened 2025-10-24
Fix instance discovery bug 5546

### LINKED ISSUE #5546: [Bug] MSAL should not crash if instance discovery results in HTTP errors
### Library version used

4.77.1

### .NET version

1. Use an airgapped authority and regional
2. Make instance discoevry call to global endpopint fail with SSL error

Actual: MSAL crashes with the SSL exception
Expected: MSAL shopuld log the error and continue without aliases

### Scenario

None

see above

### Relevant code snippets

```csharp

```

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

bgavrilMS:
Fixes #5546 

Authority validation should only fail on a specific error. Do not fail on Http errors.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
+++ b/src/client/Microsoft.Identity.Client/Instance/Discovery/InstanceDiscoveryManager.cs
@@ -208,7 +208,7 @@ namespace Microsoft.Identity.Client.Instance.Discovery
                 requestContext.Logger.Error($"[Instance Discovery] Instance discovery failed - invalid instance! ");
                 throw;
             }
-            catch (HttpRequestException e) 
+            catch (Exception e) 
             { 
                 requestContext.Logger.Warning(
                     $"[Instance Discovery] Instance Discovery failed. MSAL will continue without network instance metadata. \n\r" +


tests/Microsoft.Identity.Test.Unit/PublicApiTests/ConfidentialClientApplicationTests.cs
```diff
@@ -151,39 +151,7 @@ public async Task ConfidentialClientUsingSecretNoCacheProvidedTestAsync()
                 appCacheAccess.AssertAccessCounts(1, 1);
                 userCacheAccess.AssertAccessCounts(0, 0);
             }
-        }
-
-        [TestMethod]
-        public async Task ConfidentialClientUsingSecretNoInstanceDiscoveryTestAsync()
```
bgavrilMS, line 157: I just moved this

tests/Microsoft.Identity.Test.Unit/PublicApiTests/InstanceDiscoveryTests.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Microsoft.Identity.Test.Unit.PublicApiTests
+{
+    [TestClass]
+    public class InstanceDiscoveryTests : TestBase
+    {
+        [TestMethod]
+        public async Task ConfidentialClientUsingSecretNoInstanceDiscoveryTestAsync()
```
bgavrilMS, line 37: Do not review this one, it was just moved.

Avery-Dunn (approved):

PR #5557, opened 2025-10-28
[MSI v2] Cache MSI v2 cert 

gladjohn:
commit 1 - caches MSI v2 in-memory + tests (ready)
commit 2 - stores cert to store and if there is a cache miss, checks store (in-progress)
commit 3 - adds mutex (in-progress)

## This PR adds a minimal, internal in‑memory certificate cache for IMDS v2 mTLS PoP:

- One entry per cache key (the key is exactly what the caller passes: client_id, object_id, or resource_id), by design.
- Entries are returned only if valid ≥ 24 hours, otherwise they are ignored and evicted (mTLS tokens are valid up to 24h and must be bound to a cert valid for the full token lifetime).
- The cache returns a clone of the X509Certificate2 so callers can dispose independently. The cache owns its copy and disposes on eviction/clear.
- Normalized to UTC for time comparisons.
- Thread‑safe: ConcurrentDictionary + an external per‑key SemaphoreSlim (unchanged) prevents double‑mints.

### Why 24 hours?

mTLS PoP tokens can be valid up to 24h; reusing a cert that expires sooner would create a token that cannot be used by the resource throughout its validity. The cache therefore refuses to store certs with < 24h remaining and evicts entries that have hit their NotAfter.

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
         /// </summary>
         /// <param name="nowUtc"></param>
         /// <returns></returns>
-        public bool IsExpiredUtc(DateTimeOffset nowUtc) => nowUtc >= NotAfterUtc;
+        public bool IsExpiredUtc(DateTimeOffset nowUtc) => nowUtc >= (NotAfterUtc - MinRemainingLifetime);
 
         /// <summary>
         /// dispose the entry and its certificate.
@@ -72,14 +72,8 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
                 return; // already disposed
             }
 
-            try
-            {
-                Certificate.Dispose();
-            }
-            catch
-            {
-                // defensive: some platforms throw on double/invalid dispose; we swallow to keep eviction robust
-            }
+            // Idempotent due to the atomic guard
+            Certificate.Dispose();
         }
     }
 }
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ICertificateCache.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ICertificateCache.cs
@@ -17,16 +17,17 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
         /// Try to get a cached certificate+endpoint+clientId for the specified cacheKey.
         /// Returns true and non-null outputs if found and not expired.
         /// </summary>
-        bool TryGet(string cacheKey, out CertificateCacheValue value, ILoggerAdapter logger = null);
+        bool TryGet(
+            string cacheKey, 
+            out CertificateCacheValue value, 
+            ILoggerAdapter logger = null);
 
         /// <summary>
         /// Insert or replace the cached certificate+endpoint+clientId for cacheKey.
         /// </summary>
         void Set(
             string cacheKey,
-            X509Certificate2 certificate,
-            string endpoint,
-            string clientId,
+            in CertificateCacheValue value,
             ILoggerAdapter logger = null);
 
         /// <summary>Remove an entry if present.</summary>
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
@@ -52,22 +52,20 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
         /// <inheritdoc />
         public void Set(
             string cacheKey,
-            X509Certificate2 certificate,
-            string endpoint,
-            string clientId,
+            in CertificateCacheValue value,
             ILoggerAdapter logger = null)
         {
             ThrowIfDisposed();
             ValidateCacheKey(cacheKey);
 
-            if (certificate is null)
-                throw new ArgumentNullException(nameof(certificate));
-            if (string.IsNullOrWhiteSpace(endpoint))
-                throw new ArgumentException("Endpoint must be non-empty.", nameof(endpoint));
-            if (string.IsNullOrWhiteSpace(clientId))
-                throw new ArgumentException("ClientId must be non-empty.", nameof(clientId));
+            if (value.Certificate is null)
+                throw new ArgumentNullException(nameof(value.Certificate));
+            if (string.IsNullOrWhiteSpace(value.Endpoint))
+                throw new ArgumentException("Endpoint must be non-empty.", nameof(value.Endpoint));
+            if (string.IsNullOrWhiteSpace(value.ClientId))
+                throw new ArgumentException("ClientId must be non-empty.", nameof(value.ClientId));
 
-            var notAfterUtc = ToNotAfterUtc(certificate);
+            var notAfterUtc = ToNotAfterUtc(value.Certificate);
             var nowUtc = DateTimeOffset.UtcNow;
 
             // Enforce minimum remaining lifetime (e.g., 24h).
@@ -75,14 +73,14 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
             {
                 var remaining = notAfterUtc - nowUtc;
                 logger?.Verbose(() =>
-                    "[CertCache] Skipping certificate with insufficient remaining lifetime (" +
-                    $"{remaining.TotalHours:F2}h) (key='{Mask(cacheKey)}').");
+                    "[CertCache] Skipping certificate with insufficient remaining lifetime " +
+                    $"({remaining.TotalHours:F2}h) (key='{Mask(cacheKey)}').");
                 return;
             }
 
             // Cache owns its copy; it will dispose upon eviction.
-            var cachedCopy = new X509Certificate2(certificate);
-            var newEntry = new CertificateCacheEntry(cachedCopy, notAfterUtc, endpoint, clientId);
+            var cachedCopy = new X509Certificate2(value.Certificate);
+            var newEntry = new CertificateCacheEntry(cachedCopy, notAfterUtc, value.Endpoint, value.ClientId);
 
             _entriesByCacheKey.AddOrUpdate(
                 cacheKey,
@@ -93,7 +91,6 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
                 },
                 (_, old) =>
                 {
-                    // Defensive: don't double-dispose in weird interleavings
                     if (!old.IsDisposed)
                     {
                         old.Dispose();
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
@@ -489,8 +489,9 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
                 // 3) Mint + cache under the provided cacheKey
                 var created = await factory().ConfigureAwait(false);
 
-                s_mtlsCertificateCache.Set(
-                    cacheKey, created.Item1, created.Item2, created.Item3, logger);
+                s_mtlsCertificateCache.Set(cacheKey,
+                    new CertificateCacheValue(created.Item1, created.Item2, created.Item3),
+                    logger);
 
                 return created;
             }
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/InMemoryCertificateCacheTests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/InMemoryCertificateCacheTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Concurrent;
 using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 using Microsoft.Identity.Client.ManagedIdentity.V2;
@@ -44,7 +45,7 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
             const string ep = "https://mtls.endpoint";
             const string cid = "11111111-1111-1111-1111-111111111111";
 
-            cache.Set(key, cert, ep, cid);
+            cache.Set(key, new CertificateCacheValue(cert, ep, cid));
 
             var ok = cache.TryGet(key, out var value);
             Assert.IsTrue(ok);
@@ -72,7 +73,7 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
 
             // Certificate lifetime shorter than product threshold (24h)
             using var shortCert = CreateSelfSignedCert(TimeSpan.FromHours(1));
-            cache.Set("short-key", shortCert, "https://mtls", "client-guid");
+            cache.Set("short-key", new CertificateCacheValue(shortCert, "https://mtls", "client-guid"));
 
             var ok = cache.TryGet("short-key", out _);
             Assert.IsFalse(ok, "Cache should skip certs with remaining lifetime < 24h.");
@@ -86,8 +87,8 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
             using var certB = CreateSelfSignedCert(TimeSpan.FromDays(3), "CN=B");
 
             const string key = "replace-key";
-            cache.Set(key, certA, "https://ep", "cid");
-            cache.Set(key, certB, "https://ep", "cid");
+            cache.Set(key, new CertificateCacheValue(certA, "https://ep", "cid"));
+            cache.Set(key, new CertificateCacheValue(certB, "https://ep", "cid"));
 
             var ok = cache.TryGet(key, out var v);
             Assert.IsTrue(ok);
@@ -108,7 +109,7 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
             var cache = new InMemoryCertificateCache();
             using var cert = CreateSelfSignedCert(TimeSpan.FromDays(2));
 
-            cache.Set("k1", cert, "https://ep", "cid");
+            cache.Set("k1", new CertificateCacheValue(cert, "https://ep", "cid"));
             var removed = cache.Remove("k1");
             Assert.IsTrue(removed);
 
@@ -123,8 +124,8 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
             using var c1 = CreateSelfSignedCert(TimeSpan.FromDays(2));
             using var c2 = CreateSelfSignedCert(TimeSpan.FromDays(2));
 
-            cache.Set("k1", c1, "https://ep1", "cid1");
-            cache.Set("k2", c2, "https://ep2", "cid2");
+            cache.Set("k1", new CertificateCacheValue(c1, "https://ep1", "cid1"));
+            cache.Set("k2", new CertificateCacheValue(c2, "https://ep2", "cid2"));
 
             cache.Clear();
 
@@ -143,10 +144,10 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
             Assert.ThrowsException<ArgumentException>(() => cache.TryGet(null, out _));
 
             // Set
-            Assert.ThrowsException<ArgumentException>(() => cache.Set(" ", cert, "ep", "cid"));
-            Assert.ThrowsException<ArgumentNullException>(() => cache.Set("k", null, "ep", "cid"));
-            Assert.ThrowsException<ArgumentException>(() => cache.Set("k", cert, " ", "cid"));
-            Assert.ThrowsException<ArgumentException>(() => cache.Set("k", cert, "ep", " "));
+            Assert.ThrowsException<ArgumentException>(() => cache.Set(" ", new CertificateCacheValue(cert, "ep", "cid")));
+            Assert.ThrowsException<ArgumentNullException>(() => cache.Set("k", new CertificateCacheValue(null, "ep", "cid"))); // This will throw as expected in the test
+            Assert.ThrowsException<ArgumentException>(() => cache.Set("k", new CertificateCacheValue(cert, " ", "cid")));
+            Assert.ThrowsException<ArgumentException>(() => cache.Set("k", new CertificateCacheValue(cert, "ep", " ")));
 
             // Remove
             Assert.ThrowsException<ArgumentException>(() => cache.Remove(""));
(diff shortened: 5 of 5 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -27,9 +28,30 @@ internal class ManagedIdentityClient
         private X509Certificate2 _runtimeMtlsBindingCertificate;
         internal X509Certificate2 RuntimeMtlsBindingCertificate => Volatile.Read(ref _runtimeMtlsBindingCertificate);
 
+        // Central, process-local cache for mTLS binding (cert + endpoint + canonical client_id).
+        internal static readonly ICertificateCache s_mtlsCertificateCache = new InMemoryCertificateCache();
+
+        // Per-key async de-duplication so concurrent callers don’t double-mint.
+        internal static readonly ConcurrentDictionary<string, SemaphoreSlim> s_perKeyGates =
```
MZOLN: This design introduces unnecessary coupling and breaks encapsulation, leading to class envy. The responsibility should either reside where it’s used or be abstracted into a dedicated component.

Additionally, usage should avoid direct field access. Instead, expose functionality through well-defined internal or public methods behind a thoughtful interface.

For instance, why isn’t s_perKeyGates treated as an implementation detail of the imdsv2 component, given that it appears to be used only there? The same applies to the cache instance.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
```diff
@@ -0,0 +1,33 @@
...
+        internal static readonly TimeSpan MinRemainingLifetime = TimeSpan.FromHours(24);
+
+        public CertificateCacheEntry(X509Certificate2 certificate, DateTimeOffset notAfterUtc, string endpoint, string clientId)
+        {
+            Certificate = certificate ?? throw new MsalClientException(nameof(certificate));
+            NotAfterUtc = notAfterUtc;
+            Endpoint = endpoint ?? throw new MsalClientException(nameof(endpoint));
+            ClientId = clientId ?? throw new MsalClientException(nameof(clientId));
+        }
+
```
MZOLN, line 37: Wouldnt we need comments for public members?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ICertificateCache.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    internal interface ICertificateCache : IDisposable
+    {
+        /// <summary>
+        /// Try to get a cached certificate+endpoint+clientId for the specified cacheKey.
+        /// Returns true and non-null outputs if found and not expired.
+        /// </summary>
+        bool TryGet(
+            string cacheKey,
+            out X509Certificate2 certificate,
```
MZOLN: This is a very classic way of doing it, wouldnt it be better to use a tuple, valuetype result or even a record/structure here so that we have some flexibility with adding fields without breaking the contract?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ICertificateCache.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Process-local cache for an mTLS certificate and its endpoint.
+    /// Expiration is based solely on certificate.NotAfter.
+    /// </summary>
+    internal interface ICertificateCache : IDisposable
```
MZOLN: I'd say that it shouldnt inerit form IDisposable, but the implementation has.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+using System.Threading;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// certificate+endpoint+clientId cache stored in process memory.
+    /// </summary>
+    internal sealed class InMemoryCertificateCache : ICertificateCache
+    {
```
MZOLN, line 16: This should implement IDisposable, not the interface, imho

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+            out X509Certificate2 certificate, 
+            out string endpoint, 
+            out string clientId,
+            ILoggerAdapter logger = null)
+        {
+            certificate = null;
+            endpoint = null;
+            clientId = null;
+
+            if (cacheKey == null)
```
MZOLN: string.isnullorempty/whitespace?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+                throw new MsalClientException(nameof(cacheKey));
+            if (Volatile.Read(ref _disposed) != 0)
+                return false;
```
MZOLN: Shouldnt it be ObjectDisposedException?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// certificate+endpoint+clientId cache stored in process memory.
+    /// </summary>
+    internal sealed class InMemoryCertificateCache : ICertificateCache
+    {
+        private readonly ConcurrentDictionary<string, CertificateCacheEntry> _entriesByCacheKey =
```
MZOLN, line 17: How many entries do we predict to have here? For tiny maps or low contention a plain dictionary + coarse lock would have less overhead....

If we have read mostly and rare writes (as I imagined would be the case here) but need truly stable snapshots, immutabledictionary with copy-on write might be a better approach, as they are lock free, jsut higher allocation on writes...

But if a very small map or low contention dictionary with a single lock is often simpler and can be way faster.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+                throw new MsalClientException(nameof(cacheKey));
+            if (Volatile.Read(ref _disposed) != 0)
+                return false;
+
+            CertificateCacheEntry entry;
+            if (_entriesByCacheKey.TryGetValue(cacheKey, out entry))
+            {
+                var nowUtc = DateTimeOffset.UtcNow;
+
+                if (entry.IsExpiredUtc(nowUtc))
```
MZOLN: Could we extract it to a function to reduce nesting?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// <exception cref="MsalClientException"></exception>
+        /// <exception cref="ObjectDisposedException"></exception>
+        public void Set(
+            string cacheKey, 
+            X509Certificate2 certificate, 
+            string endpoint, 
+            string clientId,
+            ILoggerAdapter logger = null)
+        {
+            if (cacheKey == null)
```
MZOLN: String.isnullorempty/whitespace?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+
+            var notAfterUtc = ToNotAfterUtc(certificate);
+            var nowUtc = DateTimeOffset.UtcNow;
+
+            if (notAfterUtc <= nowUtc + CertificateCacheEntry.MinRemainingLifetime)
+            {
+                if (logger != null)
+                {
+                    var remaining = notAfterUtc - nowUtc;
+                    logger.Verbose(() => "[CertCache] Skipping certificate with insufficient remaining lifetime (" +
```
MZOLN: "... Skipping addition of certificate to cache, due to insufficient..."

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    internal sealed class CertificateCacheEntry : IDisposable
```
MZOLN, line 13: Could benefit form IsDisposed public property...

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+            if (_entriesByCacheKey.TryGetValue(cacheKey, out entry))
+            {
+                var nowUtc = DateTimeOffset.UtcNow;
+
+                if (entry.IsExpiredUtc(nowUtc))
+                {
+                    CertificateCacheEntry removed;
+                    if (_entriesByCacheKey.TryRemove(cacheKey, out removed))
+                    {
+                        removed.Dispose();
```
MZOLN: Looks to me like this is a commonly used pattern. Consider adding IsDisposed bool property to CacheEntry, then you could check if it already is disposed to make the code a little more defensive ?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+                cacheKey,
+                _ =>
+                {
+                    if (logger != null)
+                        logger.Verbose(() => "[CertCache] SET (key='" + Mask(cacheKey) + "').");
+                    return newEntry;
+                },
+                (_, old) =>
+                {
+                    old.Dispose();
```
MZOLN: See comemnt above... IsDisposed on cacheentry could make this code safer.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// <summary>
+        /// removes an entry fromn the cache, if present.
+        /// </summary>
+        /// <param name="cacheKey"></param>
+        /// <param name="logger"></param>
+        /// <returns></returns>
+        /// <exception cref="MsalClientException"></exception>
+        public bool Remove(string cacheKey, ILoggerAdapter logger = null)
+        {
+            if (cacheKey == null)
```
MZOLN: string.IsNullOrWhitespace/Empty (?)

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+                throw new MsalClientException(nameof(cacheKey));
+            CertificateCacheEntry entry;
+            if (_entriesByCacheKey.TryRemove(cacheKey, out entry))
+            {
+                entry.Dispose();
```
MZOLN: same

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// </summary>
+        /// <param name="logger"></param>
+        public void Clear(ILoggerAdapter logger = null)
+        {
+            foreach (var kvp in _entriesByCacheKey)
+            {
+                CertificateCacheEntry entry;
+                if (_entriesByCacheKey.TryRemove(kvp.Key, out entry))
+                {
+                    entry.Dispose();
```
MZOLN: Same

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// for testing: clears all entries.
+        /// </summary>
+        /// <param name="logger"></param>
+        public void Clear(ILoggerAdapter logger = null)
+        {
```
MZOLN, line 123: Needs _isdisposed check for consistency. Best extract it to a private method and call it from every function

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        /// <summary>
+        /// removes an entry fromn the cache, if present.
+        /// </summary>
+        /// <param name="cacheKey"></param>
+        /// <param name="logger"></param>
+        /// <returns></returns>
+        /// <exception cref="MsalClientException"></exception>
+        public bool Remove(string cacheKey, ILoggerAdapter logger = null)
+        {
```
MZOLN, line 105: Needs _disposed check for consistency. Extract it to a method and call from every public function/proprty.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+            var notAfter = cert.NotAfter;
+            if (notAfter.Kind == DateTimeKind.Unspecified)
+            {
+                notAfter = DateTime.SpecifyKind(notAfter, DateTimeKind.Local);
+            }
+            return new DateTimeOffset(notAfter.ToUniversalTime());
+        }
+
+        /// <summary>
+        /// used for logging cache keys without exposing full values.
```
MZOLN: nit: Start from capital letter

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// </summary>
+        /// <param name="s"></param>
```
MZOLN: Should we fill those up as well?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// <returns></returns>
+        private static string Mask(string s)
+        {
+            if (string.IsNullOrEmpty(s))
+                return "<empty>";
+            var take = Math.Min(8, s.Length);
```
MZOLN: so it would expose full value if it is less than 8? Is that ok?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -736,5 +740,794 @@ await managedIdentityApp.AcquireTokenForManagedIdentity(ManagedIdentityTests.Res
...
+
+        #region cached certificate tests
+        [TestMethod]
+        public async Task mTLSPop_ForceRefresh_UsesCachedCert_NoIssueCredential_PostsCanonicalClientId_AndSkipsAttestation()
+        {
+            using (new EnvVariableContext())
+            using (var httpManager = new MockHttpManager())
+            {
+                // Start clean across tests
+                ManagedIdentityClient.ResetSourceForTest();
```
MZOLN: How do you ensure that those tests dont race? Because we are using static state here, this could cause instabilities.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -27,9 +28,30 @@ internal class ManagedIdentityClient
...
+            new ConcurrentDictionary<string, SemaphoreSlim>(StringComparer.Ordinal);
+
         internal static void ResetSourceForTest()
         {
             s_sourceName = ManagedIdentitySource.None;
```
MZOLN, line 33: Since we leak the dependency to test, this utility metod can belong in test itself....
A better approach would be to expose appropriate public methods on well encapsulated components.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// dispose the entry and its certificate.
+        /// </summary>
+        public void Dispose()
+        {
+            if (Interlocked.Exchange(ref _disposed, 1) != 0)
+            {
+                return; // already disposed
+            }
+
+            try
```
MZOLN: try/catch is expensive. You won't double dispose with that atomic switch you got on line 70, would you?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ICertificateCache.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// Returns true and non-null outputs if found and not expired.
+        /// </summary>
+        bool TryGet(string cacheKey, out CertificateCacheValue value, ILoggerAdapter logger = null);
+
+        /// <summary>
+        /// Insert or replace the cached certificate+endpoint+clientId for cacheKey.
+        /// </summary>
+        void Set(
+            string cacheKey,
+            X509Certificate2 certificate,
```
MZOLN: should we adjust the set as well? This looks better actually ;)

gladjohn: agree - missed updating this

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: moved the one in IMDS V2 to it's own method and call it from here

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
gladjohn: removed this - and each test before initializing will always call `ResetSourceForTest`, so the static caches are cleared in PRODUCT.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated but this is internal

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: no, this is skipping because we want to mint new cert

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: not many entries - 60 max for a VM, but depending on the client, they all may be used at once.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: No because this is just the client_id app dev passes to MSAL. But the cert was set with ObjectDisposedException

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: yes

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InMemoryCertificateCache.cs
gladjohn: agreed, updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ICertificateCache.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/CertificateCacheEntry.cs
gladjohn: these are internal

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: udpated

PR #5559, opened 2025-10-29
Make `IMsalMtlsHttpClientFactory` interface public

cpp11nullptr:
Make `IMsalMtlsHttpClientFactory` interface public so downstream components can use it for its mTLS PoP implementation.


global.json
```diff
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "8.0.415",
+        "version": "9.0.306",
```
MZOLN: Not sure if you can do that @gladjohn ?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -150,6 +150,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Identity.Abstractions" />
```
MZOLN: Is this needed?

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -336,5 +359,30 @@ public string CreateAuthorizationHeader()
...
+        /// <code>
+        /// var authHeader = result.CreateAuthorizationHeaderBound();
+        /// HttpClientHandler handler = new();
+        /// handler.ClientCertificates.Add(authHeader.BindingCertificate);
+        /// HttpClient client = new HttpClient(handler);
+        /// client.DefaultRequestHeaders.Add("Authorization", authHeader.AuthorizationHeaderValue);
+        /// HttpResponseMessage r = await client.GetAsync(urlOfTheProtectedApi);
+        /// </code>
+        /// </example>
+        public AuthorizationHeaderInformation CreateAuthorizationHeaderBound()
```
MZOLN: Wouldnt CreateAuthorizationHeaderInformation be a better name here? In case we would add more parameters in the future it would become more flexible, vs creating dedicated methods for each use case.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -7,18 +7,18 @@
...
     /// <remarks>
     /// Implementations of this interface must be thread-safe.
     /// It is important to reuse HttpClient instances to avoid socket exhaustion.
     /// Do not create a new HttpClient for each call to <see cref="GetHttpClient(X509Certificate2)"/>.
     /// If your application requires Integrated Windows Authentication, set <see cref="HttpClientHandler.UseDefaultCredentials"/> to true.
-    /// This interface is intended for internal use by MSAL only and is designed to support MTLS scenarios.
+    /// This interface is designed to support MTLS scenarios.
     /// </remarks>
-    internal interface IMsalMtlsHttpClientFactory : IMsalHttpClientFactory
+    public interface IMsalMtlsHttpClientFactory : IMsalHttpClientFactory
```
MZOLN, line 21: Does this need to extend IMsalHttpClientFactory?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
cpp11nullptr: As it was discussed offline, it's needed to be able to add helper method for `AuthenticationResult` which will return `AuthorizationHeaderInformation`. Alternatively, this helper method could go into `Microsoft.Identity.Web` (so no dependency of `Microsoft.Identity.Abstractions` is needed there) - but it'd make it less coupled with `AuthenticationResult`. I don't have strong opinion and happy with both variants.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
cpp11nullptr: In `HttpManager`, HTTP client factory is held as `IMsalHttpClientFactory` with [casting](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/3ea20b4a0db87be0d85f32e2e6a50957d2b71d53/src/client/Microsoft.Identity.Client/Http/HttpManager.cs#L210) it to `IMsalMtlsHttpClientFactory` if client certificate is present.

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
cpp11nullptr: Good point, renamed.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
bgavrilMS: @MZOLN - since `IMsalHttpClientFactory` is a public interface, we cannot simply add a method to it. We chose to add `IMsalMtlsHttpClientFactory` instead for the new functionality instead.

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -348,7 +341,8 @@ protected async Task<AuthenticationResult> CacheTokenResponseAndCreateAuthentica
                 AuthenticationRequestParameters.RequestContext.ApiEvent,
                 account,
                 msalTokenResponse.SpaAuthCode,
-                msalTokenResponse.CreateExtensionDataStringMap());
+                msalTokenResponse.CreateExtensionDataStringMap(),
+                AuthenticationRequestParameters.AppConfig.ClientCredentialCertificate);
```
bgavrilMS: @cpp11nullptr - are you sure this is needed? I think the cert is set here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AuthScheme/PoP/MtlsPopAuthenticationOperation.cs#L41

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
```diff
@@ -338,5 +361,30 @@ public string CreateAuthorizationHeader()
...
+        /// <code>
+        /// var authHeader = result.CreateAuthorizationHeaderInformation();
+        /// HttpClientHandler handler = new();
+        /// handler.ClientCertificates.Add(authHeader.BindingCertificate);
+        /// HttpClient client = new HttpClient(handler);
+        /// client.DefaultRequestHeaders.Add("Authorization", authHeader.AuthorizationHeaderValue);
+        /// HttpResponseMessage r = await client.GetAsync(urlOfTheProtectedApi);
+        /// </code>
+        /// </example>
+        public AuthorizationHeaderInformation CreateAuthorizationHeaderInformation()
```
bgavrilMS: Since MSAL doesn't actually use `AuthorizationHeaderInformation`, does it make sense to include this in Identity.Web only?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
bgavrilMS: Yes, I would suggest to do this work in Id.Web

src/client/Microsoft.Identity.Client/AuthenticationResult.cs
cpp11nullptr: Sounds good for me.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -7,18 +7,18 @@
 namespace Microsoft.Identity.Client
 {
     /// <summary>
-    /// Internal factory responsible for creating HttpClient instances configured for mutual TLS (MTLS).
-    /// This factory is specifically intended for use within the MSAL library for secure communication with Azure AD using MTLS.
+    /// A factory responsible for creating HttpClient instances configured for mutual TLS (MTLS).
```
gladjohn:
```suggestion
    /// A factory responsible for creating HttpClient instances configured for mutual TLS (mTLS).
```

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -7,18 +7,18 @@
...
+    /// This factory is intended for use to secure communication with Azure AD using MTLS.
```
gladjohn:
```suggestion
    /// This factory is intended for use to secure communication with Azure AD using mTLS.
```

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -7,18 +7,18 @@
...
     /// For more details on HttpClient instancing, see https://learn.microsoft.com/dotnet/api/system.net.http.httpclient?view=net-7.0#instancing.
     /// </summary>
     /// <remarks>
     /// Implementations of this interface must be thread-safe.
     /// It is important to reuse HttpClient instances to avoid socket exhaustion.
     /// Do not create a new HttpClient for each call to <see cref="GetHttpClient(X509Certificate2)"/>.
     /// If your application requires Integrated Windows Authentication, set <see cref="HttpClientHandler.UseDefaultCredentials"/> to true.
-    /// This interface is intended for internal use by MSAL only and is designed to support MTLS scenarios.
+    /// This interface is designed to support MTLS scenarios.
```
gladjohn:
```suggestion
    /// This interface is designed to support mTLS scenarios.
```

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
```diff
@@ -7,18 +7,18 @@
 namespace Microsoft.Identity.Client
 {
     /// <summary>
-    /// Internal factory responsible for creating HttpClient instances configured for mutual TLS (MTLS).
-    /// This factory is specifically intended for use within the MSAL library for secure communication with Azure AD using MTLS.
+    /// A factory responsible for creating HttpClient instances configured for mutual TLS (MTLS).
+    /// This factory is intended for use to secure communication with Azure AD using MTLS.
     /// For more details on HttpClient instancing, see https://learn.microsoft.com/dotnet/api/system.net.http.httpclient?view=net-7.0#instancing.
```
gladjohn: should be updated to use the net 8 docs

gladjohn:
think the only change needed is to expose `IMsalMtlsHttpClientFactory`. The Binding cert is already exposed in Authentication Result. There are [existing tests](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs) that already validates this.

src/client/Microsoft.Identity.Client/AppConfig/IMsalMtlsHttpClientFactory.cs
cpp11nullptr: Done.

PR #5560, opened 2025-10-29
ImdsV2: Temporary Workaround for Resource Id Query Param Bug in IMDS

Robbie-Microsoft:
IMDS server is supposed to accept `msi_res_id` as a query param for Resource Id.

We are temporarily adjusting the query param: `msi_res_id` -> `mi_res_id`


src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -99,7 +101,7 @@ protected override Task<ManagedIdentityRequest> CreateRequestAsync(string resour
 
                 case AppConfig.ManagedIdentityIdType.ResourceId:
                     logger?.Info("[Managed Identity] Adding user assigned resource id to the request.");
-                    return new KeyValuePair<string, string>(Constants.ManagedIdentityResourceIdImds, userAssignedId);
+                    return new KeyValuePair<string, string>(imdsV1 ? Constants.ManagedIdentityResourceIdImds : Constants.ManagedIdentityResourceId, userAssignedId);
```
gladjohn, line 104: resource id?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -75,7 +75,8 @@ protected override Task<ManagedIdentityRequest> CreateRequestAsync(string resour
             var userAssignedIdQueryParam = GetUserAssignedIdQueryParam(
                 _requestContext.ServiceBundle.Config.ManagedIdentityId.IdType,
                 _requestContext.ServiceBundle.Config.ManagedIdentityId.UserAssignedId,
-                _requestContext.Logger);
+                _requestContext.Logger,
+                imdsV1: true);
```
gladjohn, line 79: can we add a test for this Robbie? by capturing the outgoing request

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: what?

gladjohn: yeah, it was mis-leading. sorry, just to avoid confusion should we create a new ManagedIdentityResourceIdImdsV2

Robbie-Microsoft: The constant already exists as `ManagedIdentityResourceId`. Do you want to create another constant with the same value?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: It's covered already by existing mocks

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
gladjohn: no - think out loud that will lead to more confusion later

PR #5566, opened 2025-11-03
IMDSv2 mTLS PoP: add best‑effort persisted binding‑cert cache (Windows CurrentUser\My) layered over in‑memory cache; per‑alias mutex; tests

gladjohn:
Fixes - Adds MSI v2 cert to the store (persistent cache)

The IMDSv2 mTLS PoP flow currently relies only on a process‑local cache. That means new processes (or app restarts) must re‑mint the binding certificate even when a valid one already exists on the machine. This PR layers a best‑effort, cross‑process persisted cache on top of the existing in‑memory cache to reduce reminting, while keeping token acquisition robust (never blocked by persistence).

**Testing**
unit testing

**Performance impact**
none

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
@@ -228,7 +228,7 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
 
                             // Add the original instance (carries private key if present)
                             store.Add(cert);
-                            logger?.Info(() => "[PersistentCert] Persisted certificate to CurrentUser/My.");
+                            logger.Info(() => "[PersistentCert] Persisted certificate to CurrentUser/My.");
 
                             // Conservative cleanup: remove expired entries for this alias only
                             PruneExpiredForAlias(store, alias, nowUtc, logger);


src/client/Microsoft.Identity.Client/ManagedIdentity/V2/FriendlyNameCodec.cs
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Encodes/decodes the persisted X.509 FriendlyName for MSAL mTLS certs.
+    /// Format: "MSAL|alias=cacheKey|ep=endpointBase"
+    /// Open the cert store and look at FriendlyName to see examples.
+    /// Wish we could paste a screenshot here... Maybe I can show it in code walkthroughs.
+    /// </summary>
+    internal static class FriendlyNameCodec
```
bgavrilMS: naming: MsiCertificateFriendlyNameEncoder ?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/FriendlyNameCodec.cs
```diff
@@ -0,0 +1,107 @@
...
+
+using System;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Encodes/decodes the persisted X.509 FriendlyName for MSAL mTLS certs.
+    /// Format: "MSAL|alias=cacheKey|ep=endpointBase"
+    /// Open the cert store and look at FriendlyName to see examples.
+    /// Wish we could paste a screenshot here... Maybe I can show it in code walkthroughs.
```
bgavrilMS: ?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/FriendlyNameCodec.cs
```diff
@@ -0,0 +1,107 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Encodes/decodes the persisted X.509 FriendlyName for MSAL mTLS certs.
+    /// Format: "MSAL|alias=cacheKey|ep=endpointBase"
```
bgavrilMS: Can you give more examples here? This format is not clear.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MsiCertificateFriendlyNameEncoder.cs
```diff
@@ -0,0 +1,107 @@
...
+        public const string TagEp = "ep";
+
+        /// <summary>
+        /// Encodes alias and endpointBase into friendly name.
+        /// </summary>
+        /// <param name="alias"></param>
+        /// <param name="endpointBase"></param>
+        /// <param name="friendlyName"></param>
+        /// <returns></returns>
+        public static bool TryEncode(string alias, string endpointBase, out string friendlyName)
```
bgavrilMS, line 64: nit: Avoid using these `TryXYZ` methods for cases where exceptions should be thrown.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/FriendlyNameCodec.cs
bgavrilMS: what is alias, cachekey, ep and endpoint base?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MsiCertificateFriendlyNameEncoder.cs
```diff
@@ -0,0 +1,107 @@
...
+                    endpointBase = v;
+                }
+            }
+
+            return !string.IsNullOrWhiteSpace(alias) && !string.IsNullOrWhiteSpace(endpointBase);
+        }
+
+        /// <summary>
+        /// Checks for illegal characters in alias/endpointBase.
+        /// Endpoint itself comes from IMDS and is well-formed, but we still validate.
```
bgavrilMS, line 131: Why not throw?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -457,21 +455,13 @@ private static async Task<Tuple<X509Certificate2, string, string>> GetOrCreateMt
             if (factory is null)
                 throw new ArgumentNullException(nameof(factory));
 
-            X509Certificate2 cachedCertificate;
-            string cachedEndpointBase;
-            string cachedClientId;
-
-            // 1) Only lookup by cacheKey
+            // 1) In-memory cache first
```
bgavrilMS: This 1) 2) 3) logic should be encapsualted in a different object, some sort of Cache abstraction.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,95 @@
...
+using System;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
+using Microsoft.Identity.Client.PlatformsCommon.Shared;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Cross-process lock based on a per-alias named mutex.
```
bgavrilMS: What is an alias?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,95 @@
...
+    internal static class InterprocessLock
+    {
+        public static bool TryWithAliasLock(
+            string alias,
+            TimeSpan timeout,
+            Action action,
+            Action<string> logVerbose = null)
+        {
+            var nameGlobal = GetMutexNameForAlias(alias, preferGlobal: true);
+            var nameLocal = GetMutexNameForAlias(alias, preferGlobal: false);
```
bgavrilMS: Explain in comment why you need 2 mutexes

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,95 @@
...
+                    return false;
+                }
+            }
+
+            return false;
+        }
+
+        public static string GetMutexNameForAlias(string alias, bool preferGlobal = true)
+        {
+            string suffix = HashAlias(Canonicalize(alias));
```
bgavrilMS, line 126: Do you need to hash? How long are the aliases? Maybe you can avoid hashing here.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+    /// 1) Filter by FriendlyName "MSAL|alias=&lt;cacheKey&gt;|ep=&lt;base&gt;" (scopes to *our* alias).
+    /// 2) Enforce remaining lifetime (≥ 24h) and require HasPrivateKey; pick newest NotAfter.
+    /// 3) On that single winner, read Subject CN and require a GUID (canonical client_id).
+    ///
+    /// Notes:
+    /// - We never reattach private keys. If HasPrivateKey == false, caller will mint.
+    /// - No throws; persistence must not block token acquisition.
+    /// - Windows-only; FriendlyName semantics are undefined elsewhere.
+    /// </summary>
+    internal static class PersistentCertificateStore
```
bgavrilMS: Are you overusing static? Everything seems static in this design. Does this not need mocking? And a logger?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+    {
+        public static bool TryFind(
```
bgavrilMS: This is not a good interface for a cache - use Read / Write / Delete for caching.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+            string aliasCacheKey,
+            out CertificateCacheValue value,
+            ILoggerAdapter logger = null)
+        {
+            value = default;
+
+            if (!DesktopOsHelper.IsWindows())
```
bgavrilMS: Why? The X509 stores exist on Linux as well.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+                catch
+                {
+                    items = store.Certificates.Cast<X509Certificate2>().ToArray();
+                }
+
+                X509Certificate2 best = null;
+                string bestEndpoint = null;
+                DateTime bestNotAfter = DateTime.MinValue;
+
+                foreach (var c in items)
```
bgavrilMS: please use more meaningful variable names. Not `c`

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+                store.Open(OpenFlags.OpenExistingOnly | OpenFlags.ReadOnly);
+
+                // Snapshot to avoid "collection modified" during enumeration on some providers
+                X509Certificate2[] items;
+                try
+                {
+                    items = new X509Certificate2[store.Certificates.Count];
+                    store.Certificates.CopyTo(items, 0);
+                }
+                catch
```
bgavrilMS: All of these try / catch should log

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+                        c.Dispose();
+                    }
+                }
+
+                if (best != null)
+                {
+                    // CN (GUID) → canonical client_id
+                    string cn = null;
+                    try
+                    { cn = best.GetNameInfo(X509NameType.SimpleName, false); }
```
bgavrilMS: Please format all docs and use proper indentation.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+                    catch { }
```
bgavrilMS: log all failures.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
bgavrilMS: IMO, a better design would have been: 

- a local in-memory cache logic
- a cert store logic
- an orchestrator

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+            }
+
+            return false;
+        }
+
+        public static void TryPersist(
+            string aliasCacheKey,
+            X509Certificate2 cert,
+            string endpointBase,
+            string clientId,           // unused; CN is the source of truth on read
```
bgavrilMS: If it's unused, why do you add it?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
```diff
@@ -0,0 +1,259 @@
...
+            ILoggerAdapter logger = null)
+        {
+            if (!DesktopOsHelper.IsWindows() || cert == null)
+                return;
+
+            // We only persist certs that can actually be used for mTLS later.
+            if (!cert.HasPrivateKey)
+            {
+                logger?.Verbose(() => "[PersistentCert] Not persisting: certificate has no private key.");
```
bgavrilMS: Seems like a warning to me? Or even an exception? Why would you persist certs without private key?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/StorePruner.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Removes expired entries (NotAfter less than  now) for an alias from an open X509 store.
+    /// </summary>
+    internal static class StorePruner
```
bgavrilMS: Doesn't seem worth having another (static!) class for 1 mehtod. Might as well move it to PersistentCertificaetStore? It's a tpyical cache op.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/FriendlyNameCodec.cs
gladjohn: removed it - sometimes i just talk to myself I think

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MsiCertificateFriendlyNameEncoder.cs
gladjohn: we do not want any exceptions to be thrown in this. Persisting the cert in the store is based on best-effort, we do not want to stop authentication because we could not persist to the store. We will continue with in-memory.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/FriendlyNameCodec.cs
gladjohn: added

gladjohn: added comments

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MsiCertificateFriendlyNameEncoder.cs
gladjohn: we do not want to block auth for persistence cert feature, if persistence fails, we rely on in-memory cache. Cert persistence is always best effort.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: added.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: it's the managed identity id, added comments.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: added

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: yes, because mi_res_id can run long. 

image

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
gladjohn: we don't need mocking here, but made few changes.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
gladjohn: We do not have a MSI v2 Linux story defined yet and that is for phase 2. and the Friendly name tagging won't work. IMDS team has a bug on them to fix the endpoints and we may need to remove the friendly name tag to support this on Linux.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
gladjohn: added logs

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
gladjohn: that's how it is now

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
gladjohn: removed it

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateStore.cs
gladjohn: You are correct, we do not need this check.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/StorePruner.cs
gladjohn: removed

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -195,7 +193,12 @@ public static AbstractManagedIdentity Create(RequestContext requestContext)
 
         internal ImdsV2ManagedIdentitySource(RequestContext requestContext) :
             base(requestContext, ManagedIdentitySource.ImdsV2)
-        { }
+        {
+            IPersistentCertificateCache persisted = 
+                PersistentCertificateCacheFactory.Create(requestContext.Logger);
+
+            _mtlsCache = new MtlsBindingCache(s_mtlsCertificateCache, persisted);
+        }
```
Robbie-Microsoft: Can we use dependency injection instead of putting this in the constructor? This should enhance testability.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IMtlsBindingCache.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    internal interface IMtlsBindingCache
+    {
+        Task<Tuple<X509Certificate2, string /*endpoint*/, string /*clientId*/>> GetOrCreateAsync(
```
Robbie-Microsoft: Can you create a named tuple?

```
internal class MtlsBindingInfo
{
    public X509Certificate2 Certificate { get; set; }
    public string Endpoint { get; set; }
    public string ClientId { get; set; }
}
```

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            var nameGlobal = GetMutexNameForAlias(alias, preferGlobal: true);
+            var nameLocal = GetMutexNameForAlias(alias, preferGlobal: false);
+
+            foreach (var name in new[] { nameGlobal, nameLocal })
+            {
+                try
+                {
+                    // Create or open existing
+                    using var m = new Mutex(initiallyOwned: false, name);
```
Robbie-Microsoft: what if an exception occurs before the using block exposes?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,123 @@
...
+                    // Wait to acquire
+                    bool entered;
+
+                    try
+                    {
+                        entered = m.WaitOne(timeout);
+                    }
+                    catch (AbandonedMutexException)
+                    {
+                        entered = true; // prior holder crashed
```
Robbie-Microsoft: can you add a log here?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,123 @@
...
+                    }
+
+                    if (!entered)
+                    {
+                        logVerbose?.Invoke($"[PersistentCert] Skip persist (lock busy '{name}').");
```
Robbie-Microsoft: should we log duration waited?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IMtlsBindingCache.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    internal interface IMtlsBindingCache
```
Robbie-Microsoft: Can you add XML docs here to be consistent with the rest of this PR?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IPersistentCertificateCache.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <summary>
+        /// Persists the certificate for the alias (best-effort). Implementations should
+        /// tag entries to allow alias scoping and prune expired duplicates conservatively.
+        /// </summary>
+        void Write(string alias, X509Certificate2 cert, string endpointBase, ILoggerAdapter logger = null);
+
+        /// <summary>
+        /// Prunes expired entries for the alias (best-effort).
+        /// </summary>
+        void Delete(string alias, ILoggerAdapter logger = null);
```
Robbie-Microsoft: should these return bool, like `read`, to indicate success or failure?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MsiCertificateFriendlyNameEncoder.cs
```diff
@@ -0,0 +1,146 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: Should we add logging in this file?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
```diff
@@ -0,0 +1,95 @@
...
+                    logger?.Verbose(() => "[PersistentCert] Skipping persisted cert without private key; minting new.");
+                }
+
+                // 4) Mint + back-fill mem + best-effort persist + prune
+                var created = await factory().ConfigureAwait(false);
+                var createdValue = new CertificateCacheValue(created.Item1, created.Item2, created.Item3);
+
+                _memory.Set(cacheKey, in createdValue, logger);
+                _persisted.Write(cacheKey, created.Item1, created.Item2, logger);
+                _persisted.Delete(cacheKey, logger);
```
Robbie-Microsoft, line 136: can you add a comment here? Why are you deleting right after you write?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsBindingCache.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <summary>
+        /// Inject both caches to avoid global state and enable testing.
+        /// </summary>
+        public MtlsBindingCache(ICertificateCache memory, IPersistentCertificateCache persisted)
+        {
+            _memory = memory ?? throw new ArgumentNullException(nameof(memory));
+            _persisted = persisted ?? throw new ArgumentNullException(nameof(persisted));
+        }
+
+        public async Task<Tuple<X509Certificate2, string, string>> GetOrCreateAsync(
```
Robbie-Microsoft: Can you use ValueTuple for clarity?

```
public async Task<(X509Certificate2 Certificate, string Endpoint, string ClientId)> GetOrCreateAsync(...)```

```

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
```diff
@@ -0,0 +1,95 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: should we add logging for cache hit/misses?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsBindingCache.cs
```diff
@@ -0,0 +1,95 @@
...
+            CancellationToken cancellationToken,
+            ILoggerAdapter logger)
+        {
+            if (string.IsNullOrWhiteSpace(cacheKey))
+                throw new ArgumentException("cacheKey must be non-empty.", nameof(cacheKey));
+            if (factory is null)
+                throw new ArgumentNullException(nameof(factory));
+
+            // 1) In-memory cache first
+            if (_memory.TryGet(cacheKey, out var cached, logger))
```
Robbie-Microsoft: can you give `cached` a clearer name? `cachedCert` perhaps?

Same idea for `persisted` and  `createdValue`

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/NoOpPersistentCertificateCache.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: should we add logging to this file?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Identity.Client.Core;
+using Microsoft.Identity.Client.PlatformsCommon.Shared;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    internal static class PersistentCertificateCacheFactory
+    {
+        public static IPersistentCertificateCache Create(ILoggerAdapter logger)
```
Robbie-Microsoft, line 14: you pass logger in, but it's not used. Can you add logs?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
```diff
@@ -0,0 +1,336 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: Do we want all logs in this file to be `verbose`? Should they be `info` instead?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
```diff
@@ -0,0 +1,336 @@
...
+            if (!MsiCertificateFriendlyNameEncoder.TryEncode(alias, endpointBase, out var friendlyName))
+            {
+                logger?.Verbose(() => "[PersistentCert] FriendlyName encode failed; skipping persist.");
+                return;
+            }
+
+            // Best-effort: short lock, skip if busy
+            InterprocessLock.TryWithAliasLock(
+                alias,
+                timeout: TimeSpan.FromMilliseconds(300),
```
Robbie-Microsoft, line 157: Should this be configurable? Should we add a retry loop?

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/KeyedSemaphorePool.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: If you remove a semaphore from the dictionary, should you also dispose it to avoid leaks?

```
if (_gates.TryRemove(key, out var removed))
{
    removed.Dispose();
}
```

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/FriendlyNameCodecTests.cs
```diff
@@ -0,0 +1,86 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: can you add test cases:
- Very long alias and endpoint values.
- Aliases and endpoints containing unicode or special characters (besides those already filtered).
- Double-encoded inputs or already-prefixed strings to ensure decoder resilience.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/FriendlyNameCodecTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            var bad = "NOTMSAL|alias=a|ep=b";
+            Assert.IsFalse(MsiCertificateFriendlyNameEncoder.TryDecode(bad, out _, out _));
+
+            // Missing tags
+            var missing = MsiCertificateFriendlyNameEncoder.Prefix + "alias=a";
+            Assert.IsFalse(MsiCertificateFriendlyNameEncoder.TryDecode(missing, out _, out _));
+        }
+
+        [TestMethod]
+        public void EncodeDecode_Roundtrip()
```
Robbie-Microsoft: how is this different from `EncodeDecode_RoundTrip_Succeeds`?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/FriendlyNameCodecTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            string alias = "my-alias-123";
+            string ep = "https://ep/base";
+            Assert.IsTrue(MsiCertificateFriendlyNameEncoder.TryEncode(alias, ep, out var fn));
+            Assert.IsTrue(MsiCertificateFriendlyNameEncoder.TryDecode(fn, out var a2, out var e2));
+            Assert.AreEqual(alias, a2);
+            Assert.AreEqual(ep, e2);
+        }
+
+        [TestMethod]
+        public void Encode_Rejects_Illegal()
```
Robbie-Microsoft: For `TryEncode_Rejects_IllegalChars` and `Encode_Rejects_Illegal`, can you use [DataRow] instead?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2TestStoreCleaner.cs
```diff
@@ -0,0 +1,238 @@
...
+        public static bool IsWindows => RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+
+        /// <summary>
+        /// Remove all persisted entries that look like test artifacts:
+        /// - FriendlyName starts with "MSAL|" and either has a fake endpoint or is expired-by-policy
+        /// - Subject CN matches test CNs (UAMI-20Y, SAMI-20Y, Test)
+        /// - Subject CN matches certain values AND subject DC equals the test tenant
+        /// Best-effort, no-throw.
+        /// </summary>
+        public static void RemoveAllTestArtifacts()
```
Robbie-Microsoft, line 43: Both `RemoveAllTestArtifacts()` and `RemoveAlias()` have near-identical code for snapshotting the certificates from the store. Can you refactor it into a private helper method?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2TestStoreCleaner.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+            catch
+            {
+                // best-effort
+            }
+        }
+
+        // ---------------- helpers ----------------
+
+        private static bool MatchesSubjectTrash(string cn, string dc)
```
Robbie-Microsoft, line 176: Can you add XML docs for the helper methods?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2TestStoreCleaner.cs
```diff
@@ -0,0 +1,238 @@
...
+        {
+            if (string.IsNullOrEmpty(cn))
+                return false;
+
+            // Unconditional CNs
```
Robbie-Microsoft, line 181: Why is there a difference in string comparison between conditional and unconditional CNs? (Ordinal vs. OrdinalIgnoreCase)

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/InterprocessLockTests.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: can you add additional tests:
 
- Edge cases:
   - Invalid or null alias values to ensure proper exception handling or fallback.
   - Very long alias strings to check for buffer overflows or system limitations.
   - Non-Windows platforms
- Multiple concurrent attempts
- Exception handling within action: Add tests where the action delegate throws an exception—confirm lock is released and exception bubbles as expected.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/InterprocessLockTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        private static bool IsWindows => RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+
+        [TestMethod]
+        public void GetMutexName_Format_And_Canonicalization()
+        {
+            if (!IsWindows)
+            { Assert.Inconclusive("Windows-only"); return; }
+
+            var aliasRaw = "  my-alias  ";
+            var g = InterprocessLock.GetMutexNameForAlias(aliasRaw, preferGlobal: true);
```
Robbie-Microsoft: please use global/local for readability

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/PersistentCertificateStoreUnitTests.cs
```diff
@@ -0,0 +1,553 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Robbie-Microsoft, line 1: can you add additional tests:
- for non-windows behavior, validate intended no-op/error/alternative behavior
- Add tests to validate behavior when mutex acquisition fails due to errors besides contention (e.g., invalid alias, OS failures), or when the action throws an exception—does the lock get released?
- Test that repeated calls for the same alias succeed serially—i.e., after the first lock is released, the second can acquire.
- Test boundary and edge cases for aliases: empty string, extremely long name, special characters, unicode, etc.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/PersistentCertificateStoreUnitTests.cs
```diff
@@ -0,0 +1,553 @@
...
+// Licensed under the MIT License.
```
Robbie-Microsoft, line 2: Use private helper methods to reduce repeated code for Windows checks or lock calls

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IMtlsBindingCache.cs
gladjohn: created a named tuple

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: If new Mutex(...) throws, the outer catch (Exception ex) in TryWithAliasLock now logs the failure (via the provided logging delegate) and returns false. That keeps persistence best‑effort.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: When WaitOne throws AbandonedMutexException, we now log a verbose message indicating that we’re treating the mutex as acquired, then continue.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: Wrapped the WaitOne call in a Stopwatch and log the elapsed time in the “lock busy, skip persist” path.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IMtlsBindingCache.cs
gladjohn: I added it, but @bgavrilMS does not want me to. So this has been very confusing for me

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IPersistentCertificateCache.cs
gladjohn: no, it shouldn't as this is best effort. But we do have logging.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MsiCertificateFriendlyNameEncoder.cs
gladjohn: I left the codec itself logging‑free and non‑throwing, returning false on invalid input.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
gladjohn: Write persists the newest binding for a given alias, and Delete is used to prune expired entries for that alias only, so the store doesn't grow unbounded over time. Added code comments to clarify

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsBindingCache.cs
gladjohn: dupe comment as before. fixed already

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
gladjohn: Added Verbose logging for each branch: memory hit, persistent hit, and mint. That lets us see where bindings are coming from (memory vs store vs factory) when verbose logging is enabled.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsBindingCache.cs
gladjohn: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/NoOpPersistentCertificateCache.cs
gladjohn: I left NoOpPersistentCertificateCache itself silent and log instead at factory time when it’s selected.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
gladjohn: PersistentCertificateCacheFactory.Create now logs an Info message indicating which implementation is selected: WindowsPersistentCertificateCache on Windows and NoOpPersistentCertificateCache on other platforms. That makes the persistence mode visible in logs.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
gladjohn: we are logging too many things, the necessary ones are marked as info.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
gladjohn: Kept a single short, non‑configurable timeout and no retry, and added a comment. Persistent store access is explicitly best‑effort and must not block token acquisition; if the lock is busy we skip the write/delete and fall back to the in‑memory cache instead of retrying.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/KeyedSemaphorePool.cs
gladjohn: In the current design we no longer remove semaphores from the dictionary; each key uses a single SemaphoreSlim for the lifetime of the pool. That avoids races where a semaphore could be disposed while another caller still holds a reference. We also don’t touch AvailableWaitHandle, so relying on GC for clean‑up is sufficient for our usage.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/FriendlyNameCodecTests.cs
gladjohn: added

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/FriendlyNameCodecTests.cs
gladjohn: removed

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/FriendlyNameCodecTests.cs
gladjohn: updated

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2TestStoreCleaner.cs
gladjohn: done

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2TestStoreCleaner.cs
gladjohn: the method names themselves are self-explanatory, but still added.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2TestStoreCleaner.cs
gladjohn: Switched both unconditional and conditional CN comparisons to use StringComparison.OrdinalIgnoreCase. For these test‑only certificates there’s no benefit to different semantics, and a consistent case‑insensitive comparison avoids subtle mismatches.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/InterprocessLockTests.cs
gladjohn: added

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/PersistentCertificateStoreUnitTests.cs
gladjohn: added

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/PersistentCertificateStoreUnitTests.cs
gladjohn: done

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,133 @@
...
+                var hex = new CommonCryptographyManager().CreateSha256HashHex(s);
+                // Truncate to 32 chars to fit mutex name length limits
+                return string.IsNullOrEmpty(hex)
+                    ? "0"
+                    : (hex.Length > 32 ? hex.Substring(0, 32) : hex);
+            }
+            catch
+            {
+                return "0";
+            }
```
Robbie-Microsoft: nit: should you make constants for these?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IMtlsBindingCache.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Core;
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
+{
+    /// <summary>
+    /// Abstraction over the in-memory + persisted cache for IMDSv2 mTLS binding certificates.
+    /// </summary>
+    internal interface IMtlsBindingCache
```
bgavrilMS: maybe call it mtlsCertificateCache? Binding is overloaded word

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,133 @@
...
+        public static bool TryWithAliasLock(
+            string alias,
+            TimeSpan timeout,
+            Action action,
+            Action<string> logVerbose = null)
+        {
+            var nameGlobal = GetMutexNameForAlias(alias, preferGlobal: true);
+            var nameLocal = GetMutexNameForAlias(alias, preferGlobal: false);
+
+            foreach (var name in new[] { nameGlobal, nameLocal })
```
bgavrilMS: nit: instead of a for loop, can you extract a local method and call it twice, one after the other?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                try
+                {
+                    // Create or open existing. If this throws (e.g. ACL issues),
+                    // the outer catch below logs and returns false (best-effort).
+                    using var m = new Mutex(initiallyOwned: false, name);
```
bgavrilMS: pls use meaningful var names.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IPersistentCertificateCache.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Implementations must be best-effort and non-throwing so that
+    /// certificate persistence never blocks authentication.
+    /// </summary>
+    internal interface IPersistentCertificateCache
+    {
+        /// <summary>
+        /// Reads the newest valid (≥24h remaining, has private key) entry for the alias.
+        /// Returns <c>true</c> on cache hit, <c>false</c> otherwise.
+        /// </summary>
+        bool Read(string alias, out CertificateCacheValue value, ILoggerAdapter logger = null);
```
bgavrilMS: logger should not be null. you are calling this class, ensure you don't call it with null logger

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
```diff
@@ -0,0 +1,25 @@
...
```
bgavrilMS, line 14: Can we get an extra escape hatch here. Add an env variable that forces the use of NoOpPersistentCertificateCache

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
```diff
@@ -0,0 +1,341 @@
...
+    ///   <item>
+    ///     <description>Canonical <c>client_id</c> is the GUID in the certificate <c>CN</c>.</description>
+    ///   </item>
+    /// </list>
+    /// <para><strong>Notes:</strong> operations are non-throwing and must not block token acquisition.
+    /// FriendlyName tagging semantics are Windows-only.</para>
+    /// </remarks>
+    internal sealed class WindowsPersistentCertificateCache : IPersistentCertificateCache
+    {
+        private static bool PlatformSupportsPersistentFriendlyName(ILoggerAdapter logger)
```
bgavrilMS: This is somewhat overly defensive. THe factory is reponsible for implementing this.

gladjohn: removed

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
gladjohn: added

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IPersistentCertificateCache.cs
gladjohn: udpated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
```diff
@@ -1,25 +1,31 @@
...
-            if (DesktopOsHelper.IsWindows())
+            var disable = Environment.GetEnvironmentVariable(DisableEnvVar);
+            if (!string.IsNullOrEmpty(disable) &&
+                (disable.Equals("1", StringComparison.OrdinalIgnoreCase) ||
+                 disable.Equals("true", StringComparison.OrdinalIgnoreCase)))
             {
-                logger?.Info(() =>
-                    "[PersistentCert] Windows detected; enabling persistent mTLS binding certificate cache (CurrentUser/My).");
-                return new WindowsPersistentCertificateCache();
+                logger?.Info(() => "[PersistentCert] No-op persistent cache enabled via " + DisableEnvVar + ".");
```
Robbie-Microsoft: Do you need "?" in "logger?.Info"? It doesn't appear that logger can be null.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,143 @@
...
+                    return false;
+                }
+            }
+
+            if (TryScope(globalName, out var unauthorizedGlobal))
+            {
+                return true;
+            }
+
+            if (unauthorizedGlobal)
```
Robbie-Microsoft, line 113: Is this if statement needed? TryScope is returning a bool.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
```diff
@@ -0,0 +1,143 @@
...
+        public static bool TryWithAliasLock(
+            string alias,
+            TimeSpan timeout,
+            Action action,
+            Action<string> logVerbose = null)
+        {
+            var globalName = GetMutexNameForAlias(alias, preferGlobal: true);
+            var localName = GetMutexNameForAlias(alias, preferGlobal: false);
+
+            bool TryScope(string name, out bool unauthorized)
```
Robbie-Microsoft, line 44: nit: add xml docs to this function so we can know what it does without looking at the rest of its code

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+    internal static class PersistentCertificateCacheFactory
+    {
+        private const string DisableEnvVar = "MSAL_MI_DISABLE_PERSISTENT_CERT_CACHE";
+
+        public static IPersistentCertificateCache Create(ILoggerAdapter logger)
+        {
+            var disable = Environment.GetEnvironmentVariable(DisableEnvVar);
+            if (!string.IsNullOrEmpty(disable) &&
+                (disable.Equals("1", StringComparison.OrdinalIgnoreCase) ||
+                 disable.Equals("true", StringComparison.OrdinalIgnoreCase)))
```
Robbie-Microsoft, line 19: should you include "TRUE"?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
```diff
@@ -0,0 +1,322 @@
...
+    ///   <item>
+    ///     <description>Canonical <c>client_id</c> is the GUID in the certificate <c>CN</c>.</description>
+    ///   </item>
+    /// </list>
+    /// <para><strong>Notes:</strong> operations are non-throwing and must not block token acquisition.
+    /// FriendlyName tagging semantics are Windows-only.</para>
+    /// </remarks>
+    internal sealed class WindowsPersistentCertificateCache : IPersistentCertificateCache
+    {
+        public bool Read(string alias, out CertificateCacheValue value, ILoggerAdapter logger = null)
```
Robbie-Microsoft: I saw Bogdan comment this elsewhere, should logger be able to null here? Should it always be passed in?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
gladjohn: Good question! We don’t need to explicitly include "TRUE" because StringComparison.OrdinalIgnoreCase already makes the comparison case-insensitive. So "true", "TRUE", or "True" will all match without extra logic. The intent is to support both "1" and any case variation of "true" as valid flags.

We do have test coverage with these cases.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
gladjohn: good catch - updated

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: added

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/InterprocessLock.cs
gladjohn: Good call to double‑check. TryScope’s bool answers “did we run?”, but we still need to know why the Global attempt failed.

We only try Local\ when Global is unauthorized; if Global was just busy or the action failed, we return false without falling back.

I’ll keep the if (unauthorizedGlobal) and add a small cleanup/comment to make the intent explicit.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
gladjohn: fixed all occurrences.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/PersistentCertificateCacheFactory.cs
Robbie-Microsoft: oh, oops I missed that

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
```diff
@@ -0,0 +1,322 @@
...
+                            }
+                            catch (Exception exSet)
+                            {
+                                logger.Verbose(() => "[PersistentCert] Could not set FriendlyName; skipping persist. " + exSet.Message);
+                                return;
+                            }
+
+                            // Add the original instance (carries private key if present)
+                            store.Add(cert);
+                            logger?.Info(() => "[PersistentCert] Persisted certificate to CurrentUser/My.");
```
Robbie-Microsoft: remove "?"

Robbie-Microsoft (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
gladjohn: removed

Robbie-Microsoft (approved):

PR #5376, opened 2025-07-09
Fix for #5375 - Add a FormatResultAsync  to IAuthenticationOperation

bgavrilMS:
Fixes #5375 

Create an `IAuthenticationOperation2` interface with the new method.

PR #5567, opened 2025-11-03
Enable IAuthenticationOperation2 to reject MSAL cached tokens and fetch new ones from ESTS

trwalke:
Fixes #

Adding `Task<bool> ValidateCachedTokenAsync(MsalCacheValidationData)` to `IAuthenticationOperation2`

This method will allow the implementation of IAuthenticationOperation2 (extensions) to validate if the cached token is still valid. If the token is not valid, this will tell MSAL to not use the token and to fetch a new one from ESTS. If the token is valid, MSAL will proceed as normal.

`MsalCacheValidationData` provides the data to the extension to enable it to determine if the token is valid.

**Testing**
Unit tests

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationOperation2.cs
+++ b/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationOperation2.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Identity.Client.AuthScheme
         /// <summary>
         /// Determines whether the cached token is still valid.
         /// </summary>
-        /// <param name="cachedTokenData">Data used to determine if token is valid</param>
+        /// <param name="cachedTokenData">Data used to determine if token is still valid</param>
         /// <returns></returns>
         Task<bool> ValidateCachedTokenAsync(MsalCacheValidationData cachedTokenData);
     }


bgavrilMS:
Not sure what the purpose for this is, but you cannot add a method to an itnerface. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5376

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -76,6 +77,19 @@ protected override async Task<AuthenticationResult> ExecuteAsync(CancellationTok
 
             MsalAccessTokenCacheItem cachedAccessTokenItem = await GetCachedAccessTokenAsync().ConfigureAwait(false);
 
+            // Validate the cached token using the authentication operation
```
bgavrilMS, line 80: Should be on request base?

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -161,4 +162,8 @@
     <AdditionalFiles Include="PublicAPI/$(TargetFramework)/PublicAPI.Unshipped.txt" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="AuthScheme\MsalCacheValidationData.cs" />
```
bgavrilMS: revert all changes

src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationOperation.cs
```diff
@@ -43,6 +43,13 @@ public interface IAuthenticationOperation
         /// <returns>Name and values of params</returns>
         IReadOnlyDictionary<string, string> GetTokenRequestParams();
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="cachedTokenItem"></param>
+        /// <returns></returns>
+        bool ValidateCachedToken(MsalCacheValidationData cachedTokenItem);
```
bgavrilMS: naming - how about `AcceptCachedToken` ?

bgavrilMS: And maybe make it async from the start? :)

trwalke: Updated

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
trwalke: It cant work there because each request child object implements its own token retrieval from the cache. Adding it to ClientCredential for now

src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationOperation2.cs
```diff
@@ -17,5 +17,12 @@ public interface IAuthenticationOperation2 : IAuthenticationOperation
         /// Will be invoked instead of IAuthenticationOperation.FormatResult
         /// </summary>
         Task FormatResultAsync(AuthenticationResult authenticationResult, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Determines whether the cached token is still valid.
```
bgavrilMS, line 22:
```suggestion
        /// Determines whether the cached token is still valid. Only works for client_credentials flow
```

src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationOperation2.cs
```diff
@@ -17,5 +17,12 @@ public interface IAuthenticationOperation2 : IAuthenticationOperation
...
+        /// </summary>
+        /// <param name="cachedTokenData">Data used to determine if token is valid</param>
```
gladjohn:
```suggestion
        /// <param name="cachedTokenData">Data used to determine if token is still valid</param>
```

gladjohn (approved):

PR #5569, opened 2025-11-05
Attempt to fix flaky GetRemovesExpired_Async test

Avery-Dunn:
See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5357

`GetRemovesExpired_Async` in `ThrottlingCacheTests` is flaky in our pipelines, occasionally failing due to the "k1" cache entry not being expired/removed like we expect it to. The `TestCleanup_Async` test in that class does the same expiry checks, except it isn't flaky.

My suspicion is a race condition caused by the task delay: `TestCleanup_Async` used a 50ms delay, however `GetRemovesExpired_Async` used a 1ms delay which is the same as the expiry time. A lack of precision in the timing could be causing us to check the cache before the entry is actually expired.

This PR aims to fix that by removing the delay logic entirely, and simply sets the cache entries to already be expired when they're created.

### DIFF
--- a/tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsTests.cs
+++ b/tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsTests.cs
@@ -39,10 +39,10 @@ namespace Microsoft.Identity.Test.E2E
         [TestCategory("MI_E2E_Imds")]
         [DataTestMethod]
         [DataRow(null /*SAMI*/, null, DisplayName = "AcquireToken_OnImds_Succeeds-SAMI")]
-        [DataRow("4b7a4b0b-ecb2-409e-879a-1e21a15ddaf6", "clientid", DisplayName = "AcquireToken_OnImds_Succeeds-UAMI-ClientId")]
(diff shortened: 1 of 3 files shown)


tests/Microsoft.Identity.Test.Unit/Throttling/ThrottlingCacheTests.cs
```diff
@@ -18,16 +18,15 @@ public class ThrottlingCacheTests
...
 
         [TestMethod]
-        public async Task GetRemovesExpired_Async()
+        public void GetRemovesExpired()
         {
             // Arrange
             ThrottlingCache cache = new ThrottlingCache();
             
-            cache.AddAndCleanup("k1", new ThrottlingCacheEntry(_ex1, TimeSpan.FromMilliseconds(1)), _logger); // expired
+            cache.AddAndCleanup("k1", new ThrottlingCacheEntry(_ex1, TimeSpan.FromMilliseconds(-10000)), _logger); // expired
```
gladjohn, line 26:
```suggestion
            // Use negative TTL to guarantee expiration without timing flakiness (no Task.Delay)
            cache.AddAndCleanup("k1", new ThrottlingCacheEntry(_ex1, TimeSpan.FromMilliseconds(-10000)), _logger); // expired
```

PR #5573, opened 2025-11-07
Add extensibility APIs

### LINKED ISSUE #5568: [Feature Request] Add 3 new extensiblity APIs
### MSAL client type

Confidential

### Problem statement

`ConfidentialClientApplicationBuilder.WithCertificate(IAppConfig => X509Certificate)`
 This would provide a callback to obtain the certificate based on clientid and authority/tenantId. This function would need to be evaluated each time before ExecuteAsync is called.
`ConfidentialClientApplicationBuilder.WithRetry((IAppConfig, MsalException) => bool` This would be a retry policy for ExecuteAsync. On failure to obtain a token, a bool is returned saying whether retry should be attempted. If false, exception is rethrown. If true, then a try will be done. This will be repeated until the function returns false.
`ConfidentialClientApplicationBuilder.WithObserver((IAppConfig, OneOf<AuthenticationResult, MsalException>) => void` Callback for the result of ExecuteAsync. Would return either the AuthenticationResult, or the exception. Could also be `WithOnSuccessCallback(IAppConfig)`, we don't need the failure or even the AuthenticationResult object.
 
Usage

```csharp
    public static ConfidentialClientApplicationBuilder WithManagedCertificate(
        this ConfidentialClientApplicationBuilder builder,
        ICertificateSelectionService selection)
    {
        // Obtains a IManagedCertificate instace via the info in AppConfig
        IManagedCertificate GetCert(IAppConfig appConfig) =>
            selection.GetCertificateByFmiId(new GetByFmiIdOptions(appConfig.ClientId, appConfig.TenantId));

        return builder

            // Sets the certificate to use for client credential flows.
            .WithCertificate(appConfig => GetCert(appConfig).ActiveCertificate)

            // MarkFailure sends telemetry about the failure and returns a bool if the ActiveCertificate has changed.
            // appConfig.ClientCredentialCertificate must be set correctly as MarkFailure uses it for telemetry.
            .WithRetry((appConfig, ex) => GetCert(appConfig).MarkFailure(appConfig.ClientCredentialCertificate, ex))

            // MarkSuccess sends telemetry about the success.
            // We do not need to send telemetry about failures as WithRetry handles that.
            .WithObserver((appConfig, res) =>
            {
                if (res.Successful)
                    GetCert(appConfig).MarkSuccess(appConfig.ClientCredentialCertificate);
            });
    }
```

### Proposed solution

_No response_

### Alternatives

_No response_

neha-bhargava:
Fixes #5568 

* Create the extensibility apis for dynamic cert, retry and observer

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
@@ -241,7 +241,7 @@ namespace Microsoft.Identity.Client.ManagedIdentity
             return queryParams;
         }
 
-        public static async Task<bool> ProbeImdsEndpointAsync(
+        public static async Task<(bool success, string failureReason)> ProbeImdsEndpointAsync(
             RequestContext requestContext,
             ImdsVersion imdsVersion,
             CancellationToken cancellationToken)
@@ -255,8 +255,8 @@ namespace Microsoft.Identity.Client.ManagedIdentity
             {
                 case ImdsVersion.V2:
 #if NET462
-                requestContext.Logger.Info("[Managed Identity] IMDSv2 flow is not supported on .NET Framework 4.6.2. Cryptographic operations required for managed identity authentication are unavailable on this platform. Skipping IMDSv2 probe.");
-                return false;
+                    requestContext.Logger.Info("[Managed Identity] IMDSv2 flow is not supported on .NET Framework 4.6.2. Cryptographic operations required for managed identity authentication are unavailable on this platform. Skipping IMDSv2 probe.");
+                    return (false, "IMDSv2 is not supported on .NET Framework 4.6.2");
 #else
                     apiVersionQueryParam = ImdsV2ManagedIdentitySource.ApiVersionQueryParam;
                     imdsApiVersion = ImdsV2ManagedIdentitySource.ImdsV2ApiVersion;
@@ -303,22 +303,24 @@ namespace Microsoft.Identity.Client.ManagedIdentity
                     retryPolicy: retryPolicy)
                 .ConfigureAwait(false);
             }
-            catch (Exception ex)
+            catch (Exception ex) when (ex is not OperationCanceledException)
             {
-                requestContext.Logger.Info($"[Managed Identity] {imdsStringHelper} probe endpoint failure. Exception occurred while sending request to probe endpoint: {ex}");
-                return false;
+                string failureMessage = $"{imdsStringHelper} probe failed. Exception: {ex.Message}";
+                requestContext.Logger.Info(() => $"[Managed Identity] {failureMessage}");
+                return (false, failureMessage);
             }
 
             // probe omits the "Metadata: true" header and then treats 400 Bad Request as success
             if (response.StatusCode == HttpStatusCode.BadRequest)
             {
                 requestContext.Logger.Info(() => $"[Managed Identity] {imdsStringHelper} managed identity is available.");
-                return true;
+                return (true, null);
             }
             else
             {
+                string failureMessage = $"{imdsStringHelper} probe failed. Status code: {response.StatusCode}, Body: {response.Body}";
                 requestContext.Logger.Info(() => $"[Managed Identity] {imdsStringHelper} managed identity is not available. Status code: {response.StatusCode}, Body: {response.Body}");
-                return false;
+                return (false, failureMessage);
             }
         }
     }
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
@@ -54,13 +54,15 @@ namespace Microsoft.Identity.Client.ManagedIdentity
             {
                 requestContext.Logger.Info($"[Managed Identity] Selecting managed identity source if not cached. Cached value is {s_sourceName} ");
 
+                ManagedIdentitySourceResult sourceResult = null;
                 ManagedIdentitySource source;
 
                 // If the source is not already set, determine it
                 if (s_sourceName == ManagedIdentitySource.None)
                 {
                     // First invocation: detect and cache
-                    source = await GetManagedIdentitySourceAsync(requestContext, isMtlsPopRequested, cancellationToken).ConfigureAwait(false);
+                    sourceResult = await GetManagedIdentitySourceAsync(requestContext, isMtlsPopRequested, cancellationToken).ConfigureAwait(false);
+                    source = sourceResult.Source;
                 }
                 else
                 {
@@ -96,14 +98,14 @@ namespace Microsoft.Identity.Client.ManagedIdentity
                     ManagedIdentitySource.AzureArc => AzureArcManagedIdentitySource.Create(requestContext),
                     ManagedIdentitySource.ImdsV2 => ImdsV2ManagedIdentitySource.Create(requestContext),
                     ManagedIdentitySource.Imds => ImdsManagedIdentitySource.Create(requestContext),
-                    _ => throw new MsalClientException(MsalError.ManagedIdentityAllSourcesUnavailable, MsalErrorMessage.ManagedIdentityAllSourcesUnavailable)
+                    _ => throw CreateManagedIdentityUnavailableException(sourceResult)
                 };
             }
         }
 
         // Detect managed identity source based on the availability of environment variables and csr metadata probe request.
         // This method is perf sensitive any changes should be benchmarked.
-        internal async Task<ManagedIdentitySource> GetManagedIdentitySourceAsync(
+        internal async Task<ManagedIdentitySourceResult> GetManagedIdentitySourceAsync(
             RequestContext requestContext,
             bool isMtlsPopRequested,
             CancellationToken cancellationToken)
@@ -113,36 +115,46 @@ namespace Microsoft.Identity.Client.ManagedIdentity
             if (source != ManagedIdentitySource.None)
             {
                 s_sourceName = source;
-                return source;
+                return new ManagedIdentitySourceResult(source);
             }
+
+            string imdsV2FailureReason = null;
+            string imdsV1FailureReason = null;
 
             // skip the ImdsV2 probe if MtlsPop was NOT requested
             if (isMtlsPopRequested)
             {
-                var imdsV2Response = await ImdsManagedIdentitySource.ProbeImdsEndpointAsync(requestContext, ImdsVersion.V2, cancellationToken).ConfigureAwait(false);
-                if (imdsV2Response)
+                var (imdsV2Success, imdsV2Failure) = await ImdsManagedIdentitySource.ProbeImdsEndpointAsync(requestContext, ImdsVersion.V2, cancellationToken).ConfigureAwait(false);
+                if (imdsV2Success)
                 {
                     requestContext.Logger.Info("[Managed Identity] ImdsV2 detected.");
                     s_sourceName = ManagedIdentitySource.ImdsV2;
-                    return s_sourceName;
+                    return new ManagedIdentitySourceResult(s_sourceName);
                 }
+                imdsV2FailureReason = imdsV2Failure;
             }
             else
             {
                 requestContext.Logger.Info("[Managed Identity] Mtls Pop was not requested; skipping ImdsV2 probe.");
             }
 
-            var imdsV1Response = await ImdsManagedIdentitySource.ProbeImdsEndpointAsync(requestContext, ImdsVersion.V1, cancellationToken).ConfigureAwait(false);
-            if (imdsV1Response)
+            var (imdsV1Success, imdsV1Failure) = await ImdsManagedIdentitySource.ProbeImdsEndpointAsync(requestContext, ImdsVersion.V1, cancellationToken).ConfigureAwait(false);
+            if (imdsV1Success)
             {
                 requestContext.Logger.Info("[Managed Identity] ImdsV1 detected.");
                 s_sourceName = ManagedIdentitySource.Imds;
-                return s_sourceName;
+                return new ManagedIdentitySourceResult(s_sourceName);
             }
+            imdsV1FailureReason = imdsV1Failure;
 
             requestContext.Logger.Info($"[Managed Identity] {MsalErrorMessage.ManagedIdentityAllSourcesUnavailable}");
             s_sourceName = ManagedIdentitySource.None;
-            return s_sourceName;
+            
+            return new ManagedIdentitySourceResult(s_sourceName)
+            {
+                ImdsV1FailureReason = imdsV1FailureReason,
+                ImdsV2FailureReason = imdsV2FailureReason
+            };
         }
 
         /// <summary>
@@ -230,12 +242,39 @@ namespace Microsoft.Identity.Client.ManagedIdentity
         }
 
         /// <summary>
+        /// Creates an MsalClientException for when no managed identity source is available,
+        /// including detailed failure information from IMDS probes if available.
+        /// </summary>
+        private static MsalClientException CreateManagedIdentityUnavailableException(ManagedIdentitySourceResult sourceResult)
+        {
+            string errorMessage = MsalErrorMessage.ManagedIdentityAllSourcesUnavailable;
+
+            if (sourceResult != null)
+            {
+                if (!string.IsNullOrEmpty(sourceResult.ImdsV1FailureReason) || !string.IsNullOrEmpty(sourceResult.ImdsV2FailureReason))
+                {
+                    errorMessage += " MSAL was not able to detect the Azure Instance Metadata Service (IMDS) that runs on VMs:";
+                    if (!string.IsNullOrEmpty(sourceResult.ImdsV2FailureReason))
+                    {
+                        errorMessage += $" IMDSv2: {sourceResult.ImdsV2FailureReason}.";
+                    }
+                    if (!string.IsNullOrEmpty(sourceResult.ImdsV1FailureReason))
+                    {
+                        errorMessage += $" IMDSv1: {sourceResult.ImdsV1FailureReason}.";
+                    }
+                }
+            }
+
+            return new MsalClientException(MsalError.ManagedIdentityAllSourcesUnavailable, errorMessage);
+        }
+
+        /// <summary>
         /// Sets (or replaces) the in-memory binding certificate used to prime the mtls_pop scheme on subsequent requests.
         /// The certificate is intentionally NOT disposed here to avoid invalidating caller-held references (e.g., via AuthenticationResult).
         /// </summary>
         /// <remarks>
         /// Lifetime considerations:
-        /// - The binding certificate is ephemeral and valid for the token’s binding duration.
+        /// - The binding certificate is ephemeral and valid for the token's binding duration.
         /// - If rotation occurs, older certificates will be eligible for GC once no longer referenced.
         /// - Explicit disposal can be revisited if a deterministic rotation / shutdown strategy is introduced.
         /// </remarks>
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySourceResult.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySourceResult.cs
@@ -1,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client.ManagedIdentity
+{
+    /// <summary>
+    /// Result of managed identity source detection, including the detected source and any failure information from IMDS probes.
+    /// </summary>
+    /// <remarks>
+    /// This class is returned by <see cref="ManagedIdentityApplication.GetManagedIdentitySourceAsync"/> to provide
+    /// detailed information about managed identity source detection, including failure reasons when IMDS probes fail.
+    /// This information is useful for credential chains like DefaultAzureCredential to determine whether to skip
+    /// managed identity authentication entirely.
+    /// </remarks>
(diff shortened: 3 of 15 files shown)


src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -131,6 +131,25 @@ public string ClientVersion
         internal IRetryPolicyFactory RetryPolicyFactory { get; set; }
         internal ICsrFactory CsrFactory { get; set; }
 
+        #region Extensibility Callbacks
+
+        /// <summary>
+        /// Dynamic certificate provider callback for client credential flows.
+        /// </summary>
+        public Func<IAppConfig, X509Certificate2> ClientCredentialCertificateProvider { get; set; }
```
bgavrilMS: I am not sure about `IAppConfig` here. Interfaces are difficult to evolve. Maybe we should use `AssertioNRequestOptions` ?

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs#L16

Or maybe we can have an "AppConfig" object instead.

bgavrilMS: Let's make all our extensibility points async from the start.

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationBuilderExtensions.cs
```diff
@@ -29,5 +30,103 @@ public static ConfidentialClientApplicationBuilder WithAppTokenProvider(
...
+        /// <remarks>
+        /// <para>The retry policy is only invoked for network failures, not for cached token retrievals.</para>
+        /// <para>When the policy returns <c>true</c>, MSAL will invoke the certificate provider callback again (if configured)
+        /// before making another token request, enabling certificate rotation scenarios.</para>
+        /// <para>MSAL's internal throttling and retry mechanisms will still apply, including respecting Retry-After headers.</para>
+        /// <para>To prevent infinite loops, ensure your retry policy has appropriate termination conditions.</para>
+        /// </remarks>
+        public static ConfidentialClientApplicationBuilder WithRetry(
+            this ConfidentialClientApplicationBuilder builder,
+            Func<IAppConfig, MsalException, bool> retryPolicy)
```
bgavrilMS: Same comment - IAppConfig not the rigth object.

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationBuilderExtensions.cs
```diff
@@ -29,5 +30,103 @@ public static ConfidentialClientApplicationBuilder WithAppTokenProvider(
...
+        /// <returns>The builder to chain additional configuration calls.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="retryPolicy"/> is null.</exception>
+        /// <remarks>
+        /// <para>The retry policy is only invoked for network failures, not for cached token retrievals.</para>
+        /// <para>When the policy returns <c>true</c>, MSAL will invoke the certificate provider callback again (if configured)
+        /// before making another token request, enabling certificate rotation scenarios.</para>
+        /// <para>MSAL's internal throttling and retry mechanisms will still apply, including respecting Retry-After headers.</para>
+        /// <para>To prevent infinite loops, ensure your retry policy has appropriate termination conditions.</para>
+        /// </remarks>
+        public static ConfidentialClientApplicationBuilder WithRetry(
```
bgavrilMS: Wondering if `OnBeforeFailure` and `OnSuccess` are better than `WithRetry` and `WithObserver` ?

bgavrilMS: We should make it clear that this callback is only triggered for exceptions from the sts

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava: Updated to use another object instead.

neha-bhargava: Done

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationBuilderExtensions.cs
neha-bhargava: Updated

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationBuilderExtensions.cs
```diff
@@ -29,5 +30,153 @@ public static ConfidentialClientApplicationBuilder WithAppTokenProvider(
...
+            
+            // Create a CertificateAndClaimsClientCredential with null certificate
+            // The certificate will be resolved dynamically via the provider in ResolveCertificateAsync
+            builder.Config.ClientCredential = new Microsoft.Identity.Client.Internal.ClientCredential.CertificateAndClaimsClientCredential(
+                certificate: null,
+                claimsToSign: null,
+                appendDefaultClaims: true);
+            
+            return builder;
+        }
```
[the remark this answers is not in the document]
neha-bhargava: Not doing the validation in this PR

src/client/Microsoft.Identity.Client/Extensibility/ExecutionResult.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client.Extensibility
+{
+    /// <summary>
+    /// Represents the result of a token acquisition attempt.
+    /// Used by the execution observer configured via <see cref="ConfidentialClientApplicationBuilderExtensions.OnSuccess"/>.
+    /// </summary>
+    public class ExecutionResult
```
tlupes, line 15: Can this include the credential used (primarily care about the exact certificate used). If that information is in AuthenticationResult then that is fine.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -127,14 +122,150 @@ private async Task<AuthenticationResult> GetAccessTokenAsync(
...
+                        {
+                            logger.Info("[ClientCredentialRequest] OnMsalServiceFailureCallback returned true. Retrying token request.");
+                            continue; // Retry the loop
+                        }
+                        
+                        logger.Info("[ClientCredentialRequest] OnMsalServiceFailureCallback returned false. Propagating exception.");
+                    }
+                    
+                    // Invoke OnSuccess callback with failure result
+                    await InvokeOnSuccessCallbackAsync(authResult: null, exception: serviceEx, logger).ConfigureAwait(false);
```
tlupes: Seems OnSuccess is a misnomer if it is called both for failures and success; perhaps it should be renamed?

tests/Microsoft.Identity.Test.Unit/AppConfigTests/ConfidentialClientApplicationExtensibilityApiTests.cs
```diff
@@ -0,0 +1,357 @@
...
+                    .Create(TestConstants.ClientId)
+                    .WithExperimentalFeatures()
+                    .WithCertificate((Func<ClientCredentialExtensionParameters, Task<X509Certificate2>>)null)
+                    .Build());
+
+            Assert.AreEqual("certificateProvider", ex.ParamName);
+        }
+
+        [TestMethod]
+        public void WithCertificate_AllowsMultipleCallbackRegistrations_LastOneWins()
```
tlupes, line 74: Is this how the rest of the APIs currently work? If you call WithCertificate multiple times, it just overwrites?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
neha-bhargava: Agree, I feel this can be renamed to `OnCompletion`. Since `OnServiceFailure` only handles certain service exceptions this can be triggered when either the request succeeds or if there are any other failures that are not being used to retry. @bgavrilMS thoughts on this?

tests/Microsoft.Identity.Test.Unit/AppConfigTests/ConfidentialClientApplicationExtensibilityApiTests.cs
neha-bhargava: Right now that is how it is. The test case was generated by copilot. I am going to propose another PR that will validate and throw if the cert is being set twice. Or if the dynamic cert is used then avoid static cert.

src/client/Microsoft.Identity.Client/Extensibility/ClientCredentialExtensionParameters.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace Microsoft.Identity.Client.Extensibility
+{
+    /// <summary>
+    /// Provides application configuration context to client credential extensibility callbacks.
+    /// Contains read-only information about the confidential client application.
+    /// </summary>
+#if !SUPPORTS_CONFIDENTIAL_CLIENT
+    [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]  // hide confidential client on mobile
+#endif
+    public class ClientCredentialExtensionParameters
```
bgavrilMS: Can we reuse AssertionRequestOptions ? Some of the properties don't match exactly, but it's ok if they are not set.

neha-bhargava: Updated

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -127,14 +122,148 @@ private async Task<AuthenticationResult> GetAccessTokenAsync(
...
+                    await InvokeOnSuccessCallbackAsync(authResult, exception: null, logger).ConfigureAwait(false);
+                    
+                    return authResult;
+                }
+                catch (MsalServiceException serviceEx)
+                {
+                    // Check if OnMsalServiceFailureCallback is configured
+                    if (AuthenticationRequestParameters.AppConfig.OnMsalServiceFailureCallback != null)
+                    {
+                        logger.Info("[ClientCredentialRequest] MsalServiceException caught. Invoking OnMsalServiceFailureCallback.");
```
neha-bhargava: Log the number of retries

src/client/Microsoft.Identity.Client/ApiConfig/AbstractConfidentialClientAcquireTokenParameterBuilder.cs
```diff
@@ -48,7 +48,8 @@ protected override void Validate()
             // Confidential client must have a credential
             if (ServiceBundle?.Config.ClientCredential == null &&
```
bgavrilMS, line 49: Can the "ClientCredentialCertificateProvider" be abstracted into the existing "ClientCredential" ?

bgavrilMS: In other words, you can always assume a ClientCredentialCertificateProvider is used. 

WithCertificate((_) => X509Certificate) 
Provider = delegate created by app develper

WithCertificate(myCertificate)
Provider = simple delegate that returns myCertificate

src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs
```diff
@@ -6,15 +6,34 @@
...
+        }
+
+        /// <summary>
+        /// Internal constructor that creates AssertionRequestOptions from ApplicationConfiguration
+        /// </summary>
+        /// <param name="appConfig">The application configuration</param>
+        internal AssertionRequestOptions(ApplicationConfiguration appConfig)
+        {
+            ClientID = appConfig.ClientId;
+            TenantId = appConfig.Authority?.TenantId;
```
bgavrilMS: Does this work if the app developer uses `WithTenantId` at the request level? I believe the TokenEndpoint is correctly set in that case.

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -131,6 +131,26 @@ public string ClientVersion
...
+        /// <summary>
+        /// Dynamic certificate provider callback for client credential flows.
+        /// </summary>
+        public Func<AssertionRequestOptions, Task<X509Certificate2>> ClientCredentialCertificateProvider { get; set; }
+
+        /// <summary>
+        /// MSAL service failure callback that determines whether to retry after a token acquisition failure from the identity provider.
+        /// Only invoked for MsalServiceException (errors from the Security Token Service).
+        /// </summary>
+        public Func<AssertionRequestOptions, MsalException, Task<bool>> OnMsalServiceFailureCallback { get; set; }
```
bgavrilMS: naming: I believe the convention for these callbacks is to not use the "callback" suffix.  So maybe "OnBeforeMsalServiceException"

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
```diff
@@ -131,6 +131,26 @@ public string ClientVersion
...
+
+        /// <summary>
+        /// Success callback that receives the result of token acquisition attempts (typically successful, but can include failures after retries are exhausted).
+        /// </summary>
+        public Func<AssertionRequestOptions, ExecutionResult, Task> OnSuccessCallback { get; set; }
```
bgavrilMS: Do we need this ExecutionResult object? It seems to me that OnMsalServiceFailure takes care of the error sceanrio. And so this can be `Func<AssertionRequestOptions, AuthenticationResult> OnTokenAcquisitionSuccess`

src/client/Microsoft.Identity.Client/Internal/ClientCredential/CertificateAndClaimsClientCredential.cs
```diff
@@ -35,10 +32,15 @@ public CertificateAndClaimsClientCredential(
             Certificate = certificate;
             _claimsToSign = claimsToSign;
             _appendDefaultClaims = appendDefaultClaims;
-            _base64EncodedThumbprint = Base64UrlHelpers.Encode(certificate.GetCertHash());
+ 
+            // Certificate can be null when using dynamic certificate provider
+            if (certificate != null)
+            {
+                _base64EncodedThumbprint = Base64UrlHelpers.Encode(certificate.GetCertHash());
```
bgavrilMS: _base64EncodedThumbprint  doesn't seem to in use at all

src/client/Microsoft.Identity.Client/Internal/ClientCredential/CertificateAndClaimsClientCredential.cs
bgavrilMS, line 63: This is fine for Bearer tokens, but it is going to require a bit of work for mTLS POP tokens, which is an important scenario. 

In case of MTLS POP tokens, the certificate is used to talk to Entra via mTLS. 

I recommend you use a follow-up PR to address this sceanrio. There are integration tests for mtls POP.

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
bgavrilMS: Agreed, or use OnSuccess since failure is alreay covered via OnMsalServiceException

neha-bhargava: Only the service exceptions that are handled in the callback will be handled there. Rest can still throw exception back. Will rename it to OnCompletion for clarity.

src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs
neha-bhargava: Are you saying that we should extract the tenant id from token endpoint? Or set the token endpoint instead?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava: Or OnMsalServiceExceptionProvider?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava: No, the OnMsalServiceFailure wll only handle the exceptions defined in the callback. There can be other exceptions as well that can be thrown.

src/client/Microsoft.Identity.Client/Internal/ClientCredential/CertificateAndClaimsClientCredential.cs
neha-bhargava: Dynamic cert for the MTLS POP as well?

src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs
neha-bhargava: Keeping the executionResult as is. Also sending back the cert used

src/client/Microsoft.Identity.Client/Internal/ClientCredential/CertificateAndClaimsClientCredential.cs
neha-bhargava: Discussed offline. This will go in another iteration.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs
```diff
@@ -6,15 +6,37 @@
...
+    /// </summary>
+    /// <remarks> Use the provided information to generate the client assertion payload </remarks>
 #if !SUPPORTS_CONFIDENTIAL_CLIENT
     [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]  // hide confidential client on mobile
 #endif
     public class AssertionRequestOptions {
         /// <summary>
+        /// Default constructor for AssertionRequestOptions
+        /// </summary>
+        public AssertionRequestOptions()
```
trwalke, line 20: is this needed? C# already creates this

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationBuilderExtensions.cs
```diff
@@ -29,5 +31,154 @@ public static ConfidentialClientApplicationBuilder WithAppTokenProvider(
...
+        /// Configures an async callback that is invoked when MSAL receives an error response from the identity provider (Security Token Service).
+        /// The callback determines whether MSAL should retry the token request or propagate the exception.
+        /// This callback is invoked after each service failure and can be called multiple times until it returns <c>false</c> or the request succeeds.
+        /// </summary>
+        /// <param name="builder">The confidential client application builder.</param>
+        /// <param name="onMsalServiceFailure">
+        /// An async callback that determines whether to retry after a service failure.
+        /// Receives the assertion request options and the <see cref="MsalServiceException"/> that occurred.
+        /// Returns <c>true</c> to retry the request, or <c>false</c> to stop retrying and propagate the exception.
+        /// The callback will be invoked repeatedly after each service failure until it returns <c>false</c> or the request succeeds.
```
trwalke, line 86: Is it possible for customers to get into an infinite loop? Should we warn them about this?

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
```diff
@@ -143,13 +138,72 @@ private async Task<AuthenticationResult> GetAccessTokenAsync(
         {
             await ResolveAuthorityAsync().ConfigureAwait(false);
 
-            // Get a token from AAD
-            if (ServiceBundle.Config.AppTokenProvider == null)
+            AuthenticationResult authResult = null;
+            int retryCount = 0;
+
+            // Retry loop using the retry callback if configured
+            while (true)
```
trwalke, line 145: Should there be a hard limit on retries? just to be sure we dont get into an infinite loop?

trwalke (approved):
just a few minor questions

src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs
neha-bhargava: Since I added a constructor, a default one is needed explicitly. If there is no constructor, the default one is not needed in C#

src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationBuilderExtensions.cs
neha-bhargava: Yes, and they are instructed to handle it. This is an extensibility API to provision some feature internally

src/client/Microsoft.Identity.Client/Internal/Requests/ClientCredentialRequest.cs
neha-bhargava: No, we do not want to control the number of retries for the failures handled in this callback in MSAL.

PR #5581, opened 2025-11-15
ImdsV2 is detected, even on a UAMI configuration error

### LINKED ISSUE #5565: [Bug] IMDS v2 fails to detect source correctly when identity not found
### Library version used

msal mTLS pop preview

### Issue Description:

During Managed Identity source detection, MSAL sends the initial platform-metadata GET (cred-api-version=2.0) with a client_id selector for UAMI. If that UAMI isn’t assigned to the VM, IMDS returns “Identity not assigned to VM.” MSAL misclassifies this as “IMDSv2 not available,” exits the IMDSv2 path, and throws MsalClientException.

### Expected behavior

should detect MSI v2 source when available. 

### Identity provider

### Regression

_No response_

### Solution and workarounds

Fix: Reclassify this as a UAMI configuration error rather than endpoint unavailability.

Robbie-Microsoft:
Fixes #5565

### DIFF
--- a/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
@@ -94,7 +94,7 @@ namespace Microsoft.Identity.Client.ManagedIdentity.V2
                 {
                     if (IsUamiConfigurationError(response, requestContext.ServiceBundle.Config.ManagedIdentityId))
                     {
-                        requestContext.Logger.Info("[Managed Identity] IMDSv2 endpoint is available but UAMI configuration error detected during probe. Returning empty CsrMetadata to indicate endpoint availability.");
+                        requestContext.Logger.Info("[Managed Identity] IMDSv2 endpoint is available but there is a UAMI configuration error. Returning empty CsrMetadata to indicate endpoint availability.");
                         return new CsrMetadata();
                     }
 
--- a/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
+++ b/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System.Drawing;
 using System.Globalization;
 
 namespace Microsoft.Identity.Client
--- a/src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
+++ b/src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
@@ -55,14 +55,6 @@ namespace Microsoft.Identity.Client
                 resource);
         }
 
-        // do probe for imdsv1
-
-        // currently sends without metadata header, if available will respond with 400
-
-        // figure out if there's a scenario where v1 should be chosen over v2. ask raghavendra
-
-        // make probe configurable for v1 and v2
-
         /// <inheritdoc/>
         public async Task<ManagedIdentitySource> GetManagedIdentitySourceAsync()
         {


src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -91,15 +92,31 @@ public static async Task<CsrMetadata> GetCsrMetadataAsync(
             {
                 if (probeMode)
```
gladjohn, line 93: IMDS throttles very aggressively on **200s**, but is much more forgiving of **400s**.  
For that reason we want the **IMDSv2 probe** to have a single responsibility:

> **“Does the endpoint exist and respond?”**

and nothing else.

---

Right now `GetPlatformMetadata` ends up hitting IMDSv2 **twice**, and the probe path behaves too much like a “real” call:

- It can return **200**.
- It runs identity-specific logic (**UAMI** / `IdentityNotFound`) that we only need on the second call.

---

We’d like to align the probe with the way **Azure.Identity** does it in [`ImdsManagedIdentityProbeSource`](https://github.com/Azure/azure-sdk-for-net/blob/1bf9117302a89005a7e549182992184625741cdc/sdk/identity/Azure.Identity/src/ImdsManagedIdentityProbeSource.cs#L13):

- The probe sends a **GET** to the IMDS endpoint **without** the `Metadata: true` header.
- IMDS responds with **400** very quickly.
- A **400** of that specific shape is treated as **“probe succeeded, endpoint exists”**  
  (they throw an internal `ProbeRequestResponseException` and then do the real call through `MsalManagedIdentityClient`).
- Only the **second call** is a full, valid IMDS request that can return **200** and surface identity errors.

---

## Concretely, for MSAL

### **Probe call**
- Build the request so IMDS returns a **predictable 400** (e.g., omit the `Metadata` header).
- Treat that **400** as **“IMDSv2 endpoint is available.”**
- Do **not** rely on **200** responses or run UAMI / `IdentityNotFound` logic here — the probe is **not** an identity call.

### **Second (real) call**
- Send the fully valid IMDSv2 request (with `Metadata: true` + identity hints).
- This is the **only call** that should be able to return **200**.
- This is also the **only place** where we classify **UAMI vs system-assigned** and map  
  `IdentityNotFound` vs `ManagedIdentityRequestFailed`.

---

That keeps the probe **cheap and stable** (one responsibility: endpoint exists)  
and ensures we’re not burning extra **200s** against IMDS just to detect IMDSv2.

IMDS V2 probe spec - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/docs/msi_v2/vm_vmss_credential_probe.md

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
```diff
@@ -450,5 +450,7 @@ public static string InvalidTokenProviderResponseValue(string invalidValueName)
         public const string InvalidCertificate = "The certificate received from the Imds server is invalid.";
         public const string CannotSwitchBetweenImdsVersionsForPreview = "ImdsV2 is currently experimental - A Bearer token has already been received; Please restart the application to receive a mTLS PoP token.";
         public const string MtlsPopTokenNotSupportedinImdsV1 = "A mTLS PoP token cannot be requested because the application\'s source was determined to be ImdsV1.";
+        public const string IdentityNotFound = "Managed Identity not found. To request credentials for this identity, please assign it first. Please see aka.ms/ManagedIdentityNotFound for more details";
```
gladjohn, line 453: IMDS already returns the `IdentityNotFound` error message, and MSAL simply bubbles up the IMDS payload without rewriting or overriding it. Adding our own string introduces duplication and may cause divergence from IMDS' authoritative error.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
Robbie-Microsoft: I think this should be tracked on a separate GitHub issue. This PR fixes the linked bug, that you were pushing me to fix, so that people using public preview are no longer blocked.

Do you remember our discussions where we decided to combine the probe and the metadata endpoint? We had that discussion very early on.

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
Robbie-Microsoft: This is the exact error message from Imds - It's used multiple times in this PR, including test mocks, so I made it a constant.

I could return the error message from Imds, but I would still keep this constant for the test mocks.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
bgavrilMS: If there are partners blocked by this, agreed to take this fix, but I am not sure there are?

 Otherwise, it does seem like the 400 error is the correct fix and this can be done alongisde MSIv1 probing

src/client/Microsoft.Identity.Client/MsalErrorMessage.cs
bgavrilMS: Agreed with @gladjohn - we can depend on the error code, but the error messages should from the server.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: While onboarding the partner we found they are using UAMI, and UAMI itself is not working in the environment. So no new partner onboarding till end of month.

bgavrilMS:
Are we abandoning this @Robbie-Microsoft ?

Robbie-Microsoft:
> Are we abandoning this @Robbie-Microsoft ?

Yes, that's what you and Gladwin told me.

PR #5587, opened 2025-11-18
Add draft spec for WithCertificate API

gladjohn:
Introduced a draft specification for the `WithCertificate` API, detailing goals, new types, overloads, and example usage for Bearer and mTLS authentication.

### DIFF
--- a/docs/With_Certificate_Design.md
+++ b/docs/With_Certificate_Design.md
@@ -1,0 +1,188 @@
+# Draft Spec: `WithCertificate` API – Bearer, mTLS, Dynamic Selection
+
+## 1. Goals
+
+- Keep existing `WithCertificate(X509Certificate2, bool sendX5C)` behavior **unchanged**.
+- Support:
+  - Standard **Bearer** client‑credential tokens.
+  - **mTLS Bearer** (certificate used only for TLS binding).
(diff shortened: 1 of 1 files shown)


docs/With_Certificate_Design.md
```diff
@@ -0,0 +1,188 @@
...
+  - **mTLS Bearer** (certificate used only for TLS binding).
+  - **mTLS PoP** (existing `.WithMtlsProofOfPossession()` flow).
+- Add a **delegate‑based** `WithCertificate` that can select a certificate **per request** using `IAppConfig`.
+
+## 2. New types
+
+### 2.1 `ClientCertificateUsage`
+
+```csharp
+public enum ClientCertificateUsage
```
bgavrilMS, line 17: THere is no need for this? It should be derived from the context?

gladjohn: Are you suggesting we add a a request‑level WithMutualTls() to opt into Bearer over mTLS?

bgavrilMS: Yes, WithMTLSPOP() API already exists. We should not have 2 APIs for achiving the same thing.

bgavrilMS:
@neha-bhargava will create a new API for `WithCertificate` to consolidate all the options.

Issue #5594, opened 2025-11-20
[Feature Request] IMDS Source Detection Logic Improvement


Robbie-Microsoft:
### MSAL client type

Managed identity

### Problem statement

# Current functionality
`GetManagedIdentitySourceAsync` will check for and return managed identity sources in the following order:
1. non-Imds sources (via env vars)
2. ImdsV2 (via probe)
Note: Presence of ImdsV2 implies ImdsV1 also exists in the environment.
(ImdsV2 will never exist without ImdsV1, but ImdsV1 can exist without ImdsV2 in some production environments.) 
3. return `DefaultToImds`
Azure SDK will then probe ImdsV1 to see if it's available.

# Proposed functionality
1. non-Imds sources (via env vars)
2. ImdsV2 (via probe)
3. ImdsV1 (via probe)
4. return `None`

### Detailed Work Items

- [ ] Change business logic to probe ImdsV1, before returning `None` if probe returns 400
- [ ] Change business logic to deprecate `DefaultToImds`

christothes:
>  Change business logic to probe ImdsV1, before returning None if probe returns 404

This should be - ... before returning `None` if probe returns 400 (when sending the probe request without the `Metadata` header.

Robbie-Microsoft:
@christothes We'll be adding a mandatory cancellation token param to the public API - used in the IMDS v1 and v2 probes.

```csharp
GetManagedIdentitySourceAsync(CancellationToken cancellationToken)
```

Robbie-Microsoft:
@christothes, we have to revert the change implemented in this feature request.

Before this feature request, we were previously defaulting to IMDS v1 after probing IMDS v2, and letting acquireToken throw an exception when attempting to get a token from IMDSv1.
We were now going to return a new exception after probing IMDS v2 and v1, indicating that no managed identity sources are available.

I've been told this is a breaking change and that it must be reverted. Perhaps we can re-introduce ImdsV1 Probing functionality in the next major version of MSAL.

However, we'll still be introducing the mandatory cancellation token param to the public API.

@bgavrilMS, @gladjohn please confirm here.

PR #5602, opened 2025-11-26
IMDS Source Detection Logic Improvement

Robbie-Microsoft:
Implemented ImdsV2 and ImdsV1 probing

Implements #5594

### DIFF
--- a/LibsAndSamples.sln
+++ b/LibsAndSamples.sln
@@ -194,6 +194,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MacConsoleAppWithBroker", "
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Identity.Client.MtlsPop", "src\client\Microsoft.Identity.Client.MtlsPop\Microsoft.Identity.Client.MtlsPop.csproj", "{3E1C29E5-6E67-D9B2-28DF-649A609937A2}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "WinUI3PackagedSampleApp", "tests\devapps\WinUI3PackagedSampleApp\WinUI3PackagedSampleApp.csproj", "{CE282240-0806-EB91-87E4-D791DC86DEE8}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug + MobileApps|Any CPU = Debug + MobileApps|Any CPU
@@ -1988,6 +1990,48 @@ Global
 		{3E1C29E5-6E67-D9B2-28DF-649A609937A2}.Release|x64.Build.0 = Release|Any CPU
 		{3E1C29E5-6E67-D9B2-28DF-649A609937A2}.Release|x86.ActiveCfg = Release|Any CPU
 		{3E1C29E5-6E67-D9B2-28DF-649A609937A2}.Release|x86.Build.0 = Release|Any CPU
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|Any CPU.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|Any CPU.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|ARM.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|ARM.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|ARM64.ActiveCfg = Debug|ARM64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|ARM64.Build.0 = Debug|ARM64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|iPhone.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|iPhone.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|iPhoneSimulator.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|iPhoneSimulator.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|x64.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|x64.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|x86.ActiveCfg = Debug|x86
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug + MobileApps|x86.Build.0 = Debug|x86
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|Any CPU.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|Any CPU.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|ARM.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|ARM.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|ARM64.ActiveCfg = Debug|ARM64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|ARM64.Build.0 = Debug|ARM64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|iPhone.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|iPhone.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|iPhoneSimulator.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|iPhoneSimulator.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|x64.ActiveCfg = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|x64.Build.0 = Debug|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|x86.ActiveCfg = Debug|x86
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Debug|x86.Build.0 = Debug|x86
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|Any CPU.ActiveCfg = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|Any CPU.Build.0 = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|ARM.ActiveCfg = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|ARM.Build.0 = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|ARM64.ActiveCfg = Release|ARM64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|ARM64.Build.0 = Release|ARM64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|iPhone.ActiveCfg = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|iPhone.Build.0 = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|iPhoneSimulator.ActiveCfg = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|iPhoneSimulator.Build.0 = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|x64.ActiveCfg = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|x64.Build.0 = Release|x64
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|x86.ActiveCfg = Release|x86
+		{CE282240-0806-EB91-87E4-D791DC86DEE8}.Release|x86.Build.0 = Release|x86
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -2046,6 +2090,7 @@ Global
 		{AEF6BB00-931F-4638-955D-24D735625C34} = {34BE693E-3496-45A4-B1D2-D3A0E068EEDB}
 		{DBD18BC8-72E4-47D4-BD79-8DEBD9F2C0D0} = {34BE693E-3496-45A4-B1D2-D3A0E068EEDB}
 		{3E1C29E5-6E67-D9B2-28DF-649A609937A2} = {1A37FD75-94E9-4D6F-953A-0DABBD7B49E9}
+		{CE282240-0806-EB91-87E4-D791DC86DEE8} = {34BE693E-3496-45A4-B1D2-D3A0E068EEDB}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {020399A9-DC27-4B82-9CAA-EF488665AC27}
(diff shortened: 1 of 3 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -211,5 +220,103 @@ public static Uri GetValidatedEndpoint(
...
+                response = await requestContext.ServiceBundle.HttpManager.SendRequestAsync(
+                    GetValidatedEndpoint(requestContext.Logger, imdsEndpoint, queryParams),
+                    headers,
+                    body: null,
+                    method: HttpMethod.Get,
+                    logger: requestContext.Logger,
+                    doNotThrow: false,
+                    mtlsCertificate: null,
+                    validateServerCertificate: null,
+                    cancellationToken: requestContext.UserCancellationToken,
```
bgavrilMS: Should we have a max waiting time here? Does Azure SDK use some sort of timeout after which they give up?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -103,32 +102,45 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
...
         {
             // First check env vars to avoid the probe if possible
             ManagedIdentitySource source = GetManagedIdentitySourceNoImdsV2(requestContext.Logger);
-
-            // If a source is detected via env vars, or
-            // a source wasn't detected (it defaulted to ImdsV1) and MtlsPop was NOT requested,
-            // use the source.
-            // (don't trigger the ImdsV2 probe endpoint if MtlsPop was NOT requested)
-            if (source != ManagedIdentitySource.DefaultToImds || !isMtlsPopRequested)
+#pragma warning disable CS0618 // Type or member is obsolete
```
bgavrilMS: Please add a more meaningful comment on why we need to call this obsolete value.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: Maybe this needs to be renamed like GetManagedIdentitySourceNoImds or something like that to denote that only env variables are used. Method names is better than comments.

bgavrilMS: nit: use proper comments i.e. ///

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -177,7 +189,9 @@ internal static ManagedIdentitySource GetManagedIdentitySourceNoImdsV2(ILoggerAd
             }
             else
             {
+#pragma warning disable CS0618 // Type or member is obsolete
```
bgavrilMS: Another way of handling this would be to return None here? And if Azure SDK calls the old method "GetSource", you return DefaultToImds instead of None. This would make the code in this file somewhat more readable.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
```diff
@@ -48,6 +48,7 @@ public enum ManagedIdentitySource
         /// Indicates that the source is defaulted to IMDS since no environment variables are set.
         /// This is used to detect the managed identity source.
         /// </summary>
+        [Obsolete("Use Imds instead. We are no longer defaulting to IMDS. An error will be thrown instead if no managed identity sources are available.")]
```
bgavrilMS:
```suggestion
        [Obsolete("In use only to support the now obsolete GetSource API. Will be removed in a future version. Use GetSourceAsync instead, which may return None, Imds or ImdsV2")]
```

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: Another way would be to use a nullable value, i.e. ImdsManagedIdentitySource? source. if it's null, it means we haven't initialized it.

bgavrilMS: Up to you, DefaultToImds also works but it's harder to read the code. And you've explained it to me :)

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -211,5 +220,103 @@ public static Uri GetValidatedEndpoint(
...
+                    imdsStringHelper = "IMDSv1";
+                    break;
+
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(imdsVersion), imdsVersion, null);
+            }
+
+            var queryParams = ImdsQueryParamsHelper(requestContext, apiVersionQueryParam, imdsApiVersion);
+
+            var headers = new Dictionary<string, string>
```
gladjohn, line 281: would be good to have some comments here like, 

`probe deliberately omits the Metadata: true header and then treats 400 as IMDS present`

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -211,5 +220,103 @@ public static Uri GetValidatedEndpoint(
...
+                    retryPolicy: retryPolicy)
+                .ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                requestContext.Logger.Info($"[Managed Identity] {imdsStringHelper} probe endpoint failure. Exception occurred while sending request to probe endpoint: {ex}");
+                return false;
+            }
+
+            if (response.StatusCode == HttpStatusCode.BadRequest)
```
gladjohn, line 313: I’d add a short comment near the StatusCode == BadRequest check explaining why 400 is treated as success, just for future reference.

src/client/Microsoft.Identity.Client/Http/Retry/ImdsProbeRetryPolicy.cs
```diff
@@ -5,11 +5,11 @@
 
 namespace Microsoft.Identity.Client.Http.Retry
 {
-    internal class CsrMetadataProbeRetryPolicy : ImdsRetryPolicy
+    internal class ImdsProbeRetryPolicy : ImdsRetryPolicy
```
gladjohn, line 8: consider renaming the file to `ImdsProbeRetryPolicy.cs`

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -103,32 +102,45 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
         // This method is perf sensitive any changes should be benchmarked.
         internal async Task<ManagedIdentitySource> GetManagedIdentitySourceAsync(
             RequestContext requestContext,
-            bool isMtlsPopRequested)
+            bool isMtlsPopRequested,
+            bool noImdsV2 = false)
```
gladjohn: is this used anywhere? `noImdsV2`

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -94,7 +92,8 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
                     ManagedIdentitySource.CloudShell => CloudShellManagedIdentitySource.Create(requestContext),
                     ManagedIdentitySource.AzureArc => AzureArcManagedIdentitySource.Create(requestContext),
                     ManagedIdentitySource.ImdsV2 => ImdsV2ManagedIdentitySource.Create(requestContext),
-                    _ => new ImdsManagedIdentitySource(requestContext)
+                    ManagedIdentitySource.Imds => ImdsManagedIdentitySource.Create(requestContext),
+                    _ => throw new MsalServiceException(MsalError.ManagedIdentityAllSourcesUnavailable, MsalErrorMessage.ManagedIdentityAllSourcesUnavailable)
```
gladjohn: is this really a service exception?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -611,6 +606,7 @@ public void TestCsrGeneration_BadCert_ThrowsMsalServiceException(string badCert)
             Assert.ThrowsException<MsalServiceException>(() =>
                 CsrValidator.ParseRawCsr(badCert));
         }
+        #endregion CSR Generation Tests
```
gladjohn, line 609: Are there tests to validate `GetManagedIdentitySourceAsync` does not probe again, when the application already returned it as None?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsVersion.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Client.ManagedIdentity.V2
```
gladjohn, line 4: can this be in `Microsoft.Identity.Client.ManagedIdentity` namespace?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -103,32 +102,45 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
...
+                if (imdsV2Response)
+                {
+                    requestContext.Logger.Info("[Managed Identity] ImdsV2 detected.");
+                    s_sourceName = ManagedIdentitySource.ImdsV2;
+                    return s_sourceName;
+                }
+            }
+            else
+            {
+                requestContext.Logger.Info("[Managed Identity] Mtls Pop was not requested; skipping ImdsV2 probe.");
```
gladjohn, line 132: We are really not skipping probe? can we actually do that?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
```diff
@@ -66,7 +80,17 @@ public async Task GetManagedIdentityTests(
 
                 ManagedIdentityApplication mi = miBuilder.Build() as ManagedIdentityApplication;
 
-                Assert.AreEqual(expectedManagedIdentitySource, await mi.GetManagedIdentitySourceAsync().ConfigureAwait(false));
+                if (managedIdentitySource == ManagedIdentitySource.ImdsV2)
```
gladjohn, line 85: Do we have a test where we first get a bearer and then request a POP token?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: what I mean is, "[Managed Identity] Mtls Pop was not requested; skipping ImdsV2 probe." is based on the probe result. Can we actually skip the probe itself?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: [Azure SDK has a 1 second timeout](https://github.com/Azure/azure-sdk-for-net/blob/fde1c3c42da59b46629c07b4c51ae01689e7cb02/sdk/identity/Azure.Identity/src/DefaultAzureCredentialFactory.cs#L293)

Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS: MsalClientException is more appropriate here.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsVersion.cs
bgavrilMS: Why @gladjohn ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -211,5 +220,106 @@ public static Uri GetValidatedEndpoint(
...
+            };
+
+            IRetryPolicyFactory retryPolicyFactory = requestContext.ServiceBundle.Config.RetryPolicyFactory;
+            IRetryPolicy retryPolicy = retryPolicyFactory.GetRetryPolicy(RequestType.ImdsProbe);
+
+            HttpResponse response = null;
+
+            try
+            {
+                using (var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(1)))
```
bgavrilMS: A better pattern for timeout is to let the caller (Azure SDK) provide it, via CancellationToken. So your async method GetSourceAsync() needs to take a parameetr GetSourceAsync(CancellationToken token) and you need to propagate this to the HTTP calls.

bgavrilMS: If we also want MSAL to provide a default timeout, we can combine our timeout with the one from the caller. But I think we are ok without a timeout of our own? We can always add one later.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/Http/Retry/ImdsProbeRetryPolicy.cs
Robbie-Microsoft: good catch. done

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: good catch. removed it.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: The ImdsV2 probe will never be skipped when called from the two APIs in `ManagedIdentityApplication.cs`: `GetManagedIdentitySourceAsync` and `GetManagedIdentitySource`.

This will only ever trigger during the token acquisition call, if `.WithMtlsProofOfPosession` was not used.

A use case is: acquiring a token without first calling `GetManagedIdentitySourceAsync`.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: An exception will be thrown in that case - we have a test for that already.

Robbie-Microsoft: discussed offline. keeping as-is for now.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsVersion.cs
gladjohn: The enums aren’t V2-specific, so putting them under ManagedIdentity keeps the namespace clean and avoids implying versioned behavior.

Robbie-Microsoft: Like we discussed offline, I would prefer to keep everything that touches ImdsV2 in the V2 subfolder

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: discussed offline. keeping as-is for now.

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: done

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs
Robbie-Microsoft: This is currently not possible, we throw an error. We even have a unit test for this: `ApplicationsCannotSwitchBetweenImdsVersionsForPreview`

`public const string CannotSwitchBetweenImdsVersionsForPreview = "ImdsV2 is currently experimental - A Bearer token has already been received; Please restart the application to receive a mTLS PoP token.";`

bgavrilMS (approved):

PR #5615, opened 2025-12-09
Implemented Service Exception for Imds Probe

### LINKED ISSUE #5565: [Bug] IMDS v2 fails to detect source correctly when identity not found

Robbie-Microsoft:
This should also fix the flaky test related to Imds Probe Cancellation Token

Fixes #5565

### DIFF
--- a/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
+++ b/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
@@ -477,13 +477,10 @@ namespace Microsoft.Identity.Test.Unit.ManagedIdentityTests
                 cts.Cancel();
                 var imdsProbesCancellationToken = cts.Token;
 
-                var ex =
-                    await Assert.ThrowsExceptionAsync<MsalServiceException>(async () =>
-                        await (managedIdentityApp as ManagedIdentityApplication).GetManagedIdentitySourceAsync(imdsProbesCancellationToken)
-                        .ConfigureAwait(false))
-                    .ConfigureAwait(false);
-
-                Assert.AreEqual(MsalError.ImdsServiceError, ex.ErrorCode);
+                await Assert.ThrowsExceptionAsync<TaskCanceledException>(async () =>
+                    await (managedIdentityApp as ManagedIdentityApplication).GetManagedIdentitySourceAsync(imdsProbesCancellationToken)
+                    .ConfigureAwait(false))
+                .ConfigureAwait(false);
             }
         }
         #endregion Probe Tests


bgavrilMS (approved):

gladjohn (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -305,8 +305,9 @@ public static async Task<bool> ProbeImdsEndpointAsync(
             }
             catch (Exception ex)
             {
-                requestContext.Logger.Info($"[Managed Identity] {imdsStringHelper} probe endpoint failure. Exception occurred while sending request to probe endpoint: {ex}");
-                return false;
+                throw new MsalServiceException(
```
bgavrilMS: How about this: 

1. Let cancellation exceptions pass through
2. Capture the messages why IMDS1 and IMDS2 probe failed and output them when app calls AcquireToken() . Smth like "Cannot acquire a token for managed identity. IMDSv1 discovery failed with xyz and IMDSv2 discovery failed with bar"

bgavrilMS: @christothes - do you need the new `GetSourceAsync` API to also return to you why IMDSv1 and IMDSv2 probes failed when we return `Source.None` ?

christothes: If we are able to pass the failure reason back to the caller, it would be better to have it at GetSourceAsync than at GetToken. For DefaultAzureCredential, it will determine if we skip managed identity alltogether.

bgavrilMS: Sounds good, let's return an object then @Robbie-Microsoft that has the enum and the failures of IDMSv2 and IMDSv1

Robbie-Microsoft: done

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
```diff
@@ -303,22 +303,39 @@ public static async Task<bool> ProbeImdsEndpointAsync(
                     retryPolicy: retryPolicy)
                 .ConfigureAwait(false);
             }
+            // only throw on cancellation exceptions
+            catch (OperationCanceledException ex)
+            {
+                string failureMessage = $"{imdsStringHelper} probe timed out or was cancelled.";
+                requestContext.Logger.Error($"[Managed Identity] {failureMessage}");
+
+                var exception = MsalServiceExceptionFactory.CreateManagedIdentityException(
```
bgavrilMS: Let's please not throw a new type of exception on cancellation. The cancellation docs I shared state that if the token is cancelled, then an OperationCancelledException is thrown. 

How about

```csharp
 try
    {
        await DoThingAsync(cts.Token);
    }
    catch (Exception ex) when (ex is not OperationCanceledException)
    {

          string failureMessage = $"{imdsStringHelper} probe failed. Exception: {ex.Message}";
          requestContext.Logger.Info(() => $"[Managed Identity] {failureMessage}");
          return (false, failureMessage);
    }
```

See https://blog.stephencleary.com/2022/03/cancellation-3-detecting-cancellation.html

bgavrilMS: We also don't need to log when the cancellation exception occurs

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -54,13 +54,15 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
             {
                 requestContext.Logger.Info($"[Managed Identity] Selecting managed identity source if not cached. Cached value is {s_sourceName} ");
 
+                ManagedIdentitySourceResult sourceResult = null;
                 ManagedIdentitySource source;
```
bgavrilMS, line 58: nit: why do you need this variable? you can just use `sourceResult.Source`

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -229,13 +241,40 @@ private static bool ValidateAzureArcEnvironment(string identityEndpoint, string
...
+        /// </summary>
+        private static MsalClientException CreateManagedIdentityUnavailableException(ManagedIdentitySourceResult sourceResult)
+        {
+            string errorMessage = MsalErrorMessage.ManagedIdentityAllSourcesUnavailable;
+
+            if (sourceResult != null)
+            {
+                if (!string.IsNullOrEmpty(sourceResult.ImdsV1FailureReason) || !string.IsNullOrEmpty(sourceResult.ImdsV2FailureReason))
+                {
+                    errorMessage += " IMDS probe failures:";
```
bgavrilMS: `IMDS probe` is a technical concept and ppl using the library won't understand. How about "MSAL was not able to detect the Azure Instance Metadata Service (IMDS) that runs on VMs ..."

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
Robbie-Microsoft: The source variable is necessary because sourceResult is Null when reusing the cached value (lines 68-72), so accessing sourceResult.Source directly would cause a NullReferenceException.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

src/client/Microsoft.Identity.Client/ManagedIdentity/ImdsManagedIdentitySource.cs
Robbie-Microsoft: done

bgavrilMS (approved):

4gust (approved):

PR #5558, opened 2025-10-28
Labmigration

RyAuld:
Overview
This PR represents a comprehensive migration to modernize the lab infrastructure used for authentication testing in MSAL.NET. The migration moves from legacy lab APIs to a new Key Vault-based approach for improved security, reliability, and maintainability.

🎯 Migration Goals
Security: Transition from legacy lab API to secure Key Vault-based credential management
Reliability: Improve test infrastructure stability and reduce external dependencies
Maintainability: Modernize codebase with enhanced debugging and error handling
Compatibility: Ensure seamless integration with existing test frameworks
🔧 Key Changes
Infrastructure Modernization
✅ Complete migration to new lab infrastructure
✅ Key Vault integration for secure credential management
✅ Enhanced LabUserHelper methods with improved error handling
✅ ADFS 2022 support and expanded authentication scenarios
Test Framework Improvements
✅ Username/Password integration tests updated with new lab methods
✅ OnBehalfOf tests migrated to Key Vault approach
✅ Hybrid SPA account handling fixes for idlab1@msidlab4 user scenarios
✅ Debug logging enhancements for better troubleshooting
Quality Assurance
✅ Fixed failing unit tests related to iOS cache key generation
✅ Custom JWT token scenarios for improved test coverage
✅ ObjectId validation in Key Vault operations
✅ Vault reference updates from buildautomation to ID4sKeyVault
🧪 Testing Strategy
The migration maintains backward compatibility while introducing modern practices:

All existing test scenarios preserved and enhanced
New Key Vault-based authentication flows validated
iOS cache key generation compatibility verified
Custom test token scenarios for reduced external dependencies
📈 Benefits
🔒 Enhanced Security: Key Vault-based credential management
⚡ Improved Performance: Reduced reliance on external lab APIs
🛠️ Better Debugging: Enhanced logging and error reporting
🔄 Future-Ready: Modern infrastructure supporting ongoing development
✅ Quality Assurance: All tests passing with improved reliability
🚀 Migration Status
 Core infrastructure migration completed
 Key Vault integration implemented
 Test framework updates applied
 Unit test compatibility issues resolved
 Integration test scenarios validated
 Debug and logging improvements deployed

PR #5631, opened 2026-01-05
Refactor and simplify Microsoft.Identity.Test.LabInfrastructure

Avery-Dunn:
Continues the lab migration work begun in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5558 and other PRs based on discussions and similar refactoring done in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/1002

In short, these changes remove the last pieces of the old API-based system of retrieving test config in favor of the new key vault-based approach, and should make the Microsoft.Identity.Test.LabInfrastructure package easier to maintain going forward by simplifying the various helper classes/methods

Although a lot of files were changed, the majority of them don't need a deep review: they are test classes that were only changed to use the new helper class/method names and behavior, each file received the same sort of adjustment, and no test behavior was actually changed.

The only files that need a real review are in the Microsoft.Identity.Test.LabInfrastructure namespace, and here is a brief summary of the changes:
 - Simplify the helper classes/methods and clarify what they are doing, and allow integration tests to retrieve only the configuration they need
    - `LabResponse`: Previously just a container for user/app/lab config JSON classes, deleted in favor of separate `XYZConfig` classes
    - `AppConfig` and `UserConfig`: Represents the JSON-based configurations stored in the key vault
    - `KeyVaultSecrets`: Added to store the names of secrets in our key vaults, rather than hardcoding them in helper methods
    - `KeyVaultSecretsProvider`: Minor changes to comments to clarify what the key vaults are used for
    - `LabResponseHelper`: The main helper for retrieving configuration from the key vault, and the class that received the most meaningful changes
      - Renamed from `LabUserHelper`, as it does more than user config
      - Removed numerous `GetXYZUserAsync` methods which had hardcoded secret names and, despite the name, returned a merged response containing user, app, and lab configuration
      - Added `GetUserConfigAsync` and `GetAppConfigAsync` methods to return separate `XYZConfig` instances based on key vault secret names
      - Added three `ConcurrentDictionary` fields to cache deserialized JSON config responses from the key vault
- Remove the last references to the Lab API
    - `LabApiConstants`: Previously stored query parameters and other constants for API calls, deleted because those calls are no longer made
    - `LabServiceParameters`: Previously stored numerous enums that were compared to results in API calls, deleted in favor of a simpler `LabConstants` file as the vast majority weren't actually used
    - `LabConstants`: Remnants of the enums that were in `LabServiceParameters`, left here to avoid changing tests and making this PR even larger
    - `LabServiceApi`: Previously was the main API helper class, deleted because those calls are no longer made
    - `UserQueryParameters` and `LabUserNotFoundException`: Previously used only to help format a strings in a few debug logs, deleted in favor of simpler error message behavior


tests/Microsoft.Identity.Test.LabInfrastructure/AppConfig.cs
bgavrilMS: If this is not used, let's remove it?

bgavrilMS: Remove if not used?

tests/Microsoft.Identity.Test.LabInfrastructure/KeyVaultSecrets.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Microsoft.Identity.Test.LabInfrastructure
+{
+    /// <summary>
+    /// Contains names of secrets stored in Azure Key Vault for lab configuration.
+    /// These secrets contain JSON-serialized configuration objects for users, apps, and lab environments.
+    /// </summary>
+    public static class KeyVaultSecrets
+    {
+        // User secrets
```
bgavrilMS: nit: these are names of KV entries and the values are JSON files with various metadata

tests/Microsoft.Identity.Test.LabInfrastructure/LabConfig.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Identity.Test.LabInfrastructure
+{
+    /// <summary>
+    /// Represents lab environment configuration retrieved from Key Vault.
+    /// </summary>
+    public class LabConfig
```
bgavrilMS: We discussed about removing this. The app and the user can have the tenant ID. Same for authority host.

tests/Microsoft.Identity.Test.LabInfrastructure/KeyVaultSecrets.cs
```diff
@@ -0,0 +1,39 @@
...
+        public const string UserPublicCloud = "User-PublicCloud-Config";
+        public const string UserFederated = "User-Federated-Config";
+        public const string UserPublicCloud2 = "MSAL-User-Default2-JSON";
+        public const string UserXcg = "MSAL-User-XCG-JSON";
+        public const string UserB2C = "MSAL-USER-B2C-JSON";
+        public const string UserArlington = "MSAL-USER-Arlington-JSON";
+        public const string UserCiam = "MSAL-USER-CIAM-JSON";
+        public const string UserPop = "MSAL-User-POP-JSON";
```
bgavrilMS, line 20: Token Types are enabled for all users? Do we really need a special entry for POP?

tests/Microsoft.Identity.Test.LabInfrastructure/LabConfig.cs
Avery-Dunn: Originally I kept it just to keep the PR simpler and avoid making too many changes (it was getting quite big), but since `LabConfig` was only used in a few tests I just removed it and all the references in the latest commit.

tests/Microsoft.Identity.Test.LabInfrastructure/KeyVaultSecrets.cs
Avery-Dunn: Improved the comments in the latest commit

tests/Microsoft.Identity.Test.LabInfrastructure/AppConfig.cs
Avery-Dunn: This must've been old comment, because despite what it says that property is used and is string (this used to be an inner class in the old `LabResponse`).

In the latest commit I cleaned up the comment.

Avery-Dunn: This must've been old comment, because despite what it says that property is used and is string (this used to be an inner class in the old `LabResponse`).

In the latest commit I cleaned up the comment.

tests/Microsoft.Identity.Test.LabInfrastructure/KeyVaultSecrets.cs
Avery-Dunn: I didn't test every one, but several tests fail when trying to use the basic public cloud user.

Using POP as an example, it's only used in two tests in [RuntimeBrokerTests](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/d8b23ade7ff1d8610ac07305105fcb89f41ed80a/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/RuntimeBrokerTests.cs): one of them works with the public cloud user, however the one expecting an MsalUiRequiredException fails.

So, I think we can definitely clean up some of the users and apps but it'd need some changes over on the Azure side or some changes to the tests that I didn't want to complicate this PR with.

bgavrilMS:
@MZOLN @gladjohn @cpp11nullptr - can you please review this as this introduces breaking changes to the lab API

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ManagedIdentityTests.NetFwk.cs
```diff
@@ -284,9 +285,9 @@ public async Task ManagedIdentityRequestFailureCheckAsync(MsiAzureResource azure
             //Arrange
             using (new EnvVariableContext())
             {
-                //Get the Environment Variables
+                //Get the Environment Variables from Key Vault
```
gladjohn, line 288: doesn't this change the intent of the test? previously we get the environment variable from the azure resource, now we get it from key vault?

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/PoPTests.NetFwk.cs
```diff
@@ -295,17 +294,52 @@ public async Task PopTestWithRSAAsync()
             Assert.AreEqual("RS256", alg, "The algorithm in the token header should be RS256");
         }
 
+        //[DoNotRunOnLinux] // POP is not supported on Linux
```
gladjohn: delete? if not needed?

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ManagedIdentityTests.NetFwk.cs
Avery-Dunn: I don't think it really changes the test: the old way used the same `LabServiceApi.GetLabResponseAsync` helper method to get a JSON just like all of the other old Lab API calls, the only difference was that it got it from "https://service.msidlab.com/" instead of "https://msidlab.com" 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/29428d242fe7616806e4e1d78e98baa3f59e4576/tests/Microsoft.Identity.Test.LabInfrastructure/LabUserHelper.cs#L117

The only difference is where the JSON comes from, but that's a black box to the test. So from the perspective of the test nothing has changed: it just calls a different helper method to get the same dictionary of environment variables.

(one thing that is different is that the "resource" used to be added as a query parameter to the API call, however the exact same info was returned no matter what the resource was so I didn't think it was important for the test)

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/PoPTests.NetFwk.cs
Avery-Dunn: Good catch, accidentally committed that when trying to solve a test failure.

Removed in the latest commit.

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ManagedIdentityTests.NetFwk.cs
gladjohn: The resource can be extended to a VM or a Function App in the future.

PR #5409, opened 2025-07-28
Enable mTLS Proof‑of‑Possession for Client‑Assertion Delegates

gladjohn:
Fixes #5398

**Changes proposed in this request**
This pull request introduces significant enhancements to the mutual TLS Proof-of-Possession (mTLS PoP) functionality, refactors client assertion handling, and updates public APIs to support these changes. The most important changes include adding support for certificate-bound client assertions, refactoring the `ClientCredential` classes, and introducing new public APIs for client assertion configuration.

### Enhancements to mTLS Proof-of-Possession:

- **Refactored `WithMtlsProofOfPossession` method**: Updated the method to handle multiple credential types (`CertificateClientCredential` and `ClientAssertionDelegateCredential`) and added logic for certificate-bound client assertions. Improved error handling for cases where certificates are not provided. (`src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs`, [src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.csL99-R126](diffhunk://#diff-66e58df1ddbe7bfef74ce08f426e8ebbdb78e7258838c181d5d497bd6c7236eeL99-R126))
- **Updated error messages**: Enhanced the error message for missing certificates in mTLS PoP to include guidance for configuring applications or chaining methods. (`src/client/Microsoft.Identity.Client/MsalErrorMessage.cs`, [src/client/Microsoft.Identity.Client/MsalErrorMessage.csL439-R439](diffhunk://#diff-5f7fcb836912bee0deb52d9dbdedafd36ce6fad65c20327ea497847edb7f2fdeL439-R439))

### Refactoring of Client Assertion Handling:

[...]

PR #5649, opened 2026-01-12
[DRAFT] Add request-level .WithMtlsBearer() API for mTLS bearer + client assertion flows

gladjohn:
### Summary
This PR introduces a new request-level API, **`.WithMtlsBearer()`**, to enable **mTLS at the transport layer** for confidential client flows while keeping the resulting access token as a **Bearer** token.

This is primarily intended for **client assertion delegate** scenarios where the client must present a certificate during the TLS handshake (mTLS), but does **not** require PoP/token-binding semantics.

@bgavrilMS can we also extend this for SNI? 

### Background / Related work
- Builds on the existing mTLS PoP + SNI and regional mTLS discovery work from **#4965**
- Builds on the client assertion mTLS PoP support added in **#5409**
- This completes the “mTLS assertion” story by adding the **mTLS bearer** variant.

### Design notes
- `.WithMtlsBearer()` follows the same overall pattern as `.WithMtlsProofOfPossession()`:
  - Request-level opt-in
  - Uses the same regional mTLS endpoint / SNI discovery path
  - Requires **Azure region** when used with AAD authorities (same validation behavior as mTLS PoP)
- For **client assertion delegates**, the assertion callback must return a **TokenBindingCertificate** (used strictly for TLS), otherwise MSAL throws an actionable error.
- API name not final @bgavrilMS FYI 
- TO-DO - Mark API as experimental? @bgavrilMS FYI 

### Current status
- **Draft PR**: implementation only
- **Not tested yet** (no unit tests, no manual validation)

### Next steps
1. **Add unit tests**
   - Region required behavior for AAD authorities
   - Client assertion delegate requires a non-null `TokenBindingCertificate`
   - Ensure mTLS bearer does not trigger PoP-specific behaviors (e.g. no cert in auth result, etc)
2. **Manual validation**
   - Validate against AAD regional mTLS endpoints using SNI
   - Validate client assertion delegate path end-to-end (with MSI v2) 
   - Validate regressions for existing mTLS PoP flows

### Notes / Open questions
- Confirm whether any existing gating logic relies specifically on “mTLS PoP enabled” checks and should be generalized to “mTLS enabled” checks for both PoP and Bearer.

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ManagedIdentityTests.NetFwk.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ManagedIdentityTests.NetFwk.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
 using System.Linq.Expressions;
@@ -11,6 +12,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
+using Azure.Core;
 using Castle.Core.Internal;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.AppConfig;
@@ -60,6 +62,9 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
         //non existent Resource ID of the User Assigned Identity 
         private const string Non_Existent_UamiResourceId = "/subscriptions/userAssignedIdentities/NO_ID";
 
+        private AccessToken? _labApiAccessToken;
+        private string _labAccessAppId = new KeyVaultSecretsProvider().GetSecretByName("LabVaultAppID").Value;
+
         [DataTestMethod]
         [DataRow(MsiAzureResource.WebApp, "", DisplayName = "System_Identity_Web_App")]
         [DataRow(MsiAzureResource.WebApp, UserAssignedClientID, UserAssignedIdentityId.ClientId, DisplayName = "ClientId_Web_App")]
@@ -72,7 +77,7 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
             {
                 // Fetch the env variables from Key Vault and set them locally
                 Dictionary<string, string> envVariables = 
-                    await LabResponseHelper.GetDefaultMSIEnvironmentVariablesAsync().ConfigureAwait(false);
+                    await GetEnvironmentVariablesAsync(azureResource).ConfigureAwait(false);
 
                 //Set the Environment Variables
                 SetEnvironmentVariables(envVariables);
@@ -127,7 +132,7 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
             {
                 // Fetch the env variables from Key Vault and set them locally
                 Dictionary<string, string> envVariables =
-                    await LabResponseHelper.GetDefaultMSIEnvironmentVariablesAsync().ConfigureAwait(false);
+                    await GetEnvironmentVariablesAsync(MsiAzureResource.WebApp).ConfigureAwait(false);
         
                 //Set the Environment Variables
                 SetEnvironmentVariables(envVariables);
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
```diff
@@ -40,8 +40,23 @@ internal class AcquireTokenCommonParameters
...
         public bool IsMtlsPopRequested { get; set; }
+        public bool IsMtlsBearerRequested { get; set; }
+
         internal Func<AttestationTokenInput, CancellationToken, Task<AttestationTokenResponse>> AttestationTokenProvider { get; set; }
 
+        internal async Task InitMtlsParametersAsync(IServiceBundle serviceBundle, CancellationToken ct)
+        {
+            if (IsMtlsPopRequested && IsMtlsBearerRequested)
+            {
+                throw new MsalClientException(
```
bgavrilMS: How can the app developer ever get into this situation? I think this is more of an Assert, not an exception.

src/client/Microsoft.Identity.Client/AuthScheme/PoP/MtlsBearerAuthenticationOperation.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System.Collections.Generic;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client.Internal;
+using Microsoft.Identity.Client.TelemetryCore.Internal.Events;
+using Microsoft.Identity.Client.Utils;
+
+namespace Microsoft.Identity.Client.AuthScheme.PoP
+{
+    internal sealed class MtlsBearerAuthenticationOperation : IAuthenticationOperation
```
bgavrilMS: I don't think the Bearer tokens should be tied to the mtls certificate. Only the credential is.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -113,6 +113,30 @@ public AcquireTokenForClientParameterBuilder WithMtlsProofOfPossession()
             return this;
         }
 
+        /// <summary>
+        /// Specifies that the certificate provided will be used for Bearer tokens with mTLS (Mutual TLS) authentication.
+        /// For more information, refer to the <see href="https://aka.ms/mtls-pop">Proof-of-Possession documentation</see>.
+        /// </summary>
+        /// <returns>The current instance of <see cref="AcquireTokenForClientParameterBuilder"/> to enable method chaining.</returns>
+        public AcquireTokenForClientParameterBuilder WithMtlsBearer()
```
bgavrilMS: No, we are not introducing a new API. We discussed that it is not necessary for FIC.

```csharp
cca.WithClientAssertion( (_) => { return (bound_credential, bound_certificate) }; 
cca.AcquireTokenForClient(); // goes over mtls because of the credential used
```

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/AcquireTokenCommonParameters.cs
```diff
@@ -40,8 +40,23 @@ internal class AcquireTokenCommonParameters
         public string FmiPathSuffix { get; internal set; }
         public string ClientAssertionFmiPath { get; internal set; }
         public bool IsMtlsPopRequested { get; set; }
+        public bool IsMtlsBearerRequested { get; set; }
```
bgavrilMS: All this state is not necessary. It should be computed dynamically - if the assertion has a certificate, the token request should be done over mTLS

bgavrilMS (changes requested):
No new API

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
gladjohn: Ok go it, I can convert this to a pure FIC only implementation and remove the SNI stuff out.

PR #5612, opened 2025-12-08
MSI POP Attestation

Robbie-Microsoft:
Implements #5591 

This pull request refactors the solution to replace the `Microsoft.Identity.Client.MtlsPop` project and namespace with `Microsoft.Identity.Client.KeyAttestation`. It updates all relevant references, configuration, and code to use the new name and adds attestation support to the managed identity flow. The most important changes are grouped below:

Project and Solution Structure:

* Removed the `Microsoft.Identity.Client.MtlsPop` project from the solution and replaced it with the new `Microsoft.Identity.Client.KeyAttestation` project in `LibsAndSamples.sln`. All solution configuration entries and project references have been updated accordingly. [[1]](diffhunk://#diff-658b40bd7b7140793d178fa774659068a808830dff1ac411909eb534f86a479fL195-R198) [[2]](diffhunk://#diff-658b40bd7b7140793d178fa774659068a808830dff1ac411909eb534f86a479fL1951-L1992) [[3]](diffhunk://#diff-658b40bd7b7140793d178fa774659068a808830dff1ac411909eb534f86a479fR1993-R2034) [[4]](diffhunk://#diff-658b40bd7b7140793d178fa774659068a808830dff1ac411909eb534f86a479fL2092-R2093)

Code and Namespace Refactoring:

[...]

PR #5653, opened 2026-01-13
Update assembly name for KeyAttestation packaging

gladjohn:
Update assembly name for KeyAttestation packaging

### DIFF
--- a/build/template-pack-and-sign-all-nugets.yaml
+++ b/build/template-pack-and-sign-all-nugets.yaml
@@ -44,12 +44,12 @@ steps:
     ProjectRootPath: '$(Build.SourcesDirectory)\$(MsalSourceDir)src\client'
     AssemblyName: 'Microsoft.Identity.Client.Extensions.Msal'
 
-# Sign binary and pack Microsoft.Identity.Client.MtlsPop
+# Sign binary and pack Microsoft.Identity.Client.KeyAttestation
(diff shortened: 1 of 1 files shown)


RyAuld (approved):

build/template-pack-and-sign-all-nugets.yaml
```diff
@@ -44,12 +44,12 @@ steps:
     ProjectRootPath: '$(Build.SourcesDirectory)\$(MsalSourceDir)src\client'
     AssemblyName: 'Microsoft.Identity.Client.Extensions.Msal'
 
-# Sign binary and pack Microsoft.Identity.Client.MtlsPop
+# Sign binary and pack Microsoft.Identity.Client.KeyAttestation
```
neha-bhargava, line 47: Just curious why is the name being changed? Would it not impact the existing users?

gladjohn: Good question. This is still in internal preview and we don’t have any external customers consuming this package yet, so the rename shouldn’t impact existing users in practice.

Also, the package scope has shifted from mTLS PoP support to attestation support for KeyGuard, so the new name reflects the updated functionality more accurately.

The [rename PR](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5612) is already merged—this PR is only updating the package signing to align with the new package name.

PR #5656, opened 2026-01-14
Managed Identity: make IMDS probing explicit and restore DefaultToImds fallback (no-probe by default)

### LINKED ISSUE #5651: [Bug] ManagedIdentityCredential within ChainedTokenCredential throwing exception on 4.81.0
### Library version used

4.81.0

### .NET version

.net10 x64

### Scenario

We're running our code on an Azure WebApp while developing it in VisualStudio. For authentication we're using chained `ManagedIdentityCredential` and `VisualStudioCredential`.

Yesterday I tried to upgrade to `4.81.0` and received the below exception when I run it locally through VisualStudio. The same code works fine on `4.79.2`

> Azure.Identity.AuthenticationFailedException
>   HResult=0x80131500
>   Message=The ChainedTokenCredential failed due to an unhandled exception: ManagedIdentityCredential authentication failed: All Managed Identity sources are unavailable. MSAL was not able to detect the Azure Instance Metadata Service (IMDS) that runs on VMs: IMDSv1: IMDSv1 probe failed. Exception: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. (169.254.169.254:80).
> See the troubleshooting guide for more information. https://aka.ms/azsdk/net/identity/managedidentitycredential/troubleshoot
>   Source=Azure.Identity
>   StackTrace:
>    at Azure.Identity.ChainedTokenCredential.<GetTokenImplAsync>d__7.MoveNext()
>    ....
> 
>   This exception was originally thrown at this call stack:
>     [External Code]
> 
> Inner Exception 1:
> AuthenticationFailedException: ManagedIdentityCredential authentication failed: All Managed Identity sources are unavailable. MSAL was not able to detect the Azure Instance Metadata Service (IMDS) that runs on VMs: IMDSv1: IMDSv1 probe failed. Exception: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. (169.254.169.254:80).
> See the troubleshooting guide for more information. https://aka.ms/azsdk/net/identity/managedidentitycredential/troubleshoot
> 
> Inner Exception 2:
> MsalClientException: All Managed Identity sources are unavailable. MSAL was not able to detect the Azure Instance Metadata Service (IMDS) that runs on VMs: IMDSv1: IMDSv1 probe failed. Exception: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. (169.254.169.254:80).
> 

### Relevant code snippets

```csharp
TokenCredential credential = new ChainedTokenCredential(
    new ManagedIdentityCredential(),
    new VisualStudioCredential(new VisualStudioCredentialOptions { TenantId = "..." })
);
```

### Expected behavior

I'm expecting the same behaviour as on 4.79.2.

### Identity provider

### Regression

4.79.2

### Solution and workarounds

I could write my own ChainedTokenCredential , catching the right exceptions. But thats not the point...

gladjohn:
Fixes #5651

## Description
This PR updates Managed Identity (MI) source detection to avoid implicit IMDS network probing by default and to support explicit probe semantics, addressing the chaining regression reported in #5651.

### Background
Recent changes caused MI source selection/detection to probe IMDS implicitly. On non-Azure environments (e.g., local dev), this could surface failures early and break credential chaining scenarios that expect “MI unavailable” to be non-fatal.

#### Explicit probe behavior
MI source detection now supports an explicit probe flag:
- `probe = false` (default): **no IMDS probe**
- `probe = true`: **probe IMDS** to confirm availability

#### `probe = false` (default)
- If an environment-based MI source is detected (App Service, Service Fabric, Arc, etc.), return it as before.
- If no environment-based source is detected, return and cache `ManagedIdentitySource.DefaultToImds` as a sentinel:
  - Meaning: “don’t probe; use IMDS as the default fallback source.”

#### `probe = true`
- Probe IMDS (v2 when applicable, then v1).
- If IMDS is reachable, return `ImdsV2` or `Imds`.
- If IMDS is unreachable, return `ManagedIdentitySource.None` (no fatal exception during detection), enabling callers to continue chaining.

### mTLS PoP behavior
- If mTLS PoP is requested while the source is `DefaultToImds`, MSAL throws `MsalClientException` with `MtlsPopTokenNotSupportedinImdsV1`, consistent with the existing “IMDSv1 does not support PoP” semantics.
- The preview guard preventing switching between IMDS versions is preserved (`CannotSwitchBetweenImdsVersionsForPreview`).

### Tests
- Updated unit tests to align with the explicit-probe model:
  - Tests that validate IMDS detection now call the probe overload explicitly.
  - Adjusted test helpers to avoid registering IMDS probe mocks when the code path throws before probing, preventing leftover mock assertions.

### Impact
- Restores expected default behavior for chaining: **no implicit IMDS probe**, and `DefaultToImds` is used as the default fallback sentinel.

### DIFF
--- a/tests/Microsoft.Identity.Test.LabInfrastructure/CertificateHelper.cs
+++ b/tests/Microsoft.Identity.Test.LabInfrastructure/CertificateHelper.cs
@@ -2,8 +2,8 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Runtime.InteropServices;
 using System.Security.Cryptography.X509Certificates;
-using Microsoft.Identity.Client.Extensions.Msal;
 
 namespace Microsoft.Identity.Test.LabInfrastructure
 {
@@ -39,7 +39,7 @@ namespace Microsoft.Identity.Test.LabInfrastructure
         private static X509Certificate2 FindCertificateByName(string certName, StoreLocation location, StoreName name)
         {
             // Unix LocalMachine X509Store is limited to the Root and CertificateAuthority stores
-            if (SharedUtilities.IsLinuxPlatform())
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
             {
                 var certPasswrod = Environment.GetEnvironmentVariable("CERTIFICATE_PASSWORD");
                 var certLocation = Environment.GetEnvironmentVariable("CERTIFICATE_LOCATION");
--- a/tests/Microsoft.Identity.Test.LabInfrastructure/LabResponseHelper.cs
+++ b/tests/Microsoft.Identity.Test.LabInfrastructure/LabResponseHelper.cs
@@ -15,8 +15,8 @@ namespace Microsoft.Identity.Test.LabInfrastructure
         public static KeyVaultSecretsProvider KeyVaultSecretsProviderMsid { get; }
 
         // Caches for configuration objects retrieved from Key Vault
-        private static readonly System.Collections.Concurrent.ConcurrentDictionary<string, UserConfig> s_userConfigCache = new();
-        private static readonly System.Collections.Concurrent.ConcurrentDictionary<string, AppConfig> s_appConfigCache = new();
+         private static readonly System.Collections.Concurrent.ConcurrentDictionary<string, UserConfig> s_userConfigCache = new System.Collections.Concurrent.ConcurrentDictionary<string, UserConfig>();
+        private static readonly System.Collections.Concurrent.ConcurrentDictionary<string, AppConfig> s_appConfigCache = new System.Collections.Concurrent.ConcurrentDictionary<string, AppConfig>();
 
         static LabResponseHelper()
         {


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -99,7 +109,7 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
 
                 // If the source is determined to be ImdsV1 and mTLS PoP was requested,
                 // throw an exception since ImdsV1 does not support mTLS PoP
-                if (source == ManagedIdentitySource.Imds && isMtlsPopRequested)
+                if (source == ManagedIdentitySource.DefaultToImds && isMtlsPopRequested)
```
bgavrilMS, line 112: Shouldn't this be "if source != IMDSv2 && popRequested" ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -113,18 +123,33 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
...
+                        ? ImdsV2ManagedIdentitySource.Create(requestContext)
+                        : ImdsManagedIdentitySource.Create(requestContext),
                     ManagedIdentitySource.ImdsV2 => ImdsV2ManagedIdentitySource.Create(requestContext),
                     ManagedIdentitySource.Imds => ImdsManagedIdentitySource.Create(requestContext),
                     _ => throw CreateManagedIdentityUnavailableException(sourceResult)
                 };
             }
         }
 
+        // Back-compat wrapper: previous callers who don't specify probe now get probe=false behavior (no probe).
```
bgavrilMS: This is internal API. What back-compat are we talking about?

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -56,18 +56,46 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
         }
 
         /// <inheritdoc/>
-        public async Task<ManagedIdentitySourceResult> GetManagedIdentitySourceAsync(CancellationToken cancellationToken)
+        public Task<ManagedIdentitySourceResult> GetManagedIdentitySourceAsync(CancellationToken cancellationToken)
```
bgavrilMS: No convenince APIs please.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -56,18 +56,46 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
...
+            return GetManagedIdentitySourceAsync(probe: false, cancellationToken: cancellationToken);
+        }
+
+        /// <summary>
+        /// Detects and returns the managed identity source available on the environment.
+        /// probe=false: no IMDS probe; returns DefaultToImds if no env-based MI source detected.
+        /// probe=true: probes IMDS; returns None if IMDS is unreachable.
+        /// </summary>
+        public async Task<ManagedIdentitySourceResult> GetManagedIdentitySourceAsync(
+            bool probe = false,
```
bgavrilMS: Let's not add a default here.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -56,18 +56,46 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
...
+            CancellationToken cancellationToken = default)
```
bgavrilMS: Inconsistent default usage here. I would recommend not using a default here.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: This check is intentionally scoped to IMDS v1 / DefaultToImds. When we have determined the source to be IMDS v1, if PoP is requested we fail fast with the existing “PoP not supported in IMDSv1” error rather than implicitly probing or attempting the IMDSv2 flow.

gladjohn: But let me see if there is non-IMDS (env based resource) tests that covers this scenario

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: Agreed — it’s internal, so “back-compat” wording was misleading.

I removed the wrapper and made the probe semantics explicit by adding the `probe` parameter to the internal `ManagedIdentityClient.GetManagedIdentitySourceAsync(...)` and updated call sites accordingly.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
gladjohn: Agreed.

I removed the convenience overload and now expose only the explicit API:
`GetManagedIdentitySourceAsync(bool probe, CancellationToken cancellationToken)`.
Callers must pass `probe` explicitly.

**notes for other reviewers?** we removed an API from shipped. This was internally used by Azure SDK and we have a contract with them.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
gladjohn: Agreed.

I removed defaults for both parameters — the method now requires both `probe` and `CancellationToken` to be passed explicitly (no optional defaults).

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
gladjohn: removed

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: added a test here - NonImdsSources_MtlsPopRequested_ReturnsBearer

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -55,19 +55,40 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
...
 
-            // Create a temporary RequestContext for the logger and the IMDS probe request.
+            // Create a temporary RequestContext for the logger and the optional IMDS probe request.
             var requestContext = new RequestContext(this.ServiceBundle, Guid.NewGuid(), null, cancellationToken);
 
-            // GetManagedIdentitySourceAsync might return ImdsV2 = true, but it still requires .WithMtlsProofOfPossesion on the Managed Identity Application object to hit the ImdsV2 flow
-            return await ManagedIdentityClient.GetManagedIdentitySourceAsync(requestContext, isMtlsPopRequested: true, cancellationToken).ConfigureAwait(false);
+            // Use isMtlsPopRequested: true so probe mode can detect IMDSv2 capability when available.
+            return await ManagedIdentityClient
+                .GetManagedIdentitySourceAsync(requestContext, isMtlsPopRequested: true, probe: probe, cancellationToken: cancellationToken)
```
neha-bhargava, line 90: Since isMtlsPopRequested is always true does that mean that the flow will never use IMDS v1? Or does the probe decide it? Do we also probe the v1 endpoint in this case?

gladjohn: Good question — `isMtlsPopRequested: true` here only controls the *probe order* when `probe == true`.

- If `probe == false` (default), we do **no network probe at all**, and we return `DefaultToImds` regardless of `isMtlsPopRequested`.

- If `probe == true`, then `isMtlsPopRequested: true` makes us *attempt the IMDSv2 probe first*. If that succeeds we return `ImdsV2`.
  If it fails (or is unreachable), we still **probe IMDSv1 next**, and return `Imds` if v1 is reachable.
  If both probes fail we return `None`.

So yes — in probe mode we do probe the v1 endpoint as well; `isMtlsPopRequested` does *not* block IMDSv1 detection. We have coverage for this in the probe tests where IMDSv2 probe fails and we fall back to an IMDSv1 probe + return `ManagedIdentitySource.Imds`.

gladjohn: I will setup sometime next week to go over MSI v2 changes and how it affects MSI v1.

neha-bhargava: That would be great! Thanks!

PR #5661, opened 2026-01-16
Fix MSI v2 PoP regression (pre-release) on VMs by probing IMDS for first-call mTLS PoP and add regression tests

gladjohn:
Fixes #pre-release bug

### Context
We introduced a regression where **mTLS Proof-of-Possession (PoP) token acquisition for Managed Identity** can fail on Azure VMs with:

`mtls_pop_token_not_supported_in_imds_v1`

This happens when MSAL determines the MI source as **IMDSv1** (or defaults to IMDSv1 behavior) even though the VM environment supports the **IMDSv2 PoP flow**.

This is a VM-specific scenario because VMs typically have **no environment-variable based MI signals**, so correct behavior depends on **IMDS probing**.

### Root cause
Our source-selection flow includes a *no-probe* path (returning `DefaultToImds`) intended to avoid IMDS network calls during selection for performance.

However, for PoP on a VM we must probe because:
- IMDSv2 is required for `mtls_pop`
- environment variables alone cannot confirm IMDSv2 availability on VMs

We also had a test coverage gap: several IMDSv2 tests effectively **pre-warmed the static source cache** by calling `GetManagedIdentitySourceAsync(probe: true)` during setup (via `addSourceCheck = true`). This masked the production behavior we need to validate: **first-call PoP AcquireToken must probe IMDS and select IMDSv2 automatically**.

### Fix
- Update Managed Identity source selection so that **when `.WithMtlsProofOfPossession()` is requested and no env-based source is detected**, MSAL runs **probe mode** to detect IMDSv2 (and selects IMDSv2 when available).
- Preserve preview guardrails around switching sources (e.g., “cannot switch between IMDS versions for preview”), while ensuring the **PoP-first scenario on VMs** works as expected.

### Tests added / updated
To prevent regressions, we add tests that validate **AcquireToken itself** triggers the correct probing behavior and does not depend on pre-probed cached state.

New/updated coverage includes:

1. **PoP first-call on VM probes and succeeds without any explicit `GetManagedIdentitySourceAsync(probe:true)` pre-call**
   - Validates that first `.WithMtlsProofOfPossession()` acquisition triggers IMDSv2 probing and succeeds when IMDSv2 endpoints are available.

2. **PoP requested when IMDSv2 probe fails but IMDSv1 is available returns `mtls_pop_token_not_supported_in_imds_v1`**
   - Ensures correct error semantics when only IMDSv1 exists.

3. *(Optional / if included)* **PoP requested when both IMDS probes fail returns `managed_identity_all_sources_unavailable`**
   - Ensures correct error when IMDS is unreachable.

Additionally:
- Tests intended to validate first-call behavior explicitly avoid any setup that would pre-warm `ManagedIdentityClient.s_sourceName` (e.g., disabling `addSourceCheck` for those tests).

### Why we’re fixing this now
This blocks customers from validating / adopting the Managed Identity IMDSv2 mTLS PoP flow on VMs and breaks the expected behavior of `.WithMtlsProofOfPossession()` in VM environments.

This PR restores:
- correct first-call PoP behavior on VMs
- correct error semantics when IMDSv2 is unavailable
- regression coverage that protects against future changes in source selection behavior

### Notes
- All unit tests pass after the changes, including updated IMDSv2 failure-path coverage.
- The change is scoped to PoP + VM-like environments (no env-var MI sources) and does not modify non-IMDS source behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The following table lists MSAL.NET versions currently supported and receiving se
 
 | Major Version | Last Release | Patch Release Date  | Support Phase|End of Support |
 | --------------|--------------|--------|------------|--------|
-| 4.x           | [![NuGet](https://img.shields.io/nuget/v/microsoft.identity.client.svg?style=flat-square&label=nuget&colorB=00b200)](https://www.nuget.org/packages/Microsoft.Identity.Client/)   |Monthly| Active | Not planned.<br/>✅Supported versions: from 4.72.1 to [![NuGet](https://img.shields.io/nuget/v/microsoft.identity.client.svg?style=flat-square&label=nuget&colorB=00b200)](https://www.nuget.org/packages/Microsoft.Identity.Client/)<br/>⚠️Unsupported versions &lt; `4.72.1`.|
+| 4.x           | [![NuGet](https://img.shields.io/nuget/v/microsoft.identity.client.svg?style=flat-square&label=nuget&colorB=00b200)](https://www.nuget.org/packages/Microsoft.Identity.Client/)   |Monthly| Active | Not planned.<br/>✅Supported versions: from 4.77.1 to [![NuGet](https://img.shields.io/nuget/v/microsoft.identity.client.svg?style=flat-square&label=nuget&colorB=00b200)](https://www.nuget.org/packages/Microsoft.Identity.Client/)<br/>⚠️Unsupported versions &lt; `4.77.1`.|
 
 ### Performance perspectives
 


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -64,23 +64,26 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
...
-                        probe: false,
+                        probe: mustProbeForPop, // false for Bearer default, true for PoP on VMs
                         cancellationToken)
                         .ConfigureAwait(false);
 
                     source = sourceResult.Source;
                 }
                 else
                 {
-                    // Reuse cached value
```
neha-bhargava, line 83: Is this comment no longer valid?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -64,23 +64,26 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
...
-                // If the source is not already set, determine it
-                if (s_sourceName == ManagedIdentitySource.None)
+                // If PoP is requested and the cached value is unknown/sentinel, we must probe to
+                // determine whether IMDSv2 is available (otherwise PoP will never work on VMs).
+                bool mustProbeForPop =
+                    isMtlsPopRequested &&
+                    (s_sourceName == ManagedIdentitySource.None || s_sourceName == ManagedIdentitySource.DefaultToImds);
+
+                // Determine source when not cached OR when PoP requires refining DefaultToImds.
+                if (s_sourceName == ManagedIdentitySource.None || mustProbeForPop)
```
neha-bhargava: The first condition here is redundant. It should be covered in the mustProbeForPop right?

gladjohn: mustProbeForPop is only true when PoP is requested; it includes s_sourceName == None only in that PoP case. We still need the s_sourceName == None check to cover the bearer path on first call (no cached source yet), where we must run env-based source selection. I can rename the condition to mustSelectSource to make this clearer.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: added back

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: simplified fix after some testing.

tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsTests.cs
```diff
@@ -31,9 +32,27 @@ private static IManagedIdentityApplication BuildMi(
...
+                MsalLogCallback,
+                LogLevel.Verbose,
+                enablePiiLogging: false,
+                enableDefaultPlatformLogging: false);
             return builder.Build();
         }
 
+        private static void MsalLogCallback(LogLevel level, string message, bool containsPii)
+        {
+            // Keep PII off. containsPii will be false in our config below, but keep this guard anyway.
```
bgavrilMS: Then don't add it? We should trust our own product.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -72,7 +72,7 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
                     sourceResult = await GetManagedIdentitySourceAsync(
                         requestContext,
                         isMtlsPopRequested,
-                        probe: false,
```
bgavrilMS, line 75: This whole probing logic is a never ending source of problems. This doesn't make sense. Why would we do probing only in certain scenarios? Can't we just always probe and deprecate "DefaultToImds"  (leave it only for legacy API)?

bgavrilMS (changes requested):
Too many complex scenarios and a constant source of bugs. Can we overhaul this?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
bgavrilMS, line 99: This also is a code smell. Getting the MSI source should not be influenced by the API that we use?

tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsTests.cs
gladjohn: AI generated content , was easier to add logs for testing - meant to revert it. will revert this change now

gladjohn:
closing per discussion

PR #5683, opened 2026-01-28
Telemetry for managed cert

### LINKED ISSUE #5682: [Engineering task] Update telemetry header to recognize if a managed cert is used
### Task type

Development

### Description

If a managed cert is used use extra query params to get that using key "ManagedCert" and pass the value to telemetry header. Do not send the query param to ests.

### Solution

_No response_

neha-bhargava:
Fixes #5682 

Telemetry for managed cert 

**Testing**
Unit tests

**Performance impact**

**Documentation**

### DIFF
--- a/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ClientCredentialsTests.WithRegion.cs
+++ b/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ClientCredentialsTests.WithRegion.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Identity.Test.Integration.HeadlessTests
             AuthenticationResult result = await GetAuthenticationResultAsync(appScopes).ConfigureAwait(false); // regional endpoint
             AssertTokenSourceIsIdp(result);
             AssertValidHost(true, factory);
-            AssertTelemetry(factory, $"{TelemetryConstants.HttpTelemetrySchemaVersion}|1004,{CacheRefreshReason.NoCachedAccessToken:D},centralus,3,4|0,1,1,,");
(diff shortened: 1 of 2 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -250,6 +250,13 @@ private ApiEvent InitializeApiEvent(string accountId)
             apiEvent.TokenType = AuthenticationRequestParameters.AuthenticationScheme.TelemetryTokenType;
             apiEvent.AssertionType = GetAssertionType();
 
+            if (AuthenticationRequestParameters.ExtraQueryParameters.TryGetValue(Constants.ManagedCertKey, out string managedCertValue) 
+                && !string.IsNullOrEmpty(managedCertValue))
+            {
+                apiEvent.IsManagedCertUsed = managedCertValue[0];
+            }
+            AuthenticationRequestParameters.ExtraQueryParameters.Remove(Constants.ManagedCertKey);
```
trwalke, line 258: Just curious, why to we use this collection to detect the managed cert and then remove it? we dont want to add it to the api somewhere?

neha-bhargava: Good question. I feel we can include it as a boolean param in CertificateOptions. @bgavrilMS thoughts on this?

neha-bhargava: Discussed with Bogdan and Thomas and we would like to keep it hidden. This is only for internal use.

PR #5688, opened 2026-01-30
Enhance C# guidelines with additional best practices

gladjohn:
Expanded C# development guidelines to include additional rules on naming, nullability, error handling, and testing practices.

Updated based on interactions with CoPilot on this PR - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5684

### DIFF
--- a/.clinerules/csharp-guidelines.md
+++ b/.clinerules/csharp-guidelines.md
@@ -1,117 +1,146 @@
 # C# Development Guidelines
 
-These guidelines define how to write and modify C# code in this repo. They apply to both humans and coding agents.
+These guidelines define how to write and modify C# code in this repo.
+They apply to both humans and coding agents.
 
-## General
+> Note: Some rules are repeated intentionally (Quick rules + Detailed rules) to reduce “agent drift”
+> and prevent hallucinated changes. The Detailed rules are the source of truth.
 
-- Use the C# language version and SDK **as configured by the repo**.
-  - Do **not** change `global.json` unless explicitly asked to.
-- Keep changes **minimal and scoped** to the request.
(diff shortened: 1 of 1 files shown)


.clinerules/csharp-guidelines.md
```diff
@@ -1,35 +1,117 @@
...
+These guidelines define how to write and modify C# code in this repo. They apply to both humans and coding agents.
+
 ## General
 
-* Always use the latest version C#, currently C# 13 features
-* Never change global.json unless explicitly asked to
-* Never change package.json or package-lock.json files unless explicitly asked to
-* Never change NuGet.config files unless explicitly asked to
+- Use the C# language version and SDK **as configured by the repo**.
+  - Do **not** change `global.json` unless explicitly asked to.
```
trwalke: Is this redundant because of what is on line 12?

.clinerules/csharp-guidelines.md
```diff
@@ -1,35 +1,117 @@
 # C# Development Guidelines
```
trwalke, line 1: Should we ask it to avoid using reflections? I've seen it do it a few times

neha-bhargava (approved):

.clinerules/csharp-guidelines.md
gladjohn: Good catch — it was duplicated in the same section. I’m going to restructure the doc so we have a short Quick rules section at the top, and then a single authoritative General section below. That keeps the reinforcement and helps reduce hallucinations.

.clinerules/csharp-guidelines.md
gladjohn: added an explicit “No reflection in product code”

PR #5695, opened 2026-01-31
Add synchronous overload for WithExtraBodyParameters to reduce async overhead on hot path

### LINKED ISSUE #5686: [Feature Request] Add a sync WithExtraBodyParameters extensibility method
### MSAL client type

Confidential

### Problem statement

WithExtraBodyParameters  is used on the hot path and most of the time it is used in a sync manner. However, we currently use a dictionary<string, async Func<_,string>> i.e. the values are obtained async. 

Forcing the async pattern over sync has a perf hit. 

Proposal is to add an overload `WithExtraBodyParameters(Dictionary<string, string>)`

### Proposed solution

_No response_

### Alternatives

_No response_

gladjohn:
# Add synchronous overload for WithExtraBodyParameters to reduce async overhead on hot path

---

# PR Description

## Summary

This PR adds a new synchronous overload of `WithExtraBodyParameters(Dictionary<string, string>)` to complement the existing async variant `WithExtraBodyParameters(Dictionary<string, Func<CancellationToken, Task<string>>>)`.

Fixes #5686

## Problem Statement

The current `WithExtraBodyParameters` API requires all parameter values to be obtained asynchronously via `Func<CancellationToken, Task<string>>`, even when parameters are statically known at build time. This forces the async pattern on the hot path and introduces unnecessary performance overhead in the common case where parameters don't require async computation.

## Solution

Introduce a new synchronous overload that:
- Accepts `Dictionary<string, string>` for static parameter values
- Eliminates async/await overhead for the common case
- Maintains full backward compatibility with existing async overload
- Internally converts sync parameters to equivalent async representation for cache key consistency
- Supports chaining and all existing features (FmiPath, cache, etc.)

### New API Surface

```csharp
public static AcquireTokenForClientParameterBuilder WithExtraBodyParameters(
    this AcquireTokenForClientParameterBuilder builder, 
    Dictionary<string, string> extrabodyparams)

### DIFF
--- a/src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
+++ b/src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
@@ -75,13 +75,21 @@ namespace Microsoft.Identity.Client.Extensibility
         }
 
         /// <summary>
-        /// Add extra body parameters to the token request (synchronous version). 
+        /// Add extra body parameters to the token request (synchronous version).
         /// These parameters are added to the cache key to associate these parameters with the acquired token.
         /// Use this overload when parameters are known synchronously to avoid async overhead.
         /// </summary>
-        /// <param name="builder"></param>
-        /// <param name="extrabodyparams">Dictionary of additional body parameters with static values</param>
-        /// <returns></returns>
+        /// <param name="builder">The builder instance</param>
+        /// <param name="extrabodyparams">Dictionary of static string values for additional body parameters</param>
+        /// <returns>The builder for method chaining</returns>
+        /// <remarks>
+        /// This synchronous overload is optimized for the common case where parameter values are statically known
+        /// and do not require async computation. Parameters are added directly to the HTTP request without any
+        /// async wrapping on the hot path, reducing memory allocation and GC pressure.
+        /// 
+        /// Cache key compatibility: Sync parameters with identical key/value pairs will produce the same cache keys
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs [hunk 1]
```diff
@@ -72,5 +73,44 @@ public static AcquireTokenForClientParameterBuilder WithExtraBodyParameters(
...
+                        data.BodyParameters.Add(param.Key, param.Value);
+                    }
+                }
+                return Task.CompletedTask;
+            });
+
+            // Convert to async dictionary for cache key component compatibility
+            var asyncParams = extrabodyparams.ToDictionary(
+                kvp => kvp.Key,
+                kvp => new Func<CancellationToken, Task<string>>(_ => Task.FromResult(kvp.Value))
```
bgavrilMS: You still have async over sync here, defeating the purpose of this improvement.

gladjohn: @bgavrilMS Should we refactor `CacheKeyComponents` to accept sync strings throughout the entire codebase, or is it acceptable to create async wrappers only once at configuration time (not on the hot path)?

bgavrilMS: It's not acceptable. If there are async cache key providers, they need to remain async. The solution is to support both sync and async versions of it.

PR #5702, opened 2026-02-02
Delete src/client/Microsoft.Identity.Client.ApiConfig directory

gladjohn:
Delete src/client/Microsoft.Identity.Client.ApiConfig directory


neha-bhargava (approved):

trwalke (approved):

trwalke:
@gladjohn this should be added to the correct public api shipped files

PR #5710, opened 2026-02-04
Fix fallback where OpenTelemetry instrumentation is not available

fowl2:
When initializing OTel, additionally catch `TypeLoadException` and `TypeInitializationException` which wrap existing catch of `FileNotFoundException` when only older version of `System.Diagnostics.DiagnosticSource` is loaded, such as on some Azure Functions versions.

Follow up to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4487

edit: I can see this was actually [already considered](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4486/changes#r1443839669) but it was decided to only include if the other exceptions were observed. I'm observing `TypeLoadException`.


bgavrilMS:
Thanks for the contribution @fowl2

neha-bhargava:
Merged. Thanks @fowl2

PR #5713, opened 2026-02-05
Set C# language version at 14

bgavrilMS:


Directory.Build.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>14</LangVersion>
```
trwalke: Seems like the build failed. I think you want tom use `14.0`
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

Issue #5755, opened 2026-02-12
Managed Identity V2: Inaccessible KeyGuard-backed certificate causes mTLS failures with no automatic recovery


gladjohn:
### Library version used

4.61.0

### .NET version

.NET 8.0.23, Windows 10.0.20348

### Scenario

ManagedIdentityClient - managed identity

## Issue details
In Managed Identity V2 scenarios, certificates are stored in the Windows certificate store and private keys are protected by KeyGuard.

If KeyGuard becomes inaccessible (due to machine restart, corruption, VBS failure), the certificate may remain while its private key reference is broken. This results in mTLS handshake failures caused by the inability to access the private key, producing errors like:

```
System.Net.Http.HttpRequestException: An error occurred while sending the request.
---> System.Net.Sockets.SocketException (10054): An existing connection was forcibly closed by the remote host.
```

Current logic only checks cert.HasPrivateKey and does not validate key accessibility before using a cached certificate. As a result, both persistent and in-memory caches may retain unusable certificates, leading to repeated failures and service disruptions with no automatic remediation.

### Relevant code snippets

```csharp
try {
    using var rsa = cert.GetRSAPrivateKey();
    byte[] data = Encoding.UTF8.GetBytes("test");
    rsa.SignData(data, HashAlgorithmName.SHA256, RSASignaturePadding.Pkcs1);
} catch {
    // On failure, delete cert from cert store and clear in-memory cache, fetch new cert
}
```

### Expected behavior

Before using cached certificates, MSAL should proactively validate the accessibility of the private key by attempting a signing operation. If validation fails, the certificate should be deleted from both persistent and in-memory caches, and a fresh certificate provisioned. Fallback to non-KeyGuard keys or retrying with exponential backoff should be avoided.

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

### Solution and workarounds

Add key accessibility checks at cache retrieval points in MtlsCertificateCache and WindowsPersistentCertificateCache. On validation failure, delete from all caches and trigger fresh certificate provisioning in managed identity scenarios.

### Error Logs

this is the error 

AL] True MSAL 4.61.0.0 MSAL.NetCore .NET 8.0.23 Microsoft Windows 10.0.20348 [2026-02-14 17:10:46Z - ffcf10a2-11ad-44f5-a544-b528876bb456] Starting [HttpManager] ExecuteAsync
[...]

PR #5761, opened 2026-02-14
IMDSv2 mTLS: Auto-recover from SCHANNEL handshake failures by evicting cached cert and re-minting

gladjohn:
Fixes https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5755

# PR Title

**IMDSv2 mTLS: Auto-recover from SCHANNEL handshake failures by evicting
cached cert and re-minting**

------------------------------------------------------------------------

# Summary

This PR fixes an intermittent but persistent failure in the IMDSv2 mTLS
PoP flow on Windows where MSAL reuses a cached/persisted mTLS binding
certificate whose private key becomes unusable (e.g., corrupted,
inaccessible, or otherwise rejected by SCHANNEL).

When this occurs, the TLS handshake fails and the token request
consistently errors until the bad certificate is manually removed from
the persistent store.

The fix introduces self-healing behavior: when we detect a
SCHANNEL-style failure during the token request, we remove the cached
mTLS binding (both in-memory and persisted store) and retry once. This
forces minting a fresh certificate/key binding and restores normal
operation.

------------------------------------------------------------------------

# Problem / Issue Description

In the IMDSv2 mTLS PoP flow, MSAL caches the mTLS binding (certificate +
endpoint + canonical `client_id`) in:

-   A process-local in-memory cache\
-   A best-effort Windows persistent store (`CurrentUser\My`) via
    `FriendlyName` tagging

In rare cases, the persisted certificate entry can appear valid (e.g.,
`HasPrivateKey == true`, unexpired) but still fail during TLS handshake
due to SCHANNEL rejecting the client credential.

This typically manifests as a transport-layer failure during the token
request:

    System.Net.Sockets.SocketException (10054):
    An existing connection was forcibly closed by the remote host

Wrapped as:

    MsalServiceException
    ErrorCode: managed_identity_unreachable_network

Because the cached binding is reused on subsequent requests, the failure
becomes **sticky**: every token acquisition attempt reuses the same bad
certificate and fails repeatedly until the certificate is manually
removed from the store.

------------------------------------------------------------------------

# Root Cause

The persistent cache selection logic previously treated a certificate as
usable if it:

-   Matched the alias via `FriendlyName`
-   Had sufficient remaining lifetime
-   Reported `HasPrivateKey == true`

However, on Windows this is not sufficient.

`HasPrivateKey` does **not** guarantee the private key is actually
usable by SCHANNEL. The private key can be:

-   Missing
-   Inaccessible (ACL issues)
-   Provider/KSP corrupted
-   Container mismatch
-   Hardware-backed state invalid after reboot

When SCHANNEL attempts the client-auth signature, the connection is
reset and the HTTP request fails.

Static validation cannot reliably predict this failure.

------------------------------------------------------------------------

# Fix

## 1) Detect SCHANNEL-style mTLS handshake failures

In `ImdsV2ManagedIdentitySource.AuthenticateAsync`, catch
`MsalServiceException` where:

-   `ErrorCode == managed_identity_unreachable_network`
-   The exception chain indicates SCHANNEL-style failure\
    (e.g., `SocketException 10054`, "forcibly closed by the remote
    host")

## 2) Evict the bad binding from caches

Remove the binding from:

-   The in-memory cache\
-   The Windows persistent store (delete all certs for that alias)

## 3) Retry once

Immediately retry `base.AuthenticateAsync(...)`.

Since the cached entry was removed, the flow re-mints a fresh
certificate and proceeds successfully.

Retry is strictly limited to **one attempt**.

------------------------------------------------------------------------

# Key Code Changes

## `ImdsV2ManagedIdentitySource`

-   Added SCHANNEL failure detection.
-   Added one-time retry logic with cache eviction.

## `MtlsBindingCache`

-   Added `RemoveBadCert(cacheKey, logger)` to remove from:
    -   In-memory cache
    -   Persistent store

## `WindowsPersistentCertificateCache`

-   Added `DeleteAllForAlias(alias, logger)`\
    Deletes all tagged certs for the alias (not just expired ones),
    enabling a full reset of a corrupted binding set.

------------------------------------------------------------------------

# Behavioral Impact

✅ No change for normal success paths.

✅ When SCHANNEL rejects a cached certificate, the system recovers
automatically without manual cleanup.

✅ Cache eviction is scoped to the affected alias (including attestation
tag), so unrelated bindings are not impacted.

⚠️ Retry is limited to one attempt to avoid masking genuine network
failures or causing repeated mint loops.

------------------------------------------------------------------------

# Example Failure This Fix Addresses

Typical failure:

    MsalServiceException: managed_identity_unreachable_network

Inner exceptions:

    HttpRequestException
    IOException: ... forcibly closed by the remote host
    SocketException (10054)

After this PR:

1.  The failure is detected as a SCHANNEL-style client cert rejection.
2.  The cached/persisted certificate is purged.
3.  The request is retried.
4.  A fresh mTLS binding is minted.
5.  Token acquisition succeeds.

## Testing
unit tests have been added

### DIFF
--- a/build/template-run-mi-e2e-imdsv2.yaml
+++ b/build/template-run-mi-e2e-imdsv2.yaml
@@ -6,10 +6,10 @@ steps:
 
 # Ensure the exact SDK is available on the agent
 - task: UseDotNet@2
-  displayName: 'Install .NET SDK 8.0.415'
+  displayName: 'Install .NET SDK 8.0.418'
   inputs:
     packageType: 'sdk'
-    version: '8.0.415'
+    version: '8.0.418'
     includePreviewVersions: false
     performMultiLevelLookup: true
 


src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -175,7 +175,58 @@ public override async Task<ManagedIdentityResponse> AuthenticateAsync(
...
+                try
+                {
+                    if (_mtlsCache is MtlsBindingCache mtlsCache)
+                    {
+                        mtlsCache.RemoveBadCert(certCacheKey, _requestContext.Logger);
+                    }
+                }
+                catch (Exception removalEx)
+                {
+                    _requestContext.Logger.Verbose(() => $"[ImdsV2] Error removing bad cert: {removalEx.Message}");
```
bgavrilMS, line 199: Shouldn't this be a fatal error?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
```diff
@@ -143,5 +143,31 @@ public async Task<MtlsBindingInfo> GetOrCreateAsync(
...
+        /// Removes a certificate from both in-memory and persistent cache when SCHANNEL rejects it.
+        /// </summary>
+        public void RemoveBadCert(string cacheKey, ILoggerAdapter logger)
+        {
+            try
+            {
+                _memory.Remove(cacheKey, logger);
+                logger?.Verbose(() => $"[PersistentCert] Removed bad cert from memory cache for '{cacheKey}'");
+            }
+            catch (Exception ex)
```
bgavrilMS, line 173: Shouldn't this be fatal? When do you expect to reach this code and recover?

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IPersistentCertificateCache.cs
```diff
@@ -32,5 +32,13 @@ internal interface IPersistentCertificateCache
...
         /// </summary>
         void Delete(string alias, ILoggerAdapter logger);
+
+        /// <summary>
+        /// Deletes all entries for the alias (best-effort).
+        /// Implementations should remove all bindings for the alias.
+        /// </summary>
+        /// <param name="alias"></param>
+        /// <param name="logger"></param>
+        void DeleteAllForAlias(string alias, ILoggerAdapter logger);
```
bgavrilMS, line 45: I don't undertsand the difference between `Delete` and `DeleteAll`

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
```diff
@@ -270,6 +270,63 @@ public void Delete(string alias, ILoggerAdapter logger)
...
+                                removed++;
+                                logger?.Verbose(() => $"[PersistentCert] Deleted certificate from store for alias '{alias}'");
+                            }
+                            finally
+                            {
+                                existing.Dispose();
+                            }
+                        }
+
+                        logger?.Verbose(() => $"[PersistentCert] DeleteAllForAlias completed for alias '{alias}'. Removed={removed}.");
```
bgavrilMS: Not needed. You already have logging statements to delete this.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
```diff
@@ -143,5 +143,31 @@ public async Task<MtlsBindingInfo> GetOrCreateAsync(
...
+            }
+
+            try
+            {
+                _persisted.DeleteAllForAlias(cacheKey, logger);
+                logger?.Verbose(() => $"[PersistentCert] Removed bad cert from persistent cache for '{cacheKey}'");
+            }
+            catch (Exception ex)
+            {
+                logger?.Verbose(() => $"[PersistentCert] Error removing from persistent cache: {ex.Message}");
```
trwalke, line 185: Should probably also be fatal exception?

trwalke (approved):

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
gladjohn: Cache eviction is best-effort by design; we don’t want store/lock issues to become the user-visible failure. We changed the retry path to force a one-time re-mint after RemoveBadCert, so we won’t reuse the persisted cert even if delete fails.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/WindowsPersistentCertificateCache.cs
gladjohn: removed

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/MtlsCertificateCache.cs
gladjohn: Cache eviction is best-effort by design; we don’t want store/lock issues to become the user-visible failure. We changed the retry path to force a one-time re-mint after RemoveBadCert, so we won’t reuse the persisted cert even if delete fails.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
gladjohn: Cache eviction is best-effort by design; we don’t want store/lock issues to become the user-visible failure. We changed the retry path to force a one-time re-mint after RemoveBadCert, so we won’t reuse the persisted cert even if delete fails.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/IPersistentCertificateCache.cs
gladjohn: added some more code comments. 

Delete(alias, …) is the maintenance/prune operation: it removes only expired certs for the alias to keep the Windows store bounded, while intentionally keeping the newest non-expired binding so normal cache hits still work. We call this after Write() as cleanup.

DeleteAllForAlias(alias, …) is the eviction/reset operation: it removes all certs for the alias (expired or not). This is only used on the SCHANNEL/mTLS failure path to ensure we don’t reuse a potentially bad cert and can re-mint a fresh one.

Both are best-effort and non-blocking (failures are logged and ignored).

PR #5764, opened 2026-02-19
Ensuring ValidateCachedTokenAsync works for other apis

trwalke:
This pull request introduces a unified approach for validating cached access tokens before use, improving consistency and reliability across authentication flows. The main change is the addition of a new helper method, `ValidateCachedAccessTokenAsync`, which centralizes the logic for validating cached tokens using the authentication operation (when supported). This method is now used in multiple request types, and comprehensive unit tests have been added to ensure correct behavior.

Token validation improvements:

* Added the `ValidateCachedAccessTokenAsync` method in `RequestBase.cs` to centralize and streamline validation of cached access tokens using `IAuthenticationOperation2`. The method returns the original cache item if validation passes, or null if validation fails.
* Updated `ClientCredentialRequest`, `OnBehalfOfRequest`, and `CacheSilentStrategy` to use the new helper method for validating cached access tokens, replacing previous inline validation logic. [[1]](diffhunk://#diff-655d5868568553085143b1c4df713cb7baf07c35ab78610a8ac73da1e467ce43L75-R75) [[2]](diffhunk://#diff-bf467b9439792137a4965d3a9b52423cf088346687d966681557d134a6ef1b10R74-R76) [[3]](diffhunk://#diff-04f82f33c9cf802268c7c48a869ca80b26839db606a0d729bd739887c7593bfaR51-R53)

Testing enhancements:

* Added unit tests in `AuthenticationOperationTests.cs` to verify that cached tokens are properly validated and either used or ignored based on the result of `ValidateCachedTokenAsync`. Tests cover both On-Behalf-Of and silent authentication flows, for both validation success and failure scenarios.

Code cleanup:

* Removed unnecessary `using Microsoft.Identity.Client.AuthScheme;` directive from `ClientCredentialRequest.cs` as part of refactoring.Ensuring ValidateCachedTokenAsync works for other apis

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,6 +9,7 @@
   <!-- Add packages, then highlight all lines and go to Edit > Advanced > Sort Lines -->
   <ItemGroup>
     <!-- Packages for the main MSAL code -->
+    <PackageVersion Include="Microsoft.Azure.Security.KeyGuardAttestation" Version="1.1.5" />
     <PackageVersion Include="Microsoft.CodeAnalysis.RulesetToEditorconfigConverter" Version="3.3.3" PrivateAssets="All" />
     <PackageVersion Include="Microsoft.CSharp" Version="4.5.0" />
     <PackageVersion Include="Microsoft.Identity.Client.NativeInterop" Version="$(MSALRuntimeNativeInteropVersion)" IncludeAssets="all" />
--- a/build/template-build-and-prep-automation.yaml
+++ b/build/template-build-and-prep-automation.yaml
@@ -53,6 +53,13 @@ steps:
     Contents: '**/bin/**/**'
     TargetFolder: '$(Build.SourcesDirectory)/WAM'
 
+- task: CopyFiles@2
+  displayName: 'Copy Managed Identity DevApp Dlls'
+  inputs:
+    SourceFolder: 'tests\devapps\Managed Identity apps\ManagedIdentityAppVM'
+    Contents: '**\bin\**\*'
+    TargetFolder: '$(Build.SourcesDirectory)/ManagedIdentityAppVM'
+
 - task: PublishPipelineArtifact@1
   displayName: 'Publish Artifact: drop'
   inputs:
@@ -112,6 +119,16 @@ steps:
             $errors += "Failed to find Linux runtime files in : " + $SearchInFolder + $app + "`n";
         }
 
+        # Managed Identity app runtime DLL check
+        $managedIdentityFolder = Join-Path $(Build.SourcesDirectory) 'ManagedIdentityAppVM'
+        $managedIdentityDll = Join-Path $managedIdentityFolder 'bin\Release\net8.0\runtimes\win-x64\native\AttestationClientLib.dll'
+        
+        Write-Host "Checking: $managedIdentityDll"
+        if (-not (Test-Path -Path $managedIdentityDll -PathType Leaf))
+        {
+            $errors += "Failed to find AttestationClientLib.dll at: $managedIdentityDll`n"
+        }
+
         #exit if there are errors
         IF (![string]::IsNullOrWhitespace($errors))
         {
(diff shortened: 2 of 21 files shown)


src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -570,5 +570,30 @@ private static RegionDetails CreateRegionDetails(ApiEvent apiEvent)
...
+        /// </summary>
+        internal static async Task<MsalAccessTokenCacheItem> ValidateCachedAccessTokenAsync(
+            AuthenticationRequestParameters authenticationRequestParameters,
+            MsalAccessTokenCacheItem cachedAccessTokenItem)
+        {
+            if (cachedAccessTokenItem != null &&
+                authenticationRequestParameters.AuthenticationScheme is IAuthenticationOperation2 authOp2)
+            {
+                var cacheValidationData = new MsalCacheValidationData();
+                cacheValidationData.PersistedCacheParameters = cachedAccessTokenItem.PersistedCacheParameters;
```
bgavrilMS, line 587: Isn't there some cancellation token that should be passed around?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -570,5 +570,30 @@ private static RegionDetails CreateRegionDetails(ApiEvent apiEvent)
...
+
+                if (!await authOp2.ValidateCachedTokenAsync(cacheValidationData).ConfigureAwait(false))
+                {
+                    authenticationRequestParameters.RequestContext.Logger.Info(
```
gladjohn, line 591: The centralized helper now logs `Cached token failed authentication operation validation.` generically, but previously the client-credential path logged with request-specific context like [ClientCredentialRequest] ....

For debugging and supportability, it will help to know which request type / flow this came from. 

Pass a caller name string into the helper (e.g., nameof(ClientCredentialRequest)) and include it in the log message.

Not a blocker, just makes troubleshooting easier.

tests/Microsoft.Identity.Test.Unit/PublicApiTests/AuthenticationOperationTests.cs
```diff
@@ -460,10 +460,209 @@ public async Task ValidateCachedTokenAsync_WithNullCachedItem_ValidationNotCalle
...
+                // Silent request - validation succeeds, cached token should be returned
+                var result = await cca.AcquireTokenSilent(TestConstants.s_scope, account)
+                    .WithAuthenticationOperation(authScheme)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+
+                // Assert - validation was called and succeeded, cached token was returned
+                Assert.AreEqual(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+                await authScheme.Received(1).ValidateCachedTokenAsync(Arg.Any<MsalCacheValidationData>())
+                    .ConfigureAwait(false);
```
gladjohn, line 664: Assert.AreEqual(0, httpManager.QueueSize) ?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
gladjohn, line 587: does this come from an extensibility API where external code has access to sensitive cache data? can the cache be manipulated? is that of any concern? just for my understanding.

trwalke: Added

trwalke: Yes, this is coming from an extension. That extension is `IAuthenticationOperation2`, which already has access to the AuthenticationResult from the authentication request/response. The cache cannot be manipulated in this case, this method simply allows you to reject a token in the cache if it does not meet certain criteria when MSAL does a cache look up, forcing MSAL to refresh the token instead.

gladjohn (approved):

PR #5815, opened 2026-03-06
MSAL MI: restore classic fallback behavior unless GetManagedIdentitySourceAsync() is explicitly invoked

### LINKED ISSUE #5812: MSAL MI: restore classic fallback behavior unless GetManagedIdentitySourceAsync() is explicitly invoked
### Task type

Development

### Description

# Managed Identity Source Discovery: Make `GetManagedIdentitySourceAsync()` Explicit

## Problem

Today, `GetManagedIdentitySourceAsync()` is used both internally and as a public API.

When `GetManagedIdentitySourceAsync()` runs, it does the following:

1. Check environment-variable-based sources first (Service Fabric, App Service, Azure Arc, etc.)
2. If an environment-based source is found, use it
3. If no environment-based source is found, probe IMDS v1
4. If IMDS v1 is not found, probe IMDS v2
5. If neither is found, return `NoneFound`

Because this discovery flow runs in the default path today, higher-level SDKs can get `NoneFound` instead of following the classic fallback behavior of defaulting to IMDS.

---

## Goal

Change the default token acquisition behavior so that `GetManagedIdentitySourceAsync()` becomes an **explicit discovery API**, not part of the standard managed identity token flow.

---

### 1) Standard token acquisition path (when `GetManagedIdentitySourceAsync()` is **not** called)

Use the classic behavior:

1. Check environment-variable-based sources first (Service Fabric, App Service, Azure Arc, etc.)
2. If an environment-based source is found, use it
3. Otherwise default to IMDS
4. **Do not probe** IMDS v1 or IMDS v2

This preserves backward compatibility for existing callers and avoids returning `NoneFound` in the default flow.

---

### 2) Explicit discovery path (when `GetManagedIdentitySourceAsync()` **is** called)

When `GetManagedIdentitySourceAsync()` is explicitly called:

1. Check environment-variable-based sources first (Service Fabric, App Service, Azure Arc, etc.)
2. If an environment-based source is found, use it
3. If no environment-based source is found, probe IMDS v1
4. If IMDS v1 is not found, probe IMDS v2
5. If neither IMDS v1 nor IMDS v2 is found, return `NoneFound`
6. Cache the discovered source
7. All subsequent token requests use the cached source

---

### 3) IMDS v2-specific request path (`WithMtlsProofOfPossession()` is used)

When `WithMtlsProofOfPossession()` is used on the request:

1. Do not perform discovery
2. Do not probe
3. Go directly to IMDS v2 endpoints

---

## Required changes

- Stop calling `GetManagedIdentitySourceAsync()` internally from the default managed identity token acquisition path
- Preserve the current discovery behavior for callers that explicitly invoke `GetManagedIdentitySourceAsync()`
- Ensure explicit discovery caches the resolved managed identity source
- Ensure subsequent token requests use the cached source after explicit discovery
- Ensure requests using `WithMtlsProofOfPossession()` go directly to IMDS v2
- Preserve existing source precedence for environment-based managed identity sources

---

### Standard managed identity token requests (do **not** call `GetManagedIdentitySourceAsync()`)

- Check environment-variable-based sources first
- Use the environment-based source when present
- Otherwise default to IMDS
- Do not probe IMDS v1 or IMDS v2

### Explicit calls to `GetManagedIdentitySourceAsync()`

- Check environment-variable-based sources first
- Probe IMDS v1 when no environment-based source is found
- Probe IMDS v2 if IMDS v1 is not found
- Return `NoneFound` if neither IMDS v1 nor IMDS v2 is found
- Cache the discovered source

### After explicit discovery

- Subsequent token requests use the cached source

### Requests using `WithMtlsProofOfPossession()`

- Skip discovery
- Skip probing
- Use IMDS v2 endpoints directly

---

## Notes

- The key change is that **probing should only happen when** `GetManagedIdentitySourceAsync()` **is explicitly called**
- The default path should go back to the classic behavior to avoid breaking higher-level SDKs
- `WithMtlsProofOfPossession()` should continue to be treated as an IMDS v2-specific path

### Solution

_No response_

gladjohn:
Fixes #5812 #5651

This pull request refactors and improves the managed identity source discovery and caching logic in the MSAL library. The changes clarify the distinction between explicit source discovery and implicit selection, ensure correct caching of discovery results (including "NoneFound"), and improve error handling for IMDSv2 scenarios. Several tests are updated to reflect the new discovery order and caching behavior. The most important changes are grouped below by theme.

Managed Identity Source Discovery & Caching

* Introduced explicit caching for managed identity source discovery results, including handling for "NoneFound" and separation from implicit source selection. Added `s_isSourceDiscoveryCached` and `s_cachedSourceResult` fields and updated logic in `ManagedIdentityClient` to use them. (`src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs`) [[1]](diffhunk://#diff-13058fddef4867b4d36dca3120b5b961ae64b039b32bcf5ad410a364b36fbfa8R28-R31) [[2]](diffhunk://#diff-13058fddef4867b4d36dca3120b5b961ae64b039b32bcf5ad410a364b36fbfa8R40-R41)
* Refactored `GetOrSelectManagedIdentitySourceAsync` to use cached discovery results when available, otherwise select source based on environment variables and default to IMDS without probing. Explicit discovery now probes IMDSv1 first, then IMDSv2, and caches the result. (`src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs`) [[1]](diffhunk://#diff-13058fddef4867b4d36dca3120b5b961ae64b039b32bcf5ad410a364b36fbfa8L51-R112) [[2]](diffhunk://#diff-13058fddef4867b4d36dca3120b5b961ae64b039b32bcf5ad410a364b36fbfa8L109-R135) [[3]](diffhunk://#diff-13058fddef4867b4d36dca3120b5b961ae64b039b32bcf5ad410a364b36fbfa8L118-R203)
* Updated `ManagedIdentityApplication.GetManagedIdentitySourceAsync` to use the new caching mechanism and removed unnecessary parameters. (`src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs`)

IMDSv2 Error Handling

* Improved error handling for IMDSv2 CSR endpoint: a 404 from the endpoint now throws a client exception indicating that mTLS PoP is not supported on IMDSv1-only hosts, providing clearer feedback to callers. (`src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs`)

Test Updates

* Updated unit and E2E tests to reflect new discovery logic, including correct probe order (IMDSv1 first), removal of unnecessary environment variable checks, and validation of cached discovery results. (`tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs`, `tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsV2Tests.cs`) [[1]](diffhunk://#diff-5fd48f0c03e6b55dc8c9e6782bfa5bb68367d5502151dab256a16520095f3e10L124-R140) [[2]](diffhunk://#diff-5b37959521a695f75a810a8ec471d2ae75924ea454268b81daed580f2a46bcb0L71-L77)
* Removed redundant IMDSv1 probe mocks from unit tests to align with the new discovery and caching flow. (`tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsTests.cs`) [[1]](diffhunk://#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154L46) [[2]](diffhunk://#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154L104) [[3]](diffhunk://#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154L162) [[4]](diffhunk://#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154L217) [[5]](diffhunk://#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154L272) [[6]](diffhunk://#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154L323) [[7]](diffhunk://#diff-057b5a5346431bf0d06e027a642e673ca6292a88765cc00b0e7603dec26b5154L368)

These changes make the managed identity source selection more robust, predictable, and easier to reason about, and improve developer experience when using mTLS PoP and IMDSv2.

**Testing**
unit, integration, and e2e

**Performance impact**
none

**Documentation**

### DIFF
--- a/prototype/msal-analyzer/.gitignore
+++ b/prototype/msal-analyzer/.gitignore
@@ -1,0 +1,33 @@
+# Dependencies
+node_modules/
+client/node_modules/
+server/node_modules/
+
+# Build output
+client/build/
+client/dist/
+
+# Environment files
+server/.env
+.env
+
+# Logs
+*.log
+npm-debug.log*
+
+# OS files
+.DS_Store
+Thumbs.db
+
+# Editor files
+.vscode/
+.idea/
+*.swp
+*.swo
+
+# Upload temp files
+server/uploads/
+server/reports/
+
+# Coverage
+coverage/
--- a/prototype/msal-analyzer/README.md
+++ b/prototype/msal-analyzer/README.md
@@ -1,0 +1,155 @@
+# MSAL Log Analyzer
+
+A prototype web application for analyzing Microsoft Authentication Library (MSAL) logs with AI-powered insights, interactive visualizations, and reporting.
+
+> **Important:** This application uses **Azure OpenAI** (a Microsoft-approved AI service) to analyze MSAL logs. Third-party AI services such as Anthropic Claude or OpenAI.com must **not** be used, because MSAL logs contain Microsoft business-sensitive data (tenant IDs, client IDs, account identifiers, correlation IDs, token metadata) that may not be transferred to non-Microsoft AI services under Microsoft policy.
+
+## Features
+
+- 📤 **Drag-and-drop log upload** - Easy file upload with real-time progress
+- 🤖 **AI-powered analysis** - Azure OpenAI integration for intelligent log parsing
+- 📊 **Performance metrics** - Total duration, HTTP calls, cache hits, token source
+- 🔄 **Interactive flow diagrams** - Mermaid-based module interaction visualizations
+- 📋 **Module details** - Tabular view of all modules with timings and status
+- 💡 **Key findings** - Automated insights and recommendations
+- 📄 **HTML export** - Download beautiful reports for sharing
+- 🌓 **Dark/Light theme** - Comfortable viewing in any environment
+- 📱 **Mobile responsive** - Works on all screen sizes
+
+## Tech Stack
+
+### Frontend
+- React 18
+- Tailwind CSS
+- Mermaid.js (diagrams)
+- React Dropzone (file upload)
+
+### Backend
+- Express.js
+- Azure OpenAI (Microsoft-approved AI service)
+- Multer (file handling)
+- Helmet (security headers)
+- CORS
+
+## Quick Start
+
+### Prerequisites
+- Node.js 18+
+- npm 9+
+- An [Azure OpenAI resource](https://portal.azure.com/#create/Microsoft.CognitiveServicesOpenAI) with a deployed model (e.g. `gpt-4o`)
+
+### Installation
+
+```bash
+# Clone the repository
+git clone <repository-url>
+cd msal-analyzer
+
+# Install all dependencies
+npm run install:all
+
+# Configure environment
+cp server/.env.example server/.env
+# Edit server/.env and fill in your Azure OpenAI credentials
+```
+
+### Development
+
+```bash
+# Start both frontend and backend in development mode
+npm run dev
+```
+
+- Frontend: http://localhost:3000
+- Backend API: http://localhost:3001
+
+### Production
+
+```bash
+# Build the frontend
+npm run build
+
+# Start the server (serves frontend + API)
+npm start
+```
+
+## Docker
+
+### Using Docker Compose
+
+```bash
+# Copy and configure environment
+cp server/.env.example server/.env
+# Edit server/.env and fill in your Azure OpenAI credentials
+
+# Build and start
+docker-compose -f docker/docker-compose.yml up --build
+
+# Access the app at http://localhost:3001
+```
+
+### Using Docker directly
+
+```bash
+docker build -f docker/Dockerfile -t msal-analyzer .
+docker run -p 3001:3001 \
+  -e AZURE_OPENAI_ENDPOINT=https://<your-resource>.openai.azure.com/ \
+  -e AZURE_OPENAI_API_KEY=your_key \
+  -e AZURE_OPENAI_DEPLOYMENT=gpt-4o \
+  msal-analyzer
+```
+
+## Configuration
+
+| Variable | Description | Default |
+|----------|-------------|---------|
+| `AZURE_OPENAI_ENDPOINT` | Azure OpenAI resource endpoint URL | Required for AI |
+| `AZURE_OPENAI_API_KEY` | Azure OpenAI API key | Required for AI |
+| `AZURE_OPENAI_DEPLOYMENT` | Model deployment name | `gpt-4o` |
+| `AZURE_OPENAI_API_VERSION` | Azure OpenAI API version | `2024-10-21` |
+| `PORT` | Server port | `3001` |
+| `NODE_ENV` | Environment | `development` |
+| `MAX_FILE_SIZE` | Max upload size in bytes | `10485760` (10MB) |
+
+> **Note:** Azure OpenAI credentials are optional — the application falls back to rule-based regex parsing when they are not configured.
+
+## Usage
+
+1. Open the application in your browser
+2. Drag and drop an MSAL log file (or click to browse)
+3. Wait for AI analysis to complete
+4. Explore the dashboard with:
+   - Performance metrics overview
+   - Module interaction flow diagram
+   - Module details table
+   - Key findings and recommendations
+5. Export the report as HTML for sharing
+
+## API Reference
+
+### POST /api/analyze
+Analyzes an MSAL log file.
+
+**Request:** `multipart/form-data`
+- `logFile`: The MSAL log file to analyze
+
+**Response:**
+```json
+{
+  "success": true,
+  "report": {
+    "summary": { ... },
+    "modules": [ ... ],
+    "mermaidDiagram": "...",
+    "findings": [ ... ],
+    "rawInsights": "..."
+  }
+}
+```
+
+### GET /api/health
+Health check endpoint.
+
+## License
+
+MIT
--- a/prototype/msal-analyzer/client/index.html
+++ b/prototype/msal-analyzer/client/index.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="theme-color" content="#0f172a" />
+    <meta
+      name="description"
+      content="MSAL Log Analyzer - Intelligent Microsoft Authentication Library log analysis with AI-powered insights"
+    />
+    <title>MSAL Log Analyzer</title>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/index.jsx"></script>
+  </body>
+</html>
--- a/prototype/msal-analyzer/client/package.json
+++ b/prototype/msal-analyzer/client/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "msal-analyzer-client",
+  "version": "1.0.0",
+  "private": true,
+  "dependencies": {
+    "axios": "^1.6.2",
+    "mermaid": "^10.6.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-dropzone": "^14.2.3"
+  },
+  "devDependencies": {
+    "@types/react": "^18.2.0",
+    "@types/react-dom": "^18.2.0",
+    "@vitejs/plugin-react": "^4.2.1",
+    "autoprefixer": "^10.4.16",
+    "postcss": "^8.4.32",
+    "tailwindcss": "^3.4.0",
+    "vite": "^5.0.0"
+  },
+  "scripts": {
+    "start": "vite --port 3000",
+    "build": "vite build",
+    "preview": "vite preview"
+  }
+}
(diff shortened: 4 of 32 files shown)


src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -25,6 +25,10 @@ internal class ManagedIdentityClient
         // Preview guard: once we fall back to IMDSv1 while IMDSv2 is cached,
         // disallow switching to IMDSv2 PoP in the same process (preview behavior).
         internal static bool s_imdsV1UsedForPreview = false;
+        // True only after the explicit discovery API (GetManagedIdentitySourceAsync) runs.
+        // This allows caching "NoneFound" without confusing it with "not discovered yet".
+        internal static bool s_isSourceDiscoveryCached = false;
```
bgavrilMS: Do we need another flag here? Can't s_cachedSourceResult  = null represent "not cached" ?

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -25,6 +25,10 @@ internal class ManagedIdentityClient
...
+        internal static ManagedIdentitySourceResult s_cachedSourceResult = null;
```
bgavrilMS: This class has too much surface. Ideally it should have 1-2 methods - getToken and getSource. Can you not expose internal details about how the cache is handled to ManagedIdentityApplication?

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
```diff
@@ -58,16 +58,15 @@ public AcquireTokenForManagedIdentityParameterBuilder AcquireTokenForManagedIden
         /// <inheritdoc/>
         public async Task<ManagedIdentitySourceResult> GetManagedIdentitySourceAsync(CancellationToken cancellationToken)
         {
-            if (ManagedIdentityClient.s_sourceName != ManagedIdentitySource.None)
+            if (ManagedIdentityClient.s_isSourceDiscoveryCached && ManagedIdentityClient.s_cachedSourceResult != null)
```
bgavrilMS: The details of caching bleed into this class. But it seems they can be kept in ManagedIdentityClient

bgavrilMS (changes requested):
Single responsability principle

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: Good catch — agreed. I removed `s_isSourceDiscoveryCached`  and now use `s_cachedSourceResult == null` to represent “not discovered yet”.

We still cache “NoneFound” by storing a non-null `ManagedIdentitySourceResult` with `Source=None` (and failure reasons), so we can distinguish “explicit discovery ran and found nothing” from “discovery never ran”. ResetSourceForTest clears `s_cachedSourceResult` back to null.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: removed the caching checks from MIA so it no longer reads `ManagedIdentityClient` cache internals. It now always delegates to `ManagedIdentityClient.GetManagedIdentitySourceAsync(...)`, which fully owns caching.

src/client/Microsoft.Identity.Client/ManagedIdentityApplication.cs
gladjohn: fixed. MIA no longer checks any cached fields. It now just calls `ManagedIdentityClient.GetManagedIdentitySourceAsync`, and the client handles caching internally.

gladjohn:
> Single responsability principle

Implemented the single responsibility feedback in db1dd6f (“single responsibility”):
- Removed the redundant `s_isSourceDiscoveryCached` flag (use `s_cachedSourceResult == null` sentinel instead)
- Made `s_cachedSourceResult` private
- Removed cache-handling logic from ManagedIdentityApplication (it now delegates to ManagedIdentityClient)

This keeps discovery/caching contained to ManagedIdentityClient and avoids leaking cache mechanics into the application layer.

src/client/Microsoft.Identity.Client/ManagedIdentity/V2/ImdsV2ManagedIdentitySource.cs
```diff
@@ -96,6 +96,18 @@ private static void ThrowCsrMetadataRequestException(
...
+            // A 404 from the IMDSv2 CSR endpoint indicates that the host supports only IMDSv1.
+            // This happens when WithMtlsProofOfPossession() is used without a prior
+            // GetManagedIdentitySourceAsync() call: MSAL routes directly to IMDSv2, and
+            // on an IMDSv1-only host the /getplatformmetadata endpoint does not exist.
+            // Translate to a client error so callers know mTLS PoP is not supported here.
+            if (statusCode == (int)HttpStatusCode.NotFound)
+            {
+                throw new MsalClientException(
+                    MsalError.MtlsPopTokenNotSupportedinImdsV1,
+                    MsalErrorMessage.MtlsPopTokenNotSupportedinImdsV1);
```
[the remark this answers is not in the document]
gladjohn: public const string MtlsPopTokenNotSupportedinImdsV1 = "A mTLS PoP token cannot be requested because the application\'s source was determined to be ImdsV1.";

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -115,63 +145,66 @@ private async Task<AbstractManagedIdentity> GetOrSelectManagedIdentitySourceAsyn
...
+                    _ => throw CreateManagedIdentityUnavailableException(s_cachedSourceResult)
+                });
             }
         }
 
-        // Detect managed identity source based on the availability of environment variables and csr metadata probe request.
-        // This method is perf sensitive any changes should be benchmarked.
+        private static ManagedIdentitySourceResult CacheDiscoveryResult(ManagedIdentitySourceResult result)
+        {
+            s_sourceName = result.Source;
```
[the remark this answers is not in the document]
gladjohn: Yes, `s_sourceName` is now redundant and implies a second cache.

I removed `s_sourceName` and the write in `CacheDiscoveryResult`; the only cached discovery signal is now `s_cachedSourceResult != null` (including caching NoneFound).

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -48,42 +51,68 @@ internal async Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityA
...
             bool isMtlsPopRequested,
             CancellationToken cancellationToken)
         {
             using (requestContext.Logger.LogMethodDuration())
             {
-                requestContext.Logger.Info($"[Managed Identity] Selecting managed identity source if not cached. Cached value is {s_sourceName} ");
+                requestContext.Logger.Info($"[Managed Identity] Selecting managed identity source. " + 
+                    $"Discovery cached: {s_cachedSourceResult != null}");
+
+                // Check for cancellation before potentially expensive operations
```
bgavrilMS: To many comments. Focus on commenting the "WHY", not the why. `cancellationToken.ThrowIfCancellationRequested` already encapsulates all the logic in your comment.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
```diff
@@ -48,42 +51,68 @@ internal async Task<ManagedIdentityResponse> SendTokenRequestForManagedIdentityA
...
+                // (e.g., network calls in token acquisition) to fail fast if the
+                // caller has already requested cancellation.
+                cancellationToken.ThrowIfCancellationRequested();
```
bgavrilMS, line 68: Best practice would be to use this before expensive operations that do not allow a CancellationToken themselves. Not sure if it's the case here.

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: Got it. Removed the redundant comments. cancellationToken.ThrowIfCancellationRequested() is clear on its own

src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs
gladjohn: Yes, this check is needed. The code below makes network calls (IMDSv1/v2 probes) and caches the results. We want to fail fast if cancellation was already requested, rather than performing expensive I/O that might succeed but whose results we won't use.

Issue #5852, opened 2026-03-13
[Engineering task] Enable MtlsPop_WithUnsupportedNonTenantedAuthorityAsync_ThrowsException


gladjohn:
### Task type

Test

### Description

Enable MtlsPop_WithUnsupportedNonTenantedAuthorityAsync_ThrowsException

### Solution

Enable MtlsPop_WithUnsupportedNonTenantedAuthorityAsync_ThrowsException

PR #5823, opened 2026-03-08
MSTest v4 migration - Phase 4 - Enable mTLS test - MtlsPop_WithUnsupportedNonTenantedAuthorityAsync_ThrowsException

gladjohn:
This pull request introduces stricter validation for mTLS PoP authentication in Azure Active Directory (AAD) scenarios, ensuring that only tenanted authorities are allowed and improving test coverage for these cases. The main changes include enhanced checks in `MtlsPopParametersInitializer`, updates to unit tests to validate the new behavior, and minor improvements to exception handling.

Validation improvements for mTLS PoP:

* Added validation in `MtlsPopParametersInitializer` to ensure that mTLS PoP authentication cannot be used with non-tenanted AAD authorities (`/common`, `/organizations`, `/consumers`). If such an authority is detected, an exception is thrown with a specific error message.
* Updated the region validation logic to only occur after confirming the authority is tenanted, preventing region checks for unsupported authorities.

Test enhancements:

* Enabled a previously ignored unit test in `MtlsPopTests.cs` to verify that exceptions are thrown when mTLS PoP is used with non-tenanted authorities.
* Updated the test assertion to use `Assert.ThrowsExactlyAsync` for more precise exception checking.

Minor code improvements:

* Added a missing namespace import in `MtlsPopParametersInitializer.cs` for consistency and clarity.

PR #5853, opened 2026-03-14
Unignore mTLS PoP test for non-tenanted authority

gladjohn:
**Fixes** - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5852 

**Follow up on comment from PR** #5823 - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5823#discussion_r2932843885 

## Why we are doing this

While investigating a test failure after the MSTest4 update, we found that the issue was **not** caused by MSTest4 itself.

### Root Cause Analysis:

The issue happened because two MSTest test methods in the same class had the exact same name, with one being parameterless and the other parameterized. MSTest test discovery did not handle the overloaded test names correctly, so it picked up the wrong test during execution, which made it look like the parameterized test was passing when it was not actually being run as expected.

### Fix:

Renamed the test methods so each test has a unique name and removed the naming collision, ensuring MSTest discovers and executes the correct test cases consistently.

## The real problem was that the test expected MSAL to throw:

- `MsalError.MissingTenantedAuthority`

for explicit mTLS PoP requests using non-tenanted AAD authorities like:

- `/common`
- `/organizations`
- `/consumers`

But there was **no product code** that actually performed that validation and threw that error. There is no product code implementing MissingTenantedAuthority for this mTLS PoP path. So even if the test appeared to pass before MSTest4, it was not validating the intended product behavior.

Because of that, the request continued deeper into execution and eventually failed in the test infrastructure (`MockHttpManager`) instead of failing early with the intended MSAL exception.

## What we found

By searching through the codebase, we found:

- `MsalError.MissingTenantedAuthority` exists in `MsalError.cs`
- it is part of the shipped public API
- the unit test was asserting this error
- but no product logic was throwing it for this mTLS PoP scenario

So there was a mismatch between:

- the public error surface
- the unit test expectation
- the actual product behavior

## What this change does

This change adds `MtlsPopParametersInitializer` to centralize mTLS/PoP initialization and validation logic.

It does a few important things:

- adds the missing early validation for explicit mTLS PoP with non-tenanted AAD authorities
- throws `MissingTenantedAuthority` before any HTTP call is made
- keeps region validation for AAD mTLS flows in one place
- separates explicit mTLS PoP setup from implicit bearer-over-mTLS behavior
- makes the logic easier to test and maintain

## Why this is better

With this change:

- invalid mTLS PoP requests fail early
- the product throws the expected MSAL exception
- tests no longer fail indirectly because of empty mock HTTP queues
- the behavior now matches the public API and test intent

## Summary

This is a product correctness fix that was uncovered during the MSTest4 migration investigation.

MSTest4 only exposed the issue. The real gap was that MSAL had a public error and a unit test for this case, but no product code enforcing it.


tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
```diff
@@ -168,10 +168,11 @@ public async Task MtlsPop_WithoutRegion_ThrowsException(bool setAzureRegion)
         }
 
         [TestMethod]
-        public async Task MtlsPop_WithUnsupportedNonTenantedAuthorityAsync_ThrowsException()
+        public async Task MtlsPop_WithoutRegion_ThrowsException()
```
gladjohn: wrong test name - the test was checking for a missing region from the builder

tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs
```diff
@@ -168,10 +168,11 @@ public async Task MtlsPop_WithoutRegion_ThrowsException(bool setAzureRegion)
...
         {
             IConfidentialClientApplication app = ConfidentialClientApplicationBuilder
                             .Create(TestConstants.ClientId)
```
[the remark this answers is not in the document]
gladjohn: dupe tests removed

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/MtlsPopParametersInitializer.cs
```diff
@@ -134,13 +135,23 @@ private static void InitMtlsPopParameters(
...
-                    MsalErrorMessage.MtlsPopWithoutRegion);
+                string tenant = AuthorityInfo.GetFirstPathSegment(serviceBundle.Config.Authority.AuthorityInfo.CanonicalAuthority);
+                if (AadAuthority.IsCommonOrganizationsOrConsumersTenant(tenant))
+                {
+                    throw new MsalClientException(
+                        MsalError.MissingTenantedAuthority,
+                        MsalErrorMessage.MtlsNonTenantedAuthorityNotAllowedMessage);
+                }
+
+                if (serviceBundle.Config.AzureRegion == null)
```
bgavrilMS, line 149: I believe Entra does actually support this. There is a global mtls endpoint.

gladjohn: To keep scope tight, this change is only validating that the authority is tenanted (and rejecting common/consumer authorities).

If we want to support the global mTLS endpoint by removing the regional checks, I can follow up with a separate work item/PR that includes the code changes and tests.

gladjohn: @bgavrilMS  Could you please take another look and approve if you’re okay with keeping this PR scoped to tenanted authorities only?

I’m treating global mTLS endpoint support as follow-up work so we can keep this change focused and land the validation behavior first. https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5865

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/MtlsPopParametersInitializer.cs
```diff
@@ -134,13 +135,23 @@ private static void InitMtlsPopParameters(
...
+            if (serviceBundle.Config.Authority.AuthorityInfo.AuthorityType == AuthorityType.Aad)
             {
-                throw new MsalClientException(
-                    MsalError.MtlsPopWithoutRegion,
-                    MsalErrorMessage.MtlsPopWithoutRegion);
+                string tenant = AuthorityInfo.GetFirstPathSegment(serviceBundle.Config.Authority.AuthorityInfo.CanonicalAuthority);
+                if (AadAuthority.IsCommonOrganizationsOrConsumersTenant(tenant))
+                {
+                    throw new MsalClientException(
+                        MsalError.MissingTenantedAuthority,
```
neha-bhargava, line 145: Can you add tests for this?

gladjohn: it is here in the same test - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/41a5556331a17a9304dc8f85e5d89c803d1770fb/tests/Microsoft.Identity.Test.Unit/PublicApiTests/MtlsPopTests.cs#L588 

The diff doesn't show it here very well.

gladjohn:
@microsoft-github-policy-service rerun

PR #5877, opened 2026-03-23
[Proposal] Add opt‑in MSAL.NET metrics enrichment

ssmelov:
This PR proposes an extensibility mechanism to **enrich existing MSAL OpenTelemetry metrics with additional dimensions**, without changing or replacing MSAL’s canonical metric set.

This PR is intentionally scoped as a **proposal** to illustrate the API shape and discuss trade‑offs before committing to a final design.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
+++ b/src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
@@ -1,6 +1,8 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using Microsoft.Identity.Client.Cache;
 using Microsoft.Identity.Client.Core;
@@ -90,7 +92,7 @@ namespace Microsoft.Identity.Client.Platforms.Features.OpenTelemetry
 
         public OtelInstrumentation()
         {
-            // Needed to fail fast if the runtime, like in-process Azure Functions, doesn't support OpenTelemetry 
+            // Needed to fail fast if the runtime, like in-process Azure Functions, doesn't support OpenTelemetry
             _ = Meter.Version;
         }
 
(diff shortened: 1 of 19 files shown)


src/client/Microsoft.Identity.Client/TokenAcquisitionResult.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+
+namespace Microsoft.Identity.Client
+{
+    /// <summary>
+    /// Represents the outcome of a token acquisition operation.
+    /// Carries either a successful <see cref="AuthenticationResult"/> or the
+    /// <see cref="Exception"/> that caused the request to fail — never both at the same time.
+    /// </summary>
+    public sealed class TokenAcquisitionResult
```
bgavrilMS, line 13: There already exists ExecutionResult for similar purposes. Try to reuse that.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -114,109 +127,162 @@ public void LogSuccessMetrics(
...
+                {
+                    new(TelemetryConstants.MsalVersion, MsalIdHelper.GetMsalVersion()),
+                    new(TelemetryConstants.Platform, platform),
+                    new(TelemetryConstants.ApiId, apiId),
+                    new(TelemetryConstants.TokenSource, authResultMetadata.TokenSource),
+                    new(TelemetryConstants.CacheLevel, cacheLevel),
+                    new(TelemetryConstants.CacheRefreshReason, authResultMetadata.CacheRefreshReason),
+                    new(TelemetryConstants.TokenType, authResultMetadata.TelemetryTokenType)
+                };
+                AppendExtraTags(ref tags, extraTags);
```
bgavrilMS, line 145: Main concern is max numbers of dimensions that a Record can have. Based on prev experience, this is 8. So adding to existing ones is problematic. 

Maybe a better design is to fire 1-2 new Records ?

neha-bhargava: Agree, having >8 tags causes performance issues. This is a known constraint of otel. I would suggest to create another counter. That can have the additional tags and any other already used tags that are useful

neha-bhargava:
This will be completed in another PR

PR #5887, opened 2026-03-25
Add in-process MAA token caching to PopKeyAttestor

Robbie-Microsoft:
## Summary

MAA (Microsoft Azure Attestation) tokens are acquired during mTLS binding certificate minting, requiring a native DLL call + network round-trip to the MAA endpoint. Since MAA tokens carry their
 own JWT expiry and have a longer TTL than binding certs, they can safely be reused across cert re-mints.
 
 This PR adds an in-memory cache for MAA tokens inside `PopKeyAttestor` to avoid redundant native DLL and network calls.
 
 > **Key point:** when the cert cache is warm, attestation is skipped entirely by the cert-reuse path. The MAA cache activates when the cert cache is cold (first boot, cert expiry, or explicit 
eviction) — it ensures that if multiple cert re-mints happen before the MAA token expires, only the first one calls the attestation service.
 
 ## Changes
 
 ### `PopKeyAttestor.cs`
 - Adds a static `ConcurrentDictionary<string, AttestationToken>` cache keyed by `"{endpoint}|{keyId}"`
   - `endpoint` — different MAA regional endpoints have different trust domains
   - `keyId` — `CngKey.KeyName`; ensures key rotation does not return a stale token for the new key
   - `clientId` intentionally **omitted** — MAA attests the CNG key, not the identity; SAMI and UAMI sharing the same key at the same endpoint correctly share the cached token
 - Key handle validation (`null`, `IsInvalid`, `SafeNCryptKeyHandle` cast) always runs before the cache check
 - Tokens within 5 minutes of expiry are evicted (matches `AccessTokenExpirationBuffer`); freshness check guards against `DateTimeOffset.MinValue - buffer` overflow
 - Per-key single-flight via `KeyedSemaphorePool` prevents thundering-herd redundant attestation calls
 - Cache hit/miss/store logged via `ILoggerAdapter` threaded from request context through `AttestationTokenProvider` delegate
 - Static constructor registers `ResetCacheForTest` with `ApplicationBase.RegisterResetCallback` so `ApplicationBase.ResetStateForTest()` automatically clears the MAA cache
 
 ### `ApplicationBase.cs`
 - Adds `RegisterResetCallback(Action)` + `ConcurrentBag<Action> s_resetCallbacks`
 - `ResetStateForTest()` invokes all registered callbacks after its own resets
 
 ### `AcquireTokenCommonParameters` / `AcquireTokenForManagedIdentityParameters`
 - `AttestationTokenProvider` delegate updated to include `ILoggerAdapter` and `string keyId`
 
 ### `ImdsV2ManagedIdentitySource.cs`
 - Extracts `rsaCng.Key.KeyName` and forwards it as `keyId` to the attestation delegate
 
 ### `ImdsV2Tests.cs`
 - `[TestCleanup]` calls `PopKeyAttestor.ResetCacheForTest()` (belt-and-suspenders alongside the `ApplicationBase` callback)
 - Adds 6 new tests:
   - `MaaTokenCache_Hit_DoesNotCallAttestationProviderAgain`
   - `MaaTokenCache_ExpiredToken_CallsAttestationProviderAgain`
   - `MaaTokenCache_DifferentEndpoints_CachedSeparately`
   - `MaaTokenCache_DifferentKeyIds_CachedSeparately` — two different CNG keys at the same endpoint get separate cache entries (key rotation safety)
   - `MaaTokenCache_ConcurrentCacheMiss_SingleFlightCallsProviderOnce`
   - `MaaTokenCache_CertCacheMiss_MaaTokenCacheStillFresh_DoesNotCallAttestationProviderAgain`
 
 ## Test results
 
 81/81 `ImdsV2Tests` passing on .NET 4.8 and .NET 8, no regressions.
 
 ## No public API changes
 
 Caching is entirely internal to `PopKeyAttestor`. `WithAttestationSupport()` is unchanged.


tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -948,6 +953,153 @@ await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
...
+            };
+
+            const string endpoint1 = "https://eastus.attestation.azure.net";
+            const string endpoint2 = "https://westus.attestation.azure.net";
+
+            // Use a real RSACng key handle so validation passes on first (cache-miss) calls
+            using var rsa1 = new RSACng(2048);
+            using var rsa2 = new RSACng(2048);
+            var handle1 = rsa1.Key.Handle;
+            var handle2 = rsa2.Key.Handle;
```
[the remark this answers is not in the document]
Robbie-Microsoft: feature is windows-only

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -40,39 +60,87 @@ public static Task<AttestationResult> AttestCredentialGuardAsync(
...
+        {
+            if (s_tokenCache.TryGetValue(cacheKey, out token) &&
+                token.ExpiresOn - s_expirationBuffer > DateTimeOffset.UtcNow)
+            {
+                return true;
+            }
+
+            token = null;
+            return false;
+        }
```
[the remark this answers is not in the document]
Robbie-Microsoft: fixed

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -40,39 +60,87 @@ public static Task<AttestationResult> AttestCredentialGuardAsync(
...
+
+            // Check the in-process cache before making any native/network calls.
+            // Key validation is intentionally deferred: a cache hit returns immediately
+            // without requiring (or validating) the key handle.
+            string cacheKey = BuildCacheKey(endpoint, clientId);
+            if (TryGetCachedToken(cacheKey, out AttestationToken cached))
+            {
+                return Task.FromResult(new AttestationResult(
+                    AttestationStatus.Success, cached, cached.Token, 0, null));
+            }
```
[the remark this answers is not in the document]
Robbie-Microsoft: fixed

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -40,39 +60,87 @@ public static Task<AttestationResult> AttestCredentialGuardAsync(
...
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // Check the in-process cache before making any native/network calls.
+            // Key validation is intentionally deferred: a cache hit returns immediately
+            // without requiring (or validating) the key handle.
+            string cacheKey = BuildCacheKey(endpoint, clientId);
+            if (TryGetCachedToken(cacheKey, out AttestationToken cached))
+            {
+                return Task.FromResult(new AttestationResult(
+                    AttestationStatus.Success, cached, cached.Token, 0, null));
```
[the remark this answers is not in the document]
Robbie-Microsoft: fixed

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -12,10 +13,23 @@ namespace Microsoft.Identity.Client.KeyAttestation
 {
     /// <summary>
     /// Static facade for attesting a Credential Guard/CNG key and getting a JWT back.
+    /// Caches valid MAA tokens to avoid redundant native DLL calls and network round-trips.
```
gladjohn, line 18:
```suggestion
    /// Caches MAA tokens to avoid redundant native DLL calls and network round-trips.
```

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +39,131 @@ internal static class PopKeyAttestor
         /// Tests should not run in parallel when using this hook to avoid race conditions.
         /// </remarks>
         internal static Func<string, SafeHandle, string, CancellationToken, Task<AttestationResult>> s_testAttestationProvider;
+
+        /// <summary>
+        /// Resets the MAA token cache and key locks. Call from [TestCleanup] to prevent cache state leaking between tests.
+        /// </summary>
+        internal static void ResetCacheForTest()
```
gladjohn, line 84: We have a master reset, we need to add this there as well.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +39,131 @@ internal static class PopKeyAttestor
...
-
             if (string.IsNullOrWhiteSpace(endpoint))
                 throw new ArgumentNullException(nameof(endpoint));
 
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // Fast path: check cache without acquiring any lock.
+            // Key validation is intentionally deferred: a cache hit returns immediately
+            // without requiring (or validating) the key handle.
+            string cacheKey = BuildCacheKey(endpoint, clientId);
```
gladjohn: Should the cache key also include the KeyGuard, not sure if ClientId needs to be part of the cache key ?

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -12,10 +13,23 @@ namespace Microsoft.Identity.Client.KeyAttestation
 {
     /// <summary>
     /// Static facade for attesting a Credential Guard/CNG key and getting a JWT back.
+    /// Caches valid MAA tokens to avoid redundant native DLL calls and network round-trips.
     /// Key discovery / rotation is the caller's responsibility.
     /// </summary>
     internal static class PopKeyAttestor
     {
+        // In-process cache: maps "{endpoint}|{clientId}" → AttestationToken (JWT + expiry).
```
gladjohn: do we plan to persist this cache across restarts/reboots?

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -12,10 +13,23 @@ namespace Microsoft.Identity.Client.KeyAttestation
...
+        private static readonly ConcurrentDictionary<string, AttestationToken> s_tokenCache =
+            new ConcurrentDictionary<string, AttestationToken>(StringComparer.OrdinalIgnoreCase);
+
+        // Per-key semaphores to prevent concurrent in-flight attestation for the same cache key (single-flight).
+        private static readonly ConcurrentDictionary<string, SemaphoreSlim> s_keyLocks =
+            new ConcurrentDictionary<string, SemaphoreSlim>(StringComparer.OrdinalIgnoreCase);
+
+        // Tokens within this window of expiry are considered stale and will be refreshed.
+        // Matches MSAL's AccessTokenExpirationBuffer (5 minutes).
```
gladjohn, line 58: we should do a background refresh at half-life

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +39,131 @@ internal static class PopKeyAttestor
...
             if (string.IsNullOrWhiteSpace(endpoint))
                 throw new ArgumentNullException(nameof(endpoint));
 
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // Fast path: check cache without acquiring any lock.
+            // Key validation is intentionally deferred: a cache hit returns immediately
+            // without requiring (or validating) the key handle.
+            string cacheKey = BuildCacheKey(endpoint, clientId);
+            if (TryGetCachedToken(cacheKey, out AttestationToken cached))
```
gladjohn: Could we add a couple of cache logs here for hit/miss/refresh?

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -64,6 +66,9 @@ public void ImdsV2Tests_Cleanup()
 
             // Reset test provider to ensure clean state for other tests
             PopKeyAttestor.s_testAttestationProvider = null;
+
+            // Reset MAA token cache so cached tokens don't leak between tests
+            PopKeyAttestor.ResetCacheForTest();
```
gladjohn, line 71: We have an Application Base ResetStaticCache. We need to plug in all static cache reset to that. This enables IdWeb and Az SDK to reset these as well.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -948,6 +953,189 @@ await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
             }
         }
 
+        [TestMethod]
```
gladjohn, line 960: Could we add one test that forces a cert-cache miss while the MAA cache is still fresh?

gladjohn (commented):
Thanks Robbie, looks good overall. Are we able to use any of the cache helper pattern from the cert cache here?

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
bgavrilMS: As a separate PR at most. 

Can you please help me understand why MAA caching and MAA proactive refresh is needed when we do certificate caching?

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
bgavrilMS: 1. Shouldn't the cache key contain `Key.ID` ? (or hash(key.public) ) ?

Note that it's not easy to derive a key ID from this key handle, but the API can change to accept an extra parameter.

3. Do we need endpoint and client ID in the key ID ? It's probably safer, but needs a comment. I think endpoint is not really needed though.

bgavrilMS: Not sure about this @gladjohn. You are the SME and should be suggesting the right answer. 

IMO, in-memory caching is sufficient. I would argue even in-memory caching is not needed. Can you please articulate scenarios for caching and persisted caching?

Remember that all caching needs to include: 

1. Operational perspective on how to deal with poisoned cache entry
2. Encryption at rest
3. Locks.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +39,131 @@ internal static class PopKeyAttestor
...
             string clientId,
             CancellationToken cancellationToken = default)
         {
-            if (keyHandle is null)
-                throw new ArgumentNullException(nameof(keyHandle));
-
             if (string.IsNullOrWhiteSpace(endpoint))
                 throw new ArgumentNullException(nameof(endpoint));
 
+            cancellationToken.ThrowIfCancellationRequested();
```
bgavrilMS: Is it really needed? Are you doing some long-lived op taht would benefit from cancellation?

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +39,131 @@ internal static class PopKeyAttestor
...
+            // without requiring (or validating) the key handle.
+            string cacheKey = BuildCacheKey(endpoint, clientId);
+            if (TryGetCachedToken(cacheKey, out AttestationToken cached))
+            {
+                return new AttestationResult(
+                    AttestationStatus.Success, cached, cached.Token, 0, string.Empty);
+            }
+
+            // Cache miss — validate the key handle before acquiring the lock.
+            if (keyHandle is null)
```
bgavrilMS, line 118: I think the keyHandle needs to exist. Why would a client ask for an attestation token if it doesn't have the key?

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +39,131 @@ internal static class PopKeyAttestor
...
+        /// Uses a per-key semaphore to ensure only one in-flight attestation call occurs per cache key,
+        /// preventing redundant native DLL and network calls under concurrency.
+        /// Cancellation only applies before the native call starts.
         /// </summary>
         /// <param name="endpoint">Attestation service endpoint (required).</param>
         /// <param name="keyHandle">Valid SafeNCryptKeyHandle (must remain valid for duration of call).</param>
         /// <param name="clientId">Optional client identifier (may be null/empty).</param>
         /// <param name="cancellationToken">Cancellation token (cooperative before scheduling / start).</param>
-        public static Task<AttestationResult> AttestCredentialGuardAsync(
+        public static async Task<AttestationResult> AttestCredentialGuardAsync(
```
bgavrilMS, line 106: We need logging to understand cache hit, cache miss etc.

Robbie-Microsoft: Done - PopKeyAttestor now logs cache hit, post-lock hit, cache miss, and successful cache store at Info level, with the cache key in each message. The ILoggerAdapter is threaded from the request context through the AttestationTokenProvider delegate.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: Fair point - removed the redundant cancellationToken.ThrowIfCancellationRequested() at the top. The CT is still passed to semaphore.WaitAsync and to the attestation call, but there's no longer an explicit redundant check.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: Agreed - moved all keyHandle validation (null, IsInvalid, type check) to before the cache check. The handle is always required now.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: The MAA native DLL call + network round-trip adds ~100-500ms per cert mint. Without the cache, every cert re-mint (e.g. on cert expiry every 24h) would incur that cost again. Since the CNG key is persistent and the MAA token lifetime is hours, the cache is justified. Persisted cache (across reboots) was not implemented - agreed it's overkill and introduces encryption-at-rest, poisoning, and locking concerns

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: When the cert cache is warm, MAA is not called at all (cert re-use path skips attestation). But when the cert cache is cold (first boot, cert expiry, or explicit eviction), we have to mint a new cert, which includes the MAA call. The MAA token has a longer TTL than the cert, so caching it avoids a redundant network call on the next cert re-mint. MaaTokenCache_CertCacheMiss_MaaTokenCacheStillFresh_DoesNotCallAttestationProviderAgain demonstrates this directly - it clears the cert cache, forces a re-mint, and verifies the attestation provider is not called a second time.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: done

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: Done - added ApplicationBase.RegisterResetCallback(Action) so extension packages can register their own resets without creating a circular dependency. PopKeyAttestor's static constructor now registers and AzSDK calling ApplicationBase.ResetStateForTest() will now also clear the MAA cache.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: Addressed with a comment ([see Bogdan's comment](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5887#discussion_r3009692197)). clientId is sufficient to distinguish identities; the KeyGuard attestation endpoint is already part of the key via endpoint.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: Done - same as [Bogdan's response](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5887#discussion_r3009722460).

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: Done - added MaaTokenCache_CertCacheMiss_MaaTokenCacheStillFresh_DoesNotCallAttestationProviderAgain. It clears only the cert cache between two acquires and verifies the attestation provider is still only called once.

Robbie-Microsoft:
> Are we able to use any of the cache helper pattern from the cert cache here?

The cert cache uses a two-tier (in-memory + persistent) design with a full interface hierarchy (ICertificateCache, IMtlsCertificateCache, etc.) because certs need to survive process restarts and are encrypted at rest. The MAA token cache is intentionally simpler — in-memory only, with a ConcurrentDictionary + per-key semaphore for single-flight. Reusing the cert cache pattern would bring in persistence, encryption, and MtlsBindingCache orchestration that aren't warranted for a short-lived JWT. If the requirements change (e.g. persist across reboots), we can adopt that pattern then.

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
Robbie-Microsoft: Done - added ApplicationBase.RegisterResetCallback(Action) which allows extension packages to register their own static cache reset without a circular dependency back into Microsoft.Identity.Client. PopKeyAttestor's static constructor now calls ApplicationBase.RegisterResetCallback(ResetCacheForTest), so any caller of ApplicationBase.ResetStateForTest() (including IdWeb and AzSDK) will automatically clear the MAA token cache as well.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
Robbie-Microsoft: Good points. Added a comment in BuildCacheKey explaining the design rationale:
- Key.ID not included: SafeHandle doesn't expose key material, and deriving a stable key ID would require platform-specific CNG API calls outside the current scope. Agreed this is a gap - left a note to revisit when the API can accept an explicit key ID parameter.
- endpoint included: Kept as a safety boundary since different MAA endpoints can have different trust domains (e.g. region-specific attestation). Open to dropping it if you feel strongly.
- clientId included: Needed to distinguish between different managed identities (SAMI vs different UAMIs) that may share the same process but call different attestation endpoints under the same key.

src/client/Microsoft.Identity.Client/ApplicationBase.cs
```diff
@@ -28,6 +29,19 @@ public abstract class ApplicationBase : IApplicationBase
...
+        // Allows extension packages (e.g. KeyAttestation) to register their own static-cache reset logic
+        // without introducing a circular dependency. Invoked by ResetStateForTest().
+        private static readonly ConcurrentBag<Action> s_resetCallbacks = new ConcurrentBag<Action>();
+
+        /// <summary>
+        /// Registers a callback to be invoked by <see cref="ResetStateForTest"/>.
+        /// Intended for extension packages (e.g. KeyAttestation) that own static caches MSAL cannot reference directly.
+        /// </summary>
+        internal static void RegisterResetCallback(Action callback)
+        {
```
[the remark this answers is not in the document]
Robbie-Microsoft: Added if (callback is null) throw new ArgumentNullException(...)

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -2,20 +2,41 @@
...
+        // require platform-specific CNG API calls. In practice the same CNG key is reused for
+        // the lifetime of the cert (cert cache lifetime), so endpoint+clientId is sufficient.
+        // Tracking issue for adding Key.ID: include when the API is updated to accept it.
+        private static readonly ConcurrentDictionary<string, AttestationToken> s_tokenCache =
+            new ConcurrentDictionary<string, AttestationToken>(StringComparer.OrdinalIgnoreCase);
+
+        // Per-key semaphores to prevent concurrent in-flight attestation for the same cache key (single-flight).
+        private static readonly ConcurrentDictionary<string, SemaphoreSlim> s_keyLocks =
+            new ConcurrentDictionary<string, SemaphoreSlim>(StringComparer.OrdinalIgnoreCase);
+
```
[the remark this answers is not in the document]
Robbie-Microsoft: Replaced with KeyedSemaphorePool from PlatformsCommon.Shared

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +46,145 @@ internal static class PopKeyAttestor
...
+            }
+
+            return result;
+        }
+
+        private static bool TryGetCachedToken(string cacheKey, out AttestationToken token)
+        {
+            if (s_tokenCache.TryGetValue(cacheKey, out token))
+            {
+                if (token.ExpiresOn - s_expirationBuffer > DateTimeOffset.UtcNow)
```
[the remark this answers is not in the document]
Robbie-Microsoft: Flipped to token.ExpiresOn > UtcNow + buffer

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -948,8 +953,245 @@ await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
...
+                Assert.AreEqual(1, providerCallCount, "MAA should be called once on first acquire.");
+
+                // Second acquire: cert is cached, so MAA token cache should also be hit
+                var second = await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
+                    .WithMtlsProofOfPossession()
+                    .WithAttestationSupport()
+                    .ExecuteAsync().ConfigureAwait(false);
+
+                Assert.AreEqual(1, providerCallCount, "MAA should NOT be called again when token is cached.");
+                Assert.AreEqual(TokenSource.Cache, second.AuthenticationResultMetadata.TokenSource);
```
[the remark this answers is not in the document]
Robbie-Microsoft: Added WithForceRefresh(true) + AddMocksToGetEntraTokenUsingCachedCert; asserts TokenSource.IdentityProvider

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -948,8 +953,245 @@ await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
...
+
+                // Second acquire: cert cache miss → re-mints cert, but MAA cache hit → provider NOT called again
+                AddMocksToGetEntraToken(httpManager, certificateRequestCertificate: rawCert);
+                await mi2.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
+                    .WithMtlsProofOfPossession()
+                    .WithAttestationSupport()
+                    .ExecuteAsync().ConfigureAwait(false);
+
+                Assert.AreEqual(1, providerCallCount, "MAA should NOT be called again when MAA token cache is still fresh, even after cert cache miss.");
+            }
```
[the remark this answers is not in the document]
Robbie-Microsoft: Added TokenSource.IdentityProvider assertion

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -948,8 +953,245 @@ await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
...
+
+            // Fire 5 concurrent attestation calls for the same key before any result is cached
+            var tasks = new Task<AttestationResult>[5];
+            for (int i = 0; i < tasks.Length; i++)
+                tasks[i] = PopKeyAttestor.AttestCredentialGuardAsync(endpoint, handle, "client1", cancellationToken: CancellationToken.None);
+
+            // Allow exactly one provider call to complete
+            providerGate.Release(1);
+            var results = await Task.WhenAll(tasks).ConfigureAwait(false);
+
```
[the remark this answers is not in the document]
Robbie-Microsoft: Changed to SemaphoreSlim(0, 5) and releases tasks.Length

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +46,145 @@ internal static class PopKeyAttestor
...
             var safeNCryptKeyHandle = keyHandle as SafeNCryptKeyHandle
                 ?? throw new ArgumentException("keyHandle must be a SafeNCryptKeyHandle. Only Windows CNG keys are supported.", nameof(keyHandle));
 
-            cancellationToken.ThrowIfCancellationRequested();
+            string cacheKey = BuildCacheKey(endpoint, clientId);
 
-            // Check for test provider to avoid loading native DLL in unit tests
-            if (s_testAttestationProvider != null)
+            // Fast path: check cache without acquiring any lock.
+            if (TryGetCachedToken(cacheKey, out AttestationToken cached))
```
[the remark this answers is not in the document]
Robbie-Microsoft: Will update PR description

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -25,54 +46,146 @@ internal static class PopKeyAttestor
...
+        }
+
+        /// <summary>
+        /// Resets the MAA token cache and key locks. Called automatically via
+        /// <see cref="ApplicationBase.ResetStateForTest"/>; also callable directly from [TestCleanup].
+        /// </summary>
+        internal static void ResetCacheForTest()
+        {
+            s_tokenCache.Clear();
+        }
```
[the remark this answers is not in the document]
Robbie-Microsoft: Done - updated the XML doc to reflect that only s_tokenCache is cleared. KeyedSemaphorePool is intentionally not reset: semaphores are stateless coordination primitives (they hold no cached data), and KeyedSemaphorePool deliberately omits Clear() to avoid races with concurrent callers.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
gladjohn: PDP and Storage has multiple identities on the same VM, so caching MAA token (in-memory) will surely benefit. 

There was an ask to add persistent cache for a different client. That workload will have 50+ processes spinning up on the same VM. Let me check if that plan still is on the roadmap.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
gladjohn: > Can you please help me understand why MAA caching and MAA proactive refresh is needed when we do certificate caching?

Services have multiple UAMI's running on the same VM accessing different resources. Caching MAA token (in-memory) helps re-use same cached token for different identities. KG key is one per VM and MAA token is for taht key. So caching MAA token helps.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
gladjohn: > clientId included: Needed to distinguish between different managed identities (SAMI vs different UAMIs) that may share the same process but call different attestation endpoints under the same key.

can this happen? good to check with IMDS team. But still, a cached MAA token is for the key and not for the identity.

gladjohn: design documented [here](https://microsoft.sharepoint-df.com/:fl:/g/contentstorage/CSP_3ca947e4-ffaf-45ea-945e-e79db8129325/IQC1RrDtu_MCQ4G6Zt3tYQcvAYJDJhyQ-1b3DrJF7TcJShs?e=nAmcC6&nav=cz0lMkZjb250ZW50c3RvcmFnZSUyRkNTUF8zY2E5NDdlNC1mZmFmLTQ1ZWEtOTQ1ZS1lNzlkYjgxMjkzMjUmZD1iJTIxNUVlcFBLX182a1dVWHVlZHVCS1RKVVJ4VHpHV3lfWktveDllaFI4eFRTNjEwcjZwbHlpOFRvSnJld0hFRnhHQiZmPTAxUldUSloyRlZJMllPM083VEFKQllET1RHM1hXV0NCWlAmYz0lMkYmYT1Mb29wQXBwJnA9JTQwZmx1aWR4JTJGbG9vcC1wYWdlLWNvbnRhaW5lciZ4PSU3QiUyMnclMjIlM0ElMjJUMFJUVUh4dGFXTnliM052Wm5RdWMyaGhjbVZ3YjJsdWRDMWtaaTVqYjIxOFlpRTFSV1Z3VUV0Zlh6WnJWMVZZZFdWa2RVSkxWRXBWVW5oVWVrZFhlVjlhUzI5NE9XVm9Vamg0VkZNMk1UQnlObkJzZVdrNFZHOUtjbVYzU0VWR2VFZENmREF4VWxkVVNsb3lRazlYUVRWVVIweEVVMHhhUmxsRlJWSkRWa2RPUlZGWVRWWSUzRCUyMiUyQyUyMmklMjIlM0ElMjJhODc1Y2FhMS1mOGQwLTQ4MDMtOTAyMC0yZDI1NjcxNTRkZTUlMjIlN0Q%3D)

gladjohn: You don't need a cache key at all. You just need:

```cs
private static CachedToken _cachedMaaToken;

if (_cachedMaaToken != null && !_cachedMaaToken.IsExpired)
{
    return _cachedMaaToken.Token;
}
```

bgavrilMS: The underlying key is CngKey or smth. We can pass an extra parameter to this method `(SafeHandle handle, string keyID, ...`

Robbie-Microsoft: Done - cache key is now {endpoint}|{keyId} where keyId is CngKey.KeyName. Per Gladwin's feedback, clientId was removed since MAA attests the key, not the identity. Added a test (MaaTokenCache_DifferentKeyIds_CachedSeparately) to verify two different keys at the same endpoint get separate cache entries.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -2,20 +2,46 @@
...
 {
     /// <summary>
     /// Static facade for attesting a Credential Guard/CNG key and getting a JWT back.
+    /// Caches valid MAA tokens to avoid redundant native DLL calls and network round-trips.
     /// Key discovery / rotation is the caller's responsibility.
     /// </summary>
     internal static class PopKeyAttestor
     {
+        // In-process cache: maps "{endpoint}|{clientId}|{keyId}" → AttestationToken (JWT + expiry).
+        //
```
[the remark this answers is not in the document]
Robbie-Microsoft: updated description

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
gladjohn: For now, **in-memory caching is sufficient** and gives us the main benefit: avoiding repeated MAA DLL + network cost for scenarios where multiple identities/processes run on the same VM (for example PDP / Storage style workloads).

Persisted caching feels like a separate design because it brings additional requirements:
- encryption at rest
- poisoned cache handling
- cross-process locking
- reboot-time invalidation semantics

My recommendation: **keep in-memory cache only for now** and only consider persisted cache later if we have a confirmed workload that truly needs it (for example the previously mentioned 50+ processes on the same VM scenario).

gladjohn (approved):

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -24,55 +50,152 @@ internal static class PopKeyAttestor
...
 
-            // Check for test provider to avoid loading native DLL in unit tests
-            if (s_testAttestationProvider != null)
+            // Fast path: check cache without acquiring any lock.
+            if (TryGetCachedToken(cacheKey, out AttestationToken cached))
             {
-                return s_testAttestationProvider(endpoint, keyHandle, clientId, cancellationToken);
+                logger?.Info($"[PopKeyAttestor] MAA token cache hit for '{cacheKey}'.");
+                return new AttestationResult(
+                    AttestationStatus.Success, cached, cached.Token, 0, string.Empty);
```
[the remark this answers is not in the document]
Robbie-Microsoft: All 5 Info($"...") → Info(() => $"...")

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -24,55 +50,152 @@ internal static class PopKeyAttestor
...
+                s_tokenCache.TryRemove(cacheKey, out _);
+            }
+
+            token = null;
+            return false;
+        }
+
+        private static string BuildCacheKey(string endpoint, string clientId, string keyId)
+        {
+            return $"{endpoint}|{clientId ?? string.Empty}|{keyId ?? string.Empty}";
```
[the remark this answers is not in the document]
Robbie-Microsoft: Added NormalizeEndpoint() helper (trims trailing /)

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -2,20 +2,46 @@
...
+        //   keyId     – the CNG key name (CngKey.KeyName), added so that if the same
+        //               endpoint+client ever uses two different CNG keys (e.g. after key
+        //               rotation), the stale token for the old key is not returned for the
+        //               new one. Null/empty for ephemeral (non-KSP) keys.
+        private static readonly ConcurrentDictionary<string, AttestationToken> s_tokenCache =
+            new ConcurrentDictionary<string, AttestationToken>(StringComparer.OrdinalIgnoreCase);
+
+        // Per-key async gates to prevent concurrent in-flight attestation for the same cache key (single-flight).
+        private static readonly KeyedSemaphorePool s_keyLocks = new KeyedSemaphorePool();
+
```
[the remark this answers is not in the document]
Robbie-Microsoft: Added comment explaining the pool is bounded in practice (named keys only; ephemeral keys share one slot)

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -2,20 +2,46 @@
...
+        //   endpoint  – included because the MAA service issues endpoint-specific tokens;
+        //               eastus.attestation.azure.net and westus.attestation.azure.net issue
+        //               different tokens for the same key/client combination.
+        //   clientId  – included because it is sent in the attestation request body and
+        //               affects the claims in the issued token.
+        //   keyId     – the CNG key name (CngKey.KeyName), added so that if the same
+        //               endpoint+client ever uses two different CNG keys (e.g. after key
+        //               rotation), the stale token for the old key is not returned for the
+        //               new one. Null/empty for ephemeral (non-KSP) keys.
+        private static readonly ConcurrentDictionary<string, AttestationToken> s_tokenCache =
```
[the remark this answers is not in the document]
Robbie-Microsoft: Added code comment explaining the trade-off + fixed 2 tests that were incorrectly relying on rsa.Key.KeyName (null for ephemeral keys) instead of explicit keyIds

bgavrilMS (approved):

bgavrilMS (approved):

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -24,55 +60,167 @@ internal static class PopKeyAttestor
...
 
-            cancellationToken.ThrowIfCancellationRequested();
+            string cacheKey = BuildCacheKey(endpoint, keyId);
+
+            // Note on null/empty keyId (ephemeral, non-KSP keys):
+            // When keyId is null, the cache key collapses to "{endpoint}|", which means all
+            // ephemeral keys at the same endpoint share one cache slot. In production, Credential
+            // Guard always uses persisted (named) CNG keys so keyId is never null. The degenerate
+            // case is accepted: a process can only have one in-flight attestation for the same
+            // endpoint, and the MAA token is bound to the endpoint, not the key itself.
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed. ImdsV2ManagedIdentitySource now derives keyId from a SHA-256 fingerprint of the RSA public key (Modulus+Exponent) when rsaCng.Key.KeyName is null, so each distinct ephemeral key gets its own MAA cache slot. Named (KSP-persisted) keys continue to use their KeyName unchanged. PopKeyAttestor retains a defensive bypass for any null keyId that may come from other callers.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -2,20 +2,56 @@
...
+        //               deployments) or multiple named CNG keys (e.g. during key rotation). A single
+        //               field would incorrectly return the wrong token for the second endpoint/key.
+        //               With endpoint+keyId as a compound key the cache correctly scopes each entry.
+        private static readonly ConcurrentDictionary<string, AttestationToken> s_tokenCache =
+            new ConcurrentDictionary<string, AttestationToken>(StringComparer.OrdinalIgnoreCase);
+
+        // Per-key async gates to prevent concurrent in-flight attestation for the same cache key (single-flight).
+        // The pool is bounded in practice: ephemeral (unnamed) keys bypass the cache/semaphore entirely,
+        // so only named keys ever enter the pool. A typical process uses one named key per endpoint.
+        private static readonly KeyedSemaphorePool s_keyLocks = new KeyedSemaphorePool();
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in latest commit. Updated all four comment locations (s_keyLocks field, cache key design block, keyId XML param doc, and the inline guard) to reflect that callers like ImdsV2ManagedIdentitySource always supply a non-empty keyId — either the CNG KeyName or a SHA-256 fingerprint for ephemeral keys. The 
ull/empty path is now described as a defensive fallback, not the normal ephemeral-key code path.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -24,55 +60,186 @@ internal static class PopKeyAttestor
...
                 {
-                    using var client = new AttestationClient();
-                    return client.Attest(endpoint, safeNCryptKeyHandle, clientId ?? string.Empty);
+                    logger?.Info(() => $"[PopKeyAttestor] MAA token cache hit (post-lock) for '{cacheKey}'.");
+                    return new AttestationResult(
+                        AttestationStatus.Success, cached, cached.Token, 0, string.Empty);
                 }
-                catch (Exception ex)
+
+                logger?.Info(() => $"[PopKeyAttestor] Calling attestation provider for '{cacheKey}'.");
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in latest commit. All 5 cache log calls (hit, miss, post-lock hit, calling provider, stored) are now `Verbose` instead of `Info`, and the cache key is wrapped through a new `Mask()` helper that shows only the last 8 chars + total length (e.g. `…abc12345(47)`). The helper matches the convention in `InMemoryCertificateCache`.

gladjohn (approved):

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
```diff
@@ -948,8 +957,318 @@ await mi.AcquireTokenForManagedIdentity(ManagedIdentityTests.Resource)
...
+                    .WithAttestationSupport()
+                    .ExecuteAsync().ConfigureAwait(false);
+
+                Assert.AreEqual(1, providerCallCount, "MAA should be called once on first acquire.");
+
+                // Second acquire: force bypass the MSAL token cache so we actually exercise the
+                // cert/MAA path again. Cert is still cached so no re-mint; MAA cache is hit so
+                // the attestation provider must NOT be called a second time.
+                MockHelpers.AddMocksToGetEntraTokenUsingCachedCert(
+                    httpManager,
```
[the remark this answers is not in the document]
gladjohn: agreed - we need this unit test to be added

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -24,55 +63,203 @@ internal static class PopKeyAttestor
...
+            s_tokenCache.Clear();
+        }
+
         /// <summary>
         /// Asynchronously attests a Credential Guard/CNG key with the remote attestation service and returns a JWT.
-        /// Wraps the synchronous <see cref="AttestationClient.Attest"/> in a Task.Run so callers can
-        /// avoid blocking. Cancellation only applies before the native call starts.
+        /// Returns a cached token if one is available and not within the expiration buffer.
+        /// Uses a per-key semaphore to ensure only one in-flight attestation call occurs per cache key,
+        /// preventing redundant native DLL and network calls under concurrency.
```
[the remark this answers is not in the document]
gladjohn: +1

tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ImdsV2Tests.cs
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed. The test now uses a shared TestKeyGuardManagedIdentityKeyProvider instance and resets both the cert cache (ImdsV2ManagedIdentitySource.ResetCertCacheForTest()) and the MSAL source (ManagedIdentityClient.ResetSourceForTest()) between the two acquires. The second acquire is forced through the full cert-mint + attestation path, so the MAA token cache is the only thing preventing a second provider call. See latest commit 157c5c89c.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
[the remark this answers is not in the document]
Robbie-Microsoft: Done. Added a note to the XML doc summary stating that the native attestation call is wrapped in \Task.Run\ and that cancellation can prevent the operation from starting but cannot interrupt it once under way. See latest commit 157c5c89c.

src/client/Microsoft.Identity.Client.KeyAttestation/PopKeyAttestor.cs
```diff
@@ -2,20 +2,59 @@
...
+        //               token is valid for a given key. Different managed identities (SAMI vs UAMI)
+        //               sharing the same CNG key at the same endpoint correctly share the cached token.
+        //
+        //   Why a dictionary and not a single field?
+        //               A process could theoretically use multiple MAA endpoints (e.g. in multi-region
+        //               deployments) or multiple named CNG keys (e.g. during key rotation). A single
+        //               field would incorrectly return the wrong token for the second endpoint/key.
+        //               With endpoint+keyId as a compound key the cache correctly scopes each entry.
+        private static readonly ConcurrentDictionary<string, AttestationToken> s_tokenCache =
+            new ConcurrentDictionary<string, AttestationToken>(StringComparer.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed. Switched `s_tokenCache` to `StringComparer.Ordinal` and updated `NormalizeEndpoint()` to lowercase the endpoint string before composing the cache key. This keeps host comparisons effectively case-insensitive while leaving `keyId` (CNG key name or SHA-256 fingerprint) case-sensitive — matching the `InMemoryCertificateCache` convention. See commit 2d2c6f3ad.

src/client/Microsoft.Identity.Client/ApplicationBase.cs
```diff
@@ -101,6 +118,9 @@ public static void ResetStateForTest()
 
             InMemoryPartitionedAppTokenCacheAccessor.ClearStaticCacheForTest();
             InMemoryPartitionedUserTokenCacheAccessor.ClearStaticCacheForTest();
+
+            foreach (var cb in s_resetCallbacks)
+                cb();
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed. Each callback in the `foreach` loop is now wrapped in a `try/catch`. Exceptions are collected into a `List<Exception>` and re-thrown as a single `AggregateException` after all callbacks have run, so one bad callback no longer prevents the others from executing. See commit 2d2c6f3ad.

gladjohn (approved):

PR #5888, opened 2026-03-26
Added support for WithArributeTokens 

4gust:
# Added support for `WithAttributeTokens` for CCA acquisition methods
This pull request adds support for specifying attribute tokens in token acquisition requests across multiple authentication flows in the Microsoft Identity Client library. It introduces new `WithAttributeTokens` methods to the parameter builders for authorization code, client credentials, and on-behalf-of flows, allowing developers to include custom attribute tokens in requests via the `attribute_tokens` parameter.

The most important changes include:

**New API Surface:**

* Added `WithAttributeTokens(IEnumerable<string> attributeTokens)` methods to `AcquireTokenByAuthorizationCodeParameterBuilder`, `AcquireTokenForClientParameterBuilder`, and `AcquireTokenOnBehalfOfParameterBuilder`, enabling the inclusion of attribute tokens in token requests. These methods validate input, join the tokens, and add them to the request body. [[1]](diffhunk://#diff-5197c1f0bc80f2a410918064894a2a01956c5d6426f84d680ff1492a2204f508R157-R181) [[2]](diffhunk://#diff-66e58df1ddbe7bfef74ce08f426e8ebbdb78e7258838c181d5d497bd6c7236eeR164-R189) [[3]](diffhunk://#diff-d75d5bb65b2da527bc52f06c06ffef51c55b90e988072a8cb56c337afff76dbcR166-R190)

* Updated the public API files for all supported frameworks to register the new `WithAttributeTokens` methods, ensuring they are part of the public API surface.

**OAuth2 Parameter Support:**

* Added a new constant `AttributeTokens` to the `OAuth2Parameter` class for use as the key in the request body parameter.

Fixes #

**Testing**
Unit test added.


src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
```diff
@@ -153,6 +154,31 @@ public AcquireTokenByAuthorizationCodeParameterBuilder WithCcsRoutingHint(string
...
         }
 
+        /// <summary>
+        /// Specifies attribute tokens to include in the token request.
+        /// The tokens are joined with spaces and sent as the <c>attribute_tokens</c> body parameter.
+        /// </summary>
+        /// <param name="attributeTokens">A list of attribute token strings to include in the request.</param>
+        /// <returns>The builder to chain method calls.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="attributeTokens"/> is null or contains no elements.</exception>
+        public AcquireTokenByAuthorizationCodeParameterBuilder WithAttributeTokens(IEnumerable<string> attributeTokens)
```
bgavrilMS: Should this be added to the base class ?

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
```diff
@@ -153,6 +154,31 @@ public AcquireTokenByAuthorizationCodeParameterBuilder WithCcsRoutingHint(string
...
+        {
+            if (attributeTokens == null)
+            {
+                throw new ArgumentNullException(nameof(attributeTokens));
+            }
+
+            string joinedTokens = string.Join(" ", attributeTokens);
+
+            this.OnBeforeTokenRequest((data) =>
```
bgavrilMS: This does not do the right thing in terms of caching.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
```diff
@@ -161,6 +161,44 @@ public AcquireTokenForClientParameterBuilder WithAttributes(string attributeJson
...
+        /// <summary>
+        /// Specifies attribute tokens to include in the token request.
+        /// The tokens are joined with spaces and sent as the <c>attribute_tokens</c> body parameter.
+        /// </summary>
+        /// <param name="attributeTokens">A list of attribute token strings to include in the request.</param>
+        /// <returns>The builder to chain method calls.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="attributeTokens"/> is null or contains no elements.</exception>
+        public AcquireTokenForClientParameterBuilder WithAttributeTokens(IEnumerable<string> attributeTokens)
+        {
+            if (attributeTokens == null)
```
bgavrilMS: There is a lot of duplicate code between these methods. In fact all 3 implemenations are different and 2 of them are wrong.

bgavrilMS: MSAL should be tollerant to null arguments. Just no-op. This provides a better dev ex.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
```diff
@@ -153,6 +154,31 @@ public AcquireTokenByAuthorizationCodeParameterBuilder WithCcsRoutingHint(string
...
+            {
+                data.BodyParameters.Add(OAuth2Parameter.AttributeTokens, joinedTokens);
```
bgavrilMS: Will there be enough logging if we go with this approach?

bgavrilMS (changes requested):
Too much code duplication. No token caching.

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenByAuthorizationCodeParameterBuilder.cs
4gust: updated this added caching test

src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForClientParameterBuilder.cs
4gust: updated

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +42,68 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+
+            string joinedTokens = string.Join(" ", normalizedTokens);
+
+            var extraBodyParams = new Dictionary<string, Func<CancellationToken, Task<string>>>
+            {
+                { OAuth2Parameter.AttributeTokens, _ => Task.FromResult(joinedTokens) }
+            };
+
+            builder.WithExtraBodyParametersInternal(extraBodyParams);
+
```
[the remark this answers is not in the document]
4gust: @copilot apply changes based on [this feedback](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5888#discussion_r3049701263)

tests/Microsoft.Identity.Test.Unit/PublicApiTests/WithAttributeTokensTests.cs
```diff
@@ -0,0 +1,714 @@
...
+                    .Create(ClientId)
+                    .WithAuthority(new Uri(ClientApplicationBase.DefaultAuthority), true)
+                    .WithClientSecret(TestConstants.ClientSecret)
+                    .WithHttpManager(httpManager)
+                    .WithExperimentalFeatures(true)
+                    .BuildConcrete();
+
+                httpManager.AddInstanceDiscoveryMockHandler();
+
+                var handler = httpManager.AddMockHandlerSuccessfulClientCredentialTokenResponseMessage(
```
Robbie-Microsoft, line 40: `WithAttributeTokens` does not call `ValidateUseOfExperimentalFeature()`, so `WithExperimentalFeatures(true)` is not required here. This call is vestigial from a prior implementation that routed through the experimental `WithExtraBodyParameters` API.

These calls appear on lines 39, 76, 112, 173, 209, 532, 550, 567, 589, 624, and 659. They should all be removed — leaving them in signals to readers (and future test authors) that `WithAttributeTokens` is experimental when it is not.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +42,86 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+        /// and are also included as a cache key component so that requests with different attribute tokens
+        /// are cached as distinct entries.
+        /// Null, empty, or whitespace-only token entries are ignored. A null or empty collection is a no-op.
+        /// </summary>
+        /// <typeparam name="T">The concrete builder type.</typeparam>
+        /// <param name="builder">The builder to chain options to.</param>
+        /// <param name="attributeTokens">A list of attribute token strings to include in the request. Individual tokens must not contain whitespace.</param>
+        /// <returns>The builder to chain method calls.</returns>
+        /// <exception cref="ArgumentException">Thrown when any token contains embedded whitespace.</exception>
+        public static AbstractAcquireTokenParameterBuilder<T> WithAttributeTokens<T>(
```
Robbie-Microsoft: The generic constraint `where T : AbstractAcquireTokenParameterBuilder<T>` does not restrict this method to confidential client builders. `WithAttributeTokens` is callable on public client flows such as `AcquireTokenInteractive`, `AcquireTokenByUsernamePassword`, etc.

If called from a public client builder, the `attribute_tokens` body parameter would be added to the token request (where the public-client endpoint would likely ignore it), and — more problematically — the attribute tokens would be added to the cache key via `WithAdditionalCacheKeyComponents`. This would break subsequent `AcquireTokenSilent` calls that don't include the same attribute tokens, causing unexpected cache misses.

Please either:
1. Add a XML doc note explicitly calling out this behavior is unsupported for public client flows, or
2. Restrict the constraint to `AbstractConfidentialClientAcquireTokenParameterBuilder<T>` if such a base class exists.

4gust:
@copilot resolve the merge conflicts in this pull request

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,104 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
             return builder;
         }
 
+        internal static AbstractAcquireTokenParameterBuilder<T> WithExtraBodyParametersInternal<T>(
```
bgavrilMS: This already exists as a public API: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/f13a02c325ddee2eca7135fcb720ee8493539c8c/src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs?plain=1#L52C84-L52C84

4gust: removed this and used `OnBeforeTokenRequest`

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,85 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+
+            var normalizedTokens = new List<string>();
+            foreach (string token in attributeTokens)
+            {
+                if (!string.IsNullOrWhiteSpace(token))
+                {
+                    string trimmed = token.Trim();
+                    if (trimmed.Any(char.IsWhiteSpace))
+                    {
+                        throw new ArgumentException(
```
bgavrilMS, line 100: It's not worth doing checks. Attribute tokens are signed by the STS. The STS will also validate them and error out.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,85 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+                logger.Info(() => $"[WithAttributeTokens] Attaching {count} attribute token(s) " +
+                                  "to the request body and including them in the cache key partition.");
+            }
+
+            var extraBodyParams = new Dictionary<string, Func<CancellationToken, Task<string>>>
+            {
+                { OAuth2Parameter.AttributeTokens, _ => Task.FromResult(joinedTokens) }
+            };
+
+            builder.OnBeforeTokenRequest(data =>
```
bgavrilMS: Use `WithExtraBodyParameters` ?

4gust: updated

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: removed

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,114 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+        /// <typeparam name="T">The concrete confidential client builder type.</typeparam>
+        /// <param name="builder">The builder to chain options to.</param>
+        /// <param name="extraBodyParams">List of additional body parameters.</param>
+        /// <returns>The builder to chain method calls.</returns>
+        public static AbstractConfidentialClientAcquireTokenParameterBuilder<T> WithExtraBodyParameters<T>(
+            this AbstractConfidentialClientAcquireTokenParameterBuilder<T> builder,
+            IDictionary<string, Func<CancellationToken, Task<string>>> extraBodyParams)
+            where T : AbstractConfidentialClientAcquireTokenParameterBuilder<T>
+        {
+            builder.ValidateUseOfExperimentalFeature();
```
gladjohn, line 149: Experimental feature gate is transitive: WithAttributeTokens delegates to WithExtraBodyParameters, which calls ValidateUseOfExperimentalFeature().

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,114 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+            {
+                if (logger != null && logger.IsLoggingEnabled(LogLevel.Verbose))
+                {
+                    logger.Verbose(() => "[WithAttributeTokens] collection contained no usable tokens.");
+                }
+
+                return builder;
+            }
+
+            string joinedTokens = string.Join(" ", normalizedTokens);
```
gladjohn, line 121: If the app developer passes {"B", "A"} on the first call and {"A", "B"} on the second, MSAL will produce cache keys "B A" and "A B", two different strings - the second call misses the cache and goes to the network even though the tokens are semantically
identical. 

how can we guarantee this does not happen and deliver a better devex here?

gladjohn: maybe update the xml doc?

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -49,27 +49,13 @@ public static AcquireTokenForClientParameterBuilder WithProofOfPosessionKeyId(
         /// <param name="builder"></param>
         /// <param name="extrabodyparams">List of additional body parameters</param>
         /// <returns></returns>
+        [EditorBrowsable(EditorBrowsableState.Never)] // Soft deprecate; prefer the generic overload on AbstractConfidentialClientAcquireTokenParameterBuilderExtension that also supports OBO and AuthorizationCode.
         public static AcquireTokenForClientParameterBuilder WithExtraBodyParameters(
             this AcquireTokenForClientParameterBuilder builder, 
             Dictionary<string, Func<CancellationToken, Task<string>>> extrabodyparams)
```
gladjohn, line 56: this used a `Dictionary`

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,114 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+        /// to associate these parameters with the acquired token. Works for confidential client flows
+        /// (AcquireTokenForClient, AcquireTokenOnBehalfOf, AcquireTokenByAuthorizationCode).
+        /// </summary>
+        /// <typeparam name="T">The concrete confidential client builder type.</typeparam>
+        /// <param name="builder">The builder to chain options to.</param>
+        /// <param name="extraBodyParams">List of additional body parameters.</param>
+        /// <returns>The builder to chain method calls.</returns>
+        public static AbstractConfidentialClientAcquireTokenParameterBuilder<T> WithExtraBodyParameters<T>(
+            this AbstractConfidentialClientAcquireTokenParameterBuilder<T> builder,
+            IDictionary<string, Func<CancellationToken, Task<string>>> extraBodyParams)
```
gladjohn: Inconsistent parameter types (Dictionary vs IDictionary), I tagged the old one for your reference.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,114 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+        /// a null or empty collection is a no-op.
+        /// </summary>
+        /// <typeparam name="T">The concrete confidential client builder type.</typeparam>
+        /// <param name="builder">The builder to chain options to.</param>
+        /// <param name="attributeTokens">Attribute tokens to include. Individual tokens must not contain whitespace.</param>
+        /// <returns>The builder to chain method calls.</returns>
+        /// <exception cref="ArgumentException">Thrown when any token contains embedded whitespace.</exception>
+        public static AbstractConfidentialClientAcquireTokenParameterBuilder<T> WithAttributeTokens<T>(
+            this AbstractConfidentialClientAcquireTokenParameterBuilder<T> builder,
+            IEnumerable<string> attributeTokens)
```
gladjohn, line 81: ### Type narrowing is lost in the fluent chain — consider returning `T` instead of the base type

`WithAttributeTokens<T>` and the new generic `WithExtraBodyParameters<T>` are constrained with:

```csharp
where T : AbstractConfidentialClientAcquireTokenParameterBuilder<T>
```

However, they return the base type:

```csharp
AbstractConfidentialClientAcquireTokenParameterBuilder<T>
```

instead of returning `T`.

This means the concrete builder type is erased mid-chain. As a result, concrete-only methods, such as `WithFmiPath` on `AcquireTokenForClientParameterBuilder`, become unavailable after calling `WithAttributeTokens`.

This compiles:

```csharp
app.AcquireTokenForClient(scopes)
   .WithFmiPath("path")            // concrete type ✅
   .WithAttributeTokens(tokens)    // base type — OK because it is last in the chain
   .ExecuteAsync();
```

This does not compile:

```csharp
app.AcquireTokenForClient(scopes)
   .WithAttributeTokens(tokens)    // returns base type ❌
   .WithFmiPath("path")            // CS1061: no WithFmiPath on base type
   .ExecuteAsync();
```

The test `WithAttributeTokens_ForClient_CombinedWithOtherMethods_Async` already works around this with the comment:

> `WithAttributeTokens` returns `AbstractAcquireTokenParameterBuilder<T>`, so concrete-builder methods like `WithFmiPath` must be called before it in the chain.

That ordering requirement is a usability trap. Fluent APIs are generally expected to be order-independent.

### Suggested fix

Return `T` and cast back to the concrete builder type. This is safe because of the F-bounded generic constraint:

```csharp
public static T WithAttributeTokens<T>(
    this AbstractConfidentialClientAcquireTokenParameterBuilder<T> builder,
    IEnumerable<string> attributeTokens)
    where T : AbstractConfidentialClientAcquireTokenParameterBuilder<T>
{
    ...
    builder.WithExtraBodyParameters(extraBodyParams);
    return (T)builder;
}
```

The same change should be applied to `WithExtraBodyParameters<T>`.

Since these entries are still in `PublicAPI.Unshipped.txt`, tightening the return type now should be low risk. The unshipped API files just need to be updated.

It would also be worth adding a regression test that chains the calls in the opposite order:

```csharp
app.AcquireTokenForClient(scopes)
   .WithAttributeTokens(tokens)
   .WithFmiPath("path")
   .ExecuteAsync();
```

This would lock in the expected fluent chaining behavior.

- commented by CoPilot

gladjohn (approved):
approved with comments

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: I will sort the token to generate the `joinedToken`

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,114 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+            where T : AbstractConfidentialClientAcquireTokenParameterBuilder<T>
+        {
+            ILoggerAdapter logger = builder.ServiceBundle?.ApplicationLogger;
+
+            if (attributeTokens is null)
+            {
+                if (logger != null && logger.IsLoggingEnabled(LogLevel.Verbose))
```
trwalke: Typically, we do a check to see if logging is enabled when we are needing to log something that requires a lot of computation so that we dont waste effort on creating the log message. However, in this case, the message is very simple so you can just log without checking.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +44,114 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+                            nameof(attributeTokens));
+                    }
+
+                    normalizedTokens.Add(trimmed);
+                }
+            }
+
+            if (normalizedTokens.Count == 0)
+            {
+                if (logger != null && logger.IsLoggingEnabled(LogLevel.Verbose))
```
trwalke: if statement isnt needed here for logging

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: noted.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: done

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: resolved this.=, added test for it.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
trwalke: @gladjohn Fortunately, when we created `WithAddtiionalCacheKeyComponenets`, we made sure to use a sorted list when creating the cache key so the elements of the cache key are always in the same order.

trwalke (approved):

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: We sort with keys, if the value is generated from has multiple values then we need to sort it.
In this case we have take multiple values for a single key.
Added a sorting before generating value

trwalke: Oh I see, it is nested in this case. Makes sense

Robbie-Microsoft (commented):
*Posted in error — please disregard.*

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +43,107 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+
+            if (normalizedTokens.Count == 0)
+            {
+                logger?.Verbose(() => "[WithAttributeTokens] collection contained no usable tokens.");
+                return (T)builder;
+            }
+
+            // Sort so that callers passing the same set of tokens in different orders
+            // hit the same cache entry and produce the same request body.
+            normalizedTokens.Sort(StringComparer.Ordinal);
```
Robbie-Microsoft, line 119: Duplicate tokens in the input (e.g. `["tokenA", "tokenA"]`) survive the sort and are joined as `"tokenA tokenA"`, which is a different cache key and request body than `"tokenA"`. Consider deduplicating before sorting:

```csharp
normalizedTokens = normalizedTokens.Distinct(StringComparer.Ordinal).ToList();
normalizedTokens.Sort(StringComparer.Ordinal);
```

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
```diff
@@ -42,6 +43,107 @@ public static AbstractAcquireTokenParameterBuilder<T> OnBeforeTokenRequest<T>(
...
+        /// <summary>
+        /// Sends <paramref name="attributeTokens"/> as the <c>attribute_tokens</c> body parameter
+        /// and includes them in the cache key. Null/empty/whitespace entries are ignored;
+        /// a null or empty collection is a no-op.
+        /// </summary>
+        /// <typeparam name="T">The concrete confidential client builder type.</typeparam>
+        /// <param name="builder">The builder to chain options to.</param>
+        /// <param name="attributeTokens">Attribute tokens to include. Individual tokens must not contain whitespace.</param>
+        /// <returns>The builder to chain method calls.</returns>
+        /// <exception cref="ArgumentException">Thrown when any token contains embedded whitespace.</exception>
```
Robbie-Microsoft, line 73: `WithAttributeTokens` delegates to `WithExtraBodyParameters`, which calls `ValidateUseOfExperimentalFeature()`. If the caller did not configure `.WithExperimentalFeatures()` on the app builder, this throws `MsalClientException` at runtime. The `<exception>` doc only covers `ArgumentException` — please also document the `MsalClientException` case.

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
```diff
@@ -49,27 +49,13 @@ public static AcquireTokenForClientParameterBuilder WithProofOfPosessionKeyId(
         /// <param name="builder"></param>
         /// <param name="extrabodyparams">List of additional body parameters</param>
         /// <returns></returns>
+        [EditorBrowsable(EditorBrowsableState.Never)] // Soft deprecate; prefer the generic overload on AbstractConfidentialClientAcquireTokenParameterBuilderExtension that also supports OBO and AuthorizationCode.
```
Robbie-Microsoft, line 52: `[EditorBrowsable(EditorBrowsableState.Never)]` hides this from IntelliSense but produces no compiler warning, so existing callers get no signal to migrate. Consider also adding `[Obsolete("Use AbstractConfidentialClientAcquireTokenParameterBuilderExtension.WithExtraBodyParameters instead.", false)]` to make the migration path discoverable without a build break.

Robbie-Microsoft:
**OBO cache-partition limitation is undocumented in the API**

The test `WithAttributeTokens_OnBehalfOf_SentInRequestBody_Async` includes a comment acknowledging that for OBO flows, different attribute-token sets may not coexist side-by-side in the same user token cache partition — meaning a call with tokens {X,Y} can silently evict the cached entry for {A,B}.

This is a significant behavioral footgun that isn't mentioned anywhere in the public XML docs for `WithAttributeTokens`. Callers expecting per-set cache isolation in OBO flows will be silently broken. The doc should at minimum warn that OBO cache isolation per attribute-token set is not guaranteed and that callers needing strict isolation should use separate `ConfidentialClientApplication` instances.

bgavrilMS:
Does this need to be closed @4gust ?

src/client/Microsoft.Identity.Client/Extensibility/AcquireTokenForClientBuilderExtensions.cs
4gust: Updated with this.

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: updated

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: added a `Distinct` check.

4gust:
@Robbie-Microsoft Good catch, I have updated the method documentation and added a few tests. I skipped one scenario due to a bug.

I’ve opened a tracking issue here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5963

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: resolved

src/client/Microsoft.Identity.Client/Extensibility/AbstractConfidentialClientAcquireTokenParameterBuilderExtension.cs
4gust: resolved

4gust:
@copilot resolve the merge conflicts in this pull request

gladjohn (approved):

PR #5901, opened 2026-04-03
Refactor: Simplify internal mTLS certificate passing and reduce complexity

gladjohn:
## Summary
 Reduces internal complexity around how `MtlsCertificate` (`X509Certificate2`) is carried, stored, and checked across CCA and MIA flows. No public API surface changes.
 
 ## Problem
 The `MtlsCertificate` property was threaded through many layers — constructors, parameter bags, request contexts — with redundant checks, mutable state, and parameter objects misused as return channels (set-then-null patterns). This made the mTLS PoP code difficult to reason
 about and maintain.
 
 ## Changes
 
 ### Phase 1 — Remove dead code & simplify executor layer
 - Removed dead `MtlsCertificate` property from `RequestContext` (declared but never assigned)
 - Changed `RequestContext` constructor from `X509Certificate2` to `bool isMtlsRequested`
 - Removed unused setter on `AuthenticationRequestParameters.MtlsCertificate`
 - Simplified `ClientCredentialRequest.ShouldUseCachedToken` to use `AuthScheme.KeyId` directly
 
 ### Phase 2 — Centralize cert+scheme dual-set pattern
 - Introduced `IMtlsCertificateProvider` interface on `MtlsPopAuthenticationOperation`
 - Made `AcquireTokenCommonParameters.MtlsCertificate` a smart getter that derives cert from the auth scheme (single source of truth)
 - Added `ApplyMtlsPopScheme(cert)` helper to eliminate 4 duplicated `new MtlsPopAuthenticationOperation(cert)` + `MtlsCertificate = cert` call sites
 
 ### Phase 3 — Eliminate MIA cert bubbling
 - Changed `AbstractManagedIdentity.AuthenticateAsync` to return `(ManagedIdentityResponse, X509Certificate2)` tuple — cert flows through return value instead of parameter mutation
 - Removed `AcquireTokenForManagedIdentityParameters.MtlsCertificate` (was misused as a return channel)
 - Removed redundant `IsMtlsPopRequested` copy at request time (already set by builder)
 - Moved `_isMtlsPopRequested` field from base `AbstractManagedIdentity` to `ImdsV2ManagedIdentitySource` (only consumer)
 
 ## Testing
 - **0 Failed, 1855 Passed, 19 Skipped** (net48) — matches baseline


src/client/Microsoft.Identity.Client/Internal/ClientCredential/CredentialMaterialResolver.cs
```diff
@@ -0,0 +1,108 @@
...
+                    : OAuthMode.Regular,
+                Claims = requestParams.Claims,
+                ClientCapabilities = requestParams.AppConfig.ClientCapabilities,
+                CryptographyManager = requestParams.RequestContext.ServiceBundle.PlatformProxy.CryptographyManager,
+                SendX5C = requestParams.SendX5C,
+                UseSha2 = requestParams.AuthorityManager.Authority.AuthorityInfo.IsSha2CredentialSupported,
+                ExtraClientAssertionClaims = requestParams.ExtraClientAssertionClaims,
+                ClientAssertionFmiPath = requestParams.ClientAssertionFmiPath,
+                Authority = requestParams.AuthorityInfo.CanonicalAuthority.Authority,
+                TenantId = requestParams.AuthorityManager.Authority.TenantId,
```
Robbie-Microsoft, line 103: `CredentialContext` is missing a `CorrelationId` field. `AssertionRequestOptions` has a public `CorrelationId` property that callers rely on for end-to-end tracing (e.g., propagating the same correlation ID to downstream Managed Identity token requests). It was previously populated by the now-removed internal constructor. Without this field, delegates receive `Guid.Empty` instead of the request's correlation ID.

Please add `CorrelationId = requestParams.RequestContext.CorrelationId` to both the `CredentialContext` struct definition and here in `BuildContext`, then thread it through to `AssertionRequestOptions` in both `ClientAssertionDelegateCredential` and `ClientAssertionStringDelegateCredential`.

tests/Microsoft.Identity.Test.Unit/RequestsTests/CredentialMatrixTests.cs
```diff
@@ -0,0 +1,434 @@
...
+
+        // ──────────────────────────────────────────────
+        // Row 4 – Secret + MtlsMode  (unsupported)
+        // ──────────────────────────────────────────────
+
+        [TestMethod]
+        public async Task Row4_SecretCredential_MtlsMode_ThrowsMsalClientExceptionAsync()
+        {
+            var credential = new ClientSecretCredential("my-secret");
+            MsalClientException ex = await Assert.ThrowsExactlyAsync<MsalClientException>(
```
Robbie-Microsoft, line 208: Repo convention is `AssertException.TaskThrowsAsync<MsalClientException>(...)` which also validates the error code inline via a callback, giving clearer failure messages than a separate `Assert.AreEqual` line. `Assert.ThrowsExactlyAsync` is MSTest 4.x-specific and inconsistent with the rest of the test suite.

The same pattern applies to the other `MtlsMode` exception tests at lines 242, 287, 378, 391, and 404.

tests/Microsoft.Identity.Test.Unit/RequestsTests/CredentialMatrixTests.cs
```diff
@@ -0,0 +1,434 @@
...
+        [ClassCleanup]
+        public static void ClassCleanup()
+        {
+        }
+
+        // ──────────────────────────────────────────────
+        // Helpers
+        // ──────────────────────────────────────────────
+
+        private static CredentialContext RegularContext() => new CredentialContext
```
Robbie-Microsoft, line 57: Both `RegularContext()` and `MtlsContext()` (line 72) leave `Authority = null` and `TenantId = null`. No test currently verifies that `context.Authority` and `context.TenantId` are propagated to `AssertionRequestOptions` in the delegate credential tests (e.g., Rows 7 and 9).

Once the `CanonicalAuthority.Authority` bug in `CredentialMaterialResolver` is fixed, consider adding a test that captures the `AssertionRequestOptions` passed to the delegate and asserts `opts.Authority` is non-null and correct.

PR #5906, opened 2026-04-03
fix: make System.ValueTuple conditional on net462 only (#5894)

### LINKED ISSUE #5894: [Bug] .NET 4.81 MVC Web solution : Nuget Package dependency on ValueTuple causes runtime failure
### Library version used

4.83.3

### .NET version

4.81

### Scenario

Why is it that all of a sudden we started seeing package ValueTuple appearing in Web.Config when we do a Nuget Package updates for Identity.Client or Formats.Asn1 !?

It causes Web App runtime failure and the only solution is that we REMOVE Manually from Web.Config these lines

      <dependentAssembly>
        <assemblyIdentity name="System.ValueTuple" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
      </dependentAssembly>

If we try to remove this package from the Solution, we receive an error 

**Unable to uninstall 'System.ValueTuple.4.6.2' because 'Microsoft.Identity.Client.4.83.3, System.Formats.Asn1.10.0.5' depend on it.**

When we researched the subject on GPT / google, we found that this package is part of the FRAMEWORK and should not be mentioned in Web.Config, so why Nuget Package Microsoft.Identity.Client refer to it !!!

This is breaking our CDN pipeline as we have to do manual intervention. It is really frustrating.

Please consider !!!

### Relevant code snippets

```csharp

```

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

gladjohn:
System.ValueTuple was an unconditional PackageReference, causing it to appear in the NuGet dependency group for all TFMs including net472. On .NET Framework 4.7+ ValueTuple is in-box, so the NuGet package generates harmful binding redirects in Web.Config that cause runtime failures.

Changes:
- Microsoft.Identity.Client.csproj: restrict direct ValueTuple ref to net462 (the only TFM that needs it) and suppress the transitive dep from System.Formats.Asn1 on net472 via PrivateAssets/ExcludeAssets.
- Remove unnecessary ValueTuple refs from Microsoft.Identity.Lab.Api and Microsoft.Identity.Test.Common (both netstandard2.0-only).

Fixes #5894

# 📦 PR #5906 — NuSpec Verification Report

## `System.ValueTuple` Conditional Fix

> **PR:** [fix: make System.ValueTuple conditional on net462 only (#5894)](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5906)
> **Date:** April 30, 2026
> **Package:** `Microsoft.Identity.Client 4.61.0`

---

## 🔍 NuSpec Dependency Diff

Generated `.nupkg` from both `main` and PR branch, extracted `.nuspec`, and compared dependency groups.

### `System.ValueTuple` presence per TFM

| Target Framework | main (before) | PR #5906 (after) | Status |
|:-----------------|:--------------|:-----------------|:-------|
| `.NETFramework4.6.2` | `System.ValueTuple 4.5.0` | `System.ValueTuple 4.5.0` | 🟢 Kept — not in-box on 4.6.2 |
| `.NETFramework4.7.2` | `System.ValueTuple 4.5.0` ⚠️ | **Removed** | 🟢 Fixed — in-box on 4.7+ |
| `net8.0` | `System.ValueTuple 4.5.0` ⚠️ | **Removed** | 🟢 Fixed — in-box on .NET Core+ |
| `.NETStandard2.0` | absent | absent | 🟡 No change |

### Full dependency groups (after fix)

```xml
<!-- net462 — ValueTuple KEPT (needed) -->
<group targetFramework=".NETFramework4.6.2">
  <dependency id="Microsoft.IdentityModel.Abstractions" version="8.14.0" />
  <dependency id="System.Diagnostics.DiagnosticSource" version="6.0.1" />
  <dependency id="System.Formats.Asn1" version="8.0.1" />
  <dependency id="System.ValueTuple" version="4.5.0" />     ✅
</group>

<!-- net472 — ValueTuple REMOVED (in-box) -->
<group targetFramework=".NETFramework4.7.2">
  <dependency id="Microsoft.IdentityModel.Abstractions" version="8.14.0" />
  <dependency id="System.Diagnostics.DiagnosticSource" version="6.0.1" />
  <dependency id="System.Formats.Asn1" version="8.0.1" />
  <!-- No System.ValueTuple — in-box on .NET Framework 4.7+ -->  ✅
</group>

<!-- net8.0 — ValueTuple REMOVED (in-box) -->
<group targetFramework="net8.0">
  <dependency id="Microsoft.IdentityModel.Abstractions" version="8.14.0" />
  <dependency id="System.Diagnostics.DiagnosticSource" version="6.0.1" />
  <!-- No System.ValueTuple — in-box on .NET Core+ -->  ✅
</group>
```

---

image

### DIFF
--- a/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
+++ b/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
@@ -154,7 +154,17 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.IdentityModel.Abstractions" />
+  </ItemGroup>
+
+  <!-- System.ValueTuple is in-box starting with .NET Framework 4.7; only net462 needs the package. -->
+  <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop462)'">
     <PackageReference Include="System.ValueTuple" />
+  </ItemGroup>
+
+  <!-- Suppress transitive System.ValueTuple from System.Formats.Asn1 on net472+, where it is in-box
+       and the NuGet package causes harmful binding redirects (see #5894). -->
+  <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop472)'">
+    <PackageReference Include="System.ValueTuple" PrivateAssets="all" ExcludeAssets="all" />
   </ItemGroup>
 
   <ItemGroup Label="For public api analyzer support">
--- a/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
+++ b/src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
@@ -164,7 +164,7 @@
   <!-- Suppress transitive System.ValueTuple from System.Formats.Asn1 on net472+, where it is in-box
        and the NuGet package causes harmful binding redirects (see #5894). -->
   <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop472)'">
-    <PackageReference Include="System.ValueTuple" PrivateAssets="all" ExcludeAssets="all" />
+    <PackageReference Include="System.ValueTuple" PrivateAssets="All" ExcludeAssets="All" />
   </ItemGroup>
 
   <ItemGroup Label="For public api analyzer support">
--- a/src/client/Microsoft.Identity.Lab.Api/Microsoft.Identity.Lab.Api.csproj
+++ b/src/client/Microsoft.Identity.Lab.Api/Microsoft.Identity.Lab.Api.csproj
@@ -37,7 +37,6 @@
     <PackageReference Include="System.Threading.Tasks" />
     <PackageReference Include="System.Threading.Tasks.Parallel" />
     <PackageReference Include="System.Threading.Thread" />
-    <PackageReference Include="System.ValueTuple" />
   </ItemGroup>
 
   <ItemGroup>
--- a/tests/Microsoft.Identity.Test.Common/Microsoft.Identity.Test.Common.csproj
+++ b/tests/Microsoft.Identity.Test.Common/Microsoft.Identity.Test.Common.csproj
@@ -20,7 +20,6 @@
     <PackageReference Include="System.Threading.Tasks" />
     <PackageReference Include="System.Threading.Tasks.Parallel" />
     <PackageReference Include="System.Threading.Thread" />
-    <PackageReference Include="System.ValueTuple" />
   </ItemGroup>
 
   <ItemGroup>


src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -154,9 +154,19 @@
...
+
+  <!-- System.ValueTuple is in-box starting with .NET Framework 4.7; only net462 needs the package. -->
+  <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop462)'">
     <PackageReference Include="System.ValueTuple" />
   </ItemGroup>
 
+  <!-- Suppress transitive System.ValueTuple from System.Formats.Asn1 on net472+, where it is in-box
+       and the NuGet package causes harmful binding redirects (see #5894). -->
+  <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop472)'">
+    <PackageReference Include="System.ValueTuple" PrivateAssets="all" ExcludeAssets="all" />
```
Robbie-Microsoft: Is ExcludeAssets="all" redundant? PrivateAssets="all" already prevents the package from flowing to consumers as a dependency. ExcludeAssets="all" prevents the assets from being used in the current project too.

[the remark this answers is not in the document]
Robbie-Microsoft: ExcludeAssets casing also inconsistent - should be "All" to match PrivateAssets

Robbie-Microsoft (commented):
No test validates the .nupkg metadata to confirm ValueTuple no longer appears in net472 dependency group. The fix is correct in principle, but the only way to really confirm it worked is to inspect the generated .nupkg's .nuspec. Have you manually verified the fix?

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
```diff
@@ -154,9 +154,19 @@
...
   <ItemGroup>
     <PackageReference Include="Microsoft.IdentityModel.Abstractions" />
+  </ItemGroup>
+
+  <!-- System.ValueTuple is in-box starting with .NET Framework 4.7; only net462 needs the package. -->
+  <ItemGroup Condition="'$(TargetFramework)' == '$(TargetFrameworkNetDesktop462)'">
     <PackageReference Include="System.ValueTuple" />
   </ItemGroup>
 
+  <!-- Suppress transitive System.ValueTuple from System.Formats.Asn1 on net472+, where it is in-box
```
bgavrilMS, line 164: I don't fully understand the trick you did here. What would happen if you let net47 use System.Tuple that it comes with?

gladjohn: net472 does use the in-box ValueTuple, which is good. But the problem we are dealing - is that System.Formats.Asn1 has a transitive dependency on the System.ValueTuple NuGet package, so even without us referencing it directly, NuGet still lists it in MSAL's net472 dependency group. That causes consuming apps to generate a binding redirect to the package's System.ValueTuple.dll, which conflicts with the in-box type and breaks at runtime (#5894).

src/client/Microsoft.Identity.Client/Microsoft.Identity.Client.csproj
gladjohn: Good obersvbation -  they're not redundant, they do different things:

- PrivateAssets="all" → stops the dependency from flowing to consumers of MSAL (keeps it out of the .nuspec).
- ExcludeAssets="all" → stops the package's assets from being consumed in this project's build/runtime, so we actually use the in-box ValueTuple on net472 instead of the package's DLL.

Without ExcludeAssets, MSAL itself would still bind against the package's System.ValueTuple.dll at build time on net472, which is exactly what we're trying to avoid.

Robbie-Microsoft (approved):

PR #5907, opened 2026-04-03
Add 63 unit tests for parsing/logic edge cases in IdToken, WsTrust, AuthorizationResult, ThrottlingCache, Base64Url

gladjohn:
## Summary
 
 Adds 63 unit tests across 5 test files targeting real parsing and logic edge cases in `Microsoft.Identity.Client`. Each test validates specific behavioral contracts that would catch actual regressions.
 
 Also marks `ConcurrentHashSet<T>` (copied from [i3arnon/ConcurrentHashSet](https://github.com/i3arnon/ConcurrentHashSet)) as `[ExcludeFromCodeCoverage]` since it's external code that shouldn't affect MSAL coverage metrics.
 
 ## Test Files
 
 | File | Tests | Target | Key Scenarios |
 |---|---:|---|---|
 | `WsTrustResponseExtendedTests` | 13 | `WsTrustResponse` | SOAP fault parsing, missing body/fault fallback, WsTrust13 vs 2005, SAML1 preference, skipped responses |
 | `AuthorizationResultExtendedTests` | 18 | `AuthorizationResult` | Null/empty URI, error subcodes, `msauth://` scheme, `FromPostData`, `FromStatus` factory |
 | `IdTokenExtendedTests` | 11 | `IdToken` | Null/empty/malformed JWT, claim mapping, `GetUniqueId` fallback (oid→sub), missing issuer, `ClaimsPrincipal` |
 | `Base64UrlHelpersExtendedTests` | 13 | `Base64UrlHelpers` | Null handling, URL-safe char validation, padding variants, invalid input, padded base64 decoding |
 | `ThrottlingCacheExtendedTests` | 8 | `ThrottlingCache` | Clear/IsEmpty, newer entry replacement, older-doesn't-replace-newer, cleanup of expired entries |
 
 ## Production Code Change
 
 - `ConcurrentHashSet.cs`: Added `[ExcludeFromCodeCoverage]` attribute (class is copied from external OSS repo)
 
 ## Validation
 
 - ✅ Build: 0 errors, 0 warnings
 - ✅ Tests: All 63 new tests pass on both net48 and net8.0


tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        [TestMethod]
+        public void Encode_SimpleString_NoSpecialChars()
+        {
+            string result = Base64UrlHelpers.Encode("Hello");
+            Assert.IsNotNull(result);
+            // Should not contain +, /, or =
+            Assert.DoesNotContain("+", result);
+            Assert.DoesNotContain("/", result);
+            Assert.DoesNotContain("=", result);
```
Robbie-Microsoft, line 48: I believe these arguments are reversed. The assertion always passes and doesn't actually validate URL-safety.

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        [TestMethod]
+        public void Encode_BytesThatProducePlusAndSlash()
+        {
+            // 0xFB, 0xEF, 0xBE => produces + and / in standard base64
+            byte[] data = { 0xFB, 0xEF, 0xBE };
+            string encoded = Base64UrlHelpers.Encode(data);
+            Assert.DoesNotContain("+", encoded);
+            Assert.DoesNotContain("/", encoded);
```
Robbie-Microsoft, line 77: I believe these arguments are reversed. The assertion always passes and doesn't actually validate URL-safety.

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenExtendedTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            try
+            {
+                IdToken.Parse("singlesegment");
+                Assert.Fail("Should have thrown");
+            }
+            catch (MsalClientException ex)
+            {
+                Assert.AreEqual(MsalError.InvalidJwtError, ex.ErrorCode);
+            }
+        }
```
Robbie-Microsoft: These test the same thing in two separate methods - could be one test

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenExtendedTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                Assert.Fail("Should have thrown an exception");
+            }
+            catch (MsalClientException)
+            {
+                // Expected - JSON parse error
+            }
+            catch (FormatException)
+            {
+                // Expected - invalid base64
+            }
```
Robbie-Microsoft: Why is this catching both MsalClientException and FormatException? Asserting which one is expected - the test passes regardless of which exception is thrown, making it weak.

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
gladjohn: I think the current order is correct for MSTest: Assert.DoesNotContain(string substring, string value). So Assert.DoesNotContain("+", result) is checking that result does not contain "+".

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
gladjohn: I think the current order is correct for MSTest: Assert.DoesNotContain(string substring, string value). So Assert.DoesNotContain("+", result) is checking that result does not contain "+".

tests/Microsoft.Identity.Test.Unit/Throttling/ThrottlingCacheExtendedTests.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            var cache = new ThrottlingCache();
+            var olderEntry = new ThrottlingCacheEntry(_ex1, TimeSpan.FromMinutes(10));
+
+            // Small delay to ensure different creation times
+            Thread.Sleep(10);
+            var newerEntry = new ThrottlingCacheEntry(_ex2, TimeSpan.FromMinutes(10));
+
+            cache.AddAndCleanup("k1", olderEntry, _logger);
+            cache.AddAndCleanup("k1", newerEntry, _logger);
```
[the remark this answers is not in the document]
gladjohn: Thanks for flagging this! This has been addressed — the test now uses the `ThrottlingCacheEntry` constructor with explicit `DateTimeOffset` creation times to deterministically set ordering, no `Thread.Sleep` needed. :slightly_smiling_face:

— _Response by Copilot_

tests/Microsoft.Identity.Test.Unit/Throttling/ThrottlingCacheExtendedTests.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            var expiredEntry = new ThrottlingCacheEntry(_ex1, TimeSpan.FromMilliseconds(-10000));
+            var validEntry = new ThrottlingCacheEntry(_ex2, TimeSpan.FromMinutes(10));
+
+            cache.AddAndCleanup("expired", expiredEntry, _logger);
+            cache.AddAndCleanup("valid", validEntry, _logger);
+
+            // Wait for cleanup interval to pass, then trigger cleanup via another add
+            Thread.Sleep(50);
+            cache.AddAndCleanup("trigger", new ThrottlingCacheEntry(_ex2, TimeSpan.FromMinutes(10)), _logger);
```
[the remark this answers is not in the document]
gladjohn: Good catch! This has been addressed — the test now uses `ThrottlingCache(0)` for a zero-interval cleanup trigger and `TimeSpan.FromMilliseconds(-10000)` for an already-expired entry, making it fully deterministic with no `Thread.Sleep`. :slightly_smiling_face:

— _Response by Copilot_

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
gladjohn: Hi Robbie! The current argument order is actually correct for MSTest v4. Per the [official docs](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.assert.doesnotcontain?view=mstest-net-4.1), the signature is `Assert.DoesNotContain(string substring, string value)` — substring first, value second. So `Assert.DoesNotContain('+', result)` correctly checks that `result` does not contain `'+'`. This is different from the older `StringAssert.DoesNotContain` which had the parameters in the opposite order — easy to mix up! :slightly_smiling_face:

— _Response by Copilot_

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
gladjohn: Same as above — the argument order here is correct for MSTest v4's `Assert.DoesNotContain(substring, value)` signature. `Assert.DoesNotContain('+', encoded)` correctly validates that the encoded output doesn't contain `'+'`. :slightly_smiling_face:

— _Response by Copilot_

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenExtendedTests.cs
gladjohn: Great suggestion! Merged both tests into a single `Parse_SingleSegment_ThrowsMsalClientExceptionWithCorrectErrorCode` that uses `Assert.Throws<MsalClientException>` to capture the exception and then verifies the error code — clean and concise. Pushed in the latest commit. Thanks, Robbie! :slightly_smiling_face:

— _Response by Copilot_

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenExtendedTests.cs
gladjohn: Good point! Traced through the code — `!!!invalid!!!` has length 13 (mod 4 == 1), so `Base64UrlHelpers.DecodeBytes` throws `FormatException` directly before JSON parsing is ever attempted. Updated the test to use `Assert.Throws<FormatException>` for a precise assertion. Thanks for the feedback, Robbie! :slightly_smiling_face:

— _Response by Copilot_

tests/Microsoft.Identity.Test.Unit/Throttling/ThrottlingCacheExtendedTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            cache.AddAndCleanup("k1", new ThrottlingCacheEntry(_ex1, TimeSpan.FromMinutes(10)), _logger);
+            Assert.HasCount(1, cache.CacheForTest);
+        }
+
+        [TestMethod]
+        public void Cleanup_RemovesOnlyExpiredEntries()
+        {
+            // Use a cleanup interval of 0 so cleanup triggers immediately on next add
+            var cache = new ThrottlingCache(0);
```
Robbie-Microsoft: The test uses `new ThrottlingCache(0)` (0ms interval), but `CleanCache` likely uses a strict `<` check (`_lastCleanupTime + interval < DateTimeOffset.UtcNow`). With a 0ms interval that becomes `_lastCleanupTime < UtcNow` — if two `AddAndCleanup` calls happen within the same clock tick, cleanup won't fire and the `expired` entry won't be removed, causing a false-fail.

Consider using a negative interval (e.g., `new ThrottlingCache(-1)`) to guarantee cleanup always triggers, making the test fully deterministic.

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
```diff
@@ -0,0 +1,212 @@
...
+            Assert.AreEqual(original, decoded);
+        }
+
+        [TestMethod]
+        public void Decode_NoPaddingNeeded()
+        {
+            // "AAAA" is valid base64 (length % 4 == 0), no URL-safe chars
+            byte[] result = Base64UrlHelpers.DecodeBytes("AAAA");
+            Assert.IsNotNull(result);
+            Assert.HasCount(3, result as System.Collections.IEnumerable);
```
Robbie-Microsoft: `Assert.HasCount(3, result as System.Collections.IEnumerable)` is unnecessarily complex. The `as` cast to the non-generic `IEnumerable` is redundant here — if the cast ever silently returned `null`, `HasCount` would give a misleading failure.

Since `result` is a `byte[]`, prefer the simpler and type-safe:
```csharp
Assert.AreEqual(3, result.Length);
```

bgavrilMS (changes requested):
I don't think this is a good idea. It artificially increases code coverage, but do you know they capture the correct behavior?

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
```diff
@@ -0,0 +1,212 @@
...
+
+using System;
+using System.Text;
+using Microsoft.Identity.Client.Utils;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.UtilTests
+{
+    [TestClass]
+    public class Base64UrlHelpersExtendedTests: TestBase
```
bgavrilMS, line 12: these ones look ok.

tests/Microsoft.Identity.Test.Unit/UtilTests/ConcurrentHashSetTests.cs
```diff
@@ -0,0 +1,408 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Identity.Client.Utils;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.UtilTests
+{
+    [TestClass]
+    public class ConcurrentHashSetTests : TestBase
```
bgavrilMS: You copied this class from https://github.com/i3arnon/ConcurrentHashSet/blob/main/src/ConcurrentHashSet/ConcurrentHashSet.cs

You should copy the unit tests from there if you want code coverage OR exclude this class from coverage

bgavrilMS: Or use a collection from the framework. Better collections were added to .NET . The AI suggests to use ConcurrentDictionary

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenExtendedTests.cs
```diff
@@ -0,0 +1,149 @@
...
+using System;
+using System.Text;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Internal;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.CoreTests
+{
+    [TestClass]
+    public class IdTokenExtendedTests : TestBase
```
bgavrilMS, line 14: these look ok

tests/Microsoft.Identity.Test.Unit/CacheTests/TokenCacheNotificationArgsExtendedTests.cs
```diff
@@ -0,0 +1,161 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Identity.Client;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Identity.Test.Unit.CacheTests
+{
+    [TestClass]
+    public class TokenCacheNotificationArgsExtendedTests : TestBase
```
bgavrilMS: these look ok

tests/Microsoft.Identity.Test.Unit/Throttling/ThrottlingCacheExtendedTests.cs
gladjohn: Fixed, changed to new ThrottlingCache(-1) so cleanup triggers unconditionally regardless of clock tick resolution

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
gladjohn: removed the unnecessary as System.Collections.IEnumerable cast

tests/Microsoft.Identity.Test.Unit/UtilTests/Base64UrlHelpersExtendedTests.cs
gladjohn: Thanks! Also trimmed 9 redundant round-trip tests and fixed the padding test to use literal padded input ("SGVsbG8=") instead of re-encoding

tests/Microsoft.Identity.Test.Unit/UtilTests/ConcurrentHashSetTests.cs
gladjohn: Agreed on both points. Removed the test file entirely tjhat was testing a copied OSS class doesn't validate MSAL behavior. Also added [ExcludeFromCodeCoverage] to ConcurrentHashSet<T> so it doesn't affect coverage metrics.

tests/Microsoft.Identity.Test.Unit/CoreTests/IdTokenExtendedTests.cs
gladjohn: Thanks! These are kept as-is, they test real JWT parsing edge cases.

tests/Microsoft.Identity.Test.Unit/CacheTests/TokenCacheNotificationArgsExtendedTests.cs
gladjohn: On second look, these were just testing that constructors set properties and not validating real behavior. Removed the file.

gladjohn:
> I don't think this is a good idea. It artificially increases code coverage, but do you know they capture the correct behavior?

Fair point. Dropped ~55 tests that artificially inflated coverage (ConcurrentHashSet, TokenCacheNotificationArgs constructors, TraceTelemetryConfig constants, redundant Base64Url round-trips). The remaining ~63 tests all validate real parsing/logic: JWT claim mapping and fallback, SOAP fault parsing, authorization URI handling, and throttling cache semantics. Also added [ExcludeFromCodeCoverage] to ConcurrentHashSet<T> since it's copied from i3arnon/ConcurrentHashSet.

bgavrilMS (approved):

gladjohn:
@trwalke can I get a review on this?

Robbie-Microsoft (changes requested):
The only production code change in this PR — adding [ExcludeFromCodeCoverage] to ConcurrentHashSet.cs — will conflict with PR #5975, which deletes that file entirely in favor of .NET's built-in ConcurrentDictionary<T, byte>.

Can you drop the ConcurrentHashSet.cs change from this PR (the tests are the real value here and stand on their own)?

gladjohn:
> The only production code change in this PR — adding [ExcludeFromCodeCoverage] to ConcurrentHashSet.cs — will conflict with PR #5975, which deletes that file entirely in favor of .NET's built-in ConcurrentDictionary<T, byte>.
> 
> Can you drop the ConcurrentHashSet.cs change from this PR (the tests are the real value here and stand on their own)?

@copilot - Revert the change to ConcurrentHashSet.cs on this PR branch (gladjohn/code_coverage) so the file matches main exactly. Remove the [ExcludeFromCodeCoverage] attribute and any using System.Diagnostics.CodeAnalysis; import that was added for it. After the revert, the PR should contain only the 5 new test file

Robbie-Microsoft (approved):

PR #5916, opened 2026-04-09
Make AKV call failures inconclusive instead of failing the build

gladjohn:
The AcquireTokenAndCallAKV_OnImdsV2_MtlsPoP_WithAttestation_Succeeds E2E test has been intermittently failing due to AKV returning 401 Unauthorized with "Certificate import or verification failed." — an infrastructure issue unrelated to MSAL.

This PR changes the AKV call handling so that:

 - All non-success HTTP responses from AKV are marked as Assert.Inconclusive (previously Assert.IsTrue which failed the build)
 - HttpRequestException during the AKV call is also marked Assert.Inconclusive (previously Assert.Fail)
 - The inconclusive message includes status code and response body for visibility in ADO test results

What is NOT changed: mTLS PoP token acquisition assertions remain hard failures — only the downstream AKV secret call is treated as inconclusive, since it validates infrastructure rather than MSAL logic.

### DIFF
--- a/tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsV2Tests.cs
+++ b/tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsV2Tests.cs
@@ -201,7 +201,7 @@ namespace Microsoft.Identity.Test.E2E
             }
             catch (HttpRequestException ex)
             {
-                Assert.Fail($"AKV call failed: {ex.Message}");
+                Assert.Inconclusive($"AKV call could not be completed: {ex.Message}");
             }
         }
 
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsV2Tests.cs
```diff
@@ -362,15 +362,13 @@ private static async Task CallAkvSecretAsync(
             using var response = await httpClient.GetAsync(new Uri(secretUrl)).ConfigureAwait(false);
             var responseContent = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 
-            if ((int)response.StatusCode >= 500)
+            if (!response.IsSuccessStatusCode)
             {
+                // AKV failures are infrastructure-related and should not fail the build.
+                // The core test validates mTLS PoP token acquisition; AKV is a downstream call.
```
neha-bhargava, line 368: Why have this call in E2E tests then?

neha-bhargava (approved):

tests/Microsoft.Identity.Test.E2e/ManagedIdentityImdsV2Tests.cs
gladjohn: We still need this in E2E because, right now, our team is the only one with the test infrastructure for this partner onboarding flow. So this test is helping validate the broader end-to-end scenario and gives us useful signal to share back with partners when something fails.

4gust (approved):

PR #5920, opened 2026-04-10
Add MsalRemainingTokenLifetime histogram for token expiry 

### LINKED ISSUE #5889: [Feature Request] MSAL telemetry enhancement
### MSAL client type

Confidential, Managed identity

### Problem statement

[Product Backlog Item 3554274](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3554274): [MSAL.NET Telemetry] Add time‑to‑expiry (RemainingTimeOfToken) metric

### Proposed solution

_No response_

### Alternatives

_No response_

neha-bhargava:
Add new OTel histogram (MsalRemainingTokenLifetime.1A) that records the remaining lifetime of tokens in seconds at the time of acquisition.

Tags (6/8): MsalVersionPlatform (combined), ApiId, TokenSource, CacheLevel, CacheRefreshReason, TokenType.

Also adds TelemetryTokenTypeConstants.ToDisplayString() helper for int-to- string token type mapping used by the new histogram.

Resolves: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/5889
PBI: 3554274

Fixes #

**Testing**

**Performance impact**

**Documentation**

### DIFF
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -131,6 +131,15 @@ namespace Microsoft.Identity.Client.Internal.Requests
                         durationInUs,
                         authenticationResult.AuthenticationResultMetadata,
                         AuthenticationRequestParameters.RequestContext.Logger);
+
+            ServiceBundle.PlatformProxy.OtelInstrumentation.LogRemainingTokenLifetime(
+                        ServiceBundle.PlatformProxy.GetProductName(),
+                        apiEvent.ApiId,
+                        authenticationResult.AuthenticationResultMetadata.TokenSource,
+                        GetCacheLevel(authenticationResult),
+                        authenticationResult.AuthenticationResultMetadata.CacheRefreshReason,
+                        authenticationResult.AuthenticationResultMetadata.TelemetryTokenType,
+                        authenticationResult.ExpiresOn);
         }
 
         private void LogFailureTelemetryToOtel(string errorCodeToLog, ApiEvent apiEvent, CacheRefreshReason cacheRefreshReason)
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -129,24 +129,18 @@ namespace Microsoft.Identity.Client.Internal.Requests
 
         private void LogSuccessTelemetryToOtel(AuthenticationResult authenticationResult, ApiEvent apiEvent, long durationInUs)
         {
+            CacheLevel cacheLevel = GetCacheLevel(authenticationResult);
+
             // Log metrics
             ServiceBundle.PlatformProxy.OtelInstrumentation.LogSuccessMetrics(
                         ServiceBundle.PlatformProxy.GetProductName(),
                         apiEvent.ApiId,
                         apiEvent.CallerSdkApiId,
                         apiEvent.CallerSdkVersion,
-                        GetCacheLevel(authenticationResult),
+                        cacheLevel,
                         durationInUs,
                         authenticationResult.AuthenticationResultMetadata,
-                        AuthenticationRequestParameters.RequestContext.Logger);
-
-            ServiceBundle.PlatformProxy.OtelInstrumentation.LogRemainingTokenLifetime(
-                        ServiceBundle.PlatformProxy.GetProductName(),
-                        apiEvent.ApiId,
-                        authenticationResult.AuthenticationResultMetadata.TokenSource,
-                        GetCacheLevel(authenticationResult),
-                        authenticationResult.AuthenticationResultMetadata.CacheRefreshReason,
-                        authenticationResult.AuthenticationResultMetadata.TelemetryTokenType,
+                        AuthenticationRequestParameters.RequestContext.Logger,
                         authenticationResult.ExpiresOn);
         }
 
--- a/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
@@ -139,6 +139,15 @@ namespace Microsoft.Identity.Client.Internal.Requests
                         durationInUs,
                         authenticationResult.AuthenticationResultMetadata,
                         AuthenticationRequestParameters.RequestContext.Logger);
+
+            ServiceBundle.PlatformProxy.OtelInstrumentation.LogRemainingTokenLifetime(
+                        ServiceBundle.PlatformProxy.GetProductName(),
+                        apiEvent.ApiId,
+                        authenticationResult.AuthenticationResultMetadata.TokenSource,
+                        GetCacheLevel(authenticationResult),
+                        authenticationResult.AuthenticationResultMetadata.CacheRefreshReason,
+                        authenticationResult.AuthenticationResultMetadata.TelemetryTokenType,
+                        authenticationResult.ExpiresOn);
         }
 
         private void LogFailureTelemetryToOtel(string errorCodeToLog, ApiEvent apiEvent, CacheRefreshReason cacheRefreshReason, string rawStsErrorCode = null)
--- a/src/client/Microsoft.Identity.Client/Internal/TelemetryTokenTypeConstants.cs
+++ b/src/client/Microsoft.Identity.Client/Internal/TelemetryTokenTypeConstants.cs
@@ -21,5 +21,19 @@ namespace Microsoft.Identity.Client.Internal
         public const int Extension = 5;
 
         public const int MtlsPop = 6;
+
+        /// <summary>
+        /// Maps a telemetry token type integer to its human-readable string representation.
+        /// </summary>
+        internal static string ToDisplayString(int tokenType) => tokenType switch
+        {
+            Bearer => nameof(Bearer),
+            Pop => nameof(Pop),
+            SshCert => nameof(SshCert),
+            AtPop => "ExternalBoundToken",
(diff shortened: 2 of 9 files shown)


src/client/Microsoft.Identity.Client/Internal/TelemetryTokenTypeConstants.cs
```diff
@@ -21,5 +21,19 @@ internal static class TelemetryTokenTypeConstants
...
+
+        /// <summary>
+        /// Maps a telemetry token type integer to its human-readable string representation.
+        /// </summary>
+        internal static string ToDisplayString(int tokenType) => tokenType switch
+        {
+            Bearer => nameof(Bearer),
+            Pop => nameof(Pop),
+            SshCert => nameof(SshCert),
+            AtPop => "ExternalBoundToken",
```
Robbie-Microsoft: Would it be safer and more consistent to avoid this magic string?

Robbie-Microsoft: Could it be referenced somewhere else eventually?

src/client/Microsoft.Identity.Client/Internal/Requests/RequestBase.cs
```diff
@@ -131,6 +131,15 @@ private void LogSuccessTelemetryToOtel(AuthenticationResult authenticationResult
                         durationInUs,
                         authenticationResult.AuthenticationResultMetadata,
                         AuthenticationRequestParameters.RequestContext.Logger);
+
+            ServiceBundle.PlatformProxy.OtelInstrumentation.LogRemainingTokenLifetime(
+                        ServiceBundle.PlatformProxy.GetProductName(),
+                        apiEvent.ApiId,
+                        authenticationResult.AuthenticationResultMetadata.TokenSource,
+                        GetCacheLevel(authenticationResult),
```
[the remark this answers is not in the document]
Robbie-Microsoft: +1

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -218,5 +227,28 @@ public void LogFailureMetrics(string platform,
                         new(TelemetryConstants.TokenType, tokenType));
             }
         }
+
+        public void LogRemainingTokenLifetime(
```
ssmelov, line 250: Just to clarify, why isn’t this included in LogSuccessMetrics?

src/client/Microsoft.Identity.Client/Internal/TelemetryTokenTypeConstants.cs
Robbie-Microsoft: "ExternalBoundToken" should be a named constant, not a magic string

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
```diff
@@ -577,6 +577,24 @@ private void VerifyMetrics(int expectedMetricCount, List<Metric> exportedMetrics
...
+
+                    case "MsalRemainingTokenLifetime.1A":
+                        Trace.WriteLine("Verify the metrics captured for MsalRemainingTokenLifetime.1A histogram.");
+                        Assert.AreEqual(MetricType.Histogram, exportedItem.MetricType);
+
+                        expectedTags.Add(TelemetryConstants.MsalVersionPlatform);
+                        expectedTags.Add(TelemetryConstants.ApiId);
+                        expectedTags.Add(TelemetryConstants.TokenSource);
+                        expectedTags.Add(TelemetryConstants.CacheLevel);
+                        expectedTags.Add(TelemetryConstants.CacheRefreshReason);
```
Robbie-Microsoft, line 740: Test validates tags but never asserts the recorded histogram value

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -88,6 +89,14 @@ internal class OtelInstrumentation : IOtelInstrumentation
             unit: "us",
             description: "Performance of token acquisition calls extension latency."));
 
+        /// <summary>
+        /// Histogram to record the remaining lifetime of acquired tokens in seconds.
```
Robbie-Microsoft, line 94: XML doc says "seconds" but PR description says "minutes"

Robbie-Microsoft (commented):
PR description says tags are "6/8" but only 6 are emitted. The description mentions 8 planned tags - it's unclear which 2 are missing and whether that's intentional or an oversight worth clarifying.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
neha-bhargava: I will update the PR description but this should be logged in seconds to make sure we have some relevant data. Since this is a histogram the records are stored in buckets based on the value recorded

src/client/Microsoft.Identity.Client/Internal/TelemetryTokenTypeConstants.cs
neha-bhargava: Updated and no longer valid

tests/Microsoft.Identity.Test.Unit/TelemetryTests/OTelInstrumentationTests.cs
neha-bhargava: Updated the tests

neha-bhargava:
> s are "6/8" but only 6 are emitted. The description mentions 8 planned tags - it's unclear which 2 are missing and whether that's intentional or an oversight worth clarifying.

Its 6 out of 8. The otel have a constraint where the performance is impacted if more than 8 tags are added to the telemetry. So, this statement just means that we are within the limit.

bgavrilMS (approved):

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -172,6 +182,19 @@ public void LogSuccessMetrics(
                 new(TelemetryConstants.CacheLevel, authResultMetadata.CacheLevel),
                 new(TelemetryConstants.TokenType, authResultMetadata.TelemetryTokenType));
             }
+
+            if (s_remainingTokenLifetime.Value.Enabled)
```
ssmelov: Should we also record remaining-token-lifetime metric for background token fetches? (at SilentRequestHelper.cs)

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
```diff
@@ -172,6 +182,19 @@ public void LogSuccessMetrics(
...
+            {
+                long remainingSeconds = Math.Max(0, (long)(expiresOn - DateTimeOffset.UtcNow).TotalSeconds);
+
+                s_remainingTokenLifetime.Value.Record(remainingSeconds,
+                    new(TelemetryConstants.MsalVersionPlatform, $"{MsalIdHelper.GetMsalVersion()},{platform}"),
+                    new(TelemetryConstants.ApiId, apiId),
```
Robbie-Microsoft: Every other MSAL OTel histogram uses separate `MsalVersion` and `Platform` tags. This histogram combines them into a single `MsalVersionPlatform` tag, which means consumers can't filter or group by version or platform independently in their OTel backend. If the constraint is staying within the 8-tag limit, consider dropping `CacheLevel` instead — it overlaps heavily with `TokenSource` for this metric and is the most redundant of the six tags.

neha-bhargava: Cache level is not redundant but an important dimension. The platform and version tags are combined since this combination for an app generally remains the same.

src/client/Microsoft.Identity.Client/Platforms/Features/OpenTelemetry/OtelInstrumentation.cs
neha-bhargava: Done

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ProactiveRefreshTests.cs
```diff
@@ -92,7 +92,7 @@ public async Task ProactiveRefreshTriggers_WithTelemetry_Test()
 
         private long ValidateSuccessMetrics(MeterProvider meterProvider, List<Metric> exportedMetrics)
         {
-            Assert.HasCount(5, exportedMetrics);
+            Assert.HasCount(6, exportedMetrics);
```
ssmelov, line 95: nit (non-blocking): Could we assert on the histogram's dimensions? The existing HasCount(6) confirms the metric is emitted, and OTelInstrumentationTests covers tag presence on the LogSuccessMetrics call site, but neither would catch a regression where the background path passes the wrong tag values (e.g., CacheRefreshReason or TokenSource). Happy for this to be a follow-up.

ssmelov (approved):

tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/ProactiveRefreshTests.cs
neha-bhargava: Thanks! Will include this in future PR.

PR #5923, opened 2026-04-13
Split CI pipeline into independent stages for granular rerun on failure

gladjohn:
## Problem

The entire build and test pipeline currently runs as a single stage: `MSALBuildAndTest`.

When an integration test fails and the pipeline is restarted, all jobs, including unit tests that already passed, run again from scratch. This wastes approximately 10+ minutes of CI time per rerun.

## Solution

Split the pipeline into 6 independent stages so Azure DevOps can rerun only the failed stage.

| Stage | Jobs | Depends On |
|---|---|---|
| Build | `PreBuildCheck`, `BuildAndStageProjects`, `MacBuild` | — |
| Unit Tests | `RunDesktopUnitTests`, `BuildQualityChecks` | `Build` |
| Cache Compat Tests | `BuildAndRunCacheCompatTests` | `Build` |
| Desktop Integration Tests | `RunDesktopIntegrationTests` | `Build` |
| Linux Integration Tests | `BuildandRunIntegrationTestsOnLinux` | None — builds from source |
| Managed Identity E2E Tests | `Arc`, `IMDS`, `IMDS V2` | `Build` |

Stages 2–6 run in parallel after `Build` completes.

## Changes

- `template-build-and-run-all-tests.yaml`  
  Reworked from a flat `jobs:` list into 6 `stages:` with the correct `dependsOn` relationships.

- `template-desktop-test-setup.yaml` *(new)*  
  Added a shared setup template for desktop test jobs. This includes:
  - installing .NET 8
  - downloading the build artifact
  - copying test assets
  - installing Key Vault secrets

  This template is reused by both unit and integration test jobs.

- `pipeline-pullrequest.yaml` / `pipeline-ci.yaml`  
  Updated to consume the template at the stage level.

- `pipeline-releasebuild.yaml`  
  Updated similarly, and `PackAndSign` is now a separate stage that depends on `Build`.

## Notes

- Existing templates such as `template-run-all-tests.yaml` and `template-desktop-unit-and-automation.yaml` are preserved and remain in use by OneBranch pipelines.
- `BuildQualityChecks` now runs only in the **Unit Tests** stage. Previously, it observed combined unit and integration test coverage.
- Coverage thresholds may need to be rebaselined.

### DIFF
--- a/build/pipeline-ci.yaml
+++ b/build/pipeline-ci.yaml
@@ -11,6 +11,4 @@ variables:
   ConsolidateAppCenterTests: true
 #BUILD PHASE 
 stages:
-- stage: MSALBuildAndTest
-  jobs: #Build and stage projects
-  - template: template-build-and-run-all-tests.yaml
+- template: template-build-and-run-all-tests.yaml
--- a/build/pipeline-pullrequest.yaml
+++ b/build/pipeline-pullrequest.yaml
@@ -18,6 +18,4 @@ variables:
 #BUILD PHASE 
 
 stages:
-- stage: MSALBuildAndTest
-  jobs: #Build and stage projects
-  - template: template-build-and-run-all-tests.yaml
+- template: template-build-and-run-all-tests.yaml
--- a/build/pipeline-releasebuild.yaml
+++ b/build/pipeline-releasebuild.yaml
@@ -11,17 +11,16 @@ variables:
 
 #BUILD PHASE 
 stages:
-- stage: MSALBuildAndTest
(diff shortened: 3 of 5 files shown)


trwalke (approved):
While I generally agree with the approach, there is always a trade of between parallelization and pool resources. There was a point in which build jobs were slowing down because there weren't enough agents available to run all of the builds happening in our org. I would double check the amount of agents available to us to run parallel tasks in OneBranch and normal ADO builds.

Avery-Dunn (approved):

PR #5928, opened 2026-04-15
[Proposal] Add token acquisition metrics that cover both successful and failed t…

ssmelov:
Add token acquisition metrics that cover both successful and failed token acquisition outcomes and close an observability gap by recording IDP network latency for background refresh.

### DIFF
--- a/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
+++ b/src/client/Microsoft.Identity.Client/Http/HttpManager.cs
@@ -254,15 +254,23 @@ namespace Microsoft.Identity.Client.Http
 
                 HttpClient client = GetHttpClient(bindingCertificate, validateServerCert);
 
-                using (HttpResponseMessage responseMessage =
-                    await client.SendAsync(requestMessage, cancellationToken).ConfigureAwait(false))
+                try
+                {
+                    using (HttpResponseMessage responseMessage =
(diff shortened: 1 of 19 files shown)


src/client/Microsoft.Identity.Client/Http/HttpManager.cs
```diff
@@ -254,15 +254,23 @@ private async Task<HttpResponse> ExecuteAsync(
...
-                    returnValue.UserAgent = requestMessage.Headers.UserAgent.ToString();
-                    return returnValue;
+                        HttpResponse returnValue = await CreateResponseAsync(responseMessage).ConfigureAwait(false);
+                        returnValue.UserAgent = requestMessage.Headers.UserAgent.ToString();
+                        return returnValue;
+                    }
+                }
+                catch
+                {
+                    LastRequestDurationInMs = sw.ElapsedMilliseconds;
```
neha-bhargava, line 272: finally block might be a good place to place the duration here that will cover all the cases. Here and in the RequestBase.cs as well

ssmelov: I'd keep it scoped to the network‑bound portion — putting it in finally would also include response body read.

neha-bhargava:
This work is completed in another PR

PR #5937, opened 2026-04-19
Add CorrelationId to AssertionRequestOptions for FIC 

### LINKED ISSUE #5924: [Feature Request] Support passing CorrelationId in AssertionRequestOptions for CCA and Managed Identity token requests (FIC)
### MSAL client type

Managed identity

### Problem statement

# Key objectives:

## Add CorrelationId property to AssertionRequestOptions (as Guid type)

- Make it settable and retrievable
- Update internal constructor to accept correlation ID from CCA
- Propagate correlation ID across both token acquisition legs

## Testing 
- When CCA passes a correlation ID → it flows to MI first-leg acquisition
- Ensures coherent logging and tracing across the two-leg flow

## Docs
- Update documentation and tests

### Proposed solution

_No response_

### Alternatives

_No response_

gladjohn:
Fixes #5924
 
 ## Summary
 
 Adds a `CorrelationId` (`Guid`) property to `AssertionRequestOptions` so that the CCA's correlation ID flows into assertion callbacks. This enables coherent end-to-end tracing across both legs of FIC token exchange flows when the callback implementer forwards the ID to 
downstream MI/CCA requests.
 
 ## Changes
 
 ### Production
 - **`AssertionRequestOptions.cs`** — new `public Guid CorrelationId` property + updated internal constructor
 - **6 construction sites** — propagate `CorrelationId` from `RequestContext` or `CommonParameters`:
   - `ClientAssertionDelegateCredential`
   - `ClientAssertionStringDelegateCredential`
   - `ClientCredentialRequest` (OnMsalServiceFailure + OnCompletion)
   - `MtlsPopParametersInitializer`
   - `CertificateAndClaimsClientCredential`
 - **PublicAPI.Unshipped.txt** — updated for all 6 target frameworks
 
 ### Tests
 - Updated `AssertionInputIsMutable` to cover new property
 - Updated `SignedAssertionDelegateClientCredential_NoClaims` to assert CorrelationId is populated
 - **2 new unit tests** verifying user-provided CorrelationId flows to both delegate types (string + `ClientSignedAssertion`)
 - **2 E2E tests updated** (`Sni_AssertionFlow`, `AgentGetsAppTokenForGraph`) to verify CorrelationId in FIC assertion callbacks
 
 ## Usage
 
 ```csharp
 .WithClientAssertion(async (AssertionRequestOptions options, CancellationToken ct) =>
 {
     // options.CorrelationId now carries the CCA's correlation ID
     var miResult = await miApp
         .AcquireTokenForManagedIdentity("api://AzureADTokenExchange")
         .WithCorrelationId(options.CorrelationId) // propagate for tracing
         .ExecuteAsync(ct);
 
     return new ClientSignedAssertion { Assertion = miResult.AccessToken };
 })

### DIFF
--- a/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/MtlsPopParametersInitializer.cs
+++ b/src/client/Microsoft.Identity.Client/ApiConfig/Parameters/MtlsPopParametersInitializer.cs
@@ -124,6 +124,7 @@ namespace Microsoft.Identity.Client.ApiConfig.Parameters
                 Claims = p.Claims,
                 CancellationToken = ct,
                 ClientAssertionFmiPath = p.ClientAssertionFmiPath,
+                CorrelationId = p.CorrelationId,
 
                 // Best-effort context. IMPORTANT: use AbsoluteUri, not Uri.Authority (host only).
                 TokenEndpoint = serviceBundle.Config.Authority.AuthorityInfo.CanonicalAuthority.AbsoluteUri
--- a/src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs
+++ b/src/client/Microsoft.Identity.Client/AppConfig/AssertionRequestOptions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
 using System.Collections.Generic;
 using System.Threading;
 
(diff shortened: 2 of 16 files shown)


trwalke (approved):
LGTM

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/MtlsPopParametersInitializer.cs
```diff
@@ -124,6 +124,7 @@ private static AssertionRequestOptions CreateAssertionRequestOptions(
                 Claims = p.Claims,
                 CancellationToken = ct,
                 ClientAssertionFmiPath = p.ClientAssertionFmiPath,
+                CorrelationId = p.CorrelationId,
```
Robbie-Microsoft, line 127: Nit: The other 5 construction sites source CorrelationId from requestContext.CorrelationId, but here it comes from p.CorrelationId (a CommonParameters field). These should be equivalent in
 practice, but worth confirming that p.CorrelationId is always populated when MSAL auto-generates a correlation ID (not just when the caller uses WithCorrelationId()). If there's a case where 
the auto-generated ID is set on RequestContext after CommonParameters is built, p.CorrelationId could be Guid.Empty while requestContext.CorrelationId would have the real value.

Robbie-Microsoft (approved):

src/client/Microsoft.Identity.Client/ApiConfig/Parameters/MtlsPopParametersInitializer.cs
gladjohn: Good observation! `ValidateAndCalculateApiId()` always runs before this code), and it sets `CommonParameters.CorrelationId` to either the user-provided value or `Guid.NewGuid()`. The **RequestContext** is created after **TryInitMtlsPopParametersAsync** returns, so it doesn't exist yet at this call site. Both ultimately read from the same CommonParameters.CorrelationId value, **so they're always equivalent**.

PR #5949, opened 2026-04-24
Fix HttpClient object leak described in #5948 (no socket exhaustion)

### LINKED ISSUE #5948: [Bug] The default HttpClient factory creates HttpClient objects on each request (but does not use them)
### Library version used

4.74.0

### .NET version

all

### Scenario

SimpleHttpClientFactory has a bug where it creates HttpClient objects on each request. 

```csharp
            return s_httpClientPool.GetOrAdd("non_mtls", CreateHttpClient());
```

Actual: 

- GetOrAdd is called and both arguments are evaluated. CreateHttpClient() is invoked and returns a **new** HttpClient
- The inner logic will notice a cache hit, so it will pull the old HttpClient from the cache

Expected: 

- Do not call CreateHttpClient() as it is mildly expensive

Note: Because the new HttpClient is never used, the socket exhaustion classical problem of HttpClient will not manifest. The new HttpClient will be garbage collected.

### Relevant code snippets

```csharp

```

### Expected behavior

_No response_

### Identity provider

### Regression

_No response_

### Solution and workarounds

_No response_

bgavrilMS:
Fixes #5948 

Fix a leak of HttpCLient objects from improper use of ConcurrentDictionary


gladjohn:
Nit: PR title references #5984 but should be #5948.

gladjohn (approved):

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -24,8 +25,12 @@ internal class SimpleHttpClientFactory : IMsalMtlsHttpClientFactory, IMsalSFHttp
         private static readonly ConcurrentDictionary<string, HttpClient> s_httpClientPool = new ConcurrentDictionary<string, HttpClient>();
         private static readonly object s_cacheLock = new object();
 
+        // referenced in unit tests
+        internal static int HttpClientCreationCount;
+
         private static HttpClient CreateHttpClient()
         {
+            Interlocked.Increment(ref HttpClientCreationCount);
```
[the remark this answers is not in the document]
bgavrilMS: @copilot apply changes based on [this feedback](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5949#discussion_r3138176229)

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -88,6 +93,13 @@ private static void CheckAndManageCache()
             }
         }
 
+        // referenced in unit tests
+        internal static void ResetForTest()
+        {
+            s_httpClientPool.Clear();
+            HttpClientCreationCount = 0;
+        }
```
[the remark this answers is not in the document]
bgavrilMS: @copilot apply changes based on [this feedback](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5949#discussion_r3138176259)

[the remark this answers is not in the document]
bgavrilMS: @copilot apply changes based on [this feedback](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5949#discussion_r3138176278)

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -88,6 +95,17 @@ private static void CheckAndManageCache()
...
 
+        // referenced in unit tests
+        internal static void ResetStaticStateForTest()
+        {
+            foreach (var client in s_httpClientPool.Values)
+            {
+                client.Dispose();
+            }
+            s_httpClientPool.Clear();
+            s_httpClientCreationCount = 0;
```
Robbie-Microsoft: The counter reset uses a plain assignment while the increment uses Interlocked.Increment — inconsistent and not thread-safe. Should be:
```csharp
Interlocked.Exchange(ref s_httpClientCreationCount, 0);
```

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -88,6 +95,17 @@ private static void CheckAndManageCache()
             }
         }
 
+        // referenced in unit tests
+        internal static void ResetStaticStateForTest()
+        {
+            foreach (var client in s_httpClientPool.Values)
```
Robbie-Microsoft: ResetStaticStateForTest disposes and clears the pool without holding s_cacheLock, while CheckAndManageCache does lock. If a test ever exercises concurrent code paths, there's a race between the clear here and the cache management in CheckAndManageCache. The whole body should be wrapped in lock (s_cacheLock).

Robbie-Microsoft:
The Service Fabric overload still has the unfixed eager GetOrAdd pattern:

```csharp
// GetHttpClient(Func<...> validateServerCert) — line ~111
return s_httpClientPool.GetOrAdd(key, new HttpClient(handler));
```

But there's a deeper problem here: the key is handler.GetHashCode().ToString(). Since GetHashCode() on a reference type returns the object's identity hash, and a new HttpClientHandler is created on every call, the key is **always unique** — the pool never actually hits. Every call adds a new HttpClient to the pool permanently (until the 1000-entry eviction), which is worse than the bug being fixed.

The comment above it even acknowledges this: *"The server cert rotates so we need a new HttpClient for each call."* If that's the intent, the GetOrAdd should be removed entirely and the method should just return new HttpClient(handler) directly, skipping the pool.

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpClientFactoryTests.cs
```diff
@@ -77,5 +77,28 @@ public void TestHttpClientWithMtlsCertificateAndCustomHandler()
...
+            // Before the fix, GetOrAdd(key, CreateHttpClient()) eagerly evaluates
+            // CreateHttpClient() on every call, causing unnecessary throwaway
+            // HttpClient/HttpClientHandler allocations.
+            Assert.AreEqual(1, created,
+                $"CreateHttpClient was called {created} times for 3 lookups. " +
+                "Use GetOrAdd(key, factory_delegate) to avoid creating throwaway HttpClient instances.");
+        }
+
     }
 }
```
Robbie-Microsoft, line 163: The regression test only covers GetHttpClient() (the "non_mtls" key). GetHttpClient(X509Certificate2) received the same factory-delegate fix but has no corresponding test. Consider adding a parallel test calling GetHttpClient(cert) multiple times with the same cert and asserting the HttpClientCreationCount only increments once.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -63,7 +70,7 @@ private static HttpClient CreateMtlsHttpClient(X509Certificate2 bindingCertifica
 
         public HttpClient GetHttpClient()
         {
-            return s_httpClientPool.GetOrAdd("non_mtls", CreateHttpClient());
+            return s_httpClientPool.GetOrAdd("non_mtls", _ => CreateHttpClient());
```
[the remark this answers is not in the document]
bgavrilMS: There shouldn't be that many HttpClient even under contention.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -74,7 +81,7 @@ public HttpClient GetHttpClient(X509Certificate2 x509Certificate2)
             }
 
             string key = x509Certificate2.Thumbprint;
-            return s_httpClientPool.GetOrAdd(key, CreateMtlsHttpClient(x509Certificate2));
+            return s_httpClientPool.GetOrAdd(key, _ => CreateMtlsHttpClient(x509Certificate2));
```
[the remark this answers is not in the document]
bgavrilMS: @copilot - implement this.

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpClientFactoryTests.cs
```diff
@@ -77,5 +77,28 @@ public void TestHttpClientWithMtlsCertificateAndCustomHandler()
...
+        public void TestGetHttpClient_DoesNotLeakHttpClients()
+        {
+            // Arrange - reset static state so we start from a clean pool
+            SimpleHttpClientFactory.ResetStaticStateForTest();
+            var factory = new SimpleHttpClientFactory();
+
+            // Act - call GetHttpClient multiple times with the same (default) key
+            factory.GetHttpClient();
+            factory.GetHttpClient();
+            factory.GetHttpClient();
```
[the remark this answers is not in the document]
bgavrilMS: @copilot - add a test for this.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -107,8 +136,7 @@ public HttpClient GetHttpClient(Func<HttpRequestMessage, X509Certificate2, X509C
                 }
             };
 
-            string key = handler.GetHashCode().ToString();
-            return s_httpClientPool.GetOrAdd(key, new HttpClient(handler));
+            return new HttpClient(handler);
```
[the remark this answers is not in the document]
bgavrilMS: The expensive stuff is not the creation of a new HttpClient, it's the creation of the handler? And that code hasn't changed.

Service Fabric scenario should not be HTTP heavy. What is your suggestion?

gladjohn (approved):
clean fix for the eager GetOrAdd leak, Lazy<HttpClient> with ExecutionAndPublication is the right call

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -99,16 +128,31 @@ public HttpClient GetHttpClient(Func<HttpRequestMessage, X509Certificate2, X509C
...
 
 #if NET471_OR_GREATER || NETSTANDARD || NET
-            var handler = new HttpClientHandler
-            {
-                ServerCertificateCustomValidationCallback = (message, cert, chain, sslPolicyErrors) =>
+            // Build a stable key from the delegate's method and target identity.
+            // This correctly caches clients when the same method on the same object instance
+            // is used as the callback (the common Service Fabric case), while still creating
+            // distinct clients for genuinely different callbacks.
+            string key = "sf_" +
```
bgavrilMS: @copilot - this is too complicated. Please revert everything related to Service Fabric

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -88,6 +101,22 @@ private static void CheckAndManageCache()
```
Robbie-Microsoft, line 87: `ResetStaticStateForTest` disposes `Lazy<HttpClient>` values before clearing the pool, but this method doesn't. When the pool hits the threshold in production, all existing `HttpClient` instances are abandoned to the finalizer. The same disposal loop from `ResetStaticStateForTest` belongs here.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -88,6 +101,22 @@ private static void CheckAndManageCache()
             }
         }
 
+        // referenced in unit tests
+        internal static void ResetStaticStateForTest()
+        {
```
Robbie-Microsoft, line 106: `ResetStaticStateForTest` and `HttpClientCreationCount` are test-only hooks in the production assembly with no guard. `ResetStaticStateForTest` actively disposes and clears production state — accidental non-test usage would be undetected. Is bare `internal` acceptable here, or should these be behind `#if DEBUG` / a test-only partial class?

tests/Microsoft.Identity.Test.Unit/CoreTests/HttpTests/HttpClientFactoryTests.cs
```diff
@@ -77,5 +79,85 @@ public void TestHttpClientWithMtlsCertificateAndCustomHandler()
...
+            var factory = new SimpleHttpClientFactory();
+
+            const int threadCount = 20;
+            var tasks = new List<Task<HttpClient>>(threadCount);
+
+            // Act - call GetHttpClient() concurrently from many threads at once
+            for (int i = 0; i < threadCount; i++)
+            {
+                tasks.Add(Task.Run(() => factory.GetHttpClient()));
+            }
```
Robbie-Microsoft, line 142: `Task.Run` doesn't guarantee the tasks actually race. On a serializing runner or under load, these 20 tasks may execute sequentially — and the test would pass even without the `Lazy` fix. Use a `Barrier` or `ManualResetEventSlim` to synchronize all threads immediately before they call `GetHttpClient()` so contention is guaranteed.

src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/SimpleHttpClientFactory.cs
```diff
@@ -63,7 +72,9 @@ private static HttpClient CreateMtlsHttpClient(X509Certificate2 bindingCertifica
 
         public HttpClient GetHttpClient()
         {
-            return s_httpClientPool.GetOrAdd("non_mtls", CreateHttpClient());
+            return s_httpClientPool.GetOrAdd(
+                "non_mtls",
+                _ => new Lazy<HttpClient>(CreateHttpClient, LazyThreadSafetyMode.ExecutionAndPublication)).Value;
         }
```
Robbie-Microsoft, line 78: Worth a comment here: `ConcurrentDictionary.GetOrAdd` with a factory delegate can still invoke the factory multiple times under contention — multiple threads may each create a competing `Lazy<HttpClient>` wrapper, but only one is stored. The 'losing' `Lazy` instances are never `.Value`-accessed so no `HttpClient` is constructed for them. The `ExecutionAndPublication` mode then guarantees only one `HttpClient` per *stored* `Lazy`. These are two separate race scenarios — the concurrent test only validates the second one (`LazyThreadSafetyMode`), not the first (`GetOrAdd` factory contention).

trwalke (approved):

